PR #159, opened 2014-11-06
NMS-6814: Documentation template for configuration files

indigo423:
- Migrate pull request to OpenNMS devel instead development/docs


indigo423:
Won't fix: #278 is the first example.

PR #166, opened 2014-11-13
added jvm metrics for FileDescriptors; added related graphs NMS-7172

DerTak:
http://issues.opennms.org/browse/NMS-7172
http://bamboo.internal.opennms.com:8085/browse/OPENNMS-NMS162


agalue:
DerTak, I've added a bunch of enhancements for the JMX Collector in order to deal with storeByGroup. The default settings are not the best one, as it is going to use the objectname (instead of the mbean's name, which can be used) as the name of the RRD. So, the workaround for non-JMX collectors is easy (create a different group), but in this case, another mbean will have the same objectname, which is going to confuse the default behavior of storeByGroup.

The solution should be, use the option that I made to use the mbean's name the default, and then create an upgrade task to rename all the JRBs if storeByGroup is enabled. If that sounds good, create a Jira issue for it.

PR #193, opened 2014-12-17
PJSM-253 Error when instantiating bean jaxbXml

mvrueden:
* Jira: http://issues.opennms.org/browse/PJSM-253
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-NMS190

Do not merge, still work in progress.


indigo423:
@mvrueden is this still work in progress?

RangerRick:
This is worth doing, but we should hold off doing anything until post-15.0.0.

PR #210, opened 2015-01-16
NMS-7333: Switches values retrieved incorrectly in the BSF notification strategy 

dschlenk:
I couldn't find any unit tests nor did I write any for this four line change. I did however verify that the change makes the BSFNotificationStrategy actually work. 

JIRA: http://issues.opennms.org/browse/NMS-7333

Todo:
- [ ] Code review
- [ ] Pass tests
- [ ] Documentation
- [ ] Close JIRA issue


dschlenk:
If someone wants to cherry pick this to 14 if there's another point release that would be swell too.

PR #212, opened 2015-01-21
NMS-7277: Fix WMI instance name

dschlenk:
Naming of WMI resources is now similar to how it used to be, which makes instances of things like wmiPhysicalDisk and wmiLogicalDisk  work again. 

JIRA: http://issues.opennms.org/browse/NMS-7277

Todo:
- [ ] Test WMI stats in 1.12.9 then upgrade to 14.0.4
- [ ] Code review
- [ ] Pass unit tests
- [ ] Merge to develop
- [ ] Merge to release candidate for 14.0.4
- [ ] Close JIRA issue


indigo423:
@dschlenk Can you please verify if I've understood the issue correct and does your fix resolve the JIRA issue I've assigned to this pull request?
Thank you very much tackling this issue. I've set the target for the pull request to version 14.0.4.

dschlenk:
@indigo423 Yes it does resolve that issue although I haven't written any upgrade routines to deal with changed names. However, my guess is that with this patch the names are the same as they used to be, but I haven't tested that. Basically need to: test getting WMI collecting stats for disks in ~1.12.9, then upgrade to  release-14.0.4 with this applied and verify that the resource names stay the same. I'll report back once I've done that.

RangerRick:
We're probably not doing another 14 release, instead we're working on getting 15 out.  I've merged this patch (I rebased it to be one commit from you) to foundation which is our ancestor for develop, so it should merge forward there shortly.

PR #216, opened 2015-01-31
Xmp monitor.java

rdkgit:
Update to XmpMonitor.java so that authenUser can be taken from xmp-config.xml and collectd service entry.


rdkgit:
Not sure how to only commit XmpMonitor.java from this branch.  Learning git.  Sorry.

indigo423:
This pull request is reapplied by #217 and #218

PR #220, opened 2015-02-02
NMS-7394: Outage records are not getting written to the database

j-white:
-Removes any duplicate outages and enforce uniqueness with a
partial index.
-Prevents the poller daemon from creating duplicate outages in cases where events arrive out-of-order.

JIRA: http://issues.opennms.org/browse/NMS-7394

Todo:
- [ ] Review
- [ ] Tests
- [ ] Merge to develop, 15.1 and close JIRA issue


agalue:
My impression about Pollerd in 15 is that it is very slow compared with 14. Also, even with the changes of this pull request, there are still some unexpected behavior (more details on Jira).

j-white:
@agalue Do those problems occur on 15 without this patch? If so, let's address them separately.

agalue:
I'm going to re-do the tests without your patch again. The first time I did it, I was able to reproduce the original problem, but I can do it again just in case.

agalue:
I repeated the tests again this morning, and I take back my statement when I said Pollerd was working "slow". It is working as expected.

Here are the details of the tests that I did today:

1) I configured two exact CentOS 7 VMs: one with OpenNMS 15 installed through YUM-stable, and another one with OpenNMS 16-SNAPSHOT from jira/NMS-7394 through YUM-branches.

2) I've started OpenNMS on both VMs at the same time, and pushed the same requisition in both servers at the same time.

3) Then, I took down several nodes (one at a time), to see when both OpenNMS servers see the nodeDown (and the nodeUp), and in all cases (even on those with Path-Outages) the alarms came almost at the same time. So, there are no "delays" on the workflow of Pollerd.

I think the issue with the "Not Monitored" vs "100% Available" is not related with this pull-request, so I've created another Jira issue for it.

agalue:
I did a squash merge of the jira/NMS-7394 into develop on revision 82e2861dfc697b5b42ea4b0e35dc8685b528e5fb

PR #236, opened 2015-03-06
LTS-137 fixed.

fooker:
Backporting of documentation from develop to foundation.


fooker:
Fix for http://issues.opennms.org/browse/LTS-137

PR #238, opened 2015-03-10
LTS-126: Create JasperReport template guideline

indigo423:
- Created JasperReport layout guide line for page header and footer

JIRA: http://issues.opennms.org/browse/LTS-126

Todo: 
- [ ] Define style guide for content tables
- [ ] Define color palette for charts
- [ ] Merge to Horizon
- [ ] Merge to Meridian


indigo423:
alfred, do recheck OCA

mvrueden:
Hey @indigo423 , what is the status on this issue?

I think we should put it into foundation instead of develop, because it is a LTS issue.
In addition we should also provide an empty dummy jrxml template which is consistent to the described template guideline here.

indigo423:
It is not just an LTS issue. You need a template for both to make reports maintainable and allow to apply different styles. I agree that it would be good to have a dummy JRXML but was out of scope of this issue.

mvrueden:
@indigo423 yes, that why we should merge it to `foundation` and not to `develop`. It is then automatically merged to all the right branches (Meridian, Horizon, 16.0.4, etc.) and you do not have to worry about merging it to the right places.

indigo423:
Upgraded to latest develop branch and pull request has to be reopened.

PR #241, opened 2015-03-12
NMS-7551: DiskUsageMonitor never fails with a non existing disk

roskens:
I've added a unit test for DiskUsageMonitor with a few test cases that all pass.

I've enhanced DiskUsageMonitor with a new parameter (require-type) that you can use to flag whether ALL or  ANY matches on the disk name must pass.

JIRA: http://issues.opennms.org/browse/NMS-7551
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-NMS257

Todo:
- [x] Pass unit tests
- [x] Review code
- [ ] Merge aganinst the release-15.0.2 branch


indigo423:
I've added the Bamboo link to see if this build goes green, then we can review it. I've set also the release version to 15.0.2 in the JIRA issue.

PR #245, opened 2015-03-22
NMS-7579: Remove unnecessary output from opennms-doc modules

indigo423:
- Set source document name in guide-admin
- Set source document name in guide-development
- Set source document name in guide-doc
- Set source document name in guide-install
- Set source document name in guide-user
- Set source document name in releasenotes

Each maven module generates only one index.html and one index.pdf and all other HTML and PDF files are removed.

JIRA: http://issues.opennms.org/browse/NMS-7579

Todo:
- [x] Review for deploy and packaging
- [x] Merge to develop and close JIRA issue

### DIFF
--- a/opennms-doc/guide-admin/pom.xml
+++ b/opennms-doc/guide-admin/pom.xml
@@ -20,6 +20,13 @@
                     <skipAssembly>false</skipAssembly>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.asciidoctor</groupId>
+                <artifactId>asciidoctor-maven-plugin</artifactId>
+                <configuration>
+                    <sourceDocumentName>index.adoc</sourceDocumentName>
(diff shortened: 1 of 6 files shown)


RangerRick:
I don't know anything about how asciidoc works so if you say the plugin changes the behavior, I guess I can merge it for you, but I have no value to add as a reviewer of this.  :)

indigo423:
@RangerRick The biggest issue which I can't see right now, is there any impact for the deployment, because we have only one index.html and one index.pdf for each documentation module. There are no dedicated html and pdf files for every single .adoc file anymore. So it is really something about unwanted side effects by reducing the amount of unnecessary html/pdf files.

RangerRick:
Pretty sure the deployment stuff just does the directory wholesale, so it shouldn't care.

PR #234, opened 2015-03-05
Allow choice of BSFManager run type in BSFNotificationStrategy

dschlenk:
No unit tests exist for this class, but I did verify it works IRL. Also can be cherry picked back to just about any branch 14 or later.

JIRA: ?

Todo:
- [ ] JIRA issue with number
- [x] Pass unit tests (no unit tests here)
- [ ] Review
- [ ] Merge to develop

PR #249, opened 2015-04-06
Improve BSFMonitor and BSFNotificationStrategy performance

dschlenk:
BSF performance has always been terrible since scripts are compiled and executed fresh each time. These patches will allow BSF engines that cache to actually cache by making BSFManager instances long running. I've verified that it works by patching groovy's CachingGroovyEngine to store some cache hit/miss stats and taking thread dumps at various points in time. Drops the load on a host monitoring a lot of hosts each with many services monitored by a groovy script from ~3-5 to ~0.25-0.75.

### DIFF
--- a/opennms-services/src/main/java/org/opennms/netmgt/poller/monitors/BSFMonitor.java
+++ b/opennms-services/src/main/java/org/opennms/netmgt/poller/monitors/BSFMonitor.java
@@ -126,6 +126,7 @@ import org.opennms.netmgt.poller.PollStatus;
  *
  * @author <A HREF="mailto:jay@opennms.org">Jason Aras</A>
  * @author <A HREF="mailto:jeffg@opennms.org">Jeff Gehlbach</A>
+ * @author <A HREF="mailto:dschlenk@converge-one.com</A>
  * @author <A HREF="http://www.opennms.org">OpenNMS</A>
  */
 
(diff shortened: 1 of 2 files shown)


dschlenk:
Oh, this also obsoletes PR #234.

opennms-services/src/main/java/org/opennms/netmgt/poller/monitors/BSFMonitor.java
```diff
@@ -232,11 +257,30 @@ public PollStatus poll(MonitoredService svc, Map<String,Object> map) {
             pollStatus = PollStatus.unavailable(e.getMessage());
             LOG.warn("BSFMonitor poll for service '{}' failed with unexpected throwable: {}", svc.getSvcName(), e.getMessage(), e);
         } finally { 
-            bsfManager.terminate();
+            try{
```
soleger: If you want to be totally correct here, you should loop and put the try/catch inside the loop so that if any undeclareBean() call fails, it won't prevent attempts to undeclare subsequent beans.

dschlenk:
Indeed! I did it right in the notification version, forgot to fix the monitor code to work the same way. Testing a patch right now - I'll push it here when it's tested. Thanks @soleger!

dschlenk:
Tests were successful.

RangerRick:
I just set up a fresh branch to make sure everything's still OK with all the churn we've introduced lately.  If it passes, I'll merge to develop:

http://bamboo.internal.opennms.com:8085/browse/OPENNMS-NMS284

PR #260, opened 2015-04-29
NMS-7632: Added deprecation message for LinkD

fooker:
Added deprecation message for LinkD

JIRA: http://issues.opennms.org/browse/NMS-7632

Todo:
- [ ] Code review
- [ ] Pass unit tests
- [ ] Merge to develop and close JIRA issue


soleger:
There are two spelling errors in the message: "deprectaed" -> "deprecated" and "pleas" -> "please". Other than that, it looks good.

fooker:
Typos fixed.

fooker:
The according issue is:
http://issues.opennms.org/browse/NMS-7632

soleger:
I think that it is too much to edit the service-configuration.xml file since this will force all users who have edited the file to deal with an rpmnew file when they upgrade. Linkd is already marked as enabled="false" in service-configuration.xml and I think that's enough. Also, why are the edits to the maps configurations necessary? I think these deal with the SVG maps and that isn't related to Linkd.

fooker:
I don't care much about service-configuration.xml - I'm ok with reverting it.

The SVG map is backed by the tables populated by LinkD. As EnhancedLinkD is not able to populate them, I've already asked by mail if it's worth migrating. The outcome was to get rid of it as manual link provisioning in not a valid use case.

fooker:
I'v reverted the changes to service-configuration.xml as promised.
Looks like we are ready to merge.

PR #262, opened 2015-05-04
jira/NMS-7618

christianpape:
Please review and merge fix for this issue:
http://issues.opennms.org/browse/NMS-7618


pbrane:
Works for me but, I did have to create another user and login as that user then logout and log back in as admin to get the default landing page to change.  Must be a browser cache thing.

PR #271, opened 2015-06-01
NMS-7520: Remove linkd

fooker:
This completely removes the "old" linkd from the code and the users configuration as specified in 
JIRA: http://issues.opennms.org/browse/NMS-7520.
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS28-24

The following parts has been removed:
- linkd daemon
- linkd topology provider
- SVG map
- link provisioning rest interface
- link provisioning adapters
- linkd web interface sections
- according config files

Please review and merge.


agalue:
On the changelog file I can see that only the "main" Linkd tables are removed, and there are lots of other tables that must be removed as well, like vlantable, just for naming one of them. I mean, clean the code is good, but remove the unused fields and tables from the database is actually more useful (in my opinion).

Actually, speaking about VLAN information, there are customers that require the VLAN data and that's the main reason why they are not using EnLinkd at the moment (because EnLinkd doesn't populate the VLAN information on the DB, and if it does, it doesn't expose it like the old Linkd on the node's page).

We should have another issue on Jira to track the features that exist on Linkd that need to be ported to EnLinkd, and if they are not going to be ported, explain in detail why.

mvrueden:
@agalue :+1:

fooker:
@agalue thats exactly what I think, too. Is there any meta/toplevel/parent/story-issue for these missing things?

Can you give us some more details about the remaining tables?

agalue:
I think that in order to be 100% accurate, we should ask Antonio about it. He's the only one who knows all the details of Linkd and EnLinkd.

rssntn67:
eheh,
 Dustin why do you not call me?

Antonio

fooker:
We removed a whole bunch of tables, eneties and DAOs. From our point of view we are ready for merging again.

fooker:
Maybe we should recreate this pull request as a squash merged one, as most of the commits are not usefull on its own?

mvrueden:

PR #276, opened 2015-06-03
NMS-7713: Improve links for SLA categories on start page

mvrueden:
http://issues.opennms.org/browse/NMS-7713

### DIFF
--- a/opennms-webapp/src/main/webapp/includes/categories-box.jsp
+++ b/opennms-webapp/src/main/webapp/includes/categories-box.jsp
@@ -54,6 +54,21 @@
 	m_category_list = new CategoryList();
     }
 
+	// Creates a link to the rtc/category.jsp according to the selected outagesType.
+	public String createCategoriesOutageLink(HttpServletResponse response, Category category, String outagesType, String linkTitle, String linkText) {
+		if (category.getLastUpdated() != null) {
+			if (linkTitle == null) {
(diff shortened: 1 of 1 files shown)


opennms-webapp/src/main/webapp/includes/categories-box.jsp
```diff
@@ -54,6 +54,21 @@
 	m_category_list = new CategoryList();
     }
 
+	// Creates a link to the rtc/category.jsp according to the selected outagesType.
+	public String createCategoriesOutageLink(HttpServletResponse response, Category category, String outagesType, String linkTitle, String linkText) {
+		if (category.getLastUpdated() != null) {
+			if (linkTitle == null) {
+				return String.format("<a href=\"%s\">%s</a>",
```
fooker: You can use en empty title attribute if linktitle is null to simplify this.

mvrueden: The thing is, that a title for the 2nd and 3rd column is defined.
If I manually set the title of the link to EMPTY the link shows no title when mouse overing. 
If I do not set the title when mouse overing at least the title of the column is shown. 

To me the last behaviour is closer to the original one. 
That is why I allowed the title to be null.

mvrueden:
Approvied by Seth: `looks fine to me. merge!!`

PR #278, opened 2015-06-05
NMS-7712: Migrate documentation from opennms.properties

indigo423:
Migrated all documentation from opennms.properties to Admin guide. The described default values are set in the opennms.properties. For the reason the properties are used in several parts of OpenNMS they are ordered in sections which are reflected also in the documentation part.

JIRA: http://issues.opennms.org/browse/NMS-7712

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/opennms.properties
+++ b/opennms-base-assembly/src/main/filtered/etc/opennms.properties
@@ -1,346 +1,101 @@
-#
-# This properties file allows you to set global properties for the opennms
-# application.  These properties will be set as system properties at the
-# beginning of Java startup.  Any properties set on the Java command-line
-# with -Dproperty=value will be overridden by these values.
-#
 
-# ###### ICMP ######
-# OpenNMS provides three ICMP implementations. JICMP (legacy, IPv4-only),
-# JNA (supports both IPv4 and IPv6), and JICMP6.
-#
-# The JICMP implementation is what has traditionally been used in OpenNMS
-# since 1.0, uses JNI.  It requires you to install a separate package
-# (JICMP) which contains a shared library for interfacing with your system's
-# ICMP APIs.
-#
-# The JICMP6 core library is a version of the JICMP codebase which can speak
-# ICMPv6, instead of ICMPv4. The OpeNMS JICMP6 pinger supports both ICMPv4
-# and ICMPv6.  It delegates all ICMPv4 ping requests to the original JICMP
-# JniPinger above and uses the JICMP6 library for ICMPv6 packets.  This is
-# the default for OpenNMS 1.9.90 and up.
-#
-# Finally, the JNA implementation is written from the ground up to support
-# IPv4 and IPv6, and takes advantage of the JNA project's ability to access
-# native APIs without needing to distribute separate shared libraries.  It
-# is, however, not as performant as the JICMP6 pinger, so it is not
-# recommended unless you are in an environment which requires it.  This is
-# the default ICMP implementation used in the remote poller, since it does
-# not rely on any external native code to be installed outside of the JVM.
-#
-# To use the JNI ICMPv4 interface only, use the following property setting:
+### Documentation:
+### http://docs.opennms.org/OpenNMS/${project.version}/documentation/guide-admin/index.html#file-opennms-properties
+
+### General
+
+#org.opennms.security.disableLoginSuccessEvent=false
+opennms.bin.dir=${install.bin.dir}
+java.awt.headless=true
+#org.opennms.utils.propertiesCache.enableCheckFileModified=false
+
+### ICMP
+
 #org.opennms.netmgt.icmp.pingerClass=org.opennms.netmgt.icmp.jni.JniPinger
-#
-# To use the JNA ICMPv4/ICMPv6 implementation, use the following property:
 #org.opennms.netmgt.icmp.pingerClass=org.opennms.netmgt.icmp.jna.JnaPinger
-#
-# The default is set to use the JNI ICMPv4/ICMPv6 interface like so:
 #org.opennms.netmgt.icmp.pingerClass=org.opennms.netmgt.icmp.jni6.Jni6Pinger
-
-# By default, OpenNMS will start up if either ICMPv4 *or* ICMPv6 are
-# available and initialize properly.  If you wish to force IPv4 or IPv6
-# explicitly, set one or both of these properties to true.
-#
 #org.opennms.netmgt.icmp.requireV4=detect
 #org.opennms.netmgt.icmp.requireV6=detect
 
-# ###### SNMP ######
-# OpenNMS provides two different SNMP implementations.  JoeSNMP is the
-# original OpenNMS # SNMP Library and provides SNMP v1 and v2 support.
-# SNMP4J is a new 100% Java SNMP library # that provides support for SNMP
-# v1, v2 and v3.
-#
-# To enable the JoeSnmp library uncomment the following line.
+### SNMP
+
 #org.opennms.snmp.strategyClass=org.opennms.netmgt.snmp.joesnmp.JoeSnmpStrategy
-#
-# The SNMP4J library is currently the default.  But can also be set with
-# the following line.
-#
-# Default:
 #org.opennms.snmp.strategyClass=org.opennms.netmgt.snmp.snmp4j.Snmp4JStrategy
-#
-
-# The world is full of buggy SNMP agents.  We can work around many of their
-# quirks by extending various SMI classes from SNMP4J.  The following property
-# must be set in order for any of these workarounds to be enabled.
 org.snmp4j.smisyntaxes=opennms-snmp4j-smisyntaxes.properties
-
-# When debugging SNMP problems when using the SNMP4J library, it may be
-# helpful to receive runtime exceptions from SNMP4J. These exceptions almost
-# always indicate a problem with an SNMP agent. Any that we don't catch will
-# end up in output.log, so they're disabled by default, but they may provide
-# more information (albeit without timestamps) than the messages that SNMP4J
-# logs (see snmp4j.LogFactory)
 org.opennms.snmp.snmp4j.forwardRuntimeExceptions=false
-
-# By default, SNMP4J does not do any logging internally, but it's easy to
-# enable.  See log4j.properties to adjust log levels for these messages.
 snmp4j.LogFactory=org.snmp4j.log.Log4jLogFactory
-
-# Net-SNMP agents prior to release 5.4.1 on 64-bit platforms exhibit a bug
-# that causes the discovery of a node's interfaces to fail. A workaround has
-# been made that helps SNMP4J cope with this bug. If for some reason you need
-# to disable this workaround, comment out the following line.
 org.opennms.snmp.workarounds.allow64BitIpAddress=true
-
-# DrayTek Corporation, Router Model: Vigor2820 Series devices with a
-# sysObjectID of .1.3.6.1.4.1.7367 sometimes present objects from the
-# ipAddrTable with an instance ID of "0". To substitute a valid (though still
-# dodgy) value of "0.0.0.0" when we encounter one of these, set the following
-# property to true.
 org.opennms.snmp.workarounds.allowZeroLengthIpAddress=true
-
-# Some SNMP agents claim to support SNMPv2c, but either completely fail to
-# respond to GET-BULK requests or return bad responses to them under some
-# conditions. To disable the use of GET-BULK PDUs globally in the SNMP4J
-# strategy, set this property to true. This change will adversely affect the
-# efficiency of SNMP data collection from agents that properly support
-# GET-BULK.
 org.opennms.snmp.snmp4j.noGetBulk=false
-
-# Some buggy SNMP agents fail to exclude Counter64 objects from view when
-# responding to SNMPv1 requests (as mandated by RFC3584 § 4.2.2.1). To relax
-# handling of v1 responses to permit Counter64 varbinds rather than discarding
-# them as ill-formed (per the same RFC), set this property to true.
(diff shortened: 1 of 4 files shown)


soleger:
* Many of the options that were commented out are now uncommented. I would leave all options that were previously commented out as commented out in the new version to make the new config functionally identical.
* I would prefer to follow the pattern used by PostgreSQL where all options in the config are commented out and represent the default values. This may require code changes to ensure that default values are used for certain system properties.
* I would like to see all possible values documented inside the config file itself.
* I would like to see a link to the documentation inside the config file itself.

I'm making all of the recommendations at once since any change to opennms.properties requires manual merging by our support staff for every customer and I'd rather cluster all drastic changes to the file so that they all get done at once.

opennms-base-assembly/src/main/filtered/etc/opennms.properties
```diff
@@ -1,622 +1,126 @@
...
-# To use the JNA ICMPv4/ICMPv6 implementation, use the following property:
-#org.opennms.netmgt.icmp.pingerClass=org.opennms.netmgt.icmp.jna.JnaPinger
-#
-# The default is set to use the JNI ICMPv4/ICMPv6 interface like so:
-#org.opennms.netmgt.icmp.pingerClass=org.opennms.netmgt.icmp.jni6.Jni6Pinger
+# Section :: GENERAL
+opennms.bin.dir=${install.bin.dir}
+java.awt.headless=true
+org.opennms.dao.ipinterface.findByServiceType=select distinct ipIf from OnmsIpInterface as ipIf join ipIf.monitoredServices as monSvc where monSvc.serviceType.name = ?
+org.opennms.web.aclsEnabled=false
```
soleger: This should be moved to "Section :: WEB UI".

pbrane:
+1

indigo423:
@soleger I've worked in your suggestions regarding uncommented and commented settings and moved some parts around. I've dropped a mail to opennms-discuss and opennms-docs to get some feedback about this issue and tried to explain the motivations behind this change. Thanks for your feedback, I'll dig in the PostgreSQL pattern to get more familiar with it.

opennms-base-assembly/src/main/filtered/etc/opennms.properties
```diff
@@ -1,622 +1,179 @@
...
 org.eclipse.jetty.server.Request.maxFormKeys=2000
-
-###### JETTY HTTPS SUPPORT ######
-# Details: http://www.opennms.org/index.php/Standalone_HTTPS_with_Jetty
-# If you want Jetty to provide an HTTPS listener, this is the port to listen on
-# Note that setting this property does NOT disable the plain HTTP listener,
-# which is required by Rtcd to post realtime status updates.  If you do not
-# wish to allow unsecured HTTP access to the OpenNMS web UI, you must set
-# org.opennms.netmgt.jetty.host above or use firewall rules to accomplish this.
 #org.opennms.netmgt.jetty.https-port = 8443
```
soleger: The Jetty HTTPS stuff needs to be in its own section apart from normal Jetty properties. Also, some of this might be obsolete since we now use jetty.xml to run the server. Only the properties that are referenced in the default jetty.xml file should be necessary now.

opennms-base-assembly/src/main/filtered/etc/opennms.properties
```diff
@@ -1,622 +1,179 @@
...
-# NOTE: if you enable a trouble-ticketing plugin here, you must also edit
-# $OPENNMS_HOME/jetty-webapps/opennms/WEB-INF/configuration.properties
 opennms.ticketer.plugin=org.opennms.netmgt.ticketd.NullTicketerPlugin
-
-# Comment out the NullTicketerPlugin line above and uncomment the lines below
-# to enable OtrsTicketerPlugin
-# You will need to adjust alarmTroubleTicketLinkTemplate to suit your OTRS Base URL
-# Note that you will need to install the OpenNMS module for OTRS in your OTRS server
-
 #opennms.ticketer.plugin=org.opennms.netmgt.ticketer.otrs.OtrsTicketerPlugin
```
soleger: This section needs a one-line explanation that all of these properties are used to enable OTRS ticketing.

opennms-base-assembly/src/main/filtered/etc/opennms.properties
```diff
@@ -1,622 +1,179 @@
...
 #opennms.alarmTroubleTicketEnabled = true
 #opennms.alarmTroubleTicketLinkTemplate = <a href="http://localhost/otrs/index.pl?Action=AgentTicketZoom&TicketNumber=\${id}">\${id}</a>
-
-# Comment out the NullTicketerPlugin line above and uncomment the lines below
-# to enable RtTicketerPlugin
-# You will need to adjust alarmTroubleTicketLinkTemplate to suit your RT Base URL
-
 #opennms.ticketer.plugin=org.opennms.netmgt.ticketer.rt.RtTicketerPlugin
```
soleger: This section needs a one-line explanation that all of these properties are used to enable RT ticketing.

opennms-base-assembly/src/main/filtered/etc/opennms.properties
```diff
@@ -1,622 +1,179 @@
...
 #opennms.alarmTroubleTicketEnabled = true
 #opennms.alarmTroubleTicketLinkTemplate = <a href="http://localhost/Ticket/Display.html?id=\${id}">\${id}</a>
-
-# Comment out the NullTicketerPlugin line above and uncomment the lines below
-# to enable RemedyTicketerPlugin
-# You will need to adjust alarmTroubleTicketLinkTemplate to suit your Remedy Base URL
-# The following settings are for locale it_IT
-
 #opennms.ticketer.plugin=org.opennms.netmgt.ticketer.remedy.RemedyTicketerPlugin
```
soleger: This section needs a one-line explanation that all of these properties are used to enable Remedy ticketing.

opennms-base-assembly/src/main/filtered/etc/opennms.properties
```diff
@@ -1,622 +1,179 @@
...
-opennms.report.template.dir=${install.dir}/etc
-opennms.report.dir=${install.share.dir}/reports
-opennms.report.logo=${install.webapps.dir}/images/logo.gif
-ksc.default.graphsPerLine=1
+### Eventd IPC
 
-###### EVENTD IPC ######
-# The hostname or IP address of the OpenNMS server where events should be sent.
-# Default: 127.0.0.1
 #opennms.rtc.event.proxy.host=127.0.0.1
```
soleger: These should all be in the RTC IPC section.

opennms-base-assembly/src/main/filtered/etc/opennms.properties
```diff
@@ -1,622 +1,179 @@
...
+#org.opennms.web.graphs.engine=png
 
-# (Mapquest API)
-#gwt.geocoder.class=org.opennms.features.poller.remote.gwt.server.geocoding.MapquestGeocoder
+### Mapping and Geocoding
 
-# (OpenStreetMaps API)
+gwt.maptype=OpenLayers
+gwt.apikey=
+#gwt.geocoder.class=org.opennms.features.poller.remote.gwt.server.geocoding.GoogleMapsGeocoder
```
soleger: The geocoder section needs an explanation that you are choosing one implementation among the 3 options (Google, Nominatim, Mapquest).

opennms-base-assembly/src/main/filtered/etc/opennms.properties
```diff
@@ -1,622 +1,179 @@
...
-# not rely on any external native code to be installed outside of the JVM.
-# 
-# To use the JNI ICMPv4 interface only, use the following property setting:
+
+### Documentation:
+### http://docs.opennms.org/OpenNMS/${project.version}/documentation/guide-admin/index.html#file-opennms-properties
+
+### General
+
+#org.opennms.dao.ipinterface.findByServiceType=select distinct ipIf from OnmsIpInterface as ipIf join ipIf.monitoredServices as monSvc where monSvc.serviceType.name = ?
```
soleger: This property needs to be DELETED!!! Why is there HQL in opennms.properties? This must be for some custom behavior somewhere... but whoever needs this should fork their own repo and make the change there.

opennms-doc/guide-admin/src/asciidoc/text/config-files/opennms.properties.adoc
```diff
@@ -0,0 +1,920 @@
+
+[[file-opennms-properties]]
+=== File: opennms.properties
+
+This properties file allows you to set global properties for the opennms application.
+These properties will be set as system properties at the beginning of Java startup.
+Any properties set on the Java command-line with `-Dproperty=value` will be overridden by these values.
+
+[options="autowidth"]
```
soleger: I'm not sure I like autowidth here because some of the panels are not as wide, making it difficult to scan down the document and see the property names easily. I would prefer width=100%.

opennms-doc/guide-admin/src/asciidoc/text/config-files/opennms.properties.adoc
```diff
@@ -0,0 +1,920 @@
...
+| Type             | Java property file with key value pairs
+| Location         | `${OPENNMS_HOME}/etc`
+| Restart required | yes
+|===
+
+==== General
+
+[options="header, autowidth"]
+|===
+| Property | `opennms.bin.dir`
```
soleger: Is it also possible to left-justify the property names so that they all line up vertically on the page? I don't like the centering since it makes it difficult to scan for property names.

soleger: Also, is the 'Property' column necessary? People should know just from looking at the doc that this is a system property that is contained in opennms.properties. I think it just takes up space.

indigo423: The content is left-justified with rendering in our doc, the GitHubs preview is just a very simple way to represent the adoc file and tables. It's really just a preview. If you want to see the output generated with bamboo you can run a `mvn clean package` in the `guide-admin` directory.

indigo423: I have removed the 'Property' column, I agree about the space.

opennms-doc/guide-admin/src/asciidoc/text/config-files/opennms.properties.adoc
indigo423: This is right now a more complicated topic and should be moved to a different topic for the reason it implies all tables in the doc we use right now and has implication on the PDF rendered output.

opennms-base-assembly/src/main/filtered/etc/opennms.properties
```diff
@@ -1,622 +1,179 @@
...
+#org.opennms.netmgt.collectd.DefaultCollectionAgent.loadSnmpDataOnInit=false
```
indigo423: Regarding the existing documentation this property is related to `org.opennms.dao.ipinterface.findByServiceType`. Should this property also be deleted?
I can deal with the deletion from the docs, I can't deal with the "...but whoever needs this should fork their own repo and make the change there" :)

opennms-base-assembly/src/main/filtered/etc/opennms.properties
indigo423: Done

opennms-base-assembly/src/main/filtered/etc/opennms.properties
soleger: I meant: this property needs to be deleted and we need to remove usage of the property from the source code. I don't know its exact origin but it's not acceptable to have the ability to edit HQL statements by setting a system property.

indigo423: @soleger Is this something we can solve in this pull request? The topic is migrating the existing documentation. I would recommend to open an dedicated issue to remove this property. The only thing I can do is to remove the property from the doc. This is from my point of view the last open topic in this pull request. Thanks for taking care, helps a lot.

opennms-doc/guide-admin/src/asciidoc/text/config-files/_template.adoc
```diff
@@ -0,0 +1,23 @@
...
+// REMOVE ME!! Please keep first line an empty line to make sure, the ToC can be build correctly
+==== File: {name-of-the-configuration-file}
+
+A short summary what is the configuration file is doing
+
+[options="autowidth"]
+|===
+| Type             | e.g. Java property file with key value pairs or XML file with XSD
+| Location         | `path where the configuration file is located`
+| Restart required | yes or no
```
StCyr: >"restart required" 

It's not just yes or no, you can also have "send-event.pl"

opennms-doc/guide-admin/src/asciidoc/text/config-files/_template.adoc
```diff
@@ -0,0 +1,23 @@
+
+// REMOVE ME!! Please keep first line an empty line to make sure, the ToC can be build correctly
+==== File: {name-of-the-configuration-file}
```
StCyr: Also, for xml files, there's already a lot of documentation in the schema files (.xsd files).

Couldn't you setup some automation to get some info from these files right here?

mvrueden:

indigo423:
Undecided how to proceed, we keep the documentation in the current opennms.properties file.

j-white:
Was there any specific reason why we didn't proceed with this, or just a general of consensus?

indigo423:
@j-white To sum it up very briefly, I did a proposal to move properties to official versioned documentation and remove the docs from the properties files to not have to maintain in two places. The proposal was rejected with the result to keep the documentation in the properties files so an administrator has the bits and pieces to configure properties on hand when he is editing the file.

indigo423:
See also here: https://issues.opennms.org/browse/HZN-356

PR #288, opened 2015-06-18
HZN-329: Add box with OpenNMS related Grafana Dashboards

indigo423:
This pull request allows to add a box with quick access to OpenNMS releated Grafana Dashboards. It is required to create an Grafana API key for OpenNMS. The Dashboards can be filtered by using a Grafana Dashboard Tag.

JIRA: http://issues.opennms.org/browse/HZN-329

Todo:
- [x] Review code
- [x] Merge to develop for Horizon 17


opennms-webapp/src/main/webapp/includes/grafana-box.jsp
```diff
@@ -0,0 +1,132 @@
...
+        /**
+         * ...using the search API url.
+         */
+        HttpGet httpGet = new HttpGet("/api/search/");
+        httpGet.setHeader("Authorization", "Bearer " + grafanaApiKey);
+
+        /**
+         * Invoke the request to the Grafana host.
+         */
+        HttpResponse httpResponse = httpclient.execute(httpHost, httpGet);
```
mvrueden: This is pretty obvious. Do we need a comment for that?

opennms-webapp/src/main/webapp/includes/grafana-box.jsp
```diff
@@ -0,0 +1,132 @@
...
+
+        /**
+         * Convert the complete response into a string...
+         */
+        responseString = IOUtils.toString(httpResponse.getEntity().getContent(), "UTF-8");
```
mvrueden: This is pretty obvious. Do we need a comment for that?

opennms-webapp/src/main/webapp/includes/grafana-box.jsp
```diff
@@ -0,0 +1,132 @@
...
+    }
+%>
```
mvrueden: Are the statements so important, that they deserve a multi line comment?

opennms-webapp/src/main/webapp/includes/grafana-box.jsp
j-white: Is there any reason were making a synchronous to the Grafana API on the server-side instead of using an AJAX call? The whole index page won't load until we get a response back from the Grafana server.

opennms-webapp/src/main/webapp/includes/grafana-box.jsp
```diff
@@ -0,0 +1,132 @@
...
+
+<div id="heatmap-box" class="panel panel-default">
+    <div class="panel-heading">
+        <h3 class="panel-title">Grafana Dashboards</h3>
+    </div>
+    <div class="panel-body">
+        <script type="text/javascript" src="/opennms/js/jquery/jquery-1.8.2.min.js"></script>
```
j-white: index.jsp already include a copy of jquery in the header.

opennms-webapp/src/main/webapp/includes/grafana-box.jsp
```diff
@@ -0,0 +1,132 @@
...
+        <script type="text/javascript" src="/opennms/js/jquery/ui/jquery-ui-1.8.2.custom.js"></script>
```
j-white: Is there any reason we need jquery-ui here? There doesn't appear to be.

opennms-webapp/src/main/webapp/includes/grafana-box.jsp
```diff
@@ -0,0 +1,132 @@
...
+            var grafanaTag = '<%=grafanaTag%>';
+            var obj = <%=responseString%>;
+
+            jQuery.each(obj['dashboards'], function (i, val) {
+                var showDashboard = true;
+
+                if (grafanaTag != '') {
+                    showDashboard = false;
+                    for (var index in val['tags']) {
+                        console.log(val['tags'][index]);
```
j-white: Should be removed.

j-white: console.log line should be removed

opennms-webapp/src/main/webapp/includes/grafana-box.jsp
```diff
@@ -0,0 +1,132 @@
...
+                        if (grafanaTag == val['tags'][index]) {
+                            showDashboard = true;
+                            break;
+                        }
+                    }
+                }
+
+                if (showDashboard) {
+                    document.write('<a href="<%=grafanaProtocol%>://<%=grafanaHostname%>:<%=grafanaPort%>/dashboard/db/' + val['slug'] + '"><span class="glyphicon glyphicon-signal" aria-hidden="true"></span>&nbsp;' + val['title'] + "</a><br />");
```
j-white: The list of URLs should be rendered as a [list](http://getbootstrap.com/css/#unstyled) instead of using <br />. The former makes it easier to style or theme.

j-white: Also, we should probably append to a specific div instead of using document.write.

opennms-webapp/src/main/webapp/includes/grafana-box.jsp
```diff
@@ -0,0 +1,132 @@
...
+        HttpResponse httpResponse = httpclient.execute(httpHost, httpGet);
+
+        /**
+         * Convert the complete response into a string...
+         */
+        responseString = IOUtils.toString(httpResponse.getEntity().getContent(), "UTF-8");
+    }
+%>
+
+<div id="heatmap-box" class="panel panel-default">
```
j-white: heatmap?

opennms-webapp/src/main/webapp/includes/grafana-box.jsp
```diff
@@ -0,0 +1,130 @@
...
+            && !"".equals(grafanaProtocol)
+            && ("http".equals(grafanaProtocol) || "https".equals(grafanaProtocol))) {
+        try {
+            DefaultHttpClient httpClient = new DefaultHttpClient();
+            HttpParams httpParams = httpClient.getParams();
+            /**
+             * Setting the timeouts to assure that the landing page will be loaded. Making the
+             * call via JS isn't possible due to CORS-related problems with the Grafana server.
+             */
+            HttpConnectionParams.setConnectionTimeout(httpParams, 500);
```
j-white: Should the timeout be configurable?

opennms-webapp/src/main/webapp/includes/grafana-box.jsp
```diff
@@ -0,0 +1,130 @@
...
+            HttpConnectionParams.setSoTimeout(httpParams, 500);
+            HttpHost httpHost = new HttpHost(grafanaHostname, grafanaPort, grafanaProtocol);
+            HttpGet httpGet = new HttpGet("/api/search/");
+            httpGet.setHeader("Authorization", "Bearer " + grafanaApiKey);
+            HttpResponse httpResponse = httpClient.execute(httpHost, httpGet);
+            responseString = IOUtils.toString(httpResponse.getEntity().getContent(), "UTF-8");
+        } catch (UnknownHostException e) {
+            // do nothing
+        } catch (ConnectTimeoutException e) {
+            // do nothing
```
j-white: Can we update the message in the box to reflect that the connection timed out?

opennms-base-assembly/src/main/filtered/etc/opennms.properties
```diff
@@ -620,3 +620,27 @@ excludeServiceMonitorsFromRemotePoller=DHCP,NSClient,RadiusAuth,XMP
...
+#org.opennms.grafanaBox.apiKey=
+#
+# When a tag is specified only dashboards with this given tag
+# will be displayed. When no tag is given all dashboards will
+# be displayed.
+#org.opennms.grafanaBox.tag=
+#
+# The protocol for the ReST call can also be specified
+#org.opennms.grafanaBox.protocol=http
+
```
mvrueden: Currently there is a discussion going on whether to move properties-documentation to it's own doc module.
As Ronny is mostly driving this, can you please clear with him, if you need to create an issue or something for that?

opennms-webapp/src/main/webapp/includes/grafana-box.jsp
mvrueden: I'd also like to see some kind of error handling. Catching and not doing anything is usually a bad idea. At least log it somewhere. 

You may also try multi-catch:

    try {
        throw new SomeException("Frühstück ist fertig.");
    } catch (SomeException | SomeOtherException | AndAnotherOne e) {
        LOG.error(e);
    }

j-white:
Thanks for the updates Christian. I'll merge once Bamboo goes green.

opennms-base-assembly/src/main/filtered/etc/opennms.properties
```diff
@@ -620,3 +620,31 @@ excludeServiceMonitorsFromRemotePoller=DHCP,NSClient,RadiusAuth,XMP
...
+# will be displayed. When no tag is given all dashboards will
+# be displayed.
+#org.opennms.grafanaBox.tag=
+#
+# The protocol for the ReST call can also be specified
+#org.opennms.grafanaBox.protocol=http
+#
+# Timeouts for contacting the grafana server
+#org.opennms.grafanaBox.connectionTimeout=500
+#org.opennms.grafanaBox.soTimeout=500
```
indigo423: Do you know what the `soTimeout` means?

indigo423:
If you provide Grafana through a proxy you run in problems, cause the hostname in the property is also used to generate the links for the dashboard. So there should be a second variable for grafana.proxy.host which allows to give a different host name for creating the links.

opennms-base-assembly/src/main/filtered/etc/opennms.properties
christianpape: It's the timeout on the socket level. JavaDoc for this:

Enable/disable SO_TIMEOUT with the specified timeout, in milliseconds. With this option set to a non-zero timeout, a read() call on the InputStream associated with this Socket will block for only this amount of time. If the timeout expires, a java.net.SocketTimeoutException is raised, though the Socket is still valid. The option must be enabled prior to entering the blocking operation to have effect. The timeout must be > 0. A timeout of zero is interpreted as an infinite timeout.

mvrueden:

PR #301, opened 2015-06-27
NMS-5016: Add Dell Force10 Events

mfuhrmann:
JIRA: http://issues.opennms.org/browse/NMS-5016
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS55/latest


agalue:
I'm just curious ;) which tool was used to generate the event definitions ?

PR #320, opened 2015-07-14
NMS-7798: Add Sonicwall Events

mfuhrmann:
JIRA: http://issues.opennms.org/browse/NMS-7798
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS95


soleger:
This definition is waaaaay too verbose. Each trap appears to have a single varbind and the list of possible values is several hundred values, each the same for every trap.

For each trap type, I would imagine that very few of the varbind values are possible. The list of hundreds of values needs to be reduced to the values that can actually occur for each trap.

mfuhrmann:
I will try to clean up the file.

soleger:
Can you change the UEI prefix from:
```
uei.opennms.org/traps/sonicwall
```
to:
```
uei.opennms.org/vendor/SonicWall
```
A majority of the traps use uei.opennms.org/vendor instead of uei.opennms.org/traps.

soleger:
When this is merged, it MUST be merged manually as a squash merge to avoid committing the original, enormous file into the main OpenNMS repo.

dschlenk:
FWIW typically vendor traps have uei.opennms.org/vendor/<VendorName>/traps as the prefix so as to distinguish from other event sources (syslog, etc).

indigo423:
Cleaned up vendor names and recreated pull request with a single commit solved in #352.

PR #325, opened 2015-07-21
NMS-4549: Add Santera Events

mfuhrmann:
I did not change eventconf.xml because I don't know if it is too much for including them per default.

JIRA: http://issues.opennms.org/browse/NMS-4549
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS414-1


indigo423:
What, I miss is the include in the eventconf.xml?

indigo423:
Bamboo failed for the reason we have the event files but they are not included in the eventconf.xml

http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS94-UNIT-1/test/case/128340924

mfuhrmann:
If ONMS can handle it without any problems, we should merge it. But I'm not a ONMS developer. I don't know if it is too big.
The feature list will look nice with it: "We added ~4000 events for Santera devices" :-)

dschlenk:
maybe throw it in the examples directory?

indigo423:
@dschlenk I like this idea. What do you think @mfuhrmann.

indigo423:
See comment in JIRA issue

PR #330, opened 2015-07-22
High Speed utilization graph should use ifHighSpeed instead of ifSpeed

jccomputing:
The graph uses ifSpeed and give incorrect results when the interface speed is above 4Gbps.

This PR switches to ifHighSpeed.


opennms-bamboo:
Welcome @jccomputing.

I could not find a signed OpenNMS Contributor Agreement (OCA) assigned to the username @jccomputing.

jeffgdotorg:
Thanks for the fix, @jccomputing. Can't believe that one got missed for so long...

PR #329, opened 2015-07-22
HZN-375: Added option to specifiy nameserver to DnsResolutionMonitor

fooker:
Fixes http://issues.opennms.org/browse/HZN-375.

Please review and merge.

Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS96

PR #331, opened 2015-07-23
HZN-375: Added option to specify nameserver to DnsResolutionMonitor

fooker:
Issue: http://issues.opennms.org/browse/HZN-375.
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS96

This is the same as #329 but against develop.

### DIFF
--- a/opennms-services/src/main/java/org/opennms/netmgt/poller/monitors/DNSResolutionMonitor.java
+++ b/opennms-services/src/main/java/org/opennms/netmgt/poller/monitors/DNSResolutionMonitor.java
@@ -28,115 +28,113 @@
 
 package org.opennms.netmgt.poller.monitors;
 
-import java.net.Inet4Address;
-import java.net.Inet6Address;
-import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.Arrays;
 import java.util.Map;
 
 import org.opennms.core.utils.ParameterMap;
 import org.opennms.netmgt.poller.Distributable;
 import org.opennms.netmgt.poller.MonitoredService;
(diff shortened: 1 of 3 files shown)


soleger:
I would prefer that the reason codes were changed like I mentioned in the last PR. The case where no result is returned is covered by the "Unable to resolve..." message in the old code. Am I reading this incorrectly?

fooker:
I'Ve changed the reason codes again. Do you think it's better, now?

opennms-services/src/main/java/org/opennms/netmgt/poller/monitors/DNSResolutionMonitor.java
```diff
@@ -28,115 +28,113 @@
...
+
+        // Resolving succeeded - checking results
+        final long end = System.currentTimeMillis();
+
+        // Check if result is valid
+        if (!ipv4Found && !ipv6Found) {
+            return PollStatus.unavailable("Unable to resolve host '" + name + "'");
+
+        } else  if (ipv4Required && !ipv4Found) {
+            return PollStatus.unavailable("No IPv4 Address result (A record) returned for '" + name + "'");
```
soleger: In the old monitor, this was:
```java
name + " could only be resolved to an IPv6 address"
```
indicating that an AAAA record was found but no A record was found. This is not apparent in the new message. Can we make the messages for this something like:
```java
name + " could be resolved to an IPv6 address (AAAA record) but not an IPv4 address (A record)."
```

fooker:
Ok, now I got what you are talking about - sorry. Does this look better, now?

PR #354, opened 2015-08-24
HZN-185: Review fixes for Debian installation guide

StCyr:
Please make sure the URLs are correct: You are using opennms.eu here while you are using opennms.org for RHEL-based systems

### DIFF
--- a/opennms-doc/guide-install/src/asciidoc/text/opennms/introduction.adoc
+++ b/opennms-doc/guide-install/src/asciidoc/text/opennms/introduction.adoc
@@ -72,19 +72,24 @@ TIP: Verify the release of _OpenNMS_ packages with `yum info opennms`.
 
 ==== Specific Release on Debian-based system
 
+Create a new apt source file (eg: `/etc/apt/sources.list.d/opennms.list`), and add the following 2 lines:
+
+.Package repository configuration for Debian-based systems
 [source, shell]
 ----
-vi /etc/apt/sources.list.d/opennms.list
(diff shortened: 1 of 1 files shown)


opennms-doc/guide-install/src/asciidoc/text/opennms/introduction.adoc
```diff
@@ -72,20 +72,25 @@ TIP: Verify the release of _OpenNMS_ packages with `yum info opennms`.
...
+.Package repository configuration for Debian-based systems
 [source, shell]
 ----
-vi /etc/apt/sources.list.d/opennms.list
+deb http://debian.opennms.eu ${RELEASE} main <1>
+deb-src http://debian.opennms.eu ${RELEASE} main <1>
 ----
 
-.Package repository configuration for RHEL-based systems
+<1> Replace `${RELEASE}` with a release name like `testing` or `snapshot`.
```
mvrueden: It would be nice to have all available possibilities here or at a referenced article.

mvrueden:
AFAIK the opennms.eu is a mirror of opennms.org which means, both should work.
It might be useful to add a HINT message to point that out. 
@indigo423  please correct me, if I am wrong :)

mvrueden:
It also seems that ulf is down :-/

mvrueden:
Ulf, can you please redo OCA check.

indigo423:
Using different mirrors is a different topic. This description should use the minimal default which is opennms.org.

PR #357, opened 2015-09-02
Jira/HZN-452: Use the Measurement API as a datasource in JasperReports.

mvrueden:
* Jira: http://issues.opennms.org/browse/HZN-452
* Bamboo:  http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS152


indigo423:
The links to official documentation doesn't work. They go to branches/development instead of 17.0.0-SNAPSHOT.

mvrueden:
@indigo423 I know. The issue is that I either have to hard reference them, but then they don't work anymore when released. So I decided to use a broken link in develop and have it working when released. In addition I opened an issue to adress this problem (http://issues.opennms.org/browse/NMS-7873).

PR #359, opened 2015-09-09
HZN-388: Added deletion support for resources

fooker:
Please review and merge:

Issue: http://issues.opennms.org/browse/HZN-388
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS157


j-white:
Looks good, we just need a solution when ttl=0.

Note that this won't pass in bamboo until https://github.com/OpenNMS/newts/pull/28 is merged though.

PR #360, opened 2015-09-10
jira/NMS-7844

mvrueden:
Jira: http://issues.opennms.org/browse/NMS-7844
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS161

The target branch is jira/HZN-452 and not develop, because a lot of changes have been made there regarding the jasper reports.


mvrueden:
Bamboo is failing, but that is expected, as almost all tests should be fixed in HZN-452.
Going to merge, even if Bamboo is red.

PR #377, opened 2015-10-07
Revert "NMS-7820: VerticalLabel fix for many graphs"

mvrueden:
Reverts OpenNMS/opennms#376


mvrueden:
I accidently merged to develop, but it should go to foundation instead. We also should label the labels correctly instead of label them "Count" by default.

PR #379, opened 2015-10-07
Fixed #HZN-466.

fooker:
Seperated the resource-cli tool into the contrib package and excluded
the second slf4j implementation from the main build.

JIRA: http://issues.opennms.org/browse/HZN-466
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS189


mvrueden:
Hey @fooker, looks good so far.
Unfortunately I found a couple of issues:
 * The shell script to execute the resourcecli tool is not working, because it does not find the slf4j* libraries located in `lib`. We probably do not need those anymore, because we ship a `jar-with-dependencies`
 * The contrib package is not available by default. You have to install `opennms-contrib`. Therefore we should create a seperate folder `lib/tools` which is always available, or also add the `bin/resourcecli.sh` to not be there by default, otherwise the jars would be missing.

fooker:
Now I'm using the ```jar-with-dependencies``` avoiding the whole wrapper script.

j-white:
Why do we even need slf4j-simple for this binary anyhow? Can't we just use log4j-slf4j-impl?

Also, try and avoid using  ```jar-with-dependencies```, since it increases the size of packages and the footprint on disk.

fooker:
As the tool is a pure command line tool, the logger is used to communicate with the user. If we want to switch to ```log4j-slf4j``` we have to provide an alternative config which does the same as ```slf4-simple```.

The reason to use ```jar-with-dependencies``` is that the tools in ```/conrib``` can run outside of an opennms environment.

indigo423:
Pull request name convention:

```
"HZN/NMS-<number>: subject"

detailed description detailed description detailed description detailed description
detailed description detailed description detailed description detailed description
detailed description detailed description detailed description detailed description
``

**Comment by fooker** [comment]
Usefull comment convention:
```
Specify whats your oppinnion and provide usefull hints hwo to solve the problem.

rababarabarbara blafasel rababarabarbara blafasel rababarabarbara blafasel
rababarabarbara blafasel rababarabarbara blafasel rababarabarbara blafasel
rababarabarbara blafasel rababarabarbara blafasel rababarabarbara blafasel
```

mvrueden:
I agree with @fooker. 
The idea is to have a very easy to use cli-tool.
The cli-tool may depend on libraries, which are not by default in the lib folder of OpenNMS or may not be placed there to avoid adding dependencies to the OpenNMS classpath which are only dependencies of the cli tool itself.

j-white:
In general yes, but in this case, the same behavior can be achieved using the existing libraries.

The core RPM increased in size by 24MB (from 341MB to 365MB) by allowing this particular utility to be run on it's own.

fooker:
Back to the old initial bundling aproach...

I've removed the dependency to slf4j-simple and don't build a fat-jar any more.

PR #383, opened 2015-10-09
Fix for issue NMS-7917

christianpape:
Fix for issue NMS-7917: In the case an Alarm Dashlet's filter criteria returned an empty set of alarmIds the restriction was not correctly applied to the AlarmRepository instance. Now the empty set will be replaced by a set containing only the integer zero (0), so no alarms will match which is the correct expected behavior.

Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS194
Jira: http://issues.opennms.org/browse/NMS-7917


christianpape:
Closed because this fix must be made for foundation 
and not for the develop branch. Please see PR #384 
for the correct fix.

PR #385, opened 2015-10-09
HZN-389: New event file for APC's PowerNet MIB

StCyr:
This is not yet ready to be merged: It still misses the severity of the 391 new events for example.
But at least, you can already easily have a look at my work.

JIRA: http://issues.opennms.org/browse/HZN-389
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS321


StCyr:
Ok,

so, I've done with the severity rating. 

As a rule of the thumb, I generally assigned severity (for the newly created events):

* "Normal" or "Cleared" when APC says the trap is "INFORMATIONAL"
* "Warning" when they say "WARNING"
* "Minor" when they say "SEVERE", and
* "Major" when they say "FAILURE", "CRITICAL" or "MAJOR"

though I might have not respected this mapping from time to time.

Now, I've just learned that I should not use severity "Cleared" for clearing events. So, I guess I'll to correct that.

StCyr:
Hi, I've done configuring all new events.

I'm now testing this new file on my server, and it looks OK.

Though, I think it would be good to review my work since this change adds about 200 alarms to the ~2000 already existing alarm configurations. Maybe some aren't worth adding.

I think it would be good if some other users could test this event file on their OpenNMS server to see if their alarm dashboards don't get flooded by new uninteresting alarms.

Rgds,

Cyrille

StCyr:
closing

rebased the PR against release-17.0.1

See https://github.com/OpenNMS/opennms/pull/543

PR #387, opened 2015-10-12
HZN-459: Evaluate possibilities to defer between ifHcInOctets and ifInOctets

christianpape:
HZN-459: Evaluate possibilities to defer between ifHcInOctets and ifInOctets
Introduced an fallback-attribute parameter in the Source model to allow to query for 
a secondary attribute if the primary one does not exist.

Jira: http://issues.opennms.org/browse/HZN-459
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS195-3

### DIFF
--- a/features/measurements/api/src/main/java/org/opennms/netmgt/measurements/model/Source.java
+++ b/features/measurements/api/src/main/java/org/opennms/netmgt/measurements/model/Source.java
@@ -64,6 +64,12 @@ public class Source {
     private String attribute;
 
     /**
+     * fallback Attribute name.
+     * i.e. ifInOctets
+     */
+    private String fallbackAttribute;
+
+    /**
      * Data source name. Typically the same as the attribute name, but may differ
      * if an attribute contains multiple data sources.
      *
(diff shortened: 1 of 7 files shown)


indigo423:
Fix fox und pull request convention?

indigo423:
Pull request name convention:

```
"HZN/NMS-<number>: subject"

detailed description detailed description detailed description detailed description
detailed description detailed description detailed description detailed description
detailed description detailed description detailed description detailed description
```

j-white:
Looks good. I think we also need to update the report templates to that use ifInOctets/ifOutOctets to leverage this though.

christianpape:
Hi, I think it is a good idea to create an additional issue for the modification 
of the reports because I've a very limited knowledge about the report stuff and
its creation.

mvrueden:
Hey @christianpape as we already discussed, please create that issue or assign me issue `HZN-459` then I will implement this functionality.

christianpape:
Here we go...
http://issues.opennms.org/browse/HZN-483

PR #398, opened 2015-10-18
NMS-7750: Fix for Cannot edit some Asset Info fields

muratyildiz1976:
- All asset info where transferred over the network correctly 
- The error occurred in the processing logic in the backend due to incompatibility of attributes of AssetCommand and OnmsAssetRecord classes while using BeanUtils.copyProperties method
- Fixed the issue by copying those lost attributes in a seperate step


mvrueden:
The pull request must created against foundation. Closing.

PR #402, opened 2015-10-19
HZN-453: Do not make REST Calls when running within OpenNMS

mvrueden:
* Jira: http://issues.opennms.org/browse/HZN-453
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS210


features/measurements/api/src/main/java/org/opennms/netmgt/measurements/api/MeasurementsService.java
```diff
@@ -0,0 +1,114 @@
...
+public class MeasurementsService {
+
+    private final MeasurementFetchStrategy fetchStrategy;
+    private final ExpressionEngine expressionEngine;
+    private final FilterEngine filterEngine;
+    private final QueryRequestValidator queryRequestValidator = new QueryRequestValidator();
+
+    @Autowired
+    public MeasurementsService(MeasurementFetchStrategy fetchStrategy, ExpressionEngine expressionEngine, FilterEngine filterEngine) {
+        Preconditions.checkNotNull(fetchStrategy);
```
j-white: checkNotNull returns the object, so these lines can be rewritten in the form of:
this.fetchStrategy = Preconditions.checkNotNull(fetchStrategy);

j-white:
Looks great! I'd say it's good to merge once Bamboo goes green.

features/measurements/api/src/main/java/org/opennms/netmgt/measurements/api/MeasurementsService.java
mvrueden: Hey @j-white, good catch :)

PR #400, opened 2015-10-19
NMS-7750: Fix for Cannot edit some Asset Info fields

muratyildiz1976:
- All asset info where transferred correctly
- The error occurred in the processing logic in the backend due to incompatibility of attributes of AssetCommand and OnmsAssetRecord classes while using BeanUtils.copyProperties method
- Fixed the issue by copying those lost attributes in a seperate step
- link: http://issues.opennms.org/browse/NMS-7750

PR #408, opened 2015-10-21
 NMS-7750: Fix for Cannot edit some Asset Info fields #400 

mvrueden:
* JIRA: http://issues.opennms.org/browse/NMS-7750
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS207


mvrueden:
The original pull request was created by Murat and can be found here: https://github.com/OpenNMS/opennms/pull/400.

This pull request contains a already made fix to use JRE 7 instead of JRE 8 for the jmxconfiguration-theme.

PR #410, opened 2015-10-22
HZN-503: Fixed ClassNotFound error for resourcecli command

fooker:
Simplified resourcecli script to avoid clumpsy classpath building.

JIRA: http://issues.opennms.org/browse/HZN-503
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS219


j-white:
Looks good. This fixes the problem on my test setup.

Good to merge once bamboo goes green.

PR #413, opened 2015-10-24
NMS-7820: Vertical Label Fix

mfuhrmann:
JIRA: http://issues.opennms.org/browse/NMS-7820


opennms-base-assembly/src/main/filtered/etc/snmp-graph.properties.d/opennms-graph.properties
```diff
@@ -213,7 +213,7 @@ report.OpenNMS.Capsd.ActiveSuspecThreads.AttributeReport.name=ActiveSuspectThrea
 report.OpenNMS.Capsd.ActiveSuspecThreads.AttributeReport.columns=ActiveSuspecThreads
 report.OpenNMS.Capsd.ActiveSuspecThreads.AttributeReport.type=interfaceSnmp
 report.OpenNMS.Capsd.ActiveSuspecThreads.AttributeReport.command=--title="ActiveSuspectThreads" \
```
indigo423: We should cleanup the title as well.

indigo423:
I would suggest to remove the Capsd related metrics, cause it doesn't exist anymore. Any objections?

soleger:
The Capsd definitions cannot be removed in foundation, only in OpenNMS 16 and above. I think that a lot of that work was already done in this issue: http://issues.opennms.org/browse/NMS-7649. Please revert the Capsd changes.

indigo423:
@soleger thank you for update

soleger:
Please do not merge these changes right now, the code for Meridian 2015.1.1 is effectively frozen. These fixes will have to go in after that release.

j-white:
Can we rebase these changes over the foundation-2016 branch instead? 

The capsd and mx4j related graphs are gone, so the diff should be much smaller now :)

agalue:
I'm wondering why we have things with "Capsd" on their name, considering that Capsd is gone

indigo423:
Cleaned up and updated to latest develop branch. This is PR is replaced with #534.

PR #423, opened 2015-10-27
Revert "NMS-7407: Inconsistent naming in Admin/System Information"

indigo423:
Reverts OpenNMS/opennms#422


indigo423:
Smoke test fails: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS228-SMOKE-1/test/case/139307129

PR #406, opened 2015-10-21
HZN-478: Use measurement API in statsd instead of old RRD DAOs

fooker:
Statsd failed to work with NewTS as it's using the RRD DAO directly and tries to access the (non existing) RRD Files on disk.

This changes the resource visitors used by statsd to use the measurement API instead of the RRD DOA to gether the statistic values.

Issue: http://issues.opennms.org/browse/HZN-478
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS212-2

PR #427, opened 2015-10-28
HZN-478: Use measurement API in statsd instead of old RRD DAOs

fooker:
Statsd failed to work with NewTS as it's using the RRD DAO directly and tries to access the (non existing) RRD Files on disk.

This changes the resource visitors used by statsd to use the measurement API instead of the RRD DOA to gether the statistic values.

Old pull-request: #406

Issue: http://issues.opennms.org/browse/HZN-478
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS212


j-white:
Thanks for the update, looks much better now :)

Let's merge this once Bamboo goes green.

PR #428, opened 2015-10-30
BSM-55: Initial BusinessService model.

j-white:
Initial work on modeling business services and making them readable via a new page in the Admin UI.

JIRA: http://issues.opennms.org/browse/BSM-55
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS234/latest


opennms-base-assembly/src/main/filtered/etc/create.sql
```diff
@@ -2638,3 +2638,21 @@ CREATE TABLE subcomponents (
...
+    CONSTRAINT bsm_services_pkey PRIMARY KEY (id)
+);
+
+CREATE TABLE bsm_service_attributes (
+    bsm_service_id integer NOT NULL,
+    key character varying(255) NOT NULL,
+    value TEXT NOT NULL,
+    CONSTRAINT bsm_service_attributes_pkey PRIMARY KEY (bsm_service_id, key)
+);
+
```
indigo423: @soleger Do we actually maintain the create.sql file?

soleger: Yes it is used as the basis of the database used in integration tests. It is not used by the OpenNMS packages (RPMs, DEBs, etc), the database is managed by Liquibase there. We use create.sql in the integration tests because performing the Liquibase migrations for every test would add hours to the Bamboo builds.

indigo423:
thx for explaining.

features/bsm/model/src/main/resources/changelog.xml
```diff
@@ -0,0 +1,44 @@
...
+		</createTable>
+
+		<createTable tableName="bsm_service_attributes" >
+
+			<column name="bsm_service_id" type="integer">
+			  <constraints nullable="false"/>
+			</column>
+
+			<column name="key" type="varchar(255)">
+			  <constraints nullable="false"/>
```
mvrueden: We probably need to add a unique constraint here.

features/bsm/rest/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+  <properties>
+    <skipITs>true</skipITs>
+  </properties>
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-failsafe-plugin</artifactId>
+        <version>2.16</version>
+        <configuration>
```
mvrueden: We should really define this in the root pom. Otherwise we have to change ALL plugin definitions if we at some point need to change the version.

opennms-webapp/src/main/webapp/admin/bsm/main.html
```diff
@@ -0,0 +1,44 @@
...
+					</th>
+					<th class="col-sm-5">
+						Name
+					</th>
+					<th class="col-sm-6">
+						Attributes
+					</th>
+				</tr>
+				<tr ng-repeat="item in items">
+					<td>
```
mvrueden: Please also add a (No business services set) blog, if no business services are defined yet.

mvrueden:
Hey Jesse, this looks really good. 
Please have a look at my comments.
If you have any questions, please let me know.

features/bsm/model/src/main/resources/changelog.xml
j-white: We define a compound primary key bellow.

mvrueden: I missed that :+1:

mvrueden:
Thanks @j-white for your changes. As soon as bamboo goes green, it can be merged.

PR #430, opened 2015-10-30
NMS-7899: Log a thread dump and active queries when a deadlock is detected.

j-white:
This should help us isolate the deadlocks reported in NMS-7899.

JIRA: http://issues.opennms.org/browse/NMS-7899
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS237

### DIFF
--- a/opennms-dao-api/src/main/java/org/opennms/netmgt/dao/util/DeadLockDebugger.java
+++ b/opennms-dao-api/src/main/java/org/opennms/netmgt/dao/util/DeadLockDebugger.java
@@ -1,0 +1,159 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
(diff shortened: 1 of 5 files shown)


opennms-dao-api/src/main/java/org/opennms/netmgt/dao/util/DeadLockDebugger.java
```diff
@@ -0,0 +1,159 @@
...
+                for (int i = 1; i <= metaData.getColumnCount(); i++) {
+                    if (i == 1)
+                        sb.append("|");
+                    sb.append(set.getString(i) + "|");
+                }
+                sb.append("\n");
+            }
+        } finally {
+            try {
+                set.close();
```
mvrueden: set can be null here.

PR #431, opened 2015-10-31
NMS-7949: Promote Compass app when mobile browser detected

jeffgdotorg:
JIRA: http://issues.opennms.org/browse/NMS-7949
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS241


jeffgdotorg:
Reworked iOS stuff, thanks @RangerRick. No equivalent seems to exist for Google Play so left that one as originally implemented.

PR #434, opened 2015-11-04
HZN-514: Make OTRS Integration plugin support arbitrary SoapAction HTTP header

StCyr:
JIRA: http://issues.opennms.org/browse/HZN-514
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS246

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/otrs.properties
+++ b/opennms-base-assembly/src/main/filtered/etc/otrs.properties
@@ -52,6 +52,9 @@
 # Typical OTRS v 3.1.x and above endpoint
 #otrs.endpoint=http://localhost/otrs/nph-genericinterface.pl/Webservice/GenericTicketConnector
 
+# OTRS v 3.1.x and above require a soapAction HTTP header
+otrs.soapaction=http://www.otrs.org/TicketConnector/
+
 #####################################################################
 #
 # User configuration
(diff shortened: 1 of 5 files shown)


opennms-base-assembly/src/main/filtered/etc/otrs.properties
```diff
@@ -52,6 +52,9 @@
 # Typical OTRS v 3.1.x and above endpoint
 #otrs.endpoint=http://localhost/otrs/nph-genericinterface.pl/Webservice/GenericTicketConnector
 
+# OTRS v 3.1.x and above require a soapAction HTTP header
+otrs.soapaction=http://www.otrs.org/TicketConnector/
```
indigo423: Is this something which should be commented in by default like the `ours.endpoint` property?

indigo423:
@StCyr thank you very much for fixing this issue. I've added your branch to bamboo and it went green. I'll set the label for a code review.

opennms-base-assembly/src/main/filtered/etc/otrs.properties
StCyr: I don't think so. At least, it doesn't hurt to leave it uncommented.

I guess the `otrs.endpoint` property is commented because there are 2 alternatives and one _must_ choose one otherwise integration won't work.

On the other hand, `otrs.soapaction` will only be used when the 3.1+ integration is used, and will be ignored if the 3.1- integration is used.

StCyr:
I've reviewed the bamboo failure

It doesn't seem caused by my patch...

mvrueden:
@StCyr I do not think so either. I gave bamboo a kick and it should turn green.

mvrueden:
It seems that something is wrong in bamboo. @RangerRick, can you please have a look at this?

StCyr:
This PR sucks:

soapAction HTTP header is set by the annotations in the org.otrs.ticketconnector java class.

And, much other things (like: validation of the webservice's response) depend on these annotations. See for example http://issues.opennms.org/browse/HZN-518

So, the proper implementation should be something that is able to modify annotations at runtime. It seems reflection allows that (see https://stackoverflow.com/questions/29656252/communicating-to-a-soap-webservice-with-constantly-changing-targetnamespace).

Also, this ticket's title should be renamed to something like "OTRS integration doesn't support arbitrary target namespace".

Only the "http://www.otrs.org/TicketConnector/" target namespace is supported atm.

fooker:
As this was not merged in v17.0.0 please open another pull request against branch ```release-17.0.1```.

StCyr:
This PR sucks, so I'll not create a new one against 17.0.1 or any other OpenNMS releases.

Proper implementation should allow modification of targetNamespace at runtime.

PR #441, opened 2015-11-05
HZN-508: Update WHATSNEW with OpenNMS 17 changes since Aug 11

indigo423:
Added credits to community member and added major improvements in documentation.

JIRA: http://issues.opennms.org/browse/HZN-508


indigo423:
Thank you @soleger

PR #448, opened 2015-11-09
Make RMI Bind to Localhost By Default

RangerRick:
This pull implements a custom socket factory to coerce the RMI server into binding only to localhost by default.  Additionally, it provides a default karaf configuration that uses `localhost` for RMI server and client connections.

You can override the bind port by setting the `opennms.poller.server.serverHost` property in `$OPENNMS_HOME/etc/opennms.conf`:

```
# old behavior, listen on all interfaces
ADDITIONAL_MANAGER_OPTIONS="-Dopennms.poller.server.serverHost=0.0.0.0"

# bind to a specific IP
ADDITIONAL_MANAGER_OPTIONS="-Dopennms.poller.server.serverHost=192.168.0.1"
```


RangerRick:
Alternatively, you can set opennms.poller.server.serverHost= in `opennms.properties`.  I didn't add it there by default because of Meridian's requirement not to change config files.  I'll add it to opennms.properties in 16.0.4 after this is merged.

j-white:
Looks good to me. Waiting on Bamboo.

PR #451, opened 2015-11-11
NMS-7975: Change status code to 204 instead of 404

mvrueden:
* JIra: http://issues.opennms.org/browse/NMS-7975
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS266

### DIFF
--- a/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/AbstractDaoRestService.java
+++ b/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/AbstractDaoRestService.java
@@ -139,7 +139,7 @@ public abstract class AbstractDaoRestService<T,K extends Serializable> {
 		final List<T> coll = getDao().findMatching(crit);
 
 		if (coll == null || coll.size() < 1) {
-			return Response.status(Status.NOT_FOUND).build();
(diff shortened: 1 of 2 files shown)


j-white:
Can you confirm that the 'Manage Monitoring Locations' page still works as expected with these changes? I believe it's the thing we have that currently using v2 of the REST API.

I think this line in particular might need to be updated:
    https://github.com/OpenNMS/opennms/blob/develop/opennms-webapp/src/main/webapp/js/angular-onmsList-monitoringLocation.js#L106

mvrueden:
I found three implementer of the current v2 API:
 * https://github.com/OpenNMS/opennms/blob/develop/opennms-webapp/src/main/webapp/js/angular-onmsList-monitoringLocation.js
 * https://github.com/OpenNMS/opennms/blob/develop/opennms-webapp/src/main/webapp/js/angular-onmsList-notification.js
 * https://github.com/OpenNMS/opennms/blob/develop/opennms-webapp/src/main/webapp/js/angular-onmsList-minion.js

I think only the "Manage Minions" and "Manage Monitore Locations" are used at the moment, as the new notification page is not linked by default (index.jsp vs indexNew.jsp).
However, all of these work with the new API without making any change to the client side angularjs code.

PR #454, opened 2015-11-13
NMS-7985: Change mirrors to DigitalOcean global caches

indigo423:
Changed install guide to point to new Digital Ocean global mirrors on yum.mirrors.opennms.org and debian.mirrors.opennms.org.


indigo423:
Change target branch to release 17

PR #461, opened 2015-11-18
BSM-56: Create/Update/Delete Business Services

DerTak:
Jira: http://issues.opennms.org/browse/BSM-56
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS263


smoke-test/src/test/java/org/opennms/smoketest/BSMAdminIT.java
```diff
@@ -28,70 +28,49 @@
...
-
-        // Verify that the Business Service appears in the list
-        m_driver.get(BASE_URL + "opennms/admin/bsm/index.jsp");
-        findElementByXpath("//*[contains(text(), '" + servicePrefix + "-name')]");
-
-        // The Business Service's attributes should also be displayed
-        WebElement el = findElementByXpath("//*[contains(text(), '" + servicePrefix + "-key')]");
-        WebElement parent = el.findElement(By.xpath(".."));
-        assertEquals(servicePrefix + "-key: " + servicePrefix + "-value", parent.getText());
+    @Test
```
j-white: These tests assume that they are executed in the order defined, but JUnit does not provide this guarantee.

We should either rework the logic or rename the tests and use the @FixMethodOrder annotation.

opennms-webapp/src/main/webapp/admin/bsm/js/controllers/BusinessServicesListController.js
```diff
@@ -0,0 +1,67 @@
...
+
+            $scope.bsCreate = function () {
+                if ($scope.name.length > 0) {
+                    $scope.newBS = new BusinessServices({name: $scope.name}).$save()
+                        .then(function (newBS) {
+                            listBS();
+                            $scope.name = "";
+                        })
+                        .catch(function (req) {
+                            $log.warn("create for new BusinessService failed: " + req);
```
mvrueden: please remove the `for`

smoke-test/src/test/java/org/opennms/smoketest/BSMAdminIT.java
```diff
@@ -28,70 +28,49 @@
...
 
 public class BSMAdminIT extends OpenNMSSeleniumTestCase {
 
-    @Test
-    public void canReadBusinessServices() throws ClientProtocolException, IOException {
-        // Deferring this check until we can delete entries via the UI
-//        // Verify that a message is presented when there are no business services defined
-//        m_driver.get(BASE_URL + "opennms/admin/bsm/index.jsp");
-//        findElementByXpath("//*[contains(text(), 'There are no business services defined.')]");
+    public static final String RENAMED_SERVICE_NAME = "renamed_service";
```
mvrueden: Why is this declared public and everything else private?

smoke-test/src/test/java/org/opennms/smoketest/BSMAdminIT.java
mvrueden: Please use @FixMethodOrder

smoke-test/src/test/java/org/opennms/smoketest/BSMAdminIT.java
```diff
@@ -28,70 +28,49 @@
...
+        findElementById("delete-" + BASIC_SERVICE_NAME).click();
+        findElementByXpath("//*[contains(text(), 'There are no business services defined.')]");
+    }
 
-            try (CloseableHttpResponse response = httpClient.execute(request)) {
-                assertEquals(201, response.getStatusLine().getStatusCode());
-            }
-        }
+    @Test
+    public void canEditBusinessServices() {
```
mvrueden: Can you please seperate the code segments and add a comment to each of the segments what it is doing?
E.g.:

    //create new service
    ...
    // rename created service
    ...
    // delete created service
    ...

j-white:
I tried the build and it works well.

smoke-test/src/test/java/org/opennms/smoketest/BSMAdminIT.java
DerTak: fixed

PR #462, opened 2015-11-18
create a JAAS login module to use the OpenNMS UserManager

RangerRick:
I apologize in advance to whomever reviews this pull request.

So here's the deal.  Take a look at `container/jaas-login-module` -- in there is the Platonic Ideal of a simple JAAS plugin that loads in Karaf as a bundle and can authenticate against our Spring Security config.  It uses services, and only needs a bundle activator to inject the context into the login module because login modules are classloaded and unaware of OSGi, but is otherwise pretty clean.

The problem comes when you attempt to use the same module *outside* of Karaf, because of the way we load it.  When Karaf registers a JAAS module, it actually wraps it in a proxy `LoginModule` that lets it lazy-initialize.  The problem is, that proxy `LoginModule` isn't in the system classpath, it comes from Karaf, so I couldn't share it with the outside world.

Conversely, I couldn't forego Karaf JAAS initialization altogether, because Karaf initializes a JAAS `Configuration` class that lets it load and unload modules as bundles come and go.

The wonderful design of JAAS initialization in the JVM is such that the `Configuration` class is a singleton, so when Karaf sets it, it owns it.

So.

This commit does a number of horrible things to get around all of these architectural issues.

* I implemented my *own* `Configuration` module, that I initialize alongside the rest of Spring Security except it sits and waits (up to 2 minutes) for the Karaf one to appear and then puts a façade in front of it that short-circuits the Karaf one and returns our own configuration if attempting to access the "opennms" realm.  You can override the 2 minute wait time by setting `-Dorg.opennms.web.springframework.security.jaas-timeout=millis`
* I had to implement my own proxy `LoginModule` as well, because the way login modules get statically initialized means I can't rely on the classpath being what I want.  I put it in opennms-bootstrap and then the bootstrap process injects a clean `$OPENNMS_HOME/lib`-aware classpath statically into the proxy module.  This proxy module then classloads the real one the first time it's initialized.
* I sacrificed one cow, one goat, three chickens, and 17 llamas, so that should be covered.

Someday, when we finish inverting our bootstrap such that we just run inside Karaf, the rest of this crap can all go away and we can just leave the `jaas-login-module` project intact.

*sigh*


RangerRick:
Dangit, I always do this.  Wrong branch.  :)

PR #463, opened 2015-11-18
HZN-531: Implement JAAS Authentication Module for JMX RMI and Karaf

RangerRick:
I apologize in advance to whomever reviews this pull request.

So here's the deal.  Take a look at `container/jaas-login-module` -- in there is the Platonic Ideal of a simple JAAS plugin that loads in Karaf as a bundle and can authenticate against our Spring Security config.  It uses services, and only needs a bundle activator to inject the context into the login module because login modules are classloaded and unaware of OSGi, but is otherwise pretty clean.

The problem comes when you attempt to use the same module *outside* of Karaf, because of the way we load it.  When Karaf registers a JAAS module, it actually wraps it in a proxy `LoginModule` that lets it lazy-initialize.  The problem is, that proxy `LoginModule` isn't in the system classpath, it comes from Karaf, so I couldn't share it with the outside world.

Conversely, I couldn't forego Karaf JAAS initialization altogether, because Karaf initializes a JAAS `Configuration` class that lets it load and unload modules as bundles come and go.

The wonderful design of JAAS initialization in the JVM is such that the `Configuration` class is a singleton, so when Karaf sets it, it owns it.

So.

This commit does a number of horrible things to get around all of these architectural issues.

* I implemented my *own* `Configuration` module, that I initialize alongside the rest of Spring Security except it sits and waits (up to 2 minutes) for the Karaf one to appear and then puts a façade in front of it that short-circuits the Karaf one and returns our own configuration if attempting to access the "opennms" realm.  You can override the 2 minute wait time by setting `-Dorg.opennms.web.springframework.security.jaas-timeout=millis`
* I had to implement my own proxy `LoginModule` as well, because the way login modules get statically initialized means I can't rely on the classpath being what I want.  I put it in opennms-bootstrap and then the bootstrap process injects a clean `$OPENNMS_HOME/lib`-aware classpath statically into the proxy module.  This proxy module then classloads the real one the first time it's initialized.
* I sacrificed one cow, one goat, three chickens, and 17 llamas, so that should be covered.

Someday, when we finish inverting our bootstrap such that we just run inside Karaf, the rest of this crap can all go away and we can just leave the `jaas-login-module` project intact.

*sigh*

JIRA: http://issues.opennms.org/browse/HZN-531
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS258


dschlenk:
As someone who has instances that take ~5 minutes to start, I wonder if that default timeout is a little low? Depends on when this part gets loaded of course. I'll try to spin up a test and see what happens - what would break if the wait time expires?

RangerRick:
The timeout is the time between when the spring security context (inside jetty) is initialized, and the jaas-login-module bundle is loaded inside Karaf.  Most long startups are happening because of event or provisioning stuff long before you get to the webapp bit.  You would only go past that if it takes more than 2 minutes from when monitor.log says it's starting Jetty and finished starting it.  Seems safe, but I could always make it longer.

The only thing that would break AFAIK is trying to access JMX; it would require a username and password but be unable to access the auth mechanism.  Karaf shell would be unaffected.

indigo423:
Is there an JIRA issue for this pull request?

RangerRick:
@indigo423 yup: http://issues.opennms.org/browse/HZN-531

dschlenk:
@RangerRick Yep, from what I can tell they are slow to start due to collectd things. Jetty seems to start relatively quickly. Thanks for the additional detail!

indigo423:
Ok added the links to bamboo, JIRA and formatted the pull request respectively.

features/springframework-security/src/main/java/org/opennms/web/springframework/security/AuthorityPrincipal.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Override
+    public String getName() {
+        return m_name;
+    }
+
+
+    @Override
+    public int hashCode() {
+        final int prime = 31;
```
mvrueden: Please use Objects.hash(...) (http://docs.oracle.com/javase/7/docs/api/java/util/Objects.html#hash%28java.lang.Object...%29) instead of manually hash coding everything.

features/springframework-security/src/main/java/org/opennms/web/springframework/security/AuthorityPrincipal.java
```diff
@@ -0,0 +1,55 @@
...
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+        if (getClass() != obj.getClass()) {
+            return false;
+        }
+        final AuthorityPrincipal other = (AuthorityPrincipal) obj;
+        if (m_name == null) {
```
mvrueden: Please use Objects.equals(...) (http://docs.oracle.com/javase/7/docs/api/java/util/Objects.html#equals%28java.lang.Object,%20java.lang.Object%29) instead of manually check for equality.

mvrueden:
Can you please add some documentation on how this is supposed to work?
Or at least the property to control the "give up" timeout should be documented.

features/springframework-security/src/main/java/org/opennms/web/springframework/security/OpenNMSLoginModule.java
```diff
@@ -0,0 +1,158 @@
...
+        m_subject = subject;
+        m_callbackHandler = callbackHandler;
+        m_sharedState = (Map<String, ? super Object>) sharedState;
+        m_options = (Map<String, ? super Object>) options;
+        m_debug = Boolean.parseBoolean((String) options.get("debug"));
+    }
+
+    @Override
+    public boolean login() throws LoginException {
+        LOG.debug("OpenNMSLoginModule: login()");
```
mvrueden: Is there a way to reuse the "login" logic by generalizing it and use it in both login modules? 
Or at least lets add a comment to explain why we are doing this :)

RangerRick: A little, but a surprising amount of this code is not shared.  :)  Since the Karaf version has to extend the abstract karaf login module, but the other one needs to implement it fully itself.  I've pushed an update that fixes the unit tests and attempts to normalize the code as much as possible, but it's still pretty gross.

RangerRick:
FYI, @soleger intends to merge his other JMX work into this branch so we can review it all before pushing it to 17.  That said, what I have so far has gone green in Bamboo, so this should be OK once his change are read.

features/springframework-security/src/main/java/org/opennms/web/springframework/security/AuthorityPrincipal.java
```diff
@@ -0,0 +1,46 @@
+package org.opennms.web.springframework.security;
```
soleger: Add license.

features/springframework-security/src/main/java/org/opennms/web/springframework/security/LoginHandler.java
```diff
@@ -0,0 +1,24 @@
+package org.opennms.web.springframework.security;
```
soleger: Add license.

features/springframework-security/src/main/java/org/opennms/web/springframework/security/LoginModuleUtils.java
```diff
@@ -0,0 +1,101 @@
+package org.opennms.web.springframework.security;
```
soleger: Add license.

mvrueden:
To prevent license missing issues I edited the code templates of java files, so they by default always have a license created. Maybe we should do the same for the ecilpse folks?

opennms-bootstrap/src/main/java/org/opennms/bootstrap/Bootstrap.java
```diff
@@ -395,6 +395,8 @@ private static void configureRMI(final ClassLoader cl) throws IOException {
             m_rmiServerSocketFactory = new HostRMIServerSocketFactory("localhost");
             RMISocketFactory.setSocketFactory(m_rmiServerSocketFactory);
         }
+
+        OpenNMSProxyLoginModule.setClassloader(cl);
```
soleger: Please add a descriptive comment here about why this is necessary so that people reading or updating the code understand this.

PR #466, opened 2015-11-20
HZN-523: Requisitions with a space in the name are no longer readable

RangerRick:
URL-handling changed in the ReST API in 17, most likely related to the move to CXF.

These changes add some unit tests to validate behavior across all the ForeignSourceRepository implementations, as well as handling the new %20 in URLs.  In the case where the import URL has a %20 in it, Provisiond checks if it's a valid file, and if not, tries translating it and loads that instead.

JIRA: http://issues.opennms.org/browse/HZN-523
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS284

### DIFF
--- a/opennms-provision/opennms-provision-persistence/src/main/java/org/opennms/netmgt/provision/persist/AbstractForeignSourceRepository.java
+++ b/opennms-provision/opennms-provision-persistence/src/main/java/org/opennms/netmgt/provision/persist/AbstractForeignSourceRepository.java
@@ -39,6 +39,7 @@ import org.opennms.core.utils.ConfigFileConstants;
 import org.opennms.core.xml.JaxbUtils;
 import org.opennms.netmgt.provision.persist.foreignsource.ForeignSource;
 import org.opennms.netmgt.provision.persist.requisition.Requisition;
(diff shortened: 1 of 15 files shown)


agalue:
I think that both, the foreign source and the foreign ID cannot contain any of the following characters:

/
\
?
:
&

The above characters are either reserved characters on URLs, or filesystem divisors when using storeByForeignSource.

As an addition to that, we can enhance the Angular UI to reject those characters.

RangerRick:
For historical reasons, it's actually hard to get rid of :, because I guarantee we have customers importing old requisitions with that in it because of the old default imported: one.

I do agree those are all bad, but for fixing the immediate regression, I say we only roll with what I've done here, and open a new issue on adding better handling of the other characters.

agalue:
Guess that thanks to these "historical reasons", the users in question (asumming they exist) will never be able to use the ReST API or the storeByForeignSource with JRobin/RRDtool.

I just made a suggestion based on my experience with support, but go ahead and do what you think is better.

RangerRick:
Oh I definitely think we should look at fixing the other stuff, but it's going to be a LOT more work.  This is to at least bring things back in line with 16 For Now (TM).

PR #474, opened 2015-11-26
NMS-8004: Install iplike stored procedure

indigo423:
Create user for OpenNMS Postgres database and add super user permission to allow installation of C language stored procedure IPLIKE with `./install -dis`. After first start of OpenNMS the opennms user is locked down to a normal user removing super user role and removing possibility to create other databases.

JIRA: http://issues.opennms.org/browse/NMS-8004
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS294


RangerRick:
Seems reasonable to me.

PR #481, opened 2015-12-02
HZN-550: Use only first value in X-Forwarded-Host

fooker:
The header can contain multiple values in the case of proxy chaining.
The first value is the on of the outer-most proxy.

JIRA: http://jira.opennms.org/browse/HZN-550
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS312

### DIFF
--- a/opennms-web-api/src/main/java/org/opennms/web/api/Util.java
+++ b/opennms-web-api/src/main/java/org/opennms/web/api/Util.java
@@ -164,9 +164,13 @@ public abstract class Util extends Object {
      */
     public static String getHostHeader(final HttpServletRequest request) {
         for (int i = 0; i < hostHeaders.length; ++i) {
-        	final String ret = request.getHeader(hostHeaders[i]);
-            if (ret != null) {
-                return ret;
+            // Get the first value in the header (support for proxy-chaining)
+            final String header = request.getHeader(hostHeaders[i]);
+            if (header != null) {
+                final String[] values = header.split(", *");
+                if (values.length >= 1) {
+                    return values[0];
+                }
             }
         }
         return request.getServerName() + ":" + Integer.toString(request.getServerPort());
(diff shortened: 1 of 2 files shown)


mvrueden:
Hey @fooker, what about a shiny unit test for this?
There is already a UtilTest class available.

fooker:
I've already looked at the tests. I think trying to mock the ```HttpServletRequest``` is not worth the effort.

In adition, the basic funktionality is covered by the smoke tests (it works without a proxy).

mvrueden:
You can just add another method with only getting the headers as parameter. THen you do not have to mock the servlet request.

opennms-web-api/src/main/java/org/opennms/web/api/Util.java
```diff
@@ -164,9 +164,10 @@ protected static String substituteUrl(final HttpServletRequest request, final St
      */
     public static String getHostHeader(final HttpServletRequest request) {
         for (int i = 0; i < hostHeaders.length; ++i) {
-        	final String ret = request.getHeader(hostHeaders[i]);
-            if (ret != null) {
-                return ret;
+            // Get the first value in the header (support for proxy-chaining)
+            final Enumeration<String> headers = request.getHeaders(hostHeaders[i]);
+            if (headers.hasMoreElements()) {
+                return headers.nextElement();
```
j-white: The behavior appears to be identical: we're still using the value from the first occurrence of the header.

I understand that the value of the X-Forwarded-Host header may many (comma separated) values, but I don't see how this change solves the problem.
Aren't the values in X-Forwarded-Host 
It still looks like we're using the value from the first occurence

fooker: The difference is that ```getHeaders``` returns the values of the header already splitted whereas ```getHeader``` returns the concatinated format.

fooker:
Looks like the implementation of ```getHeaders``` is not consistent. Therefore I do manual splitting.
I also added some unit tests.

mvrueden:
Thanks for adding the test. Looks good to me. @j-white what do you think?

j-white:
Cherry-picked to foundation-2016 and renamed the JIRA issue to NMS-8029.

PR #478, opened 2015-12-02
HZN-344: Provide a tool for importing existing .rrd / .jrb metrics into Newts

fooker:
This is a follow-up of: https://github.com/OpenNMS/opennms/pull/473

Added a tool for converting existing metric repositories to NewTS.

JIRA: http://issues.opennms.org/browse/HZN-344
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS235

PR #473, opened 2015-11-26
HZN-344: Provide a tool for importing existing .rrd / .jrb metrics into Newts

fooker:
Added a tool for converting existing metric repositories to NewTS.

JIRA: http://issues.opennms.org/browse/HZN-344
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS235

PR #487, opened 2015-12-03
HZN-344: Provide a tool for importing existing .rrd / .jrb metrics into Newts

fooker:
This is a follow-up of: #473 and #478.

Added a tool for converting existing metric repositories to NewTS.

JIRA: http://issues.opennms.org/browse/HZN-344
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS438


j-white:
Following up on my previous comment about the failure with the .jrb files:
Shouldn't we avoid trying to process to the .jrb files when the '-t' flag is not set? I know of a few deployments that have both .jrb and .rrd files laying around in the folders.

features/newts-repository-converter/src/main/java/org/opennms/features/newts/converter/NewtsConverter.java
```diff
@@ -0,0 +1,613 @@
...
+    private NewtsConverter(final String... args) {
+        final Options options = new Options();
+
+        final Option helpOption = new Option("h", "help", false, "Print this help");
+        options.addOption(helpOption);
+
+        final Option opennmsHomeOption = new Option("o", "onms-home", true, "OpenNMS Home Directory (defaults to /opt/opennms)");
+        options.addOption(opennmsHomeOption);
+
+        final Option rrdPathOption = new Option("r", "rrd-dir", true, "The path to the RRD data (defaults to /var/opennms/rrd)");
```
j-white: Can we make the default RRD path relative to onms-home?

j-white:
I also ran into exceptions of the form (I can provide the .jrbs if needed):
```
Exception in thread "ForkJoinPool-1-worker-7" java.lang.IllegalArgumentException: value (-533718590) is outside the range for an unsigned long value
        at com.google.common.base.Preconditions.checkArgument(Preconditions.java:148)
        at com.google.common.primitives.UnsignedLong.valueOf(UnsignedLong.java:95)11:25:15.761 [ForkJoinPool-1-worker-7] INFO  org.opennms.features.newts.conv
erter.NewtsConverter - Processing resource: dir=/home/jesse/git/opennms/target/opennms/share/rrd/snmp/1, file=memShared, group=ucd-memory

        at org.opennms.features.newts.converter.NewtsConverter.injectSamplesToNewts(NewtsConverter.java:506)
        at org.opennms.features.newts.converter.NewtsConverter.processResource(NewtsConverter.java:464)
        at org.opennms.features.newts.converter.NewtsConverter.lambda$processStoreByMetricResource$6(NewtsConverter.java:427)
        at org.opennms.features.newts.converter.NewtsConverter$$Lambda$15/1675318936.run(Unknown Source)
        at java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1402)
        at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
        at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
        at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1689)
        at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:157)
```

j-white:
After running the tool I hit the exception above, and it appears to have completed:
```
11:25:16.047 [Main] INFO  org.opennms.features.newts.converter.NewtsConverter - Conversion Finished: metrics: 178, samples: 2252, time: 6 sec 
```

but the process just hangs in the foreground.

features/newts-repository-converter/src/main/java/org/opennms/features/newts/converter/NewtsConverter.java
```diff
@@ -0,0 +1,613 @@
...
+        try (final BufferedReader r = Files.newBufferedReader(path.resolve(String.format("%s.meta", metric)))) {
+            meta.load(r);
+
+        } catch (final IOException e) {
+            throw Throwables.propagate(e);
+        }
+
+        final String group = meta.getProperty("GROUP");
+        if (group == null) {
+            throw NewtsConverterError.create("No group information found - please verify storeByGroup settings");
```
j-white: It's possible that one or more archives do not have the GROUP defined in their meta files i.e. a some metrics were collected for an interface that no longer exists.

I think we should log the error and indicate the name of the file in question, but continue to process the remaining archives when this happens.

j-white:
It looks like the tool is also converting the metrics to use storeByFs.

We should fail if storeByFs is not enabled, similar to how we fail is Newts is not enabled.

j-white:
Also, we need to use Newts 1.3.1 instead of 1.3.2-SNAPSHOT, it's going to be a few weeks before 1.3.2 is released.

j-white:
Aside from the comments above, it appears to be working pretty well.

agalue:
The original design of the tool was to "assume" storeByFS is enabled and also storeByGroup is enabled, because someone told me that Newts does the same assumptions. I didn't know that it has to be explicitly enabled in opennms.properties in order to use it.

j-white:
Newts always uses storeByGroup, but the same is not true for storeByFS.

agalue:
Is that documented somewhere ?

j-white:
There's a note on the recommendation to use storeByFS here:
  http://docs.opennms.org/opennms/releases/17.0.0/guide-admin/guide-admin.html#_configuration_2

No mention of storeByGroup though.

agalue:
bookmarked!

Regarding, storeByGroup, that's why I mentioned here ;)

features/newts-repository-converter/src/main/java/org/opennms/features/newts/converter/NewtsConverter.java
fooker: Done

fooker:
Now, storeByForeignSource is enforced.

features/newts-repository-converter/src/main/java/org/opennms/features/newts/converter/NewtsConverter.java
fooker: I switched to use logging instead of exceptions to keep the converter running.

fooker:
I've (hopefully) fixed the negative values issue. @j-white, can you try again?

fooker:
Looks like exceptions in a for join pool caus the thread to stay alive. Fixed hanging at exit.

fooker:
Now, the ```-s``` is non-optional. This makes the behaviour more clear.

j-white:
I'm still running into some issues (HEAD is b3940a604da):
* I get an NPE when running './bin/newts-repository-converter -s false'
* We still try and process JRB files when '-t rrd' is set, although the conversion no longer stops. We should only be processing files with the '.rrd' extension in this case.
* The tool still isn't exiting when an exception is thrown during processing

fooker:
The tool searches for ```.meta``` files and tries to parse the according ```.rrd``` or ```.jrb``` file (depending on the flag). If the file was not found, the tool throwed a cryptical parser error.

* NPE fixed
* Now it will print a file not found message if the .jrb / .rrd file was not found (or the wrong flag was set)
* Enforced exiting

fooker:
Bamboo is failing because we are missing some NewTS fixes comming in v1.3.3. We should wait until these fixes are released and bump the version before merging.

j-white:
foundation-2016 now includes the release of Newts 1.3.3. I've merged foundation-2016 into this branch and re-instated the Bamboo build.

Assuming the tests pass, we should probably rebase and squash this branch before merging.

fooker:
Rebased ontop foundation-2016 and squashed.

j-white:
Looks like we're in good shape now. Will merge this once Bamboo goes green.

agalue:
w00t :)

pbrane:
> On Feb 17, 2016, at 2:19 PM, Alejandro Galue <notifications@github.com> wrote:
> 
> w00t :)
> 

w00t w00t!

soleger:
Once this percolates through Bamboo, we'll include it as part of the 17.1.0 release.

PR #489, opened 2015-12-03
HZN-552: Fixed grafana-box.jsp to use the new Grafana API

christianpape:
Jira: http://issues.opennms.org/browse/HZN-552
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS317


j-white:
I assume this works with Grafana v2.5.0 now.

Can we add a note in the docs that references the supported API version(s)?

PR #490, opened 2015-12-03
HZN-551:Define optional event parameters on the event definition

agalue:
Define optional event parameters on the event definition.

JIRA: http://issues.opennms.org/browse/HZN-551
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS314

### DIFF
--- a/features/events/daemon/src/main/java/org/opennms/netmgt/eventd/EventExpander.java
+++ b/features/events/daemon/src/main/java/org/opennms/netmgt/eventd/EventExpander.java
@@ -42,10 +42,13 @@ import org.opennms.netmgt.xml.event.Event;
 import org.opennms.netmgt.xml.event.Header;
 import org.opennms.netmgt.xml.event.Logmsg;
 import org.opennms.netmgt.xml.event.Operaction;
+import org.opennms.netmgt.xml.event.Parm;
 import org.opennms.netmgt.xml.event.Tticket;
 import org.opennms.netmgt.xml.event.UpdateField;
+import org.opennms.netmgt.xml.event.Value;
 import org.opennms.netmgt.xml.eventconf.Decode;
 import org.opennms.netmgt.xml.eventconf.Maskelement;
+import org.opennms.netmgt.xml.eventconf.Parameter;
 import org.opennms.netmgt.xml.eventconf.Varbindsdecode;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.util.Assert;
@@ -722,6 +725,22 @@ public final class EventExpander implements org.opennms.netmgt.dao.api.EventExpa
                 }
                 
                 e.setAlarmData(alarmData);
+            }
+
+            if (econf.getParameterCollection() != null && econf.getParameterCount() > 0) {
+                if (e.getParmCollection() == null) {
+                    e.setParmCollection(new ArrayList<Parm>());
+                }
+                for (Parameter p : econf.getParameterCollection()) {
+                    Parm parm = new Parm();
+                    parm.setParmName(p.getName());
+                    Value v = new Value();
+                    v.setContent(p.getValue());
+                    v.setType("string");
+                    v.setEncoding("text");
+                    parm.setValue(v);
+                    e.addParm(parm);
+                }
             }
         }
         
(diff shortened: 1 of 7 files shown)


features/events/daemon/src/main/java/org/opennms/netmgt/eventd/EventExpander.java
```diff
@@ -723,6 +726,22 @@ public synchronized void expandEvent(Event e) {
...
+                }
+                for (Parameter p : econf.getParameterCollection()) {
+                    Parm parm = new Parm();
+                    parm.setParmName(p.getName());
+                    Value v = new Value();
+                    v.setContent(p.getValue());
+                    v.setType("string");
+                    v.setEncoding("text");
+                    parm.setValue(v);
+                    e.addParm(parm);
```
j-white: What's the expected behavior when an event already has a parameter with the same name?

agalue:
Good question. If you have the answer, let me know and I make it happen.

agalue:
The parameters is an ArrayList, so without changes, it will be a duplicate. That being said. This is something that a human being is going to manually configure. So, if the name exist, it is their fault, not OpenNMS fault.

My $0.02

j-white:
I would opt for trying to avoid duplicate parameters with the same name.

Maybe we should treat the value in the eventconf as a default, and only set it if a parameter with the same name doesn't already exist.

agalue:
Most of the new event definitions are associated with SNMP Traps (which already have a pre-defined and well known set of parameters that use the OID, not a readable text, as the parameter name), or threshold related events (which already have a well known set of parameters).

This feature is intended to simplify the configuration of potential new event listeners that will take actions based on the parameters in questions. That means, the OpenNMS administrator has to choose the name . For traps it is not a problem, because I don't think that a human will choose an OID instead of a human readable text for the parameter name.

For the thresholds, please allow me to understand why someone will try to override an existing parameters, considering that all the existing parameters should exist ?

In my opinion, the default behavior is more than OK, because if someone is using a name that is already used, this is considered a bad configuration, and like on any other configuration file in OpenNMS, the user should fix it.

Now, if you have a better idea, that won't break the use case and purpose of this enhancement, go ahead and change the code.

soleger:
It does sound like a corner case that anyone would send an event that contains a new parm value for an event where the parm value is already defined in the eventconf file. However, I don't think it makes sense to end up with two copies of the same parm key in one event.

We can either overwrite the default value with the value from the event, always keep the default value, or add some extra config that can change the behavior. I'd recommend overwriting the value with the value from the incoming event. That means that if, for whatever reason, the event producer wants to override the default behavior, they can do that on an event-by-event basis.

agalue:
Seth, those sound good but those enhancements are out of the scope of the purpose of this particular enhancement. I suggest to create other Jira issues for them.

Now, if you see a duplicate parameters on a event on the DB, well that's user's fault, and he or she should fix it. Once again, if you have another idea (keeping in mind the scope of the request), update the code. Then, I'll make sure that the use case I need still works.

indigo423:
HashMap behavior? Parameter name as key and the last value with the same name wins?

pbrane:
> On Dec 3, 2015, at 5:27 PM, Ronny Trommer <notifications@github.com> wrote:
> 
> HashMap behavior? Parameter name as key and the last value with the same name wins?
> 
Here’s my $0.02.

There are pretty many events that your software generates that have parameters that never get displayed or persisted because the eventconf has no reference to them.  So, there is a chance that would not be the user’s fault if they inadvertently defined a parameter that was already being used by the system.

However, the use case for this, as Alejandro explains, is that the user is probably adding very custom parameters for some integration behavior and if they are doing that, they probably have a pretty good idea what they’re doing and this is probably a “tail wagging the dog” issue and not a problem.

That being said, if they did inadvertently define a parameter that was being used by the system, the code that is creates these new parameters should:

1) Write a WARN log message.
2) Keep the original parameter since it was created by the software is probably critical to the system.

The user will be able to see that their parameter isn’t getting set to the defined value and they’ll figure it out.

Additionally, someone should go through all the code that references the EventBuilder and make sure that all the parameters that get created are defined/documented in the appropriate eventconf entry.  Internship project?


David

agalue:
David, thank you very much for your comment. I agree with you, I'll introduce those changes, and create a new pull request.

agalue:
Here is the new one: https://github.com/OpenNMS/opennms/pull/493

PR #492, opened 2015-12-04
BSM-62: BusinessService AdminPage

DerTak:
Contains the subtasks BSM-70 and BSM-71

Issues:
http://issues.opennms.org/browse/BSM-62
http://issues.opennms.org/browse/BSM-70
http://issues.opennms.org/browse/BSM-71

Bamboo:
http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS319


pbrane:
> On Dec 4, 2015, at 08:20, Markus Neumann <notifications@github.com> wrote:
> 
> Bamboo:
> Bamboo is deleting the BSM-62 build after each run, so there is no reasonable link at the moment.

Is that normal?

DerTak:
We are investigating the BSM-62 vs Bamboo problem.

smoke-test/src/test/java/org/opennms/smoketest/BSMAdminIT.java
```diff
@@ -1,88 +1,197 @@
...
 @FixMethodOrder(MethodSorters.NAME_ASCENDING)
 public class BSMAdminIT extends OpenNMSSeleniumTestCase {
 
     private final String RENAMED_SERVICE_NAME = "renamed_service";
-    private final String BSM_LIST_URL = BASE_URL + "opennms/admin/bsm/index.jsp";
+    private final String BSM_ADMIN_URL = BASE_URL + "opennms/admin/bsm/adminpage.jsp";
     private final String BASIC_SERVICE_NAME = "BasicService";
-    private final String BASIC_SERVICE_NAME_TO_CHANGE = "BasicServiceToChange";
+
+    private final String ADD_BUTTON_XPATH = "/html/body/div[3]/div[5]/div/div/div[5]/div/div/div/div[3]/div/div[2]/div/div[2]/div[1]/span/span";
```
j-white: Can we come up with better xpath expressions for these? Any small layout changes changes are bound to break these tests.

agalue: I agree with Jesse. It is always better to minimize as much as you can either the XPath or the CSS Selector used to find an HTML element. Also, I would use the easier from both selection technologies.

Here is a hint. With FireBug, on the console, you can write something like $x("blahblah") where blah blah is an XPath, to verify your selection.

j-white:
Can we rename the page title and links to be 'Manage Business Services' instead of 'Business Service Monitoring Admin Page'?

Also, when looking at the list of business services, there's a mismatch between the size of the columns in the rows, and those in the header. (I can provide a screenshot if this description is not clear)

DerTak:
screenshots would be good.

smoke-test/src/test/java/org/opennms/smoketest/BSMAdminIT.java
DerTak: Fixed: Base-Xpath of the IpServiceSelector is now used to reduce the xpath expressions.

j-white:
I can't find a consistent way to reproduce it, let's forget it about it for now.

j-white:
Looks good overall, let's merge once Bamboo goes green.

PR #494, opened 2015-12-04
NMS-8015: remove SMS reflector from tools and RXTX from build

RangerRick:
While the SMS reflector has been removed for a few major releases, the RXTX dependencies have remained.  Remove RXTX and remove the SMS reflector from opennms-tools since it has gone unmaintained and unused.


pbrane:
I think it is quite possible that someone could be using the RXTX stuff for "out of band" notifications.  I would, if I had time to implement it.  Could we pull this into an opennms-tools branch or is it there already?

RangerRick:
It's there already, and no one has even checked if it still works for 2 major releases.  It would take some work to modernize it and fix the rxtx dependencies to be in the tools version rather than in OpenNMS proper.

pbrane:
Okay.

soleger:
We can also do OOB SMS notifications by following Tarus' recent blog post:

http://www.adventuresinoss.com/2015/09/03/send-an-sms-with-opennms/

pbrane:
Yeah, I was looking for plugin support within the JVM.  But, I agree, that works, too.

David

PR #495, opened 2015-12-04
NMS-8018: Vacuumd sends automation events before transaction is closed

j-white:
JIRA: http://issues.opennms.org/browse/NMS-8018
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS320/latest


agalue:
Looks good to me. I'll merge it when Bamboo goes green.

PR #504, opened 2015-12-10
NMS-8032: Describe behavior and usage of Alarms

indigo423:
Created document folder structure and built a diagram which describe state changes of an Alarm.

JIRA: http://issues.opennms.org/browse/NMS-8032
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS329


pbrane:
A couple of corrections:
- Indeterminate and Normal severity Alarms shouldn't be shown going to a clear severity/state.
- Clear state/severity Alarms shouldn't be shown going to a Normal severity.

*Just a side note:
We recently had a conversation around confusion of Alarm status vs. state vs. severity.  We eventually supposed that we should use two statuses to make a state determination: Operational Status and Administrative Status.  Then, a severity can be applied to the determined state.  That is not in current model, however.*

indigo423:
This work in progress is base for some issues triggered in opennms-discuss: http://opennms.markmail.org/message/7ifupkvjh6arhtms?q=opennms+list:net%2Esourceforge%2Elists%2Eopennms-discuss+order:date-backward&page=1

indigo423:
@pbrane 
It is possible to click on "Clear" in the WebUI on a "Normal" and "Indeterminate" Alarm. By Correction you mean this should be filed as a bug in the application?
It is possible to escalate a Cleared Alarm in the WebUI? Same here?

pbrane:
> On Dec 10, 2015, at 3:46 AM, Ronny Trommer <notifications@github.com> wrote:
> 
> It is possible to click on "Clear" in the WebUI on a "Normal" and "Indeterminate" Alarm. By Correction you mean this should be filed as a bug in the application?
> 
My description is by convention/best practices currently in the alarm workflow of Alarmd+Vacuumd (where it is currently held).
> It is possible to escalate a Cleared Alarm in the WebUI? Same here?
> 
I know that programmatically it is possible to do these things, it’s just not the convention built into the automations.  If a user changes the severity of an alarm, then they should understand this convention which is why I mention it for your documentation.  Of course, the user could change the escalation and clearing conventions that come default in Vacuumd and therefore may want to do as you have described in the UI to meet their specific design.  (Just because you can in the UI doesn’t mean you should ;)  We could introduce a “standard behavior or policy” for the UI and make it configurable.  Better yet, we should have this workflow as a service that the UI and other components can implement.  Add that to the list ;)

agalue:
+1 to everything said by pbrane :)

indigo423:
Sorry, I have to nag again, I don't understand how your comment answers my question :)

It is possible to Clear a "Normal" and "Indeterminate" event in the Web UI. This is nothing what is related to a users "custom" configuration. The application introduces this behavior, which I understand from your explanation as not correct?

The same thing to escalate a Cleared Alarm in the WebUI. 

I don't talk about the possibilities you can bend configuration or customizations. I mean the introduced behavior which is given by the web application? If both ways are not intended, we should file a bug and fix the WebUI to do not allow this workflows?

Until now I have more reasons to file a bug against the Alarm Web UI. Is this not intended a explanation would be good :)

indigo423:
I would definitely would make the "standard behavior or policy" or default behavior the intended. You both have probably use cases in mind which I probably don't, so ok to me.

Thanks for clarification.

agalue:
> On Dec 10, 2015, at 4:41 PM, Ronny Trommer <notifications@github.com> wrote:
> 
> Sorry, I have to nag again, I don't understand how your comment answers my question :)
> 
> It is possible to Clear a "Normal" and "Indeterminate" event in the Web UI. This is nothing what is related to a users "custom" configuration. The application introduces this behavior, which I understand from your explanation as not correct?
> 
That's right. Unless a user is 100% sure about what is about to do, clearing alarms from the WebUI makes sense only for alarms related with "problems without resolution" (i.e. alarm-type=3).

Indeterminate is the default severity used when you don't have any idea what severity an event should have, like the MIB compiler. MIBs don't have that information, and of course, the severity will be Indeterminate. To make the event useful, it should have a severity.
> The same thing to escalate a Cleared Alarm in the WebUI. 
> 
This doesn't make sense. If an alarm is cleared and you can escalate it through the WebUI, then you've found a bug.

Alejandro Galue
agalue@opennms.org <mailto:agalue@opennms.org>
PGP Key Fingerprint: 5293 6234 1E75 DF30 7821  1823 87AF 972E DAF8 BE2C

j-white:
Is this still work in progress? :)

j-white:
@indigo423 what's left to be done here? Aside from your merge, there hasn't been any changes since last year, and this PR is still marked as work in progress.

indigo423:
Reopen when branch is ready to be merged

PR #505, opened 2015-12-10
NMS-8024: Add support for OSGi-based Ticketing plugins.

j-white:
* Moved ticketd out of opennms-services and into it's own module
* Added a new OSGiBasedTicketerPlugin that uses a plugin from the OSGi service registry
* Updated the JIRA ticketing plugin to use the latest client API bindings
* Moved the JIRA ticket plugin to use the new OSGi interface
* Documented how to setup the JIRA plugin

JIRA: http://issues.opennms.org/browse/NMS-8024
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS325


features/ticketing/daemon/src/main/java/org/opennms/netmgt/ticketd/DefaultTicketerServiceLayer.java
```diff
@@ -114,16 +102,16 @@ public void afterPropertiesSet() throws Exception {
     public void cancelTicketForAlarm(int alarmId, String ticketId) {
         OnmsAlarm alarm = m_alarmDao.get(alarmId);
         if (alarm == null) {
-            throw new ObjectRetrievalFailureException("Unable to locate Alarm with ID: "+alarmId, null);
+            LOG.error("No alarm with id {} was found. Ticket with id '{}' will not be canceled.", alarmId, ticketId);
```
soleger: Oh no, it's the Canadian spelling of 'canceled'!

dependencies/pom.xml
```diff
@@ -28,7 +28,6 @@
     <module>javamail</module>
     <module>jaxb</module>
     <module>jcifs</module>
-    <module>jersey</module>
     <module>jersey-client</module>
```
soleger: With these changes, the only things left using jersey-client-dependencies appear to be features/ticketing and features/resource-cli. Does that sound correct?

j-white: features/reporting/repository still relies on it too. We could work on migrating these to use cxf though.

dependencies/jersey/pom.xml
```diff
@@ -1,155 +0,0 @@
...
-  <repositories>
-    <repository>
-      <snapshots><enabled>false</enabled></snapshots>
-      <releases><enabled>true</enabled></releases>
-      <id>maven2-repository.dev.java.net</id>
-      <name>Java.net Repository for Maven</name>
-      <url>http://maven.opennms.org/content/repositories/maven2-repository.dev.java.net/</url>
-    </repository>
-  </repositories>
-</project>
```
mvrueden: R.I.P.

features/ticketing/api/pom.xml
```diff
@@ -8,6 +8,21 @@
...
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <extensions>true</extensions>
+        <configuration>
+          <instructions>
+            <Bundle-RequiredExecutionEnvironment>JavaSE-1.8</Bundle-RequiredExecutionEnvironment>
```
mvrueden: This is not really part of this pull request, but I saw this now more often.
We have properties for everything, except this. How about having a JdkVersion8String and a JdkVersion7String or something like that defined in the root pom. Would make migration easier in the future.

features/ticketing/jira-integration/pom.xml
```diff
@@ -7,55 +7,36 @@
...
+              <Bundle-RequiredExecutionEnvironment>JavaSE-1.8</Bundle-RequiredExecutionEnvironment>
+            </instructions>
+          </configuration>
+        </plugin>
+        <plugin>
+            <groupId>org.opennms.maven.plugins</groupId>
+            <artifactId>features-maven-plugin</artifactId>
+            <configuration>
+                <bundles>
+                    <bundle>mvn:joda-time/joda-time/${jodaTimeVersion}</bundle>
```
mvrueden: You do not have to define the bundles manually if they are defined in the dependencies as not provided and are real bundles.

features/ticketing/api/pom.xml
soleger: I agree. We should call it osgiExecutionEnvironment just to make its usage explicit. Then we can change it later on to Java-1.9 and the variable name doesn't change.

mvrueden: We also need another one, maybe `osgiExecutionEnvironmentLegacy` for the previous version of the JRE. We have at least one bundle which requires JRE 7 and not 8.

PR #507, opened 2015-12-10
BSM-80: Update the Business Service Model to depend on Business Services

fooker:
Jira: http://issues.opennms.org/browse/BSM-80
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS326/latest

### DIFF
--- a/features/bsm/service/impl/src/test/java/org/opennms/netmgt/bsm/service/internal/BusinessServiceManagerImplIT.java
+++ b/features/bsm/service/impl/src/test/java/org/opennms/netmgt/bsm/service/internal/BusinessServiceManagerImplIT.java
@@ -59,6 +59,8 @@ import org.springframework.transaction.annotation.Transactional;
 
 import com.google.common.collect.Lists;
 
+import java.util.Collections;
+
 @RunWith(OpenNMSJUnit4ClassRunner.class)
 @ContextConfiguration(locations = {
         "classpath:/META-INF/opennms/applicationContext-commonConfigs.xml",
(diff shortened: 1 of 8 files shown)


j-white:
Looks good to me.

features/bsm/service/impl/src/test/java/org/opennms/netmgt/bsm/service/internal/BusinessServiceManagerImplIT.java
```diff
@@ -153,6 +159,65 @@ public void testGetOperationalStatusForBusinessService() {
...
+                            businessServiceManager.getFeasibleChildServices(businessServiceManager.getById(serviceId1)));
+        Assert.assertEquals(ImmutableSet.of(businessServiceManager.getById(serviceId3)),
+                            businessServiceManager.getFeasibleChildServices(businessServiceManager.getById(serviceId2)));
+        Assert.assertEquals(ImmutableSet.of(),
+                            businessServiceManager.getFeasibleChildServices(businessServiceManager.getById(serviceId3)));
+    }
+
+    @Test
+    public void testLoopCreation() {
+        thrown.expect(IllegalArgumentException.class);
```
mvrueden: This is not sufficient.
An IllegalArgumentException can be thrown in so many occasians. Please either throw a more meaningfull exception such as "LoopDetectedException" or verify that the IllegalArgumentException is exactly the exception you expect.

mvrueden:
One side-node: This branch should be merged to BSM-62 instead of features/bsm.

features/bsm/service/impl/src/test/java/org/opennms/netmgt/bsm/service/internal/BusinessServiceManagerImplIT.java
fooker: I don't think new exception types are a good idea - but we should discuss this somewhere else. To ensure the right exception is triggered, I will check themessage, too.

fooker:
I think we can merge this to ```jira/BSM-61``` manually to avoid creating another pull request.

fooker:
Merged manually to ```jira/BSM-61```.

fooker:
Allow deletion of services which are still assigned to other services.

mvrueden:
I am a bit confused here. @fooker mentioned that a change to the code still needs to be made (verify the right exception is thrown in the tests). This change was never made here, but the pull request is marked as REVIEW OK and was merged manually to `jira/BSM-61`.

mvrueden:
I added the `Do NOT merge` label to avoid merging it to `features/bsm` as it needs to be manually merged to `jira/BSM-61`

j-white:
The code for BSM-80 is already merged into the BSM-61 branch.

PR #508, opened 2015-12-11
HZN-555: Fixed SQL statement to include all categories

christianpape:
Jira: http://issues.opennms.org/browse/HZN-555
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS332


mvrueden:
Looks good to me, but can you please provide a test for the bug?

PR #509, opened 2015-12-11
Fix for Enhancement NMS-8030

agalue:
Improve the SNMP data collection config parsing to give more flexibility to the users.

Bamboo is green:
http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS331

Technically, this enhancement can be considered as a performance improvement because I reduced the time spent building the effective configuration. That means, the initial time reading datacollection-config.xml and the time spent when re-reading the configuration has been reduced.

In case it is necessary to merge this into foundation for Meridian 2015, I think it won't be hard.

### DIFF
--- a/opennms-config/src/main/java/org/opennms/netmgt/config/DataCollectionConfigParser.java
+++ b/opennms-config/src/main/java/org/opennms/netmgt/config/DataCollectionConfigParser.java
@@ -96,7 +96,11 @@ public class DataCollectionConfigParser {
                         throwException("You must specify at least the data collection group name or system definition name for the include-collection attribute", null);
                     } else {
                         // Include One system definition
-                        addSystemDef(collection, include.getSystemDef());
+                        SystemDef systemDef = getSystemDef(include.getSystemDef());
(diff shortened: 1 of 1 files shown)


j-white:
Are there any tests that verifies the new functionality?

agalue:
No but I can modify my playground to be a valid JUnit test.

For me a feature works, if I can run OpenNMS and verify the functionality, as seeing a JUnit test passing is not good enough for me. 

Assuming that's the only requirement I'm going to add the test and merge the branch directly. If you have objections let me know ASAP

agalue:
commit efdd527d07cfe84cce627c20be0cbd39d3febd92
Author: Alejandro Galue <agalue@opennms.org>
Date:   Fri Dec 11 16:54:33 2015 -0500

    Adding JUnit test to verify the new functionality (to make Jesse happy).

j-white:
Looks good, but let's try and avoid passive aggressive comments in our commit messages next time.

mvrueden:
Yes, it made me happy too :)

PR #513, opened 2015-12-15
BSM-81: Update the Business Service Admin Page to be able to build a Business Service Hierarchy

christianpape:
Jira: http://issues.opennms.org/browse/BSM-81
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS387


DerTak:
Create two BS, test1 and test2. Add test2 as a member of test1. Try to delete test2.
The result is a red marker on the delete button an a hibernate exception.

j-white:
Similar behavior happens if you try and delete a node after one of it's IP services have been associated to a BS.

Looks like we need to add "on delete cascade" to some of our foreign keys.

christianpape:
Exactly, but in my opinion this should be handled by the service, not by the UI. I can react on exceptions and/or check pre-conditions via service method calls. The ReST API will face the same problems, right?

fooker:
I think ```ON DELETE CASCADE``` is ok for the node relation. For BS, thereshould be some kind of hint to the user showing which services are affected.

pbrane:
> On Dec 15, 2015, at 9:28 AM, Christian Pape <notifications@github.com> wrote:
> 
> Exactly, but in my opinion this should be handled by the service, not by the UI. I can react on exceptions and/or check pre-conditions via service method calls. The ReST API will face the same problems, right?

+1

pbrane:
> On Dec 15, 2015, at 9:31 AM, Dustin Frisch <notifications@github.com> wrote:
> 
> I think ON DELETE CASCADE is ok for the node relation. For BS, thereshould be some kind of hint to the user showing which services are affected.

Business logic in the DB is scary though often it seems necessary… not sure why.  It is a place that the user can see the business logic by viewing the DB schema.  However, documentation should be the place where the user should view this information: JavaDoc and User Docs.  It gets scary trying to maintain this logic in 2 places: within the service itself and in the object model.  Those can often clash or get out of sync.

David

mvrueden:
To the node foreign key relation problem:
Do we really only want to add ON CASCADE definitions? 
This is the implementing solution for the problem, but is that the correct solution?
I mean a business service is defined which the user wanted the way it is defined.
Now we delete nodes and ip services and if we remove the association of the ip services from the business service we change the business service and affect maybe other business services.
We should discuss a way how to make this as transparent to the user as possible as this is definetly something one does not think about!
Maybe we can send events or notifications in these cases.
Or maybe we duplicate the reduction keys of the removed ip services and attach them to the business service.

DerTak:
The "what to do if nodes or interfaces get deleted" stuff is not in the scope of BSM-81. We should move this discussion into BSM-90, the related issue in jira. (I'll copy your comment over).

PR #517, opened 2015-12-16
HZN-563: Improve flexibility to the syslog alarm northbounder

agalue:
This set of code changes provide the ability to apply powerful filters over the syslog northbound destinations, as well as the ability to reload the configuration.

The reload feature has some limitations due to the way the syslog northbound interface was implemented. Because each declared destination is going to be an independent implementation of the Northbounder interface, you cannot rename, add or remove destinations without restart OpenNMS. But, you can change the internal settings of any destination like IP, port, protocol, etc.

Of course, all the filters will be reloaded properly.

The code has JUnit tests to validate all the features introduced, and all of them has been tested on a lab environment.

JIRA: http://issues.opennms.org/browse/HZN-563
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS337

### DIFF
--- a/opennms-alarms/api/src/main/java/org/opennms/netmgt/alarmd/api/NorthboundAlarm.java
+++ b/opennms-alarms/api/src/main/java/org/opennms/netmgt/alarmd/api/NorthboundAlarm.java
@@ -29,14 +29,18 @@
 package org.opennms.netmgt.alarmd.api;
 
 import java.net.InetAddress;
+import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
+import org.opennms.netmgt.events.api.EventParameterUtils;
 import org.opennms.netmgt.model.OnmsAlarm;
 import org.opennms.netmgt.model.OnmsDistPoller;
 import org.opennms.netmgt.model.OnmsSeverity;
 import org.opennms.netmgt.model.TroubleTicketState;
+import org.opennms.netmgt.xml.event.Parm;
 
 /**
  * Wraps the OnmsAlarm into a more generic Alarm instance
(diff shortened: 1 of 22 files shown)


opennms-alarms/api/src/main/java/org/opennms/netmgt/alarmd/api/NorthboundAlarm.java
```diff
@@ -280,7 +284,11 @@ public static x733AlarmType get(int id) {
 	
     private final Integer m_id;
     private final String m_uei;
-	private Integer m_nodeId;
```
j-white: The ElasticSearch NorthBounder (merged in 18.0.0) relies on the nodeId.

See https://github.com/OpenNMS/opennms/blob/develop/features/elasticsearch/event-forwarder/src/main/java/org/opennms/features/elasticsearch/eventforwarder/internal/NodeCache.java for details.

j-white: Nevermind, it's still there.

j-white:
Looks good, as you mentioned in the Jira issue, the reload() functionality is a bit limited though.

I think it can be solved fairly easily by registering the SyslogNorthbounderManager as Northbounder and handling the reload events there instead.

agalue:
@j-white your idea is brilliant! I'm going to close this pull request and re-implement the reload feature that way.

agalue:
I modified the code to properly register/unregister the northbounders when the configuration is changed, so now there are not restrictions or limitations on the reload config.

I moved the filter list inside the destination because it makes more sense, and the configuration looks cleaner now.

Finally, I reformatted all the code of opennms-alarms/syslog-northbounder to have proper indentation and simple JavaDocs (to make it look more readable). This is the reason why you may see that all the classes have been changed.

agalue:
Pending:

Create JUnit test to verify the registration/unregistration of the northbounders to be sure they work.

Possible Enhancement:

Check if the content of a destination was changed, and only re-register it, if the content was changed.

PR #522, opened 2016-01-04
Happy New Year 2016

jeffgdotorg:
Copyright year updates in web UI


agalue:
The header with the date exist on all the .java files (and several but not all the .jsp as well). Should we update them as well ?

jeffgdotorg:
The dates in the non-user-facing files get updated only when something else in those files is changed.

agalue:
I asked because you changed the headers I'm referring here on the JSPs involved in this pull request, but either way, that sounds good to me.

indigo423:
I bet @RangerRick has a boilerplate update script which fixes all that :) :cross finger:

RangerRick:
I have a script that gets you 90% there, but needs a lot of hand management.  :)

PR #523, opened 2016-01-04
Features/pimp graph results

gregigorius:
can someone help me? i want to make a plugin purposed to compare the result between xx to yy time with aa to bb time. 
(ex: January 3rd 2016 12 PM to January 3rd 2016 13 PM compare with January 2nd 2016 12 PM to January 2nd 2016 13 PM)
thank you


indigo423:
Hi @gregigorius,

i tried to understand your issue. What is exactly your use case? Do you try to get RRDtool time shift implemented? Like here: http://rrdtool.vandenbogaerdt.nl/shift.php

To get a good result, since Grafana 2.0 there is a time shift feature available: http://docs.grafana.org/v2.0/guides/whats-new-in-v2/

If you want some wider feedback, feel free to discuss your topic on our opennms-discuss mailing list which you can find here: https://www.opennms.org/wiki/Mailing_lists

Thank you very much for your contribution.

PR #524, opened 2016-01-06
NMS-4433: Add alarm threshold box to start page

indigo423:
The patch for the threshold alarm box is migrated to OpenNMS Horizon. The stylesheets are migrated to use new bootstrap theme. The logic executing native Java code in the JSP is refactored to a ThresholdBoxController and the box just represents the view. The icon images are normalized.

Added configuration parameter to opennms.properties to allow enabling and disabling the threshold alarm box. Allow to configure the UEI part how threshold events are evaluated.

JIRA: http://issues.opennms.org/browse/NMS-4433
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS353


opennms-webapp/src/main/java/org/opennms/web/controller/alarm/ThresholdBoxController.java
```diff
@@ -0,0 +1,135 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected ModelAndView handleRequestInternal(HttpServletRequest request, HttpServletResponse response) throws Exception {
+
+        int rows = DEFAULT_ROWS;
+
+        if (System.getProperty(THRESHOLD_ALARM_COUNT_PROPERTY) != null) {
```
j-white: This block can be replaced with:
int rows = Integer.getInteger(THRESHOLD_ALARM_COUNT_PROPERTY, DEFAULT_ROWS);

indigo423: Thank you very much :+1:

soleger: Aren't the Boolean.getBoolean(), Integer.getInteger() and Long.getLong() methods considered a notorious landmine in the Java API? Sonar probably has a rule that flags any usage of them. :)

http://konigsberg.blogspot.com/2008/04/integergetinteger-are-you-kidding-me.html

I like the Boolean.getBoolean("true") == Boolean.FALSE example.

indigo423: As @j-white mentioned, I use now the signature of getInteger(String nm, int val) to have the possibility if the user set property is not found, the default value is assigned. What is your suggestion to make this better?

agalue:
A limitation I remember from the original implementation is that it is impossible to assume that all the threshold related events are going to have a unified UEI (or share the same prefix). How are you going to identify threshold events ?

indigo423:
You can configure the prefix now with a property: 
``# opennms.thresholdAlarm.uei.filter = "uei.opennms.org/threshold/``

There is some unconfigurable convention left for the specific threshold types:

- relativeChangeExceeded
- absoluteChangeExceeded
- highThresholdExceeded
- lowThresholdExceeded

It is possible to make those also customizable through an property?

Btw.: This is one of the examples why the current UI doesn't give you good "monitoring information", you can do theoretical everything you want but practical not :)

agalue:
I would say it is better to have a more complex way for matching, because you're assuming that all the UEIs will have the same prefix, and I can bet whatever you want that this won't be the case on any average OpenNMS customer ;)

That being said, you can use a set of SPEL Expressions and start evaluating them in order. If one matches, then the event is accepted. As an example on how to use SPEL, check out my enhancements on the Syslog Northbounder Implementation.

indigo423:
Investigate to set the events-source field in https://github.com/OpenNMS/opennms/blob/develop/opennms-services/src/main/java/org/opennms/netmgt/threshd/AbstractThresholdEvaluatorState.java

indigo423:
no way for me to get this feature merged.

PR #525, opened 2016-01-07
NMS-8068: Add support for marshalling NorthboundAlarms to XML

j-white:
JIRA: http://issues.opennms.org/browse/NMS-8068
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS368-3


agalue:
Looks good to me. I would say: ship it !!!!

And ship it soon, as I have to update the other northbounder implementation prior merging them after the change on the ipAddr field ;)

PR #528, opened 2016-01-08
BSM-86: adding/removing reduction keys in the Web UI

christianpape:
jira: http://issues.opennms.org/browse/BSM-86
bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS385


christianpape:
Sorry, wrong target branch. Created another PR #529 against BSM-14.

PR #529, opened 2016-01-08
BSM-86: adding/removing reduction keys in the Web UI

christianpape:
jira: http://issues.opennms.org/browse/BSM-86
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS385


pbrane:
Christian, just checking to see if you saw my comments/changes to the BSM-86 issue.

christianpape:
Yes, I've seen the changes. The reduction keys are simple strings that are associated with business services and this is of course a many-to-many relationship, so multiple reduction keys can be associated with a single business service. The different roles, their operations or permissions on the hierarchy is not in the scope of this sprint.

PR #531, opened 2016-01-08
BSM-93: Allow Applications to be selected in the Topology UI

j-white:
JIRA: http://issues.opennms.org/browse/BSM-93
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS384


features/topology-map/plugins/org.opennms.features.topology.plugins.topo.application/src/main/java/org/opennms/features/topology/plugins/topo/application/browsers/ApplicationSelectionLinkGenerator.java
```diff
@@ -67,16 +67,21 @@ public Object generateCell(final Table source, final Object itemId, Object colum
...
-							if (applicationCriteria == null) {
-								applicationCriteria = new ApplicationCriteria();
-								graphContainer.addCriteria(applicationCriteria);
-							}
-							if (!Objects.equals(applicationCriteria.getApplicationId(), String.valueOf(idProperty.getValue()))) {
-								applicationCriteria.setApplicationId(String.valueOf(idProperty.getValue()));
-								graphContainer.setDirty(true);
-								graphContainer.redoLayout();
-							}
+                        UI ui = UI.getCurrent();
```
mvrueden: Can you please add a comment why you are doing this instead of using the graphContainer.

PR #537, opened 2016-01-11
NMS-7978: Add threshold description and formatting XML files

indigo423:
This is a proposed patch that does the following:

1. thresholds.xml
  * Add descriptions to all thresholds to add documentation on what each threshold is for
  * Change whitespace and ordering to match how the OpenNMS web UI generates the file; I used the web GUI to open each threshold in the editor and then saved it WITH NO CHANGES (other than the description) to get OpenNMS just to rewrite the file formatting. The issue here is it is much easier to diff a new version of the file with the one in production if the whitespace and XML formatting match how the GUI edits the file.

2. eventconf.xml
  * Added a comment about file ordering for documentation purposes
  * Remove extraneous blank lines

JIRA: http://issues.opennms.org/browse/NMS-7978
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS406-1

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/thresholds.xml
+++ b/opennms-base-assembly/src/main/filtered/etc/thresholds.xml
@@ -1,70 +1,123 @@
-<?xml version="1.0"?>
-<thresholding-config>
-        <group name="mib2"
-                rrdRepository = "${install.share.dir}/rrd/snmp/">
-                <threshold type="high" ds-name="tcpInErrors" ds-type="node" ds-label="" value="1" rearm="0" trigger="1"/>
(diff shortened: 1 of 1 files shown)


indigo423:
@TimMillerDyck I have removed to eventconf.xml changes for two reasons: 

- don't give unnecessary merge conflicts for update from 17.0.0 to 17.0.1
- We try to avoid comments in XML files, cause more of them get generated through the UI and ReST API, comments get removed and introduce update conflicts

Does this make sense to you?

Thanks for your contribution!

TimMillerDyck:
@indigo423 Hi Ronny, thanks for your revisions and for merging the patch! I get the point about XML comments being lost on file rewriting down the road. I'm just big on comments. :)
Regards, Tim

PR #541, opened 2016-01-11
BSM-88: Implement alarm table polling functionality

mvrueden:
* JIRA: http://issues.opennms.org/browse/BSM-88
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS388

Please note that with this PR issue http://issues.opennms.org/browse/BSM-79 is also fixed.


opennms-base-assembly/src/main/filtered/etc/create.sql
```diff
@@ -2666,3 +2666,13 @@ CREATE TABLE bsm_service_ifservices (
   REFERENCES bsm_service (id) ON DELETE CASCADE
 );
 
+CREATE TABLE bsm_service_children (
```
j-white: Why is this necessary?

mvrueden: Actually we don't.

PR #542, opened 2016-01-12
BSM-78: Add possibility to reload the Business State Machine

mvrueden:
* JIRA: http://issues.opennms.org/browse/BSM-78
* BAMBOO: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS410


j-white:
I think we should also add the ability to the reload the state machine via the REST API.

mvrueden:
:+1:

PR #543, opened 2016-01-12
NMS-8078: Updated to APC PowerNet MIB 4.1.6 events

StCyr:
New PR against release-17.0.1

Previous one (https://github.com/OpenNMS/opennms/pull/385) was against develop.

Caution:
=======

This PR adds about about 200 alarms to the ~2000 already existing alarm configurations. Maybe some aren't worth adding.

I think it would be good if some other users could test this event file on their OpenNMS server to see if their alarm dashboards don't get flooded by new uninteresting alarms.


indigo423:
We have reopened this issue to target it for OpenNMS Horizon 18.

PR #548, opened 2016-01-13
HZN-570: Expose the NBI's configuration through ReST

agalue:
Besides implementing the new ReST end points, several enhancements have been added to the implementations of the NBI's configuration objects, besides other refactoring to simplify the solution and de-duplicate code.

JIRA: http://issues.opennms.org/browse/HZN-570
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS370


j-white:
Would it be possible to document the new endpoints in the REST section of the development guide?

j-white:
The default values don't appear to be handled properly in the classes that were converted from castor.See https://www.opennms.org/wiki/Converting_a_Castor_Class_to_JAXB#Dealing_with_Defaults.

agalue:
Sure! I can update the docs. 

About the defaults: that wiki page is outdated. That is not required anymore, and my JUnit tests validate that the way I'm handling the defaults is correct and works as expected. 

I can add explicit tests to validate the defaults if that helps.

j-white:
I agree that it will work, but the code now has the following caveats:
* The generated has* and delete* functions don't make sense anymore.
* The serialized XML will always include the default values, which differs from the previous behavior.

agalue:
Ahhhh, you're right.

Those are "side effects" of my solution, but that doesn't mean the solution is incorrect. That being said, do you think it worth invest time modifying the code just to avoid these side effects ?

agalue:
I added the develop docs for the ReST end-points.

Please let me know if I should invest time dealing with the defaults on the code.

agalue:
When Bamboo goes green, it is safe to merge this pull request:

http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS370/latest

j-white:
I would fix the defaults now since we're unlikely to come back to it. Your call.

agalue:
That was a lot of code changes, so let's wait for Bamboo to be sure that I'm not missing something.

pbrane:
> On Jan 13, 2016, at 5:03 PM, Alejandro Galue <notifications@github.com> wrote:
> 
> That was a lot of code changes, so let's wait for Bamboo to be sure that I'm not missing something.
> 
Thanks, Alejandro!


David

agalue:
@pbrane  You're very welcome!

There are still some busted stuff because of the spring context that I have to fix, but other than that, everything seems to be good.

If anyone has more concerns or things I should change please let me know.

PR #552, opened 2016-01-20
NMS-8092: Add OpenWrt syslog and related event definitions

TimMillerDyck:
This is a pull request for the addition of syslog messages and related events for OpenWrt (a Linux distribution for routers and other embedded devices) and various OpenWrt packages that also report status through syslog.

JIRA: http://issues.opennms.org/browse/NMS-8092
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS426-2

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/syslog/OpenWrt.syslog.xml
+++ b/opennms-base-assembly/src/main/filtered/etc/syslog/OpenWrt.syslog.xml
@@ -1,0 +1,108 @@
+<?xml version="1.0"?>
+
+<!--
+
+OpenWrt (the Linux distribution) and other OpenWrt software package syslog messages
+
+-->
+
+<syslogd-configuration-group>
+    <ueiList>
+
+        <!-- core OS syslog messages -->
+
+        <!-- system startup -->
+        <ueiMatch>
+            <!-- OpenWrt 12.09 version: this ifup message is used as proxy for system startup as the more relevant system startup log messages appear before remote syslogging is working -->
+            <severity>Notice</severity>
+            <process-match expression="^ifup$" />
+            <match type="substr" expression="Enabling Router Solicitations on loopback" />
+            <uei>uei.opennms.org/vendor/openwrt/syslog/SystemStarting</uei>
+        </ueiMatch>
+        <ueiMatch>
+            <!-- OpenWrt 14.07 version: this logread message is used as proxy for system startup as the more relevant system startup log messages appear before remote syslogging has started -->
+            <severity>Emergency</severity>
+            <process-match expression="logread" />
+            <match type="substr" expression="Logread connected to" />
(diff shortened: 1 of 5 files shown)


opennms-base-assembly/src/main/filtered/etc/syslog/OpenWrt.syslog.xml
```diff
@@ -0,0 +1,108 @@
...
+            <parameter-assignment matching-group="1" parameter-name="interfaceName" />
+        </ueiMatch>
+
+        <!-- The only syslog sending mechanism OpenWrt has is to send ALL syslog messages to a remote syslog server. This results in a high
+             volume of syslog messages being received and stored in OpenNMS. To avoid this, the statement below will discard (and not save)
+             all syslog messages received from the specified devices below (other than those already captured above).
+               OpenWrt 12.09: list the device by its hostname (lowercase)
+               OpenWrt 14.07 and later: list the device by the source IP address it uses to send the syslog packet -->
+        <ueiMatch>
+            <hostname-match expression="(openwrt01|192\.0\.2\.1).*" />
```
indigo423: Is this something which should be modified by the user?

indigo423:
Really nice enhancement thank you! I've put it into bamboo to see if it goes green.

opennms-base-assembly/src/main/filtered/etc/syslog/OpenWrt.syslog.xml
TimMillerDyck: @indigo423 Hi Ronny, yes, the hostname-match expression is something the end-user sysadmin would typically modify by adding the hostname/IP address of their OpenWrt device once they have configured OpenWrt to send syslog messages to OpenNMS. The problem is that OpenWrt has no syslog filtering on its send side, so OpenNMS gets all the syslog messages from the whole OS. The hostname-match above is just to avoid the event list clutter that results in OpenNMS by dropping all syslog messages after the specific desired ones have already been handled. -Tim

indigo423:
@TimMillerDyck I've upgraded the bamboo branch to latest develop and added the configuration file to the integration test.

TimMillerDyck:
@indigo423 Hi Ronny, I just added a similar syslog test exclusion for the OpenWrt pull. Thanks. -Tim

jeffgdotorg:
Looks good, @indigo423's comments about the hostname thing notwithstanding.

TimMillerDyck:
Hi Jeff, would a note in the release readme be the right place to highlight this, for those that want to use this functionality? I could write one if desired. It’s really a lack of selectivity in OpenWrt’s syslog functionality that is the issue, not anything with OpenNMS.

 

Regards,

Tim

TimMillerDyck:
@indigo423 Hi Ronny, checking in to see if there anything else I should do regarding this patch submission. Thanks, Tim

PR #553, opened 2016-01-21
BSM-95: Create an Ops Board Dashlet for Business Services

christianpape:
Jira: http://issues.opennms.org/browse/BSM-95
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS425


features/bsm/service/api/src/main/java/org/opennms/netmgt/bsm/service/BusinessServiceSearchCriteria.java
```diff
@@ -0,0 +1,49 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with OpenNMS(R).  If not, see:
+ * http://www.gnu.org/licenses/
+ *
+ * For more information contact:
+ * OpenNMS(R) Licensing <license@opennms.org>
+ * http://www.opennms.org/
+ * http://www.opennms.com/
```
mvrueden: The url must be idented. Please check other license headers to match.

features/bsm/service/api/src/main/java/org/opennms/netmgt/bsm/service/BusinessServiceSearchCriteriaBuilder.java
```diff
@@ -0,0 +1,302 @@
...
+     */
+    private Order m_order = Order.Name;
+    /**
+     * the results limit
+     */
+    private int m_limit = 0;
+    /**
+     * ascending or descending?
+     */
+    private boolean m_ascending = true;
```
mvrueden: Can you make this an enum?

features/vaadin-dashlets/dashlet-bsm/src/main/java/org/opennms/features/vaadin/dashboard/dashlets/BSMConfigHelper.java
```diff
@@ -0,0 +1,146 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with OpenNMS(R).  If not, see:
+ * http://www.gnu.org/licenses/
+ *
+ * For more information contact:
+ * OpenNMS(R) Licensing <license@opennms.org>
+ * http://www.opennms.org/
+ * http://www.opennms.com/
+ *******************************************************************************/
```
mvrueden: License header does not match default header (urls)

features/vaadin-dashlets/dashlet-bsm/src/main/java/org/opennms/features/vaadin/dashboard/dashlets/BSMConfigHelper.java
```diff
@@ -0,0 +1,146 @@
...
+     * Returns a boolean value for a given key from the map where "1", "true" and "yes"
+     * are interpeted as boolean true, otherwise false.
+     *
+     * @param map the map to be used
+     * @param key the key
+     * @return the boolean value
+     */
+    public static boolean getBooleanForKey(Map<String, String> map, String key) {
+        String value = map.get(key);
+        return ("true".equals(value) || "yes".equals(value) || "1".equals(value));
```
mvrueden: why do the values "true", "yes" and "1" equal true?

features/vaadin-dashlets/dashlet-bsm/src/main/java/org/opennms/features/vaadin/dashboard/dashlets/BSMConfigHelper.java
```diff
@@ -0,0 +1,146 @@
...
+     * @param map the map to be used
+     * @param key the key
+     * @return the int value, 0 if not parsable
+     */
+    public static int getIntForKey(Map<String, String> map, String key) {
+        String value = map.get(key);
+        try {
+            return Integer.parseInt(value);
+        } catch (NumberFormatException e) {
+            return 0;
```
mvrueden: We really should not do this. If the value is not an int, but you expect an int, throw an exception and deal with that exception in the UI. The helper should not return 0 in case of an error.

features/vaadin-dashlets/dashlet-bsm/src/main/java/org/opennms/features/vaadin/dashboard/dashlets/BSMConfigHelper.java
```diff
@@ -0,0 +1,146 @@
...
+
+        boolean filterByAttribute = getBooleanForKey(map, "filterByAttribute");
+        String attributeKey = getStringForKey(map, "attributeKey");
+        String attributeValue = getStringForKey(map, "attributeValue");
+
+        boolean filterBySeverity = getBooleanForKey(map, "filterBySeverity");
+
+        String severityValue = getStringForKey(map, "severityValue");
+
+        if (severityValue == null || "".equals(severityValue)) {
```
mvrueden: Please use Strings.isEmptyOrNull(..)

features/vaadin-dashlets/dashlet-bsm/src/main/java/org/opennms/features/vaadin/dashboard/dashlets/BSMConfigurationWindow.java
```diff
@@ -0,0 +1,432 @@
...
+                    m_dashletSpec.getParameters().put("orderBy", BusinessServiceSearchCriteriaBuilder.Order.Name.name());
+                }
+
+                if (m_ascDesc.getValue() != null) {
+                    m_dashletSpec.getParameters().put("orderAsc", "Ascending".equals(m_ascDesc.getValue())?"true":"false");
+                } else {
+                    m_dashletSpec.getParameters().put("orderAsc", "true");
+                }
+
+                m_dashletSpec.getParameters().put("resultsLimit", m_limitTextField.getValue().toString());
```
mvrueden: The generation of the spec looks wrong. 
Most of the code looks the same and should be generalized a bit more.

features/vaadin-dashlets/dashlet-bsm/src/main/java/org/opennms/features/vaadin/dashboard/dashlets/BSMDashlet.java
```diff
@@ -0,0 +1,219 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with OpenNMS(R).  If not, see:
+ * http://www.gnu.org/licenses/
+ *
+ * For more information contact:
+ * OpenNMS(R) Licensing <license@opennms.org>
+ * http://www.opennms.org/
+ * http://www.opennms.com/
+ *******************************************************************************/
```
mvrueden: The license header does not match the default. The urls are not idented properly.

features/vaadin-dashlets/dashlet-bsm/src/main/java/org/opennms/features/vaadin/dashboard/dashlets/BSMDashlet.java
```diff
@@ -0,0 +1,219 @@
...
+     *
+     * @param dashletSpec            the {@link DashletSpec} to be used
+     * @param businessServiceManager the {@link BusinessServiceManager} to be used
+     */
+    public BSMDashlet(String name, DashletSpec dashletSpec, BusinessServiceManager businessServiceManager) {
+        super(name, dashletSpec);
+        /**
+         * Setting the member fields
+         */
+        m_businessServiceManager = businessServiceManager;
```
mvrueden: This should be initialized with TransactionAwareProxyFactoryBean.

features/vaadin-dashlets/dashlet-bsm/src/main/java/org/opennms/features/vaadin/dashboard/dashlets/BSMDashletFactory.java
```diff
@@ -0,0 +1,87 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2013-2014 The OpenNMS Group, Inc.
```
mvrueden: This class is newly created in 2016. Please use 2016 here.

features/vaadin-dashlets/dashlet-bsm/src/main/java/org/opennms/features/vaadin/dashboard/dashlets/BSMDashletFactory.java
```diff
@@ -0,0 +1,87 @@
...
+ * OpenNMS(R) is Copyright (C) 1999-2014 The OpenNMS Group, Inc.
```
mvrueden: This should be 1999-2016.

features/vaadin-dashlets/dashlet-bsm/src/main/java/org/opennms/features/vaadin/dashboard/dashlets/BSMConfigHelper.java
christianpape: Because the config file can also be modified by hand. So, true/yes/1 will result in a boolean true.

fooker: Then you should ad "on" to the list and use equalsIgnoreCase(...).

features/vaadin-dashlets/dashlet-bsm/src/main/java/org/opennms/features/vaadin/dashboard/dashlets/BSMConfigHelper.java
christianpape: This method will only be used in the case for the results limit in this dashlet. A value of 0 means no results limit. So, it is ok to not limit the results in the case of a parse error.

features/bsm/service/impl/src/main/java/org/opennms/netmgt/bsm/service/internal/BusinessServiceManagerImpl.java
```diff
@@ -294,4 +295,13 @@ private OnmsMonitoredService getMonitoredService(Integer serviceId) throws NoSuc
         }
         return monitoredService;
     }
+
+    @Override
+    public List<BusinessService> search(BusinessServiceSearchCriteria businessServiceSearchCriteria) {
+        if (businessServiceSearchCriteria == null) {
```
mvrueden: Please use `Objects.requireNonNull(businessSearchCriteria)`

features/bsm/service/api/src/main/java/org/opennms/netmgt/bsm/service/BusinessServiceSearchCriteriaBuilder.java
```diff
@@ -0,0 +1,318 @@
...
+
+    /**
+     * Default constructor
+     */
+    public BusinessServiceSearchCriteriaBuilder() {
+    }
+
+    @Override
+    public List<BusinessService> apply(BusinessServiceManager businessServiceManager, List<BusinessService> businessServiceDTOs) {
+        Stream<BusinessService> s = businessServiceDTOs.stream();
```
mvrueden: Can we change the behaviour that it uses "businessServiceManager.findMatching(Criteria)" instead.
This method should then convert the BusinessServiceSearchCriteria to a general Criteria, which then can be used by the underlying BusinessServiceDAO to filter at the database level.

features/vaadin-dashlets/dashlet-bsm/src/main/java/org/opennms/features/vaadin/dashboard/dashlets/BSMDashlet.java
```diff
@@ -0,0 +1,220 @@
...
+                    Page.getCurrent().getStyles().add(".severity.Normal{ color:#000000; background-color: #D7E1CD; border-top: 2px solid #336600 !important; background-image: url(images/bgNormal.png); }");
+                    Page.getCurrent().getStyles().add(".severity.Cleared { color:#000000; background-color: #EEEEEE; border-top: 2px solid #999999 !important; background-image: url(images/bgCleared.png); }");
+                    Page.getCurrent().getStyles().add(".severity.bright { color:#000000; padding-left: 20px; background-position: top left; background-repeat: repeat-y; }");
+                    Page.getCurrent().getStyles().add(".severity.nobright { color:#000000; background-image: none;}");
+                }
+
+                @Override
+                public void refresh() {
+                    m_gridLayout.removeAllComponents();
+                    injectStyles();
```
j-white: The same styles get injected on the page every time the dashlet is refreshed. Since these are never removed, the DOM just keeps growing with copies of the same styles every refresh.

(Many of the other existing dashlets have this same problem.)

features/vaadin-dashlets/dashlet-bsm/src/main/java/org/opennms/features/vaadin/dashboard/dashlets/BSMDashlet.java
```diff
@@ -0,0 +1,220 @@
...
```
j-white: Similar problem with injectStyles() above.

features/bsm/service/api/src/main/java/org/opennms/netmgt/bsm/service/BusinessServiceSearchCriteriaBuilder.java
j-white: +1

j-white:
I tried it out and it's working great. Thanks @christianpape.

features/bsm/service/api/src/main/java/org/opennms/netmgt/bsm/service/BusinessServiceSearchCriteriaBuilder.java
christianpape: Due to the acceptance criterias of this issue it is required to filter by regular expressions and to filter by Business Service state. So, for now I cannot use the criteria API to accomplish this. If we want to change this we have to change the BSM model and criteria API too.

j-white: Understood. How about we relax the requirement to use a LIKE statement, which is supported by the existing Criteria API? If more advanced filtering is required, the user can resort to setting and using the attribute fields.

mvrueden: We could do that, but then we cannot filter for the severity/status of the Business Services as we do not persist it. I'd say let's not worry about it at the moment as we have to come back to address this later anyways.

pbrane: > On Jan 27, 2016, at 3:06 AM, mvrueden <notifications@github.com> wrote:
> 
> We could do that, but then we cannot filter for the severity/status of the Business Services as we do not persist it. I'd say let's not worry about it at the moment as we have to come back to address this later anyways.
> 
This issue effects all the ops boards, right?

j-white: Gotcha. Alright, let's revisit this later.

I believe that only this dashlet is affected. We're trying to find a elegant way of filtering Business Services through the service layer.

j-white:
We'll merge this once Bamboo is fixed the branch goes green again.

features/bsm/service/api/src/main/java/org/opennms/netmgt/bsm/service/BusinessServiceSearchCriteriaBuilder.java
mvrueden: NOTE: Other ops boards are not affected as they all use the Criteria objects and therefore use the underlying DAOs to filter.

PR #555, opened 2016-01-28
NMS-8097: Fix Oracle Java install guide

indigo423:
It is not possible since Java 1.8.0 b72 to install with add-apt-repository. Updated the install guide to create source.list file manually.

JIRA: http://issues.opennms.org/browse/NMS-8097
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS435


indigo423:
@j-white does it make sense to fix this bug in foundation-2016?

PR #559, opened 2016-02-02
BSM-24: Users can adjust the default status propagation logic for each BS in the BSM hierachy (Map and Reduce)

mvrueden:
* JIRA: http://issues.opennms.org/browse/BSM-24
* BAMBOO: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS416


j-white:
There are a lot of changes here, but generally it looks good. I think we're safe to merge once Bamboo goes green with the changes in 6fee83d.

PR #560, opened 2016-02-02
BSM-92: Get a quick overview of the existing feature sets of the Business Services feature

mvrueden:
* JIRA: http://issues.opennms.org/browse/BSM-92
* Bamboo: not required (only documentation changes)


j-white:
Looks great.

PR #563, opened 2016-02-03
BSM-120: Include "interfaceDown" events in status calculation of IP Service Edges

mvrueden:
* JIRA: http://issues.opennms.org/browse/BSM-120
* BAMBOO: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS445


j-white:
We need also need add:
  * A new verifyReductionKey() statement in org.opennms.netmgt.bsm.daemon.Bsmd
  * A reference to the reduction key in the documentation (the other two are already referenced)

mvrueden:
:+1:

PR #566, opened 2016-02-03
HZN-584: Initialize WS-Management collector documentation

indigo423:
Build structure for performance data collection. The structure follows the same pattern as in Service Assurance. The description for the WS-Man detector is separated in a dedicated file to allow easier refactoring if we have a structure for Provisioning and all Detectors.


indigo423:
Wrong target.

PR #571, opened 2016-02-04
BSM-130: Business Service Table cannot be sorted by id or name

mvrueden:
* JIRA: http://issues.opennms.org/browse/BSM-130
 * BAMBOO: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS452


mvrueden:
Yes makes totally sense. Changed to LinkedHashSet.

mvrueden:
As we are close to fixing BSM-104 I'd say lets not fix this pull request and wait for BSM-104.

PR #574, opened 2016-02-04
BSM-131: BSM Ops Board status elements do not use 100% height if other elements use more than 1 row

christianpape:
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS454
Jira: http://issues.opennms.org/browse/BSM-131


mvrueden:
I tried it and it works fine. I'd say lets merge this as soon as Bamboo goes green. (ASABGG)

PR #577, opened 2016-02-07
NMS-8112: init script "configtest" exit value is always 1

bobwest:
Propagate return value from checkXmlFiles() to init script exit value,
so "configtest" exits with 0 for success and non-zero for failure.

### DIFF
--- a/opennms-base-assembly/src/main/filtered/bin/opennms
+++ b/opennms-base-assembly/src/main/filtered/bin/opennms
@@ -851,8 +851,16 @@ case "$COMMAND" in
 
 	configtest)
 		$opennms_echo "Validating XML files: \c"
-		checkXmlFiles || exit $?
-		echo "PASSED"
+		checkXmlFiles
+		ret=$?
+		case $ret in
+			0)
+				echo "PASSED"
+				;;
+			*)
+				# nothing
+				;;
+		esac
 		;;
 
 	pause)


opennms-base-assembly/src/main/filtered/bin/opennms
```diff
@@ -851,8 +851,16 @@ case "$COMMAND" in
...
-		checkXmlFiles || exit $?
-		echo "PASSED"
+		checkXmlFiles
+		ret=$?
+		case $ret in
+			0)
+				echo "PASSED"
+				;;
+			*)
+				# nothing
```
indigo423: Should we give a warning here for the reason the checkXmlFiles did not pass?

indigo423:
Thanks for spotting this.

opennms-base-assembly/src/main/filtered/bin/opennms
bobwest: checkXmlFiles() itself is quite verbose when XML errors are identified.  I'm not sure that distinguishing between return values 3 and 6 would give any more meaningful information for the user -- they both mean that there were one or more XML syntax errors.

indigo423:
@RangerRick from my point of view we should target this fix for foundation-2016 to have this issue fixed in 17.0.1.

PR #583, opened 2016-02-09
HZN-559: Refactoring of syslogd

Delmonte3161:
JIRA: http://issues.opennms.org/browse/HZN-559
Bamboo: ?


features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/SyslogReceiverCamelNettyImpl.java
```diff
@@ -123,38 +128,63 @@ public void stop() throws InterruptedException {
...
                     );
-
+                    
                     from(from)
                     //.convertBodyTo(java.nio.ByteBuffer.class)
                     .process(new Processor() {
                         @Override
                         public void process(Exchange exchange) throws Exception {
-                            ChannelBuffer buffer = exchange.getIn().getBody(ChannelBuffer.class);
+                            ChannelBuffer buffer = ChannelBuffers.copiedBuffer(exchange.getIn().getBody().toString().getBytes());
```
soleger: Is there a reason that this change is necessary? I think that you can keep the original line since the message body should already be a ChannelBuffer.

MS043660: When we try to get the exchange body and try to provide to syslogconnection we end up in Null pointer exception,Since body type is of String and exchange.getIn.getBody returns an object which is null since we apply Channel buffer class on it,while in the above case i take the string in body and construct the ChannelBuffer from string.

soleger:
These changes have been merged into the OpenNMS/opennms:feature-refactor-syslogd branch. Closing PR.

PR #584, opened 2016-02-10
BSM-132: Cannot Edit an Edge after creation

christianpape:
Jira: http://issues.opennms.org/browse/BSM-132
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS462

### DIFF
--- a/features/bsm/vaadin-adminpage/src/main/java/org/opennms/netmgt/bsm/vaadin/adminpage/BusinessServiceEdgeEditWindow.java
+++ b/features/bsm/vaadin-adminpage/src/main/java/org/opennms/netmgt/bsm/vaadin/adminpage/BusinessServiceEdgeEditWindow.java
@@ -28,12 +28,18 @@
 
 package org.opennms.netmgt.bsm.vaadin.adminpage;
 
-import com.google.common.collect.Ordering;
+import java.util.Collection;
+import java.util.Set;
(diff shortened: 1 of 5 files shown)


features/bsm/vaadin-adminpage/src/main/java/org/opennms/netmgt/bsm/vaadin/adminpage/BusinessServiceEdgeEditWindow.java
```diff
@@ -312,6 +318,37 @@ public BusinessServiceEdgeEditWindow(final BusinessService businessService,
...
+                    }
+
+                    case REDUCTION_KEY: {
+                        Set<ReductionKeyEdge> edges = businessService.getReductionKeyEdges();
+                        edges.remove(edge);
+                        businessService.setReductionKeyEdges(edges);
+                        break;
+                    }
+                }
+            }
```
mvrueden: Can we move this to the BusinessServiceManager or BusinessService object? 
How about creating a `getEdges(EdgeType)` and `setEdges(EdgeType, Collection)` method?

mvrueden:
Please also provide a smoke test to verify that you really can edit an edge.
I see two test cases:
1. edit a already persisted edge
2. edit a transient (non persisted) edge

PR #585, opened 2016-02-10
BSM-133: Hierarchy layout breaks when two different "hierarchies" depend on the same reduction key

mvrueden:
* JIRA: http://issues.opennms.org/browse/BSM-133
 * BAMBOO: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS464

### DIFF
--- a/features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/jung/HierarchyLayout.java
+++ b/features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/jung/HierarchyLayout.java
@@ -48,8 +48,8 @@ import edu.uci.ics.jung.graph.Graph;
 
 /**
  * A custom layout to be able to draw a hierarchy.
- * This implementation is inspired by {@link edu.uci.ics.jung.algorithms.layout.TreeLayout}.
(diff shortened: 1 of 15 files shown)


features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/jung/HierarchyLayout.java
```diff
@@ -0,0 +1,213 @@
...
+import com.google.common.base.Preconditions;
+
+import edu.uci.ics.jung.algorithms.layout.Layout;
+import edu.uci.ics.jung.graph.Forest;
+import edu.uci.ics.jung.graph.Graph;
+
+/**
+ * A custom layout to be able to draw a hierarchy.
+ * This implementation is inspired by {@link edu.uci.ics.jung.algorithms.layout.TreeLayout}.
+ * Unfortunately this implementation only works if 2 parents do not have the same child (see {@link edu.uci.ics.jung.graph.util.TreeUtils#getRoots(Forest)}).
```
j-white: Is this still true?

mvrueden: I am refering to the `TreeLayout` and not `this`. 
Maybe that is not clear enough.

PR #587, opened 2016-02-10
NMS-6404: Add support for publishing events on the bus via REST.

j-white:
Included a sample python script in contrib/ that can be used to generate
and publish events via REST.

JIRA: http://issues.opennms.org/browse/NMS-6404
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS466-2

### DIFF
--- a/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/EventRestService.java
+++ b/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/EventRestService.java
@@ -34,6 +34,7 @@ import java.util.Date;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.FormParam;
 import javax.ws.rs.GET;
+import javax.ws.rs.POST;
 import javax.ws.rs.PUT;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
@@ -51,6 +52,7 @@ import org.joda.time.format.ISODateTimeFormat;
 import org.opennms.core.criteria.Alias.JoinType;
 import org.opennms.core.criteria.CriteriaBuilder;
 import org.opennms.netmgt.dao.api.EventDao;
+import org.opennms.netmgt.events.api.EventIpcManager;
 import org.opennms.netmgt.model.OnmsEvent;
 import org.opennms.netmgt.model.OnmsEventCollection;
 import org.opennms.web.rest.support.MultivaluedMapImpl;
@@ -66,6 +68,9 @@ public class EventRestService extends OnmsRestService {
 
     @Autowired
     private EventDao m_eventDao;
+
+    @Autowired
+    private EventIpcManager m_eventForwarder;
 
     /**
      * <p>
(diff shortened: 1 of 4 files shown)


jeffgdotorg:
Tests?

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/EventRestService.java
```diff
@@ -258,6 +263,14 @@ private void processEventAck(final SecurityContext securityContext, final OnmsEv
         m_eventDao.save(event);
     }
 
+    @POST
+    @Consumes({MediaType.APPLICATION_XML, MediaType.APPLICATION_JSON, MediaType.APPLICATION_ATOM_XML})
+    @Transactional
+    public Response publishEvent(final org.opennms.netmgt.xml.event.Event event) {
+        m_eventForwarder.sendNow(event);
```
agalue: It might be a good idea if you add the sendNow between try/catch and then build an appropriate response object on the catch if something wrong happens.

j-white:
Tests?

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/EventRestService.java
```diff
@@ -67,6 +69,9 @@
     @Autowired
     private EventDao m_eventDao;
 
+    @Autowired
+    private EventIpcManager m_eventForwarder;
```
soleger: You might change this to be an EventForwarder type instead of the full EventIpcManager super-interface.

j-white: Autowiring the EventForwarder fails because of multiple implementations.
Adding a Qualifier fixes this, but causes the tests to fail.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/EventRestService.java
j-white: sendNow will only throw an exception if we get a RejectedExecutionException from the thread pool, in which case the default behavior of a 500 is probably suitable.

Did you have something specific in mind?

agalue: That's not the way I would do it, but guess it should work. It is up to you. Check how I implemented a similar thing on the ReST API for NBIs.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/EventRestService.java
agalue: I typically use EventProxy. Check how I did it for the NBIs:

https://github.com/OpenNMS/opennms/blob/foundation-2016/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/config/SyslogNorthbounderConfigurationResource.java

PR #591, opened 2016-02-11
BSM-110: Remove the Business Services Masterpage

mvrueden:
* JIRA: http://issues.opennms.org/browse/BSM-110
 * BAMBOO: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS471


mvrueden:
Forgot to delete the smoke tests....

PR #594, opened 2016-02-15
BSM-143: Add "Highest Severity Above" reduction function

mvrueden:
* JIRA: http://issues.opennms.org/browse/BSM-143
 * BAMBOO: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS475


j-white:
That's a lot of changes for adding a single reduction function ;)

PR #596, opened 2016-02-16
BSM-144: Update the header of the Ops Board to use the same styling as the rest of the application

christianpape:
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS479
Jira: http://issues.opennms.org/browse/BSM-144

I did an implementation without using the OnmsHeaderProvider because 
the Ops Board/Panel has its own input fields and make no use of the original 
OpenNMS menu bar.

### DIFF
--- a/features/vaadin-dashboard/src/main/java/org/opennms/features/vaadin/dashboard/ui/HeaderLayout.java
+++ b/features/vaadin-dashboard/src/main/java/org/opennms/features/vaadin/dashboard/ui/HeaderLayout.java
@@ -36,12 +36,12 @@ import org.opennms.features.vaadin.dashboard.ui.wallboard.WallboardView;
 import com.vaadin.data.Property;
 import com.vaadin.navigator.View;
 import com.vaadin.navigator.ViewChangeListener;
-import com.vaadin.server.ExternalResource;
 import com.vaadin.shared.ui.MarginInfo;
+import com.vaadin.shared.ui.label.ContentMode;
 import com.vaadin.ui.Alignment;
 import com.vaadin.ui.Button;
 import com.vaadin.ui.HorizontalLayout;
-import com.vaadin.ui.Link;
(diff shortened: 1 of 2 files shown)


features/vaadin-dashboard/src/main/java/org/opennms/features/vaadin/dashboard/ui/HeaderLayout.java
```diff
@@ -59,17 +65,19 @@ public HeaderLayout() {
...
         setSpacing(true);
         setWidth("100%");
+        setHeight(64, Unit.PIXELS);
 
         /**
          * Adding the logo
          */
         Link link = new Link(null, new ExternalResource("/opennms/index.jsp"));
-        link.setIcon(new ThemeResource("img/logo.png"));
+        link.setIcon(new ExternalResource("/opennms/images/horizon_logo.svg"));
```
mvrueden: Can you set the icon based on css? 
This is definetly causing trouble for a Meridian release.

christianpape: It's defined in a ftl-template and not in a scss/css-File. This is a general problem and cannot be addressed in this issue.

j-white:
The problem with this approach is that it makes it difficult to maintain alternate branding (i.e. Meridian).

Can't we just make the necessary changes to the OnmsHeaderProvider API?

christianpape:
The logo is now configurable in the dashboard's theme CSS-file. Is this okay? I've really no clue how to insert Vaadin-based components in the header's HTML-fragment provided by the OnmsHeaderProvider.

j-white:
Yeah, that's better. Can you re-target this patch for foundation-2016?

j-white:
I'll cherry pick this to foundation-2016 once it's merged.

PR #598, opened 2016-02-16
HZN-487: Modularized Discovery based on Apache Camel

soleger:
This branch contains a refactor of the Discovery subsystem so that it is implemented in Apache Camel and uses an ActiveMQ queue (per location) to enqueue discovery ping jobs.

Remaining work not part of this pull request: Clean up any code that has been obsoleted by this refactor and complete a set of blueprint files for running the discovery jobs on a Minion container.


j-white:
I gave this a try and it worked.

There are still some rough edges, but it looks like most of those are captured in the HZN-487 epic.

PR #599, opened 2016-02-17
BSM-145: Configure the Business Service attributes using the Admin UI

christianpape:
Jira: http://issues.opennms.org/browse/BSM-145
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS478

### DIFF
--- a/features/vaadin-components/core/src/main/java/org/opennms/netmgt/vaadin/core/KeyValueInputDialogWindow.java
+++ b/features/vaadin-components/core/src/main/java/org/opennms/netmgt/vaadin/core/KeyValueInputDialogWindow.java
@@ -1,8 +1,8 @@
 /*******************************************************************************
  * This file is part of OpenNMS(R).
  *
- * Copyright (C) 2015 The OpenNMS Group, Inc.
- * OpenNMS(R) is Copyright (C) 1999-2015 The OpenNMS Group, Inc.
+ * Copyright (C) 2016 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2016 The OpenNMS Group, Inc.
  *
  * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
  *
--- a/features/vaadin-components/core/src/main/java/org/opennms/netmgt/vaadin/core/KeyValueInputDialogWindow.java
+++ b/features/vaadin-components/core/src/main/java/org/opennms/netmgt/vaadin/core/KeyValueInputDialogWindow.java
@@ -1,0 +1,392 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2015 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2015 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License,
+ * or (at your option) any later version.
+ *
+ * OpenNMS(R) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with OpenNMS(R).  If not, see:
+ *      http://www.gnu.org/licenses/
+ *
+ * For more information contact:
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.netmgt.vaadin.core;
+
+import com.vaadin.data.Validator;
+import com.vaadin.event.ShortcutAction;
+import com.vaadin.ui.Alignment;
+import com.vaadin.ui.Button;
+import com.vaadin.ui.FormLayout;
+import com.vaadin.ui.HorizontalLayout;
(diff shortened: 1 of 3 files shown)


features/vaadin-components/core/src/main/java/org/opennms/netmgt/vaadin/core/KeyValueInputDialogWindow.java
```diff
@@ -0,0 +1,392 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2015 The OpenNMS Group, Inc.
```
mvrueden: should be 2016

features/vaadin-components/core/src/main/java/org/opennms/netmgt/vaadin/core/KeyValueInputDialogWindow.java
```diff
@@ -0,0 +1,392 @@
...
+ * OpenNMS(R) is Copyright (C) 1999-2015 The OpenNMS Group, Inc.
```
mvrueden: should be 2016

smoke-test/src/test/java/org/opennms/smoketest/BSMAdminIT.java
```diff
@@ -1,29 +1,29 @@
...
+ *      http://www.gnu.org/licenses/
+ *
  * For more information contact:
- * OpenNMS(R) Licensing <license@opennms.org>
- * http://www.opennms.org/
- * http://www.opennms.com/
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
  *******************************************************************************/
```
mvrueden: :+1:

smoke-test/src/test/java/org/opennms/smoketest/BSMAdminIT.java
```diff
@@ -116,6 +128,24 @@ public BsmAdminPageEditWindow addReductionKeyEdge(String reductionKeyText, Strin
             return this;
         }
 
+        public BsmAdminPageEditWindow addAttribute(String key, String value) throws InterruptedException {
+            newAttributeWindow()
+                    .key(key)
+                    .value(value)
+                    .confirm();
+            wait.until(ExpectedConditions.elementToBeClickable(By.id("addAttributeButton")));
```
mvrueden: This is kinda a duplicate, as we already wait in the confirm() method or is there a particular reason to wait again for another element?

smoke-test/src/test/java/org/opennms/smoketest/BSMAdminIT.java
```diff
@@ -250,6 +280,26 @@ public BsmAdminPageEdgeEditWindow confirm() {
...
+    private class BsmAdminPageAttributeEditWindow {
+        public BsmAdminPageAttributeEditWindow key(String key) throws InterruptedException {
+            findElementById("keyField").clear();
+            findElementById("keyField").sendKeys(key);
+            return this;
+        }
+
+        public BsmAdminPageAttributeEditWindow value(String value) {
+            findElementById("valueField").clear();
+            findElementById("valueField").sendKeys(String.valueOf(value));
```
mvrueden: `value` already is a String, so we do not need `String.valueOf(value)` here.

smoke-test/src/test/java/org/opennms/smoketest/BSMAdminIT.java
```diff
@@ -637,6 +687,34 @@ public void testCanEditPersistedEdge() throws InterruptedException {
...
 
+    @Test
+    public void testCanEditTransientAttribute() throws InterruptedException {
+        // create Business Service with one Attribute
+        final String serviceName = createUniqueBusinessServiceName();
+        BsmAdminPageEditWindow bsmAdminPageEditWindow = bsmAdminPage.openNewDialog(serviceName).addAttribute("foo", "bar");
+        wait.until(pageContainsText("foo=bar"));
+
+        // cancel
+        bsmAdminPageEditWindow.cancel();
```
mvrueden: Please verify afterwards that the attribute `foo = bar` has not been set and please delete the Business Service afterwards.

smoke-test/src/test/java/org/opennms/smoketest/BSMAdminIT.java
```diff
@@ -116,6 +128,24 @@ public BsmAdminPageEditWindow addReductionKeyEdge(String reductionKeyText, Strin
...
+            wait.until(ExpectedConditions.elementToBeClickable(By.id("addAttributeButton")));
+            return this;
+        }
+
+        public BsmAdminPageEditWindow editAttribute(String key, String value) throws InterruptedException {
+            getSelectWebElement("attributeList").selectByVisibleText(key);
+            editAttributeWindow()
+                    .value(value)
+                    .confirm();
+            wait.until(ExpectedConditions.elementToBeClickable(By.id("addAttributeButton")));
```
mvrueden: This is kinda a duplicate, as we already wait in the confirm() method or is there a particular reason to wait again for another element?

smoke-test/src/test/java/org/opennms/smoketest/BSMAdminIT.java
christianpape: Sorry, copy & paste mistake

mvrueden:
I tested this and found a couple of things:
 * By default "Key" is set as the default "Key". Let's not do that.
 * If you add an attribute, let's say "Key=Value", and then click `Add Attribute` again, the validator states `String must not be empty` but probably should say something like this: `Key is already defined, please use another name for Key`
 * The value can be empty. Should we enforce an input here? (`input.length() >=1?`)

christianpape:
Fixed the validators and messages but I do not enforce a non-empty input for the attribute's value. In my opinion we should not do this. An empty value has another semantic than an undefined one.

j-white:
I also think that we should allow empty values.

I did notice that the input is focused in the "Value" field after clicking on "Add Attribute". Can we change the default focus to be on the "Key" field instead?

PR #601, opened 2016-02-17
NMS-8146: "show interfaces" link doesn't remember SNMP MIB-2 attribute filter.

StCyr:
PRELIMINARY ADVISORY: I don't know exactly why it works, but it works.

You can test the fix by modifying the corresponding .jsp file on a running OpenNMS instance (and restarting it).

JIRA: http://issues.opennms.org/browse/NMS-8146
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS586-1


j-white:
I reinstated the Bamboo build, let's wait until that passes before merging.

PR #608, opened 2016-02-19
NMS-8151: Remove all Jersey artifacts from the system class-path.

j-white:
JIRA: http://issues.opennms.org/browse/NMS-8151
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS491-3


opennms-webapp-rest/src/main/webapp/WEB-INF/web.xml
```diff
@@ -53,29 +52,6 @@
...
-      <param-name>com.sun.jersey.spi.container.ContainerRequestFilters</param-name>
-      <param-value>com.sun.jersey.api.container.filter.GZIPContentEncodingFilter</param-value>
-    </init-param>
-    <init-param>
-      <param-name>com.sun.jersey.spi.container.ContainerResponseFilters</param-name>
-      <param-value>com.sun.jersey.api.container.filter.GZIPContentEncodingFilter</param-value>
-    </init-param>
-    <load-on-startup>1</load-on-startup>
-  </servlet>
-  -->
```
agalue: Does this mean that CXF supports compression by default ? Or, is that already enabled on CXF ?

j-white: Those lines were previously commented. GZIP compression is already enabled on CXF.

agalue: Excellent!

Everything else seems to be OK to me.

PR #613, opened 2016-02-24
BSM-159: Add EdgeStatusProvider for Business Services

mvrueden:
* JIRA: http://issues.opennms.org/browse/BSM-159
 * BAMBOO: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS497


features/bsm/service/impl/src/main/java/org/opennms/netmgt/bsm/service/internal/DefaultBusinessServiceStateMachine.java
```diff
@@ -356,4 +364,14 @@ public String transform(GraphEdge edge) {
             m_rwLock.readLock().unlock();
         }
     }
+
+    @Override
+    public BusinessServiceGraph getGraph() {
+        m_rwLock.readLock().lock();
+        try {
+            return m_g;
```
j-white: Can we return a copy of the graph instead of a reference? There's also no point in using a read lock when returning the reference. I'm mostly worried that callers may alter the graph in some way.

mvrueden: As discuessed, at the moment there is no easy/nice way to create a copy of the graph. 
I added a java doc comment to inform potential callers about not modifiing the graph. 
We need to re-visit this at same point in the future.

PR #615, opened 2016-02-25
BSM-156: Add friendly names for reduction key edges

christianpape:
Jira: http://issues.opennms.org/browse/BSM-156
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS503


features/bsm/service/impl/src/main/java/org/opennms/netmgt/bsm/service/internal/edge/ChildEdgeImpl.java
```diff
@@ -70,4 +70,9 @@ public String toString() {
                 .add("child", getChild() == null ? null : getChild().getId())
                 .toString();
     }
+
+    @Override
+    public String getFriendlyName() {
+        return null;
```
mvrueden: Can we add a comment, that the friendly name of business services do not have a friendly name.
Otherwise one may think it was forgotten to be implemented?

mvrueden:
In the UI we enforce a length of the friendly name of >= 1 && <= MAX_LENGTH.
But we do not do that on the service/rest layer.
Let's do that too.

features/bsm/service/impl/src/main/java/org/opennms/netmgt/bsm/service/internal/edge/ChildEdgeImpl.java
pbrane: > On Feb 26, 2016, at 4:59 AM, mvrueden <notifications@github.com> wrote:
> 
> Can we add a comment, that the friendly name of business services do not have a friendly name. Otherwise one may think it was forgotten to be implemented

Can we automatically generate/suggest one based on the reduction key?

:David

David Hustace
The OpenNMS Group, Inc.

mvrueden: Yes, we can do that for ReductionKeyEdges.

christianpape:
I don't think so. In the UI I have setNullSettingAllowed(true), setNullRepresentation("") and setMaxLength(MAXLENGTH). So I think this results in something like >=0 && <=MAXLENGTH since
setRequired() is not set to true.

pbrane:
> On Feb 26, 2016, at 10:44 AM, Christian Pape <notifications@github.com> wrote:
> 
> I don't think so. In the UI I have setNullSettingAllowed(true), setNullRepresentation("") and setMaxLength(MAXLENGTH). So I think this results in something like >=0 && <=MAXLENGTH since
> setRequired() is not set to true.

And that means we can’t programmatically suggest something >= 0 and <=MAXLENGTH

:David


David Hustace
The OpenNMS Group, Inc.

christianpape:
O I'm sorry - my response was not related to friendly name suggestion thing. It was a response to Mvr's comment. Of course we can programmatically suggest a friendly name. We just need some kind of rules for the generation of these suggestions.

pbrane:
> On Feb 26, 2016, at 10:50 AM, Christian Pape <notifications@github.com> wrote:
> 
> We just need some kind of rules for the generation of these suggestions.

Sure.  Maybe we can look at it during our hangout in a few minutes.

features/bsm/persistence/api/pom.xml
```diff
@@ -38,5 +38,14 @@
...
       <artifactId>guava</artifactId>
     </dependency>
+    <dependency>
+      <groupId>javax.validation</groupId>
+      <artifactId>validation-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.hibernate</groupId>
+      <artifactId>hibernate-validator</artifactId>
+      <version>4.3.2.Final</version>
```
mvrueden: The hibernate-validator is used some place else in bsm. Please define a `hibernateValidatorVersion` property in the root pom and re-use the property everywhere else please.

mvrueden:
Can we also add a test for the `BusinessServiceRestServiceIT` to set a friendly name. Also with a friendly name length > 30. 
The same would be nice for the service layer, to ensure the hibernate-validator actually works.

PR #617, opened 2016-02-25
Fix for Bug NMS-8165 - The ACL feature is broken.

agalue:
All the tests from OnmsDaoIT are passing with this change, Bamboo doesn't complain, and the ACL feature is now fully working, after removing the {alias} reference from the @Filter defined inside OnmsOutage.

Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS508
Jira: http://issues.opennms.org/browse/NMS-8165


j-white:
Can we add an integration test to catch the error, so similar bugs don't show up in the future?
If you feel it's out of scope we can create another JIRA issue to track it.

agalue:
I love the idea of having a JUnit test to verify the ACLs, but I can't find a way to do it. Any pointers ?

j-white:
Take a look at [AuthorizationIT.java](https://github.com/OpenNMS/opennms/blob/develop/opennms-dao/src/test/java/org/opennms/netmgt/dao/AuthorizationIT.java). We should be able to extend that for outages.

PR #624, opened 2016-02-29
NMS-8079: The OnmsDaoContainer does not update its cache correctly, leading to a NumberFormatException

mvrueden:
* JIRA: http://issues.opennms.org/browse/NMS-8079
* BAMBOO: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS519


soleger:
Using the DISTINCT_ROOT_ENTITY transformer has some unusual side effects that can cause problems if you are not ready for them. For instance, if you perform a query with a limit of 10, you may only receive 5 results once the transformer deduplicates the results. I do not want to add the DISTINCT_ROOT_ENTITY transformer to AbstractDaoHibernate because of these side effects.

There are other techniques that you can use to ensure that only a single row is returned for a query. The most useful is to specify a join criteria on any aliases to OneToMany related tables so that only one row is returned for each root entity.

soleger:
There is an example of using a join criteria on an alias inside:
org.opennms.web.rest.v2.NotificationRestService

mvrueden:
I reverted my changes and applied another patch.

PR #628, opened 2016-02-29
NMS-8160: Remove Quick-Add node to avoid confusions and avoid breaking the ReST API

agalue:
Considering the side effects on the Requisitions ReST API, I decided to remove everything related with the Quick-Add node view.

Jira: http://issues.opennms.org/browse/NMS-8160
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS512

**IMPORTANT**: The solution for [NMS-8163](http://issues.opennms.org/browse/NMS-8163) will have the AngularJS version of the Quick-Add Node feature, but this has to be merged first.


agalue:
I've added a comment on the Jira issue. Adding a modal window to the new UI for "Quick-Add Node" is so easy that I could have it in almost no time.

Now, properly implement a stand alone UI for this requires changes on the Requisitions ReST API, to save resources, besides the effort of re-implementing things that are already on the new UI, which was not designed to be shared across several JS apps.

indigo423:
Ulf can you please recheck the contributors OCA?

agalue:
As part of the solution for NMS-8163, I already have an AngularJS version of the Quick-Add Node feature fully working and tested, implemented using the current ReST API.

PR #635, opened 2016-03-07
NMS-8205: Generate events for alarm lifecyle changes.

j-white:
JIRA: http://issues.opennms.org/browse/NMS-8205
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS526


mvrueden:
In http://issues.opennms.org/browse/NMS-8205 it is stated, that `Alarmd` and `Vacuumd` need changes to make BSM work properly. `Vacuumd` was changed in `features/bsm` whereas `Alarmd` is changed in `develop`. Shouldn't we change both daemons in `features/bsm` instead?

j-white:
This is a reworked version of the changes that were made in `features/bsm`. I opted to merge these changes into develop separately, since their are relatively independent.

PR #636, opened 2016-03-07
NMS-8211: Add support for retrieving nodes with a filter query.

j-white:
JIRA: http://issues.opennms.org/browse/NMS-8211
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS527

### DIFF
--- a/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/NodeRestService.java
+++ b/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/NodeRestService.java
@@ -29,6 +29,7 @@
 package org.opennms.web.rest.v1;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/NodeRestService.java
```diff
@@ -123,6 +129,22 @@ public OnmsNodeList getNodes(@Context final UriInfo uriInfo) {
...
+                return coll;
+            }
+
+            // Apply the criteria, without the filter rule
+            params.remove("filterRule");
+            final CriteriaBuilder filterRuleCriteriaBuilder = getCriteriaBuilder(params);
+            filterRuleCriteriaBuilder.in("id", filteredNodeIds);
+            filterRuleCriteriaBuilder.distinct();
+
+            crit = filterRuleCriteriaBuilder.toCriteria();
```
mvrueden: To me this looks almost like the first `nodeId` implementation.
Can we either use the same criteria as above, or at least add a comment why we filter differently (and why we need it)?

PR #638, opened 2016-03-09
HZN-491, HZN-494, HZN-602: Final Discovery refactors for Minion

soleger:
These are the last blueprint changes necessary to modularize Discovery enough to run on Minion. It includes the blueprint, features.xml additions, and ITs for the blueprints.


core/icmp-jna/src/main/java/org/opennms/jicmp/standalone/PingReply.java
```diff
@@ -33,6 +33,9 @@
 /**
  * PingReply
  *
+ * @deprecated This interface has a lot of overlap with {@link EchoPacket}...
+ * one of them needs to be moved into an API project.
+ * 
  * @author brozow
  */
 interface PingReply {
```
mvrueden: Please also add `@Deprecated`. The Annotation is interpreted by the IDE, whereas the javadoc bit is for documentation purposes.

features/discovery/src/test/java/org/opennms/netmgt/discovery/DiscovererBlueprintIT.java
```diff
@@ -0,0 +1,162 @@
...
+    /**
+     * Use Aries Blueprint synchronous mode to avoid a blueprint deadlock bug.
+     * 
+     * @see https://issues.apache.org/jira/browse/ARIES-1051
+     * @see https://access.redhat.com/site/solutions/640943
+     */
+    @Override
+    public void doPreSetup() throws Exception
+    {
+        System.setProperty( "org.apache.aries.blueprint.synchronous", Boolean.TRUE.toString() );
```
mvrueden: Is each test invoked with another JVM fork? If not, we should reset the properties to the default values after the test is done.

soleger: They are normally forked.

core/icmp-jna/src/main/java/org/opennms/jicmp/standalone/PingReply.java
soleger: I'm going to defer this because we need to go through all of our code and add the @Deprecated annotation. It's not regularly used, at least in older code.

PR #639, opened 2016-03-09
BSM-167: Fixed BundleException

christianpape:
Jira: http://issues.opennms.org/browse/BSM-167
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS533


mvrueden:
Can we add a test to catch this?

PR #640, opened 2016-03-09
NMS-8194: Normalize the HTTP return codes of the ReST API according to best practices.

agalue:
This contains all the changes necessaries to normalize the return codes returned by each request against the ReST API, according to the best practices of designing and implementing ReST APIs.

This work is required for any third-party integration that relies on the ReST API (like the new Requisitions UI).

Please review this carefully, and let me know if everything is consistent and I'm not missing something important. The rules used are documented on Jira (and also on the official docs, as part of this pull request).

* JIRA: http://issues.opennms.org/browse/NMS-8194
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS521


j-white:
Looks great. Can we also update the REST endpoint for the Measurements API with the same behavior? (See features/measurements/rest/src/main/java/org/opennms/web/rest/v1/MeasurementsRestService.java)

agalue:
> On Mar 10, 2016, at 7:07 AM, Jesse White <notifications@github.com> wrote:
> Looks great. Can we also update the REST endpoint for the Measurements API with the same behavior? (See features/measurements/rest/src/main/java/org/opennms/web/rest/v1/MeasurementsRestService.java)
> 

Thanks a lot Jesse!

I checked it and it looks ok to me at first sight, but I'll look into the code more deeper today to be sure about the BAD_REQUEST on some of the exceptions.

agalue:
I just realized that the NCS feature has a ReST end point that requires a lot of work. I'm going to fix it today, besides double-checking the measurements ReST end-point.

RangerRick:
Did a cursory look-over of these patches and it looks really nice to me.  MAN it's gonna be nice having non-sucky ReST stuff!

agalue:
> On Mar 14, 2016, at 3:18 PM, Benjamin Reed <notifications@github.com> wrote:
> 
> Did a cursory look-over of these patches and it looks really nice to me.

Thanks!

> MAN it's gonna be nice having non-sucky ReST stuff!


Indeed ;)

j-white:
Looks like this is good to merge. There's a merge conflict we need to resolve first though.

j-white:
It took a few retries, but the smoke tests finally passed.

PR #641, opened 2016-03-10
NMS-8215: Reload chart configuration instead of Notifd configuration

indigo423:
Reload Chart configuration instead of Notification configuration.

* JIRA: http://issues.opennms.org/browse/NMS-8215
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS537

### DIFF
--- a/opennms-config/src/main/java/org/opennms/netmgt/config/ChartConfigFactory.java
+++ b/opennms-config/src/main/java/org/opennms/netmgt/config/ChartConfigFactory.java
@@ -115,7 +115,7 @@ public class ChartConfigFactory extends ChartConfigManager {
     @Override
     public void update() throws IOException, MarshalException, ValidationException {
         if (m_lastModified != m_chartConfigFile.lastModified()) {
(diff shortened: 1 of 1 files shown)


indigo423:
@RangerRick The issue was tagged to 18, do you changed already the auto merge behavior? Is this the correct target branch then?

RangerRick:
@indigo423 I don't understand the question.  There is no change to auto-merge because develop was and is already 18.  foundation-2016 auto-merges to develop and that's it.

If you're just asking if foundation-2016 is the correct place to merge bugfixes, then yes, definitely.  That gets them into 18 and meridian 2016.

indigo423:
@RangerRick does it go also to 17.1.1? You mentioned something about 17.1.1 is not auto merged anymore?

agalue:
> On Mar 10, 2016, at 5:25 PM, Ronny Trommer <notifications@github.com> wrote:
> 
> does it go also to 17.1.1?

17.1.1 has been released, so no it doesn't ;)

PR #642, opened 2016-03-10
BSM-114: Delete orphan reduce and map functions

fooker:
The persistence layer is (due to an hibernate bug) not able to delete
oprhan map function instances and reduce function instances. To work
around, the service layer ensures they are deleted.

JIRA: http://issues.opennms.org/browse/BSM-114
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS543


mvrueden:
Looks good to me. I added a changeset to the changelog.xml to delete already orphan map/reduce entries. I say, as soon as Bamboo goes green, let's merge this.

PR #648, opened 2016-03-11
BSM-173: Add a "Problems Box" for applications to the OpenNMS start page.

mvrueden:
* JIRA: http://issues.opennms.org/browse/BSM-173
* BAMBOO: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS545


j-white:
The implementations looks good, however I feel like we're duplicating most of the features we develop for Business Services again for Applications.

Since Applicatons are special cases of Business Services, we can probably just go ahead and remove the Applications concept altogether.

pbrane:
I've got another thought on this but we can talk about it.

David

mvrueden:
@j-white I agree. One thing we may keep in mind is, that the Application concept is very simple, whereas the Business Services concept is more complex. (I know "mimimimi").

PR #651, opened 2016-03-15
BSM-175: Fixed element naming for REST objects

fooker:
* JIRA: http://issues.opennms.org/browse/BSM-175
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS549


mvrueden:
We should also modify the schema file.

fooker:
Bamboo is green, so can we merge this?

PR #652, opened 2016-03-16
NMS-8152: Collect and publish anonymous statistics to stats.opennms.org

j-white:
JIRA: http://issues.opennms.org/browse/NMS-8152
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS490


indigo423:
Just FYI: The "Let's encrypt" certificate used at stats.opennms.org is not trusted by browsers.

features/datachoices/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -0,0 +1,91 @@
...
+                http://aries.apache.org/blueprint/xmlns/blueprint-cm/v1.1.0
+                http://aries.apache.org/schemas/blueprint-cm/blueprint-cm-1.1.0.xsd
+">
+
+    <cm:property-placeholder persistent-id="org.opennms.features.datachoices.cfg">
+        <cm:default-properties>
+            <!-- The Oracle JDK does not currently recognize the Let's Encrypt Certificate Chain
+                 in default installations (the 'ISRG Root X1' root certificate in particular)
+                 so we resort to using HTTP. -->
+            <cm:property name="url" value="http://stats.opennms.org/datachoices/"/>
```
indigo423: If you go to this URL it redirects to https:// should there be

features/datachoices/src/main/java/org/opennms/features/datachoices/internal/StateManager.java
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Maintains state in cfg file:
+ *   systemid=000-00-00-0000
+ *   enabled=true
+ *   acknowledged-by=admin
+ *   acknowledged-at=13943454545
+ *
+ * Same file also holds:
+ *   url=https://stats.opennms.org:3542
```
indigo423: Seems like a left over

j-white: The default value may be different, but the comment is still relevant.

features/datachoices/src/main/resources/OSGI-INF/blueprint/blueprint.xml
j-white: That URL shouldn't redirect. It's intended to be HTTP for the reason outlined in the comment.

features/datachoices/pom.xml
```diff
@@ -0,0 +1,101 @@
...
+            <artifactId>jackson-mapper-asl</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.opennms</groupId>
+            <artifactId>opennms-web-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.freemarker</groupId>
+            <artifactId>freemarker</artifactId>
+            <version>${freemarkerVersion}</version>
```
mvrueden: I am not sure about the scope here. The bundle is wrapped above, but has no scope provided. This may conflict with the maven-bundle-plugin. Can you verify that the current scope is correct?

features/datachoices/src/main/java/org/opennms/features/datachoices/shell/internal/DisplayUsageReportCommand.java
```diff
@@ -0,0 +1,55 @@
...
+
+    private UsageStatisticsReporter m_usageStatisticsReporter;
+
+    @Override
+    protected Object doExecute() throws Exception {
+        long then = System.currentTimeMillis();
+        String reportAsJson = m_usageStatisticsReporter.generateReport().toJson(true);
+        long delta = System.currentTimeMillis() - then;
+
+        System.out.printf("Generated usage statitics reports in %.2f seconds:\n%s\n", delta / 1000f, reportAsJson);
```
mvrueden: Just a thought: How about also using the logging system for this.

features/datachoices/src/main/java/org/opennms/features/datachoices/web/internal/ModalInjector.java
```diff
@@ -0,0 +1,92 @@
...
+        }
+        return null;
+    }
+
+    protected static String generateModalHtml(boolean showOnLoad) throws IOException, TemplateException {
+        Configuration cfg = new Configuration(Configuration.VERSION_2_3_21);
+        ClassTemplateLoader ctl = new ClassTemplateLoader(ModalInjector.class, "/web");
+        cfg.setTemplateLoader(ctl);
+        // Load the template
+        Template template = cfg.getTemplate("modal.ftl.html");
```
mvrueden: Shouldn't we name it `modal.html.ftl` instead?

opennms-doc/guide-admin/src/asciidoc/text/datachoices.adoc
```diff
@@ -0,0 +1,24 @@
...
+:imagesdir: ../images
+
+The `Data Choices` module collects and publishes anonymous usage statistics to https://stats.opennms.org.
+
+When a user with the `Admin` role logs into the system for the first time, they will be prompted as to whether or not they want to opt-in to publish these statistics.
+Statistics will only be published once an `Administrator` has opted-in.
+
+Usage statistics can later be disabled by accessing the 'Data Choices' link in the 'Admin' menu.
+
+When enabled, the following anonymous statistics will be collected and publish on system starutp and every 24 hours after:
```
mvrueden: `startup` vs `starutp`

opennms-web-api/src/main/java/org/opennms/web/api/Injector.java
```diff
@@ -0,0 +1,48 @@
...
+package org.opennms.web.api;
+
+import javax.servlet.http.HttpServletRequest;
+
+/**
+ * Allows services to inject HTML content at arbitrary URLs.
+ *
+ * @author jwhite
+ */
+public interface Injector {
```
mvrueden: Can we name it "HtmlInjector" or something like that. I immediatly had `google guice` or `spring` injection in mind when reading `Injector`.

features/datachoices/src/main/java/org/opennms/features/datachoices/internal/UsageStatisticsReporter.java
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    public UsageStatisticsReportDTO generateReport() {
+        final SystemInfoUtils sysInfoUtils = new SystemInfoUtils();
+        final UsageStatisticsReportDTO usageStatisticsReport = new UsageStatisticsReportDTO();
+        // Unique system identifier
+        try {
+            usageStatisticsReport.setSystemId(m_stateManager.getOrGenerateSystemId());
+        } catch (IOException e) {
+            LOG.warn("An error occurred while retrieving the system id.", e);
```
mvrueden: Should we still continue?

j-white:
In case your interested, the current code for the backend is available here:
  https://github.com/j-white/usage-stats-handler/blob/master/app.js

features/datachoices/src/main/java/org/opennms/features/datachoices/web/internal/ModalInjector.java
j-white: The syntax highlighting continue to work if it ends in .html :)

features/datachoices/pom.xml
j-white: What's wrong with the default compile scope?

mvrueden: I thought that if you leave the scope to "compile" alltogether with the bundle plugin it copies over all the files and that would make no sense if you define the maven dependency manually in the features plugin, but it turns out, it doesn't copy the classes from the compile scoped dependency, so nothing is wrong.

mvrueden:
I tested this and it works. So :+1: from me.

j-white:
Looks like I rebased over the wrong branch. I'm going to build a new branch with this change set and merge it manually.

PR #655, opened 2016-03-21
BSM-186: Enhance the BSM architecture (refactoring)

mvrueden:
* JIRA: http://issues.opennms.org/browse/BSM-186
* BAMBOO: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS558


features/bsm/persistence/api/src/main/java/org/opennms/netmgt/bsm/persistence/api/BusinessServiceEdgeEntity.java
```diff
@@ -180,4 +180,9 @@ public boolean equalsDefinition(BusinessServiceEdgeEntity other) {
                 && getMapFunction().equalsDefinition(other.getMapFunction());
         return equals;
     }
+
+    @Override
+    public <T> T accept(EdgeEntityVisitor<T> visitor) {
+        return null; // we MUST overwrite this in all sub-classes properly
```
j-white: Can we throw a run-time exception instead of returning null? If we can't, can we make a comment?

j-white:
Looks good, I think this will make things a lot easier to maintain in the long run.

features/bsm/persistence/api/src/main/java/org/opennms/netmgt/bsm/persistence/api/BusinessServiceEdgeEntity.java
mvrueden: I didn't throw an exception because the class is not abstract (Hibernate issue) and therefore the method may be called. It is very unlikely, as in the end there should only be sub class instances, but at some point you have a (Abstract)BusinessServciceEdgeEntity.
In those cases, even if it may not be ideal, we should not throw an exception.
I change the comment to make that more clear.

PR #657, opened 2016-03-21
NMS-8226: Upgrade SNMP4J to version 2.4

jeffgdotorg:
http://issues.opennms.org/browse/NMS-8226
Upgrade SNMP4J to 2.4.3 and SNMP4J-Agent to 2.4.2. Thanks to Martin Laercher for the suggestion and testing.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1265,6 +1265,8 @@
     <osgiVersion>5.0.0</osgiVersion>
     <osgiEnterpriseVersion>5.0.0</osgiEnterpriseVersion>
     <slf4jVersion>1.7.7</slf4jVersion>
+    <snmp4jVersion>2.4.3</snmp4jVersion>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -1265,6 +1265,8 @@
     <osgiVersion>5.0.0</osgiVersion>
     <osgiEnterpriseVersion>5.0.0</osgiEnterpriseVersion>
     <slf4jVersion>1.7.7</slf4jVersion>
+    <snmp4jVersion>2.4.3</snmp4jVersion>
+    <snmp4jagentVersion>2.4.2</snmp4jagentVersion>
```
mvrueden: Is there a particular reason why we use a different version?

jeffgdotorg: snmp4j and snmp4j-agent are separate artifacts. At the moment, the versions above are the latest for the corresponding projects.

PR #659, opened 2016-03-23
NMS-8163: Requisitions UI Enhancements

agalue:
This one contains lots of enhancements to the Requisitions UI, and a few enhancements to the Requisitions ReST API.

Here are the highlights:

* Upgrade AngularJS to 1.5.2 (latest version)
* Upgrade Angular UI Bootstrap to 1.2.5 (latest version)
* Using JS library from http://ip-address.js.org to validate IP address (the current RegEx is wrong for IPv6).
* Adding support for reset default foreign source
* Adding support for cloning foreign source definition
* Adding support for reseting the foreign source definition of a given requisition
* Implementing Quick-Add node as part of the current UI, as a stand alone JSP (to replace the old and broken implementation)
* Improving all the icons
* Adding deployed requisition ReST end point to accurately update the UI based on the current state of the nodes from the database, and adding support for this on the Requisitions UI.
* Refactoring the whole logic of building the requisition and nodes state using the deployed stats, instead of the deployed requisitions.
* At any time, you can refresh the deployed stats for all the requisitions, or just a single one, without the need of reloading the whole UI (which is a way to track the progress of the import phase, when adding a new requisition).
* The node page has now two different layouts, and a button to change it ;)
* Enhanced validations for policies and detectors parameters, as well as categories and asset fields.
* Lots of improvements on the Karma/Jasmine tests. These are not executed through the build process, and require to manually run "grunt test" from the UI directory.
* The CasperJS tests has also been improved to test more key features of the UI, but unfortunately not all of them.
* All buttons are now actually buttons, and not submit buttons (which is the default, if you don't specify the type on a button tag). Thanks to this, hitting enter, won't do any unwanted server side requests.

I will appreciate a lot if you can checkout a branch (jira/NMS-8163), give it a try, and give me your impressions. I manually tested all the features with a large set of requisitions (700+ with more than 11000+ nodes in total) using org.opennms.provisiond.repositoryImplementation=fastFile in opennms.properties to accelerate the responsiveness of the ReST API for insanely big requisitions.

* JIRA: http://issues.opennms.org/browse/NMS-8163
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS499/latest


indigo423:
Just a hint, if you add the issue number in front of the pull request, helps a lot to figure out which issue was fixed in a pull request. You can simply type the issue number in the pull request search in GitHub and it is easy to figure out when an where a pull request was merged and in which branch. We don't have a JIRA <-> GitHub integration so this sort of things make live easier especially if you want double check release notes or other things.

j-white:
Looks pretty slick overall.

Here's what I found while playing around with the changes:
1) Quick-add node fails for users with only the "provision" role

To reproduce:
Create a user in the "provision" role
Access the quick-add node page
Attempt to add a node to an existing requisition
Add fails with  "Cannot update SNMP credentials"

It looks like we need to edit the Spring security rules to allow users in the provision role to access /opennms/rest/snmpConfig

2) The Quick-add UI disappears when you click on the "Quick-Add Node" link in the admin menu if you're already on the page

When clicking on the link a second time, the trailing slash is missing from the URL, the UI is restored when you add it back.

3) Synchronizing requisitions that have special characters in the name (including spaces) fail

I realize that this is not a problem with the requisition UI itself, but perhaps we should add additional validation to a requisition's name when creating a *new* one.

agalue:
@j-white thanks a lot! You're right, I probably forgot to update Spring Security. Quick-Add Node should work with ROLE_PROVISION. I'll take a look at this, and the synchronizing issues.

agalue:
@j-white, I fixed the first 2 issues.

The service used by the QuickAddNodeController is the same service used by the Requisitions UI. That means, the synchronization process is exactly the same. For now, please try using requisitions with just alphanumeric characters for the foreign source and the foreign IDs, until we figure out how to properly validate the fields.

I think I should modify the Requisitions Service to URL encode the name of the foreign source and the foreign ID, to avoid URL issues when using special characters.

Also, in case you haven't figured yet, we should NEVER create a requisition called "default" or "deployed", or use the character "/" as part of them for obvious reasons.

Adding an angular validation with custom code is super easy, I'm open for suggestions about how we should validate the requisitions names, and the foreign IDs. Currently, there are no validations for the foreign source name, and the validation for the foreign ID is just to enforce the fact that it should be unique per requisition.

agalue:
It seems like the problem with the synchronization is the ReST API, not the UI or the Quick-Add node feature, when the foreign source name contain spaces, I was able to reproduce the problem, and found that there is no need to URL encode the names (that's done automatically by AngularJS).

agalue:
This is a server side problem ;)
I'm going to look at it next week.
```
Caused by: java.lang.NullPointerException
    at org.opennms.web.svclayer.support.DefaultRequisitionAccessService$RequisitionAccessor.createSnapshot(DefaultRequisitionAccessService.java:464) ~[opennms-web-api-17.0.1-SNAPSHOT.jar:?]
    at org.opennms.web.svclayer.support.DefaultRequisitionAccessService$25.call(DefaultRequisitionAccessService.java:805) ~[opennms-web-api-17.0.1-SNAPSHOT.jar:?]
    at org.opennms.web.svclayer.support.DefaultRequisitionAccessService$25.call(DefaultRequisitionAccessService.java:803) ~[opennms-web-api-17.0.1-SNAPSHOT.jar:?]
    at java.util.concurrent.FutureTask.run(FutureTask.java:266) ~[?:1.8.0_74]
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) ~[?:1.8.0_74]
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617) ~[?:1.8.0_74]
```

agalue:
I think the following is obvious, but because of this server side error, Horizon 17 won't work for all those users who have requisitions with spaces and/or other characters on their names. This is also true if you try to manage the requisitions through the ReST API directly or indirectly through provision.pl

agalue:
@j-white I fixed the NPE generated while creating the snapshots when the requisition has spaces (or any other special characters) on its name on revision f068cdb5fb07488dd7f09d249fd4b95be4002e63. Now, you should be able to use any special character on requisition names that are going to be URL encoded.

PR #662, opened 2016-03-24
BSM-150: Fix broken Node Maps

j-white:
JIRA: http://issues.opennms.org/browse/BSM-150
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS566/latest

### DIFF
--- a/features/topology-map/org.opennms.features.topology.api/src/main/java/org/opennms/features/topology/api/browsers/OnmsVaadinContainer.java
+++ b/features/topology-map/org.opennms.features.topology.api/src/main/java/org/opennms/features/topology/api/browsers/OnmsVaadinContainer.java
@@ -36,6 +36,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.TreeMap;
 
 import org.opennms.core.criteria.Criteria;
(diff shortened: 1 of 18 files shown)


features/topology-map/org.opennms.features.topology.api/src/main/java/org/opennms/features/topology/api/browsers/OnmsVaadinContainer.java
```diff
@@ -173,15 +177,22 @@ public K getItemId(int rowIndex) {
...
-            if (beans == null) return;
-            int rowNumber = page.getStart();
-            for (T eachBean : beans) {
-                addItem(rowNumber, getId(eachBean), eachBean);
-                doItemAddedCallBack(rowNumber, getId(eachBean), eachBean);
-                rowNumber++;
-            }
+            // Wrap the reload in the single transaction context to ensure
+            // that all of the callbacks are executed within a session
+            m_transactionTemplate.execute(new TransactionCallbackWithoutResult() {
```
mvrueden: I thought about this myself.
What about moving this to the `OnmsDaoContainerDatasource` and remove the `doItemAddedCallback` method alltogether, as I think that logic should live in `OnmsDaoContainerDatasource` too.

j-white: That could work. As you mentioned, we'll need to remove the doItemAddedCallBack() function though.

PR #663, opened 2016-03-24
BSM-177: Alarm Label is too long

j-white:
JIRA: http://issues.opennms.org/browse/BSM-177
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS567/latest

### DIFF
--- a/features/topology-map/plugins/org.opennms.features.topology.plugins.topo.bsm/src/main/java/org/opennms/features/topology/plugins/topo/bsm/ReductionKeyVertex.java
+++ b/features/topology-map/plugins/org.opennms.features.topology.plugins.topo.bsm/src/main/java/org/opennms/features/topology/plugins/topo/bsm/ReductionKeyVertex.java
@@ -29,6 +29,8 @@
 package org.opennms.features.topology.plugins.topo.bsm;
 
 import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


features/topology-map/plugins/org.opennms.features.topology.plugins.topo.bsm/src/main/java/org/opennms/features/topology/plugins/topo/bsm/ReductionKeyVertex.java
```diff
@@ -44,12 +49,23 @@ public ReductionKeyVertex(GraphVertex graphVertex) {
...
 
+    protected static String getLabelFromReductionKey(String reductionKey) {
+        String label;
+        Matcher m = REDUCTION_KEY_LABEL_PATTERN.matcher(reductionKey);
+        if (m.matches()) {
+            label = String.format("%s:%s", m.group(1), m.group(2));
+        } else {
+            label = reductionKey;
+        }
+        return label.substring(0, Math.min(label.length(), MAX_LABEL_LENGTH));
```
mvrueden: Can we substring the label with css instead?

j-white: We can limit the size of the text box with CSS, which isn't ideal. I'm not aware of any way of doing a substring per say.

PR #666, opened 2016-03-28
NMS-7880: Karaf integration tests

soleger:
This PR resolves the dependency issues with the KarafTestCase tests. The tests that load OpenNMS and Minion features have been moved into the opennms-full-assembly project.


container/features/src/main/resources/features-minion.xml
```diff
@@ -8,10 +8,16 @@
     <!-- OpenNMS Features -->
     <repository>mvn:org.opennms.karaf/opennms/${project.version}/xml/features</repository>
 
+    <!-- TODO: Maybe this is already defined somewhere else -->
+    <feature name="minion-core-api" description="OpenNMS :: Minion :: Core :: API" version="${project.version}">
```
j-white: Can we move this to the features files in features/minion/core/features?

container/features/src/main/resources/features.xml
```diff
@@ -392,6 +411,11 @@
       <bundle>mvn:org.opennms.core/org.opennms.core.web/${project.version}</bundle>
     </feature>
 
+    <!-- TODO: Remove dependencies on this Minion API bundle inside this features.xml -->
+    <feature name="minion-core-api" description="OpenNMS :: Minion :: Core :: API" version="${project.version}">
```
j-white: Can we move this to the features files in features/minion/core/features?

opennms-full-assembly/src/test/java/org/opennms/assemblies/karaf/FeatureInstallKarafIT.java
```diff
@@ -0,0 +1,150 @@
...
+        installFeature("opennms-rrd-jrobin");
+        installFeature("opennms-snmp");
+        installFeature("opennms-syslogd");
+        //installFeature("opennms-syslogd-handler-default");
+        installFeature("opennms-syslogd-handler-kafka");
+        installFeature("opennms-syslogd-handler-minion");
+        // Syslog listeners can only be installed one at a time
+        //installFeature("opennms-syslogd-listener-camel-netty");
+        //installFeature("opennms-syslogd-listener-javanet");
+        //installFeature("opennms-syslogd-listener-nio");
```
j-white: Can we try installing these separately?

opennms-full-assembly/src/test/java/org/opennms/assemblies/karaf/FeatureInstallKarafIT.java
```diff
@@ -0,0 +1,150 @@
...
+@RunWith(PaxExam.class)
+@ExamReactorStrategy(PerMethod.class)
+public class FeatureInstallKarafIT extends KarafTestCase {
+
+    /**
+     * This test attempts to install all features from the OpenNMS
+     * features.xml.
+     */
+    @Test
+    public void testInstallAllOpenNMSFeatures() {
```
j-white: Should we trying installing all of these together and individually?

soleger: The OnmsFeatureKarafIT test does install each feature individually so we're doing it both ways. 

I'm not sure if this monolithic test serves much of a purpose other than failing if there are severe interdependency issues (like multiple versions of Spring or other libraries being loaded at once). But I guess that's a good reason to try and load all of the features in the same container like we're doing here.

opennms-full-assembly/src/test/java/org/opennms/assemblies/karaf/FeatureInstallKarafIT.java
soleger: I didn't add the opennms-syslogd-listener-* features to OnmsFeatureKarafIT or MinionFeatureKarafIT yet but they should be added.

soleger: I just tried to add these and there's an issue: currently the Syslogd features rely on minion-core-api  because of the reused ActiveMQ component. We need to write a new test inside the project that constructs the Minion features repo for Syslogd that performs these tests. I'd like to defer that until after this PR is merged.

container/features/src/main/resources/features-minion.xml
soleger: Yes, this feature should be there. I'd like to do that after this PR is merged because we need to write Minion-only ITs inside the features/minion repo project anyway.

PR #677, opened 2016-03-31
BSM-148: Add information panel to Topology UI

fooker:
JIRA: http://issues.opennms.org/browse/BSM-148
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS555

This provides the infrastructure and APIs for selection responsive info
panels in the topology UI. Registered panel items can provied UI
components in relation to the selected vertex or edge.

A detailed panel for all BSM components is provided showing the element
details and the status including an inside view of the status
calculations.

In addition, a generic panel for technical details is provieded for all
selections, regardless of the specific type.


j-white:
Looks great. The only comment I have is that elements looks off center when the panel is collapsed. Perhaps we should redraw the layout when toggling the panel.

j-white:
I'm going to start some work which is based on this branch, so please try and avoid any force pushes :)

fooker:
I've created Issue BSM-200 for the offset problem to get this merged faster.

mvrueden:
Re-centering the graph is not an easy task. 
A `redoLayout` will not have any effect as it is applied BEFORE the new width/height information is available. Issue issues.opennms.org/browse/BSM-200 was created to address this out of this issue's scope.

PR #681, opened 2016-04-01
NMS-8266: SNMP collection systemDefs for Juniper SRX210he2, SRX100h

jeffgdotorg:
JIRA: http://issues.opennms.org/browse/NMS-8266
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS583


indigo423:
Looks good to me, we should let bamboo crunch the Unit tests for data collection configs. It "should" pass without issues ;)

PR #683, opened 2016-04-01
BSM-137: Simulate state calculation in the Topology UI

j-white:
JIRA: http://issues.opennms.org/browse/BSM-137
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS554-13


indigo423:
#677 is merged :)

mvrueden:
I talked to Jesse and this is what we'd like to add in addition:
 * Visiual indicator that you are actually in simulation mode (at the moment one does not know)
 * Disable the Simulation options in the menu when BSM Topology is NOT selected

PR #684, opened 2016-04-04
Updated PostgreSQL JDBC driver to version 9.4.1208.

soleger:
Version bump to 9.4.1208.


mvrueden:
Can we create an issue for this for transparency?

soleger:
This crashed and burned in Bamboo so I'm going to close this PR until we time to resolve the issues. It's unusual that a PostgreSQL driver bump broke but it could be an incompatibility in our version of Liquibase.

PR #689, opened 2016-04-05
Removed old Linkd tables from create.sql

soleger:
There were still stale Linkd tables in the create.sql file. I removed them and tests still work fine.

* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS578


indigo423:
Do we have a issue for that?

mvrueden:
This technically belongs to the "remove-linkd" issue @fooker and @christianpape worked on a while ago. This is a clean up, which originally was forgotten. I don't think it is necessary to create an issue for this. I also do not think we should re-open the old one for this either. 
As Ronny likes to say "Just my 0.02 _enter your currency of choice_"

PR #690, opened 2016-04-05
NMS-8271: Fix unexpected CorrelatorTest test failures

jeffgdotorg:
http://issues.opennms.org/browse/NMS-8271


agalue:
I tested the solution this time. I can compile it and all the tests (ITs and non-ITs) are passing.

PR #692, opened 2016-04-06
NMS-8198: Change Provisioning UI unsaved state indication

indigo423:
Remove bright table color and changed font color from white to black.

* JIRA: http://issues.opennms.org/browse/NMS-8198
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS597


j-white:
I'll save this one for @agalue

agalue:
Looks ok to me. I mean, the change is not designed for the Requisitions UI, it is designed for anything on the OpenNMS UI that uses Bootstrap and the "danger" class for coloring things.

I'm not aware of other OpenNMS parts that are using that class, so I would say .... "Ship it!"

PR #694, opened 2016-04-07
BSM-201: Show ... when reduction key label is cut

mvrueden:
* JIRA: http://issues.opennms.org/browse/BSM-201
* BAMBOO: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS599


features/topology-map/plugins/org.opennms.features.topology.plugins.topo.bsm/src/main/java/org/opennms/features/topology/plugins/topo/bsm/ReductionKeyVertex.java
```diff
@@ -62,7 +62,10 @@ protected static String getLabelFromReductionKey(String reductionKey) {
         } else {
             label = reductionKey;
         }
-        return label.substring(0, Math.min(label.length(), MAX_LABEL_LENGTH));
+        if (label.length() > MAX_LABEL_LENGTH) {
+            return label.substring(0, MAX_LABEL_LENGTH - "...".length()) + "...";
```
fooker: Using a constant for the ellipsis or using StringUtils.abbreviate if available?

PR #698, opened 2016-04-08
HZN-701: Use FontAwesome Icons from Vaadin instead

mvrueden:
* JIRA: http://issues.opennms.org/browse/HZN-701
 * BAMBOO: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS603


mvrueden:
This should be merged after https://github.com/OpenNMS/opennms/pull/696

mvrueden:
It seems that the changes in this branch cause a weird issue:
If you move the topology over to the right the SVG background overlays the toolbar.
It is only temporarily as the toolbar immediately redraws, but it feels wrong.
@indigo423 is going to verify this.
Until then, let's NOT merge this.

mvrueden:
The toolbar is going to change with "Project Atlas". Therefore I re-create this pull request against `features/atlas` (https://github.com/OpenNMS/opennms/pull/784) instead. The mentioned problems here should be no issue anymore.

PR #700, opened 2016-04-11
NMS-8221: Refresh now dirties the graph container

fooker:
* JIRA: http://issues.opennms.org/browse/NMS-8211
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS608

### DIFF
--- a/features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/operations/RedoLayoutOperation.java
+++ b/features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/operations/RedoLayoutOperation.java
@@ -30,6 +30,7 @@ package org.opennms.features.topology.app.internal.operations;
 
 import java.util.List;
 
+import org.opennms.features.topology.api.GraphContainer;
 import org.opennms.features.topology.api.Operation;
 import org.opennms.features.topology.api.OperationContext;
 import org.opennms.features.topology.api.OperationContext.DisplayLocation;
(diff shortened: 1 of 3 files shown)


features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/operations/RedoLayoutOperation.java
```diff
@@ -41,8 +41,8 @@
 	@Override
     public void execute(List<VertexRef> targets, OperationContext operationContext) {
         if (operationContext != null && operationContext.getGraphContainer() != null) {
-            //operationContext.getGraphContainer().getBaseTopology().refresh();
-            //operationContext.getGraphContainer().setDirty(true);
+            operationContext.getGraphContainer().getBaseTopology().refresh();
```
mvrueden: Can we then also change the TopologyUI auto refresher logic to use the "RedoLayoutOperation" instead?
See https://github.com/OpenNMS/opennms/blob/develop/features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/TopologyUI.java#L166

mvrueden:
Can you also please change the commit message as it is only half true. 
The real fix here is that we "reload" the topology provider which was not reloaded previously.
The setDirty(true) forces the graph to reload (from the Topology Provider)

fooker:
I've changed the code to use the same function on both actions. Fixing the commit message can be done during squash merging.

PR #701, opened 2016-04-11
NMS-8043: Ping from topology view does not work (results in 404 Page not found)

mvrueden:
* JIRA: http://issues.opennms.org/browse/NMS-8043
* BAMBOO: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS610

### DIFF
--- a/opennms-webapp/src/main/java/org/opennms/web/map/ExecCommandController.java
+++ b/opennms-webapp/src/main/java/org/opennms/web/map/ExecCommandController.java
@@ -1,0 +1,238 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2007-2014 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2014 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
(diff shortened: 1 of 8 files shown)


opennms-webapp/src/main/java/org/opennms/web/map/ExecCommandController.java
```diff
@@ -0,0 +1,238 @@
...
+/**
+ * <p>ExecCommandController class.</p>
+ *
+ * @author mmigliore
+ * @author antonio
+ *
+ * this class provides to create pages for executing ping and traceroute commands
+ * @version $Id: $
+ * @since 1.8.1
+ * @deprecated DO NOT use anymore. See http://issues.opennms.org/browse/NMS-8245 why.
```
j-white: Do you really want to bring this class back?

j-white:
I'm not a big fan of bringing the ExecCommandController back, passing user input to Runtime.exec() is never a good idea.

The Pinger service is now available as an OSGi service. I'd opt for using this instead:
https://github.com/OpenNMS/opennms/blob/develop/opennms-icmp/opennms-icmp-jni/src/main/resources/OSGI-INF/blueprint/blueprint.xml#L13

mvrueden:
I agree. 
We should revisit this.
The JniPinger is the better soluting.
However we have to deal with the "long running task" problem in Vaadin or the Request-Handler. 
However this is not really doable in the time frame for release 18.0.0. 
We should revisit this in Horizon 19.0.0.
Closing Pull Request and deleting branch.

PR #702, opened 2016-04-11
Fix NMS-6549: Eventd does not honor reloadDaemonConfig event

jeffgdotorg:
* JIRA: http://issues.opennms.org/browse/NMS-6549
* BAMBOO: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS611

### DIFF
--- a/features/events/daemon/src/main/java/org/opennms/netmgt/eventd/BroadcastEventProcessor.java
+++ b/features/events/daemon/src/main/java/org/opennms/netmgt/eventd/BroadcastEventProcessor.java
@@ -34,6 +34,7 @@ import org.opennms.netmgt.events.api.EventIpcManager;
 import org.opennms.netmgt.events.api.EventListener;
 import org.opennms.netmgt.model.events.EventBuilder;
 import org.opennms.netmgt.xml.event.Event;
+import org.opennms.netmgt.xml.event.Parm;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.util.Assert;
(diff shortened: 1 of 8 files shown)


features/events/daemon/src/main/java/org/opennms/netmgt/eventd/BroadcastEventProcessor.java
```diff
@@ -147,7 +148,8 @@ private boolean isReloadConfigEvent(Event event) {
         boolean isTarget = false;
         
         if (EventConstants.RELOAD_DAEMON_CONFIG_UEI.equals(event.getUei())) {
-            if ("Eventd".equalsIgnoreCase(AbstractEventUtil.getInstance().getValueOfParm(EventConstants.PARM_DAEMON_NAME, event))) {
+            Parm target = event.getParm(EventConstants.PARM_DAEMON_NAME);
+            if (target != null && "Eventd".equalsIgnoreCase(target.getValue().getContent())) {
```
mvrueden: Dumb question: But can `target.getValue()` be null?
If so, we should check that, too.

mvrueden:
Can we write a unit test for this?

features/events/daemon/src/main/java/org/opennms/netmgt/eventd/BroadcastEventProcessor.java
agalue: That is theoretically impossible. In fact, if that can be null, there are hundreds of lines of codes in OpenNMS that would require a NPE check.

agalue:
> On Apr 12, 2016, at 8:31 AM, mvrueden <notifications@github.com> wrote:
> 
> Can we write a unit test for this?


Done!

features/events/daemon/src/main/java/org/opennms/netmgt/eventd/BroadcastEventProcessor.java
pbrane: > On Apr 12, 2016, at 8:47 AM, Alejandro Galue <notifications@github.com> wrote:
> 
> >          boolean isTarget = false;
> >
> >          if (EventConstants.RELOAD_DAEMON_CONFIG_UEI.equals(event.getUei())) {
> > -            if ("Eventd".equalsIgnoreCase(AbstractEventUtil.getInstance().getValueOfParm(EventConstants.PARM_DAEMON_NAME, event))) {
> > +            Parm target = event.getParm(EventConstants.PARM_DAEMON_NAME);
> > +            if (target != null && "Eventd".equalsIgnoreCase(target.getValue().getContent())) {
> 
> That is theoretically impossible. In fact, if that can be null, there are hundreds of lines of codes in OpenNMS that would require a NPE check.

It is *in fact* impossible since a boolean is a primitive.

:D

jeffgdotorg:
Targeted wrong branch again. Burning it down.

PR #710, opened 2016-04-12
NMS-8296: Be able to re-order the policies on a requisition through the UI

agalue:
* JIRA: http://issues.opennms.org/browse/NMS-8296
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS621

I ended up by implementing a very simple but useful way to move elements on angular tables that have pagination and filter components.

This is intended to re-order the detectors or policies on a foreign source definition. The elements will be applied in the exact same order they are defined, so order is important, specially for policies.


agalue:
I tested it on my lab and it is working fine. Hope someone else can test it, to be sure I'm not missing something.

I preferred this instead of drag-and-drop, because a drag-and-drop will be weird on paginated/filtered lists.

PR #711, opened 2016-04-13
NMS-8313: Do not modify the bounding box after listener were notified

mvrueden:
* JIRA: http://issues.opennms.org/browse/NMS-8313
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS622

This patch should also fix the following issues:
 * http://issues.opennms.org/browse/NMS-8299
 * http://issues.opennms.org/browse/NMS-8300
 * http://issues.opennms.org/browse/NMS-8314


j-white:
I tried the patch and it worked. Waiting on Bamboo.

PR #722, opened 2016-04-17
HZN-708: BSM Admin UI Enhancements

j-white:
JIRA: http://issues.opennms.org/browse/HZN-708
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS627

Helps improve the usability of the BSM Admin UI by:
* Removing the ID column
* Using a TreeTable instead of a normal table to help visualize the hierarchy
* Adding a link to open a particular service in the Topology UI
* Using searchable combo boxes instead of list selects for adding edges

### DIFF
--- a/features/bsm/vaadin-adminpage/src/main/java/org/opennms/netmgt/bsm/vaadin/adminpage/BusinessServiceMainLayout.java
+++ b/features/bsm/vaadin-adminpage/src/main/java/org/opennms/netmgt/bsm/vaadin/adminpage/BusinessServiceMainLayout.java
@@ -156,18 +156,18 @@ public class BusinessServiceMainLayout extends VerticalLayout {
                         });
                     } else {
                         new org.opennms.netmgt.vaadin.core.ConfirmationDialog()
-                                .withOkAction(UIHelper.getCurrent(TransactionAwareUI.class).wrapInTransactionProxy(new org.opennms.netmgt.vaadin.core.ConfirmationDialog.Action() {
-                                    @Override
-                                    public void execute(org.opennms.netmgt.vaadin.core.ConfirmationDialog window) {
-                                        m_businessServiceManager.getBusinessServiceById((Long) itemId).delete();
-                                        refreshTable();
-                                    }
-                                }))
-                                .withOkLabel("Delete anyway")
-                                .withCancelLabel("Cancel")
-                                .withCaption("Warning")
-                                .withDescription("This entry is referencing or is referenced by other Business Services! Do you really want to delete this entry?")
-                                .open();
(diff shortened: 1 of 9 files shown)


pbrane:
> On Apr 17, 2016, at 1:09 PM, Jesse White <notifications@github.com> wrote:
> 
> Helps improve the usability of the BSM Admin UI by:
> 
> 	• Removing the ID column
> 	• Using a TreeTable instead of a normal table to help visualize the hierarchy
> 	• Adding a link to open a particular service in the Topology UI
> 	• Using searchable combo boxes instead of list selects for adding edges
> 

Awesomeness.

features/bsm/vaadin-adminpage/src/main/java/org/opennms/netmgt/bsm/vaadin/adminpage/BusinessServiceMainLayout.java
```diff
@@ -59,121 +78,197 @@
...
+
+            @Override
+            public Object generateCell(Table source, Object itemId, Object columnId) {
+                final HorizontalLayout layout = new HorizontalLayout();
+                final BusinessServiceStateMachine stateMachine = m_businessServiceManager.getStateMachine();
+                try {
+                    final BusinessService businessService = m_beanContainer.getItem(itemId).getBean().getBusinessService();
+                    final Status status = stateMachine.getOperationalStatus(businessService);
+                    if (status != null) {
+                        final URIBuilder builder = new URIBuilder("http://host/opennms/topology");
```
fooker: Ouch!

features/bsm/vaadin-adminpage/src/main/java/org/opennms/netmgt/bsm/vaadin/adminpage/BusinessServiceMainLayout.java
```diff
@@ -59,121 +78,197 @@
...
+                        builder.addParameter("focus-vertices", businessService.getId().toString());
+                        builder.addParameter("szl", "1");
+                        builder.addParameter("layout", "Hierarchy Layout");
+                        builder.addParameter("provider", "Business Services");
+                        final URL url = builder.build().toURL();
+                        final Link link = new Link("View in Topology UI", new ExternalResource(
```
fooker: Can we make the links opennms-ish green?

fooker:
I like the idea of using a tree view for navigation, but it feels kinda clumsy for komplex graphs.

Lets imagine the following situation:
I have added two BSs called ```root1``` and ```root2``` without any further configuration. In addition, I have added a BS called ```shared``` and make it a child to both root BSs. Now, I have added another BS called ```child```.
At first, ```child``` is shown at the root level, as it's not in relation to any other BS. If I make it a child of ```shared```, it disappears from the visible part of the table and can only be found deep in the tree. To make it more strange, the element can be found twice - once for each root.

I'm still not sure how to handel the or show it in a better way...

j-white:
Trying to flatten a complex graph into a table view isn't easy, but I do think that the tree view provides many benefits over the naive list of services.

PR #723, opened 2016-04-18
NMS-8334: Implement org.opennms.timeseries.strategy=evaluate to facilitate the sizing process

agalue:
Implement org.opennms.timeseries.strategy=evaluate to facilitate the sizing process

The idea is to use the these strategy when designing the cluster for Cassandra or doing the sizing for the storage backend for RRDtool.

This strategy is going to work like a normal persister, but its purpose is just write data on a log file called EvaluationMetrics.log. The frequency on which the data will be added to the log can be controlled through a property called org.opennms.evaluate.dump_stats_frequency (5 minutes by default)

Also, on a higher frequency controlled by org.opennms.evaluate.dump_cache_frequency (60 minutes by default) the content of the shared cache used to count resources, groups, nodes, interfaces and attributes will be written on a log file called EvaluationMetrics-Cache.log.

Sample Data for EvaluationMetrics.log

```
2016-04-18 13:56:25,480 INFO  [metrics-logger-reporter-1-thread-1] EvaluationMetrics: type=GAUGE, name=evaluate.groups, value=282
2016-04-18 13:56:25,482 INFO  [metrics-logger-reporter-1-thread-1] EvaluationMetrics: type=GAUGE, name=evaluate.interfaces, value=9
2016-04-18 13:56:25,482 INFO  [metrics-logger-reporter-1-thread-1] EvaluationMetrics: type=GAUGE, name=evaluate.nodes, value=9
2016-04-18 13:56:25,482 INFO  [metrics-logger-reporter-1-thread-1] EvaluationMetrics: type=GAUGE, name=evaluate.numeric-attributes, value=1361
2016-04-18 13:56:25,482 INFO  [metrics-logger-reporter-1-thread-1] EvaluationMetrics: type=GAUGE, name=evaluate.resources, value=140
2016-04-18 13:56:25,482 INFO  [metrics-logger-reporter-1-thread-1] EvaluationMetrics: type=GAUGE, name=evaluate.string-attributes, value=289
2016-04-18 13:56:25,483 INFO  [metrics-logger-reporter-1-thread-1] EvaluationMetrics: type=METER, name=evaluate.samples, count=13179, mean_rate=43.928074974234306, m1=45.14303088528703, m5=27.7474824322603, m15=12.442160095821313, rate_unit=events/second
```

The groups, resources and attributes contain data from Pollerd and Collectd. The data for "nodes" is the distinct collection resource parents found (which is associated with the nodes that have a Collectable Services associated in collectd-configuration.xml). The data for "interfaces" is the distinct IP addresses found (which is associated with the IP interfaces that have Monitored Services associated in poller-configuration.xml).

* JIRA: http://issues.opennms.org/browse/NMS-8334
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS624


j-white:
Looks great.

Since we're adding a new log file, should we also add an entry to the default **log4j2.xml**?

agalue:
@j-white This feature (which is not enabled by default), is intended to be used once and for a short period of time. Having this in mind, I think it doesn't worth having a definition for log4j2.xml, or how useful would be having one. Why do you want to add one ?

j-white:
I was just asking since we have entries for all of the other log files we produce.

agalue:
Not true, the SNMP Asset Provider, and the SNMP Inventory Adapter are just two examples of things not covered by log4j2.xml that should be covered, but in this case I think we don't need one.

PR #729, opened 2016-04-20
NMS-5915: Change HTTP client code to retry on any exception

soleger:
Because we are a network management framework, we want to retry operations even if low-level protocol exceptions have been thrown. The solution proposed by Alejandro has been incorporated into the httpclient API since this bug was opened and because Ben wrote the HttpClientWrapper class a few years ago, this retry logic will be applied in most places that we use httpclient (PSM, HttpCollector, etc).

* JIRA: http://issues.opennms.org/browse/NMS-5915
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS645

### DIFF
--- a/core/web/src/main/java/org/opennms/core/web/HttpClientWrapper.java
+++ b/core/web/src/main/java/org/opennms/core/web/HttpClientWrapper.java
@@ -415,7 +415,7 @@ public class HttpClientWrapper implements Closeable {
                 requestConfigBuilder.setConnectTimeout(m_connectionTimeout);
             }
             if (m_retries != null) {
-                httpClientBuilder.setRetryHandler(new HttpRequestRetryOnExceptionHandler(m_retries, true));
+                httpClientBuilder.setRetryHandler(new HttpRequestRetryOnExceptionHandler(m_retries, false));
             }
             if (m_sslContext.size() != 0) {
                 configureSSLContext(httpClientBuilder);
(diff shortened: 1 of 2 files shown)


core/web/src/main/java/org/opennms/core/web/HttpClientWrapper.java
```diff
@@ -388,7 +415,7 @@ public CloseableHttpClient getClient() {
                 requestConfigBuilder.setConnectTimeout(m_connectionTimeout);
             }
             if (m_retries != null) {
-                httpClientBuilder.setRetryHandler(new DefaultHttpRequestRetryHandler(m_retries, false));
+                httpClientBuilder.setRetryHandler(new HttpRequestRetryOnExceptionHandler(m_retries, true));
```
j-white: This client is used in many different contexts, I'm afraid that flipping requestSentRetryEnabled from false to true might have some unintended consequences.

soleger: Here's a list of all of the stuff using HttpClientWrapper:
```
JUnitHttpServerTest
HttpClientWrapper
UsageStatisticsReporter
NominatimGeocoderService
NominatimGeocoder
HypericAckProcessor
HttpNorthbounder
HttpCollector
NCSNorthbounder
HttpNotificationStrategy
HttpPostMonitor
PageSequenceMonitor
WebMonitor
WebClient
RequestTracker
HttpUrlConnection
HttpUrlConnectionIT
```
For a lot of these, requestSentRetryEnabled would be fine since it's just idempotent read operations but I guess it could cause problems with alarm forwarding and notifications in some rare cases. I'll turn it back off.

PR #730, opened 2016-04-21
NMS-8098: Unwrapped exceptions to make nicer PSM failure reasons

soleger:
The PSM was catching an httpclient exception and always returning failure message "I/O Error" even though the root cause exceptions provided more precise failure codes. I added code to unwrap the exceptions so that the failure codes are more useful.

I also added a catch(Throwable) to the main PSM poll() method so that it's more resilient to unexpected exceptions (like config typos, etc).

* JIRA: http://issues.opennms.org/browse/NMS-8098
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS646/latest

### DIFF
--- a/features/poller/api/src/main/java/org/opennms/netmgt/poller/PollStatus.java
+++ b/features/poller/api/src/main/java/org/opennms/netmgt/poller/PollStatus.java
@@ -45,6 +45,8 @@ import javax.persistence.Transient;
 public class PollStatus implements Serializable {
     private static final long serialVersionUID = 3L;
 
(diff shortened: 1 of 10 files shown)


pbrane:
> On Apr 20, 2016, at 10:01 PM, Seth Leger <notifications@github.com> wrote:
> 
> The PSM was catching an httpclient exception and always returning failure message "I/O Error" even though the root cause exceptions provided more precise failure codes. I added code to unwrap the exceptions so that the failure codes are more useful.

Nice!  Thanks.

PR #743, opened 2016-04-26
NMS-8373: Upgrade Backshift and add support for {diffTime}

j-white:
JIRA: http://issues.opennms.org/browse/NMS-8373
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS666


features/measurements/impl/src/main/java/org/opennms/netmgt/measurements/impl/JEXLExpressionEngine.java
```diff
@@ -99,6 +99,9 @@ public void applyExpressions(final QueryRequest request, final FetchResults resu
         final Map<String, double[]> columns = results.getColumns();
         final int numRows = timestamps.length;
 
+        // Calculate the time span
+        jexlValues.put("__diff_time", numRows < 1 ? 0d : timestamps[numRows-1] - timestamps[0]);
```
agalue: During my tests, I figured out that you have to divide the difference between the two timestamps by 1000 in order to get the same numbers VDEF provides.

agalue: For example, ifInOctets is expressed in "bytes per second" (any counter is treated as a rate), if you want "bytes", you have to multiply by the amount of "seconds", not the amount of "milliseconds".

j-white: I divide __diff_time by 1000 here instead: https://github.com/OpenNMS/backshift/blob/v1.1.2/src/Backshift.Utilities.RpnToJexlConverter.js#L145

That way all time units in the Measurement API remain in millis.

agalue: Perfect!

PR #747, opened 2016-04-27
HZN-726: Changed ActiveMQ broker to only listen for vm://localhost

soleger:
This commit changes the OpenNMS-side JMS connections to run over the vm://localhost connector instead of TCP. The TCP port will be disabled by default. I also fixed the Syslog Minion code to use the shared JMS connection from minion-core.

* JIRA: http://issues.opennms.org/browse/HZN-726
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS669

### DIFF
--- a/features/discovery/blueprint-discovery.xml
+++ b/features/discovery/blueprint-discovery.xml
@@ -56,9 +56,9 @@
 		<argument ref="eventForwarder" />
 	</bean>
 
-	<bean id="discovery-activemq" class="org.apache.activemq.camel.component.ActiveMQComponent">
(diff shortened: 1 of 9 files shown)


j-white:
Looks good.

We should also update the paragraph that explains how to enable the TCP connector, since the statements there are no longer valid. See **opennms-doc/guide-install/src/asciidoc/text/minion/introduction.adoc**.

pbrane:
Awesome!

j-white:
Thanks for the doc update!

PR #750, opened 2016-04-27
HZN-544: Add jmx-datacollection-config.d/ for defining custom JMX datacollection definitions

j-white:
Here we split the default jmx-datacollection-config.xml into multiple files, and allow the configuration to be extended by adding new files in jmx-datacollection-config.d/. Additional files do not need to be explicitly included, making it easy to overlay custom configuration without touching any of the distribution files.

JIRA: http://issues.opennms.org/browse/HZN-544
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS674


agalue:
Looks very good!

It seems like everything inside jmx-datacollection.d will be loaded automatically, and there is no need to add reference to those files, right ?

A long time ago, @DerTak added some sample configuration to examples/jvm-datacollection. Maybe those files require some love as they are based on the code I made (also a long time ago) to physically add references to the external files (i.e. import-mbeans).

j-white:
Correct, everything in jmx-datacollection.d is automagically loaded, and reloaded on change.

It looks like the "import-mbeans" statements no longer work though, I guess I'll fix those since you may still want to share mbean definitions between collections.

indigo423:
this is really nice, we have splitted the basic JVM metrics from the OpenNMS ones. They are right now in the examples repository. We can target the JVM config cleanup for 19.

indigo423:
the testJmxDatacollectionConfig zz001testAllConfigs failed :/

mvrueden:
This changes may also impact the JMX Configuration CLI and UI. I am going to verify this on Monday.

indigo423:
@mvrueden does it have effect on the JMX config CLI and UI?

agalue:
> On Apr 27, 2016, at 4:35 PM, Jesse White <notifications@github.com> wrote:
> 
> It looks like the "import-mbeans" statements no longer work though, I guess I'll fix those since you may still want to share mbean definitions between collections.
> 
Great, as that's why import-beans exist, besides that it might be possible that some users are using this feature, to avoid copy/paste the same thing across several JMX collections.

mvrueden:
I just verified. 
The JMX Configuration Generator CLI and UI should still work as expected.
The documentation does not need to be updated as well.
Sorry, it took me a bit longer than anticipated.

PR #751, opened 2016-04-27
NMS-8076: Added links to local, offline docs to webapp

soleger:
These changes will add links to the online docs to the "Help/Support" page. If you install the opennms-docs package, the links will redirect to local copies of the docs so that you don't need internet access to view them.

* JIRA: http://issues.opennms.org/browse/NMS-8076
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS675

### DIFF
--- a/opennms-webapp/src/main/webapp/WEB-INF/jsp/support/index.jsp
+++ b/opennms-webapp/src/main/webapp/WEB-INF/jsp/support/index.jsp
@@ -43,6 +43,26 @@ session="true"
   <jsp:param name="breadcrumb" value="Support" />
 </jsp:include>
 
+<script type="text/javascript">
+  // Shorthand for 'onload'
+  $(function() {
+    // Check to see if the URL provided by the opennms-docs package is present on the system.
+    // If so, display the links to the offline docs symlinked from /usr/share/doc/opennms-${version}.
+    $.ajax({
+      url: 'docs/guide-install/index.html',
+      method: 'HEAD',
+      error: function()
+      {
+        $('#online-documentation').css("display", "inline-block");
+      },
+      success: function()
+      {
+        $('#offline-documentation').css("display", "inline-block");
+      }
+    });
+  });
+</script>
+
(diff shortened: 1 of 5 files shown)


opennms-webapp/src/main/webapp/WEB-INF/jsp/support/index.jsp
```diff
@@ -185,11 +205,37 @@ session="true"
...
+  <div class="panel-heading">
+    <h3 class="panel-title">Documentation</h3>
+  </div>
+  <div class="panel-body">
+    <span id="online-documentation" style="display:none;">
+    <ul class="list-unstyled">
+      <li><a href="http://docs.opennms.org/opennms/releases/latest/guide-install/guide-install.html">Installation Guide</a></li>
+      <li><a href="http://docs.opennms.org/opennms/releases/latest/guide-user/guide-user.html">Users Guide</a></li>
+      <li><a href="http://docs.opennms.org/opennms/releases/latest/guide-admin/guide-admin.html">Administrators Guide</a></li>
+      <li><a href="http://docs.opennms.org/opennms/releases/latest/guide-development/guide-development.html">Developers Guide</a></li>
```
indigo423: The only think I've seen the links to the online documentation goes to "latest" instead of the specific version you have running. It can lead to features and functions which don't exist in the version you've running. We don't have a way to ask for a version number right now. Just a quick idea, we could create just one link to the online docs which goes to http://docs.opennms.org/opennms/releases/ and the user has to pick the version he had installed by himself.

mvrueden: We should definetly link to the version currently installed and not to `latest`

soleger: Because this requires additional code changes, I've opened a new issue to cover this.

http://issues.opennms.org/browse/NMS-8402

PR #765, opened 2016-05-02
HZN-732: Add smoke tests for the Topology UI

j-white:
JIRA: http://issues.opennms.org/browse/HZN-732
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS679


smoke-test/src/test/java/org/opennms/smoketest/BSMAdminIT.java
```diff
@@ -61,59 +62,61 @@
     /**
      * Class to control the inputs of the "Business Service Edit"-Window
      */
-    private class BsmAdminPageEditWindow {
+    public static class BsmAdminPageEditWindow {
```
mvrueden: Can we make the classes not inner classes and move them to the src package instead?

smoke-test/src/test/java/org/opennms/smoketest/TopologyIT.java
```diff
@@ -0,0 +1,383 @@
...
+        @Override
+        public String toString() {
+            return String.format("VisibleVertex[label='%s']", label);
+        }
+    }
+
+    /**
+     * Controls the workflow of the "Topology UI" Page
+     */
+    public static class TopologyUiPage {
```
mvrueden: Can we name this `TopologyUIPage` to be consistend with the `TopologyUI` class?

smoke-test/src/test/java/org/opennms/smoketest/TopologyIT.java
```diff
@@ -0,0 +1,383 @@
...
+                                + " Unable to determine if the item is checked or unchecked.");
+                }
+            } else {
+                throw new RuntimeException("Element has no CSS classes!"
+                        + " Unable to determine if the item is checked or unchecked.");
+            }
+        }
+    }
+
+    public static class TopologyUiSearchResults {
```
mvrueden: Can we rename this to `TopologyUISearchResults` just to be consistend with the `TopologyUI` class?

smoke-test/src/test/java/org/opennms/smoketest/BSMAdminIT.java
j-white: Most of the method they need access to are *protected*, so we need to keep them in the same package. That said, I'd rather keep them grouped together here than breaking them out in *org.opennms.smoketest*.

j-white:
I'm going to retarget this to the features/atlas branch to make sure we address any possible intermittent failures before it's merged back to develop.

PR #768, opened 2016-05-03
Cleanup spaces and tabs in new lines to reduce merge conflicts

indigo423:
Cleanup introduced white spaces on new lines and before line breaks which makes diffs harder during updates.

* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS693


indigo423:
If we should have an issue in JIRA for this kind of things, please let me know and I'll create one.

PR #771, opened 2016-05-03
HZN-732: Add smoke tests for the Topology UI

j-white:
JIRA: http://issues.opennms.org/browse/HZN-732
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS679


mvrueden:
This looks good to me. If Bamboo goes green, from my point of view, we can merge this.

PR #773, opened 2016-05-03
HZN-735: Add a SearchProvider for the GraphMLTopologyProvider

j-white:
JIRA: http://issues.opennms.org/browse/HZN-735
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS690

### DIFF
--- a/features/topology-map/plugins/org.opennms.features.topology.plugins.topo.graphml/src/main/java/org/opennms/features/topology/plugins/topo/graphml/GraphMLSearchProvider.java
+++ b/features/topology-map/plugins/org.opennms.features.topology.plugins.topo.graphml/src/main/java/org/opennms/features/topology/plugins/topo/graphml/GraphMLSearchProvider.java
@@ -29,34 +29,21 @@
 package org.opennms.features.topology.plugins.topo.graphml;
 
 import java.util.ArrayList;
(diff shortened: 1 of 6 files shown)


features/topology-map/plugins/org.opennms.features.topology.plugins.topo.graphml/src/main/java/org/opennms/features/topology/plugins/topo/graphml/GraphMLSearchProvider.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Override
+    public String getSearchProviderNamespace() {
+        return graphMLTopologyProvider.getVertexNamespace();
+    }
+
+    @Override
+    public List<? extends VertexRef> queryVertices(SearchQuery searchQuery, GraphContainer container) {
+        final List<GraphMLVertex> matchingVertices = new ArrayList<>();
+        final GraphML graphML = graphMLTopologyProvider.getGraphML();
```
mvrueden: I think we should use `graphMLTopologyProvider.getVertices()` and filter on the result instead.
But we can also leave it for now, as this is going to change anyways as soon as we have the layering in place.

PR #774, opened 2016-05-04
HZN-568: More final changes to get Minion Syslogd working :)

soleger:
The opennms-syslogd-handler-default feature on OpenNMS installed some OSGi bundles that conflicted with classes in the system classpath. These changes should resolve that problem and hopefully get the Minion system tests working.

* JIRA: http://issues.opennms.org/browse/HZN-568
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS682

### DIFF
--- a/container/features/src/main/resources/features.xml
+++ b/container/features/src/main/resources/features.xml
@@ -735,6 +735,11 @@
       <feature>camel-blueprint</feature>
       <feature>camel-http</feature>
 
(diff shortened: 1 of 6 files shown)


soleger:
I re-verified that the RPM artifacts from the latest Bamboo for this branch DO fix the MST failures. Finally! :)

PR #780, opened 2016-05-08
NMS-8408: Created theme for Horizon and Meridian

indigo423:
Created default template which was generated from https://github.com/OpenNMS/asciidoctor-stylesheet-factory.

* JIRA: http://issues.opennms.org/browse/NMS-8408
* Bamboo:

### DIFF
--- a/opennms-doc/opennms-horizon.css
+++ b/opennms-doc/opennms-horizon.css
@@ -1,0 +1,677 @@
+/*! normalize.css v2.1.2 | MIT License | git.io/normalize */
+/* ========================================================================== HTML5 display definitions ========================================================================== */
+/** Correct `block` display not defined in IE 8/9. */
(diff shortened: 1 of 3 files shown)


indigo423:
* Cleanup unused default asciidoctor theme
* In sections with deep structure like the Provisioning Guide the sections are not good to read
* Indication with fixed fonts for file names or commands was better in the old version, the orange on white is harder to read and distracts
* Admonition for Info had a neutral blue color and is now green which doesn't fit to the rest of the admonitions. it should keep the neutral blue color or should have a neutral color which fits into the scheme.
* Should the PDF theme adapted as well?

soleger:
Closing since nobody is working on this. I already deleted the stylesheet factory fork that this was based on a while ago.

PR #790, opened 2016-05-11
HZN-742 write junits for blueprint-syslog-handler-kafka.xml

cernerpradeep:
Hi Seth,

I was not able to push to the branch that I created from opennms:opennms develop branch because of access. Since I this code needs to be reviewed, I am creating this PR. Once I get the access, I will create fresh PR. However request you to review the code.

JIRA: http://issues.opennms.org/browse/HZN-742
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-DEL42


features/events/syslog/pom.xml
```diff
@@ -260,5 +266,30 @@
...
+	 <groupId>log4j</groupId>
+	 <artifactId>log4j</artifactId>
+	 <version>1.2.15</version>
+	 <scope>test</scope>
+	</dependency>
+    
+	<dependency>
+	 <groupId>org.slf4j</groupId>
+	 <artifactId>slf4j-log4j12</artifactId>
+	 <version>1.7.21</version>
```
mvrueden: Is there a reason to define the version explicitly?
We usually either use the version-properties e.g. `${slf4jVersion}`(see https://github.com/OpenNMS/opennms/blob/develop/pom.xml#L1294) defined in the root pom, or do not define a version at all - as it is defined in the root pom in most cases.

features/events/syslog/pom.xml
```diff
@@ -260,5 +266,30 @@
...
+	 <groupId>org.apache.curator</groupId>
+	 <artifactId>curator-test</artifactId>
+	 <version>2.0.0-incubating</version>
+	 <scope>test</scope>
+	</dependency>
+    
+    <dependency>
+	 <groupId>log4j</groupId>
+	 <artifactId>log4j</artifactId>
+	 <version>1.2.15</version>
```
mvrueden: Is there a reason to define the version explicitly?
We usually either use the version-properties e.g. ${log4jVersion}(see https://github.com/OpenNMS/opennms/blob/develop/pom.xml#L1282) defined in the root pom, or do not define a version at all - as it is defined in the root pom in most cases.

cernerpradeep: Yes. I tried doing it like that. But I get this error
**Missing artifact log4j:log4j:jar:99.99.99-use-log4j2**. But I when mention the version the errors go away.

mvrueden: Can you try `<version>${log4jVersion}</version>` instead?
This should work.

soleger: The log4j:log4j artifact should be removed. It's the old log4j 1.x artifact so you need to remove it and add exclusions to all artifacts that require log4j:log4j.

To see which dependencies require it, you can run 'mvn dependency:tree' and see which JARs have log4j:log4j as a child.

features/events/syslog/pom.xml
```diff
@@ -110,6 +110,12 @@
     <dependency>
       <groupId>org.opennms.core</groupId>
       <artifactId>org.opennms.core.daemon</artifactId>
+     <exclusions>
```
soleger: This isn't right, you should leave this in. log4j-over-slf4j redirects log4j 1.x calls to slf4j which is what we want to do inside OpenNMS.

features/events/syslog/pom.xml
```diff
@@ -260,5 +266,30 @@
         </exclusion>
       </exclusions>
     </dependency>
+    
+    <!-- HZN-742 -->
+    <dependency>
+	 <groupId>org.apache.curator</groupId>
+	 <artifactId>curator-test</artifactId>
+	 <version>2.0.0-incubating</version>
+	 <scope>test</scope>
```
soleger: You need to exclude log4j:log4j from this curator dependency. Since it is the only thing new here, it must be causing the log4j dependency problems. Just add an <exclusions><exclusion>...</exclusion></exclusions> block with the log4j groupId and artifactId.

soleger:
Pradeep, this is the correct way to make the pull request. After the code passes in Bamboo, I can merge it right into the OpenNMS/opennms repo. Thanks, Seth

features/events/syslog/pom.xml
cernerpradeep: Yes, this is taken care. I excluded the log4j and slf4j, it worked!. Thanks

soleger:
Hi Pradeep, I just merged OpenNMS/opennms:develop back into your branch. Please pull to get the changes.

pom.xml
```diff
@@ -1253,6 +1253,7 @@
     <commonsLoggingVersion>99.99.99-use-jcl-over-slf4j</commonsLoggingVersion>
     <commonsMath3Version>3.5</commonsMath3Version>
     <commonsValidatorVersion>1.3.1</commonsValidatorVersion>
+    <curatorVersion>2.0.0-incubating</curatorVersion>
```
soleger: This version seems old, can you upgrade to a later version? 2.10.0, 3.0.0, or 3.1.0 are available.

features/events/syslog/src/test/java/org/opennms/netmgt/syslogd/SyslogdHandlerKafkaIT.java
```diff
@@ -28,24 +28,112 @@
...
 import org.junit.runner.RunWith;
 import org.opennms.core.test.OpenNMSJUnit4ClassRunner;
-import org.opennms.core.test.camel.CamelBlueprintTest;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.test.context.ContextConfiguration;
 
 @RunWith(OpenNMSJUnit4ClassRunner.class)
 @ContextConfiguration(locations = { "classpath:/META-INF/opennms/emptyContext.xml" })
-public class SyslogdHandlerKafkaIT extends CamelBlueprintTest {
```
soleger: Please use CamelBlueprintTest here. You can still override the doPreSetup() method. If you use CamelBlueprintTest, you can remove the useAdviceWith(), isUseDebugger() and isMockEndpoints() methods.

soleger:
This PR has been superseded by later PRs rebased against develop. Closing PR.

PR #792, opened 2016-05-11
HZN-751: Allow to use the Measurement API to query for values in templates

christianpape:
This branch is based on HZN-729. So, this should only be merged after HZN-729 is merged successfully.

JIRA: http://issues.opennms.org/browse/HZN-751
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS710/latest


mvrueden:
This should only be merged AFTER HZN-729 is merged successfully.

mvrueden:
Jesse pointed something out:

> If we add graphs, can we try and do it using our 'graph.js' so that it can work with both rrdgraph and Backshift?
> Here's  an example graph definition: https://github.com/OpenNMS/opennms/blob/develop/opennms-webapp/src/main/webapp/WEB-INF/jsp/graph/results.jsp#L217
> The following snippet sets up graph.js to process the graph-container divs: https://github.com/OpenNMS/opennms/blob/develop/opennms-webapp/src/main/webapp/includes/bootstrap.jsp#L120

christianpape:
I did not manage to get the graph.js stuff running in templates. It seems that JavaScript is not executed when placed inside a Vaadin Label even when declared with ContentMode.HTML. Please see JIRA issue for more details...

PR #798, opened 2016-05-12
Fixed status values inside JDBCMonitor and WebMonitor

soleger:
Fixes:
- If JDBCMonitor is configured with a mismatching JDBC driver and URL, it always returns "Available"
- WebMonitor returns "Available" when an exception (including connect exceptions, timeouts, etc) happens during polling

Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS722/


mvrueden:
Looks good to me. 
The error message `Wrong kind of JDBC driver ({})` sounds weird to me. 
Maybe just `Wrong JDBC driver`?
But it could also be just that I am not a native speaker ;)

indigo423:
Is it worth to add this to JIRA for release notes?

soleger:
I added a JIRA for the WebMonitor problem (which is the most severe since it can return "Available" when the service is down). http://issues.opennms.org/browse/NMS-8427

PR #802, opened 2016-05-17
HZN-760: Write smoke tests for the GraphML topology provider

mvrueden:
* JIRA:http://issues.opennms.org/browse/HZN-760
* BAMBOO: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS728/


features/graphml/src/main/java/org/graphdrawing/graphml/DataExtensionType.java
```diff
@@ -0,0 +1,85 @@
+//
```
j-white: Should add our copyright headers on these files?

PR #805, opened 2016-05-17
NMS-8404: Consolidate the node[] and nodeSource[] resource types

j-white:
JIRA: http://issues.opennms.org/browse/NMS-8404
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS727


opennms-dao/src/main/java/org/opennms/netmgt/dao/support/AbstractTopLevelResourceType.java
```diff
@@ -47,7 +48,10 @@ public boolean isResourceTypeOnParent(OnmsResource parent) {
 
     @Override
     public List<OnmsResource> getResourcesForParent(OnmsResource parent) {
-        return getTopLevelResources();
+        if (parent == null) {
```
agalue: Is this correct ?

j-white: Yep, "top-level" resources can't have parents.

agalue: Ok. So, why having an OnmsResource as a parameter ?

j-white: It's a generic method inherited from OnmsResourceType. Not all resource types are "top-level" resources.

agalue: I understand, but the parameter is never used. Having this in mind, is it correct to assume that there are implementations of AbstractTopLevelResourceType that override this method and use the provided parameter ?

j-white: Implementations of AbstractTopLevelResourceType do not override this method.

agalue: Then, why having that parameter instead of modifying the method signature to be the following ?

 public List<OnmsResource> getResourcesForParent();

j-white: That parameter is used by other resource types:
https://github.com/OpenNMS/opennms/blob/develop/opennms-dao/src/main/java/org/opennms/netmgt/dao/support/InterfaceSnmpResourceType.java#L117

agalue: That's what I tried to ask a few comments before ;) thanks!

agalue:
Looks good to me, let's wait for Bamboo.

opennms-dao/src/main/java/org/opennms/netmgt/dao/support/AbstractTopLevelResourceType.java
j-white: Cool, sorry for the confusion :)

jeffgdotorg:
Sorry I haven't been in the mix on this one; doing remote consulting for a customer today. Thanks for the amazing and quick work!

PR #806, opened 2016-05-17
NMS-8429: Add a new control to open graphs in their own window.

j-white:
JIRA: http://issues.opennms.org/browse/NMS-8429
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS733-1


agalue:
I tried this and works perfect!

I'd love to be able to remove the horrible GWT icon for the KSC-Add report feture in the near future, to make all the icons use the same format.

Either way, I'll merge as soon as Bamboo goes green.

agalue:
One thing I noticed is that the icon appears when showing a single graph. I think this can be easily avoided by checking if "reports" is either "all" or have more than one value. If this is the case, show the button, otherwise hide it (as it is being shown only one graph).

PR #808, opened 2016-05-18
NMS-8393: changes to address clickjacking vulnerability

gallenc:
This issue addresses a perceived cross scripting  'clickjacking' vulnerability described in  http://issues.opennms.org/browse/NMS-8393

This change is needed for an enterprise customer who has tested this patch against meridian 2015 and will be upgrading to meridian 2016

User instructions for configuring the patch are also described on the OpenNMS wiki at
https://www.opennms.org/wiki/Nessus_Security_Fixes

This patch simply adds the required jetty-rewrite jar to the opennms/lib directory and additional options in the /etc/examples/jetty.xml. This makes it easier for the user to apply this configuration but by default this feature is not turned on. The user must follow the instructions in https://www.opennms.org/wiki/Nessus_Security_Fixes to uncomment the relevant sections in jetty.xml

This fix has been tested in production with no perceived issues but it is not enabled by default as there may be unforeseen impacts to the UI or on users who embed OpenNMS screens as x-frames in other pages.

* JIRA:  http://issues.opennms.org/browse/NMS-8393
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS725-2

### DIFF
--- a/opennms-jetty/pom.xml
+++ b/opennms-jetty/pom.xml
@@ -91,6 +91,11 @@
       <version>${jettyVersion}</version>
     </dependency>
     <dependency>
+      <groupId>org.eclipse.jetty</groupId>
+      <artifactId>jetty-rewrite</artifactId>
+      <version>${jettyVersion}</version>
+    </dependency>
+    <dependency>
       <groupId>org.eclipse.jdt.core.compiler</groupId>
       <artifactId>ecj</artifactId>
       <scope>runtime</scope>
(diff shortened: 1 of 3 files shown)


RangerRick:
I'm not sure the consequences of this, but is there a reason we wouldn't enable it by default?

gallenc:
Hi,

I am not sure of all the consequences but I think the header setting
prevents browsers from  opening other sources in iframes. It might
effect users of the wall board who are pulling in content from other
sources. However I haven't checked all the possibilities. I thought for
now leave this as default off and let people report any problems if they
use it.

Craig

j-white:
Given that we're unsure of the full scope of consequences I think we're better off leaving it off by default for now.

PR #812, opened 2016-05-18
NMS-8430: Fixed issue where the number of ring buffer entries exceeds the capacity

j-white:
JIRA: http://issues.opennms.org/browse/NMS-8430
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS743


features/newts/src/main/java/org/opennms/netmgt/newts/NewtsWriter.java
```diff
@@ -165,13 +172,13 @@ public void insert(List<Sample> samples) {
                     .map(s -> s.getResource().getId())
```
soleger: This stream() call is called during each failure to enqueue on the ring buffer. According to the ratelimitedlogger docs, if you were to wrap it in an object whose toString() returned the value, the call would be avoided unless the ratelimitedlogger decides to actually print the message. That might save you a lot of overhead here when the ring buffer is full.

j-white: Good point, I went ahead and wrapped it in a toString() method.

PR #815, opened 2016-05-20
HZN-754: Moved generic info panel item provider to topology-app

christianpape:
Jira: http://issues.opennms.org/browse/HZN-754
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS737


features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/info/GenericInfoPanelItemProvider.java
```diff
@@ -205,33 +203,17 @@ public int getOrder() {
                 }
             }
         }
-        if (vertex instanceof GraphMLVertex) {
-            final GraphMLVertex atlasVertex = (GraphMLVertex) vertex;
-            context.putAll(atlasVertex.getProperties());
-        }
+
+        context.putAll(vertex.getProperties());
```
mvrueden: Can we use reflections instead of only populating the context?

PR #817, opened 2016-05-20
HZN-769: Recommend Cassandra 3.0.x instead of 2.1.x

j-white:
JIRA: http://issues.opennms.org/browse/HZN-769
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS751

This PR updates the installation guide to install Cassandra 3.0.x instead of 2.1.x. We also include some additional links and clarify some of the assumptions that were made.

The default configuration files were also updated to remove the required categories, and remove the version from the package names.


opennms-base-assembly/src/main/filtered/etc/collectd-configuration.xml
```diff
@@ -3,37 +3,29 @@
 <collectd-configuration
         threads="50">
 
-    <package name="cassandra21x">
-        <filter><![CDATA[(IPADDR != '0.0.0.0') & (categoryName == 'Cassandra21x')]]></filter>
+    <package name="cassandra-via-jmx">
+        <filter>IPADDR != '0.0.0.0'</filter>
```
indigo423: 👍 to bind it just on the service name

opennms-base-assembly/src/main/filtered/etc/collectd-configuration.xml
```diff
@@ -3,37 +3,29 @@
...
-            <parameter key="friendly-name" value="cassandra21x"/>
-            <parameter key="collection" value="cassandra21x"/>
+            <parameter key="collection" value="cassandra30x"/>
+            <parameter key="friendly-name" value="cassandra"/>
             <parameter key="thresholding-enabled" value="true"/>
             <parameter key="factory" value="PASSWORD-CLEAR"/>
             <parameter key="username" value="cassandra-username"/>
             <parameter key="password" value="cassandra-password"/>
         </service>
-    </package>
```
indigo423: 👍 to have them in one package.

opennms-base-assembly/src/main/filtered/etc/collectd-configuration.xml
```diff
@@ -3,37 +3,29 @@
...
             <parameter key="retry" value="2"/>
             <parameter key="timeout" value="3000"/>
             <parameter key="protocol" value="rmi"/>
             <parameter key="urlPath" value="/jmxrmi"/>
-            <parameter key="rrd-base-name" value="cassandra21x"/>
-            <parameter key="ds-name" value="cassandra21x"/>
-            <parameter key="friendly-name" value="cassandra21x"/>
-            <parameter key="collection" value="cassandra21x"/>
+            <parameter key="collection" value="cassandra30x"/>
+            <parameter key="friendly-name" value="cassandra"/>
```
indigo423: Just a question, is there anything we have to take care of, in case we move from cassandra 21x to cassandra3x on a system using a similar friendly name?

indigo423:
I like the changes. Get Cassandra and Newts in one configuration packages and get rid of the need for categories is very good in my point of view. I like this improvement beside the question I have 👍 for me.

opennms-base-assembly/src/main/filtered/etc/collectd-configuration.xml
j-white: Good point. We should probably mention this in the release notes.

opennms-base-assembly/src/main/filtered/etc/poller-configuration.xml
```diff
@@ -36,31 +35,14 @@
...
       <parameter key="retry" value="2"/>
       <parameter key="timeout" value="3000"/>
       <parameter key="protocol" value="rmi"/>
       <parameter key="urlPath" value="/jmxrmi"/>
-      <parameter key="rrd-base-name" value="cassandra21x-newts"/>
-      <parameter key="ds-name" value="cassandra21x-newts"/>
-      <parameter key="friendly-name" value="cassandra21x-newts"/>
-      <parameter key="collection" value="cassandra21x-newts"/>
+      <parameter key="rrd-base-name" value="jmx-cassandra-newts"/>
+      <parameter key="ds-name" value="jmx-cassandra-newts"/>
```
agalue: What if someone wants to monitor a Cassandra cluster from an OpenNMS instance storing data on RRD files ? In this case, we should consider the limitation of 19 characters for the DS names.

agalue:
I agree with @indigo423 with the new structure for collectd-configuration.xml and poller-configuration.xml. Now, we should definitively put some notes on the release notes, if we're not planning on having an upgrade tool task to preserve the collected statistics for those users that are already using Cassandra and want to upgrade to a newer version of OpenNMS, preserving the data.

opennms-base-assembly/src/main/filtered/etc/poller-configuration.xml
j-white: "jmx-cassandra-newts" is 19 characters long, am I missing something?

agalue:
Fortunately, Java8 is your friend ... All Attribs and CompAttribs are OK in terms of alias size:

```
    @Test
    public void checkJmxConfig() throws Exception {
        JmxDatacollectionConfig cfg = JaxbUtils.unmarshal(JmxDatacollectionConfig.class, ConfigFileConstants.getFile(ConfigFileConstants.JMX_DATA_COLLECTION_CONF_FILE_NAME));
        List<String> collections = cfg.getJmxCollectionList().stream().map(c -> c.getName()).collect(Collectors.toList());
        JMXDataCollectionConfigFactory.init();
        
        collections.forEach(c -> {
            long badDs = JMXDataCollectionConfigFactory.getInstance().getJmxCollection(c).getMbeans().stream()
                    .flatMap(m -> m.getAttribList().stream())
                    .filter(a -> a.getAlias().length() > 19 && !a.getType().toLowerCase().contains("string"))
                    .peek(a -> System.err.println(a.getAlias()))
                    .count();
            badDs += JMXDataCollectionConfigFactory.getInstance().getJmxCollection(c).getMbeans().stream()
                    .flatMap(m -> m.getCompAttribList().stream())
                    .filter(a -> a.getAlias().length() > 19 && !a.getType().toLowerCase().contains("string"))
                    .peek(a -> System.err.println(a.getAlias()))
                    .count();
            Assert.assertEquals(0, badDs);            
        });
    }
```
That means, I'll merge the change.

agalue:
@j-white  This has to be manually merged to develop because the JMX config was divided into multiple files there, so guess I let you do it.

PR #818, opened 2016-05-20
NMS-8363: Newts converter: Skip values that fall outside the supported range.

j-white:
JIRA: http://issues.opennms.org/browse/NMS-8363
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS750


j-white:
Retargeting this at foundation-2016.

PR #821, opened 2016-05-23
NMS-8400: add common activemq component

cgorantla:
activemq component to be used by osgi services.

* JIRA: http://issues.opennms.org/browse/nms-8400
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS746/


features/activemq/component/pom.xml
```diff
@@ -0,0 +1,31 @@
...
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.opennms.features</groupId>
+    <artifactId>org.opennms.features.activemq</artifactId>
+    <version>19.0.0-SNAPSHOT</version>
+  </parent>
+  <packaging>bundle</packaging>
+  <groupId>org.opennms.features.activemq</groupId>
+  <artifactId>component</artifactId>
```
soleger: Can you change this artifactId to "org.opennms.features.activemq.component"? That will make the JAR name easier to identify (since it would be difficult to tell what "component-19.0.0.jar" contains just by looking at its name).

cgorantla:
Updated artifact to org.opennms.features.activemq.component in 4694571

PR #833, opened 2016-05-27
NMS-8456: Tweaked performance of syslog Camel routes

soleger:
Disabled "persistent" JMS messages for syslog to avoid request-reply cycle for each message. Added concurrentConsumers to seda and activemq endpoints to thread the message handling. Verified that all routes are executing asynchronously.

* JIRA: http://issues.opennms.org/browse/NMS-8456
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS766


soleger:
I've cherry-picked this change to release-18.0.1, closing PR.

PR #836, opened 2016-06-01
HZN-784:  Add basic detectors to minion shell

cgorantla:
basic detectors can be launched from minion shell

* JIRA: http://issues.opennms.org/browse/hzn-784


container/features/src/main/resources/features-minion.xml
```diff
@@ -47,5 +47,15 @@
 	  <bundle>blueprint:mvn:org.opennms.features.minion/core-impl/${project.version}/xml/blueprint-heartbeat</bundle>
 	  <feature>minion-core-api</feature>
 	</feature>
+
+    <feature name="opennms-detector-minion" description="OpenNMS :: Minion :: Detector" version="${project.version}">
```
j-white: This feature should be renamed to *opennms-detector-simple* and moved into the other features.xml file since it is not Minion specific.

features/minion/core/features/src/main/resources/features.xml
```diff
@@ -27,5 +30,50 @@
     <feature name="minion-shell" description="Minion :: Shell" version="${project.version}">
         <feature>minion-core</feature>
         <bundle>mvn:org.opennms.features.minion/shell/${project.version}</bundle>
+        <feature version="[4.0,4.1)">spring</feature>
```
j-white: Good job on putting this list together, but there are way too many dependencies here. These should be broken out into separate features, many of which already exist.

features/minion/core/shell/src/main/java/org/opennms/minion/core/shell/DetectorTypeCompleter.java
```diff
@@ -0,0 +1,60 @@
...
+import org.apache.karaf.shell.api.action.lifecycle.Service;
+import org.apache.karaf.shell.api.console.CommandLine;
+import org.apache.karaf.shell.api.console.Completer;
+import org.apache.karaf.shell.api.console.Session;
+import org.apache.karaf.shell.support.completers.StringsCompleter;
+
+@Service
+public class DetectorTypeCompleter implements Completer {
+
+    public static final String ICMP = "icmp";
```
j-white: We should avoid hard coding the list of detectors and discover them from the service registry instead,

opennms-provision/opennms-detector-simple/blueprint-detector.xml
```diff
@@ -0,0 +1,50 @@
...
+		http://aries.apache.org/blueprint/xmlns/blueprint-ext/v1.1.0
+		http://aries.apache.org/schemas/blueprint-ext/blueprint-ext-1.1.xsd
+">
+
+	<bean id="icmp" class="org.opennms.netmgt.provision.detector.icmp.IcmpDetector">
+	</bean>
+	<service id="icmpDetector" ref="icmp"
+		interface="org.opennms.netmgt.provision.SyncServiceDetector">
+		<service-properties>
+			<entry key="alias" value="detector.icmp" />
```
soleger: Can we change these aliases to be:
```
<entry key="implementation" value="org.opennms.netmgt.provision.detector.icmp.IcmpDetector" />
```
instead of using another constant for the value?

features/minion/core/features/src/main/resources/features.xml
```diff
@@ -1,6 +1,9 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <features name="minion-core-features" xmlns="http://karaf.apache.org/xmlns/features/v1.2.0">
 
+    <!-- OpenNMS Features -->
+    <repository>mvn:org.opennms.karaf/opennms/${project.version}/xml/features</repository>
```
j-white: We don't want to import these dependencies with the "core" features, let's create a new project for the provisioning shell commands add with the "default" features instead.

j-white:
Looks good for a first pass. The only thing missing from HZN-784 is the ability to set the detector's attributes.

container/features/src/main/resources/features.xml
```diff
@@ -848,4 +848,16 @@
         <bundle>mvn:org.opennms.features.ticketing/org.opennms.features.ticketing.api/${project.version}</bundle>
         <bundle>mvn:org.opennms.features/opennms-integration-tsrm/${project.version}</bundle>
     </feature>
+    
+    
+    <feature name="opennms-detector-simple" description="OpenNMS :: Minion :: Detector" version="${project.version}">
```
j-white: The description here should be `OpenNMS :: Provision :: Simple Detector`

features/minion/provisiond-shell/src/main/java/org/opennms/minion/provisiond/shell/MinionDetector.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    private void setAttributes(Map<String, String> properties,
+            SyncServiceDetector detector) {
+
+        String ipMatch = properties.get("ipMatch");
+        String port = properties.get("port");
+        String timeout = properties.get("timeout");
+
+        detector.setIpMatch(ipMatch);
```
j-white: We should set the attributes the same way that provisiond does when it instantiates the detectetors. See https://github.com/OpenNMS/opennms/blob/develop/opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/DefaultPluginRegistry.java#L139

cgorantla:
Added ability to set attributes and  moved Provisiond detector to a different project

PR #837, opened 2016-06-06
Clean up task API for Provisiond

soleger:
To investigate:

* http://issues.opennms.org/browse/NMS-5593
* http://issues.opennms.org/browse/NMS-6760
* http://issues.opennms.org/browse/NMS-7860

I've cleaned up the task classes behind Provisiond. There are interfaces for things that should have interfaces and I've changed the single-threaded completion service in DefaultTaskCoordinator to use a single-threaded Executor that is invoked with CompletableFuture. This makes the code simpler and easier to read and should log and handle exceptions better.

* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS771


RangerRick:
Looks good to me.  Seems like all pretty straightforward changes.  I say ship it!

agalue:
Some of the referenced issues have Meridian on its target, and this PR is based on develop. So, either we should rebase this PR to foundation-2016, or change the issues to have Meridian-2017 and Horizon 19 on their target versions.

RangerRick:
Yeah, when he was talking about it in scrum this morning, I think the idea was that we're way too close to Meridian release to risk changes to provisiond now, so it would have to be targeted to Meridian 2017 or maybe backported to 2016.1.1 or something in the future after we've confirmed it's solid in 19.0.0 or something.

soleger:
Yeah, let's defer the JIRA issues to Horizon 19. My hope is that the improved logging in these commits might catch problems in individual detectors or tasks for people running 19. Then we can go back to foundation-2016 and make smaller, targeted changes to fix, for instance, one detector that might be causing a task to fail to complete.

I also didn't want to merge to foundation-2016 because like I said, I'm still not 100% sure there aren't other task ordering or exception handling problems in addition to the ones I fixed here. Since it's not a 100% fix, let's try it out in Horizon first.

agalue:
Makes sense! let's update those Jira issues :)

PR #838, opened 2016-06-07
NMS-8485: dockerized smoke tests

RangerRick:
This pull request switches the smoke tests in foundation-2016 to use docker containers to run OpenNMS and PostgreSQL, rather than relying on a locally-installed OpenNMS when bamboo runs.

Note that by default, the tests will still run the old way (assuming 127.0.0.1:8980 is an OpenNMS system).  To enable dockerized tests, pass -Dorg.opennms.smoketest.docker=true on the maven command-line.

```
../compile.pl -Dorg.opennms.smoketest.docker=true -t verify
```

This branch uses the [newer smoke-test API branch, version 2-SNAPSHOT](https://github.com/OpenNMS/smoke-tests/blob/2/test-api/src/main/java/org/opennms/smoketest/OpenNMSSeleniumTestCase.java).

* JIRA: http://issues.opennms.org/browse/NMS-8485
* Bamboo:http://bamboo.internal.opennms.com:8085/browse/OPENNMS-DST0-17/

### DIFF
--- a/smoke-test/src/test/java/org/opennms/smoketest/JmxConfigurationGeneratorIT.java
+++ b/smoke-test/src/test/java/org/opennms/smoketest/JmxConfigurationGeneratorIT.java
@@ -1,6 +1,7 @@
 package org.opennms.smoketest;
 
 import java.util.Collection;
+import java.util.concurrent.TimeUnit;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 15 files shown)


RangerRick:
Also note, most of the real work happened in the [smoke-test project](https://github.com/OpenNMS/smoke-tests/blob/2/test-api/src/main/java/org/opennms/smoketest/OpenNMSSeleniumTestCase.java), which is isolated from being used unless our Bamboo smoke-test scripts detect the usage of `2-SNAPSHOT` in `smoke-test/pom.xml`.

RangerRick:
DO NOT merge until the following bamboo branches are green:

* foundation-2016+docker: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-DST0-17/
* release-18.0.1+docker: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS779-1/
* develop+docker: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS773-4/

Once they all pass and this pull request is merged, we can merge the release-18.0.1+docker branch which includes all of the foundation changes plus the minion system tests.  This should be done in a separate PR.

smoke-test/src/test/java/org/opennms/smoketest/JmxConfigurationGeneratorIT.java
```diff
@@ -106,39 +110,79 @@ public void verifyCompMemberSelection() throws InterruptedException {
...
+        boolean found = false;
+        do {
+            setVaadinValue("port", "18980");
+            setVaadinValue("authenticateUser", "admin");
+            setVaadinValue("authenticatePassword", "admin");
+
+            // go to next page
+            waitForElement(By.id("next")).click();
+
+            try {
```
soleger: Can you add some comments to all of these lines to outline why you had to do all of these backflips? It might be difficult to understand for anyone coming into this code later.

RangerRick: good call; done

PR #841, opened 2016-06-08
NMS-8484 - Add custom string attributes based on indirect and complex SNMP Indices

agalue:
Add custom string attributes based on indirect and complex SNMP Indices

* JIRA: http://issues.opennms.org/browse/NMS-8484
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS777

The idea is be able to add additional string attributes (i.e. those that are going to be stored in strings.properties) to use them when labeling the resource and inside of graph templates.

There are several MIB tables that don't have references to the source, just the statistics, and the source is embedded inside the index. This index should be decomposed and use the partial components on another SNMP tables in order to figure out the source.

Here is an example from AIRESPACE-WIRELESS-MIB:

Let's say you want to collect data from:

```
bsnAPIfLoadNumOfClients OBJECT-TYPE
    SYNTAX          Integer32
    MAX-ACCESS      read-only
    STATUS          current
    DESCRIPTION
        "This is the number of clients attached to this Airespace
        AP at the last measurement interval(This comes from 
        APF)"
    ::= { bsnAPIfLoadParametersEntry 4 }
```

The table on which the above metric is defined it is called bsnAPIfLoadParametersTable, and its index is a composed index formed by: bsnAPDot3MacAddress and bsnAPIfSlotId. It would be more useful to show on the resource label (and the graphs) when listing the resources associated with this metric, the name of the access point and the slot.

Currently, this is not possible (which is the reason of this PR).

The name of the AP is defined on another table indexed by bsnAPDot3MacAddress:

```
bsnAPName OBJECT-TYPE
    SYNTAX          OCTET STRING (SIZE  (0..32))
    MAX-ACCESS      read-write
    STATUS          current
    DESCRIPTION
        "Name assigned to this AP. If an AP is not configured its
        factory default name will be ap:<last three byte of 
        MACAddress> eg. ap:af:12:be"
    ::= { bsnAPEntry 3 }
```

With the changes described on this PR (which are optional and non intrusive), you can put the following on your data collection configuration:

```
<resourceType name="bsnAPEntry" label="Cisco Wireless AP" resourceLabel="${bsnAPName} (index ${index})">
    <persistenceSelectorStrategy class="org.opennms.netmgt.collection.support.PersistAllSelectorStrategy" />
    <storageStrategy class="org.opennms.netmgt.collection.support.IndexStorageStrategy" />
</resourceType>

<resourceType name="bsnAPIfLoadParametersEntry" label="Cisco Wireless AP Resources" resourceLabel="AP ${bsnAPName} Slot ${slotNumber} (index ${index})">
    <persistenceSelectorStrategy class="org.opennms.netmgt.collection.support.PersistAllSelectorStrategy" />
    <storageStrategy class="org.opennms.netmgt.collection.support.IndexStorageStrategy" />
</resourceType>

<group name="bsnAPTable" ifType="all">
    <mibObj oid=".1.3.6.1.4.1.14179.2.2.1.1.3" instance="bsnAPEntry" alias="bsnAPName" type="string" />
</group>

<group name="bsnAPIfLoadParametersTable" ifType="all">
    <mibObj oid=".1.3.6.1.4.1.14179.2.2.13.1.4" instance="bsnAPIfLoadParametersEntry" alias="bsnAPIfLoadNumOfCli" type="integer" />
    <property instance="bsnAPIfLoadParametersEntry" alias="bsnAPName">
        <parameter key="source-type" value="bsnAPEntry" />
        <parameter key="source-alias" value="bsnAPName" />
        <parameter key="index-pattern" value="^(.+)\.\d+$" />
    </property>
    <property instance="bsnAPIfLoadParametersEntry" alias="slotNumber" class-name="org.opennms.netmgt.collectd.IndexSplitPropertyExtender">
        <parameter key="index-pattern" value="^.+\.(\d+)$" />
    </property>
</group>
```

Now, the resources associated with bsnAPIfLoadParametersEntry are going to have 2 new string properties: bsnAPName and slotNumber. If the class-name is not specified on a property tag, the default will be: org.opennms.netmgt.collectd.RegExPropertyExtender.

With this approach we can implement something to handle even more complex scenarios like Cisco QoS.

Because this solution is non intrusive, I think it is safe to add it to meridian 2016 (which is why the solution is based on foundation-2016).


opennms-config-jaxb/src/main/java/org/opennms/netmgt/config/datacollection/Group.java
```diff
@@ -271,6 +283,13 @@ public boolean equals(final Object obj) {
         } else if (!m_mibObjects.equals(other.m_mibObjects)) {
             return false;
         }
+        if (m_mibObjProperties == null) {
+            if (other.m_mibObjProperties != null) {
+                return false;
+            }
+        } else if (!m_mibObjProperties.equals(other.m_mibObjProperties)) {
```
j-white: MibObjProperty does not implement hashCode() and equals(), so this may not work as expected.

agalue: > On Jun 8, 2016, at 12:19 PM, Jesse White <notifications@github.com> wrote:
> 
> MibObjProperty does not implement hashCode() and equals(), so this may not work as expected.


Good point!

I'll implement those methods, using similar logic like the other datacollection-config classes.

PR #843, opened 2016-06-09
HZN-802: Move the DistPollerDao shim, fix system IDs on Minion syslog events

soleger:
To let the Syslogd code use the same API to look up the local system ID, I moved the DistPollerDao shim out of Discovery and into a common bundle that can be reused by all shared OpenNMS/Minion modules. This PR also adds the Minion ID to SyslogConnection messages so that they are persisted by Eventd correctly.

I also moved consolidated the empty Spring and blueprint contexts in side the core/test-api bundles so that they are easier to reuse.

* JIRA: http://issues.opennms.org/browse/HZN-802
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS776


RangerRick:
Looks pretty straightforward to me... Mostly just rearranging code.

PR #848, opened 2016-06-09
hzn-818: Add factory classes for detectors 

cgorantla:
Add factory classes for detectors in order to support multiple instances being created 
* JIRA: http://issues.opennms.org/browse/hzn-818
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS784-3

### DIFF
--- a/opennms-provision/opennms-detector-simple/blueprint-detector.xml
+++ b/opennms-provision/opennms-detector-simple/blueprint-detector.xml
@@ -15,7 +15,8 @@
 	<service id="icmpDetector" ref="icmp"
 		interface="org.opennms.netmgt.provision.SyncServiceDetector">
 		<service-properties>
-           <entry key="implementation" value="org.opennms.netmgt.provision.detector.icmp.IcmpDetector" />
+			<entry key="implementation"
+				value="org.opennms.netmgt.provision.detector.icmp.IcmpDetector" />
 		</service-properties>
 	</service>
 
@@ -24,7 +25,8 @@
 	<service id="loopDetector" ref="loop"
 		interface="org.opennms.netmgt.provision.SyncServiceDetector">
 		<service-properties>
-			<entry key="implementation" value="org.opennms.netmgt.provision.detector.loop.LoopDetector" />
+			<entry key="implementation"
+				value="org.opennms.netmgt.provision.detector.loop.LoopDetector" />
 		</service-properties>
 	</service>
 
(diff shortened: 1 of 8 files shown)


opennms-provision/opennms-detector-simple/blueprint-detector.xml
```diff
@@ -15,7 +15,8 @@
 	<service id="icmpDetector" ref="icmp"
```
j-white: Do we still need the detectors to be exposed if we have the factories?

opennms-provision/opennms-detector-simple/src/main/java/org/opennms/netmgt/provision/detector/factory/IcmpDetectorFactory.java
```diff
@@ -0,0 +1,42 @@
...
+package org.opennms.netmgt.provision.detector.factory;
+
+import org.opennms.netmgt.provision.DetectorFactory;
+import org.opennms.netmgt.provision.detector.icmp.IcmpDetector;
+
+public class IcmpDetectorFactory implements DetectorFactory<IcmpDetector> {
+
+    @Override
+    public IcmpDetector createDetector() {
+        IcmpDetector detector = new IcmpDetector();
```
j-white: We can get rid of these temporary variables and just `return new *;`

opennms-provision/opennms-detector-simple/blueprint-detector.xml
cgorantla: yes, using them for getting the names in the Completer,  I can use the same factory there as well.

opennms-provision/opennms-detector-simple/blueprint-detector.xml
```diff
@@ -43,7 +46,52 @@
...
+	</service>
+
+	<bean id="loopFactory"
+		class="org.opennms.netmgt.provision.detector.factory.LoopDetectorFactory">
+	</bean>
+	<service id="loopDetectorFactory" ref="loopFactory"
+		interface="org.opennms.netmgt.provision.DetectorFactory">
+		<service-properties>
+			<entry key="implementation"
+				value="org.opennms.netmgt.provision.detector.factory.LoopDetectorFactory" />
```
j-white: I would use the class name of the detector here instead.

cgorantla:
review comments are taken care in https://github.com/OpenNMS/opennms/pull/852

PR #849, opened 2016-06-10
NMS-8477: NRT graph interval should default to > 1s

jeffgdotorg:
* JIRA: http://issues.opennms.org/browse/NMS-8477


jeffgdotorg:
Wrong base. Again. I always do that.

PR #853, opened 2016-06-13
NMS-8376: Better synchronization in Poller event processing

soleger:
The DefaultPollerContext was using synchronized() blocks to control access to the list of pending event updates. This commit changes the code to use ConcurrentLinkedQueue so that block-level synchronization isn't necessary and improves the synchronization inside the onEvent() method so that it iterates less and also uses ConcurrentLinkedQueue inside PendingPollEvent.

* JIRA: http://issues.opennms.org/browse/NMS-8376
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS793


j-white:
Looks good, and should solve the issue mentioned in NMS-8376.

I'm am nervous about making these changes directly in foundation-2016 when we're so close to a Meridian release though.

soleger:
These changes have been merged into foundation-2016. Closing PR.

PR #859, opened 2016-06-15
HZN-792: Add breadcrumbs for layer navigation

mvrueden:
* JIRA: http://issues.opennms.org/browse/HZN-792
* BAMBOO: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS770


j-white:
As discussed, we should probably add some smoke tests for this :)

mvrueden:
I added smoke tests and also fixed the issue where the source was not correct (source vs target).
Should be mergeable if bamboo goes green.

PR #860, opened 2016-06-16
NMS-8493: Removed some system classpath feature dependencies

soleger:
Several Karaf features were installing bundles that conflicted with the system classpath on OpenNMS. These bundles are now excluded from the features.

* JIRA: http://issues.opennms.org/browse/NMS-8493
* Bamboo: TBD


soleger:
Cherry-picked this back to develop. Closing PR.

PR #866, opened 2016-06-20
NMS-8503: Do not override attributes. Migrate test to use annotations.

mvrueden:
* JIRA: http://issues.opennms.org/browse/NMS-8503
* BAMBOO: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS808


mvrueden:
Ronny verified this and the attributes are now available.

PR #869, opened 2016-06-21
Changed artifactIds for container, ncs, and topology-map projects

soleger:
I changed the Maven artifact names for the JARs created in the container, ncs, and topology-map projects so that they are fully qualified (ie. org.opennms.features.topology.plugins.topo.history) so that they are easier to differentiate in the Eclipse UI when you have lots of modules imported. It will also make it less likely that JAR names will overlap in the future.

This will make the project a little more developer-friendly in preparation for Dev-Jam.

* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS813


RangerRick:
Looks pretty straightforward to me, and smoke tests passed so it should be safe.

agalue:
we should do the same for the rest of the projects.

PR #874, opened 2016-06-23
NMS-8520: Add changeDate for updating incident

cgorantla:
* JIRA: http://issues.opennms.org/browse/nms-8520
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS819


indigo423:
@cgorantla the problem is the "The ticket change date is the same as the reported date" the code change just reflect something about changedBy field which is not the problem?

indigo423:
@cgorantla Thank you very much, I'll verify this issue today.

indigo423:
@cgorantla I've build and ran the code and it works. Function wise +1 from me.

PR #876, opened 2016-06-23
Simplified Scriptd by using java.util.concurrent classes

soleger:
Cleanup in Scriptd to simplify the code.

* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS816


opennms-services/src/main/java/org/opennms/netmgt/scriptd/Executor.java
```diff
@@ -61,173 +63,119 @@
...
-        m_eventScriptMap = new ConcurrentHashMap<String,List<EventScript>>();
-
-        for (int i = 0; i < scripts.length; i++) {
-            Uei[] ueis = scripts[i].getUei();
+        for (EventScript script : m_config.getEventScripts()) {
+            Uei[] ueis = script.getUei();
 
             if (ueis.length == 0) {
-                m_eventScripts.add(scripts[i]);
+                m_eventScripts.add(script);
```
RangerRick: Is there any way loadConfig() could be called multiple times?  Since it's a COW list, could we end up with duplicates if we reload configs?

soleger: Yep, I think this is a problem with the way that I made the list final. I'll work on a fix for that.

PR #879, opened 2016-06-27
NMS-8526: Only replace UUID on initial install

soleger:
Right now, the Minion RPM spec will generate a new UUID during every installation. Instead, this should only be done once during the first installation.

* JIRA: http://issues.opennms.org/browse/NMS-8526
* Bamboo: TBD


soleger:
Merged into develop, closing PR.

PR #882, opened 2016-06-28
NMS-8345: Created opennms-webapp-remoting RPM/DEB

soleger:
This RPM will reduce the size of the opennms-webapp-jetty RPM to speed up installations for people who are not using the Remote Poller.

* JIRA: http://issues.opennms.org/browse/NMS-8345
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS828


soleger:
Oops, I guess I need to do this for Debian as well. ;-)

PR #883, opened 2016-06-28
NMS-8519: Fixed incorrect HRULE syntax

soleger:
Fixed type on HRULE.

* JIRA: http://issues.opennms.org/browse/NMS-8519
* Bamboo: TBD


soleger:
Merged into foundation, closing PR.

PR #884, opened 2016-06-29
NMS-8488: SQL statement and connection cleanup in a finally block

soleger:
Moved JDBC connection and statement cleanup into a finally block so that JDBC connections aren't left open.

* JIRA: http://issues.opennms.org/browse/NMS-8488
* Bamboo: TBD


soleger:
Merged into foundation, closing PR.

PR #893, opened 2016-07-05
NMS-8588: Only send physical bounds if changed.

mvrueden:
* JIRA: http://issues.opennms.org/browse/NMS-8588
* BAMBOO: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS842


j-white:
Works good, no more stuttering :+1: 

Waiting on Bamboo.

PR #894, opened 2016-07-06
NMS-8537: Fix Discovery's per-range/specific/includeUrl foreignSource and location

soleger:
During refactoring of Discovery for Minion in OpenNMS 18, we broke the (undocumented) ability to have individual ranges, specific IPs, and include URLs produce nodes in separate foreign sources. We need to fix this regression and also add the ability to configure these fields in the discovery UI.

* JIRA: http://issues.opennms.org/browse/NMS-8537
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS844


mvrueden:
If we are going to fix a undocumented feature, we should also add a minor documentation for that feature :)

opennms-config/src/main/java/org/opennms/netmgt/config/DiscoveryConfigFactory.java
```diff
@@ -570,17 +574,20 @@ public long getIntraPacketDelay() {
...
-            protected boolean matches(final IPPollAddress item) {
-                return !isExcluded(item.getAddress());
-            }
-
-        };
+        return StreamSupport.stream(
+                Spliterators.spliteratorUnknownSize(it, ORDERED | IMMUTABLE), false
+            )
+            // Filter out excluded addresses
+            .filter(item -> { return !isExcluded(item.getAddress());})
```
mvrueden: This can even be shorter: `.filter(item -> !isExcluded(item.getAddress()))`

PR #895, opened 2016-07-06
NMS-8474: Run info pannels in single transaction

fooker:
* JIRA: http://issues.opennms.org/browse/NMS-8474
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS840

All info pannel providers are wrapped in a single transaction to ensure
consistent values shown in all pannels.


mvrueden:
The issue also addresses the GraphML Custom Edge Status Provider. Is that call already wrapped?

j-white:
The calls to computeEdgeStatus() appear to be already wrapped in a transaction context.

PR #896, opened 2016-07-06
NMS-8589: Fix Memory Leak and Thread Leak for TopologyUI and NodeMapsApplication

mvrueden:
* JIRA: http://issues.opennms.org/browse/NMS-8589
* BAMBOO: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS845


soleger:
Is it possible to get rid of ```NodeMapsApplication.addRefresher()``` also? It looks like it is only used to trigger the refresh in MapWidgetComponent if I'm reading the code correctly.

mvrueden:
No. The `addRefresher()` is basically telling the client to poll every 5 seconds for changes, whereas the Scheduler is server side, to update the backend, every 5 minutes only.

PR #898, opened 2016-07-06
HZN-834: Add a new "Grid Layout"

j-white:
JIRA: http://issues.opennms.org/browse/HZN-834
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS846


mvrueden:
The issue is targeted for the feature `atlas`. At the moment the Grid Layout would not really work as expected, as the `GraphMLVertex` does not implement `LevelAware`. Is this intentionally?

j-white:
We could implement `LevelAware` if we want to give the user control of the order of the vertices. I'll go ahead and add this.

Ignoring X, and Y is intentional.

PR #899, opened 2016-07-06
Add a new command that removes all of the _other_ vertices from focus.

j-white:
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS847/latest

There's no JIRA issue for this, but I found this particular useful when trying to isolate a vertex when many vertices were already in focus.

### DIFF
--- a/features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/operations/RemoveFocusOtherVerticesOperation.java
+++ b/features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/operations/RemoveFocusOtherVerticesOperation.java
@@ -1,0 +1,109 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2016 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2016 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License,
+ * or (at your option) any later version.
+ *
+ * OpenNMS(R) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with OpenNMS(R).  If not, see:
+ *      http://www.gnu.org/licenses/
+ *
+ * For more information contact:
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.features.topology.app.internal.operations;
+
+import java.util.List;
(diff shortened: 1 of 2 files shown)


features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/operations/RemoveFocusOtherVerticesOperation.java
```diff
@@ -0,0 +1,108 @@
...
+import org.opennms.features.topology.api.support.VertexHopGraphProvider.VertexHopCriteria;
+import org.opennms.features.topology.api.support.VertexHopGraphProvider.WrappedVertexHopCriteria;
+import org.opennms.features.topology.api.topo.Criteria;
+import org.opennms.features.topology.api.topo.VertexRef;
+
+public class RemoveFocusOtherVerticesOperation implements Constants, Operation {
+
+    @Override
+    public void execute(List<VertexRef> targets, final OperationContext operationContext) {
+        if (targets == null || targets.isEmpty()) {
```
mvrueden: This is already checked via `enabled()` and does not need to be checked twice.

features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/operations/RemoveFocusOtherVerticesOperation.java
```diff
@@ -0,0 +1,108 @@
...
+                    shouldRemove = false;
+                    break;
+                }
+            }
+
+            if (shouldRemove) {
+                graphContainer.removeCriteria(eachCriteria);
+            }
+        }
+        graphContainer.redoLayout();
```
mvrueden: We should only invoke `redoLayout` if we actually removed a criteria, otherwise nothing changed.

features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/operations/RemoveFocusOtherVerticesOperation.java
```diff
@@ -0,0 +1,108 @@
...
+import org.opennms.features.topology.api.GraphContainer;
+import org.opennms.features.topology.api.Operation;
+import org.opennms.features.topology.api.OperationContext;
+import org.opennms.features.topology.api.support.VertexHopGraphProvider;
+import org.opennms.features.topology.api.support.VertexHopGraphProvider.VertexHopCriteria;
+import org.opennms.features.topology.api.support.VertexHopGraphProvider.WrappedVertexHopCriteria;
+import org.opennms.features.topology.api.topo.Criteria;
+import org.opennms.features.topology.api.topo.VertexRef;
+
+public class RemoveFocusOtherVerticesOperation implements Constants, Operation {
```
mvrueden: Is there a particular reason for implementing `Constants`?

features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/operations/RemoveFocusOtherVerticesOperation.java
j-white: Good point.

features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/operations/RemoveFocusOtherVerticesOperation.java
j-white: Nope, this was copied from RemoveFocusVerticesOperation. I'll remove it.

PR #900, opened 2016-07-07
Fixed 2 typos in the measurement ReST API doc

StCyr:
Hello,

I don't think this PR deserves a JIRA issue: It just fixes 2 typo in the doc

### DIFF
--- a/opennms-doc/guide-development/src/asciidoc/text/rest/measurements.adoc
+++ b/opennms-doc/guide-development/src/asciidoc/text/rest/measurements.adoc
@@ -16,8 +16,11 @@ The following table shows all supported query string parameters and their defaul
 [options="header"]
 |===
 | name               | default   | comment
-| start              | -14400000 | Timestamp in milliseconds. If < 0, the effective value will be `end - start`
-                                  (i.e.: default value is 4 hours ago).
+| start              | -14400000 | Timestamp in milliseconds.
+
+                                   If > 0, the timestamp is relative to the UNIX epoch (January 1st 1970 00:00:00 AM).
+                                   
+                                   If < 0, the timestamp is relative to the `end` option (i.e.: default value is 4 hours ago).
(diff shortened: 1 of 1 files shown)


opennms-doc/guide-development/src/asciidoc/text/rest/measurements.adoc
```diff
@@ -16,8 +16,9 @@ The following table shows all supported query string parameters and their defaul
 [options="header"]
 |===
 | name               | default   | comment
-| start              | -14400000 | Timestamp in milliseconds. If < 0, the effective value will be (end + start).
-| end                | 0         | Timestamp in milliseconds. If <= 0, the effective value will be the current timestamp.
+| start              | -14400000 | Timestamp in milliseconds. If < 0, the effective value will be `end - start`
```
mvrueden: `end + start` is definitely correct. The reason for this is, that start can be negative. If it were `-` a negative value would result in `end - -start = end + start`. This seems wrong for the default value of `-14400000`.

StCyr: What I understand is:

positive value => start is relative to EPOCH
negative value => start is relative to "end"

the example under the table describes such a behaviour.

Cyr

2016-07-07 9:19 GMT+02:00 mvrueden <notifications@github.com>:

j-white: The other updates you made are fine, but `end - start` is not accurate.

Let's assume that end=0 and start=-14400000 (the defaults), and now=1468273176000.

The effective values are:
end=now=1468273176000
start=end + start=1468273176000 + (-14400000)=1468258776000

StCyr:
Thanks @mvrueden and @j-white for your review.

I totally understand what you mean. 

What do you think about my newest change?

j-white:
Cherry-picked to foundation-2016.

PR #905, opened 2016-07-09
NMS-2656: Ack/Unack/All control on Notification list (plus Angular list of Outages)

soleger:
This branch has a combination of improvements to the prototype Angular list pages:

- Partial fix for NMS-2656: Adds Ack/Unack/All selector to the notification list
- Adds acknowledging support to the notification list
- Notification list has 99% feature parity with the old JSP list (filtering/sorting by IP address doesn't work yet)
- Added RESTv2 service for outages plus an Angular list of outages
- Added nodeId, nodeLabel, foreignSource, and foreignId to XML/JSON serialization of outages
- Added ability to do IS NULL/IS NOT NULL in FIQL searches

To get to the new list pages, visit indexNew.jsp instead of index.jsp.

* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS839


opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/OutageRestService.java
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Basic Web Service using REST for {@link OnmsOutage} entity.
+ *
+ * @author Seth
+ */
+@Component
+@Path("outages")
+@Transactional
+public class OutageRestService extends AbstractDaoRestService<OnmsOutage,Integer> {
```
j-white: I don't think it makes sense to allow outages to be created or modified here. May we should have a read-only version of the AbstractDaoRestService?

soleger: I added an issue to track this improvement:

http://issues.opennms.org/browse/HZN-841

PR #912, opened 2016-07-12
NMS-8200 only try `ulimit -n` if it would raise the number of filehandles

RangerRick:
This patch changes the `$OPENNMS_HOME/bin/install` script to only attempt a `ulimit -n` call to raise the ulimit if the number is higher than the current value.

* JIRA: http://issues.opennms.org/browse/NMS-8200
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS859


soleger:
Why does this PR have 2 commits in it? It looks like the bugfix is just commit c664648c66fb25f09ad4e8ed60237f87af3d365d.

agalue:
I agree with @soleger. It seems like you're accidentally including a fix that is already merged into foundation-2016, unless you're changing the Hardware Inventory Provisioning Adapter here ;)

RangerRick:
Oh, it's because I made the PR against `develop`, rather than `foundation`.  >_<  I guess that commit made it to foundation but hasn't made it to develop (or got fixed/cherry-picked over).

RangerRick:
https://github.com/OpenNMS/opennms/pull/913

PR #913, opened 2016-07-12
NMS-8200 only attempt to set ulimit if it is higher than current

RangerRick:
This patch changes the $OPENNMS_HOME/bin/install script to only attempt a ulimit -n call to raise the ulimit if the number is higher than the current value.

* JIRA: http://issues.opennms.org/browse/NMS-8200
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS859

### DIFF
--- a/opennms-base-assembly/src/main/filtered/bin/install
+++ b/opennms-base-assembly/src/main/filtered/bin/install
@@ -6,11 +6,17 @@ OPENNMS_BINDIR="${install.bin.dir}"
 RUNAS="root"
 
 ULIMIT=`which ulimit 2>/dev/null`
-if [ -n $ULIMIT ]; then
-	for SIZE in 1024 2048 4096 8192 unlimited; do
-		ulimit -n "$SIZE" >/dev/null 2>&1
-		if [ $? -gt 0 ]; then
+if [ -n $ULIMIT ] && [ -x $ULIMIT ]; then
+	for SIZE in 1024 2048 4096 8192 16384 20480; do
+		let CURRENT=`ulimit -n`
+		if [ $CURRENT = "unlimited" ]; then
 			break
+		fi
+		if [ $CURRENT -lt $SIZE ]; then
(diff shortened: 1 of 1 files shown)


opennms-base-assembly/src/main/filtered/bin/install
```diff
@@ -6,12 +6,18 @@ OPENNMS_BINDIR="${install.bin.dir}"
 RUNAS="root"
 
 ULIMIT=`which ulimit 2>/dev/null`
```
indigo423: Just to verify:
```
meridian-sh ~ # cat /etc/redhat-release
CentOS Linux release 7.2.1511 (Core)
meridian-sh ~ # bash -x test.sh
++ which ulimit
+ ULIMIT=
+ echo

meridian-sh ~ # cat test.sh
ULIMIT=`which ulimit 2>/dev/null`
echo $ULIMIT
```

this is empty? you can execute it and it gives 

```
meridian-sh ~ # ulimit
unlimited
```
This is intended?

indigo423:
IMHO, this solves not the problem to require root to set the ulimit. Instead of setting the ulimit, I would print a warning message and let the administrator adjust the limits with operating system mechanisms and remove the requirement running as root.

opennms-base-assembly/src/main/filtered/bin/install
RangerRick: Good catch.  I just changed it to execute `ulimit` instead and then check the return value.

RangerRick:
Instal needs to be run as root because of ownership and upgrade stuff anyways, doesn't it?  Removing root is a much bigger scope and is orthogonal to this issue.

indigo423:
yeah the comment intention was just to remove at least one little part and make it easier for the next one who needs to remove the root requirements from other parts :)

RangerRick:
Considering this is going to touch new meridian 2015/2016 and horizon 18.0.1 I am trying to do as little as possible to the script.  :)

PR #915, opened 2016-07-13
NMS-8597: Make the gwt.openlayers.url and gwt.openlayers.custom.attri…

mvrueden:
This issue makes the tileserver url and the tile layer attribution properties configurable.
In addition the missing configuration for the node maps is added.

Please try out both, the Node Maps and the Remote Poller Maps, as on my local system the Node Maps looks weird (not centered when first visited), but I do not know if it is a new issue or already existed before.

Before testing this, please clear your browser cache.

* JIRA: http://issues.opennms.org/browse/NMS-8597
* BAMBOO: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS862

### DIFF
--- a/features/remote-poller-gwt/src/main/resources/org/opennms/features/poller/remote/gwt/public/app.jsp
+++ b/features/remote-poller-gwt/src/main/resources/org/opennms/features/poller/remote/gwt/public/app.jsp
@@ -31,10 +31,35 @@
 
 <%@page language="java" contentType="text/html; charset=UTF-8" %>
 <%@page import="java.net.URLEncoder"%>
+<%@ page import="org.apache.commons.lang.StringEscapeUtils" %>
+
+<%!
+	public static boolean isValidConfiguration() {
+		return isUrlValid() && isAttributionValid();
+	}
+
+	public static boolean isUrlValid() {
+		return isDefined(System.getProperty("gwt.openlayers.url"));
+	}
+
+	public static boolean isAttributionValid() {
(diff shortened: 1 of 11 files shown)


features/remote-poller-gwt/src/main/resources/org/opennms/features/poller/remote/gwt/public/app.jsp
```diff
@@ -31,11 +31,36 @@
...
+
+	public static boolean isDefined(String input) {
+		return input != null && !"".equals(input);
+	}
+%>
 <%
+
   String mapImplementation = System.getProperty("gwt.maptype", "");
-  String openlayersUrl = System.getProperty("gwt.openlayers.url", "http://otile1.mqcdn.com/tiles/1.0.0/osm/${z}/${x}/${y}.png");
+  String openlayersUrl = System.getProperty("gwt.openlayers.url", "");
```
RangerRick: Shouldn't we make this default to the tiles.opennms.org URL rather than giving broken maps on startup?

RangerRick:
I just tried building with JDK7 (which foundation supports) and it fails because you use java8 features:

```java
        return options.entrySet().stream()
                .map(eachEntry -> new Option(eachEntry.getKey(), eachEntry.getValue()))
                .collect(Collectors.toList());
```

agalue:
Tested on a VM and it works (I can see the maps and the requests against tiles.opennms.org through FireBug).

PR #935, opened 2016-07-20
NMS-8043: Ping from topology view does not work (results in 404 Page not found)

mvrueden:
* JIRA: http://issues.opennms.org/browse/NMS-8043
* BAMBOO: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS880


mvrueden:
Closing as it should be created against `foundation-2016` but is targeted against `release-18.0.1`.

PR #940, opened 2016-07-21
NMS-8616: Add caution section for wrong DNS A Resource Record

indigo423:
Give users a hint to verify DNS A record settings. Wrong A record settings can prevent OpenNMS from start and a exception is shown.
- JIRA: http://issues.opennms.org/browse/NMS-8616
- Bamboo:


j-white:
This is a a good first step, but do we know how to reproduce this? If so, I think we should should focus on trying to fix it or at least improving the error message in the generate exception.

mvrueden:

opennms-bamboo:
Welcome @web-flow.

I could not find a signed OpenNMS Contributor Agreement (OCA) assigned to the username @web-flow.

PR #944, opened 2016-07-25
Fix jmx config generator tests

rssntn67:
On java 8 jmx config geneator tests are failing. At least on my mac.

So with the help of @mvrueden  we have make some changes to fix it


rssntn67:
created issue NMS-8622. Closing because the quickl fix provided does not work.

PR #947, opened 2016-07-27
HZN-852: Created inline Trend charts for displaying database-based statistics

christianpape:
- JIRA: http://issues.opennms.org/browse/HZN-852
- Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS879


indigo423:
Jira issue missing for release notes

indigo423:
👍 for me for the docs.

opennms-config-jaxb/src/main/java/org/opennms/netmgt/config/trend/TrendConfiguration.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.List;
+
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlElement;
+import javax.xml.bind.annotation.XmlRootElement;
+
+@XmlRootElement(name="trend-configuration")
+@XmlAccessorType(XmlAccessType.NONE)
+public class TrendConfiguration implements Serializable {
```
j-white: We should add XML Marshaling/Unmarshaling tests for this object.

opennms-base-assembly/src/main/filtered/etc/trend-configuration.xml
```diff
@@ -0,0 +1,232 @@
...
+            <trend-attribute key="sparkMinSpotColor" value="white"/>
+            <trend-attribute key="sparkMaxSpotColor" value="white"/>
+            <trend-attribute key="sparkSpotRadius" value="3"/>
+            <trend-attribute key="sparkHighlightSpotColor" value="white"/>
+            <trend-attribute key="sparkHighlightLineColor" value="white"/>
+        </trend-attributes>
+        <descriptionLink>outage/list.htm?outtype=current</descriptionLink>
+        <description>${intValue[23]} NODES WITH OUTAGE(S)</description>
+        <query>
+            <![CDATA[
```
j-white: Can we add integration tests to verify the default queries in the configuration file?

Otherwise, these may be forgotten if/when we make a schema change.

agalue:
Just a note about JavaScript dependencies...

Adding third-party dependencies to `opennms-webapp/src/main/webapp/js` will be deprecated once the JavaScript work is merged into `develop`. I recommend to start working on a proper way even if that work is not merged or while it is merged, to have some kind of organization in the meantime.

So, the third-party libraries should be placed like `bower` would do it on `opennms-webapp/src/main/webapp/lib` (think about this directly like a trimmed version of `bower_components`). On `opennms-webapp/src/main/webapp/js`, you should ONLY add proprietary code, like the code you have to implement in JavaScript to make your app work without issues.

Also, it is important to be sure that any `jQuery` related work should be based on 2.2.x, and if `jquery-ui` is required, please use 1.12.0.

Makes sense ?

It was precisely all jquery/jquery-ui and extensions of jquery-ui the hardest thing to make it work with the new structure, and please avoid "custom" bundles, please use whatever is available through `bower` as a reference.

opennms-base-assembly/src/main/filtered/etc/trend-configuration.xml
christianpape: done!

indigo423:
for the reason the background color does not have any functionality, can we use a non-functional color as background for the spark lines? Especially in NOCs people associate the colors to something like "Green" everything as it is "OK". I got several feedbacks where we use  For the reason the trend widgets are just informational they should have a function neutral color which is not something like green, yellow, organge red. We have already some problems like this in the Opsboard and would like to avoid this type of problems at the beginning: http://issues.opennms.org/browse/NMS-7911

christianpape:
Yep, so please create a draft for neutral elements thats fits harmonic in the Horizon/green and Meridian/blue design. Do not focus only on isolated elements - it must also work for the landing page.

opennms-webapp/src/main/webapp/WEB-INF/jsp/trend/trend.jsp
```diff
@@ -0,0 +1,105 @@
...
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+--%>
+
+<%@ page language="java" contentType="text/html" session="true" import="org.opennms.netmgt.config.trend.TrendAttribute"%>
+<%@ taglib uri="http://java.sun.com/jsp/jstl/core" prefix="c" %>
+<%@ taglib uri="http://java.sun.com/jsp/jstl/functions" prefix="fn" %>
+
+<style type="text/css">
```
indigo423: Styling should be defined in SCSS file which uses global styling variables in opennms-webapp/src/main/scss/properties.scss  and not in JSP page.

opennms-webapp/src/main/webapp/WEB-INF/jsp/trend/trend.jsp
```diff
@@ -0,0 +1,105 @@
...
+
+</style>
+
+<div class="alert alert-trend" role="alert">
+    <table cellpadding="0" cellspacing="0" width="100%" border="0">
+        <tr>
+            <td width="1%">
+                <h1 style="margin:0;"><span class="glyphicon ${trendDefinition.icon}" aria-hidden="true"></span></h1>
+            </td>
+            <td style="white-space: nowrap; padding-left:5px; padding-right:5px;">
```
indigo423: Inline styling should use styles from SCSS file using styling variables in opennms-webapp/src/main/scss/properties.scss

opennms-webapp/src/main/webapp/WEB-INF/jsp/trend/trend.jsp
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    .jqstooltip {
+        width: auto !important;
+        height: auto !important;
+    }
+
+</style>
+
+<div class="alert alert-trend" role="alert">
```
indigo423: Inline styling should use styles from SCSS file using styling variables in opennms-webapp/src/main/scss/properties.scss

opennms-webapp/src/main/webapp/trend/single-trend-box.jsp
```diff
@@ -0,0 +1,63 @@
...
+--%>
+
+<%@page language="java"
+        contentType="text/html"
+        session="true"%>
+
+<%@taglib uri="http://java.sun.com/jsp/jstl/core" prefix="c" %>
+<%@taglib uri="http://java.sun.com/jsp/jstl/fmt" prefix="fmt" %>
+
+<style type="text/css">
```
indigo423: Would it make sense to move trend line styling to CSS file?

mvrueden:

j-white:
Is this a work in progress?

christianpape:
For me this is done. Ronny did not liked my styling/coloring. So he changed it and started a discussion on the mailing list. I don't know what the outcome/status is...

j-white:
@indigo423 what are we doing here?

PR #948, opened 2016-07-27
NMS-8631: Add Didactum data collection

mfuhrmann:
* JIRA: http://issues.opennms.org/browse/NMS-8631

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/snmp-graph.properties.d/didactum-graph.properties
+++ b/opennms-base-assembly/src/main/filtered/etc/snmp-graph.properties.d/didactum-graph.properties
@@ -1,0 +1,58 @@
+##############################################################################
+##
+##  Please add report definition in a new line to make it easier
+##  for script based sanity checks
+##
+##################################################
+
+reports=DidactumMonitoringSystem 
+
+######
+###### Reports for Didactum Monitoring System
+######
+
+report.DidactumMonitoringSystem.name=Didactum Monitoring System Internal Analog Sensor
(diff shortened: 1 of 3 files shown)


opennms-base-assembly/src/main/filtered/etc/snmp-graph.properties.d/didactum-graph.properties
```diff
@@ -0,0 +1,58 @@
+##############################################################################
+##
+##  Please add report definition in a new line to make it easier
+##  for script based sanity checks
+##
+##################################################
+
+reports=DidactumMonitoringSystem
```
indigo423: to be more compliant with the current configs, I would suggest to rename to `didactum.monitoringSystem`

opennms-base-assembly/src/main/filtered/etc/snmp-graph.properties.d/didactum-graph.properties
```diff
@@ -0,0 +1,58 @@
...
+##
+##################################################
+
+reports=DidactumMonitoringSystem 
+
+######
+###### Reports for Didactum Monitoring System
+######
+
+report.DidactumMonitoringSystem.name=Didactum Monitoring System Internal Analog Sensor
```
indigo423: * `didactum.monitoringSystem.internal.sensors.analog.name`
* `didactum.monitoringSystem.can.sensors.analog.name`
* `didactum.monitoringSystem.rs.sensors.analog.name`

indigo423:
Moved to https://github.com/opennms-config-modules/didactum-monitoring-system-2

PR #952, opened 2016-07-29
NMS-8624: Make enlinkd location aware

rssntn67:
NMS-8624: Make enlinkd location aware

Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS904-3

### DIFF
--- a/opennms-services/src/main/java/org/opennms/netmgt/enlinkd/NodeDiscoveryBridge.java
+++ b/opennms-services/src/main/java/org/opennms/netmgt/enlinkd/NodeDiscoveryBridge.java
@@ -28,6 +28,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.concurrent.ExecutionException;
 
 import org.opennms.netmgt.enlinkd.snmp.CiscoVtpTracker;
 import org.opennms.netmgt.enlinkd.snmp.CiscoVtpVlanTableTracker;
@@ -42,8 +43,6 @@ import org.opennms.netmgt.model.BridgeMacLink;
 import org.opennms.netmgt.model.BridgeMacLink.BridgeDot1qTpFdbStatus;
 import org.opennms.netmgt.model.BridgeStpLink;
 import org.opennms.netmgt.snmp.SnmpAgentConfig;
-import org.opennms.netmgt.snmp.SnmpUtils;
-import org.opennms.netmgt.snmp.SnmpWalker;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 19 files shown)


opennms-services/src/main/java/org/opennms/netmgt/enlinkd/NodeDiscoveryBridge.java
```diff
@@ -142,79 +141,74 @@ protected void runCollection() {
...
+            final Dot1dBaseTracker dot1dbase = new Dot1dBaseTracker();
+            
+            try {
+                m_linkd.getLocationAwareSnmpClient().walk(getPeer(), dot1dbase).
+                withDescription("dot1dbase").
+                atLocation(getLocation()).
+                execute().
+                get();
+           } catch (ExecutionException e) {
+                    // pass
```
j-white: We should log the exception, and treat it similar to the timeout behavior we had previously.

mvrueden:
Can we actually make it one commit, or distinguish the commit messages.

rssntn67:
Safely Merged after bamboo get green

opennms-services/src/test/java/org/opennms/netmgt/enlinkd/EnLinkdSnmpIT.java
```diff
@@ -173,17 +177,13 @@ public void testCdpGlobalGroupCollection() throws Exception {
...
         try {
-            walker.waitFor();
-            if (walker.timedOut()) {
-                LOG.info("run:Aborting Cdp Linkd node scan : Agent timed out while scanning the {} table", trackerName);
-            }  else if (walker.failed()) {
-                LOG.info("run:Aborting Cdp Linkd node scan : Agent failed while scanning the {} table: {}", trackerName,walker.getErrorMessage());
-            }
+            m_client.walk(config,cdpGlobalGroup)
+            .withDescription(trackerName)
+            .atLocation(null)
```
j-white: Location should not be null here.

opennms-services/src/test/java/org/opennms/netmgt/enlinkd/EnLinkdSnmpIT.java
```diff
@@ -205,17 +205,13 @@ public void testCdpGlobalGroupCollectionWithGlobalIdFormat() throws Exception {
...
```
j-white: Location should not be null here.

opennms-services/src/test/java/org/opennms/netmgt/enlinkd/EnLinkdSnmpIT.java
j-white: Nevermind, I just realized this is a test.

PR #954, opened 2016-07-29
Decrease build time by ignoring jar signing and pdf doc generation by default

mvrueden:
This changes should improve the default build time.
A new profile `run-expensive-tasks` is added, which disables jar signing and the pdf generation.
The profile is enabled by bamboo.

* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS897


pom.xml
```diff
@@ -1036,29 +1036,6 @@
           </dependency>
         </dependencies>
       </plugin>
-      <!-- jacoco code coverage testing library -->
```
soleger: Please reinstate this section of the POM. It is useful if you want to run Jacoco for analyzing code coverage during testing. You could add a profile that enables Jacoco as well.

PR #955, opened 2016-07-29
NMS-8640: Organize and centralize all JavaScript dependencies used in opennms-webapp

agalue:
# DEV-JAM! DEV-JAM! DEV-JAM! DEV-JAM! DEV-JAM! DEV-JAM!

* JIRA: http://issues.opennms.org/browse/NMS-8640
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS894/latest

This looks like a big change, mostly because all the third-party JavaScript/CSS files have been updated and reorganized.

Now, all the actual dependencies should be declared in bower.json.

When compiling opennms-webapp, Maven is going to install nodejs if it is not installed. Then it will install all the nodejs dependencies (like bower and gulp), then it will execute a special gulp task to copy from the bower_components folder only the required JS/CSS/Font/Image files required into src/webapp/lib.

The idea is to only have on src/webapp/lib (which is translated into $OPENNMS_HOME/jetty-webapps/opennms/lib) the required files without all the unnecessary overhead pulled by bower from the external sources.

In some scenarios, the files exposed by each external library/component are not enough and require some tuning. If you take a look at the bottom of the bower.json, you'll see how to instruct gulp which files are required from the bower_components directory for each library.

All the JSP that require JavaScript, as well as all the Vaadin apps that use JavaScript have been tested, and Bamboo went green (so all the smoke-tests are working).

The only pending work is adding documentation for how to develop JavaScript related applications inside OpenNMS, and how to add new dependencies.

Outside the "webapp/lib", the proprietary code (i.e. OpenNMS code) should be placed on "webapp/js".

No JavaScript code should exist outside the above directories.

The only one exception is a library called cropper (which is 10 years old, and I can't find a suitable bower version for it), used to zoom and crop RRD/JRobin images. The JS related files live on webapp/graph/cropper.


j-white:
Looks good to me. A huge improvement over the current setup.

I think we should merge this directly to `develop` though, or wait until `dev-jam-2016` is merged since there are already a lot of changes in that branch.

agalue:
Thanks! and I agree with @j-white.

The problem was that I couldn't create a PR against `develop` because it complained with "not a clean merge", so I did it against the base branch `dev-jam-2016`. Maybe we can try merging `develop` into `dev-jam-2016`, then I can merge `dev-jam-2016` to this branch, and then I can try to create a new PR based on `develop`. Does that sound correct ?

PR #956, opened 2016-08-01
Dev-Jam 2016

j-white:
JIRA: Dev Jam!
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS890-13

Changes made in this branch include:
* Make tests happy in non-US/eastern timezones
* Moved failsafe configuration to the top-level pom
* Cleanup temporary databases after tests, fixes [NMS-7879](http://jira.opennms.org/browse/NMS-7879)


core/daemon/pom.xml
```diff
@@ -14,7 +14,6 @@
   The OpenNMS Daemon launches and acts as a container for OpenNMS Services such as the poller or collector.
   </description>
   <properties>
```
soleger: We can get rid of the entire properties block here (and in other POMs with similar changes).

core/test-api/db/src/main/java/org/opennms/core/test/db/TemporaryDatabasePostgreSQL.java
```diff
@@ -117,19 +129,35 @@ public TemporaryDatabasePostgreSQL(String testDatabase, String driver, String ur
...
+            throw new TemporaryDatabaseException("The '" + OPENNMS_UNIT_TEST_PROPERTY + "' property is set to true, " +
+                    "however this class is only suitable for integration tests, not unit tests. " +
+                    "Please refactor to not use a database or move to an integration test (name your test class *IT.java). " +
+                    "See http://wiki.opennms.org/wiki/Test_conventions for details");
+        }
+    }
+
+    protected static String getDatabaseName(Object hashMe) {
+        // Append the current object's hashcode to make this value truly unique
+        return String.format("opennms_test_%s_%06d_%s", System.currentTimeMillis(), System.nanoTime(), Math.abs(hashMe.hashCode()));
```
soleger: I don't think we need ```System.currentTimeMillis()``` here, ```System.nanoTime()``` should be sufficient by itself to guarantee a unique value. Is there any reason to include it?

opennms-config/src/main/java/org/opennms/netmgt/config/SnmpPeerFactory.java
```diff
@@ -77,83 +77,80 @@
...
     /**
-     * The config class loaded from the config file
+     * This member is set to true if the configuration file has been loaded.
      */
-    private SnmpConfig m_config;
+    private static volatile boolean s_loaded = false;
 
-    private static FileReloadContainer<SnmpConfig> m_container;
-    private static FileReloadCallback<SnmpConfig> m_callback;
+    private final ReadWriteUpdateLock m_globalLock = new ReentrantReadWriteUpdateLock();
```
soleger: This change doesn't look correct. We're using the ReadWriteUpdateLock to protect access to the singleton so I think this lock needs to remain static. Can you explain any of the changes in this class?

core/daemon/pom.xml
soleger: I'll write a quick ed script for this.

core/test-api/db/src/main/java/org/opennms/core/test/db/TemporaryDatabasePostgreSQL.java
deejgregor: System.currentTimeMillis() provides a wall clock reference for coordinating with external entities (test logs, etc.) and System.nanoTime() to ensure uniqueness. I could move the System.currentTimeMillis() to the blame table.

soleger: OK, that makes sense. Leave it where it is.

opennms-config/src/main/java/org/opennms/netmgt/config/SnmpPeerFactory.java
deejgregor: SnmpPeerFactory was causing test failures in unrelated downstream tests because state was being held in the m_container/m_callback static fields pointing at stale configuration data.

I moved as much state as I could out of static fields and into the object instance, keeping only the singleton and loaded fields as static like we have in most of the other configuration factories. I left configFile a singleton because I wasn't exactly sure how it was used and didn't want to find out at this time.

The read/write locks are still there protecting read and writes, but they're on the object instance. setInstance/setFile are the two methods that can cause the class in the singleton replaced, and by my searching, they're only called in tests.

soleger:
This branch now contains Alejandro's JS changes. Once it goes green in Bamboo, we'll merge everything to develop.

PR #957, opened 2016-08-01
hzn-838: Add smoke tests for provision on minion

cgorantla:
* JIRA: http://issues.opennms.org/browse/hzn-838
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS867-SMOKE-14

### DIFF
--- a/smoke-test/src/test/java/org/opennms/smoketest/minion/DetectorsCommandTest.java
+++ b/smoke-test/src/test/java/org/opennms/smoketest/minion/DetectorsCommandTest.java
@@ -1,0 +1,158 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ * <p>
+ * Copyright (C) 2016 The OpenNMS Group, Inc. OpenNMS(R) is Copyright (C)
+ * 1999-2016 The OpenNMS Group, Inc.
+ * <p>
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ * <p>
+ * OpenNMS(R) is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU Affero General Public License as published by the Free
+ * Software Foundation, either version 3 of the License, or (at your option)
+ * any later version.
+ * <p>
+ * OpenNMS(R) is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License
+ * for more details.
+ * <p>
+ * You should have received a copy of the GNU Affero General Public License
+ * along with OpenNMS(R). If not, see: http://www.gnu.org/licenses/
+ * <p>
+ * For more information contact: OpenNMS(R) Licensing
+ * <license@opennms.org> http://www.opennms.org/ http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.smoketest.minion;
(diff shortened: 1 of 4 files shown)


smoke-test/src/test/java/org/opennms/smoketest/minion/DetectorsCommandTest.java
```diff
@@ -0,0 +1,158 @@
...
+import org.opennms.test.system.api.TestEnvironment;
+import org.opennms.test.system.api.TestEnvironmentBuilder;
+
+public class DetectorsCommandTest {
+
+    private static TestEnvironment m_testEnvironment;
+
+    private static final Logger LOG = LoggerFactory.getLogger(DetectorsCommandTest.class);
+
+    private static final String DETECTORS = "BGP_Session: org.opennms.netmgt.provision.detector.snmp.BgpSessionDetector\n"
```
j-white: Can we make this a Map instead? i.e.:

```java
Map<String, String> expectedDetectors = ImmutableMap.of(
    "BGP_Session", "org.opennms.netmgt.provision.detector.snmp.BgpSessionDetector"
    );
```

smoke-test/src/test/java/org/opennms/smoketest/minion/DetectorsCommandTest.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    @Test
+    public void LoadDetectorsOnMinionTest() throws Exception {
+
+        final InetSocketAddress sshAddr = m_testEnvironment.getServiceAddress(ContainerAlias.MINION, 8201);
+        try (final SshClient sshClient = new SshClient(sshAddr, "admin", "admin");) {
+            PrintStream pipe = sshClient.openShell();
+            pipe.println("provision:list-detectors");
+            Thread.sleep(2000);
```
j-white: We should grab the output using the same pattern used here: https://github.com/OpenNMS/opennms/blob/54c153ed33867efcf8bb9eacbad75345246f02f7/smoke-test/src/test/java/org/opennms/smoketest/minion/SyslogTest.java#L122

This avoids requiring the explicit sleep.

smoke-test/src/test/java/org/opennms/smoketest/minion/DetectorsCommandTest.java
```diff
@@ -0,0 +1,158 @@
...
+
+    @Before
+    public void checkForDocker() {
+        Assume.assumeTrue(OpenNMSSeleniumTestCase.isDockerEnabled());
+    }
+
+    @Test
+    public void LoadDetectorsOnMinionTest() throws Exception {
+
+        final InetSocketAddress sshAddr = m_testEnvironment.getServiceAddress(ContainerAlias.MINION, 8201);
```
j-white: We should run the same test in the OpenNMS shell.

smoke-test/src/test/java/org/opennms/smoketest/minion/DetectorsCommandTest.java
```diff
@@ -0,0 +1,158 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ * <p>
```
j-white: License header shouldn't contain <p> tags. See https://github.com/OpenNMS/opennms/blob/54c153ed33867efcf8bb9eacbad75345246f02f7/smoke-test/src/test/java/org/opennms/smoketest/minion/SyslogTest.java#L1 for a good example.

PR #961, opened 2016-08-03
NMS-8651: Modified uei.opennms.org/nodes/pathOutage event definition

StCyr:
Modified uei.opennms.org/nodes/pathOutage event definition to reduce on critical path rather than nodes with unknown status behind it.

* JIRA: http://issues.opennms.org/browse/NMS-8651


StCyr:
damned wrong branch

PR #970, opened 2016-08-09
NMS-8641: Replace GWT with Angular: IP/SNMP Interfaces panel on the node page

agalue:
Replace GWT with Angular: IP/SNMP Interfaces panel on the node page.

* JIRA: http://issues.opennms.org/browse/NMS-8641
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS909

Please let me know if there is something else that might be useful to add or remove.


j-white:
Looks good and works well.

Should we add a simple smoke test for this component?

agalue:
A smoke test will be a nice feature, but unfortunately, not simple.

There are no smoke tests for the node page at all (and of course no smoke tests for the old GWT implementation which is why all the tests are passing), and we need real data in order to see information rendered on the node page and do some validations.

I'll do some tests to see if we can have something working.

agalue:
I figured out a way to test it.

Because it is hard to guarantee that the smoke tests will have access to a node with SNMP capabilities to perform an actual discover/detection, I ended up using the Node's ReST end point to manually create a node, the IP interfaces and the SNMP interfaces from a JUnit/Smoke test. Then, I should have enough information available to perform the test. I'm working on it, an update on this PR will be available soon.

agalue:
As this one was tested and it has a smoke test to validate the rendering and the content, I think it is ready to be merged.

j-white:
Thanks for the update. 

Sorry I didn't merge it soon, I didn't realize the tests were already implemented.

agalue:
No worries, and thanks!!

PR #971, opened 2016-08-09
NMS-8646: Replace GWT with Angular: choose resources page

agalue:
Replace GWT with Angular: choose resources page

* JIRA: http://issues.opennms.org/browse/NMS-8646
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS911

Please let me know if there is something else that might be useful to add or remove.


agalue:
The UI will sort the resource types, and also the resources per resource type alphabetically. That means, the Node Level Resources and Interfaces will appear at the end, not at the beginning. I can change the sorting logic to mimic the current behavior which is: force SNMP Node Level to be the first, SNMP Interfaces to be the second, and then the rest of the resource types ordered.

Thoughts ?

agalue:
A JUnit/Smoke Test has been added to verify that the new link from the node page works, and also the choose resources page is rendered.

Because it is hard to have real JRB/RRD available on smoke tests, the code only verifies that the page is rendered, and the banner shown when there are no resources is displayed.

I've merged develop into this branch.

j-white:
Thanks for adding the tests.

Testing resource related pages using smoke tests should be easier once we add support for push metrics.

PR #972, opened 2016-08-10
HZN-613: Make Discovery consider locations other than localhost

j-white:
JIRA: http://issues.opennms.org/browse/HZN-613
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS924


mvrueden:
In HZN-613 it is mentioned that methods `Provisioner.doAddNode()` and `Provisioner.handleAddNode()` are of interest. These seem to be left-overs of Cupsd, which does not exist anymore. I did not remove them while hacking the "Discovery Location Awareness". However I think we should rip those methods out.

PR #974, opened 2016-08-11
NMS-8644: Replace GWT with Angular: add to KSC report on the resource graphs page.

agalue:
Replace GWT with Angular: add to KSC report on the resource graphs page.

* JIRA: http://issues.opennms.org/browse/NMS-8644
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS912

Instead of using a popover window shown right after the button, I've used a modal window with more options and some validation (to avoid adding the same graph twice to a KSC report).


agalue:
I've merged develop into the branch, and also added a karma test to verify the controllers, as it is not possible to implement a smoke test due to the fact that we don't have real data available.

PR #975, opened 2016-08-11
NMS-8642/NMS-8643: Replace GWT with Angular: search box for resources and KSC reports.

agalue:
As the Angular solution is very small, and this is only used on the front page, I decided to put both components together on the same Angular application.

NMS-8642: Replace GWT with Angular: search box for resources page on front page.
NMS-8643: Replace GWT with Angular: search box for KSC reports on front page.

* JIRA: http://issues.opennms.org/browse/NMS-8642
* JIRA: http://issues.opennms.org/browse/NMS-8643
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS925

*IMPORTANT:*

* https://github.com/OpenNMS/opennms/pull/971 has to be merged first, as the target page for the resources page is the new one based on Angular, not the old one based on GWT.
* As discussed, the typeahead components won't show paginated results, as those components are intended to be used when you partially know the name of the node to show the resources, or the name of the KSC report to show it.


agalue:
A JUnit/Smoke Test has been added to verify that the search boxes are rendered on the index page.

I've merged develop into this branch to fix issues with CoreWeb.gwt.xml (all the GWT-to-angular apps will be modifying this file).

PR #977, opened 2016-08-11
HZN-861: Update the queue name format to support multliple OpenNMS instances

j-white:
JIRA: http://issues.opennms.org/browse/HZN-861
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS928


core/lib/src/main/java/org/opennms/core/utils/SystemInfoUtils.java
```diff
@@ -40,12 +40,18 @@
 public class SystemInfoUtils {
     private static final Logger LOG = LoggerFactory.getLogger(SystemInfoUtils.class);
 
+    public static final String OPENNMS_INSTANCE_ID_SYS_PROP = "org.opennms.instance.id";
+    public static final String DEFAULT_INSTANCE_ID = "OpenNMS";
```
soleger: Long-term, this part of the model should be merged into the OnmsMonitoringSystem that represents the OpenNMS system.

PR #982, opened 2016-08-18
NMS-8678: Upgrade Jetty to 9.3.11.v20160721

j-white:
JIRA: http://issues.opennms.org/browse/NMS-8678
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS889-7


soleger:
Looks great! Finally, a servlet container that came out after 2012. 👍

agalue:
Looks very good!

I'm curious about something: does the jetty.xml file from the examples directory requires an update ? If so, guess we might need an upgrade task to fix the existing copy on /opt/opennms/etc

j-white:
jetty.xml in the examples directory is a copy of `opennms-jetty/src/main/resources/org/opennms/netmgt/jetty/jetty.xml`, so yes that has changed.

Any customization made via system properties will continue to work as it did before the upgrade. However, local modifications to jetty.xml will have to be rewritten.

What tasks should the upgrade script perform? I don't think we'll be able to handle automatically patching any custom jetty.xml files.

agalue:
No problem. If the kind of changes cannot be automated, then the user should do it manually. Now, I suggest to put a big note on the release notes to let the user know that they might need to tweak jetty.xml after the upgrade.

pbrane:
> On Aug 19, 2016, at 7:41 AM, Alejandro Galue <notifications@github.com> wrote:
> 
> No problem. If the kind of changes cannot be automated, then the user should do it manually. Now, I suggest to put a big note on the release notes to let the user know that they might need to tweak jetty.xml after the upgrade.
> 
Will they get an rpm.new file?

:David

agalue:
> On Aug 19, 2016, at 9:41 AM, p-brane <notifications@github.com> wrote:
> 
> Will they get an rpm.new file?

I'd expect a .rpmnew on the examples directory, but not on /opt/opennms/etc because it doesn't exist by default. The user will have to fix it manually due to the API/Library upgrade if I understood correctly.

PR #984, opened 2016-08-20
NMS-8617 ICMP Enhancements

RangerRick:
This patchset adds support for 3 major features in the ICMP subsystem:

1. The ability to run as non-root on Linux and Mac OS X
2. The ability to set the DSCP traffic control bits on outgoing ICMP packets
3. The ability to set the "Don't Fragment" bit on outgoing ICMP packets

* JIRA: http://issues.opennms.org/browse/NMS-8617
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS935


agalue:
Does this mean we can update the install and upgrade script to avoid require being root to execute them ?

RangerRick:
We will have to do some more work to make installation/etc. correct for non-root.  We've had upgrade scripts touch RRDs and metadata and such before, and configs that might have been written out as root.  Also we'd need to decide how to deal with being unable to listen for traps as non-root (AFAIK, these non-root features only apply to ICMP, not any old UDP port). But it's at least a first step.

I wouldn't be comfortable making it a "feature" of 19, but maybe for 20.  It does, at least, mean the remote poller can run as non-root.

container/features/src/main/resources/features.xml
```diff
@@ -357,12 +357,12 @@
       <bundle>wrap:mvn:org.springframework.webflow/spring-webflow/${springWebFlowVersion}</bundle>
     </feature>
 
-    <feature name="jicmp" description="jicmp" version="1.4.4">
-      <bundle>wrap:mvn:org.opennms/jicmp-api/1.4.4$Export-Package=org.opennms.protocols.icmp,org.opennms.protocols.ip</bundle>
+    <feature name="jicmp" description="jicmp" version="2.0.0.SNAPSHOT">
```
j-white: Should be pointing to 2.0.0 instead?

container/features/src/main/resources/features.xml
```diff
@@ -357,12 +357,13 @@
       <bundle>wrap:mvn:org.springframework.webflow/spring-webflow/${springWebFlowVersion}</bundle>
     </feature>
 
-    <feature name="jicmp" description="jicmp" version="1.4.4">
-      <bundle>wrap:mvn:org.opennms/jicmp-api/1.4.4$Export-Package=org.opennms.protocols.icmp,org.opennms.protocols.ip</bundle>
+    <feature name="jicmp" description="jicmp" version="2.0.0">
+      <feature>guava</feature>
+      <bundle>wrap:mvn:org.opennms/jicmp-api/2.0.0$Export-Package=org.opennms.protocols.icmp,org.opennms.protocols.ip</bundle>
```
soleger: Can we change these JARs into bundle artifacts so that we don't have to use the ```wrap:``` protocol any more? That would be nice.

features/minion/repository/pom.xml
```diff
@@ -218,12 +218,15 @@
 
     <repositories>
         <repository>
-            <snapshots>
-                <enabled>false</enabled>
-            </snapshots>
-            <releases>
-                <enabled>true</enabled>
-            </releases>
+            <snapshots><enabled>true</enabled></snapshots>
```
soleger: We can probably get rid of these changes now that jicmp has been released at version 2.0.0.

opennms-icmp/opennms-icmp-jni/pom.xml
```diff
@@ -58,6 +58,13 @@
 
   <repositories>
     <repository>
+      <snapshots><enabled>true</enabled></snapshots>
```
soleger: Likewise, I think we can get rid of this now that jicmp has been released.

opennms-icmp/opennms-icmp-jni6/pom.xml
```diff
@@ -58,6 +58,13 @@
 
   <repositories>
     <repository>
+      <snapshots><enabled>true</enabled></snapshots>
```
soleger: Ditto. :)

PR #985, opened 2016-08-24
HZN-857: Change Trapd's SNMPv3 REST operations to use RestClient

MS043660:
Added features for handling REST calls through Restclient.
Change log : 
- features.xml
- blueprint-trapd-listener.xml
- ScvEnabledRestClientImpl.java


features/events/traps/blueprint-trapd-listener.xml
```diff
@@ -43,6 +56,23 @@
...
+
+    <!-- A canonical source for this Minion's identity -->
+    <bean id="minionIdentity" class="org.opennms.minion.core.impl.MinionIdentityImpl">
+       <argument value="[[location]]"/>
+       <argument value="[[id]]"/>
+    </bean>
+    <service ref="minionIdentity" interface="org.opennms.minion.core.api.MinionIdentity"/>
+
+    <!-- REST Client used to communicate with the OpenNMS controller -->
+    <bean id="restClient" class="org.opennms.minion.core.impl.ScvEnabledRestClientImpl">
```
j-white: Is there any specific reason why we need to redefine this bean instead of pulling it from the service registry?

MS043660: This i added since i didnt find the service when i installed the blueprint.Now am re working on the JIRA and will remove this part of code.

container/features/src/main/resources/features.xml
```diff
@@ -826,10 +826,19 @@
     <feature name="opennms-trapd-listener" description="OpenNMS :: Trapd :: Listener" version="${project.version}">
       <feature>camel-blueprint</feature>
       <feature>opennms-trapd</feature>
-
+       <bundle>mvn:org.opennms.features.minion/core-impl/${project.version}</bundle>
```
j-white: We should depend on the "minion-core-api" feature instead of the core-impl bundle and the scv feature.

container/features/src/main/resources/features.xml
```diff
@@ -826,10 +826,19 @@
...
+	  <feature>scv</feature>
       <bundle>blueprint:mvn:org.opennms.features.events/org.opennms.features.events.traps/${project.version}/xml/blueprint-trapd-listener</bundle>
     </feature>
 
+	<feature name="scv" description="Secure Credentials Vault"
```
j-white: The 'scv' feature is already defined elsewhere.

features/events/traps/pom.xml
```diff
@@ -274,5 +274,15 @@
         </exclusion>
       </exclusions>
     </dependency>
+      <dependency>
+            <groupId>org.opennms.features.minion</groupId>
+            <artifactId>core-impl</artifactId>
```
j-white: We should only need to depend on core-api.

features/minion/core/impl/src/main/java/org/opennms/minion/core/impl/ScvEnabledRestClientImpl.java
```diff
@@ -68,31 +70,47 @@ public ScvEnabledRestClientImpl(String url, SecureCredentialsVault scv, String s
...
+			authCache.put(target, basicAuth);
+
+			HttpClientContext localContext = HttpClientContext.create();
+			localContext.setAuthCache(authCache);
+
+			response = httpclient.execute(target, httpget, localContext);
+			return response;
+	}
+    
+    @Override
```
j-white: Inconsistent formatting.

MS043660: I use Arsdigita which was mentioned..I will recheck again.

features/events/traps/pom.xml
MS043660: Yes i have removed it already and am about to push that part of code.

container/features/src/main/resources/features.xml
MS043660: But strange thing is minion-core,scv are not visible even after installing minion-core-api

container/features/src/main/resources/features.xml
MS043660: Yes ,this is also removed already.

j-white:
Thanks for the updates @MS043660, this looks much better and simpler now.

MS043660:
@j-white : But doubt is Bamboo failing and no where its related to changes done !!!!!

soleger:
I just merged develop into your branch, let's see if Bamboo is any better. If not, it is probably due to OSGi dependency changes that were introduced by your changes and this will need to be resolved before the branch can be merged.

MS043660:
Actually all OSGI changes are reverted,now it's just plain ,I don't think this will affect smoke tests,But still let's wait for Bamboo results and thank you .

mvrueden:

PR #986, opened 2016-08-24
NMS-8645: Replace GWT with Angular: assets page

agalue:
Replace GWT with Angular: assets page

* JIRA: http://issues.opennms.org/browse/NMS-8645
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS939

The Angular version looks exactly the same, there is a smoke test to validate the rendering.
Besides updating the asset fields, it could update the requisition if the nodes has foreignSource and foreignID fields if the user wants to do it.

The main difference is that all the templates and the controller are dynamic. There is a JSON file that holds how the layout will be rendered, and which fields will be used on each panel. That way, adding new asset fields to the database will be extremely easy now.


agalue:
**IMPORTANT** The smoke tests were failing because a change was required on the smoke-tests project in order to send PUT requests. I just pushed the change, so Bamboo should be OK soon.

agalue:
Finally Bamboo is green, and this is ready to be merged. Objections ?

PR #989, opened 2016-08-30
HZN-865: SNMP4J method that will convert PDU into byte[]

soleger:
@DP044946 Here is the method to convert a PDU into a byte[]. I'll merge this to develop and then we can update your trap serialization branch with the changes.

* JIRA: http://issues.opennms.org/browse/HZN-865
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS949


core/snmp/integration-tests/src/test/java/org/opennms/netmgt/snmp/snmp4j/Snmp4JDummyTransportTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+	@Before
+	public void setupSnmp4jLogging() {
+		MockLogAppender.setupLogging(true, "DEBUG");
+	}
+
+	@Test
+	public void testDummyTransport() throws Exception {
+		PDU pdu = makePdu();
+		LOG.debug(SnmpUtils.getHexString(Snmp4JUtils.convertPduToBytes(InetAddressUtils.ONE_TWENTY_SEVEN, 162, "hello", pdu)));
```
j-white: Should we assert something here?

PR #993, opened 2016-08-31
NMS-8647: Replace GWT with Angular: choose resources (part of KSC editor wizard)

agalue:
Replace GWT with Angular: choose resources (part of KSC editor wizard)

* JIRA: http://issues.opennms.org/browse/NMS-8647
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS944

I will appreciate if someone can play with the KSC reports to be sure they work as expected. I haven't found issues on my testing VM but I just want to make sure everything is OK.

The default page for reports, and the default page for "Reports -> Resource Reports" was updated as well.

As you can see, there are LOTS of files that have been removed as those are not required anymore.


agalue:
Because there is no way to make smoke-tests for this as we cannot have real data, I'm going to make a simple Karma/Jasmine to test the very basic functionality of the Angular controllers, prior merging this PR.

PR #996, opened 2016-09-02
NMS-8527: Find a faster way of determining root-cause or impacting vertices

j-white:
JIRA: http://issues.opennms.org/browse/NMS-8527
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS951


j-white:
Ping @christianpape and @fooker. Could you please lease review this, since you are most familiar with the BSM code.

features/bsm/service/impl/src/main/java/org/opennms/netmgt/bsm/service/internal/DefaultBusinessServiceStateMachine.java
```diff
@@ -213,20 +226,22 @@ private void reduceUpdateAndPropagateVertex(BusinessServiceGraph graph, GraphVer
...
 
         // Multiply the statuses based on their relative weight
-        List<Status> statuses = Lists.newArrayList();
+        List<StatusWithIndex> statuses = Lists.newArrayList();
+        int k = 0;
         for (Entry<GraphEdge, Status> entry : edgesWithStatus.entrySet()) {
             int relativeWeight = Math.floorDiv(entry.getKey().getWeight(), gcd);
             for (int i = 0; i < relativeWeight; i++) {
-                statuses.add(entry.getValue());
+                statuses.add(new StatusWithIndex(entry.getValue(), k));
```
fooker: This looks fishy as it generates the index by the position of an element in a set.

Is there any mechanism ensuring the set does not change while this graph is active?

j-white: We hold a read lock on the graph when calculating the root cause or impact.

fooker: OK

PR #997, opened 2016-09-02
NMS-8406: Don't override existing system properties when initializing the webapp.

j-white:
JIRA: http://issues.opennms.org/browse/NMS-8406
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS954/latest


soleger:
Should we also have this fix in the foundation branches?

j-white:
This fix is only needed for properties set via *opennms.properties.d*, which was added in 18.0.0, so back-porting it shouldn't be necessary.

PR #998, opened 2016-09-02
NMS-8709: Add null check for JMX agent

soleger:
* JIRA: http://issues.opennms.org/browse/NMS-8709
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS956


j-white:
Oops. Look like this one snuck in when we refactored the detectors for Minion.

Looks good!

PR #1001, opened 2016-09-06
Partial fix for HZN-776: Minion JMX metrics

soleger:
Partial JMX metrics for the Minion modules. These changes provide graphs for syslog and trap Camel routes.

TODO: Follow the same pattern for the RPC Camel routes.
TODO: Create graphs for message processing times.

* JIRA: http://issues.opennms.org/browse/HZN-776
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-DEL56


opennms-base-assembly/src/main/filtered/etc/jmx-datacollection-config.d/kafka.xml
```diff
@@ -0,0 +1,38 @@
...
+    <jmx-collection name="jmx-kafka">
+        <rrd step="300">
+            <rra>RRA:AVERAGE:0.5:1:2016</rra>
+            <rra>RRA:AVERAGE:0.5:12:1488</rra>
+            <rra>RRA:AVERAGE:0.5:288:366</rra>
+            <rra>RRA:MAX:0.5:288:366</rra>
+            <rra>RRA:MIN:0.5:288:366</rra>
+        </rrd>
+        <mbeans>
+            <mbean name="JVM Memory" objectname="java.lang:type=OperatingSystem">
```
j-white: You could also use the '<import-mbeans>' tag to avoid having to redefine these in every collection. See https://github.com/OpenNMS/opennms/blob/develop/opennms-base-assembly/src/main/filtered/etc/examples/jvm-datacollection/jmx-datacollection-config.d/activemq.xml#L11 for an example.

opennms-base-assembly/src/main/filtered/etc/collectd-configuration.xml
```diff
@@ -137,16 +153,18 @@
...
     <collector service="WS-Man" class-name="org.opennms.netmgt.collectd.WsManCollector"/>
-    <collector service="OpenNMS-JVM" class-name="org.opennms.netmgt.collectd.Jsr160Collector"/>
-    <collector service="Minion-JVM" class-name="org.opennms.netmgt.collectd.Jsr160Collector"/>
     <collector service="VMware-VirtualMachine" class-name="org.opennms.netmgt.collectd.VmwareCollector"/>
     <collector service="VMware-HostSystem" class-name="org.opennms.netmgt.collectd.VmwareCollector"/>
     <collector service="VMwareCim-HostSystem" class-name="org.opennms.netmgt.collectd.VmwareCimCollector"/>
-    <collector service="PostgreSQL" class-name="org.opennms.netmgt.collectd.JdbcCollector"/>
-</collectd-configuration>
 
+    <!-- TODO: Rename to 'JMX-OpenNMS' -->
```
j-white: TODOs in the default configs?

opennms-base-assembly/src/main/filtered/etc/jmx-datacollection-config.d/kafka.xml
soleger: I'll take a look at this because I would definitely like to consolidate the standard JVM MBean definitions so that they can be included inside JMX-OpenNMS, JMX-Minion, JMX-Kafka, JMX-JBoss, etc.

opennms-base-assembly/src/main/filtered/etc/collectd-configuration.xml
soleger: I'm going to leave this here for now because I think (for consistency's sake) that we should address this in 19 but it will require an upgrade script that can update the database.

PR #1004, opened 2016-09-07
HZN-776: Add JMX collections and graphs for remaining Minion metrics

soleger:
This PR contains work to wrap up the Minion data collections. It includes processing time metrics for syslog and trap metrics and will also contain route metrics for the RPC services when it is complete.

* JIRA: http://issues.opennms.org/browse/HZN-776
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-DEL56

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/snmp-graph.properties.d/opennms-minion-graph.properties
+++ b/opennms-base-assembly/src/main/filtered/etc/snmp-graph.properties.d/opennms-minion-graph.properties
@@ -1,4 +1,8 @@
 reports=\
+OpenNMS.Minion.RPC.Server.Detect.Exchanges, \
+OpenNMS.Minion.RPC.Server.Detect.ProcessingTime, \
+OpenNMS.Minion.RPC.Server.SNMP.Exchanges, \
+OpenNMS.Minion.RPC.Server.SNMP.ProcessingTime, \
 OpenNMS.Minion.Syslogd.Listener.Exchanges, \
 OpenNMS.Minion.Syslogd.Marshaller.Exchanges, \
 OpenNMS.Minion.Syslogd.JMS.Exchanges, \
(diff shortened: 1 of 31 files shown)


opennms-base-assembly/src/main/filtered/etc/snmp-graph.properties.d/opennms-minion-graph.properties
```diff
@@ -116,8 +112,65 @@ report.OpenNMS.Minion.Syslogd.Kafka.Exchanges.command=--title="Syslog Messages S
...
+ GPRINT:listLastSec:MIN:" Min \\: %8.2lf %s" \
+ GPRINT:listLastSec:MAX:" Max \\: %8.2lf %s\\n" \
+ STACK:marLastSec#729fcf:"Process       " \
+ GPRINT:marLastSec:AVERAGE:" Avg \\: %8.2lf %s" \
+ GPRINT:marLastSec:MIN:" Min \\: %8.2lf %s" \
+ GPRINT:marLastSec:MAX:" Max \\: %8.2lf %s\\n" \
+ STACK:mqLastSec#73d216:"Send via JMS  " \
+ GPRINT:mqLastSec:AVERAGE:" Avg \\: %8.2lf %s" \
+ GPRINT:mqLastSec:MIN:" Min \\: %8.2lf %s" \
+ GPRINT:mqLastSec:MAX:" Max \\: %8.2lf %s\\n" \
```
indigo423: Just out of curiosity: Is the last `\` necessary? See https://github.com/OpenNMS/opennms/blob/develop/opennms-base-assembly/src/main/filtered/etc/snmp-graph.properties.d/mib2-graph.properties#L82

soleger: No, and it does unexpectedly extend the line continuation. However, it makes it easier to cut-and-paste the lines to make additional line/area defs. You just have to make sure to always put a blank line afterwards.

I'll go and clean these up because after looking at all of the graph configs, only about 2% have this trailing slash. I'll extend the unit tests to detect this so that we're consistent.

soleger:
OK, this PR is ready to merge once Bamboo runs. It has message counter and processing time graphs for:

- Syslogs
- Traps
- Detector RPC
- SNMP RPC

Adding DNS, Poller, and Discovery RPC should be trivial when they are available but we can do this in later PRs.

PR #1005, opened 2016-09-08
NMS-8688: Create a node representing each Minion for monitoring

christianpape:
* JIRA: http://issues.opennms.org/browse/NMS-8688
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS946

### DIFF
--- a/opennms-services/src/main/java/org/opennms/netmgt/poller/monitors/MinionHeartbeatMonitor.java
+++ b/opennms-services/src/main/java/org/opennms/netmgt/poller/monitors/MinionHeartbeatMonitor.java
@@ -1,0 +1,74 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2016 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2016 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License,
+ * or (at your option) any later version.
+ *
+ * OpenNMS(R) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with OpenNMS(R).  If not, see:
+ *      http://www.gnu.org/licenses/
+ *
+ * For more information contact:
+ *     OpenNMS(R) Licensing <license@opennms.org>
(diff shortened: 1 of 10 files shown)


opennms-services/src/main/java/org/opennms/netmgt/poller/monitors/MinionHeartbeatMonitor.java
```diff
@@ -0,0 +1,74 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.Map;
+import com.google.common.base.Supplier;
+
+public class MinionHeartbeatMonitor extends AbstractServiceMonitor {
+
+    private static final Logger LOG = LoggerFactory.getLogger(MinionHeartbeatMonitor.class);
+
+    private final Supplier<NodeDao> nodeDao = Suppliers.memoize(() -> BeanUtils.getBean("daoContext", "nodeDao", NodeDao.class));
```
soleger: Can you guys go through the codebase and update our code to use this everywhere?? :)

pbrane: > On Sep 8, 2016, at 11:25 AM, Seth Leger <notifications@github.com> wrote:
> 
> Can you guys go through the codebase and update our code to use this everywhere?? :
> 

Shouldn’t we be using java.util.function.Supplier instead?

:D

j-white:
Is there any specific reason we create a FS per location, instead of adding all Minions to a single FS?

christianpape:
You can put all minions in one FS by just removing the %s in the opennms.minion.provisioning.foreignSourcePattern.

j-white:
Gotcha. Are there any objections to using a single FS by default?

opennms-services/src/main/java/org/opennms/netmgt/poller/monitors/MinionHeartbeatMonitor.java
fooker: AFAIK, there is no `Suppliers.memoize` in java 8. Sorry about that. :-(

christianpape:
Default pattern changed to use the foreign source "Minions" without any location.

smoke-test/src/test/java/org/opennms/smoketest/minion/MinionHeartBeatTest.java
```diff
@@ -81,12 +87,24 @@ public void minionHeartBeatTestForLastUpdated() {
...
+		await().atMost(900, SECONDS)
+			   .pollInterval(5, SECONDS)
+			   .until(DaoUtils.countMatchingCallable(minionDao,
+													 new CriteriaBuilder(OnmsMinion.class).ge("lastUpdated", startOfTest).toCriteria()),
+					  greaterThan(0));
+
+		await().atMost(90, SECONDS)
+			   .pollInterval(5, SECONDS)
+			   .until(DaoUtils.countMatchingCallable(nodeDao,
+													 new CriteriaBuilder(OnmsNode.class).eq("foreignSource", "Minions@MINION")
```
j-white: I'm guessing this will need to changed too, now that the default format has changed.

j-white:
Looks good. Waiting on Bamboo.

PR #1006, opened 2016-09-08
NMS-8718: vmwarecimquery and vmwareconfigbuilder tools is not working

indigo423:
* JIRA: http://issues.opennms.org/browse/NMS-8718

Changed include from vijava against yavijava library and included libraries in class path for SLF4J logging.


indigo423:
Verified by community member.

PR #1009, opened 2016-09-09
HZN-871: Merge magic-users.properties into users.xml

agalue:
Merge magic-users.properties into users.xml

* JIRA: http://issues.opennms.org/browse/HZN-871
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS955

The idea behind this feature is to be able to configure the "Security Roles" trough either the WebUI or ReST, without the need to modify a file and restart OpenNMS.

To avoid confusions, the roles in groups.xml have been renamed to be called "On-Call Roles" on the WebUI, because that's what they are. That way, they won't be confused with "Security Roles".

The "Security Roles" are hardcoded on org.opennms.web.api.Authentication, and I don't think we immediately have a need to make them configurable, but this would be easy to do, in case we need it in the future (we might have something like security-roles.xml, and we can even control it through the WebUI and ReST as well).

I tested the WebUI Access, the Security Roles configuration, the ReST API and Compass to be sure that everything is OK, on a VM running the branch **jira/HNZ-871**.

The docs in **opennms-docs** have been updated.

The JMX access with either admin or a user that have _ROLE_JMX_ in _jmxremote.access_ is also working.

There is an upgrade task to merge the current content of _magic-users.properties_ into _users.xml_, assuming that users declared on _magic-users.properties_ and not declared in _users.xml_ will be ignored.

By default, if a user in _users.xml_ doesn't have any role, _ROLE_USER_ will be added at runtime, but if the user has at least one role, the configured list of roles will be used. That means, if someone wants a user with _ROLE_REST_ that also requires access to the WebUI, the administrator has to physically add _ROLE_USER_ to that user.

I appreciate of someone can test it prior merging it, but from my point of view, this is ready to be merged.


j-white:
I merged develop into this branch this morning, and resolved the merge conflicts. 

I noticed a couple minor things while testing:
1) The "RTC" user shows up in the users list now, but it is possible to delete it the user. I think we should disable the "Delete" button like we do for the "Admin" user.
2) The upgrade script does not recognize custom roles. If a user has a role called "Metrics" associated with one or more users, the upgrade script will print out a literal "Warning: Invalid role {}. Ignoring..." and the "Metrics" role will not be associated with the existing users.

Aside from that, the feature works well and is a nice improvement over editing magic-users.properties.

agalue:
> On Oct 4, 2016, at 9:33 AM, Jesse White <notifications@github.com> wrote:
> 1) The "RTC" user shows up in the users list now, but it is possible to delete it the user. I think we should disable the "Delete" button like we do for the "Admin" user.
> 
Good catch! I'll fix it.
> 2) The upgrade script does not recognize custom roles. If a user has a role called "Metrics" associated with one or more users, the upgrade script will print out a literal "Warning: Invalid role {}. Ignoring..." and the "Metrics" role will not be associated with the existing users.
> 
Got it, I'll fix this as well.
> Aside from that, the feature works well and is a nice improvement over editing magic-users.properties.
> 
Thanks!

mvrueden:

PR #1010, opened 2016-09-12
NMS-8693: Add ReST Elastic Search Alarms to Foundation2016

gallenc:
NMS-8693:   Add ReST Elastic Search Alarms to Foundation2016
This pull request adds the following OSGi plugins  into OpenNMS Foundation 2016 and updates documentation This is targeting the next point release of Meridian for a customer.

alarm change notifier - generates events when alarms change state

opennms-es-rest - elastic search integration using the ReST interface 

It also adds an internal-plugins-descriptor OSGi module so that both of these plugins can be installed by users from the plugin manager gui.

* JIRA: http://issues.opennms.org/browse/NMS-8693
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS940


features/alarm-change-notifier/main-module/pom.xml
```diff
@@ -0,0 +1,232 @@
...
+  <modelVersion>4.0.0</modelVersion>
+  <artifactId>alarm-change-notifier.main-module</artifactId>
+  <name>alarm-change-notifier.main-module</name>
+  <description>Main Module for alarm-change-notifier</description>
+  <packaging>bundle</packaging>
+
+  <build>
+    <plugins>
+
+      <plugin>
```
j-white: Unnecessary. This is defined in the top-level pom.

features/alarm-change-notifier/main-module/pom.xml
```diff
@@ -0,0 +1,232 @@
...
+        <configuration>
+          <target>1.8</target>
+          <source>1.8</source>
+        </configuration>
+      </plugin>
+
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <version>2.5.3</version>
```
j-white: Please remove the version here, this is managed in the top-level pom.

features/alarm-change-notifier/main-module/src/main/java/org/opennms/plugins/dbnotifier/DbNotifierDataSourceFactory.java
```diff
@@ -0,0 +1,211 @@
...
+				}
+
+				if(istream==null){	
+					LOG.debug("loading database config from classpath file"+OPENNMS_DATASOURCE_CONFIG_FILE_NAME);
+					istream= this.getClass().getClassLoader().getResourceAsStream(OPENNMS_DATASOURCE_CONFIG_FILE_NAME);
+					if(istream==null) throw new RuntimeException("could not load database config from classpath "+OPENNMS_DATASOURCE_CONFIG_FILE_NAME);
+				}
+
+
+				DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
```
j-white: We should use the existing DAO for reading *opennms-datasources.xml* instead of parsing it manually here.

features/opennms-es-rest/main-module/src/main/java/org/opennms/plugins/elasticsearch/rest/archive/OpenNMSHistoricEventsToEs.java
```diff
@@ -0,0 +1,160 @@
...
+		super();
+	}
+
+	/**
+	 * sends events to elastic search returns true if successful
+	 * @return
+	 */
+	public String sendEventsToEs(){
+
+		OnmsRestEventsClient onmsRestEventsClient = new OnmsRestEventsClient();
```
j-white: Can we just use the Event DAO here instead of reaching in via REST?

features/opennms-es-rest/main-module/src/main/java/org/opennms/plugins/elasticsearch/rest/archive/XmlOnmsEvent.java
```diff
@@ -0,0 +1,958 @@
...
+import org.opennms.netmgt.xml.event.Parm;
+import org.opennms.netmgt.xml.event.Value;
+
+
+/**
+ * <p>XmlOnmsEvent class.</p>
+ * This class can unmarshal xml from OpenNMS ReST events interface
+ */
+@XmlRootElement(name="event")
+public class XmlOnmsEvent  implements Serializable {
```
j-white: Is there any reason we can't use the existing OnmsEvent class?

gallenc: I tried to use the the OnmsEvent class but it has a  extra object translation code which means that it doesn't marshal and un-marshall as a simple JAXB pojo class. I found it wasn't possible to make a simple JAXB client using OnmsEvent. All I needed was something to interrogate the OpenNMS rest interface and return a class with native fields.

features/opennms-es-rest/main-module/src/main/java/org/opennms/plugins/elasticsearch/rest/archive/OpenNMSHistoricEventsToEs.java
gallenc: The intent of this is to allow us to import old events into Elastic search from any OpenNMS instance using ReST. So we can import locally and remotely. This allows us to test at the customer site without changing their current Meridian system.
Secondly I looked at the eventdao and the only usable method getEventsAfterDate(final List<String> ueiList, final Date date)  seems to require an explicit list of uei's (i.e. no wild cards for uei). So difficult to get assess to a paged list of all events in the system. Using the ReST interface was the simplest way to get to a paged list of all of the events  and it allows import from external systems.

features/alarm-change-notifier/main-module/src/main/java/org/opennms/plugins/dbnotifier/DbNotifierDataSourceFactory.java
gallenc: I looked at the existing dao but there way no way to get access to all  the raw database information in order to construct the URI needed to set up jdbc-ng.
I needed to get access to the raw definitions of the opennms database hostname, port, database name, username and password to access postgres. This is used to construct a different data source url for use with jdbc-ng. The only way to get that information was to parse the raw opennms-datasources.xml file. It is not accessible from the dao (no suitable get methods on the data source)

j-white: You can unmarshal the DataSourceConfiguration from a input stream using:
    `CastorUtils.unmarshal(DataSourceConfiguration.class, fileInputStream);`

features/opennms-es-rest/main-module/src/main/java/org/opennms/plugins/elasticsearch/rest/archive/OpenNMSHistoricEventsToEs.java
j-white: You can use the CriteriaBuilder with all of the Hibernate-based DAOs to perform arbitrary queries. In this case, you could use something like the following:

```
CriteriaBuilder cb = new CriteriaBuilder(OnmsEvent.class);
cb.lt("eventTime", new Date());
m_eventDao.findMatching(cb.toCriteria());
```

features/opennms-es-rest/main-module/src/main/java/org/opennms/plugins/elasticsearch/rest/archive/XmlOnmsEvent.java
j-white: Can you elaborate on the unmarshaling problems?

If I retrieve a list of events from demo using:
`curl -u demo:demo https://demo.opennms.org/opennms/rest/events > /tmp/events.xml`

The following snippet successfully unmarshals the results:
````
final OnmsEventCollection events = JaxbUtils.unmarshal(OnmsEventCollection.class, new File("/tmp/events.xml"));
System.err.println(events.getCount());
System.err.println(events.get(0).getEventUei());
```

features/alarm-change-notifier/main-module/src/main/java/org/opennms/plugins/dbnotifier/DbNotifierDataSourceFactory.java
gallenc: change done

features/opennms-es-rest/main-module/src/main/java/org/opennms/plugins/elasticsearch/rest/archive/XmlOnmsEvent.java
gallenc: OK I have remove the XMLOnmsEvent files

features/alarm-change-notifier/main-module/pom.xml
gallenc: fixed

features/opennms-es-rest/main-module/src/main/java/org/opennms/plugins/elasticsearch/rest/archive/OpenNMSHistoricEventsToEs.java
gallenc: As discussed, this command can import events into elastic search from remote OpenNMS instances using their rest interfaces. I could use the DAO and criteria for the local machine but this method works for both remote and local so I don't propose to change anything at this time.

features/alarm-change-notifier/main-module/src/test/java/org/opennms/plugins/dbnotifier/test/manual/DbChangeNotifierTest.java
```diff
@@ -0,0 +1,122 @@
...
+import org.junit.Test;
+
+import org.opennms.plugins.dbnotifier.DatabaseChangeNotifier;
+import org.opennms.plugins.dbnotifier.DbNotificationClientQueueImpl;
+import org.opennms.plugins.dbnotifier.DbNotifierDataSourceFactory;
+import org.opennms.plugins.dbnotifier.NotificationClient;
+import org.opennms.plugins.dbnotifier.alarmnotifier.AlarmChangeNotificationClient;
+
+
+public class DbChangeNotifierTest {
```
j-white: This test will never fail since all of the exception are caught. It also assumes that an existing database is provisioned on the localhost, which may not be the case.

features/alarm-change-notifier/main-module/src/test/java/org/opennms/plugins/dbnotifier/test/manual/SimpleJsonTest.java
```diff
@@ -0,0 +1,257 @@
...
+
+		// "suppressedtime":"2016-07-04 16:19:14+01","systemid":"00000000-0000-0000-0000-000000000000","suppresseduntil":"2016-07-04 16:19:14+01","description":"Generic Raspberry Pi Alarm Raise 1","mouseovertext":null,"x733probablecause":0,"lasteventid":4184,"lasteventtime":"2016-07-05 12:00:02+01","managedobjectinstance":null,"alarmacktime":"2016-07-05 13:38:51.753+01","qosalarmstate":null,"ipaddr":"127.0.0.1","alarmackuser":"admin","nodeid":null,"firsteventtime":"2016-07-04 16:19:14+01","severity":5,"ifindex":null,"alarmtype":1,"x733alarmtype":null,"logmsg":"Generic Raspberry Pi Alarm Raise 1","tticketid":null,"firstautomationtime":null,"clearkey":null,"managedobjecttype":null,"eventuei":"uei.opennms.org\/application\/generic\/piAlarmRaise","counter":3,"applicationdn":null,"operinstruct":"","ossprimarykey":null,"stickymemo":null,"tticketstate":null,"alarmid":285,"serviceid":null,"reductionkey":"uei.opennms.org\/application\/generic\/piAlarmRaise:0:127.0.0.1:1","suppresseduser":null,"lastautomationtime":null,"eventparms":"PiIoId=1(string,text)"
+		//String testString = "[{},{}]";
+
+		String testString = "["+testString1+","+testString2+"]";
```
j-white: These tests still pass if I set `String testString  = "";`

j-white:
I just noticed that all of tests are disabled by default and flagged as "manual". Is there any reason not to put the effort in to make these run regularly?

opennms-base-assembly/src/main/filtered/etc/events/AlarmChangeNotifierEvents.xml
```diff
@@ -0,0 +1,259 @@
...
+  <event>
+    <uei>uei.opennms.org/plugin/AlarmChangeNotificationEvent/AlarmDeleted</uei>
+    <event-label>Alarm Change Notification Event : Alarm Deleted</event-label>
+    <descr>
+    &lt;p&gt;Alarm %parm[alarmid]% Deleted&lt;p&gt;
+    
+    &lt;p&gt;Deleted Alarm Values&lt;p&gt;
+    &lt;p&gt;%parm[oldalarmvalues]%&lt;p&gt;
+    </descr>
+    <logmsg dest="logndisplay">
```
j-white: Do we need to persist these to the database? `donotpersist` may be more suitable here.

j-white:
Can we remove all the `maven-compiler-plugin` definitions from the poms. I don't think there's any reason to override the source and target versions in any of these projects.

features/alarm-change-notifier/main-module/src/test/java/org/opennms/plugins/dbnotifier/test/manual/DbChangeNotifierTest.java
gallenc: This is simply a manual test to allow triggers to be tested in the database. It's not intended as an auto test. Will need to look at automatic integration testing later when we have a captive opennms to generate alarm changes

features/alarm-change-notifier/main-module/src/test/java/org/opennms/plugins/dbnotifier/test/manual/SimpleJsonTest.java
gallenc: This is simply a manual developer test to ensure we can parse the database JSON notification content into an alarm.

opennms-base-assembly/src/main/filtered/etc/events/AlarmChangeNotifierEvents.xml
gallenc: The alarm change notifier is a stand alone module which adds value by providing alarm change events which users can use to see who or what changed an alarm on a device. 
By default the events are persisted so that  the alarm life cycle can be read from the OpenNMS event database.  These events could be used in reports or searched for by users. The events also show that the alarm change notifier is working when it is deployed. Users can modify the alarm change event definitions to donotpersist if they find them annoying but I would prefer them to be persisted by default.

PR #1014, opened 2016-09-14
NMS-8739: top 20 nodes by I/O wait

unicolet:
See issue [NMS-8739](http://issues.opennms.org/browse/NMS-8739).

A simple jasper reports that shows the top 20 nodes by I/O Wait.
Works for net-snmp nodes (sorry, no windows) and requires that Store By Foreign Source is enabled.

Crafted with love in Fulda by me and @indigo423 
- JIRA: http://issues.opennms.org/browse/NMS-8739

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/database-reports.xml
+++ b/opennms-base-assembly/src/main/filtered/etc/database-reports.xml
@@ -7,35 +7,40 @@
                 report-service="availabilityReportService" description="standard opennms report in calendar format" />
         <report id="defaultClassicReport" display-name="Default classic report"
                 report-service="availabilityReportService" description="standard opennms report in tabular format" />
-        <report id="Early-Morning-Report" display-name="Early morning report" online="true" 
+        <report id="Early-Morning-Report" display-name="Early morning report" online="true"
                 report-service="jasperReportService" description="Global overview of outages, notifications and events in last 24 hours" />
-        <report id="Response-Time-Summary-Report" display-name="Response Time Summary for node" online="true" 
+        <report id="Response-Time-Summary-Report" display-name="Response Time Summary for node" online="true"
                 report-service="jasperReportService" description="Response Time by node across one or more surveillance categories. Note: % can be used as a place holder for any string literal" />
-        <report id="Node-Availability-Report" display-name="Availability by node" online="true" 
+        <report id="Node-Availability-Report" display-name="Availability by node" online="true"
                 report-service="jasperReportService" description="Availability by node across one or more surveillance categories. Note: % can be used as a place holder for any string literal" />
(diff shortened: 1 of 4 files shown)


indigo423:
My personal point of view. I would recommend to make a dedicated repository for this kind of reports for several reasons:

* The report serves a special use case and is not possible to enable it by default
* You need to have OpenNMS to be configured in a certain way -> README in the repo
* The report needs to maintain a own life cycle and a github repo provides all the tooling, version numbers, depends on a specific version of OpenNMS, issue tracking etc. etc.

soleger:
@indigo423 I'm not clear what your comments mean from the PR. Does this report depend on something repo outside of the OpenNMS codebase?

indigo423:
I would create a GitHub repository, drop in the JRXML file and a README how to install it and how to configure OpenNMS to be able to use the report. Here is a suggestion, make a version of the report and say compatible with Horizon 18.x.y / Meridian 2016.x.y. I would vote to maintain this kind of reports outside of the opennms core repository. Let them be forkable for others and let people open issues on the  report itself for improvements and not entangle the lifecycle of the report with OpenNMS own lifecycle.

mvrueden:

j-white:
Pushed into `pr/jira/NMS-8739`.

indigo423:
@j-white 
I suggest to let the report template live in something like here: https://github.com/opennms-forge/jasper-reports-top-io-wait

It allows people to fork, improve and open issues for it as it's own. It is also documented how to install the report in any version of OpenNMS Horizon / OpenNMS Meridian. It doesn't has impact on support when it is maintained inside the opennms repository.

j-white:
I'm all for it. The report is disabled by default anyhow.

@unicolet, what do you think?

unicolet:
@j-white I prefer to have it in opennms and to enable it, if needed :-)

indigo423:
I'll follow the majority of votes, don't take my opinion as a blocker :)

mvrueden:
Maintaining existing reports is quite honest a pain in the ass. Every JasperReport version upgrade, requires us to maintain the reports. 

Therefore, I think if a report can be enabled by default and is a really good report, which helps most of the people using opennms out there, it is reasonable to pull it into the opennms codebase. If it cannot be enabled by default, because it requires a certain setup, or it is a special report, only useful for a minority of the opennms users, I don't think we should pull it into the code base. It is up for discussion, if we put it into an "examples" or "contrib" folder, instead, though.

In general I like the idea of having a "opennms/reports" repository instead. 

As this report is not enabled by default, I vote against pulling it into the OpenNMS code base.
If required, I can create a "reports" repository and move it there.

j-white:
After some more discussion we've agreed to merge this into the code-base since we don't currently have a better workflow for handling custom/contributed reports.

I've pushed the changes to `pr/jira/NMS-8739`. Waiting on Bamboo.

PR #1017, opened 2016-09-14
NMS-8708: Trapd reloads SNMPv3 config when no changes have occurred

MS043660:
Change Log-
Updating of TrapRecieverImpl class where earlier it was reloading Trapd everytime on Trap Configuration xml loading through rest.


soleger:
I just merged develop into this branch to see if the Bamboo failures are addressed.

features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapReceiverImpl.java
```diff
@@ -106,7 +106,11 @@ public void setTrapdConfig(TrapdConfiguration newTrapdConfig) {
      */
     public static boolean compareSnmpV3UsersMap(Map<String, SnmpV3User> existingSnmpV3UserMap, Map<String, SnmpV3User> updatedSnmpV3Usermap) {
 
-        if (existingSnmpV3UserMap.isEmpty() || updatedSnmpV3Usermap.isEmpty()) {
+        if ((existingSnmpV3UserMap.isEmpty() && updatedSnmpV3Usermap.isEmpty())
+                || (!existingSnmpV3UserMap.isEmpty() && updatedSnmpV3Usermap.isEmpty())) {
```
soleger: This statement doesn't make sense. It reduces to ```updatedSnmpV3Usermap.isEmpty()```. Is that the intended check?

features/events/traps/src/test/java/org/opennms/netmgt/trapd/TrapdListenerBlueprintIT.java
```diff
@@ -66,7 +75,19 @@
 	private final TrapNotificationLatch m_handler = new TrapNotificationLatch(3);
 
 	private int m_port;
-
+	
+        private TrapdConfigBean trapConfiguration;
```
soleger: Move these private fields into each test method. Reusing fields between tests is a bad design.

features/events/traps/src/test/java/org/opennms/netmgt/trapd/TrapdListenerBlueprintIT.java
```diff
@@ -195,4 +216,104 @@ public void setTrapIdentity(TrapIdentity trapIdentity) {
...
+                        "onUpdate").bean(trapReciever, "setTrapdConfig").to("mock:result");
+            }
+        });
+        camelContext.start();
+        
+        for (int i = 0; i < 10; i++) {
+        mockEndPoint = camelContext.getEndpoint("mock:result", MockEndpoint.class);
+        mockEndPoint.expectedMessageCount(1);
+            producerTemplate = camelContext.createProducerTemplate();
+            producerTemplate.start();
```
soleger: I don't think that this loop is doing anything, you're just starting the producerTemplate over and over without sending any messages. Change this test to use a timer or something so that you can invoke the HTTP route multiple times.

features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapReceiverImpl.java
MS043660: This is intended because earlier this was the step which i missed,ie when both list  are empty,still it was considering that some data is there to get updated and hence it was restarting everytime.Now once this check is done we are ensuring that for empty case also we dont need to check anything and this works well.

features/events/traps/src/test/java/org/opennms/netmgt/trapd/TrapdListenerBlueprintIT.java
MS043660: Ok,This has been addressed.I thought i can reuse this.Now i have reverted back the changes and written as suggested.

mvrueden:

j-white:
There are merge conflicts there, can we try merging develop to this branch?

MS043660:
@j-white : Do you want me to merger develop and push the code ??

j-white:
Yeah, exactly.

MS043660:
@j-white : Done.

features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapReceiverImpl.java
j-white: I also find this code to be confusing, why don't just implement a proper equals method in SnmpV3User instead?

MS043660:
This has been fixed.

Regards,
Malatesh Sudarshan
Software Engineer, CTS I&O Platforms, InfraOSS
malatesh.sudarshan@cerner.com<mailto:malatesh.sudarshan@cerner.com> | +91 8553333802 mobile
Cerner Corporation | www.cerner.com<http://www.cerner.com/>

MS043660:
The review comment @j-white  has been addressed and simplified.

j-white:
The logic here still appears to be unnecessarily complicated, why can't we just implement a proper equals() method in the SnmpV3User objects?

MS043660:
@j-white : I am now confused,Could you please let me know with respect to which code part your telling to use equals because currently there is only two loops to check whether maps are empty and return the items.

j-white:
Right, we are trying to determine if there are any difference between `newSnmpV3Users` and `existingSnmpV3Users`.

What I'm saying is this can be rewritten simply as `Objects.equals(existingSnmpV3Users, newSnmpV3Users)`, provided that the `SnmpV3User` class implements a proper equals() method.

In my opinion, this is a better solution than manually comparing the maps and their contents.

features/events/traps/pom.xml
```diff
@@ -76,6 +76,11 @@
     </plugins>
   </build>
   <dependencies>
+   <dependency>
+        <groupId>javax.ws.rs</groupId>
+        <artifactId>javax.ws.rs-api</artifactId>
+        <version>2.0.1</version>
```
RangerRick: We already define the `javax.ws.rs-api` version in the top-level `pom.xml` file, so this line can be removed (to avoid future version conflicts).

features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapReceiverImpl.java
```diff
@@ -97,6 +107,19 @@ public void setTrapdConfig(TrapdConfiguration newTrapdConfig) {
         }
     }
 
+	@Override
+	public boolean equals(Object obj) {
```
RangerRick: If you implement `equals()` you should also implement `hashCode()`

features/events/traps/pom.xml
```diff
@@ -76,11 +76,6 @@
     </plugins>
   </build>
   <dependencies>
-   <dependency>
-        <groupId>javax.ws.rs</groupId>
-        <artifactId>javax.ws.rs-api</artifactId>
-        <version>2.0.1</version>
```
MS043660: @RangerRick : This has been deleted but when we  tried to trace back for this change we were  unable to find in any previous commit.This looks bit strange.

features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapReceiverImpl.java
j-white: Also, equals() is defined on the wrong class, it should be defined on SnmpV3User instead.

MS043660: @j-white :Sorry there has been mistake during code merge.Some files have got deleted and hence certain changes have not got reflected,Just now i got to know those.I have pushed the new code,Which you can review now.

soleger:
Closed in favor of PR #1147.

PR #1018, opened 2016-09-14
NMS-8741: Weather icons for categories box

unicolet:
A bit of JS to add an icon (normal:sunny, warning:partly cloudy, critical:storms) to the categories box in the main page.

image

* JIRA: http://issues.opennms.org/browse/NMS-8741


christianpape:
Seriously?

unicolet:
Seriously what?

christianpape:
Sorry, I didn't want to appear impolite but I did not get how percentage availability relate to weather icons. Of course, sun is good and rain is bad, I know, but from here it is only small step to animated Gifs with kittens on the start page :-)

unicolet:
@christianpape the thresholds for icons are exactly the same as those defined in ``categories.xml`` for the text so there is not a split meaning for icons and text. I would have liked to have more levels (6 I think) like in Jenkins but then the icons and the text might have assumed different meanings if a user customized ``categories.xml``.

Actually this was a request from a customer of mine. And, I mean, what's wrong with cats? :D

christianpape:
Ok, got it. I'm just no fan of weather icons, sorry. But of course this is a matter of taste.

soleger:
I'm not a fan either since the icons correspond exactly to the severity that is communicated by the color of the availability pane. This makes them redundant.

I can understand making a modification like this at customer request but I don't think we need it in the default OpenNMS codebase.

soleger:
Closing PR.

PR #1019, opened 2016-09-14
NMS-8722: XSS vulnerability in display of SNMP sysName and trap varbind values

jeffgdotorg:
* JIRA: http://issues.opennms.org/browse/NMS-8722


j-white:
Did we also intend to escape the SNMP attributes on the node.jsp page?

PR #1027, opened 2016-09-19
HZN-766: Send Trap & Syslog messages using XML instead of serialized POJOs

DP044946:
Commiting code for sending Trap & Syslog messages using XML instead of serialized POJOs


core/camel/src/main/java/org/opennms/core/camel/MinionDTO.java
```diff
@@ -0,0 +1,118 @@
+/*******************************************************************************
```
cgorantla: There's already DTO for Minion , may be you can use that, https://github.com/OpenNMS/opennms/blob/develop/features/minion/heartbeat/common/src/main/java/org/opennms/minion/heartbeat/common/MinionIdentityDTO.java

soleger: @cgorantla This DTO serves a different purpose. It is a base class that implements a generic message pattern similar to a Camel exchange: it has a Map<String,String> of header values and then a <body> containing an arbitrary message. We're using it to normalize the message format for syslogs and traps. We'll probably look at combining this with the RPC/Sink DTOs in the future.

features/events/syslog/src/test/java/org/opennms/netmgt/syslogd/SyslogDTOMapperTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+package org.opennms.netmgt.syslogd;
+
+import org.junit.Test;
+
+public class SyslogDTOMapperTest {
+
+	@Test
+	public void test() {
+		// TODO: Implement test
```
j-white: Can we either implement this, or remove it?

features/events/traps/src/test/java/org/opennms/netmgt/trapd/TrapDTOMapperTest.java
```diff
@@ -0,0 +1,108 @@
...
+		mapper.setDistPollerDao(distPollerDao);
+
+		TrapDTO trapDto = mapper.object2dto(snmp4JV2cTrap);
+		System.out.println("trapDto is : " + trapDto);
+		System.out.println("trapDto.getBody() is : " + trapDto.getBody());
+		System.out.println("trapDto.getCommunity() is : " + trapDto.getHeader(TrapDTO.COMMUNITY));
+
+		TrapNotification snmp4JV2cTrap1 = TrapDTOToObjectProcessor.dto2object(trapDto);
+
+		// TODO: Add assertions to make sure that conversion worked
```
j-white: Can we either implement this, or remove it? This test doesn't currently test anything.

j-white:
We should also marshal/unmarshal tests for the DTOs that were introduced here.

soleger:
The unmarshalling tests were causing me a lot of problems because of HashMap ordering in the headers. I tried to add an XmlAdapter so that the maps were always unmarshalled in alphabetical order but this triggered some other OSGi classloading issues that I don't have time to address right now.

I've created a new branch that I'll merge and then I'll open a new issue for writing marshal/unmarshal tests for the syslog and trap DTOs.

https://issues.opennms.org/browse/HZN-906

PR #1028, opened 2016-09-20
NMS-8697: Collect Elasticsearch stats via REST using the XmlCollector

pk015603:
Nms 8697: Collect Elasticsearch stats via REST using the XmlCollector

Elasticsearch exposes its management metrics over HTTP/REST/JSON. Here we have written a XML data collector which will read the JSON response and create the graph metrics.

https://issues.opennms.org/browse/NMS-8697


opennms-base-assembly/pom.xml
```diff
@@ -708,6 +708,11 @@
       <groupId>org.jsoup</groupId>
       <artifactId>jsoup</artifactId>
     </dependency>
+    <dependency>
```
soleger: Undo this change, it is not necessary.

pk015603: I got the dependency issue, so i added this dependency.

soleger: Can you describe the issue?

pk015603:
Sure, I was about to ask about this, how we need to dynamically submit the ipAddress? Sure i will change to {ipaddr}. Does it take localhost or can we configure some other IP Address?

opennms-base-assembly/pom.xml
pk015603: You are mentioning about jsoup dependency or protocols? I used collector class 'org.opennms.protocols.xml.collector.XmlCollector' in collectd and this class is present in this particular package.

opennms-base-assembly/src/main/filtered/etc/xml-datacollection-config.xml
```diff
@@ -0,0 +1,15 @@
...
+                           xmlns="http://xmlns.opennms.org/xsd/config/xml-datacollection">
+    <xml-collection name="elasticsearch-stats">
+        <rrd step="300">
+            <rra>RRA:AVERAGE:0.5:1:2016</rra>
+            <rra>RRA:AVERAGE:0.5:12:1488</rra>
+            <rra>RRA:AVERAGE:0.5:288:366</rra>
+            <rra>RRA:MAX:0.5:288:366</rra>
+            <rra>RRA:MIN:0.5:288:366</rra>
+        </rrd>
+        <xml-source url="http://localhost:9200/_cluster/stats">
```
soleger: "localhost" should be changed to "{ipaddr}" so that the IP address being collected is substituted for the value in the URL.

soleger:
One issue here that may be causing some confusion: the XML collector has its own Maven assembly in the protocols/xml project. For this reason, the xml-datacollection-config.xml and elasticsearch-stats.xml files need to be merged with the existing content in the protocols/xml project.

indigo423:
I would vote -1 for merging it into default configs.

soleger:
@indigo423 Do you have a reason for your preference? I think that with the amount of Elasticsearch integrations that we have built into the codebase (and the significant interest in these integrations from users), we need to treat monitoring Elasticsearch as a top-tier priority, similar to the OpenNMS JMX monitoring.

soleger:
I squash merged these changes. Closing PR.

commit 68dfe0b7d929309182a3242dc5bff90b3d2fbf56

PR #1032, opened 2016-09-21
NMS-8742: Add documentation for Mattermost/Slack notifications

jeffgdotorg:
* JIRA: https://issues.opennms.org/browse/NMS-8742

In order to have a place to hang these docs, I wrote a (rather basic) new chapter on notifications for the admin guide.  It needs fleshing out, but it's useful in its present form.

### DIFF
--- a/opennms-doc/guide-admin/src/asciidoc/index.adoc
+++ b/opennms-doc/guide-admin/src/asciidoc/index.adoc
@@ -125,6 +125,14 @@ include::text/performance-data-collection/collectors/wsman/detector.adoc[]
 == Events
 include::text/events.adoc[]
 
+[[ga-notifications]]
+== Notifications
+include::text/notifications/introduction.adoc[]
+include::text/notifications/getting-started.adoc[]
+include::text/notifications/concepts.adoc[]
+include::text/notifications/strategies/mattermost.adoc[]
+include::text/notifications/strategies/slack.adoc[]
+
 [ga-provisioning]
 == Provisioning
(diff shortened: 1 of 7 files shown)


opennms-doc/guide-admin/src/asciidoc/index.adoc
```diff
@@ -125,6 +125,15 @@ include::text/performance-data-collection/collectors/wsman/detector.adoc[]
 == Events
 include::text/events.adoc[]
 
+[[ga-notifications]]
+== Notifications
+include::text/notifications/introduction.adoc[]
+include::text/notifications/getting-started.adoc[]
+include::text/notifications/concepts.adoc[]
+include::text/notifications/bonus-strategies.adoc[]
+include::text/notifications/strategies/mattermost.adoc[]
```
indigo423: I've noticed we use 3 different terms:
* Notification Commands
* Bonus Methods
* Strategies

I would suggest to use just one term and call it "Notification Commands" which is the same as the file name to make it clearer where it needs to be configured and rename the directory from "strategies" to "commands" to make it easier where all extensions need to be documented and get self maintained when new extensions are written.

Otherwise I would remove the "Bonus Strategies" and  create for each Notification Command a file in the "commands" directory to have a self-maintaining list of extensions as they come. In the introduction I would also eliminate the list which lists all existing Notification Commands which would be then duplicated and have to be maintained as well.

We get quite deep in our structure, I would suggest to refactor the "Concept" layer out and use "Introduction" section for the more generic stuff.

PR #1034, opened 2016-09-21
HZN-879: location specific snmp-config

cgorantla:
Adding an attribute location ( Minion location ) to Definition 
Modify DAO to load it.

* JIRA: http://issues.opennms.org/browse/hzn-879
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS980


opennms-config-jaxb/src/main/java/org/opennms/netmgt/config/snmp/SnmpConfigVisitor.java
```diff
@@ -45,4 +45,6 @@
     
     void visitIpMatches(List<String> ipMatches);
     void visitIpMatchesFinished();
+
+    String getLocation();
```
j-white: This should not be part of the visitor interface.

opennms-config-jaxb/src/main/resources/xsds/snmp-config.xsd
```diff
@@ -248,7 +248,14 @@
                 <documentation>Match Octets (as in IPLIKE)</documentation>
               </annotation>
             </element>
-          </sequence>
+		  </sequence>
```
j-white: Can we fix the indentation to match the existing document?

opennms-config-jaxb/src/main/java/org/opennms/netmgt/config/snmp/Definition.java
```diff
@@ -184,10 +197,23 @@ public boolean equals(final Object obj) {
...
+        } else if (!m_location.equals(other.m_location)) {
+            return false;
+        }
+
         return true;
     }
 
     public void visit(final SnmpConfigVisitor visitor) {
+
+        if ((visitor.getLocation() != null) && (!visitor.getLocation().equals(m_location))) {
```
j-white: This logic is in the wrong place, it should be up to the visitor implementation to decide if it wants to visit a definition or not.

opennms-config-jaxb/src/main/java/org/opennms/netmgt/config/snmp/SnmpConfigVisitor.java
cgorantla: done

opennms-config-jaxb/src/main/resources/xsds/snmp-config.xsd
cgorantla: done

opennms-config-jaxb/src/main/java/org/opennms/netmgt/config/snmp/Definition.java
cgorantla: removed this logic

j-white:
Looks good, waiting on Bamboo.

PR #1037, opened 2016-09-22
HZN-882: use /etc/sysconfig for setting Minion options

RangerRick:
This PR adds an `/etc/sysconfig/minion` file which mirrors the Karaf startup environment variables used for overriding various options on launch.

* JIRA: https://issues.opennms.org/browse/HZN-882
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS996

### DIFF
--- a/tools/packages/minion/minion.spec
+++ b/tools/packages/minion/minion.spec
@@ -155,8 +155,11 @@ rm -rf %{buildroot}%{minioninstprefix}/demos
 
 # Copy over the run script
 mkdir -p %{buildroot}%{_initrddir}
-sed -e 's,@INSTPREFIX@,%{minioninstprefix},g' %{_builddir}/%{_name}-%{version}-%{release}/tools/packages/minion/minion.init > "%{buildroot}%{_initrddir}"/minion
+sed -e 's,@INSTPREFIX@,%{minioninstprefix},g' -e 's,@SYSCONFDIR@,%{_sysconfdir}/sysconfig,g'  %{_builddir}/%{_name}-%{version}-%{release}/tools/packages/minion/minion.init > "%{buildroot}%{_initrddir}"/minion
 chmod 755 "%{buildroot}%{_initrddir}"/minion
+
+install -d -m 755 %{buildroot}%{_sysconfdir}/sysconfig
+install -m 644 "%{_builddir}/%{_name}-%{version}-%{release}/tools/packages/minion/minion.sysconfig" "%{buildroot}%{_sysconfdir}/sysconfig/minion"
 
 # Extract the core repository
 mkdir -p %{buildroot}%{minionrepoprefix}/core
@@ -195,6 +198,7 @@ rm -rf %{buildroot}
 %files container -f %{_tmppath}/files.container
 %defattr(664 root root 775)
 %attr(755,root,root) %{_initrddir}/minion
+%attr(644,root,root) %config(noreplace) %{_sysconfdir}/sysconfig/minion
 %attr(644,root,root) %{minioninstprefix}/etc/featuresBoot.d/.readme
 
 %post container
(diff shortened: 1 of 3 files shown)


tools/packages/minion/minion.spec
```diff
@@ -155,9 +155,12 @@ rm -rf %{buildroot}%{minioninstprefix}/demos
 
 # Copy over the run script
 mkdir -p %{buildroot}%{_initrddir}
-sed -e 's,@INSTPREFIX@,%{minioninstprefix},g' %{_builddir}/%{_name}-%{version}-%{release}/tools/packages/minion/minion.init > "%{buildroot}%{_initrddir}"/minion
+sed -e 's,@INSTPREFIX@,%{minioninstprefix},g' -e 's,@SYSCONFDIR@,%{_sysconfdir}/sysconfig,g'  %{_builddir}/%{_name}-%{version}-%{release}/tools/packages/minion/minion.init > "%{buildroot}%{_initrddir}"/minion
```
soleger: ooh i didn't know you could pass two scripts on the same sed command line, neato :)

indigo423:
Is `sysconfig` the right place for application specific configuration?

fooker:
What's wrong with `/etc/opennms/minion`? It integrates nicely with existing opennms installs on the same host and works on distributions that does not have sysconfig.

-- 


-----Original Message-----

pbrane:
> On Sep 23, 2016, at 8:11 AM, Ronny Trommer <notifications@github.com> wrote:
> 
> Is sysconfig the right place for application specific configuration?
> 
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 

Remote poller has been like this for 10 years. :)

RangerRick:
@indigo423 Are you thinking `sysctl`?  `/etc/sysconfig` is literally _the_ packaging-level place to put "startup configuration" on redhat.

RangerRick:
Also, I looked into whether we needed to do any `opennms.properties.d` kind of stuff but really, non of that really needs to happen *before* the Karaf launch, and everything after should be done through Karaf properties/configs.

RangerRick:
@fooker Running a minion on the same system as OpenNMS seems kind of like a degenerate case in the first place; pretty much the only people doing that would be developers, wouldn't it?  I could see putting it into `$MINION_HOME/etc`, but I opted for the system-provided way for doing such things instead, since it's the standard.  Like I said in the JIRA issue, we can use `/etc/default/minion` on Debian/Ubuntu, which is their way of doing the same thing.  Just haven't done the Debian packaging yet so haven't gotten to implementing that bit.

soleger:
This looks fine to me. I really don't want to drop files into /etc. There are LSB guidelines and other packaging guidelines that we need to research before we do that. Ben is following the SysVInit conventions on Red Hat by using /etc/sysconfig.

indigo423:
the problem with this kind of config, it adds additional problems especially now when you abstractions on top of configuration aka Configuration Management. The LSB design of directories was never built to deal with that. The fact Linux distributions deals differently with this kind of stuff like Red Hat with `/etc/sysconfig` and Debian with `/etc/default` is nowadays totally pain in the ass to deal with also with dealing in stuff like Docker, rkt or whatever. It would so much easier to get our configuration out of the way of this weird things and make it for OpenNMS as easy as possible be generic and distribution independent configurable. Moving it just to a relative directory below $something/opennms would make life for automation people much easier and less painful.

RangerRick:
We could make it *also* use a `minion.conf` in `$MINION_HOME/etc` if this is really such a burden, but that seems pretty ridiculous to me.  All this time OpenNMS has been doing it wrong by being this big monolithic piece of junk with its own etc even for system-level things, and I'm really surprised at getting pushback from finally supporting the platform-recommended way of providing startup configuration.

When you're making a docker container, you're picking a distribution anyways, so how does it make it more difficult when there will only be one way to configure it in that base anyways?

The reason we have no Debian package is not because of this, but because no one has had the time to write one yet (it's on my TODO list).

PR #1038, opened 2016-09-23
Remove unnecessary brokerUri configs from the Minion smoke tests

soleger:
* Bamboo: TBD


RangerRick:
Will this break 18.x tests?  Do we need to rev the smoke-test API?

soleger:
This isn't an issue, it's just a change to the smoke test code because we use ```queuingservice``` now.

PR #1039, opened 2016-09-23
HZN-618: be more aggressive about stopping Karaf

RangerRick:
This PR does a cleaner version of the stop/kill/kill-9 procedure that the OpenNMS init script uses for making sure `stop` really stops.

* JIRA: http://issues.opennms.org/browse/HZN-618
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS998

### DIFF
--- a/tools/packages/minion/minion.init
+++ b/tools/packages/minion/minion.init
@@ -21,13 +21,42 @@ DESC="Minion"
 PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
 MINION_HOME="@INSTPREFIX@"
 SYSCONFDIR="@SYSCONFDIR@"
+STOP_TIMEOUT=10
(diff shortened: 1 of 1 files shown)


tools/packages/minion/minion.init
```diff
@@ -21,15 +21,44 @@ DESC="Minion"
 PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
 MINION_HOME="@INSTPREFIX@"
 SYSCONFDIR="@SYSCONFDIR@"
+STOP_TIMEOUT=10
```
soleger: Seems like this variable isn't named correctly, since it's the number of retries. Maybe STOP_RETRIES instead?

RangerRick: I suppose.  I was just copying the one in the `opennms` init script.

PR #1042, opened 2016-09-26
HZN-445: Use JAAS to authenticate ActiveMQ connections

j-white:
JIRA: https://issues.opennms.org/browse/HZN-445
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS993-4

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/magic-users.properties
+++ b/opennms-base-assembly/src/main/filtered/etc/magic-users.properties
@@ -37,7 +37,7 @@ user.rtc.password=rtc
 
 # A comma-separated list of role keys.  A role.{KEY}.name and 
 # role.{KEY}.users property must be set for each key in this property.
-roles=rtc, admin, rouser, dashboard, provision, remoting, rest, asset, mobile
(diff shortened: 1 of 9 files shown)


opennms-base-assembly/src/main/filtered/etc/magic-users.properties
```diff
@@ -73,3 +73,7 @@ role.rest.users=iphone
 
 role.mobile.name=OpenNMS Mobile User
 role.mobile.users=
+
+role.minion.name=OpenNMS Minion User
```
agalue: Considering that https://github.com/OpenNMS/opennms/pull/1009 has been waiting for a while (and it is pretty much ready to be merged), guess we should decide if this one will be merged first, and then I have to rework the solution for HZN-871, or merge 1009 first, and then use the new methodology here (which is a lot simpler and easy to use).

soleger: Won't the upgrade script in #1009 handle all of the changes?

PR #1046, opened 2016-09-27
HZN-902: Integrate location based Defintions in provision/poller/Rest APIs

cgorantla:
* JIRA: http://issues.opennms.org/browse/hzn-902
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1003/latest

         Also fixes a bug in retrieving definitions


opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/SnmpConfigRestService.java
```diff
@@ -121,12 +122,12 @@ protected void tearDown() {
     @GET
     @Produces({MediaType.APPLICATION_XML, MediaType.APPLICATION_JSON, MediaType.APPLICATION_ATOM_XML})
     @Path("{ipAddr}")
-    public SnmpInfo getSnmpInfo(@PathParam("ipAddr") String ipAddr) {
+    public SnmpInfo getSnmpInfo(@PathParam("ipAddr") String ipAddr, @QueryParam("location")String location) {
```
j-white: We should document this extra param here: https://github.com/OpenNMS/opennms/blob/develop/opennms-doc/guide-development/src/asciidoc/text/rest/snmp_configuration.adoc

core/test-api/snmp/src/main/java/org/opennms/core/test/snmp/ProxySnmpAgentConfigFactory.java
```diff
@@ -54,17 +54,25 @@ public ProxySnmpAgentConfigFactory(final InputStream config) throws MarshalExcep
...
-			LOG.debug("No agent address mapping found for {}!  Try adding a @JUnitSnmpAgent(host=\"{}\", resource=\"...\" entry...", addressString, addressString);
-    		return super.getAgentConfig(address);
-    		// throw new IllegalArgumentException("No agent address mapping found for " + addressString);
-    	}
+        return getAgentConfig(address, null);
+    }
+
+    @Override
+    public SnmpAgentConfig getAgentConfig(final InetAddress address, String location) {
+        final SnmpAgentConfigProxyMapper mapper = SnmpAgentConfigProxyMapper.getInstance();
```
soleger: Is there any particular reason that SnmpAgentConfigProxyMapper needs to use this singleton pattern?

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/SnmpConfigRestService.java
cgorantla: Done

core/test-api/snmp/src/main/java/org/opennms/core/test/snmp/ProxySnmpAgentConfigFactory.java
cgorantla: I used already existing code but looks like SnmpAgentConfigProxyMapper is used to store InetAddress, SnmpAgentAddress in HashMap  which will be subsequently used to retrieve corresponding agent Address

fooker: This method introduces `location` which is unused in the code below.

PR #1050, opened 2016-09-29
NMS-8327:  Implement explicit timeout  for JMXDetector 

cgorantla:
* JIRA: http://issues.opennms.org/browse/nms-8327
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1010-4


cgorantla:
Verified by setting timeout on JSR160Detector and it properly takes given timeout

core/jmx/impl/src/main/java/org/opennms/netmgt/jmx/impl/connection/connectors/Jsr160ConnectionFactory.java
```diff
@@ -72,21 +83,36 @@ public static JmxServerConnectionWrapper getMBeanServerConnection(Map<String,?>
...
         if (factory == null || factory.equals("STANDARD")) {
-                url = getUrl(address, port, protocol, urlPath);
+                final JMXServiceURL url = getUrl(address, port, protocol, urlPath);
                 
                 // Connect a JSR 160 JMXConnector to the server side
-                JMXConnector connector = JMXConnectorFactory.connect(url);
-                MBeanServerConnection connection = connector.getMBeanServerConnection();
-                
-                return new Jsr160ConnectionWrapper(connector, connection);
+                ExecutorService executor = Executors.newCachedThreadPool();
```
j-white: I would make this executor static and final instead of re-creating it on every call.

core/jmx/impl/src/main/java/org/opennms/netmgt/jmx/impl/connection/connectors/Jsr160ConnectionFactory.java
```diff
@@ -57,6 +65,11 @@
     
     private static final Logger LOG = LoggerFactory.getLogger(Jsr160ConnectionFactory.class);
 
+    // Set default timeout to 2 secs as specified in AbstractDetector
+    private static final long DEFAULT_TIMEOUT = 2000;
```
j-white: This connection factory is also used by other components.

Can we have it use 2 seconds as the default timeout when invoked by a detector, and 30 seconds otherwise?

core/jmx/impl/src/main/java/org/opennms/netmgt/jmx/impl/connection/connectors/Jsr160ConnectionFactory.java
```diff
@@ -72,21 +85,36 @@ public static JmxServerConnectionWrapper getMBeanServerConnection(Map<String,?>
...
+                     connector = future.get(timeout, TimeUnit.MILLISECONDS);
+                     MBeanServerConnection connection = connector.getMBeanServerConnection();
+                     return new Jsr160ConnectionWrapper(connector, connection);
+                } catch (InterruptedException | ExecutionException | TimeoutException e) {
+                    LOG.info("Exception connecting JMXConnectorFactory url {} , Error: {}", url, e.getMessage());
+                    throw new ConnectException("Error connecting JMXConnectionFactory  " + url);
+                } finally {
+                    future.cancel(true);
+                }
+
```
mvrueden: These lines (105 to 116) look identical to lines (143 to 154). Can we make them a method?

core/jmx/impl/src/main/java/org/opennms/netmgt/jmx/impl/connection/connectors/Jsr160ConnectionFactory.java
cgorantla: Done

core/jmx/impl/src/main/java/org/opennms/netmgt/jmx/impl/connection/connectors/Jsr160ConnectionFactory.java
cgorantla: Done

j-white:
I'm waiting for the develop branch to go green again in Bamboo before merging this one in.

core/jmx/impl/src/main/java/org/opennms/netmgt/jmx/impl/connection/connectors/Jsr160ConnectionFactory.java
mvrueden: I disagree. It is a (massive) duplication of code. The only difference - you said it yourself - is the future. Looks like the perfect example to introduce a method.

core/jmx/impl/src/main/java/org/opennms/netmgt/jmx/impl/connection/connectors/Jsr160ConnectionFactory.java
```diff
@@ -72,21 +85,36 @@ public static JmxServerConnectionWrapper getMBeanServerConnection(Map<String,?>
...
+                 JMXConnector connector = null;
+                 try {
+                     connector = future.get(timeout, TimeUnit.MILLISECONDS);
+                     MBeanServerConnection connection = connector.getMBeanServerConnection();
+                     return new Jsr160ConnectionWrapper(connector, connection);
+                } catch (InterruptedException | ExecutionException | TimeoutException e) {
+                    LOG.info("Exception connecting JMXConnectorFactory url {} , Error: {}", url, e.getMessage());
+                    throw new ConnectException("Error connecting JMXConnectionFactory  " + url);
+                } finally {
+                    future.cancel(true);
```
soleger: ```cancel()``` returns a boolean and we should log this value to show whether the cancel was successful or not.

cgorantla: typically, it will return false since the task has completed normally.  I don't see much value in logging this since it could be misleading.

core/jmx/impl/src/main/java/org/opennms/netmgt/jmx/impl/connection/connectors/Jsr160ConnectionFactory.java
cgorantla: @mvrueden  Even in try/catch , In one case future returns a connector where as in the other it's already there.  So in order to keep the existing code intact and not create new issues, I m retaining the way it is.

core/jmx/impl/src/main/java/org/opennms/netmgt/jmx/impl/connection/connectors/Jsr160ConnectionFactory.java
soleger: We should definitely log whether the cancel() is successful or not if the get() times out. Maybe the cancel() should be moved inside the catch() stanza?

soleger: This has to be addressed. If we don't log the result of the cancel() and the cancels fail (for whatever reason) and these threads end up leaking file handles, we won't have any logging to tell us what is going on. I realize that it's a bit pedantic but I think it's also good defensive programming.

j-white: Sorry, I missed this. We'll add the missing log statements, there's no reason not to.

cgorantla: Added log statement in PR : https://github.com/OpenNMS/opennms/pull/1066

PR #1051, opened 2016-09-30
NMS-8760: Override shutdown strategy

cgorantla:
* JIRA: http://issues.opennms.org/browse/nms-8760


soleger:
This also needs to be done in the blueprint that fetches the Trapd SNMPv3 configs and might need to be done in the RPC contexts as well (if it's not addressed there).

soleger:
I'm going to go ahead and merge these changes so far since I think that these timeouts are causing some of the periodic failures in Bamboo.

PR #1053, opened 2016-10-03
NMS-8779: Installer script doesn't work with PostgreSQL 9.6

agalue:
* JIRA: http://issues.opennms.org/browse/NMS-8779
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1020

The "CREATE USER" statement used to have "CREATEDB CREATEUSER". CREATEUSER has to be replaced with CREATEROLE.

Now, neither of those flags are required for the opennms user because the database is created by the provided superuser, as you can see in the Migrator class. For this reason, they have been removed. OpenNMS works as expected.

The JDBC driver has been updated to avoid issues with the installer.

### DIFF
--- a/core/schema/src/main/java/org/opennms/core/schema/Migrator.java
+++ b/core/schema/src/main/java/org/opennms/core/schema/Migrator.java
@@ -334,7 +334,7 @@ public class Migrator {
         try {
             c = m_adminDataSource.getConnection();
             st = c.createStatement();
-            st.execute("CREATE USER " + migration.getDatabaseUser() + " WITH PASSWORD '" + migration.getDatabasePassword() + "' CREATEDB CREATEUSER");
+            st.execute("CREATE USER " + migration.getDatabaseUser() + " WITH PASSWORD '" + migration.getDatabasePassword() + "'");
         } catch (final SQLException e) {
             throw new MigrationException("an error occurred creating the OpenNMS user", e);
         } finally {
--- a/pom.xml
+++ b/pom.xml
@@ -1374,7 +1374,7 @@
     <paxExamVersion>4.8.0</paxExamVersion>
     <paxSwissboxVersion>1.8.2</paxSwissboxVersion>
     <protobufVersion>2.6.1</protobufVersion>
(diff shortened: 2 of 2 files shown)


soleger:
This should be a foundation-2016 PR, right? It already has the 1208 version so it should be seamless to upgrade to 1211.

agalue:
> On Oct 5, 2016, at 9:02 AM, Seth Leger <notifications@github.com> wrote:
> 
> This should be a foundation-2016 PR, right? It already has the 1208 version so it should be seamless to upgrade to 1211.


I don't see the required changes on the main pom.xml for Meridian 2016.1.3, so that means, a lot more changes are required to make this work on foundation-2016. That's the reason why I did this against develop. If you want me to spend time porting the full solution for JDBC/C3P0/Hickory (because that piece of change contains the proper handling of the JDBC/C3P0 versions across all the POMs and OSGi features), I can give it a try, but I would consider this enough for its own PR after merging this one.

Thoughts ?

agalue:
@soleger I just figured that I was wrong, and you're right. The C3P0/Hickory changes are available in foundation-2016 and Meridian 2016. So, we can rebase this against foundation-2016. I've heard that rebase is possible without closing and re-creating the PR, but I don't know how.

agalue:
I tried, but the commit log went crazy, maybe close this PR and re-create a new one against foundation-201 is better. Thoughts ?

soleger:
I created a jira/NMS-8779-foundation-2016 branch with the change cherry-picked back to foundation-2016. If it passes Bamboo, I'll merge it there and we'll let it auto-merge to develop.

agalue:
Thanks a lot!

RangerRick:
Oop, reviewed this before looking at the comments.  ;)  Looks good, but if it's happening in `foundation-2016` already, I'll just leave it alone.

soleger:
I've merged these changes into foundation-2016 but they'll have to be hand-merged into develop because of a slight conflict in pom.xml. I'm going to close this PR since we won't need it after the hand-merge.

PR #1054, opened 2016-10-03
NMS-7106: If a target node is rebooted the RRD/JRB files contains spikes because the sysUpTime check is not working.

agalue:
Maybe we should rename that issue because the sysUpTime detection was working. The problem is only related with thresholds.

* JIRA: http://issues.opennms.org/browse/NMS-7106
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1021

The sysUpTime check was working, and that was preventing the persistence of the current collection-set if a restart was detected. This is crucial for avoiding spikes. Now, the counters cache was not reinitialized which is the reason of this change.

The fix was tested on my lab and it works as expected.


RangerRick:
Looks alright to me.  Looking at the current code, it's hilariously doing nothing.  :/

PR #1057, opened 2016-10-04
NMS-8545: Apache Kafka JMX collections and graph definitions

soleger:
This PR contains JMX collections and graph definitions for all of the key Kafka metrics mentioned in the Kafka operations documentation:

https://kafka.apache.org/090/ops.html

* JIRA: http://issues.opennms.org/browse/NMS-8545
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1022

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/snmp-graph.properties.d/kafka-graphs.properties
+++ b/opennms-base-assembly/src/main/filtered/etc/snmp-graph.properties.d/kafka-graphs.properties
@@ -310,29 +310,35 @@ report.kafka.purgatoryFetch.command=--title="Purgatory Size: Fetch" \
  GPRINT:value:MAX:" Max \\: %8.2lf %s\\n"
 
 
-report.kafka.netProcAvgIdle.name=Network Processor Average Idle Percentage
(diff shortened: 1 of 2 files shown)


opennms-base-assembly/src/main/filtered/etc/snmp-graph.properties.d/kafka-graphs.properties
```diff
@@ -1,52 +1,583 @@
...
+ --vertical-label="Requests" \
+ DEF:value={rrd1}:purgatoryFetch:AVERAGE \
+ AREA:value#edd400 \
+ LINE2:value#c4a000:"Fetch requests" \
+ GPRINT:value:AVERAGE:" Avg \\: %8.2lf %s" \
+ GPRINT:value:MIN:" Min \\: %8.2lf %s" \
+ GPRINT:value:MAX:" Max \\: %8.2lf %s\\n"
+
+
+report.kafka.netProcAvgIdle.name=Network Processor Average Idle Percentage
```
soleger: I'm going to change this... who graphs an idle percentage? I'm going to invert it so that it's a usage percentage. That makes 100% more sense.

PR #1061, opened 2016-10-06
NMS-8767: Revert Discovery behavior back to create non-provisioned nodes

soleger:
These changes allow Discovery to generate newSuspect events without foreignSource parameters so that non-provisioned nodes can be created. It will also include an upgrade script that will convert any nodes in the "default" requisition into non-provisioned nodes.

* JIRA: http://issues.opennms.org/browse/NMS-8767
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1029

### DIFF
--- a/features/discovery/src/main/java/org/opennms/netmgt/discovery/actors/RangeChunker.java
+++ b/features/discovery/src/main/java/org/opennms/netmgt/discovery/actors/RangeChunker.java
@@ -112,8 +112,12 @@ public class RangeChunker {
         final double packetsPerSecond = (config.getPacketsPerSecond() > 0.0) ? config.getPacketsPerSecond() : DiscoveryConfigFactory.DEFAULT_PACKETS_PER_SECOND;
 
         // If the foreign source for the discovery config is not set than use 
-        // the default foreign source
-        final String foreignSourceFromConfig = (config.getForeignSource() == null || "".equals(config.getForeignSource().trim())) ? "default" : config.getForeignSource().trim();
+        // a value of null so that non-requisitioned nodes are created.
+        //
+        // TODO: Use the "default" foreign source instead so that we can move
+        // away from using non-requisitioned nodes.
+        //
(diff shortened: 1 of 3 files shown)


agalue:
Looks good to me. I'll give it try using the RPMs on a VM.

agalue:
It works!

I just want to mention that this fix requires an upgrade task to delete the default requisition, remove the content of foreignSource/foreignID from the node's table, and fix the names on the database using DNS/SNMP (in case the requisition was synchronized).

agalue:
On a clean VM with the RPMs from http://yum.opennms.org/branches/jira-NMS-8767/ I was able to verify that the old behavior is back. The discover is now sending newSuspect events without the default requisition, and the nodes are created properly with null on their foreignSource and foreignId.

Of course we have duplicates, but we can handle them more easily now.

soleger:
I'll commit an upgrade script for this but it is very tricky to write if anybody has a requisition called "default". I think that this should be a command-line utility or something that is only included with 18.0.3,4,5... that you can run manually if necessary (since, for instance, 19 will never have this problem).

agalue:
sounds reasonable, but I doubt that if someone has a requisition called default, they will have a foreignId equals to the nodeId if the requisition is manually created, as the probability to succeed on that is extremely low if not impossible. That's why "SELECT * FROM node WHERE foreignSource = 'default' AND cast(nodeId AS text) = foreignId" gives you exactly what you need to fix. What do you think ? I can work on the upgrade task. The only hard part that I would avoid for now is fixing the names, if they were changed, to avoid having an expensive task that requires DNS/SNMP access, but the upgrade task can warn about it, because in this case, the nodelabel will be equal to the primary IP address (which is also an exact SQL query).

agalue:
Despite the fact that affected users will have to start over with a fresh inventory (and maybe fresh data), I would say merge this, and document how to manually fix the problems on Jira (instead of doing it through an update task), as I think we need this ASAP.

PR #1064, opened 2016-10-06
HZN-755: Fixed resource graph redndering in info panels

fooker:
* JIRA: http://issues.opennms.org/browse/755
* Bamboo: http://bamboo.internal.opennms.com:8085/chain/result/viewChainResult.action?planKey=OPENNMS-ONMS1030&buildNumber=1


j-white:
Tested with both rrdgraph and Backshift. Thanks for update @fooker!

PR #1065, opened 2016-10-06
NMS-8790: The search page for events is not working as expected.

agalue:
* JIRA: http://issues.opennms.org/browse/NMS-8790

The solution was verified on a VM running Meridian 2016.1.3


soleger:
This is not targeted to the right branch, right? It needs to be foundation-2016.

agalue:
> On Oct 7, 2016, at 12:19 AM, Seth Leger <notifications@github.com> wrote:
> 
> This is not targeted to the right branch, right? It needs to be foundation-2016.


You're absolutely right. I'll close this one and create another one based on foundation-2016.

agalue:
Done! Check https://github.com/OpenNMS/opennms/pull/1067

PR #1066, opened 2016-10-07
NMS-8327: Add log for task completion

cgorantla:
* JIRA: http://issues.opennms.org/browse/NMS-8327
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1034


core/jmx/impl/src/main/java/org/opennms/netmgt/jmx/impl/connection/connectors/Jsr160ConnectionFactory.java
```diff
@@ -110,9 +110,9 @@ public JMXConnector call() throws IOException {
                      return new Jsr160ConnectionWrapper(connector, connection);
                 } catch (InterruptedException | ExecutionException | TimeoutException e) {
                     LOG.info("Exception connecting JMXConnectorFactory url {} , Error: {}", url, e.getMessage());
-                    throw new ConnectException("Error connecting JMXConnectionFactory  " + url);
-                } finally {
                     future.cancel(true);
+                    LOG.info(" Trying to close the connection, the task {}", future.isDone() ? "was completed" : "was not completed");
```
soleger: According to the javadocs, isDone() will always return true here:

> After this method returns, subsequent calls to isDone() will always return true. Subsequent calls to isCancelled() will always return true if this method returned true.

 I would check isCancelled() instead and say "the task was cancelled" or "the task could not be cancelled".

soleger:
Looks good to me!

PR #1069, opened 2016-10-10
NMS-8796: replace all - in resource names to _.

roskens:
NMS-8796: Fix PersistRegexSelectorStrategy by transforming all resource names when we add them to the context variable map.

This lets you use something like the following in a resourceType (where ns-dskPath is written as ns_dskPath):

``` xml
      <persistenceSelectorStrategy class="org.opennms.netmgt.collectd.PersistRegexSelectorStrategy">
        <parameter key="match-expression" value="not(#ns_dskPath matches '^(/dev/shm|/run|/sys/fs/cgroup|/tmp).*')" />
      </persistenceSelectorStrategy>
```
- JIRA: http://issues.opennms.org/browse/NMS-8796
- Bamboo: Do not worry, we add a link to our continuous integration system here

### DIFF
--- a/opennms-services/src/main/java/org/opennms/netmgt/collectd/PersistRegexSelectorStrategy.java
+++ b/opennms-services/src/main/java/org/opennms/netmgt/collectd/PersistRegexSelectorStrategy.java
@@ -71,7 +71,7 @@ public class PersistRegexSelectorStrategy implements PersistenceSelectorStrategy
         @Override
         public void visitAttribute(CollectionAttribute attribute) {
             if (StringAttributeType.supportsType(attribute.getType()))
-                context.setVariable(attribute.getName(), attribute.getStringValue());
+                context.setVariable(attribute.getName().replace('-','_'), attribute.getStringValue());
         }
 
         public StandardEvaluationContext getEvaluationContext() {


soleger:
Isn't the parsing failing because Spring's SPEL is interpreting the '-' as a minus sign? It seems like there's a better way to do this.

agalue:
Years ago, I've implemented a workaround for threshold expressions, as you can see here:

https://github.com/OpenNMS/opennms/blob/develop/opennms-services/src/test/java/org/opennms/netmgt/threshd/ExpressionConfigWrapperTest.java#L153

The reason for this was avoid implementing complex scripts to fix the RRDs and the graph templates if you change the alias of one or several MibObjs (for Newts it will be a lot more complex). I'm not sure if that is available for the PersistSelectorStrategy, but I would do something similar, or do something at runtime, but avoid at all cost changing configuration files.

j-white:
This works for '-', but we could run into similar problems with other special characters too. Applying the same workaround that @agalue referenced seems like a reasonable (and consistent) approach.

We could continue to put the attributes with their existing names in the context, and place an additional "attributes" map in the context that would support looking up the attributes by name.

jeffgdotorg:
We covered this PR in today's bug meeting and settled on the approach from @agalue since it feels cleaner and avoids potential problems with other special characters (for instance, somebody might use a dot in a metric name, which might be interpreted by JEXL as a field-access operator). Thanks for the proposal @roskens, I'll buy you a beer next time I'm in town :)

PR #1072, opened 2016-10-11
HZN-895: Fix the history when selecting layers and preserve breadcrumbs as well

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-895
* BAMBOO: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1038


pbrane:
Markus, is this a fix for history and bread crumbs?

mvrueden:
This only addresses the history.
But in order to restore the history correctly, the breadcrumbs should be also part of it.

PR #1073, opened 2016-10-11
HZN-858: Implement a Pinger RPC Module able to ping from local and remote locations

cgorantla:
* JIRA: http://issues.opennms.org/browse/hzn-858
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1036


cgorantla:
Integrating into topology is not part of this PR

PR #1075, opened 2016-10-12
HZN-316: Add location filter to node list in UI

christianpape:
JIRA: https://issues.opennms.org/browse/HZN-316
Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1040/latest


j-white:
Looks good. The only problem I noticed is that the name of the selection in the drop-down box gets trimmed, so "All locations" shows up as "All location".

I attached a screenshot in the associated JIRA issue.

PR #1076, opened 2016-10-13
HZN-621: Minion should run as non-root by default

RangerRick:
This PR makes it so the default Minion RPMs run as non-root, using non-privileged ports for syslogd and trapd.  It also includes documentation on configuring Linux to allow non-root ping, and RHEL7 documentation on port-forwarding.

This PR also includes Minion assembly changes that were originally made for `HZN-616` but that was put back into the backlog for now.

* JIRA: http://issues.opennms.org/browse/HZN-621
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1035


opennms-assemblies/minion/src/main/filtered/etc/org.opennms.minion.controller.cfg
```diff
@@ -0,0 +1,2 @@
+location=MINION
```
j-white: Should this file be here?

PR #1084, opened 2016-10-21
HZN-924: Create and use a date string for time zone in use by system running test

dschlenk:
Without this patch, executing features/poller/client-rpc/src/test/java/org/opennms/netmgt/poller/client/rpc/PollerResponseDTOTest.java in a timezone other than US eastern results in a failure

* JIRA: http://issues.opennms.org/browse/HZN-924
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1051


features/poller/client-rpc/src/test/java/org/opennms/netmgt/poller/client/rpc/PollerResponseDTOTest.java
```diff
@@ -45,12 +49,16 @@ public PollerResponseDTOTest(PollerResponseDTO sampleObject, String sampleXml) {
 
     @Parameters
     public static Collection<Object[]> data() throws Exception {
+        String localEpoch = ((new Date(0)).toInstant())
```
j-white: We do similar date manipulation in other XML tests too:
https://github.com/OpenNMS/opennms/blob/9dfa101190a0483f9edc0c83b4f4efceed99399e/opennms-alarms/api/src/test/java/org/opennms/netmgt/alarmd/api/NorthboundAlarmTest.java#L76

Can we make this method generic and move into the `org.opennms.core.test-api.xml` project instead?

Bonus points for refactoring the NorthboundAlarmTest to use it too :)

dschlenk: A fine plan. Will do.

dschlenk:
@j-white refactored!

j-white:
Thanks for the update, waiting on Bamboo, and we'll squash merge once it's green.

dschlenk:
after chat discussion, moving method to core/lib since there are other places ISO-8601 formatted date strings would be useful

PR #1088, opened 2016-10-25
HZN-930: Allow login via GET requests

fooker:
This allows the send a GET request to ```/opennms//j_spring_security_check?j_username=$MYUSERNAME&j_password=$MYPASSWORD``` and returns a logged in session.

This is useful for Auto-Login.

* JIRA: https://issues.opennms.org/browse/HZN-930
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1054


mvrueden:
Can you please provide a jira issue describing what this is all about.
Please also prefix each commit with the according jira issue number. 
This makes development more transparent and allows later on to figure out which issue was merged in which release version.

PR #1090, opened 2016-10-25
NMS-8846: update ticket ID and state via REST

dschlenk:
I'd be happy to update the documentation too but wasn't sure what version this'll make it into and wanted to call it out there. 

* JIRA: http://issues.opennms.org/browse/NMS-8846


dschlenk:
oh fart this rebase went poorly. Let's try this again in a new branch.

PR #1091, opened 2016-10-25
NMS-8846: update ticket ID and state via REST

dschlenk:
tiny patch that lets us set ticket ID and state on an alarm via REST. (Full use case: emit alarms via the JMS NBI that get processed into tickets, then use this to make a poor (wo)man's bidirectional ticket integration)
- JIRA: http://issues.opennms.org/browse/NMS-8846
- Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1072-1

### DIFF
--- a/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/AlarmRestService.java
+++ b/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/AlarmRestService.java
@@ -44,6 +44,8 @@ import javax.ws.rs.core.Response.Status;
 import javax.ws.rs.core.SecurityContext;
 import javax.ws.rs.core.UriInfo;
 
+import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang3.EnumUtils;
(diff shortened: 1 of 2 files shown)


j-white:
Looks reasonable, I pushed the changes into `pr/jira/NMS-8846`, we'll see what Bamboo thinks.

We should also try and update the documentation when adding new parameters to a REST endpoint, but none of the existing attributes are documented, so I guess it's alright if there's aren't documented either :)

dschlenk:
@j-white I did attempt to update the docs but then realized I didn't know what release to reference for the changes. This'll get into 19 it looks like? I'll write something up and do a separate PR.

j-white:
Yeah, this is targeted for 19.0.0. 

In the future, you can include the documentation updates in the same PR as the code changes, so they will end up in whichever release the code is merged.

dschlenk:
@j-white Will do. In the docs for the alarm endpoint there's sections for what changes have taken place in what version, so I was going to follow that pattern but didn't know if this would make it into 19 or not.

PR #1095, opened 2016-10-27
HZN-926: Deprecate the getAgentConfig without location

cgorantla:
HZN-926: Deprecate the SnmpAgentConfigFactory.getAgentConfig(InetAddress) method

* JIRA: http://issues.opennms.org/browse/HZN-926
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1055


opennms-provision/opennms-snmp-scanners/src/main/java/org/opennms/netmgt/provision/scan/snmp/AbstractSnmpScanner.java
```diff
@@ -113,11 +113,12 @@ protected void onInit() {
     @Override
     public void scan(ScanContext context) throws InterruptedException {
         InetAddress agentAddress = context.getAgentAddress("SNMP");
+        String location = context.getSysLocation();
```
j-white: The sysLocation (http://www.alvestrand.no/objectid/1.3.6.1.2.1.1.6.html) is not necessarily the same as the node's location.

PR #1102, opened 2016-11-03
NMS-8823: Outage ReST service forNode use case calculates dates incorrectly

agalue:
Outage ReST service forNode use case calculates dates incorrectly

* JIRA: http://issues.opennms.org/browse/NMS-8823


agalue:
It would be good to port this solution to foundation (for Meridian 2015). The reason why I didn't do it is because opennms-webapp-rest doesn't exist in foundation, besides other ReST changes.

j-white:
Looks good. Shouldn't we target `foundation-2016` with this one instead of `develop`?

agalue:
GitHub is not smart enough to figure out the base branch.

j-white:
FYI, you can change the branch now without needing to re-create the PR: https://github.com/blog/2224-change-the-base-branch-of-a-pull-request

forestfhb:
Know nothing!

PR #1103, opened 2016-11-04
NMS-8823: Outage ReST service forNode use case calculates dates incorrectly

agalue:
Outage ReST service forNode use case calculates dates incorrectly

* JIRA: http://issues.opennms.org/browse/NMS-8823
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1082


agalue:
It would be good to port this solution to foundation (for Meridian 2015). The reason why I didn't do it is because opennms-webapp-rest doesn't exist in foundation, besides other ReST changes.

PR #1106, opened 2016-11-04
NMS-8863: NonUniqueResultException when retrieving CDP links

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-8863
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1085

### DIFF
--- a/opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/CdpElementDaoHibernate.java
+++ b/opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/CdpElementDaoHibernate.java
@@ -28,8 +28,11 @@
 
 package org.opennms.netmgt.dao.hibernate;
 
+import java.util.List;
+
 import org.opennms.netmgt.dao.api.CdpElementDao;
 import org.opennms.netmgt.model.CdpElement;
+import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/CdpElementDaoHibernate.java
```diff
@@ -57,7 +60,12 @@ public CdpElement findByNodeId(Integer id) {
 
 	@Override
 	public CdpElement findByGlobalDeviceId(String deviceId) {
-        return findUnique("from CdpElement rec where rec.cdpGlobalDeviceId = ? ", deviceId);
+        List<CdpElement> elements = find("from CdpElement rec where rec.cdpGlobalDeviceId = ? ", deviceId);
+        if (elements.size() > 1) {
+            LoggerFactory.getLogger(getClass()).warn("Expected one CdpElement for device with id {} but found {}. Using CdpElement {} and ignoring others.", deviceId, elements.size(), elements.get(0));
+        }
+        return elements.isEmpty() ? null : elements.get(0);
```
j-white: I think we should make sure that the list of sorted by some criteria, so that we consistently choose the same element if/when there are many.

PR #1108, opened 2016-11-04
NMS-8671 - handle SNMP error-status > 5

RangerRick:
This PR changes the SNMP tracker code to handle handle all known `error-status` codes from [RFC 3416](https://tools.ietf.org/html/rfc3416).

For the most part it treats the write-related responses as fatal (throwing an exception) but attempts retry for most other errors, in the hopes they may be transient.

While I have tested this with a few simple (mostly Linux-based) devices, I'd love a second set of eyes to not only look through the patches, but give this branch a shot and confirm I don't do anything heinous.

There is probably more refactoring that could be done to reduce duplicate code between the `AggregateTracker`, `ColumnTracker`, and `SingleInstanceTracker` classes, but I didn't attempt more rearranging for the sake of keeping the changes small.  The only change that touches more code was changing the various hardcoded codes from `static int constant`s to a single `enum` that is referenced throughout.

* JIRA: http://issues.opennms.org/browse/NMS-8671

### DIFF
--- a/core/snmp/api/src/main/java/org/opennms/netmgt/snmp/ErrorStatus.java
+++ b/core/snmp/api/src/main/java/org/opennms/netmgt/snmp/ErrorStatus.java
@@ -1,0 +1,198 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2016 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2014 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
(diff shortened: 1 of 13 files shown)


RangerRick:
Packages are testable using `branches/jira-NMS-8671` in place of `stable` your debian or yum config.

core/snmp/api/src/main/java/org/opennms/netmgt/snmp/ErrorStatus.java
```diff
@@ -0,0 +1,170 @@
+package org.opennms.netmgt.snmp;
```
j-white: Missing copyright header.

core/snmp/api/src/main/java/org/opennms/netmgt/snmp/ErrorStatusException.java
```diff
@@ -0,0 +1,18 @@
+package org.opennms.netmgt.snmp;
```
j-white: Missing copyright header.

RangerRick:
`nms01` is running this code and isn't showing any SNMP-related errors.  I say ship it! :shipit:

PR #1116, opened 2016-11-09
NMS-8875: Topology is slow due to geocoding failures

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-8875

### DIFF
--- a/features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/info/CoordinateResolver.java
+++ b/features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/info/CoordinateResolver.java
@@ -75,9 +75,12 @@ public class CoordinateResolver {
 
     private final NodeDao nodeDao;
 
-    CoordinateResolver(GeocoderService geocoderService, NodeDao nodeDao) {
+    private final boolean resolveCoordinatesFromAddressString;
+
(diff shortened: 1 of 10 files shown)


features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/info/CoordinateResolver.java
```diff
@@ -98,22 +99,21 @@ Result resolve(Collection<Integer> nodeIds) {
...
-                    return null;
-                }));
-
-        // Convert nodesWithLongLat
+        // Retrieve coordinates for nodes with a adress, but not logitude/latitude
+        Map<Integer, Coordinates> nodeIdToCoordinateMapping = new HashMap<>();
+        for (OnmsNode eachNode : nodesWithAddress) {
+            final String addressString = geoLocation(eachNode).asAddressString();
+            try {
+                org.opennms.features.geocoder.Coordinates coordinates = geocoderService.getCoordinates(addressString);
```
j-white: The timeout will help reduce page load times, but won't eliminate the slowness all together, since we are still blocking the page while attempting to geocode each node in series.

Further, if this geocoding fails during the initial page load, it will still be reattempted every-time the view changes.

I think we need a better way of handling this.

mvrueden:
I added the possibility to disable the resolution of geolocations based on the address string if no long/lat information is available. In addition I documented the properties available to configure the Topology Map (and removed properties not useful to configure at all).

PR #1122, opened 2016-11-09
NMS-8879: Updating assets/categories through ReST affects Forced Unmanaged services.

agalue:
Updating assets/categories through ReST affects Forced Unmanaged services.

* JIRA: http://issues.opennms.org/browse/NMS-8879
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1105

The solution has been tested on a VM running Meridian 2016.

### DIFF
--- a/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/AssetRecordResource.java
+++ b/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/AssetRecordResource.java
@@ -129,7 +129,7 @@ public class AssetRecordResource extends OnmsRestService {
         }
         if (modified) {
             LOG.debug("updateAssetRecord: assetRecord {} updated", assetRecord);
-            m_assetRecordDao.update(assetRecord);
+            m_assetRecordDao.saveOrUpdate(assetRecord);
             try {
                 sendEvent(EventConstants.ASSET_INFO_CHANGED_EVENT_UEI, node.getId());
             } catch (EventProxyException e) {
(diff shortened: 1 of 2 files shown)


agalue:
Because the projects are different in `foundation` and `foundation-2016`, as soon as this PR is merged, I can cherry-pick it on `foundation` (to avoid GIT issues), and then manually apply the fix on the correct place, as I think this should be also fixed for `Meridian 2015`.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/AssetRecordResource.java
```diff
@@ -125,7 +129,7 @@ public Response updateAssetRecord(@PathParam("nodeCriteria") final String nodeCr
         }
         if (modified) {
             LOG.debug("updateAssetRecord: assetRecord {} updated", assetRecord);
-            m_nodeDao.saveOrUpdate(node);
+            m_assetRecordDao.update(assetRecord);
```
j-white: Is it possible that the node did not have already have an asset record? In this case we would need a save instead of an update.

agalue:
> On Nov 10, 2016, at 4:41 PM, Jesse White <notifications@github.com> wrote:
> Is it possible that the node did not have already have an asset record? In this case we would need a save instead of an update.
> 
If you take a look at the beginning of the updateAssetRecord, you'll see that an exception will be thrown (BAD_REQUEST) if there are no OnmsAssetRecord associated with the node. That means, the entity has to exist, so the usage of "update" is correct on this case.

j-white:
`getAssetRecord()` will never return `null` though, it will create a new object if one wasn't created previously: https://github.com/OpenNMS/opennms/blob/foundation-2016/opennms-model/src/main/java/org/opennms/netmgt/model/OnmsNode.java#L741

PR #1128, opened 2016-11-10
NMS-8883: handle duplicate IPs in separate locations cleanly

RangerRick:
This PR makes sure that we compare against location name when determining whether a node already exists upon a `newSuspect` call (specifically, `createUndiscoveredNode` in `Provisiond`).

* JIRA: http://issues.opennms.org/browse/NMS-8883


j-white:
:+1: from me, this should also be merged into `release-18.0.3`.

RangerRick:
Turns out 18.0.3 doesn't even have support for location in the `createUndiscoveredNode()` method, so it's a no-go.  Develop only.

PR #1129, opened 2016-11-11
NMS-8885: Fix checknsc NoClassDefFoundError for LoggerFactory

indigo423:
Add SLF4J API and SLF4J Log4J libraries to class path call.

* JIRA: http://issues.opennms.org/browse/NMS-8885


soleger:
So this patch will allow the checknsc code to use the log4j 1.X API via slf4j, is that correct? If so, the patch looks good to me.

RangerRick:
Agreed.  I'll merge it.

PR #1132, opened 2016-11-14
HZN-939: Refactor Discovery to use the Ping Sweep RPC module.

j-white:
I've refactored Discovery to use Ping Sweep RPC module in favor of the existing pattern so that it better aligns with all of the other modules we've written for Minion.

JIRA: https://issues.opennms.org/browse/HZN-939


features/discovery/src/test/java/org/opennms/netmgt/discovery/messages/DiscoveryJobTest.java
```diff
@@ -1,116 +0,0 @@
...
-    @Test
-    public void testTimeoutLargerThanIntegerMaxValue1() throws Exception {
-        List<IPPollRange> m_ranges = new ArrayList<IPPollRange>();
-        IPPollRange ipPollRange = new IPPollRange(null, null, "0.0.0.0", "89.0.0.0", 1, 0);
-        m_ranges.add(ipPollRange);
-        DiscoveryJob discoveryJob = new DiscoveryJob(m_ranges, "Bogus FS", "Bogus Location", Double.MAX_VALUE);
-        assertEquals(Integer.MAX_VALUE, discoveryJob.calculateTaskTimeout());
-    } 
-
-    @Test
```
j-white: I ported an analogue of the testTimeoutIntegerMaxValueMinus1() test (I haven't been able to come up with a range that gives exactly Long.MAX_VALUE - 1).

I'm having trouble understanding what the other tests verify that isn't already covered though.

soleger: The other tests just throw some values other than 1 into the timeout/retries values to make sure that the BigInteger multiplication is correct.

PR #1133, opened 2016-11-14
HZN-946: Allow the default RPC TTL to be increased

christianpape:
* JIRA: https://issues.opennms.org/browse/HZN-946
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1115

### DIFF
--- a/core/ipc/rpc/camel-impl/src/main/java/org/opennms/core/rpc/camel/CamelRpcClientPreProcessor.java
+++ b/core/ipc/rpc/camel-impl/src/main/java/org/opennms/core/rpc/camel/CamelRpcClientPreProcessor.java
@@ -33,8 +33,28 @@ import org.apache.camel.Processor;
 import org.opennms.core.camel.JmsQueueNameFactory;
 import org.opennms.core.rpc.api.RpcRequest;
 import org.opennms.core.rpc.api.RpcResponse;
+import org.opennms.core.utils.PropertiesUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class CamelRpcClientPreProcessor implements Processor {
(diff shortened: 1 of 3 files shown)


core/ipc/rpc/camel-impl/src/main/java/org/opennms/core/rpc/camel/CamelRpcClientPreProcessor.java
```diff
@@ -43,7 +63,7 @@ public void process(Exchange exchange) {
         final JmsQueueNameFactory queueNameFactory = new JmsQueueNameFactory(CamelRpcConstants.JMS_QUEUE_PREFIX,
                 wrapper.getModule().getId(), wrapper.getRequest().getLocation());
         exchange.getIn().setHeader(CamelRpcConstants.JMS_QUEUE_NAME_HEADER, queueNameFactory.getName());
-        exchange.getIn().setHeader(CamelRpcConstants.CAMEL_JMS_REQUEST_TIMEOUT_HEADER, wrapper.getRequest().getTimeToLiveMs());
+        exchange.getIn().setHeader(CamelRpcConstants.CAMEL_JMS_REQUEST_TIMEOUT_HEADER, CAMEL_JMS_REQUEST_TIMEOUT == 0 ? wrapper.getRequest().getTimeToLiveMs() : CAMEL_JMS_REQUEST_TIMEOUT);
```
j-white: If the request specifies a timeout (i.e. `getTimeToLiveMs() != null`)  we should use is instead of the value of `org.opennms.jms.timeout` property.

j-white: If `org.opennms.jms.timeout` is not set, and `getTimeToLiveMs()` is `null`, we can resort to using Camels's default which is 20 seconds.

PR #1137, opened 2016-11-16
HZN-940: Create upgrade task for updating 'localhost' to 'Default' in discovery-configuration.xml

christianpape:
* JIRA: https://issues.opennms.org/browse/HZN-940
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1120

### DIFF
--- a/core/upgrade/src/main/java/org/opennms/upgrade/implementations/DiscoveryConfigurationLocationMigratorOffline.java
+++ b/core/upgrade/src/main/java/org/opennms/upgrade/implementations/DiscoveryConfigurationLocationMigratorOffline.java
@@ -1,0 +1,120 @@
+package org.opennms.upgrade.implementations;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.StringWriter;
+import java.io.Writer;
+import java.nio.file.Paths;
+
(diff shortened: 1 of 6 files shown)


core/upgrade/src/main/java/org/opennms/upgrade/implementations/DiscoveryConfigurationLocationMigratorOffline.java
```diff
@@ -0,0 +1,120 @@
...
+
+    @Override
+    public String getDescription() {
+        return "Changes the name for the default location from 'localhost' to 'Default'. See HZN-940.";
+    }
+
+    @Override
+    public void preExecute() throws OnmsUpgradeException {
+        log("Backing up discovery-configuration.xml\n");
+        m_backupFile = zipFile(m_configFile);
```
mvrueden: Just for clarification: In our meeting yesterday I mentioned that `m_backupFile` is used in the `postExecute()` or `rollback()` phase and if `zipFile()` threw an exception we may run into a NPE. I verified today, that this can never occur, as if `preExecute()` fails, the execution is aborted.

j-white:
Looks good to me, just missing the copyright headers in the new source files.

PR #1138, opened 2016-11-16
HZN-936: Only create a transaction when necessary

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-936


features/events/daemon/src/main/java/org/opennms/netmgt/eventd/processor/expandable/EventTemplate.java
```diff
@@ -0,0 +1,137 @@
...
+        final String collect = tokens.stream()
+                .map(t -> t.expand(event, decode))
+                .collect(Collectors.joining());
+        return collect;
+    }
+
+    // If we find any token which requires a transaction, the template itself requires a transaction as well
+    @Override
+    public boolean requiresTransaction() {
+        return tokens.stream().filter(t -> t.requiresTransaction()).findAny().isPresent();
```
j-white: Can we pre-compute this value instead? The `tokens` are immutable and set in the constructor.

features/events/daemon/src/main/java/org/opennms/netmgt/eventd/processor/expandable/EventTemplate.java
```diff
@@ -0,0 +1,137 @@
...
+    private final List<ExpandableToken> tokens = Lists.newArrayList();
+
+    private final EventUtil eventUtil;
+
+    private final ExpandableParameterResolverRegistry parameterResolverRegistry;
+
+    public EventTemplate(String input, EventUtil eventUtil) {
+        this.input = Objects.requireNonNull(input);
+        this.eventUtil = Objects.requireNonNull(eventUtil);
+        this.parameterResolverRegistry = new ExpandableParameterResolverRegistry();
```
j-white: Can we re-use the `ExpandableParameterResolverRegistry` instead of re-creating it for every `EventTemplate`. We could even pull it from the `EventUtil` instance that's being passed instead of passing it as another argument here.

features/events/daemon/src/main/java/org/opennms/netmgt/eventd/processor/expandable/EventTemplate.java
```diff
@@ -0,0 +1,137 @@
...
+ * For each placeholder in the {@link EventTemplate#input} a {@link ExpandableToken} is created.
+ * The expandable state of the {@link EventTemplate} is the expanded state of all tokens ({@link EventTemplate#tokens}.
+ */
+public class EventTemplate implements ExpandableToken {
+
+    private static final Logger LOG = LoggerFactory.getLogger(EventTemplate.class);
+
+    private static final char PERCENT = '%';
+
+    private static final String WHITESPACE_PATTERN = ".*\\s(?s).*";
```
j-white: While we're modifying this code... can we precompile this regex instead of re-compiling it on every call to `parm.matches`?

features/events/daemon/src/main/java/org/opennms/netmgt/eventd/AbstractEventUtil.java
```diff
@@ -371,239 +369,19 @@ public static String escape(String inStr, char delimchar) {
...
-			retParmVal = "";
-			if (event.getNodeid() > 0) {
-				try {
-					String foreignSource = getForeignSource(event.getNodeid());
-					if (foreignSource != null) {
-						retParmVal = WebSecurityUtils.sanitizeString(foreignSource);
-					}
-				} catch (SQLException e) {
-					// do nothing
+	private final LoadingCache<String, EventTemplate> eventTemplateCache = CacheBuilder.newBuilder()
```
j-white: Multiple instance of this bean are created in the runtime (i.e. https://github.com/OpenNMS/opennms/blob/develop/opennms-services/src/main/resources/META-INF/opennms/applicationContext-notifd.xml#L65) so each of these will end up with their own cache.

Can we find a way to de-duplicate the caches, or ideally, de-duplicate these beans?

soleger:
+1, looks good to me.

PR #1147, opened 2016-11-18
NMS-8708: Make sure Trapd config isn't reloaded unnecessarily

soleger:
Added equals()/hashCode() and Map comparison to make sure that the collection isn't reloaded when it has not changed. There are also some additional tests for the new code.

* JIRA: http://issues.opennms.org/browse/NMS-8708
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1136

### DIFF
--- a/core/snmp/api/src/main/java/org/opennms/netmgt/snmp/SnmpV3User.java
+++ b/core/snmp/api/src/main/java/org/opennms/netmgt/snmp/SnmpV3User.java
@@ -131,4 +131,16 @@ public class SnmpV3User {
             Objects.equals(getPrivPassPhrase(), that.getPrivPassPhrase()) &&
             Objects.equals(getPrivProtocol(), that.getPrivProtocol());
     }
+
+    @Override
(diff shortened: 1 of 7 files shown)


core/snmp/api/src/main/java/org/opennms/netmgt/snmp/SnmpV3User.java
```diff
@@ -106,4 +108,39 @@ public void setPrivProtocol(String privacyProtocol) {
...
+
+    @Override
+    public boolean equals(final Object o) {
+        if (o == null) {
+            return false;
+        }
+        if (o == this) {
+            return true;
+        }
+        if (this.getClass() != o.getClass()) {
```
mvrueden: Can we use `instanceof` instead?
Usually getClass() should not be a problem, but can cause issues when a class is proxied or when subclassed.

j-white:
+1 from me

PR #1150, opened 2016-11-22
NMS-8900: Spurious Trapd error logs: unknown SNMP protocol version 'v2'

jeffgdotorg:
* JIRA: http://issues.opennms.org/browse/NMS-8900

- Consider "v2" and "v2c" equivalent when accounting for traps by version
- Update Trapd integration test to verify v2 traps received as well as v1


j-white:
Shouldn't this be targeted for `foundation-20`6` instead?

jeffgdotorg:
Tried rebasing to foundation-2016 but it went sideways. Eff it, will close and recreate.

PR #1151, opened 2016-11-22
NMS-8803, HZN-811: Load Minion listeners from featuresBoot, fix zkclient

soleger:
Wrapped blueprints for syslog and trap listeners in JARs so that they can be loaded from featuresBoot. Fixed missing zkclient dependency from camel-karaf.

* JIRA: http://issues.opennms.org/browse/NMS-8803
* JIRA: http://issues.opennms.org/browse/HZN-811
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1142

### DIFF
--- a/container/features/src/main/resources/features.xml
+++ b/container/features/src/main/resources/features.xml
@@ -802,6 +802,8 @@
 
       <feature>dropwizard-metrics</feature>
       <feature>lmax-disruptor</feature>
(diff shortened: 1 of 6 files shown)


j-white:
We should also update the relevant section in the install guide: http://docs.opennms.org/opennms/branches/develop/guide-install/guide-install.html#_configure_syslog

soleger:
I'm opening the docs tweak as another issue.

PR #1152, opened 2016-11-22
NMS-8900: Spurious Trapd error logs: unknown SNMP protocol version 'v2'

jeffgdotorg:
* JIRA: http://issues.opennms.org/browse/NMS-8900

- Consider "v2" and "v2c" equivalent when accounting for traps by version
- Update Trapd integration test to verify v2 traps received as well as v1pull/1150


j-white:
+1 from me, waiting on Bamboo

PR #1154, opened 2016-11-23
NMS-8907: Invalid graph templates

agalue:
Invalid graph templates

* JIRA: http://issues.opennms.org/browse/NMS-8907
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1154

The changes have been verified with a custom script attached to the Jira
issue that executes a "rrdtool graph" command using the templates.

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/snmp-graph.properties.d/akcp-graph.properties
+++ b/opennms-base-assembly/src/main/filtered/etc/snmp-graph.properties.d/akcp-graph.properties
@@ -256,7 +256,7 @@ report.akcp.securityProbe.airflow.command=--title="(Index: {secAirflowIdx}) Airf
  DEF:hcrit={rrd3}:secAirflowHCrit:AVERAGE \
  DEF:lwarn={rrd4}:secAirflowLWarn:AVERAGE \
  DEF:lcrit={rrd5}:secAirflowLCrit:AVERAGE \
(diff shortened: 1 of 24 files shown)


j-white:
Should this be targeted for `foundation-2016` instead of `develop`?

agalue:
Not all the templates have been fixed. The remaining ones are some kind of "placeholders" (i.e. graph templates without actual graph content, just "comments"), others related to big DS (i.e. more than 19 characters, which also requires changes on datacollection-groups), and some templates with VDEF entries.

The reason why I'm ignoring the last ones is because we're supposed to migrate VDEF related templates to use {diffTime} as VDEF is not supported by backshift.

That being said, I think it is ok to merge this to fix the current broken templates.

pbrane:
> On Nov 24, 2016, at 10:05 AM, Alejandro Galue <notifications@github.com> wrote:
> 
> The reason why I'm ignoring the last ones is because we're supposed to migrate VDEF related templates to use {diffTime} as VDEF is not supported by backshift.

I think that has been fixed in the past week.  I’m not sure if that fix is only an H19 thing.


:D

agalue:
> On Nov 24, 2016, at 11:34 AM, David Hustace <notifications@github.com> wrote:
> 
> I think that has been fixed in the past week. I’m not sure if that fix is only an H19 thing.


That’s exactly why I didn’t touch those templates.

PR #1160, opened 2016-11-28
HZN-933: Fix transaction-per-EventLog inside HibernateEventWriter

soleger:
The logging changes in 19ee8497d5cacc4343c2ba5ce6f2dc59645ca6e1 busted the @Transactional behavior in HibernateEventWriter because the annotation moved to a private method. This fixes the problem by using TransactionOperations.

* JIRA: http://issues.opennms.org/browse/HZN-933
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1158

### DIFF
--- a/features/events/daemon/src/main/java/org/opennms/netmgt/eventd/processor/HibernateEventWriter.java
+++ b/features/events/daemon/src/main/java/org/opennms/netmgt/eventd/processor/HibernateEventWriter.java
@@ -32,6 +32,7 @@ import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
 import java.util.Objects;
+import java.util.concurrent.atomic.AtomicReference;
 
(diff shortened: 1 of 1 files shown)


j-white:
The Aspectj proxy allows private methods to be transactional: https://github.com/OpenNMS/opennms/blob/develop/features/events/daemon/src/main/resources/META-INF/opennms/applicationContext-eventDaemon.xml#L21

j-white:
I double checked, and the transactional annotation is not being applied here, `org.springframework.transaction.support.TransactionSynchronizationManager.isActualTransactionActive()` returns false inside the `processInTransaction()` call.

j-white:
Cherry-picked to release-19.0.0 and merged into develop.

soleger:
Thanks!

PR #1162, opened 2016-11-28
NMS-8903: Better logging for SNMP traps with processing problems

jeffgdotorg:
* JIRA: http://issues.opennms.org/browse/NMS-8903

- Add toString() methods to TrapInformation subclasses
- Add helpful logging when runtime exception caught during trap processing


core/snmp/api/src/main/java/org/opennms/netmgt/snmp/TrapInformation.java
```diff
@@ -109,9 +111,15 @@ protected void processTrap() {
...
-        for (int i = 0; i < getPduLength(); i++) {
-            processVarBindAt(i);
-        } // end for loop
+        int i = 0;
+        try {
+            for (i = 0; i < getPduLength(); i++) {
+                processVarBindAt(i);
+            } // end for loop
+        } catch (RuntimeException rte) {
+            LOG.error("Caught {} while processing varbind {} of trap; rethrowing. Trap info: {}", rte.getClass(), i+1, toString());
```
soleger: I would change this to a ```catch(Throwable)``` and also log the Throwable's stack trace since it seems like this sort of exception should be highly unexpected.

PR #1164, opened 2016-11-29
HZN-315: Add system/location column to event list in UI

christianpape:
* JIRA: https://issues.opennms.org/browse/HZN-315
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1149


mvrueden:
I tried this and it works 👍

PR #1166, opened 2016-11-29
NMS-8490: Use updated version of backshift.js

fooker:
* JIRA: http://issues.opennms.org/browse/NMS-8490


j-white:
This should also be merged into `foundation-2016`.

j-white:
I tried adding a VDEF to the `OpenNMS Queued Updates` graph definition:
```
report.onms.queued.updates.name=OpenNMS Queued Updates
report.onms.queued.updates.columns=ONMSQueUpdates
report.onms.queued.updates.type=interfaceSnmp
report.onms.queued.updates.command=--title="OpenNMS Queued Updates" \
 --vertical-label="Updates per second" \
 DEF:updates={rrd1}:ONMSQueUpdates:AVERAGE \
 LINE1:updates#0000ff:"Completed Updates" \
 VDEF:totupdates=updates,TOTAL \
 GPRINT:updates:AVERAGE:" Avg  \\: %8.2lf %s" \
 GPRINT:totupdates:AVERAGE:"Total \\: %8.2lf %s" \
 GPRINT:updates:MAX:"Max  \\: %8.2lf %s\\n"
```

JRobin renders the graph correctly, but Backshift fails with:
```
backshift.onms.min.js:2 Uncaught TypeError: Cannot read property '0' of undefined(…)
(anonymous function) @ backshift.onms.min.js:2
consolidate@ backshift.onms.min.js:2
updateValues @ backshift.onms.min.js:2
...
```

fooker:
According to the rrdtool documentation, the depricated form of (G)PRINT (the one with a correlation function) must specify a DEF or a VDEV. In opposite, the new version (without a correlation function) must use a CDEF.

I would like to interpret this realy strict. Therefore I propose to add an error message instead of supporting this.

I will check the behavior of rrdtool on this tomorrow and report back.

-- 


-----Original Message-----

fooker:
Switch VDEF and CDEF on the above gibberish...

-- 


-----Original Message-----

fooker:
rrdtool fails in the above example:
```
ERROR: Unused Arguments ""Total \\: %8.2lf %s""
```

For reference: https://oss.oetiker.ch/rrdtool/doc/rrdgraph_graph.en.html
>  It is an error to specify a vname generated by a DEF or CDEF.

j-white:
Cool. I just tried that since we use a similar pattern in some of the existing graphs with VDEF statements, so we'll need to revisit those.

We also need some better handling, but that might go beyond the scope of this issue. We already have an issue logged here: https://issues.opennms.org/browse/BSHIFT-10

fooker:
I don't think that we need to rework the existing graphs for now. They all use the deprecated, but still working, version of GPRINT by specifying a DEF/CDEF and a correlation function.

Only the few ones outlined in the the issue needs to be checked, but I don't have an environment for that. Maybe @agalue can help out on that?

soleger:
I went ahead and merged the Backshift update with VDEF support. If there are any further problems, we can track those as new issues.

PR #1169, opened 2016-11-29
NMS-8915: ArrayIndexOutOfBoundsException while processing some trap varbinds

jeffgdotorg:
* JIRA: http://issues.opennms.org/browse/NMS-8915

- Update SyntaxToEvent to catch ArrayIndexOutOfBoundsException when checking
  whether a sequence is printable UTF-8
- Update SyntaxToEventTest to provide coverage of values which are not
  valid UTF-8 sequences


opennms-model/src/main/java/org/opennms/netmgt/model/events/snmp/SyntaxToEvent.java
```diff
@@ -111,7 +111,13 @@ public static Parm processSyntax(final String name, final SnmpValue value) {
             if (m_syntaxToEvents[i].getTypeId() == -1 || m_syntaxToEvents[i].getTypeId() == value.getType()) {
                 val.setType(m_syntaxToEvents[i].getType());
                 String encoding = null;
-                if (value.isDisplayable()) {
+                boolean displayable = false;
+                try {
+                    displayable = value.isDisplayable();
+                } catch (ArrayIndexOutOfBoundsException aioobe) {
+                    // Eat it
```
soleger: Can you put a comment here that says "This should not be necessary when NMS-7547 is fixed"?

jeffgdotorg: Done.

PR #557, opened 2016-01-31
NMS-5069: Monitor disk space on more Net-SNMP devices and include in threshold package

TimMillerDyck:
Include Net-SNMP (OID 1.3.6.1.4.1.8072) devices (such as Linux sytems) in the hrstorage threshold package and collect mib2-host-resources-storage data for Net-SNMP devices as this is a more reliable way of getting disk space data than using the net-snmp-disk MIB.

This change allows disk space to be monitored successfully on more devices. Examples are given in the JIRA ticket.

JIRA: http://issues.opennms.org/browse/NMS-5069

PR #1172, opened 2016-11-30
NMS-5069:  Documentation for Collectd and SNMP collector

StCyr:
This is an attempt to fill the need for documenting the change in NMS-5069.

I've created a basic structure for documenting Collectd and the various OpenNMS Collectors. If it's okay for you, I'll continue documenting the "SNMP Collection" concept, so that, in the end, it'll document the "old Net-SNMP agent && huge filesystems" issue.

See also https://github.com/OpenNMS/opennms/pull/557

* JIRA: https://issues.opennms.org/browse/NMS-5069
* Bamboo:


indigo423:
@StCyr I'm checking out your branch and assigned myself for the review to get it merged. Thanks a lot for your contribution.

StCyr:
Great, thanks.

Just remember, the goal of this pr was to document some weird behaviour that may occur when collecting hrStorage snmp data from devices with very large filesystem (as explained in NMS-5069).

Also, I would like to document VMware data collection while I'm busy with it. I'll create another pr therefore.

Cheers,

Cyrille

indigo423:
@StCyr I can help you with the VMware data collection, would be awesome to get it into the official docs. We just wrote the paper about it.

StCyr:
Yeah I read it.

My idea was to copy it for most parts.

But atm my vmware datacollection can't work because my local user is
"opennms@vsphere.local" and the vmare:// URI in requistion doesn't support
"@" in username. And, I'm trying to fix this problem.

That's why I was asking how I should do to add parameters to config file. I
would like to turn this (which doesn't work):

  <requisition-def import-name="vmware-requisition"

import-url-resource="vmware://opennms%40vsphere.local:xxxxxxx@vcenter.xxxxxxx
/VCenterImport?key=OpenNMSImport;value=yes">
    <cron-schedule>0 18 12 * * ? *</cron-schedule>
  </requisition-def>

to this:

  <requisition-def import-name="vmware-requisition"
      import-url-resource="vmware://
vcenter.fw.belnet.be/VCenterImport?key=OpenNMSImport;value=yes"
username="opennms@vsphere.local" password="xxxxxxxxx">
    <cron-schedule>0 18 12 * * ? *</cron-schedule>
  </requisition-def>


What do you think?

Cyrille

2016-12-08 9:48 GMT+01:00 Ronny Trommer <notifications@github.com>:

indigo423:
@StCyr we should move the vmware topic to a different thread :)

I've converted some images from the OpenNMS book which are related to this piece of documentation. We need to reference them in text. Do they make sense for you?

StCyr:
Images make sense to me

pbrane:
> On Dec 8, 2016, at 3:56 AM, Cyrille Bollu <notifications@github.com> wrote:
> 
> But atm my vmware datacollection can't work because my local user is
> "opennms@vsphere.local" and the vmare:// URI in requistion doesn't support
> "@" in username. And, I'm trying to fix this problem.

Is this a problem with vmware accepting the URL or is it a problem with our marshaling of the configuration and conversion to a proper URI for vmware?  If it is the latter, I would prefer to fix the marshaling.

StCyr:
Hi @pbrane 

We (@indigo and I) have moved this vmware converstation on the chat.

A good place to discuss this is also the https://issues.opennms.org/browse/NMS-6530 issue which is a symptom of this problem. 

Can we continue discuss over there? I'd love to have your comments.

The problem, I think, is indeed with the marshaling. But, there doesn't seem to be an easy way to fix it.

indigo423:
Cleaned up and reopened under correct issue number #1193

PR #1176, opened 2016-12-01
HZN-931: only reinitialize SNMPv3 users on update

RangerRick:
This PR makes it so the trapdConfig SNMPv3 user update code *only* updates the SNMPv3 user list, rather than consuming all of the `TrapdConfiguration`.  This keeps the Minion from attempting to reinitialize the SNMP listener with the hostname and port coming from the OpenNMS host, rather than its own config.

* JIRA: http://issues.opennms.org/browse/HZN-931


j-white:
+1 from me, waiting on Bamboo

PR #1180, opened 2016-12-05
NMS-8907: Fix invalid graph templates (Second Phase)

agalue:
Fix invalid graph templates.

* JIRA: http://issues.opennms.org/browse/NMS-8907

The only pending templates to fix are the big DS and invalid DS, as that requires changes on *datacollection-config.xml files.

Remaining errors:

* ERROR(fortinet.fgApFTPStatsEntry.connection): Invalid DS fgApFTPMaxConnections (it has more than 19 characters)
* ERROR(fortinet.fgWebChDskStsEntry): Invalid DS fgWebCacheDiskMisses (it has more than 19 characters)
* ERROR(netapp.diskio): Invalid DS naMiscLowDiskReadBytes (it has more than 19 characters)

A new PR against release-19.0.0 will be created to fix the invalid DS once this one is merged.

All the errors fixed so far have been verified against latest RRDtool 1.5.x. If someone notice errors when using JRobin or Backshift, then those libraries have to be fixed to be 100% compliant with RRDtool.

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/snmp-graph.properties.d/aix-graph.properties
+++ b/opennms-base-assembly/src/main/filtered/etc/snmp-graph.properties.d/aix-graph.properties
@@ -5,8 +5,7 @@
 ##
 ##################################################
 
-reports=aix.teaser, \
-aix.cpuUtilization, \
+reports=aix.cpuUtilization, \
 aix.procNum, \
 aix.subsystems, \
 aix.subservers, \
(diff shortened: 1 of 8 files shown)


agalue:
I can't find this branch on Bamboo. I think it is because I've re-used the name of this branch. I mean, this is the second PR with the same branch name (i.e. jira/NMS-8907)

j-white:
I went ahead and re-added the branch in Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1189-1

PR #1181, opened 2016-12-06
NMS-4552: Make OpenNMS able to auto-compile jasper subreports

StCyr:
This is a port of the patch provided in NMS-4452  to make OpenNMS be able to also automaticaly compile Jasper subreports at run-time (rather than only the main Jasper reports).

Pay attention that it seems the original patch wouldn't work with computed subreportExpression parameters
(See https://issues.opennms.org/browse/NMS-4552?focusedCommentId=29503&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-29503).

Though, a workaround to this later issue was proposed (see https://issues.opennms.org/browse/NMS-4552?focusedCommentId=29510&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-29510)

Note that this commit compiles against the release-19.0.0 branch here.

Cyrille

* JIRA: https://issues.opennms.org/browse/NMS-4552

### DIFF
--- a/features/reporting/jasper-reports/src/main/java/org/opennms/reporting/jasperreports/svclayer/JasperReportService.java
+++ b/features/reporting/jasper-reports/src/main/java/org/opennms/reporting/jasperreports/svclayer/JasperReportService.java
@@ -66,6 +66,7 @@ import net.sf.jasperreports.engine.JRExporterParameter;
 import net.sf.jasperreports.engine.JRParameter;
 import net.sf.jasperreports.engine.JRPropertiesUtil;
 import net.sf.jasperreports.engine.JRReport;
+import net.sf.jasperreports.engine.JRSubreport;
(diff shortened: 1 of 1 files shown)


mvrueden:
I cannot remember, but there was an issue with compiling the subreports on the fly. I am going to test this and may stumble upon the problem.

mvrueden:
I had a look at this PR and as mentioned above the issue with auto compiling subreports is that we need to recursively check for subreports and compile them if they have not yet been compiled. In addition we also have to check each already compiled subreport if that report may contain a not yet compiled report. 

Besides that the `lastModified` check would result in always compiling the subreports, as the source file and the compiled file never have the same modified time. 

Closing this PR in favor of PR https://github.com/OpenNMS/opennms/pull/1229 which addresses the above issues and also uses a more complex subreport expression.

PR #1182, opened 2016-12-06
HZN-937: Introduce TopologyService abstraction to better control access to the underlying (Meta)TopologyProviders

mvrueden:
This issue addresses some problems with the Topology Map architecture. The main issue is within the "MergingGraphProvider" which is now gone, as mixing multiple namespaces is not supported anymore. In addition to that, only "one" (instead of number of sessions) now can trigger a refresh/reload of the underlying Topology Provider. This is intended to be a first step, not the final solution,  as  there is still plenty of work to be done.

* JIRA: https://issues.opennms.org/browse/HZN-937
* BAMBOO: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1186


mvrueden:
I encountered a weird `java.lang.UnsupportedOperationException: Internal error: In use HttpService is not an WebContainer (from Pax Web)` exception when starting OpenNMS after I merged with develop. Marking as `WIP` and `Do NOT merge` until I can figure out what is causing this issue.

features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/service/DefaultTopologyService.java
```diff
@@ -0,0 +1,239 @@
+
```
j-white: Copyright header.

features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/service/DefaultTopologyService.java
```diff
@@ -0,0 +1,239 @@
...
+                .newBuilder()
+                .expireAfterWrite(Math.max(0, cacheTimeout), TimeUnit.SECONDS)
+                // Due to graphml we may end up having a ton of graph providers.
+                // Caching is limited to 50, which is still plenty
+                .maximumSize(50)
+                .build(new CacheLoader<GraphProviderKey, GraphProvider>() {
+                    @Override
+                    public GraphProvider load(GraphProviderKey key) throws Exception {
+                        final MetaTopologyProvider metaTopologyProvider = getMetaTopologyProvider(key.getMetaTopologyId());
+                        metaTopologyProvider.reload(key.getNamespace());
```
j-white: How do we handle concurrency now that the MetaTopologyProvider instances are shared?

Looking at the `GraphMLMetaTopologyProvider`, everything is cleared in the first part of the reload() function. Which means that anyone accessing the graph while it's being reloaded will end up without any edges/vertices, or existing session may end up in some inconsistent state.

mvrueden: This issue was already there before the refactoring, but it is now more visible. The problem here is that it is not possible to either cache the complete graph (e.g. there is no `TopologyProvider.getGraph()` method) or the TopologyProvider cannot be instantiated/cloned to have another copy of it. Maybe we can create a `TopologyProviderDelegate` in order to achieve this. But to address this properly we have to change the `TopologyProvider` API, which in the end is probably something we want to do anyways, but will be much out of scope for the changes here.

features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/service/DefaultTopologyService.java
mvrueden: Yep. I finally found out why they are sometimes missing (-:

features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/service/DefaultTopologyService.java
mvrueden: The `LoadingCache` blocks if a key is looked up while the value of a key is already resolved. I changed the behaviour to only load/refresh the `GraphProvider` and not the whole `MetaTopologyProvider` to avoid conurrency problems. 

> If another call to get(K) or getUnchecked(K) is currently loading the value for key, simply waits for that thread to finish and returns its loaded value. Note that multiple threads can concurrently load values for distinct keys

See https://google.github.io/guava/releases/19.0/api/docs/com/google/common/cache/LoadingCache.html#get(K) for more details

mvrueden:
I added some more changes and fixed some issues I encountered while playing with it.
The tests should hopefully now pass.
The following issues still exist:
 * When initially on the Topology Map, the viewport is not calculated correctly, resulting in a weird "center on map" or "show entire map" behaviour
 * When uninstalling providers (via osgi) switching to another (not the one removed) provider nothing happens, but the menu updates, with no provider selected. Before the refactorings an exception was thrown forcing the user to re-login. With the "new behaviour" we may end up having an invalid ui state. This needs some more love.

mvrueden:
I addressed the issues as described above.

In addition I was able to show a popup window if the selected topology provider is no longer available, which is with graphml more likely than it was before. 
The behaviour is, that if the current topology provider is no longer available to fall back to the default one (enhanced linkd). If it is not available, try any other. If no provider is available, invalidate the user's session and have her log out and log in again, as we cannot do anything anymore for her.

j-white:
After adding a business service and clicking the "View in Topology" link, I was prompted with a "No Focus Defined" message. The business service is properly displayed when clicking this same link 30 seconds later.

I realize that this is due the topology cache, but I feel like should improve scenario. If I see a link for "View in Topology" I expect to see the element when I click the link and not a "No Focus Defined" message.

mvrueden:
Yeah, I stumbled upon the same thing. 
Maybe we can add some server side code to allow invalidating the cache. A force reload from the user's perspective would neglect the changes made here.

mvrueden:
I added an event listener to allow listening for either "bsm daemon reload successfully" or "reload topology" events. This should allow internal components to force a refresh if needed.

features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/service/TopologyServiceEventListener.java
```diff
@@ -0,0 +1,85 @@
...
+        if (e.getUei().equals(EventConstants.RELOAD_DAEMON_CONFIG_SUCCESSFUL_UEI)) {
+            String daemonName = EventUtils.getParm(e, EventConstants.PARM_DAEMON_NAME);
+            if (daemonName != null && "bsmd".equalsIgnoreCase(daemonName)) {
+                topologyService.invalidate("bsm");
+            }
+        }
+    }
+
+    public void init() {
+        eventIpcManager.addEventListener(this);
```
j-white: We should set the collection of UEIs we're interested in, instead of a getting a callback for every single event.

mvrueden:
@j-white was able to trigger a NPE. See http://paste.opennms.eu/?05965175742b8a65#x+4kmYw9a33XpfW0/0jfhBf+NTagg8ihTHneZtWINuA= for more details.
How to reproduce: 
 * Fresh install of HZN-937
 * Go to the BSM Admin Page
 * Create a new BSM
 * Reload
 * View in Topology 
 * Boom

j-white:
Everything is working well for me now. :+1:

PR #1184, opened 2016-12-07
HZN-314: Add location column to outage list in UI

christianpape:
* JIRA: https://issues.opennms.org/browse/HZN-314
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1187


soleger:
Can you add some basic tests to DaoWebOutageRepositoryIT that use the new location sorting? Otherwise it looks good.

I did notice that you didn't add any Hibernate aliases for {{outage.node.location}} but it also looks like the outage web code is still using JDBC. Is that correct?

christianpape:
Added some tests. The outage web stuff still uses JDBC, the few Hibernate-based methods already had an alias for the node stuff.

PR #1187, opened 2016-12-07
NMS-8671: SNMP `error-status` Bug Fixes and Enhancements

RangerRick:
This PR adds a few bug fixes and enhancements on top of the previous work done for [NMS-8671](https://issues.opennms.org/browse/NMS-8671).

* Change any `SET*`-related errors to be non-fatal, but to have 0 retries.  They are basically always going to be invalid in our case since OpenNMS doesn't perform any SETs, but we will generally want to continue without bailing to see if we can get more data despite a non-compliant response from an agent.
* Add `reportFatalErr()` and `reportNonFatalErr()` log methods to the `CollectionTracker` API so that all logging happens on the thread that is processing the request.
* Change the `ErrorStatus` handling to be configurable (sane defaults, but now it is possible to override the global SNMP client behavior to handle `error-status` responses in a different way).
* Add a "Special Cases and Workarounds" section to the documentation, and document the `error-status` customization there.

* JIRA: http://issues.opennms.org/browse/NMS-8671

The default SNMP error-status handlers perform as follows:
    
| error-status            | Fatal? | Retry? |
|---|---|---|
| noError(0)              | false  | false  |
| tooBig(1)               | false  | true   |
| noSuchName(2)           | false  | true   |
| badValue(3)             | false  | false  |
| readOnly(4)             | false  | false  |
| genErr(5)               | false  | true   |
| noAccess(6)             | false  | true   |
| wrongType(7)            | false  | false  |
| wrongLength(8)          | false  | false  |
| wrongEncoding(9)        | false  | false  |
| wrongValue(10)          | false  | false  |
| noCreation(11)          | false  | false  |
| inconsistentValue(12)   | false  | false  |
| resourceUnavailable(13) | false  | false  |
| commitFailed(14)        | false  | false  |
| undoFailed(15)          | false  | false  |
| authorizationError(16)  | false  | true   |
| notWritable(17)         | false  | false  |
| inconsistentName(18)    | false  | false  |
   
You can override this behavior by setting a property inside `${OPENNMS_HOME}/etc/opennms.properties` in the form:
    
```properties
org.opennms.netmgt.snmp.errorStatus.[statusCode].[type]
```
    
For example, to make authorizationError(16) throw an exception and not retry, you would set:

```properties
org.opennms.netmgt.snmp.errorStatus.16.fatal=true
org.opennms.netmgt.snmp.errorStatus.16.retry=false
```

### DIFF
--- a/core/snmp/api/src/main/java/org/opennms/netmgt/snmp/AggregateTracker.java
+++ b/core/snmp/api/src/main/java/org/opennms/netmgt/snmp/AggregateTracker.java
@@ -32,9 +32,6 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 public class AggregateTracker extends CollectionTracker {
     private static final class ChildTrackerPduBuilder extends PduBuilder {
         private List<SnmpObjId> m_oids = new ArrayList<SnmpObjId>();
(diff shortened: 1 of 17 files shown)


RangerRick:
@jeffgdotorg is this alright?  Should I go ahead and merge it?

PR #1188, opened 2016-12-07
NMS-8936: Identify time-series strategy in support/about page

jeffgdotorg:
* JIRA: http://issues.opennms.org/browse/NMS-8936


core/lib/src/main/java/org/opennms/core/utils/TimeSeries.java
```diff
@@ -27,6 +27,20 @@
         public String getName() {
             return m_name;
         }
+
+        public String getDescr() {
+            switch (m_name) {
```
j-white: This switch can be avoided by passing the description directly in the constructor i.e.:

    RRD(RRD_TIME_SERIES_STRATEGY_NAME, "RRD (RRDTool or JRobin)"),

jeffgdotorg: Thanks @j-white. Suggested change pushed, re-running branch in Bamboo.

PR #1194, opened 2016-12-08
NMS-6530: Fix how vmware integration handles user credentials

StCyr:
This is work in progress to let @mvrueden have a look at what I'm doing.

I still need to fix vmwarereqtool

* JIRA: https://issues.opennms.org/browse/NMS-6530


StCyr:
I've now updated vmwarereqtool.

The change supports passing user credentials via URL query parameters (like in 'vmware://vcenter.domain.tld/VCenterImport?_OpenNMSImport=true;username=opennms@vsphere.local;password=xxxxxxxx') but doesn't support the old way of passing user credentials (via the URL's UserInfo section, as in 'vmware://username:password@vcenter.fw.belnet.be/VCenterImport?_OpenNMSImport=true')

Is it OK for you?

Cyrille

mvrueden:
Can you also fix the exception message here: https://github.com/StCyr/opennms/blob/fa8224f87def3751ea4543f198f4e422dcfc445d/integrations/opennms-vmware/src/main/java/org/opennms/netmgt/provision/service/vmware/VmwareRequisitionUrlConnection.java#L208

integrations/opennms-vmware/src/main/java/org/opennms/netmgt/provision/service/vmware/VmwareRequisitionUrlConnection.java
```diff
@@ -130,11 +130,22 @@ public VmwareRequisitionUrlConnection(URL url) throws MalformedURLException, Rem
...
 
+        // Old or new user credentials handling scheme
+	if (url.getUserInfo() != null && !url.getUserInfo().isEmpty()) {
+            logger.warn("Old user credentials handling scheme detected. I'm gonna use it but you'd better adapt your URL to the new query parameter scheme 'vmware://<vcenter_server_fqdn>?username=<username>;password=<password>;....'");
+            m_username = getUsername();
+            m_password = getPassword();
+        } else {
+            m_username = m_args.get("username");
+            m_password = m_args.get("password");
+        }
```
mvrueden: Can we also add a warning if both methods are used?

StCyr: Well, there's already one: 

```
Logger.warn("Old user credentials handling scheme detected. I'm gonna use it but you'd better adapt your URL to the new query parameter scheme 'vmware://<vcenter_server_fqdn>?username=<username>;password=<password>;....'");
```

It will be printed whenever the old user credentials handling scheme is detected. So, also when both the old and the new schemes are specified.

I personaly don't see the necessity to add another test and log 

Cyrille

mvrueden: You are the user, if the log message is an enough indicator for you, that is fine for me.

j-white:
@mvrueden is everything good here now?

If this is backwards compatible (which it appears to be), I think we should re-target this for `foundation-2016`.

StCyr:
I tried rebasing my branch on `foundation-2016`. And, after lots of conflict solved with `git checkout --theirs`, here's the result

Apparently I did something wrong :-p

Can you advice on the next steps? 

Thanks

mvrueden:
I'll have a look at it (-:

StCyr:
At first sight, `git diff ^35ccfe1f92725abc7993a365cdd4cc345ac544de` looks OK

mvrueden:
Cherry-picked the changes from cyrille's branch and created a new branch in opennms. See PR https://github.com/OpenNMS/opennms/pull/1259. Closing this one.

PR #1195, opened 2016-12-08
NMS-8942: Preserve column order in measurements API

fooker:
* JIRA: http://issues.opennms.org/browse/NMS-8942
* Bamboo: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1198


fooker:
This type of fix will never work. There is no way to have an order between sources and expressions in the current request structure.

PR #1196, opened 2016-12-08
NMS-6287: [RFC] - Use the same format for confs as opennms writes

luizluca:
Files written by opennms has a different format than
the format that comes from etc-pristine. This generates
unecessary configuration conflits on each upgrade.

This pull request simply let OpenNMS write a bunch of config files and use this result as pristine configuration.

BTW, what is the correct branch to use?

I guess some problems are not only on pristine files but also on opennms xml writer (like the cases where xml header got removed).

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/collectd-configuration.xml
+++ b/opennms-base-assembly/src/main/filtered/etc/collectd-configuration.xml
@@ -1,148 +1,128 @@
-<?xml version="1.0"?>
-<?castor class-name="org.opennms.netmgt.collectd.CollectdConfiguration"?>
-<collectd-configuration
-        threads="50">
-
-    <package name="cassandra21x">
-        <filter><![CDATA[(IPADDR != '0.0.0.0') & (categoryName == 'Cassandra21x')]]></filter>
-        <service name="JMX-Cassandra" interval="300000" user-defined="false" status="on">
-            <parameter key="port" value="7199"/>
-            <parameter key="retry" value="2"/>
-            <parameter key="timeout" value="3000"/>
-            <parameter key="protocol" value="rmi"/>
-            <parameter key="urlPath" value="/jmxrmi"/>
-            <parameter key="rrd-base-name" value="cassandra21x"/>
-            <parameter key="ds-name" value="cassandra21x"/>
-            <parameter key="friendly-name" value="cassandra21x"/>
-            <parameter key="collection" value="cassandra21x"/>
-            <parameter key="thresholding-enabled" value="true"/>
-            <parameter key="factory" value="PASSWORD-CLEAR"/>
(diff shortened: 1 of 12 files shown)


opennms-bamboo:
Welcome @luizluca.

I could not find a signed OpenNMS Contributor Agreement (OCA) assigned to the username @luizluca.

luizluca:
BTW, are all "<created>Thursday, December 8, 2016 6:54:51 PM GMT</created>" really necessary?

indigo423:
@luizluca Is it possible to retarget your pull request against branch "develop", which is the next major release version. To get your changes accepted is possible for you to sign our OCA which you can find here: https://github.com/OpenNMS/opennms/blob/develop/CONTRIBUTING.md

Thank you very much for your contribution.

luizluca:
The OCA was already resolved. What is the best place to discuss? Jira? Maillist? Github?
The PR conflicts with some collectd and poller changes. I don't know if it is better to just skip them, redo all my work on a develop snapshot. (BTW, https://wiki.opennms.org/wiki/Download is inconsistent and I could not find a binary snapshot)

Maybe instead of trying to fix the config, we can avoid it to become a problem.

@indigo423 , on JIRA I detailed each configuration result after editing on OpenNMS. I have some doubts about what would be the expected XML output, like if it needs to include XML header, castor, etc. It seems that some XML uses a different codepath for building XML. "There can be only one"

It would be nice to have all XML writers to use a common codepath and an external tool that would let devs reformat an XML just like OpenNMS would. Something similar to "xmllint --format" but using OpenNMS conventions (that does not seems to exist). This tool could even become a standard test, avoiding commits that changes pristine configuration which would be reformated by OpenNMS.

So, the steps to fix it definitely that I could see are:
1. What is the correct XML format to use? Why are files written differently?
2. Uniform all file written by OpenNMS to the same format
3. Create a simple tool that reformats XML file and reformat all configs
4. Add an test that tries to reformat all confs and fails if they change.

indigo423:
@luizluca the opennms-discuss mailinglist is the best place to discuss such a topic. I know @agalue spend already some time in doing configuration normalisation. The steps you described are totally right, I can also see something to built into git or bamboo which automatically normalise the XML files with git hooks on commit or Bamboo has a step for XML file normalisation so anyone can do whatever he wants, but as soon it gets committed it gets normalised automatically.

I'm totally with you to get this topic addressed cause it makes configuration merges unnecessarily hard.

j-white:
I think we all agree that we should address this, and we already have some code that will help with the normalization. That said, I'm going to close this PR. We are going to warn users about the upcoming configuration file changes in the release notes for 20.0.0 (see NMS-9401) and make the changes in 21.0.0 (see NMS-6287).

PR #1199, opened 2016-12-13
NMS-8726: Optimizations uncovered with Minion syslog test

soleger:
These optimizations were uncovered while testing the Minion syslog code.

* JIRA: http://issues.opennms.org/browse/NMS-8726


j-white:
Wouldn't a `com.google.common.cache.LoadingCache` be more appropriate instead of the `ConcurrentHashMap`s?

PR #1204, opened 2016-12-15
HZN-962: Sink API Updates

j-white:
JIRA: https://issues.opennms.org/browse/HZN-962

Here we update and refactor parts of the Sink API, renaming `MessageProducer` to `MessageDispatcher` and adding built-in support for aggregation and asynchronous dispatching.

This forms the foundation for moving Syslogd and Trapd over to use the Sink API.


core/ipc/sink/common/src/main/java/org/opennms/core/ipc/sink/common/AsyncDispatcherImpl.java
```diff
@@ -0,0 +1,163 @@
...
+     * Blocks calls to {@link LinkedBlockingQueue#offer(Object)}, converting
+     * these to calls to {@link LinkedBlockingQueue#put(Object)}.
+     *
+     * This is used to block the calling thread when trying to execute
+     * tasks when the queue is already full.
+     *
+     * Make sure that corePoolSize == maximumPoolSize if using
+     * this in an ExecutorService.
+     */
+    private static class OfferBlockingQueue<E> extends LinkedBlockingQueue<E> {
```
soleger: I've been reading that it's better to use a RejectedExecutionHandler that does the put() instead of changing the semantics of the queue. Then you don't have to worry about what corePoolSize is set to.

j-white: That method also has some drawbacks though, some of which are noted here: http://stackoverflow.com/a/3518588

I'm not against changing it, but it seems that we would be swapping some caveats for others.

soleger: Correction: you do have to worry what corePoolSize is set to with the RejectedExecutionHandler (it must be non-zero or the pool could shrink to zero and deadlock). I don't really care about the implementation but it's probably a good idea to paste the stackexchange links into the code that explain the decision:

Here's the one that I saw that rebuts the link you provided:

http://stackoverflow.com/a/32123535

j-white: Makes sense. I added some more comments and referenced both of these links.

PR #1209, opened 2016-12-19
HZN-983: Make the 'hawtio-offline' feature available on the default repository for Minion

j-white:
JIRA: https://issues.opennms.org/browse/HZN-983

This patch allows includes Hawtio in the default repository, which some in handy for debugging.

We do not enable it by default in the RPMs.

It can be installed by running `feature:install hawtio-offline` in the Karaf shell (or adding it to featuresBoot.d).

### DIFF
--- a/features/minion/repository/pom.xml
+++ b/features/minion/repository/pom.xml
@@ -11,6 +11,10 @@
     <packaging>pom</packaging>
     <name>OpenNMS :: Features :: Minion :: Default Repository</name>
 
+    <properties>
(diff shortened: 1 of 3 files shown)


soleger:
This does increase the size of the RPMs from 83MB to 109MB.

j-white:
Worth every byte.

If we want to reduce the size of the RPMs in the future, we can split them out, but we'll probably want to look into addressing [HZN-624](https://issues.opennms.org/browse/HZN-624) first.

PR #1210, opened 2016-12-19
NMS-8962: Skip /var/lock/subsys actions when running as unprivileged user

jeffgdotorg:
* JIRA: http://issues.opennms.org/browse/NMS-8962

Should be zero-impact, but review requested since start/stop is pretty visible and a bit fragile. Bamboo pending.


j-white:
Should this be targeted for `release-19.0.0`?

jeffgdotorg:
Bamboo is green.

PR #1211, opened 2016-12-20
HZN-966: Update Syslogd to the the Sink API for dispatching messages

j-white:
JIRA: https://issues.opennms.org/browse/HZN-966


features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/ByteBufferXmlAdapter.java
```diff
@@ -28,13 +28,20 @@
...
+public class ByteBufferXmlAdapter extends XmlAdapter<byte[], ByteBuffer> {
+
+    @Override
+    public ByteBuffer unmarshal(byte[] bytes) throws Exception {
+        return ByteBuffer.wrap(bytes);
+    }
+
+    @Override
+    public byte[] marshal(ByteBuffer bb) throws Exception {
+        return bb.array();
```
soleger: This call relies on optional behavior of ByteBuffer. It's safer to use the original code:
```
    @XmlAttribute
    public byte[] getBytes() {
        m_bytes.rewind();
        byte[] retval = new byte[m_bytes.remaining()];
        m_bytes.get(retval);
        m_bytes.rewind();
        return retval;
    }
```
If you want to preserve the position of the buffer, you'll have to add code for that (but it's probably not a problem here).

features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/SyslogSinkConsumer.java
```diff
@@ -0,0 +1,210 @@
...
+                                && (parm.getValue().getContent() != null)) {
+                            LOG.trace("    ({}, {})", parm.getParmName().trim(), parm.getValue().getContent().trim());
+                        }
+                    }
+                    LOG.trace("  }");
+                }
+                LOG.trace("}");
+            }
+        }
+        eventForwarder.sendNowSync(eventLog);
```
soleger: Can you explain to me why sendNowSync() is necessary here? Why is it necessary for Syslogd's code (and no other event producer) to specify the semantics of the Eventd call? Is it not sufficient to give Eventd a bounded event processing queue with blockOnFull behavior?

features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/SyslogReceiverNioDisruptorImpl.java
```diff
@@ -1,426 +0,0 @@
-/*******************************************************************************
```
j-white: Unlike the JavaNet implementation, there is currently no way of using this one on either Minion (no blueprint) or OpenNMS (the class name isn't configurable) without code changes. It would also require significant refactoring to work with the Sink API.

That said, I don't see the point in maintaining it currently.

If we want to integrate it properly at some point in the future, we can reference the code available in any of the 18.x tags.

features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/SyslogSinkConsumer.java
j-white: Unlike other event producers, the messages handled by this consumer can live in an external broker.
We send synchronously in order to perform as much processing as possible on the current thread before grabbing another message.

soleger: In the case of a vanilla OpenNMS instance, the messages will be routed without an external broker. In this case, it is my understanding that the only buffering that will occur is at the kernel level in the socket buffers and then inside Netty's worker queue. We don't currently expose configuration for the Netty queue, is that correct?

j-white: Buffering is handled by the Sink API.

From the Netty worker thread, we dispatch the message to a queue managed by the Sink API. This queue is configurable in size and blocks when full. It is served by a thread pool with a configurable number of threads.

When running in OpenNMS, those threads invoke the SyslogSinkConsumer#handleMessage method directly. When running in Minion, those threads publish the message to the broker.

PR #1213, opened 2016-12-20
NMS-8957: Alarmd: Only start transactions for events with alarm data

soleger:
This should make Alarmd's event listener more efficient.

* JIRA: http://issues.opennms.org/browse/NMS-8957

### DIFF
--- a/opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
+++ b/opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
@@ -40,6 +40,9 @@ import org.opennms.netmgt.xml.event.Event;
 import org.opennms.netmgt.xml.event.UpdateField;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.transaction.TransactionStatus;
+import org.springframework.transaction.support.TransactionCallback;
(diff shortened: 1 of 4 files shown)


opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
```diff
@@ -62,28 +84,39 @@ public OnmsAlarm persist(Event event) {
...
-    
+
         String reductionKey = event.getAlarmData().getReductionKey();
         LOG.debug("addOrReduceEventAsAlarm: looking for existing reduction key: {}", reductionKey);
         OnmsAlarm alarm = m_alarmDao.findByReductionKey(reductionKey);
 
         EventBuilder ebldr = null;
         if (alarm == null) {
-            LOG.debug("addOrReduceEventAsAlarm: reductionKey:{} not found, instantiating new alarm", reductionKey);
+            if (LOG.isDebugEnabled()) {
```
j-white: What's the point of wrapping log messages like these with an extra conditional?

soleger: I was changing all of the debug() calls to avoid the overhead of creating the arguments but in this particular log message, there is no extra overhead since the value is an existing variable. I guess I was overzealous. :)

PR #1217, opened 2016-12-29
NMS-8756: Exception in LatencyStoringServiceMonitorAdaptor

rssntn67:
Changed the Hibernate Query to use
equals instead of LIKE

* JIRA: http://issues.opennms.org/browse/NMS-8756


rssntn67:
Hey Ranger, I have moved the like hibernate to = do you think is it right?

I have a test env with 20 nodes and the fix seems fine.

PR #1218, opened 2017-01-03
NMS-8944: Statsd aggegators not reset between workers

jm-k:
Aggregators are declared as 'enum's : this makes them static and so shared between all statsd worker threads.

This PR declares them as a class intances ;...

JIRA: https://issues.opennms.org/browse/NMS-8944

### DIFF
--- a/opennms-dao/src/main/java/org/opennms/netmgt/dao/support/RrdStatisticAttributeVisitor.java
+++ b/opennms-dao/src/main/java/org/opennms/netmgt/dao/support/RrdStatisticAttributeVisitor.java
@@ -65,65 +65,95 @@ public class RrdStatisticAttributeVisitor implements AttributeVisitor, Initializ
         double getValue();
         void aggregate(final double v);
     }
-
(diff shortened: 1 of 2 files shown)


j-white:
@fooker can you review this? This appears to be related to the changes you made in https://github.com/OpenNMS/opennms/pull/427.

j-white:
Also, this fix should be targeted for the `foundation-2016` branch.

fooker:
The solution seems to be valid. Only remark is, that one could use `java.util.function.Supplier<Aggregator>`  and lambda statements instead of the custom `EnumAggregator` interface.

What about tests?

fooker:
I will take this over and write an unit test. Hope it's ok for you, @jm-k.

fooker:
Unit test added.

fooker:
This should go to foundation-2016, right?

j-white:
Yeah, we should retarget this for `foundation-2016`.

fooker:
OK, I will rebase.

PR #1220, opened 2017-01-03
NMS-8972: Make the RRAs tunable for the metrics:stress command when using for RRDtool.

agalue:
Make the RRAs tunable for the metrics:stress command when using for
RRDtool.

As a bonus, improve the name of the metric names to be unique, and avoid
issues on ds.properties when testing RRDtool.

* JIRA: http://issues.opennms.org/browse/NMS-8972


j-white:
:+1:

PR #1224, opened 2017-01-05
HZN-963: Add Kafka support to the Sink IPC

j-white:
JIRA: https://issues.opennms.org/browse/HZN-963


core/test-api/kafka/src/main/java/org/opennms/core/test/kafka/JUnitKafkaServer.java
```diff
@@ -0,0 +1,171 @@
...
+        while (current.get() < max) {
+            try (final ServerSocket socket = new ServerSocket(current.get())) {
+                return socket.getLocalPort();
+            } catch (final Throwable e) {}
+            current.incrementAndGet();
+        }
+        throw new IllegalStateException("Can't find an available network port");
+    }
+
+    private static String getLocalhost() {
```
soleger: Can we use ```InetAddressUtils.getLocalHostAddressAsString()``` here? I don't recall it ever returning link-local addresses.

j-white: Maybe @RangerRick can elaborate here.

RangerRick: I was specifically matching the way that Apache Curator sets up localhost because I was getting weird mismatches running the tests on my machine using the InetAddressUtils version.

PR #1226, opened 2017-01-07
HZN-980: Add 'blockWhenFull' configuration property to Eventd

soleger:
When a queueSize is specified in Eventd and this blockWhenFull property is set to 'true', Eventd will throttle incoming events by blocking to avoid discarding events or exhausting memory.

* JIRA: http://issues.opennms.org/browse/HZN-980

### DIFF
--- a/features/events/daemon/src/main/java/org/opennms/netmgt/eventd/EventIpcManagerDefaultImpl.java
+++ b/features/events/daemon/src/main/java/org/opennms/netmgt/eventd/EventIpcManagerDefaultImpl.java
@@ -103,7 +103,7 @@ public class EventIpcManagerDefaultImpl implements EventIpcManager, EventIpcBroa
                 if (!super.offer(event)) {
                     // If the queue is full, run on current thread
                     LOG.warn("Execution queue is full, executing Runnable on current thread");
-                    event.run();
+                    Logging.withPrefix(Eventd.LOG4J_CATEGORY, event);
                 }
                 return true;
             } else {
@@ -255,7 +255,11 @@ public class EventIpcManagerDefaultImpl implements EventIpcManager, EventIpcBroa
                         // Make sure we restore our log4j logging prefix after onEvent is called
                         Map<String,String> mdc = Logging.getCopyOfContextMap();
                         try {
-                            m_listener.onEvent(event);
+                            // Synchronize on the listener because the onEvent() methods are not
+                            // always threadsafe
+                            synchronized(m_listener) {
+                                m_listener.onEvent(event);
+                            }
                         } finally {
                             Logging.setContextMap(mdc);
                         }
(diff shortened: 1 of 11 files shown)


j-white:
With 'blockWhenFull' enabled in eventd, it appears that the following could occur:

1. Possible deadlocks

    If an event listener is also an event producer (i.e. alarmd), you could end up in situations where the event listener is blocked while trying to send an event.

2. A single "slow" event listener can cause unrelated event producers to block

    For example, if the Elasticsearch event listener can't keep up with the stream of events, all of Eventd's threads will end up being blocked, and other services such as a Pollerd could block when sending nodeDown events.

soleger:
1. I think I've solved the deadlock issue with the second commit.

2. Without this configurable change, the behavior of a slow event listener is to eventually exhaust the heap and trigger OutOfMemoryErrors (when there is no queueSize set) or to discard the latest events that are received by the listener (when there is a queueSize set but blockOnFull is false). I think that the blockWhenFull behavior will be desirable when event durability is a primary concern. It will block event producers (maybe "throttle" is a more accurate term) but the alternative is unpredictable loss of events in the listeners.

features/events/daemon/src/main/java/org/opennms/netmgt/eventd/EventIpcManagerDefaultImpl.java
```diff
@@ -71,10 +73,52 @@
...
+
+        @Override
+        public boolean offer(Runnable event) {
+            // If we're inside an event listener...
+            if (EVENT_LISTENER_DEPTH.get().intValue() > 0) {
+                // Attempt to offer
+                if (!super.offer(event)) {
+                    // If the queue is full, run on current thread
+                    LOG.warn("Execution queue is full, executing Runnable on current thread");
+                    event.run();
```
j-white: The logging prefix needs to be set before running this.

j-white:
With the solution to 1), it is now possible for multiple threads to be calling a single event listener, but not all of event listeners are thread safe.

features/events/daemon/src/main/java/org/opennms/netmgt/eventd/EventIpcManagerDefaultImpl.java
soleger: I assume that you mean to the Eventd prefix so that the log ends up in eventd.log.

j-white: Yeah, the prefix also needs to be set for the event listeners. See https://github.com/OpenNMS/opennms/blob/ea6fc2e7981cd3a49f2c3395e2d2ee3412fa9689/features/events/daemon/src/main/java/org/opennms/netmgt/eventd/EventIpcManagerDefaultImpl.java#L232.

features/events/daemon/src/main/java/org/opennms/netmgt/eventd/EventIpcManagerDefaultImpl.java
```diff
@@ -255,7 +255,11 @@ public void run() {
                         // Make sure we restore our log4j logging prefix after onEvent is called
                         Map<String,String> mdc = Logging.getCopyOfContextMap();
                         try {
-                            m_listener.onEvent(event);
+                            // Synchronize on the listener because the onEvent() methods are not
+                            // always threadsafe
+                            synchronized(m_listener) {
```
j-white: Won't we end up in the same deadlock situation with this?

soleger:
Added deadlock tests to release-19.0.0, closing this PR.

PR #1229, opened 2017-01-10
NMS-4552: Auto-compile jasper subreports if needed

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-4552

@StCyr originally created PR https://github.com/OpenNMS/opennms/pull/1181 to address this issue. The provided solution did not compile all subreports (e.g. if a subreport contains another subreport). In addition it required the subreport expression to be a simple string, but it may be something like `$P{SUBREPORT_DIR} + "/subreports/"`. Besides that the patch did not auto-compile recursively, or considered sub-reports of already compiled subreports.

This PR addresses the above limitations, adds documentation and fixes a broken report.


opennms-base-assembly/pom.xml
```diff
@@ -93,6 +93,10 @@
           </execution>
         </executions>
       </plugin>
+      <!--
```
StCyr: While running "assemble.pl" I stumbled upon the following log lines:

```
INFO] Compiling report file: subreports/ResponseTimeSummary_subreport.jrxml
[INFO] Compiling report file: InterfaceAvailabilityReport.jrxml
[INFO] Compiling report file: ResponseTimeCharts.jrxml
[INFO] Compiling report file: DiskUsageForCTX.jrxml
[INFO] Compiling report file: subreports/PeakTraffic_subreport.jrxml
[INFO] Compiling report file: subreports/Top25PercentDown_subreport.jrxml
[INFO] Compiling report file: ResponseTimeSummary.jrxml
[INFO] Compiling report file: AvailabilitySummary.jrxml
[INFO] Compiling report file: subreports/ResponseTimeSummary_Response_Offenders_subreport.jrxml
[INFO] Compiling report file: subreports/DiskUsageForCTXServers_subreport1.jrxml
[INFO] Compiling report file: subreports/ResponseTimeSummary_Availability_subreport.jrxml
[INFO] Compiling report file: Early-Morning-Report.jrxml
[INFO] Compiling report file: subreports/NodeId_to_NodeLabel_subreport.jrxml
[INFO] Compiling report file: NodeAvailabilityReport.jrxml
[INFO] Compiling report file: subreports/ResponseTime_subreport1.jrxml
[INFO] Compiling report file: subreports/AvailabilitySummaryChart_subreport.jrxml
[INFO] Compiling report file: subreports/ResponseTimeSummary_Availability_Offenders_subreport.jrxml
[INFO] Compiling report file: sample-report.jrxml
[INFO] Compiling report file: ResponseTime.jrxml
[INFO] Compiling report file: AssetManagementMaintStrategy.jrxml
[INFO] Compiling report file: subreports/95thPercentileTrafficRate_subreport.jrxml
[INFO] Compiling report file: SerialInterfaceUtilizationSummary.jrxml
[INFO] Compiling report file: TotalBytesTransferredByInterface.jrxml
[INFO] Compiling report file: AssetManagementMaintExpired.jrxml
[INFO] Compiling report file: subreports/InterfaceAvailabilityReport_subreport1.jrxml
[INFO] Compiling report file: TopIOWait.jrxml
[INFO] Compiling report file: AveragePeakTrafficRates.jrxml
[INFO] Compiling report file: subreports/TotalBytesTransferredByInterface_subreport1.jrxml
[INFO] Compiling report file: SnmpInterfaceOperAvailabilityReport.jrxml
[INFO] Compiled 30 report design files.
```

It seems jasper reports are still compiled

mvrueden: Yes, they are still compiled to ensure that the reports are compileable. This is a build step only.

StCyr: at least they seem to be added to the .deb files

```
root@opennms-dev:/home/cyr/jicmp6# ls -l /etc/opennms/report-templates/
total 2488
-rw-r--r-- 1 support root     73028 janv. 11 11:35 AssetManagementMaintExpired.jasper
-rw-r--r-- 1 support root     27931 nov.   3  2015 AssetManagementMaintExpired.jrxml
-rw-r--r-- 1 support root     61276 janv. 11 11:35 AssetManagementMaintStrategy.jasper
-rw-r--r-- 1 support root     16381 nov.   3  2015 AssetManagementMaintStrategy.jrxml
drwxr-xr-x 4 support root      4096 janv. 10  2013 assets
-rw-r--r-- 1 support root     70075 janv. 11 11:35 AvailabilitySummary.jasper
-rw-r--r-- 1 support root     32195 nov.   3  2015 AvailabilitySummary.jrxml
```

j-white:
If we modify a report template in a new release, do we have to worry about deleting any related `.jasper` files?

mvrueden:
Yes, if it is an already compiled subreport.

j-white:
How do you suggest we handle that?

One option could be to delete all .jasper files on upgrade as part of the packaging scripts, but then we may risk deleting user's custom reports.

mvrueden:
For now I'd suggest always to ship the jasper and the jrxml file. This way it is a consistent setup. Users can then still fiddle with reports and have them automatically recompile by simply deleting the according jasper files.

StCyr:
@mvrueden, Getting back to your comment about my LastModified() test: 

```
if(!compiledSubreport.exists() || compiledSubreport.lastModified() < sourceSubreport.lastModified()) {
			        LOG.info(sourceSubreportName + " needs to be compiled.");
			     JasperCompileManager.compileReportToFile(sourceSubreportName,compiledSubreportName);
			        LOG.info(sourceSubreportName + " compiled.");
                            }
```

I don't understand why you said it won't work.

I mean, if you modify the source .jrxml file, its mtime will be updated and thus will be greater than the corresponding .jasper file's mtime. The test would thus succeed.

mvrueden:
Yesterday, in my mind it made total sense, why it would not work. But today it does not anymore. It should work, you are right.

StCyr:
Hello Markus,

I've tested your branch.

It works, but I've identified the following issues:

1. The "master" report is compiled everytime it is used; To the contrary of what's done with subreports, there's no test to check if a compiled (.jasper) version of the master report already exists ;
2. The compiled master report is not saved on th filesystem: We're missing some `JRSaver.saveObject` before the `return report` statement in `JasperReportService.java::compileReport()` (line 693);
3. I believe it would be better to use the LastModified() test so that users are not required to do the extra step of deleting the compiled report for their changes in the source report to be taken into account;
4. There's a bug in line 655 in `JasperReportService.java::compileSubreportsRecursively()`. It should be written `LOG.debug("Subreport '{}' compiled", compiledSubreportName);`.

Cyrille

mvrueden:
First and second was the original way and I did not change it. In my opinion it is the better approach and with subreports it should be done the same, but Jasper does not support it.

I agree with 3 and 4

StCyr:
Ah? Strange... It must have changed in version 19; I'm pretty sure .jasper files are also created for master reports in version 18

mvrueden:
Yes they were and will always be created. But they where never used by the report engine. In general the report engine is broken and therefore I don't want to spend too much time fiddling with it. The main goal here is to get the sub-reports re-compiled, which the pull request fullfills. Anything beyond that we should re-think the report engine in general (in my opinion).

mvrueden:
I addressed 3 and 4. Can you re-test it, please.

StCyr:
I'm compiling now and will test it tomorrow.

Cheers

mvrueden:
As soon as bamboo goes green, this can be merged.

PR #1232, opened 2017-01-10
HZN-985 / HZN-987: add node location to alarm/notification UI

RangerRick:
This PR adds location to the alarm and notification UIs.  It also fixes up some issues with the event browser showing locations and adds support for showing both the node and event source locations, and contains a clarification in the outage UI.

Where appropriate, I always refer to a location associated with a node as a "Node Location".  Otherwise event locations are either just "Location" or "Event Source Location".

JIRA: 
* http://issues.opennms.org/browse/HZN-985
* http://issues.opennms.org/browse/HZN-987

Related JIRA issues:
* http://issues.opennms.org/browse/HZN-314
* http://issues.opennms.org/browse/HZN-315

### DIFF
--- a/opennms-webapp/src/main/java/org/opennms/web/event/filter/NegativeLocationFilter.java
+++ b/opennms-webapp/src/main/java/org/opennms/web/event/filter/NegativeLocationFilter.java
@@ -53,7 +53,7 @@ public class NegativeLocationFilter extends NotEqualOrNullFilter<String> {
     @Override
     public boolean equals(Object obj) {
         if (obj == null) return false;
-        if (!(obj instanceof NegativeNodeLocationFilter)) return false;
+        if (!(obj instanceof NegativeLocationFilter)) return false;
         return (this.toString().equals(obj.toString()));
     }
 }
--- a/opennms-webapp/src/main/java/org/opennms/web/event/filter/NegativeLocationFilter.java
+++ b/opennms-webapp/src/main/java/org/opennms/web/event/filter/NegativeLocationFilter.java
@@ -32,7 +32,7 @@ import org.opennms.web.filter.NotEqualOrNullFilter;
 import org.opennms.web.filter.SQLType;
 
 public class NegativeLocationFilter extends NotEqualOrNullFilter<String> {
-    public static final String TYPE = "locationNot";
+    public static final String TYPE = "locationnot";
     private String m_location;
 
     public NegativeLocationFilter(String location) {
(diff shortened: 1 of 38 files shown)


opennms-webapp/src/main/java/org/opennms/web/event/filter/NegativeLocationFilter.java
```diff
@@ -53,7 +53,7 @@ public String toString() {
     @Override
     public boolean equals(Object obj) {
         if (obj == null) return false;
-        if (!(obj instanceof NegativeLocationFilter)) return false;
+        if (!(obj instanceof NegativeNodeLocationFilter)) return false;
```
agalue: I think it should be NegativeLocationFilter (like the class name), as there is another class called NegativeNodeLocationFilter

RangerRick: Oop, good catch... cut&paste error.  There is so much code duplication in these damn filters.  :/

agalue:
I tested this branch with my test environment, getting data locally, and through a minion. The filters seems to be working fine on events, alarms, notification and outages.

On the alarms page (full details), it is not clear the notion of "Location" and "Node Location" as there are no explicit columns for them like you can see on the events. Maybe add a small label would help.

agalue:
The event detail page shows the location, but not the "Node Location".
The alarm detail page doesn't show the location and node location.
The notification detail page doesn't show the location.

RangerRick:
As for the alarm detail page, I'm not really sure how to label it without taking up too much space, beyond what's in the headers.  Unless we want to completely rearrange that page, or start putting icons next to them or something.

As far as the detail pages, it is true, I did not even look at them, because it didn't occur to me to do so, since that wasn't what the issues said to do.  :)  I can definitely add that info tho'.

RangerRick:
On second thought, maybe I should rearrange the alarm page to be like the event one and put the log message on a second row.  The rows wouldn't be any taller, they'd just be laid out better.

agalue:
> On Jan 11, 2017, at 10:26 AM, Benjamin Reed <notifications@github.com> wrote:
> 
> On second thought, maybe I should rearrange the alarm page to be like the event one and put the log message on a second row. The rows wouldn't be any taller, they'd just be laid out better.

Sounds good to me

RangerRick:
OK, the latest commits add info to the detail pages, as well as cleaning up the alarm list page and such.

RangerRick:
Event List:
event-list

Event Detail:
event-detail

Alarm List:
alarm-list

Alarm Detail:
alarm-detail

Notification List:
notification-list

Notification Detail:
notification-detail

Outage Detail:
outage-detail

agalue:
Looks good to me, ship it!

RangerRick:
I'll wait for the latest bamboo build to finish and merge it.  Thanks for checking things out.

PR #1235, opened 2017-01-12
NMS-8989: Make surveillance category box scrollable in node detail page

indigo423:
When a node is assigned to a lot of surveillance categories, important information like recent events and outages get moved out of the view and requires scrolling through the whole node detail page. A max-height is set and a scroll bar just for the surveillance box is shown.

* JIRA: http://issues.opennms.org/browse/NMS-8989


indigo423:
Splitted code formatting and change into two separate commits to make it easier to track just the code change.

PR #1236, opened 2017-01-17
NMS-9000: "invert-status" parameter is missing in documentation

mfuhrmann:
* JIRA: https://issues.opennms.org/browse/NMS-9000


j-white:
The `invert-status` is applicable to all monitors.

Should we add it to every table, or create a new section for "common" parameters?

indigo423:
I would vote to add it once and like the idea with "common" parameters like retry and timeout.

mfuhrmann:
@indigo423 This might be a good idea for those people who read the documentation from the beginning to the end. My typical approach is to search a monitor on the left side in the tree, clicking on it and then I want to see all parameters for this poller. Novices could come in the situation to overlook the common section by this way. To be honest, retry and timeout isn't very confusing, but the inverter is really really nice and worth to present on a silver plate :+1: 
Either way I can fix the documentation when we come to a decision.

Feedback welcome!

indigo423:
We can make a link as reference to the "common" section in each poller?

indigo423:
Reopened in #1240

PR #1238, opened 2017-01-19
NMS-8947: Close HttpConnection after usage

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-8947

### DIFF
--- a/features/ticketing/jira-integration/pom.xml
+++ b/features/ticketing/jira-integration/pom.xml
@@ -70,5 +70,23 @@
             <artifactId>junit</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.opennms.dependencies</groupId>
+            <artifactId>spring-dependencies</artifactId>
+            <type>pom</type>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.glassfish.jersey.core</groupId>
+            <artifactId>jersey-common</artifactId>
+            <version>2.4.1</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.opennms.dependencies</groupId>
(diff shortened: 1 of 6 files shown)


j-white:
The `fieldInfoCache` also invokes `getConnection()`, but that one doesn't appear to be closed.

j-white:
:+1: from me, waiting on Bamboo

features/ticketing/jira-integration/pom.xml
```diff
@@ -70,5 +70,23 @@
...
+            <type>pom</type>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.glassfish.jersey.core</groupId>
+            <artifactId>jersey-common</artifactId>
+            <version>2.4.1</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
```
soleger: This dep is a dupe.

features/ticketing/jira-integration/pom.xml
```diff
@@ -70,5 +70,23 @@
             <artifactId>junit</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.opennms.dependencies</groupId>
+            <artifactId>spring-dependencies</artifactId>
+            <type>pom</type>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
```
soleger: I don't see why this jersey dep was added.

mvrueden:
The Jersey and Spring dependencies were added, as they are required by the `JiraTicketerPlugin` to work. Usually the default implementation from Atlassian is providing those dependencies. However we repackage it to make it a osgi bundle. While doing that we do not include spring and jersey. In order to make the tests work, we need those dependencies, otherwise `ClassNotFoundExceptions` are thrown all over the place.

PR #1239, opened 2017-01-20
HZN-995: Update collectors to use strict attribute types

j-white:
JIRA: https://issues.opennms.org/browse/HZN-995

Here we update the `CollectionSet`s to use enums for types instead of strings and provide an `XmlAdapter` used to parse the types when the configurations are unmarshaled.

This refactor is part of a larger effort to help support all of the available collector implementations on Minion (see HZN-994).


soleger:
Looks good to me!

PR #1244, opened 2017-01-25
HZN-620: change minion to run under the JRE

RangerRick:
This PR changes the Minion packages to depend on `jre-1.8.0` rather than `jdk-1.8.0`.

* JIRA: http://issues.opennms.org/browse/HZN-620


RangerRick:
Note: I had previously merged my in-progress BSM test fixes so this could go green.  I've now rebased it to `release-19.0.0` with the `@Ignore` bits, so it should pass again and be mergeable.

PR #1245, opened 2017-01-25
NMS-8085: Use platform independent resource path

fooker:
The NIO Path library was used to identify a resource. These paths are platform dependend and underly the rules of the OS running on. As resource path can contain arbitary element names, the API was changed to use ResourcePath instead.

This also replaces collons in the path with underscores to fix the original problem of storing IPv6 addresses on windows.

* JIRA: http://issues.opennms.org/browse/NMS-8085


features/collection/persistence-rrd/pom.xml
```diff
@@ -31,6 +31,10 @@
       <groupId>org.opennms.features.collection</groupId>
       <artifactId>org.opennms.features.collection.api</artifactId>
     </dependency>
+    <!--<dependency>-->
```
j-white: Nit: Can we just remove this block?

opennms-dao/pom.xml
```diff
@@ -114,7 +114,6 @@
     <dependency>
       <groupId>org.opennms.features.collection</groupId>
       <artifactId>org.opennms.features.collection.persistence.rrd</artifactId>
-      <scope>test</scope>
```
j-white: Is this necessary? Having openms-dao depend on rrd persistence doesn't seem right.

fooker: It's just for accessing https://github.com/OpenNMS/opennms/pull/1245/files#diff-0a564fd806f532496281b87014032ac2R367. I can't do it the other way around because the dependency still exists for testing.

j-white: Maybe we could move that method to the ResourcePath class itself instead? i.e. `resourcePath.toToFilesystemPath()`

fooker: I've done that first, but it looks strange because it's only related to the filesystem implementation. I will have a look if I can find a better place to do this.

j-white:
Thanks for the update!

This looks good to be merged once the conflicts are resolved.

fooker:
I've resolved the conflicts and bamboo is still outstanding (even if it's shown as passed here).

soleger:
Does this require any upgrade scripts? It looks like the path information should stay the same, right?

fooker:
The paths stays the same on systems where the colon is supported. On systems which requires translation, the files have never been written - therefore, no update script is required.

fooker:
Bamboo is green and latest development has been merged.

PR #1246, opened 2017-01-26
HZN-911: Add Geographical Status Map to Index Page

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-911


mvrueden:
I found an issue: The heatmap-box is no longer working as it requires requirejs, but we do no longer include it on the index.jsp. I have to see what the other boxes such as grafana, trendline, etc. do.

pbrane:
I saw that, as well. I thought it was just me because I could get it to work, occasionally, if I constantly refreshed it like a mad jack rabbit on cocaine.

David

mvrueden:
Should work now.

RangerRick:
Just checked out the branch and this looks pretty awesome. 👍

features/geolocation/feature/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+            <artifactId>org.opennms.features.geolocation.services</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.opennms</groupId>
+            <artifactId>opennms-dao-api</artifactId>
```
mvrueden: AFAIK this should be picked up via custom.properties and should not be part of the bundle/feature.

soleger: Agreed.

j-white:
Looks good to me. Great work @mvrueden.

opennms-webapp/src/main/webapp/lib/bootstrap-sass-official/assets/stylesheets/bootstrap/_variables.scss
```diff
@@ -355,7 +355,7 @@ $container-lg:                 $container-large-desktop !default;
 //##
 
 // Basics of a navbar
-$navbar-height:                    50px !default;
+$navbar-height:                    65px !default;
```
j-white: If we need to update this variable, we should set it on our theme instead of modifying it here.

j-white:
After compiling this branch, I'm left with a number of modified files in `opennms-webapp/src/main/webapp/lib`.

mvrueden:
I updated the pom dependencies, as mentioned above. In addition I reverted the navbar height changes, as they have side-effects e.g. at the topology map and it is not crucial. As soon as bamboo goes green, this can be merged, from my point of view. Please squash merge.

mvrueden:
@StCyr tested this as well and confirmed that it works.

opennms-webapp/src/main/webapp/index.jsp
```diff
@@ -63,7 +64,7 @@
 	<!-- Middle Column -->
 	<div class="col-md-6" id="index-contentmiddle">
 		<%
-			String centerUrl = System.getProperty("org.opennms.web.console.centerUrl", "/includes/categories-box.jsp");
+			String centerUrl = System.getProperty("org.opennms.web.console.centerUrl", "/includes/categories-box.jsp,/geomap/map-box.jsp");
```
christianpape: Please also change the default in the opennms.properties file: "The default is the view of SLM/RTC categories: /includes/categories-box.jsp."

mvrueden: Too late, already merged, please open an issue for this, if this is required.

PR #1247, opened 2017-01-26
NMS-9025: In Threshd, added some debug statements + slight code refactoring

StCyr:
The goal of this PR is to add some debug statements in Threshd's code.

While doing so, I also rewrote the logic of 2 functions.

* JIRA: http://issues.opennms.org/browse/NMS-9025

### DIFF
--- a/opennms-services/src/main/java/org/opennms/netmgt/threshd/ThresholdingSet.java
+++ b/opennms-services/src/main/java/org/opennms/netmgt/threshd/ThresholdingSet.java
@@ -321,7 +321,7 @@ public class ThresholdingSet {
         // Find the filters for threshold definition for selected group/dataSource
         ResourceFilter[] filters = thresholdEntity.getThresholdConfig().getBasethresholddef().getResourceFilter();
         if (filters.length == 0) {
-            LOG.info("passedThresholdFilters: No filters found for threshold {}", thresholdEntity.toString());
+            LOG.info("passedThresholdFilters: No filters found for threshold {}", thresholdEntity);
             return true;
         }
 
--- a/opennms-services/src/main/java/org/opennms/netmgt/threshd/ThresholdingSet.java
+++ b/opennms-services/src/main/java/org/opennms/netmgt/threshd/ThresholdingSet.java
@@ -320,15 +320,19 @@ public class ThresholdingSet {
     protected boolean passedThresholdFilters(CollectionResourceWrapper resource, ThresholdEntity thresholdEntity) {
         // Find the filters for threshold definition for selected group/dataSource
         ResourceFilter[] filters = thresholdEntity.getThresholdConfig().getBasethresholddef().getResourceFilter();
-        if (filters.length == 0) return true;
+        if (filters.length == 0) {
+            LOG.info("passedThresholdFilters: No filters found for threshold {}", thresholdEntity.toString());
+            return true;
+        }
+
         // Threshold definition with filters must match ThresholdEntity (checking DataSource and ResourceType)
-        LOG.debug("passedThresholdFilters: applying {} filters to resource {}", filters.length, resource);
-        int count = 1;
         String operator = thresholdEntity.getThresholdConfig().getBasethresholddef().getFilterOperator().toLowerCase();
-        boolean andResult = true;
+        LOG.debug("passedThresholdFilters: applying {} filters to resource {} with logical connection operator '{}'", filters.length, resource, operator);
+        int count = 0;
+        boolean result = true;
         for (ResourceFilter f : filters) {
-            LOG.debug("passedThresholdFilters: filter #{}: field={}, regex='{}'", count, f.getField(), f.getContent());
             count++;
+            LOG.debug("passedThresholdFilters: filter #{}: field={}, regex='{}'", count, f.getField(), f.getContent());
             // Read Resource Attribute and apply filter rules if attribute is not null
             String attr = resource.getFieldValue(f.getField());
             if (attr != null) {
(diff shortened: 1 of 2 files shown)


StCyr:
Yes I can do that.

But, I may not have been clear in my initial comment, and, though I rewrote
some logic to make it easier to read (imo), I didn't expect to change the
logics' output.

Maybe just re-wording the PR could be sufficient?

Cyrille

2017-01-27 10:07 GMT+01:00 Ronny Trommer <notifications@github.com>:

indigo423:
Yes, in this case rewording the PR and JIRA Issue would be sufficient. Thanks.

opennms-services/src/main/java/org/opennms/netmgt/threshd/ThresholdingSet.java
```diff
@@ -320,15 +320,19 @@ public final boolean isNodeInOutage() {
     protected boolean passedThresholdFilters(CollectionResourceWrapper resource, ThresholdEntity thresholdEntity) {
         // Find the filters for threshold definition for selected group/dataSource
         ResourceFilter[] filters = thresholdEntity.getThresholdConfig().getBasethresholddef().getResourceFilter();
-        if (filters.length == 0) return true;
+        if (filters.length == 0) {
+            LOG.info("passedThresholdFilters: No filters found for threshold {}", thresholdEntity.toString());
```
j-white: There's no need for the `.toString()` here.

j-white:
A number of threshold related test fail with these changes. See http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1295-1#new-failed-tests.

StCyr:
Strange, I don't see what's wrong. Maybe some optimization forces us to split the test in two?

So, I should be doing this:

```
 if (ds == null || "".equals(ds)) {
            return null;
}

if ("nodeid".equalsIgnoreCase(ds)) {
       return Integer.toString(m_nodeId);
} else if ("ipaddress".equalsIgnoreCase(ds)) {
       return m_hostAddress;
} else if ("iflabel".equalsIgnoreCase(ds)) {
      return getIfLabel();
} else if ("id".equalsIgnoreCase(ds)) {
       return m_resource.getPath().getFileName().toString();
}
```

instead of:

```
 if (ds == null || "".equals(ds)) {
            return null;
        } else if ("nodeid".equalsIgnoreCase(ds)) {
            return Integer.toString(m_nodeId);
        } else if ("ipaddress".equalsIgnoreCase(ds)) {
            return m_hostAddress;
        } else if ("iflabel".equalsIgnoreCase(ds)) {
            return getIfLabel();
        } else if ("id".equalsIgnoreCase(ds)) {
            return m_resource.getPath().getFileName().toString();
        }
```

StCyr:
ok, corrected the code: `mvn -Dtest=ThresholdEvaluatorAbsoluteChangeTest test` doesn't report any error anymore

RangerRick:
Looks like a few more failures in the latest build: http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1295/latest

StCyr:
Thing is I don't see how my changes relate to the testThresholsFiltersOnGenericResource test?!

The other failure seems to be caused by a LOG.warn statement. And it'll most probably be fixed if I change it to LOG.info I guess

opennms-services/src/main/java/org/opennms/netmgt/threshd/ThresholdingSet.java
```diff
@@ -320,15 +320,19 @@ public final boolean isNodeInOutage() {
...
+            return true;
+        }
+
         // Threshold definition with filters must match ThresholdEntity (checking DataSource and ResourceType)
-        LOG.debug("passedThresholdFilters: applying {} filters to resource {}", filters.length, resource);
-        int count = 1;
         String operator = thresholdEntity.getThresholdConfig().getBasethresholddef().getFilterOperator().toLowerCase();
-        boolean andResult = true;
+        LOG.debug("passedThresholdFilters: applying {} filters to resource {} with logical connection operator '{}'", filters.length, resource, operator);
+        int count = 0;
```
agalue: It used to be count=1, why did you change it ?

opennms-services/src/main/java/org/opennms/netmgt/threshd/ThresholdingSet.java
```diff
@@ -337,13 +341,18 @@ protected boolean passedThresholdFilters(CollectionResourceWrapper resource, Thr
                     final Matcher m = p.matcher(attr);
                     boolean pass = m.matches();
                     LOG.debug("passedThresholdFilters: the value of {} is {}. Pass filter? {}", f.getField(), attr, pass);
+
                     if (operator.equals("or") && pass) {
-                        return true;
+                        LOG.debug("passedThresholdFilters: Filter #{} passed and logical connection operator is 'or', no need to apply other filters", count);
+                        break;
```
agalue: Why are you adding a "break" statement here ? If the operator was "or", obviously it will never be equal to "and" ;)

opennms-services/src/main/java/org/opennms/netmgt/threshd/ThresholdingSet.java
```diff
@@ -356,10 +365,14 @@ protected boolean passedThresholdFilters(CollectionResourceWrapper resource, Thr
                 }
             }
         }
-        if (operator.equals("and") && andResult) {
+
+        if (result) {
```
agalue: Why have you changed the logic here ? the "result" boolean makes sense ONLY when the operator was a NaN  (which is why I originally called it "andResult").

My suggestion is, restore all the original logic and only focus on adding the additional logging statements you want.

mvrueden:
The code changes made here obfuscate the meaning unnecessarily.
At the moment I'd vote to not merge it.
@StCyr can you tell us, what you want to achieve?

RangerRick:

j-white:
Closing pending further changes - the tests are still failing and no feedback in 3 months.

PR #1255, opened 2017-01-27
NMS-9030: Add Admin breadcrumb in Manage Minions page

indigo423:
Added Admin breadcrumb into Manage Minions page in Admin section.

* JIRA: http://issues.opennms.org/browse/NMS-9030


indigo423:
Not approved @soleger: 

> ACLs don't restrict those pages to only admin users because normal users should be able to view them. i think that Admin breadcrumbs will confuse normal users and if you need to get back to /admin from there, you should be able to use the back button

PR #1257, opened 2017-01-27
NMS-9026: Fixed JavaScript error in Heatmap

christianpape:
JIRA: https://issues.opennms.org/browse/NMS-9026

### DIFF
--- a/opennms-webapp/src/main/webapp/heatmap/heatmap-box.jsp
+++ b/opennms-webapp/src/main/webapp/heatmap/heatmap-box.jsp
@@ -118,7 +118,22 @@
     <div id="treemap"></div>
 
     <script type="text/javascript">
-        window.onload = function() {
+
+        function addOnLoadEvent(funcToAdd) {
+            var oldOnLoadFunction = window.onload;
+            if (typeof window.onload != 'function') {
+                window.onload = funcToAdd;
+            } else {
+                window.onload = function() {
+                    if (oldOnLoadFunction) {
+                        oldOnLoadFunction();
+                    }
+                    funcToAdd();
+                }
+            }
+        }
+
+        addOnLoadEvent(function() {
             require(['jquery', 'jquery-ui/jquery-ui', '../js/jquery.ui.treemap'], function( $ ) {
                 var mouseclickHandler = function (e, data) {
                     var nodes = data.nodes;
@@ -193,7 +208,7 @@
                     });
                 });
             });
-        }
+        });
     </script>
     <div class="panel-footer">
         <div class="row">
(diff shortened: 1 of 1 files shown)


indigo423:
Should we fix this bug in foundation-2016? https://github.com/OpenNMS/opennms/tree/foundation-2016/opennms-webapp/src/main/webapp/heatmap

mvrueden:
I agree, this should be re-targeted for `foundation-2016`

christianpape:
Rebasing changes to foundation-2016

christianpape:
Sorry, I checked this and it looks like the requirejs changes are not in foundation-2016.

opennms-webapp/src/main/webapp/heatmap/heatmap-box.jsp
```diff
@@ -118,80 +118,84 @@
...
+                    $.getJSON(url, function (data) {
+                        children = data.children;
+                        refresh();
+                    });
+                });
             });
-        });
+        }
+
+        setTimeout(initialize, 250);
```
j-white: What do we need to wait for exactly?

christianpape: I found this workaround on stackoverflow.com for these "$.widget is not a function" errors. But it also works by encapsulating the code inside a window.onload function. So, I changed this...

christianpape:
The error does not occur in foundation-2016 because the foundation's code is not using requirejs in the Heatmap yet.

indigo423:
@christianpape nice, thanks for investigation.

opennms-webapp/src/main/webapp/heatmap/heatmap-box.jsp
```diff
@@ -118,80 +118,82 @@
...
-                location.href = "<%=request.getRequestURI()%>?mode=<%=mode%>&heatmap=nodesByCategory&category=" + nodes[0].id;
-                <%
-                  }
-
-                  if ("monitoredServices".equals(heatmap)) {
-                %>
-                location.href = "<%=request.getRequestURI()%>?mode=<%=mode%>&heatmap=nodesByMonitoredService&monitoredService=" + nodes[0].id;
-                <%
-                  }
+        window.onload = function() {
```
j-white: This may break any other scripts on the page that rely on `window.onload`, since only a single function can be set.

If we leverage `window.onload`, we should make sure that any other function which was previously assigned is also called.

christianpape: Good point

j-white:
Thanks for the updates. :+1+ from me.

PR #1252, opened 2017-01-27
HZN-997: Update collectors to use the CollectionSetBuilder

j-white:
JIRA: https://issues.opennms.org/browse/HZN-997

Here we update all of the collectors (except the SnmpCollector) to use the CollectionSetBuilder when generating CollectionSets.

(This will allow these collectors to be invoked with an RPC to a Minion.)

PR #1260, opened 2017-01-30
HZN-1001: Refactor ServiceCollector implementations out of opennms-services

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1001

Here we move the HttpCollector and the JMXCollector out of `opennms-services` and into `features/collection/collectors` to make it possible to load these in the Minion container.

This patch builds off of the work in https://github.com/OpenNMS/opennms/pull/1252.


j-white:
I'm going to rework this, since it depends on https://github.com/OpenNMS/opennms/pull/1252.

PR #1264, opened 2017-01-31
NMS-9023: Rework the attribute to jira field mapping

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-9023


mvrueden:
Waiting for testers to confirm that this works.

mvrueden:
We decided to merge it without feedback. `18.0.4` should be released ASAP.

PR #1265, opened 2017-02-01
NMS-9065: Fixed the Asset pages JSPs to handle ACL filtering

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-9065

### DIFF
--- a/opennms-webapp/src/main/webapp/asset/modify.jsp
+++ b/opennms-webapp/src/main/webapp/asset/modify.jsp
@@ -29,7 +29,7 @@
 
 --%>
 
-<%@page language="java" contentType="text/html" session="true"  %>
+<%@page language="java" contentType="text/html" session="true" import="org.opennms.web.springframework.security.AclUtils"%>
 
 <jsp:include page="/includes/bootstrap.jsp" flush="false">
   <jsp:param name="title" value="Modify Asset" />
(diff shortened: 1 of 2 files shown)


agalue:
Looks good to me, if the karma tests and the smoke tests are passing, go ahead and merge it.

agalue:
IMPORTANT: The Angular version of the asset page is available ONLY on 19. On 18 and older, the GWT version is used. Have you checked both versions ?

christianpape:
I tested it 5min ago and the problem seems also exist with 19. The fix for the asset/nodelist.jsp here will also work for 19. This will prevent the nodes to be listed but if you know the nodeId you can still open the corresponding asset page.

christianpape:
So, the idea is to create a second branch/PR with similar fixes for release-19.0.0. Do you agree?

agalue:
> On Feb 1, 2017, at 10:04 AM, Christian Pape <notifications@github.com> wrote:
> 
> I tested it 5min ago and the problem seems also exist with 19. The fix for the asset/nodelist.jsp here will also work for 19. This will prevent the nodes to be listed but if you know the nodeId you can still open the corresponding asset page.

Excellent!

agalue:
> On Feb 1, 2017, at 10:19 AM, Christian Pape <notifications@github.com> wrote:
> 
> So, the idea is to create a second branch/PR with similar fixes for release-19.0.0. Do you agree?

Makes sense if a patch for 18 doesn’t apply for 19 (because of GWT, or something else).

PR #1268, opened 2017-02-01
NMS-5450: Escape colons in IPv6 directory paths in Jasper reports

soleger:
Added util method to escape colons in IP addresses so that reports can access RRD files in directories with IPv6 addresses in the path.

* JIRA: http://issues.opennms.org/browse/NMS-5450


j-white:
Looks good, we had the same problem / solution in the Measurements API in https://github.com/OpenNMS/opennms/commit/67ebcb29a9b8f043f0971f7c03a4fc34719cb054.

PR #1271, opened 2017-02-02
NMS-9071: Fixed IPv6 handling in Jsr160ConnectionFactory

soleger:
This fixes JMX polling and collections against IPv6 addresses.

* JIRA: http://issues.opennms.org/browse/NMS-9071

### DIFF
--- a/opennms-services/src/main/java/org/opennms/protocols/jmx/connectors/Jsr160ConnectionFactory.java
+++ b/opennms-services/src/main/java/org/opennms/protocols/jmx/connectors/Jsr160ConnectionFactory.java
@@ -28,10 +28,11 @@
 
 package org.opennms.protocols.jmx.connectors;
 
-import java.net.Inet6Address;
+import java.io.IOException;
 import java.net.InetAddress;
 import java.net.MalformedURLException;
-import java.util.*;
+import java.util.HashMap;
+import java.util.Map;
 
 import javax.management.MBeanServerConnection;
 import javax.management.remote.JMXConnector;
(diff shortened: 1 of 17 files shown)


j-white:
It looks like some of the other connection factories suffer from a similar problem, i.e. :https://github.com/OpenNMS/opennms/blob/12cf597161d740cb8e1c8bbc2cdfe8c4163fd58e/opennms-services/src/main/java/org/opennms/protocols/jmx/connectors/JMXSecureConnectionFactory.java#L87

opennms-services/src/main/java/org/opennms/protocols/jmx/connectors/Jsr160ConnectionFactory.java
```diff
@@ -222,11 +223,23 @@ private static JMXServiceURL getUrl(InetAddress address, int port, String protoc
...
         }
         return url;
-    }    
+    }
+
+    /**
+     * Method that wraps IPv6 addresses in square brackets so that they are parsed
+     * correctly by the {@link JMXServiceURL} class.
+     */
+    private static String toUrlIpAddress(InetAddress addr) {
```
mvrueden: Can't we just fix this in `InetAddressUtils.str(address)` In general?

RangerRick: I had asked about that when he talked about this in scrum, but the problem is that the square brackets are only for URLs, not general display of IP addresses.  He could add a specific "url-friendly" method to `InetAddressUtils`, I suppose, but it's not appropriate for `str`.

soleger: I'm going to have to move this method to ```InetAddressUtils``` since the provisioning detectors also need it but it will still be a separate ```toUrlIpAddress()``` method.

PR #1284, opened 2017-02-09
NMS-9093: Cleanup OpenNMS default events

indigo423:
Removed IETF specific events from opennms.events.xml and provided separate files and included them in the default eventconf.xml which allows to remove them easily if not needed. Also reorganised OpenNMS internal event definitions into separate files for easier maintainability.

Capsd and Linkd events are still provided but are probably deprecated cause both components don't exist in the product anymore.

* JIRA: http://issues.opennms.org/browse/NMS-9093


soleger:
I would also rename ```ncs-component.events.xml``` to ```opennms.ncs.events.xml``` since they are events generated by an OpenNMS component.

indigo423:
Renamed ncs-component.events.xml to opennms.ncs-components as @soleger suggested.

agalue:
Here are my thoughts ....

I would merge the `caspd` events and the `importer` events into `opennms.provisioning.events.xml`; because, effectively, it is `Provisiond` who generates those events now.

There are events that are effectively generated by `Pollerd` on two different files: `opennms.nodes.events.xml` and `opennms.poller.events.xml`. I would put them on the same place.

`opennms.remote.poller.events.xml` contains Minion events, maybe the name of the file is not very accurate.

Guess we might merge `opennms.linkd.events.xml` and `opennms.topology-status.events.xml` as both are related to the same "source".

That makes me think, it doesn't make too much sense to have `opennms.internal.events.xml`, and then other files (like the above 2) that are also "internal" events generated by core OpenNMS daemons, maybe the "internal" file needs more love.

indigo423:
@agalue thanks for feedback, I built in all your suggestion and just left one as it was the opennms.caspd.events.xml. For the reason we don't have it anymore in the system it would be very easy to delete. Thanks a lot.

agalue:
Do not confuse things here. The events with the word "capsd" on them are now handled by Provisiond (the fact that Capsd doesn't exist has nothing to do with the UEIs). You can search for them on the code if you want.

indigo423:
@agalue thanks for the hint Provisiond uses capsd events. Should we migrate those legacy events to something more useful? There will be no one starting with opennms >=16 which is already a few years old to remember what Capsd is and I don't want to explain old fart stories :)

agalue:
You mean change UEIs and definition content ? That requires code changes, and either an upgrade task or a liquibase change to fix the UEI on existing databases. Are you willing to do it ?

indigo423:
I would recommend to do the capsd UEI change in a separate issue and different branch.

j-white:
Anything else pending here?

indigo423:
All review comments are reflected in the changes beside the rename of capsd events which requires code changes and should be handled in a different issue.

pbrane:
XML and XSDs have always followed the 2 space indentation standard. Not sure about that reference.

PR #1285, opened 2017-02-09
HZN-1016: Make geolocation resolution part of provisioning

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1016

This issue removes the geolocation lookup of address strings from the node map application and makes it part of the provisioning process. It can be disabled by the according property. The Geolocation (Rest) Service no longer supports resolution of missing coordinates.

The remote poller app has also some geolocation lookup, which is completely different from the rest of opennms and was not touched here. 

This is targeted for 20.0.0.
Please try it out and let me know what you think.

Thank you


j-white:
Worked well for me. I've merged develop back into this branch, will wait for Bamboo again before merging.

PR #1287, opened 2017-02-09
HZN-1019: Add a new requisition:// URL handler

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1019

Adds a new 'requisition://' URL handler which support retrieving requisitions from Minion via RPC and a Karaf shell command that can be used to show the results.

Currently only the 'file' and 'http' handlers are available. DNS and Vmware will be added within the scope of https://issues.opennms.org/browse/HZN-1018.


mvrueden:
Looks good to me.

PR #1289, opened 2017-02-10
HZN-626: Fixed ActiveMQ counters and graph definitions

soleger:
* JIRA: http://issues.opennms.org/browse/HZN-626


j-white:
We also need to include upgrade instructions here now, since the RRDs need to be recreated in order for the attribute types to be changed.

RangerRick:

soleger:
These counters will all be zero unless you're using a Minion, correct? I just want to make sure that I'm phrasing the release notes correctly.

j-white:
Mostly zeros unless Minion is used. There are still some connections and consumers without any Minions though.

PR #1291, opened 2017-02-11
Features/17.1.1 enlinkd upgrade

rssntn67:
NMS-8751: Poller Node Down without outages

* JIRA: http://issues.opennms.org/browse/NMS-8751


rssntn67:
Not good merge

PR #1292, opened 2017-02-11
NMS-8751: Poller Node Down without outage

rssntn67:
NMS-8751: Poller Node Down without outage

* JIRA: http://issues.opennms.org/browse/NMS-8751

### DIFF
--- a/opennms-services/src/main/java/org/opennms/netmgt/poller/PollerEventProcessor.java
+++ b/opennms-services/src/main/java/org/opennms/netmgt/poller/PollerEventProcessor.java
@@ -726,6 +726,8 @@ final class PollerEventProcessor implements EventListener {
         for (final Service databaseService : databaseServices) {
             if (polledServices.contains(databaseService)) {
                 LOG.debug("{} is being skipped. Already scheduled.", databaseService);
+                //FIXME Discuss with Jesse
+                continue;
             }
 
             LOG.debug("{} is being scheduled (or rescheduled) for polling.", databaseService);
--- a/opennms-services/src/main/java/org/opennms/netmgt/poller/QueryManagerDaoImpl.java
+++ b/opennms-services/src/main/java/org/opennms/netmgt/poller/QueryManagerDaoImpl.java
@@ -318,4 +318,29 @@ public class QueryManagerDaoImpl implements QueryManager {
                     serviceName, nodeId,  ipAddr, status, e);
         }
     }
+    
+    @Override
+    public void updateServiceLastGood(int nodeId, String ipAddr, String serviceName, Date updateTime) {
+        try {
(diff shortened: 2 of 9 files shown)


opennms-services/src/main/java/org/opennms/netmgt/poller/PollerEventProcessor.java
```diff
@@ -726,6 +726,8 @@ private void serviceReschedule(Long nodeId, String nodeLabel, Event sourceEvent,
         for (final Service databaseService : databaseServices) {
             if (polledServices.contains(databaseService)) {
                 LOG.debug("{} is being skipped. Already scheduled.", databaseService);
+                //FIXME Discuss with Jesse
```
j-white: Let's discuss. It looks like I may have forgot the continue statement when I made the changes in NMS-7761.

opennms-services/src/main/java/org/opennms/netmgt/poller/pollables/PollableElement.java
```diff
@@ -526,6 +526,8 @@ protected void doInheritParentalCause() {
             // I have no cause but my parent is down.. mark me as down as well
             // I already have a cause that's larger than myself then inherit as well
             setCause(parentalCause);
+            if (getStatus().isUp())
+            	createOutage(parentalCause);
```
j-white: Can you elaborate on why this was necessary?

opennms-services/src/main/java/org/opennms/netmgt/poller/QueryManagerDaoImpl.java
```diff
@@ -318,4 +318,33 @@ public void updateServiceStatus(int nodeId, String ipAddr, String serviceName, S
                     serviceName, nodeId,  ipAddr, status, e);
         }
     }
+    
+    @Override
+    public void updateServiceLastGood(int nodeId, String ipAddr, String serviceName, Date updateTime) {
+        try {
+            OnmsMonitoredService service = m_monitoredServiceDao.get(nodeId, InetAddress.getByName(ipAddr), serviceName);
+            service.setLastGood(updateTime);;
```
j-white: Is updating the lastGood and lastFail columns only informational, or do they have some formal purpose?

I noticed the columns were already in the database, but appeared to be unused prior to this.

j-white:
We should also add test cases for the scenarios described in the JIRA issue.

opennms-services/src/main/java/org/opennms/netmgt/poller/PollerEventProcessor.java
j-white: While investigating another pollerd related bug I came across this same problem. See https://github.com/OpenNMS/opennms/pull/1311.

rssntn67:
I think you do this in the other pull request. Let me close without merging

PR #1297, opened 2017-02-14
NMS-8934: Fix comparing IPv6 and IPv4 addresses.

fooker:
This fixes comparing IPv6 with IPV4 addresses (and vice versa) by returning `false` instead of failing with an exception.

* JIRA: http://issues.opennms.org/browse/NMS-8934


j-white:
Worked well here. The exception that was reported originally no longer occurs.

I added a few more test cases, and merged in the latest `foundation-2016`.

Waiting on Bamboo.

PR #1302, opened 2017-02-14
NMS-9102: fix pre-existing 'localhost' monitoring location

RangerRick:
This PR fixes an issue with upgrades to 19 when the user has defined a `localhost` location in `monitoring-locations.xml` or equivalent.

* JIRA: http://issues.opennms.org/browse/NMS-9102


soleger:
Looks good, people that install 19.0.0 may have to manually remove the row for ```install -dis``` to complete but upgrades to 19.0.1 will complete without issue.

RangerRick:
I actually tried an upgrade that way and it worked fine, since those IDs had already been marked as running in the DB.

PR #1303, opened 2017-02-14
NMS-9090: Collection timing enhancements

tdefilip:
- Use a boolean flag to control whether we subtract that time that a collection service run took in how long we sleep before the next service run.  A value of true for "strictInterval" means we do the subtraction.
- Use a boolean flag to control whether we use the start of the collection service run or the current time for all RRD updates we do for a run.  A value of true for "useCollectionStartTime" means we use the start time.

JIRA: http://issues.opennms.org/browse/NMS-9090

### DIFF
--- a/features/collection/persistence-rrd/src/main/java/org/opennms/netmgt/collection/persistence/rrd/RrdPersistOperationBuilder.java
+++ b/features/collection/persistence-rrd/src/main/java/org/opennms/netmgt/collection/persistence/rrd/RrdPersistOperationBuilder.java
@@ -73,7 +73,6 @@ public class RrdPersistOperationBuilder implements PersistOperationBuilder {
     private final Map<CollectionAttributeType, Number> m_declarations;
     private final Map<String, String> m_metaData = new LinkedHashMap<String, String>();
     private TimeKeeper m_timeKeeper = new DefaultTimeKeeper();
-    private long m_collectionTime;
 
     /**
      * RRDTool defined Data Source Types NOTE: "DERIVE" and "ABSOLUTE" not
@@ -115,16 +114,6 @@ public class RrdPersistOperationBuilder implements PersistOperationBuilder {
      */
     public RrdRepository getRepository() {
         return m_repository;
-    }
-
-    /**
-     * <p>setCollectionTime</p>
-     *
-     * @param collectionTime Long representing the event time for this
-     * persister.
-     */
-    public void setCollectionTime(long collectionTime) {
-        m_collectionTime = collectionTime;
     }
 
     private File getResourceDir(ResourceIdentifier resource) throws FileNotFoundException {
@@ -222,9 +211,7 @@ public class RrdPersistOperationBuilder implements PersistOperationBuilder {
             List<RrdDataSource> dataSources = getDataSources();
             if (dataSources != null && dataSources.size() > 0) {
                 createRRD(m_rrdStrategy, ownerName, absolutePath, m_rrdName, getRepository().getStep(), dataSources, getRepository().getRraList(), m_metaData);
-                long rrdTime = Boolean.getBoolean("org.opennms.netmgt.collection.persistence.rrd.useCollectionStartTime")
-                    ? m_collectionTime : m_timeKeeper.getCurrentTime();
-                updateRRD(m_rrdStrategy, ownerName, absolutePath, m_rrdName, rrdTime, getValues());
+                updateRRD(m_rrdStrategy, ownerName, absolutePath, m_rrdName, m_timeKeeper.getCurrentTime(), getValues());
             }
         } catch (FileNotFoundException e) {
             LoggerFactory.getLogger(getClass()).warn("Could not get resource directory: " + e.getMessage(), e);
--- a/features/collection/persistence-rrd/src/main/java/org/opennms/netmgt/collection/persistence/rrd/RrdPersistOperationBuilder.java
+++ b/features/collection/persistence-rrd/src/main/java/org/opennms/netmgt/collection/persistence/rrd/RrdPersistOperationBuilder.java
@@ -73,6 +73,7 @@ public class RrdPersistOperationBuilder implements PersistOperationBuilder {
     private final Map<CollectionAttributeType, Number> m_declarations;
     private final Map<String, String> m_metaData = new LinkedHashMap<String, String>();
     private TimeKeeper m_timeKeeper = new DefaultTimeKeeper();
+    private long m_collectionTime;
 
     /**
      * RRDTool defined Data Source Types NOTE: "DERIVE" and "ABSOLUTE" not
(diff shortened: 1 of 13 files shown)


features/collection/persistence-rrd/src/main/java/org/opennms/netmgt/collection/persistence/rrd/RrdPersistOperationBuilder.java
```diff
@@ -211,7 +222,9 @@ public void commit() throws PersistException {
             List<RrdDataSource> dataSources = getDataSources();
             if (dataSources != null && dataSources.size() > 0) {
                 createRRD(m_rrdStrategy, ownerName, absolutePath, m_rrdName, getRepository().getStep(), dataSources, getRepository().getRraList(), m_metaData);
-                updateRRD(m_rrdStrategy, ownerName, absolutePath, m_rrdName, m_timeKeeper.getCurrentTime(), getValues());
+                long rrdTime = Boolean.getBoolean("org.opennms.netmgt.collection.persistence.rrd.useCollectionStartTime")
+                    ? m_collectionTime : m_timeKeeper.getCurrentTime();
```
j-white: I would opt for overriding the TimeKeeper for the CollectionResource objects in the CollectionSet instead of explicitly passing the time through the persistence layers and performing this logic here.

opennms-services/src/main/java/org/opennms/netmgt/collectd/CollectableService.java
```diff
@@ -354,8 +361,13 @@ private void doRun() {
...
         }
         
+	// If we are doing strict interval, determine how long the collection
+        // has taken, so we can cut that off of the service interval
+	long diff = 0;
+        if (strictInterval)
+            diff = System.currentTimeMillis() - m_lastScheduledCollectionTime;
     	// Reschedule the service
-        m_scheduler.schedule(m_spec.getInterval(), getReadyRunnable());
+        m_scheduler.schedule(m_spec.getInterval() - diff, getReadyRunnable());
```
j-white: What happens if/when `diff > m_spec.getInterval()`?

j-white:
Thanks for the contribution @tdefilip!

features/collection/persistence-rrd/src/main/java/org/opennms/netmgt/collection/persistence/rrd/RrdPersistOperationBuilder.java
tdefilip: OK, I will look into the TimeKeeper and its use.

opennms-services/src/main/java/org/opennms/netmgt/collectd/CollectableService.java
tdefilip: Good point.  I will limit diff to be m_spec.getInterval() at most, so that the difference between them is always >= 0.

tdefilip:
Thanks for the feedback!  Once I make the necessary changes, do I re-pull here or do I file a different pull request?

j-white:
You can keep this same PR. The changes will show up here if as long as you keep pushing to the same branch.

features/collection/persistence-rrd/src/main/java/org/opennms/netmgt/collection/persistence/rrd/RrdPersistOperationBuilder.java
tdefilip: If I'm reading the code correctly, it looks like the best way to go about this is to add a setTimeKeeper method to CollectionSet and have all classes that implement CollectionSet set the TimeKeeper on their resources.  We can then create a constant timekeeper in CollectableSerivce with the collection time there and have that be the timekeeper we set on all resources via the previously mentioned method.

j-white: There are a lot of CollectionResource objects that would need updating. I was thinking we could wrap those objects instead.

The code in CollectableService could be replaced with something like:

```java
if (Boolean.getBoolean("org.opennms.netmgt.collection.persistence.useCollectionStartTime")) {
  final ConstantTimeKeeper timeKeeper = new ConstantTimeKeeper(new Date(0));
  // Wrap the persister visitor such that calls to CollectionResource.getTimeKeeper() return the given timeKeeper
  persister = wrapResourcesWithTimekeeper(persister, timeKeeper);           
}
```

where `wrapResourcesWithTimekeeper` is defined as:
```java
private static CollectionSetVisitor wrapResourcesWithTimekeeper(CollectionSetVisitor visitor, TimeKeeper timeKeeper) {
        // Wrap the given visitor and intercept the calls to visit the resources
        final CollectionSetVisitor wrappedVisitor = new CollectionSetVisitorWrapper(visitor) {
            private CollectionResource wrappedResource;

            @Override
            public void visitResource(CollectionResource resource) {
                // Wrap the given resource and return the custom timekeeper
                wrappedResource = new CollectionResourceWrapper(resource) {
                    @Override
                    public TimeKeeper getTimeKeeper() {
                        return timeKeeper;
                    }
                };
                getDelegate().visitResource(wrappedResource);
            }

            @Override
            public void completeResource(CollectionResource resource) {
                getDelegate().completeResource(wrappedResource);
            }
        };
        return wrappedVisitor;
    }
```

`org.opennms.netmgt.collection.support.CollectionSetVisitorWrapper` already exists, we would just need to create the analogue for the CollectionResource.

tdefilip: That doesn't seem to fully work.  The method that we care about here is BasePersister::createBuilder(), as that is where we take the resource's timekeeper and make it the builder's timekeeper.  createBuilder() is called by GroupPersister::visitGroup() and OneToOnePersister::visitAttribute(), and in both cases they use the group's/attribute's existing resource as the argument passed to createBuilder().  So modifying what is passed to the visitResource/completeResource calls doesn't do us any good here since neither of those affect what the builder gets.

I'm thinking we need to intercept visitGroup/completeGroup and visitAttribute/completeAttribute too, and in those calls we need to set the group's/attribute's resource in the visitXXX() calls to the wrapper, and then change it back during the completeXXX() calls.  Does that sound right (and safe, specifically setting an resetting the group's/attribute's resource)?

tdefilip:
OK, I have made the necessary changes and pushed them for review.

opennms-services/src/main/java/org/opennms/netmgt/collectd/CollectableService.java
```diff
@@ -354,8 +371,15 @@ private void doRun() {
             }
         }
         
+	// If we are doing strict interval, determine how long the collection
+        // has taken, so we can cut that off of the service interval
+	long diff = 0;
+        if (strictInterval) {
+            diff = System.currentTimeMillis() - m_lastScheduledCollectionTime;
+            diff = Math.max(diff, m_spec.getInterval());
```
j-white: Should this be `Math.min` instead of `Math.max`? Otherwise diff can still be > than m_spec.getInterval(), resulting an a negative schedule time bellow.

j-white:
Thanks for the updates. Now other persistence strategies can take advantage of the same feature.

All we're missing now is a few JUnit tests to validate the updated behavior of the CollectableService class.

roskens:
Does m_lastScheduledCollectionTime need to be passed as an argument to doCollection()? Its part of the class, so doCollection() can just use its value directly, right?

tdefilip:
Good point.  Fixed.

j-white:
Reopened as https://github.com/OpenNMS/opennms/pull/1320.

PR #1307, opened 2017-02-15
HZN-1005: JMXCollector support for Minion

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1005

Here we add support for loading the JMXCollector (and friends) in the Minion container.


mvrueden:
Looks good to me.

PR #1309, opened 2017-02-16
NMS-9109: The upgrade task for magic.users doesn't work on RPM based upgrades

agalue:
The upgrade task for magic.users doesn't work on RPM based upgrades.

* JIRA: http://issues.opennms.org/browse/NMS-9109


j-white:
Looks good to me.

Is there anything else we need to do for users upgrading from 19.0.0 to 19.0.1 which already ran the previous version of the upgrade task?

agalue:
Users will have to manually edit `opennms-upgrade-status.properties`, and remove the entry with `MagicUsersMigratorOffline`, in order to re-run the upgrade task.

PR #1313, opened 2017-02-19
NMS-9072: XmlCollector use parameters defined in the service definition in xml-source url

roskens:
NMS-9072: Pass the service parameters through and allow them to be used in the URLs and Request headers/data.

* JIRA: http://issues.opennms.org/browse/NMS-9072

### DIFF
--- a/protocols/xml/src/main/java/org/opennms/protocols/xml/collector/Sftp3gppXmlCollectionHandler.java
+++ b/protocols/xml/src/main/java/org/opennms/protocols/xml/collector/Sftp3gppXmlCollectionHandler.java
@@ -1,8 +1,8 @@
 /*******************************************************************************
  * This file is part of OpenNMS(R).
  *
- * Copyright (C) 2011-2014 The OpenNMS Group, Inc.
- * OpenNMS(R) is Copyright (C) 1999-2014 The OpenNMS Group, Inc.
+ * Copyright (C) 2011-2017 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2017 The OpenNMS Group, Inc.
  *
  * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
  *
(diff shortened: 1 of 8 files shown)


mvrueden:
This was originally targeted to be merged to master. Master should be protected and is for releases only. I changed the branch to be develop instead. Thank you (-:

protocols/xml/src/main/java/org/opennms/protocols/xml/collector/Sftp3gppXmlCollectionHandler.java
```diff
@@ -76,6 +78,20 @@ public XmlCollectionSet collect(CollectionAgent agent, XmlDataCollection collect
...
+        for(Map.Entry<String, Object> entry : parameters.entrySet()) {
+            if(entry.getKey().equals("SERVICE"))
+                continue;
+            if(entry.getKey().equals("collection"))
+                continue;
+            if(entry.getKey().equals("xml-collection"))
+                continue;
+            if(entry.getKey().equals("handler-class"))
+                continue;
+            params.put(entry.getKey(), ParameterMap.getKeyedString(parameters, entry.getKey(), ""));
```
mvrueden: This seems to be duplicated code, as it is used multiple times. Can we create a filter function or something like that? e.g:

``` java
// TODO add javadoc
// TODO verify that it actually works
public Map<String, String> filterParameters(Map<String, Object> input) {
   Set<String> blackList = Sets.newHashSet("SERVICE", "collection", "xml-collection", "handler-class");
   return   input.entrySet().stream()
                      .filter(e -> !blackList.contains(e.getKey())
                      .collect(Collectors.toMap(e -> e.getKey(), e -> ParameterMep.getKeyedString(input, e.getKey(), ""));
}
```

roskens: Easy enough to do.

roskens:
Rebased against develop, and squashed all commits into one.

soleger:
I've rebased against foundation-2017 and pushed to a branch in our repo:

https://github.com/OpenNMS/opennms/tree/jira/NMS-9072

I'll merge after it passes Bamboo.

soleger:
Merged the local branch, closing this PR.

PR #1314, opened 2017-02-19
NMS-9120: Selection of non-disk RRD strategies

tdefilip:
- Add a "userrd" properties which controls whether the RRD strategy writes to the DB.  If false, only the TCP strtagy is used if usetcp is true.  If usetcp is false, the strategy is null.

JIRA: http://issues.opennms.org/browse/NMS-9120


j-white:
:+1:

tdefilip:
Thanks!

PR #1316, opened 2017-02-20
NMS-9123: Mark org.opennms.core.ipc.sink.xml bundle as provided

soleger:
* JIRA: http://issues.opennms.org/browse/NMS-9123


j-white:
Verified using the output.log of the smoke test runs.

PR #1317, opened 2017-02-20
NMS-9116: Fix RPC integration tests

soleger:
Made the RPC ITs consistent so that they can run without flapping. Please merge after this runs a couple more times in Bamboo. Not a blocker for 19.0.1.

* JIRA: http://issues.opennms.org/browse/NMS-9116


j-white:
Thanks for fixing these - these tests appear to be running fine in the scheduled builds so far.

PR #1318, opened 2017-02-20
NMS-5334, NMS-2207: Update editoutages.jsp to allow users options when picking nodes for an outage

roskens:
NMS-5334: Category and Scheduled Outage GUI Enhancment Request
NMS-2207: allow categories when creating downtime schedules

    I added a couple of enhancements to the Scheduled Outages page that should make things a little bit easier when selecting nodes.

    In the nodelabel search box, you can now prefix the search string with a couple of special symbols to alter the search behavior:

    '@' will let you search for surveillance categories
    '#' Will search for nodes who are in the specified surveillance category
    '$' will let you search for requisition names
    '%' will search for nodes in a requisition that starts with the search string.

* JIRA: http://issues.opennms.org/browse/NMS-5334

### DIFF
--- a/opennms-webapp/src/main/webapp/admin/sched-outages/editoutage.jsp
+++ b/opennms-webapp/src/main/webapp/admin/sched-outages/editoutage.jsp
@@ -459,6 +459,18 @@ Could not find an outage to edit because no outage name parameter was specified 
 				String newNode = request.getParameter("newNode");
 				if (newNode == null || "".equals(newNode.trim())) {
 					// No node was specified
+				} else if (newNode.startsWith("@")) {
+					String category = newNode.substring(1);
+					for (OnmsNode node : NetworkElementFactory.getInstance(getServletContext()).getNodesWithCategories(new String[] { category }, false)) {
+						addNode(theOutage, node.getId());
+					}
+				} else if (newNode.startsWith("$")) {
+					String foreignSource = newNode.substring(1);
+					for (OnmsNode node : NetworkElementFactory.getInstance(getServletContext()).getAllNodes()) {
+						if (node.getForeignSource() != null && node.getForeignSource().equals(foreignSource)) {
+							addNode(theOutage, node.getId());
+						}
+					}
 				} else {
 					int newNodeId = WebSecurityUtils.safeParseInt(newNode);
 					addNode(theOutage, newNodeId);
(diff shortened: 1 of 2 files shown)


mvrueden:
This was originally targeted to be merged to master. Master should be protected and is for releases only. I changed the branch to be develop instead. Thank you (-:

mvrueden:
Can we not just search for all the things, without having to prefix the search? And reflect that in the result accordingly?

pbrane:
> On Feb 22, 2017, at 3:13 AM, mvrueden <notifications@github.com> wrote:
> 
> Can we not just search for all the things, without having to prefix the search? And reflect that in the result accordingly?

+1

David Hustace
The OpenNMS Group, Inc.

roskens:
Quick and dirty this was for me. I didn't have the luxury of time to make it more user-friendly a while back when I wrote this for work. This solved the issue for us of being able to select categories/requisitions for scheduled outages since the open JIRA issues weren't being worked on.


Yeah, I think it would be better to have this be more user-friendly. NMS-8698 is out there to re-write /admin/sched-outages/ as an Angular app... My recommendation is that a full review of all open issues for "scheduled outages" be done before the rewrite is be done. That way it can implement a lot of the requests people have had over the years.

Also, realize that Poll-Outages only supports two types of entries: nodes and interfaces. So if you search for a category or requisition, then you have to store the current list of nodes in that category/requisition, and you lose any knowledge that you picked that category/requisition.

jeffgdotorg:
I've just tested this PR after merging {{develop}} to @roskens branch. Only the `$` (select all nodes in a given requisition) use case seems to work. I took a video illustrating my experience, [attached in Jira](https://issues.opennms.org/secure/attachment/17766/vokoscreen-2017-06-08_11-14-20.mp4).

j-white:
@jeffgdotorg do you think this is good to merge now?

indigo423:
The way to select Nodes, Categories and Requisitions using `#`, `@` and `$` is a little bit unintuitive. I think when we overload the input field with multiple input types than we should use the same functionality as in the geographical map which does something like:

`category=opennms-eu`

Beside this, it is probably the issue which removes one of the biggest UI glitches in the Web UI :)

👍

mvrueden:
We encountered some issues when working on the bootstrap 3 to 4 Migration on this page as well. 
Maybe we can tackle them afterwards?

What do you guys think?

indigo423:
agree with @mvrueden, I don't think it make sense to block the bootstrap migration by issues like this which exist for years.

mvrueden:
I think we can all agree that the current Scheduled Outage page is a mess and needs some re-writing.
However, as this PR is now open for almost two years, my question @indigo423, @jeffgdotorg, @pbrane is:  Do we accept this change in terms of usability/use-case.

PR #1319, opened 2017-02-20
NMS-6118 : Add additional relative time periods to choose from.

roskens:
NMS-6118 : Add additional relative time periods to choose from.
 * Last Hour
 * Last 2 Hours
 * Last 4 Hours
 * Last 8 Hours
 * Last 12 Hours

* JIRA: http://issues.opennms.org/browse/NMS-6118

### DIFF
--- a/opennms-web-api/src/main/java/org/opennms/web/svclayer/model/RelativeTimePeriod.java
+++ b/opennms-web-api/src/main/java/org/opennms/web/svclayer/model/RelativeTimePeriod.java
@@ -1,8 +1,8 @@
 /*******************************************************************************
  * This file is part of OpenNMS(R).
  *
- * Copyright (C) 2006-2014 The OpenNMS Group, Inc.
- * OpenNMS(R) is Copyright (C) 1999-2014 The OpenNMS Group, Inc.
+ * Copyright (C) 2006-2017 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2017 The OpenNMS Group, Inc.
  *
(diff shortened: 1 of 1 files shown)


opennms-web-api/src/main/java/org/opennms/web/svclayer/model/RelativeTimePeriod.java
```diff
@@ -48,6 +48,16 @@
     
     static {
         s_defaultPeriods = new RelativeTimePeriod[] {
+                new RelativeTimePeriod("1_hour", "Last Hour", Calendar.HOUR,
```
mvrueden: can we make them only use one line? This looks weird. Can we also name them `last_1_hour` or something like that?

mvrueden:
This was originally targeted to be merged to `master`. Master should be protected and is for releases only. I changed the branch to be `develop` instead. Thank you (-:

opennms-web-api/src/main/java/org/opennms/web/svclayer/model/RelativeTimePeriod.java
roskens: Do you want me to change the existing ones to one line also? Looks like the rest of the file is formatted for max 80 columns.

mvrueden: It is up to you (-: I would appreciate it (-:

roskens:
Rebase'd against develop.
Changed lines to single per RelativeTimePeriod

j-white:
Looks good to me. Pushed to `pr/jira/NMS-6118`. Waiting on Bamboo.

PR #1320, opened 2017-02-21
NMS-9090: Collection timing enhancements

j-white:
JIRA: http://issues.opennms.org/browse/NMS-9090

- Use a boolean flag to control whether we subtract that time that a collection service run took in how long we sleep before the next service run.  A value of true for "strictInterval" means we do the subtraction.
- Use a boolean flag to control whether we use the start of the collection service run or the current time for all RRD updates we do for a run.  A value of true for "useCollectionStartTime" means we use the start time.

(Originally https://github.com/OpenNMS/opennms/pull/1303)


j-white:
Please squash when merging.

tdefilip:
Do I need to do something?

j-white:
I think we're all good here @tdefilip, just waiting for another set of eyes before merging.

mvrueden:
Looks good to me as well.

tdefilip:
Thanks!

PR #1325, opened 2017-02-23
NMS-9134: convert attr-summary to JAXB

RangerRick:
This PR converts `attr-summary.xsd` to JAXB.

* JIRA: http://issues.opennms.org/browse/NMS-9134


opennms-webapp/src/main/java/org/opennms/netmgt/config/attrsummary/descriptors/AttributeDescriptor.java
```diff
@@ -0,0 +1,327 @@
...
+
+
+import org.opennms.netmgt.config.attrsummary.Attribute;
+
+/**
+ * Class AttributeDescriptor.
+ * 
+ * @version $Revision$ $Date$
+ */
+@SuppressWarnings("all") public class AttributeDescriptor extends org.exolab.castor.xml.util.XMLClassDescriptorImpl {
```
j-white: Can we remove the descriptors too? We want to be able to remove all the Castor references.

RangerRick: Sure, I'll convert it to an XmlNoCastorTest then.

opennms-webapp/src/main/java/org/opennms/netmgt/config/attrsummary/.castor.cdr
```diff
@@ -0,0 +1,4 @@
+#Thu Feb 23 12:24:55 EST 2017
```
j-white: This can probably be deleted too.

PR #1326, opened 2017-02-23
NMS-9094: SNMP data collection definitions for Cisco ASA5585-SSP-60 devices

jeffgdotorg:
* JIRA: http://issues.opennms.org/browse/NMS-9094


jeffgdotorg:
Doing over to avoid merging spam to foundation

PR #1328, opened 2017-02-24
NMS-9134: Convert opennms-datasources.xml from Castor to JAXB.

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9134


mvrueden:
Can we not catch `Throwable`, but `Exception` instead? Otherwise we would end up catching `OutOfMemoryError`s and rethrow them, which is a bad idea. 
See https://docs.oracle.com/javase/7/docs/api/index.html?java/lang/OutOfMemoryError.html

j-white:
Good point. Fixed.

PR #1329, opened 2017-02-24
NMS-9134: Relocate event.xsd (which already uses JAXB instead of Castor)

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9134


mvrueden:
Looks good to me (-:

PR #1332, opened 2017-02-24
NMS-9128: Asset to topology mapper plugin

gallenc:
* JIRA: https://issues.opennms.org/browse/NMS-9128

This pull request adds a topology mapper plugin to OpenNMS foundation-2017 / opennms-19.x
This is required for a customer planning to upgrade to 19 and then to Meridian 2017. This is why it has been contributed to the  foundation-2017 stream. This is a low risk feature since it is not installed unless requested by the user and no OpenNMS 19 code has been modified (other than the feature lists).

The plugin is not installed unless requested by the user either as a feature through the ssh consol or as a plugin through the opennms plugin manager ui. The plugin can generate graphml topoloogies for import into the OpenNMS topology display. It uses the nodedao to access node and asset table information which it then used to generate the topology. Topologies are only generated when the plugin is activated and or a create topology event (uei.opennms.plugins/assettopology/create) is sent. Topologies are only imported when a topology install event is sent (uei.opennms.plugins/assettopology/install). 

Full documentation has been contributed to the opennms admin guide.

Bamboo results:
http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1348


mvrueden:
The `GraphMLTopologyProvider` was originally introduced to allow users to create their own topologies very easy and customize them to their needs.
While implementing the provider a lot of features were introduced which are also very useful in other use cases. For example the possibility to define a topology in a very easy way.

That beeing said, in the mid-term, a lot of the capabilities of the graphml model and the capabilities of the `GraphMLTopologyProvider` will very likely replace the current internal Topology model, as it is more flexible. However the new internal topology model will probably not be GraphML.

In the long run it would be nice if there were an easy (maybe programmable) API to create your own topologies.

Having that in mind and looking at the current proposal as an OpenNMS internal module, I see the following problems:
 * The Java objects are created from the graphml schema. This was already done and therefore they are duplicated. Please use the following module: https://github.com/OpenNMS/opennms/tree/269739ee8287cc18b5645de679b60015882809ed/features/graphml 
 * An internal GraphProvider should not use the rest service. It should instead use the  `GraphmlRepository` instead. This can be very easy exposed to OSGI. See https://github.com/OpenNMS/opennms/blob/269739ee8287cc18b5645de679b60015882809ed/opennms-webapp-rest/src/main/java/org/opennms/web/rest/support/graphml/GraphmlRepository.java for more information.
 * The `GraphMLTopologyAssetProvider` should be a plugin of `features/topology-map/plugins`. See https://github.com/OpenNMS/opennms/tree/269739ee8287cc18b5645de679b60015882809ed/features/topology-map/plugins 
 * Ideally this would implement the interface `GraphProvider` (see SimpleTopologyProvider) and be its own Provider, but if it uses the `GraphmlRepository` for now, that would be fine too, as a lot of capabilities are already available by the `GraphMLTopologyProvider`
 
If we address the above issues it would be much easier to support the provider in the future as well. 
In the current state it would make it hard to implement the mid/long-term goals without having to worry about this implementation all the time.

gallenc:
Thanks for reviewing this contribution. 

This plugin has been created in order to meet a customer commitment to allow the 'Atlas' feature (a.k.a. graphml topology provider) to be used to render a view of their network based upon fields in the asset table. 

(The requirements are recorded in the internal project https://issues.opennms.org/browse/BTAH-14 and related issues). 

Originally  a script was suggested to render graphml from the asset table. However the complexity of the rendering algorithm meant that a simple script would not be able to sustainably achieve this task. Creating a plugin which encapsulates the code with unit tests and provides a very simple user interface for the customer sees an optimum solution for now. Since it only uses the graphml model, I think it is very decoupled from the internal topology architecture and as such should not be difficult to sustain as OpenNMS moves forwards.

With respect to the specific issues raised, I have the following comments. 

1.  "The GraphMLTopologyProvider was originally introduced to allow users to create their own topologies very easy and customize them to their needs".

Answer: Actually, although the graphml interface is very flexible and uses a standards based data model it is still quite hard for customers to generate graphml for a graph containing a large number of layers or nodes. The intention of this plugin is to easily generate usable graphml.xml graphs which can optionally be modified by the customer before deployment. Hence it generates graphml.xml files for customers which can be modified and deployed through the OpenNMS ReST API rather than through an inaccessible internal api. The install and uninstall events are a convenience if the ReST API is not used by the customer directly.

2.   "However the new internal topology model will probably not be GraphML"".
I deliberate have not touched the internal topology model for this reason. The plugin interacts with OpenNMS in a very decoupled way through the published GraphML api which customers use (and the nodesdao and events dao which are quite simple and relatively stable) - this should make the code relatively stable with respect to internal topology model changes and the mid/long-term goals. 

3.  duplicated objects  created from the graphml schema
I agree there are substantially duplicate jaxb objects. I tried very hard to use the internally generated OpenNMS jaxb objects you reference but found that I had a lot of problems assessing them through a jersey client from within the plugin in the OSGi container. After several days I gave up and decided to generate my own jaxb model within the OSGi plugin . These classes are private to the plugin and do not interfere with the rest of OpenNMS. Since they are generated from the GraphML standard, I do not expect them to change over the lifetime of OpenNMS unless a new version of the GraphML standard is adopted.

2.     An internal GraphProvider should not use the rest service. 
Answer -  unless there is an end point which can directly import the graphml file, the ReST service is the only option given the requirement to make the graphml files editable by the customer before deployment.

3.  "Ideally this would implement the interface GraphProvider (see SimpleTopologyProvider)"
See discussion above -  I am not trying to create a new topology provider - simply to make the exiting well designed graphml provider easier to use in a production environment without modifying any core OpenNMs code.

4.  "The GraphMLTopologyAssetProvider should be a plugin of features/topology-map/plugins".
I think the term plugin is a bit overloaded in this context.
To position this contribution as a plugin in features/topology-map/plugins would  I think be misleading and confusing. This code does not implement any of the internal topology-map API's and it does not extend the topology map code in any way. It simply generates a graphml configuration from the topology table.

This brings me to a discussion of the intent of the plugin architecture I have proposed. The intention is to allow optional features to be developed and deployed as plugins into a given OpenNMS release. The intention is to allow third parties to also be able to develop plugins on top of stable OpenNMS releases relatively decoupled from the core OpenNMS development cycle. 

These optional features should have minimal coupling to core OpenNMS code and not require any changes to a given OpenNMS release in order to be deployed. As such they should be low risk to the development cycle and to the customer since they can be un-deployed if they pose a problem. This is particularly important if we are delivering incremental customer centric functionality on top of meridian releases which only release annually. (this custoemr has been promised (OpenNMS 19 moving to meridian 2017)
This particular plug-in could be deployed by a customer on top of OpenNMS 19 in the form of a Kar or an osgi feature deployed from maven through the Karaf UI.  However I have proposed that such plugins be also offered as bundled with OpenNMS. In this case the plugin is still an optional feature but it just happens to be delivered with a particular OpenNMS release. 

I hope this makes sense.

mvrueden:
I understand, but nevertheless if the changes I suggested above are not made I vote against merging this. 

If a customer still requires it, we can find a `3rd-party repository` where we can host such integrations. As you said, it would be almost no effort to use it in a running instance of OpenNMS. This is how we dealt with custom `GraphMLProviders` for other customers as well.

pbrane:
> On Feb 26, 2017, at 11:02 AM, mvrueden <notifications@github.com> wrote:
> 
> I understand, but nevertheless if the changes I suggested above are not made I vote against merging this.
> 
> If a customer still requires it, we can find a 3rd-party repository where we can host such integrations. As you said, it would be almost no effort to use it in a running instance of OpenNMS. This is how we dealt with custom GraphMLProviders for other customers as well.

Since:

a) this is a nice feature
b) this is a feature sponsored by a support customer

I believe we should get that feature into to the project in a supportable fashion.


Let’s move this discussion to a call on, Monday.


David

gallenc:
Asset Topology Plugin appears now to be complete and awaiting final review / merge
- Added ability to store and regenerate graph definitions as well as graphs
- Documentation updated

soleger:
I'm rebasing this against foundation-2017 and will merge if it goes green in Bamboo.

mvrueden:
I reverted the merge as the "fix for ConfigAdmin not removing .cfg file entries" commit seems weird and I want to take a look before merging. Thank you.

PR #1333, opened 2017-02-24
NMS-8973: Enlinkd fails to discovery Bridge Topology

rssntn67:
NMS-8973: Enlinkd fails to discovery Bridge Topology
JIRA: http://issues.opennms.org/browse/NMS-8973


rssntn67:
The fix has been done. I need to fix the test because I have added a new check to add link over a connection. So usually we need to have that on each shared segment the mac addresses must be in the forwarding table of each bridge/port. I'm tired just now but not so far from to be finished!

rssntn67:
As far as I understand this is fixed. The bridge topology discovery has now no issues.

PR #1334, opened 2017-02-24
NMS-9145: Add JMX instrumentation for the Drools Correlator to understand the the working memory of each rule-set (a.k.a. engine)

agalue:
Add JMX instrumentation for the Drools Correlator to understand the the
working memory of each rule-set (a.k.a. engine)

* JIRA: http://issues.opennms.org/browse/NMS-9145

### DIFF
--- a/opennms-correlation/drools-correlation-engine/src/main/java/org/opennms/netmgt/correlation/drools/DroolsCorrelationEngine.java
+++ b/opennms-correlation/drools-correlation-engine/src/main/java/org/opennms/netmgt/correlation/drools/DroolsCorrelationEngine.java
@@ -55,6 +55,10 @@ import org.opennms.netmgt.correlation.AbstractCorrelationEngine;
 import org.opennms.netmgt.xml.event.Event;
 import org.springframework.core.io.Resource;
 
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.Meter;
+import com.codahale.metrics.MetricRegistry;
+
 /**
  * <p>DroolsCorrelationEngine class.</p>
  *
@@ -71,13 +75,24 @@ public class DroolsCorrelationEngine extends AbstractCorrelationEngine {
     private String m_name;
     private String m_assertBehaviour;
     private String m_eventProcessingMode;
+
+    private final Meter m_eventsMeter;
     
+    public DroolsCorrelationEngine(final String name, final MetricRegistry metricRegistry) {
+        super();
+        this.m_name = name;
+        final Gauge<Integer> size = () -> { return getMemorySize(); };
+        metricRegistry.register(MetricRegistry.name(name, "working-memory-size"), size);
+        m_eventsMeter = metricRegistry.meter(MetricRegistry.name(name, "events"));
+    }
+
     /** {@inheritDoc} */
     @Override
     public synchronized void correlate(final Event e) {
 	LOG.debug("Begin correlation for Event {} uei: {}", e.getDbid(), e.getUei());
         m_workingMemory.insert(e);
         m_workingMemory.fireAllRules();
+        m_eventsMeter.mark();
 	LOG.debug("End correlation for Event {} uei: {}", e.getDbid(), e.getUei());
     }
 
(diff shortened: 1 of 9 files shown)


opennms-correlation/drools-correlation-engine/src/main/java/org/opennms/netmgt/correlation/drools/DroolsCorrelationEngine.java
```diff
@@ -71,13 +75,24 @@
     private String m_name;
     private String m_assertBehaviour;
     private String m_eventProcessingMode;
+
+    private final Meter m_eventsMeter;
     
+    public DroolsCorrelationEngine(final String name, final MetricRegistry metricRegistry) {
+        super();
```
j-white: Nit: An explicit call to super() is not needed.

j-white:
Looks good.

This will conflict when merging forward, we'll need to replace the call to `getMemorySize()` with `getKieSessionObjects().size()`.

agalue:
> On Feb 27, 2017, at 9:30 AM, Jesse White <notifications@github.com> wrote:
> 
> Nit: An explicit call to super() is not needed.
> 
Ok! Guess I have to review my eclipse automations ;)

agalue:
> On Feb 27, 2017, at 9:33 AM, Jesse White <notifications@github.com> wrote:
> Looks good.
> 
Thanks!

> This will conflict when merging forward, we'll need to replace the call to getMemorySize() with getKieSessionObjects().size().
> 
Ok! I this true even if we are NOT using KIE session ? We’re instantiating a StatefulSession, not a KIE Session:

https://github.com/OpenNMS/opennms/blob/foundation-2016/opennms-correlation/drools-correlation-engine/src/main/java/org/opennms/netmgt/correlation/drools/DroolsCorrelationEngine.java#L161 <https://github.com/OpenNMS/opennms/blob/foundation-2016/opennms-correlation/drools-correlation-engine/src/main/java/org/opennms/netmgt/correlation/drools/DroolsCorrelationEngine.java#L161>

agalue:
> On Feb 27, 2017, at 9:33 AM, Jesse White <notifications@github.com> wrote:
> 
> we'll need to replace the call to getMemorySize()


BTW, that method was already there (in fact, since 2007 implemented by Matt), which is why I used it. I noticed that the code base in Horizon 19 is different. I think that’s why you’re mentioning this. So, does that mean we need 2 versions of this solution ? One for foundation-2016 and another for foundation2017 ?

j-white:
That was just a note for when we merge foundation-2016 into foundation-2017. We don't need another solution/PR, we can just address that during the merge.

agalue:
> On Feb 27, 2017, at 9:50 AM, Jesse White <notifications@github.com> wrote:
> 
> That was just a note for when we merge foundation-2016 into foundation-2017. We don't need another solution/PR, we can just address that during the merge.

OK!

I found that I can replace the implementation of `getMemorySize()` with `m_workingMemory.getFactCount();`. That might facilitate the manual merge (and I’ll be more than happy to deal with the manual merge once this is merged).

agalue:
It turns out that Drools has its own JMX instrumentation, only one line of code is needed (I already tested it):

```
ruleBaseConfig.setOption(MBeansOption.ENABLED)
```

Here is the source:
https://www.packtpub.com/mapt/book/application-development/9781849511964/2/ch02lvl1sec21/Monitoring+knowledge+with+JMX

Thoughts?

j-white:
I didn't realize that Drools had built-in JMX support. If they provide the metrics your looking for, I'd opt for using the built-in functionality instead of writing our own.

agalue:
> On Feb 28, 2017, at 2:50 PM, Jesse White <notifications@github.com> wrote:
> 
> I didn't realize that Drools had built-in JMX support. If they provide the metrics your looking for, I'd opt for using the built-in functionality instead of writing our own.

+1 The cool thing is that there is no need to do code changes for this. A simple property on opennms.conf is all what we need to use it.

The thing is, you’ll have only UUID to identify the sessions (which is useless). In other words, it is not possible to know the target engine of a given set of statistics. At least my stats have the engine name as part of the name. I’ve been struggling with this the whole day, and I haven't found a way to set the name or the ID of a KIE Session.

j-white:
It looks like the MBeans can only use the UUID in versions >= 6.5.0: https://issues.jboss.org/browse/DROOLS-1235

Which isn't very useful in our case. I think we'll need to provide our own then.

agalue:
OK.

So, let's keep two custom metrics with DropWizard: `fact-count` and `pending-tasks-count`. If besides that, a user wants to use the built-in stats, just add `-Dkie.mbeans=enabled` to `opennms.conf`.

PR #1340, opened 2017-02-28
NMS-9166: Fix permissions on opennms.service file for systemd

roskens:
NMS-9166: Fix permissions on opennms.service file for systemd

* JIRA: http://issues.opennms.org/browse/NMS-9166


RangerRick:
This should probably be against `foundation`, rather than `develop`, I would imagine...

soleger:
Closing this PR to create a new one against foundation.

soleger:
New PR is #1351.

PR #1341, opened 2017-02-28
NMS-9084: Add missing certAlias for ssl with jetty

roskens:
NMS-9084: Add missing certAlias for ssl with jetty

* JIRA: http://issues.opennms.org/browse/NMS-9084


RangerRick:
hm, actually, this should probably at least go back to `foundation-2017`

RangerRick:
(I went ahead and cherry-picked it there.)

j-white:
Does the behavior change at all if the system property is not set?

RangerRick:
It doesn't appear to have a default according to the Jetty docs, so it shouldn't, in theory...  :)

PR #1344, opened 2017-03-01
NMS-9134: convert categories.xsd to JAXB

RangerRick:
* JIRA: http://issues.opennms.org/browse/NMS-9134


opennms-config-model/src/main/java/org/opennms/netmgt/config/categories/package-info.java
```diff
@@ -0,0 +1,6 @@
+@XmlSchema(
```
j-white: Copyleft header.

j-white:
:+1: from me, just need to fix that merge conflict.

RangerRick:
OK, conflict is resolved, this should be mergeable.

PR #1353, opened 2017-03-01
NMS-9172: Fix the executions of Drools rules when using streaming (CEP, Temporal Reasoning)

agalue:
Fix the executions of Drools rules when using streaming (CEP, Temporal Reasoning)

* JIRA: http://issues.opennms.org/browse/NMS-9172


agalue:
This solution can be ported to `foundation-2016` but requires some love, as the structure of `DroolsCorrelationEngine` differs. As a matter of fact, in theory, it could also be ported to `foundation`.

I'm not sure how good is Drools Fusion (CEP) on Drools 5 (for `foundation`), but maybe I could cherry-pick this on `foundation-2016` and apply the minor fixes required, after merging this PR; as I was able to verify that the solution works on a VM running Meridian 2016.1.4 (and on another VM with Horizon 19.0.1).

agalue:
The JUnit test that is failing (i.e. `FusionRulesTest`, which was added a long time ago) is in theory not well implemented. When testing Fusion/CEP rules, you have to use Psudo clocks, to fully control the test; otherwise, the real time will be used. Because of this, and the fact that the test was not properly built, it fails, as the stream mode should now be working properly which was not the case when the test in question was created (i.e. the test is not considering this fact).

I could re-work this JUnit test, but I extensively tested the solution on a real working environment (with Horizon and Meridian), and several books with best practices recommend this implementation (which is why I've created this PR, beside the ability of properly use Drools Fusion / Stream Mode within OpenNMS).

j-white:
Looks good. We just need to resolve the conflicts before merging.

agalue:
> On Mar 7, 2017, at 11:09 AM, Jesse White <notifications@github.com> wrote:
> 
> Looks good.

Great!

> We just need to resolve the conflicts before merging.


Which conflicts ? This should pass clean against `foundation-2017`. Porting this to `foundation-2016` should be a manual process (and I’ll do this after this PR is merged).

j-white:
Github says otherwise.

agalue:
> On Mar 7, 2017, at 11:19 AM, Jesse White <notifications@github.com> wrote:
> 
> Github says otherwise


Done!

agalue:
Please do a squash merge (that will simplify the manual for on `foundation-2016`).

PR #1354, opened 2017-03-01
Minor cleanup

soleger:
I cleaned up a bunch of stashes and sandboxes a week ago and found a handful of minor changes worth committing.

### DIFF
--- a/core/db/src/main/java/org/opennms/core/db/DataSourceFactory.java
+++ b/core/db/src/main/java/org/opennms/core/db/DataSourceFactory.java
@@ -58,16 +58,17 @@ import org.slf4j.LoggerFactory;
  */
 public abstract class DataSourceFactory {
 
-	private static final Logger LOG = LoggerFactory.getLogger(DataSourceFactory.class);
+    private static final Logger LOG = LoggerFactory.getLogger(DataSourceFactory.class);
 
-       //private static final Class<?> DEFAULT_FACTORY_CLASS = AtomikosDataSourceFactory.class;
-       private static final Class<?> DEFAULT_FACTORY_CLASS = C3P0ConnectionFactory.class;
+    //private static final Class<?> DEFAULT_FACTORY_CLASS = AtomikosDataSourceFactory.class;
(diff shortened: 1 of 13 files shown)


core/db/src/main/java/org/opennms/core/db/DataSourceFactory.java
```diff
@@ -205,7 +206,10 @@ public static void setInstance(final DataSource ds) {
      * @param ds a {@link javax.sql.DataSource} object.
      */
     public static synchronized void setInstance(final String dsName, final DataSource ds) {
-        m_dataSources.put(dsName, ds);
+        DataSource oldDs = m_dataSources.put(dsName, ds);
+        if (oldDs != null) {
+            // TODO: Call the closer Runnable?
```
mvrueden: If this is a clean up commit, there should not be any `TODO`s left.

core/snmp/api/src/main/java/org/opennms/netmgt/snmp/InetAddrUtils.java
```diff
@@ -31,6 +31,9 @@
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 
+/**
+ * @deprecated Use {@link InetAddressUtils} instead.
+ */
 public abstract class InetAddrUtils {
```
mvrueden: Can you please also add the `@Deprecated` annotation if it is deprecated. And it is also always nice to know why it was deprecated.

core/db/src/main/java/org/opennms/core/db/DataSourceFactory.java
soleger: The TODO just tags this line as a possible problem that needs further investigation.

core/snmp/api/src/main/java/org/opennms/netmgt/snmp/InetAddrUtils.java
soleger: It is deprecated because our ```InetAddressUtils``` is a widely-used class for these sorts of operations, hence the comment.

PR #1361, opened 2017-03-04
NMS-9165: Make sure all closeable objects are closed after use.

roskens:
NMS-9165: log entry in uncategorized.log

This pull request updates /include/grafana-box.jsp to use the newer 4.3 HttpComponents HttpClient interface and ensures that any open I/O streams are closed after use.

* JIRA: http://issues.opennms.org/browse/NMS-9165


j-white:
Looks good. Pushed to `pr/jira/NMS-9165`.

PR #1364, opened 2017-03-06
NMS-9159: Alarms integration improvements

gallenc:
* JIRA: https://issues.opennms.org/browse/NMS-9159

A number of minor improvements identified by a beta user to the Alarm Change Notifier and Elastic Search Interface to make it more usable when generating reports. 
These include
    additional node asset fields in ES events - region,site,rack, location lat and long
    severity as integer type instead of text in ES index definitions
    alarm duration calculated after alarm cleared in index
    acknowledge duration calculated time to  acknowledge alarm
In addition this patch also includes significant additions to the documentation of the mapping of OpenNMS alarms and events to elastic search and some clarifications in the plugin manager documentation.

Please note that smoke tests are failing for this branch but in discussion with Seth the problem is not in this patch but upstream tests "Minion syslogs over kafka to es rest"
http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1378


opennms-base-assembly/src/main/filtered/etc/org.opennms.plugin.elasticsearch.rest.forwarder.cfg
```diff
@@ -0,0 +1,54 @@
+# org.opennms.plugin.elasticsearch.rest.forwarder.cfg
```
j-white: Since this is just documentation, I would omit adding this to the base assembly.

opennms-doc/guide-admin/src/asciidoc/text/plugins/opennms-es-rest-mapping.adoc
```diff
@@ -0,0 +1,2295 @@
...
+
+===== Table of Index Mapping
+
+The following table describes the mapping of simple
+_{opennms-product-name}_ events to the Raw Events Index and the mapping of Alarm
+Change Events to the Alarm Change Events index and  to the Alarms
+index
+++++
+
+<table width="786" cellpadding="2" cellspacing="0">
```
j-white: Shouldn't we use an asciidoc table here instead of HTML?

soleger: Yeah these tables need to be rewritten in Asciidoc, this is way too verbose.

gallenc: So the intention is to give a mapping which a user can follow which shows how OpenNMSs events end up in the elastic search indexes. This was asked for by the lead customer. After a lot of thought, the table seemed the easiest way to explain what is going on.  It is complex.
 I originally tried to do this in asciidoc but it was way too complicated to do the editing. In the end I resorted to editing the doc in libre office and generating the html for inclusion. If there was an easy way to either convert the libre office doc directly to asciidoc or editing the asciidoc table as easily as in office, I would gladly do it. Believe me I have tried.
Can you comment on the usefulness and readability of the generated documentation.

opennms-base-assembly/src/main/filtered/etc/org.opennms.plugin.elasticsearch.rest.forwarder.cfg
soleger: I'll trim this file out, it is redundant with the admin doc descriptions of the fields and I think we've agreed before in discussions with @indigo423 that moving documentation out of the bundled config files and into the proper documentation is the best way to handle this.

gallenc:
see NMS-9276 for fix to reverted changes

PR #1368, opened 2017-03-08
NMS-9201: Changed SyslogParser to use the local timezone

soleger:
Assuming that syslog timestamps are in UTC is not correct: timestamps without a timezone should be assumed to be in the local timezone of the machine.

* JIRA: http://issues.opennms.org/browse/NMS-9201


RangerRick:
Seems reasonable to me...

PR #1369, opened 2017-03-08
HZN-1034: Added synchronous execution mode to EventIpcBroadcastProcessor

soleger:
This will make the sendNowSync() call fully synchronous.

* JIRA: http://issues.opennms.org/browse/HZN-1034


pbrane:
Wow, that’s so awesome.

j-white:
We should add tests to verify that `sendNowSync` doesn't return until the event handlers are finished, or something similar.

Also, the current solution doesn't compile.

PR #1371, opened 2017-03-10
NMS-7552: PollStatus as an event

roskens:
NMS-7552: Add support for sending the pollstatus result as an event, conditional on a system property being true.

* JIRA: http://issues.opennms.org/browse/NMS-7552

http://bamboo.internal.opennms.com:8085/browse/OPENNMS-ONMS1415

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/events/opennms.pollerd.events.xml
+++ b/opennms-base-assembly/src/main/filtered/etc/events/opennms.pollerd.events.xml
@@ -618,4 +618,17 @@
         </logmsg>
         <severity>Normal</severity>
     </event>
+    <event>
+        <uei>uei.opennms.org/internal/poller/servicePollStatus</uei>
+        <event-label>OpenNMS-defined poller event: servicePollStatus</event-label>
+        <descr>
+            &lt;p&gt;Service Polling Status Event for the %service% service
+            on interface %interface% as covered by a poller package, and
+            managed.
+        </descr>
+        <logmsg dest='logndisplay'>
+            Polling Status of "%parm[statusName]%" with reason "%parm[reason]%" for %service% service on interface %interface%. Poll completed in %parm[responseTime]%ms.
+        </logmsg>
+        <severity>Normal</severity>
+    </event>
 </events>
--- a/opennms-base-assembly/src/main/filtered/etc/opennms.properties
+++ b/opennms-base-assembly/src/main/filtered/etc/opennms.properties
@@ -102,7 +102,7 @@ org.opennms.snmp.workarounds.allowZeroLengthIpAddress=true
 org.opennms.snmp.snmp4j.noGetBulk=false
 
 # Some buggy SNMP agents fail to exclude Counter64 objects from view when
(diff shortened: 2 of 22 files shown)


opennms-services/src/main/java/org/opennms/netmgt/poller/pollables/PollContext.java
```diff
@@ -78,10 +79,10 @@
      * @param address a {@link java.net.InetAddress} object.
      * @param svcName a {@link java.lang.String} object.
      * @param date a {@link java.util.Date} object.
-     * @param reason a {@link java.lang.String} object.
+     * @param params a {@link java.util.Map} object containing {@link java.ling.String} key to {@link java.ling.String} value mappings.
      * @return a {@link org.opennms.netmgt.xml.event.Event} object.
      */
-    public Event createEvent(String uei, int nodeId, InetAddress address, String svcName, Date date, String reason);
+    public Event createEvent(String uei, int nodeId, InetAddress address, String svcName, Date date, Map<String,String> params);
```
j-white: Can we just add a new method that takes a Map here instead? That way the existing calls can stay untouched.

opennms-base-assembly/src/main/filtered/etc/events/opennms.pollerd.events.xml
```diff
@@ -618,4 +618,17 @@
...
     </event>
+    <event>
+        <uei>uei.opennms.org/internal/poller/servicePollStatus</uei>
+        <event-label>OpenNMS-defined poller event: servicePollStatus</event-label>
+        <descr>
+            &lt;p&gt;Service Polling Status Event for the %service% service
+            on interface %interface% as covered by a poller package, and
+            managed.
+        </descr>
+        <logmsg dest='logndisplay'>
```
j-white: We think this should be made 'dbonly' be default in order to avoid spamming the event log in the Web UI when this is enabled.

opennms-base-assembly/src/main/filtered/etc/opennms.properties
```diff
@@ -596,6 +596,10 @@ opennms.eventlist.showCount=false
 #opennms.poller.server.serverPort=1199
 #opennms.poller.server.registryPort=1099
 
+# The enablePollStatusEvent is a flag that globally enables sending a special
```
j-white: Can we also add a note that warns users about the potential impact when enabling this on a large system monitoring thousands of services?

opennms-dao-mock/src/main/java/org/opennms/netmgt/dao/mock/EventAnticipator.java
```diff
@@ -116,6 +145,16 @@ public synchronized void anticipateEvent(Event event, boolean checkUnanticipated
         notifyAll();
     }
 
+    public synchronized void ignoreEventUei(final String uei) {
```
j-white: These changes don't appear to be used anywhere, are they still needed?

j-white:
Closing due to lack of activity. PR is still pending changes.

PR #1379, opened 2017-03-15
NMS-8601: Changed links in requisition UI to buttons and made them right-aligned

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-8601

### DIFF
--- a/opennms-webapp/src/main/scss/opennms-theme.scss
+++ b/opennms-webapp/src/main/scss/opennms-theme.scss
@@ -138,6 +138,10 @@ a:focus {
   background-color: #444;
   border-color: #333;
 }
+.btn-primary[disabled] {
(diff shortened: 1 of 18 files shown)


agalue:
Looks reasonable to me. I don't remember if there are smoke tests checking those things, but if Bamboo passes, I think this is good to be merged.

mvrueden:
I tried out this branch, looks good in general, but the font of the icon label is weird. See jira issue for more details.

mvrueden:
The buttons look good now. The only thing which still bothers me is the "tooltip" in the "more actions" popup as it exactly says the same as each item is telling me. It should be removed from my point of view.

mvrueden:
Looks good now 👍

PR #1380, opened 2017-03-15
NMS-9197, NMS-9210: Add retries, timeouts, and exception handling for Elasticsearch forwarding

soleger:
These changes fix exception handling in the ```opennms-es-rest``` feature and add ```timeout``` and ```retries``` configuration params so that ES operations can be retried if the connection to the ES cluster fails.

The commits also include an enhancement to handle multiple ES URLs by providing a comma-separated list of URLs.

* JIRA: http://issues.opennms.org/browse/NMS-9197
* JIRA: http://issues.opennms.org/browse/NMS-9210

### DIFF
--- a/features/opennms-es-rest/main-module/src/main/java/org/opennms/plugins/elasticsearch/rest/RestClientFactory.java
+++ b/features/opennms-es-rest/main-module/src/main/java/org/opennms/plugins/elasticsearch/rest/RestClientFactory.java
@@ -91,8 +91,8 @@ public class RestClientFactory {
 				timeoutTracker.startAttempt();
 				try {
 					return m_delegate.execute(clientRequest);
-				} catch (Throwable e) {
-					LOG.warn("Exception while trying to execute REST operation", e);
+				} catch (Exception e) {
+					LOG.warn("Exception while trying to execute REST operation (attempt {})", timeoutTracker.getAttempt() + 1, e);
 				}
 			}
 			return null;
--- a/features/opennms-es-rest/main-module/src/main/java/org/opennms/plugins/elasticsearch/rest/RestClientFactory.java
+++ b/features/opennms-es-rest/main-module/src/main/java/org/opennms/plugins/elasticsearch/rest/RestClientFactory.java
@@ -28,37 +28,130 @@
 
 package org.opennms.plugins.elasticsearch.rest;
 
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.opennms.core.utils.TimeoutTracker;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.searchbox.action.Action;
 import io.searchbox.client.JestClient;
 import io.searchbox.client.JestClientFactory;
+import io.searchbox.client.JestResult;
+import io.searchbox.client.JestResultHandler;
 import io.searchbox.client.config.HttpClientConfig;
 
+/**
+ * This factory wraps the {@link JestClientFactory} to provide instances of
+ * {@link JestClient}.
+ */
 public class RestClientFactory {
-	
-	private JestClientFactory factory=null;
-	
-	
-	
+
+	private JestClientFactory factory = null;
+
+	private AtomicInteger m_timeout = new AtomicInteger(0);
+	private AtomicInteger m_retries = new AtomicInteger(0);
+
+	private static class OnmsJestClient implements JestClient {
+
+		private static final Logger LOG = LoggerFactory.getLogger(OnmsJestClient.class);
+
+		private final JestClient m_delegate;
+
+		private final int m_retries;
+
+		private final int m_timeout;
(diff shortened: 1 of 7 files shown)


features/opennms-es-rest/main-module/src/main/java/org/opennms/plugins/elasticsearch/rest/RestClientFactory.java
```diff
@@ -28,37 +28,130 @@
...
+			Map<String,Object> params = new HashMap<>();
+			params.put("strict-timeout", Boolean.TRUE);
+
+			TimeoutTracker timeoutTracker = new TimeoutTracker(params, m_retries, m_timeout);
+
+			for (timeoutTracker.reset(); timeoutTracker.shouldRetry(); timeoutTracker.nextAttempt()) {
+				timeoutTracker.startAttempt();
+				try {
+					return m_delegate.execute(clientRequest);
+				} catch (Throwable e) {
```
j-white: Shoudn't we just try catching Exceptions instead? Otherwise, we may end up catching OutOfMemoryErrors, etc...

features/opennms-es-rest/main-module/src/main/java/org/opennms/plugins/elasticsearch/rest/RestClientFactory.java
```diff
@@ -28,37 +28,130 @@
...
+					LOG.warn("Exception while trying to execute REST operation", e);
```
j-white: Can we include the retry/attempt number in the log message?

features/opennms-es-rest/main-module/src/main/java/org/opennms/plugins/elasticsearch/rest/RestClientFactory.java
```diff
@@ -28,37 +28,130 @@
...
+			// attempts even if the operation returns more quickly than the timeout
+			Map<String,Object> params = new HashMap<>();
+			params.put("strict-timeout", Boolean.TRUE);
+
+			TimeoutTracker timeoutTracker = new TimeoutTracker(params, m_retries, m_timeout);
+
+			for (timeoutTracker.reset(); timeoutTracker.shouldRetry(); timeoutTracker.nextAttempt()) {
+				timeoutTracker.startAttempt();
+				try {
+					return m_delegate.execute(clientRequest);
```
j-white: Depending on the error that's thrown here, we may actually end up indexing the same documents multiple times. It may make more sense to retry only in certain scenarios, i.e. when the connection attempt fails.

For example, we can end up with "Read timed out" errors if a request takes a long time to process, and even if the request is eventually processed, we'll keep retrying.

j-white:
Looks like the error I pasted above should be solved by https://github.com/OpenNMS/opennms/pull/1378.

features/opennms-es-rest/main-module/src/main/resources/OSGI-INF/blueprint/mainModuleBlueprint.xml
```diff
@@ -29,6 +29,8 @@
       <cm:property name="archiveAlarmChangeEvents" value="true" />
       <cm:property name="archiveOldAlarmValues" value="true" />
       <cm:property name="archiveNewAlarmValues" value="true" />
+      <cm:property name="timeout" value="5000" /> <!-- 5 second timeout for Elasticsearch operations -->
```
j-white: What do you think about renaming this to `retryTimeout` instead? Since it doesn't actually affect the timeout of the client operations, i.e. the connections or the reads.

features/opennms-es-rest/main-module/src/main/java/org/opennms/plugins/elasticsearch/rest/RestClientFactory.java
soleger: The impact of what you're describing is limited to _doNotPersist_ events: any event that has an OpenNMS database ID will be indexed with that ID. If the record is indexed multiple times, you will just end up with multiple identical 'versions' of the same record but still, only a single record will be returned by ES queries.

It sounds like you're saying that we should not retry if we get a 'Read Timed Out' error but certainly in many cases when ES doesn't respond, it's likely that the records were not indexed properly. This sounds like a valid retry scenario.

features/opennms-es-rest/main-module/src/main/resources/OSGI-INF/blueprint/mainModuleBlueprint.xml
soleger: The timeout now does affect connection timeouts and I've added a ```socketTimeout``` parameter that controls the read timeout (which probably percolates down to SO_TIMEOUT inside httpclient).

features/opennms-es-rest/main-module/src/main/java/org/opennms/plugins/elasticsearch/rest/RestClientFactory.java
j-white: Make sense. If a single record is returned by the ES queries, then I don't see any problem with the current approach.

PR #1383, opened 2017-03-16
NMS-9135: Remove Castor and related dependencies

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9135

This patch removes the remaining references to Castor, now that all of the objects have been converted to use JAXB.


RangerRick:
It's a lot to digest but as far as I can tell this looks good.  👍

PR #1385, opened 2017-03-17
HZN-1020: Refactor the DNS protocol handler to implement the RequisitionProvider interface.

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1020

Here we break out the `dns://` protocol handler used to generate requisitions into it's own module and refactor it to implement the `RequisitionProvider`.

The handling the `dns://` style URLs remains the same, but the handler is now additionally available using the `requisition://` style URLs, allow it to work when using Minion.


mvrueden:
Merge conflicts (-:

opennms-provision/opennms-requisition-dns/src/main/java/org/opennms/netmgt/provision/service/dns/DnsRequisitionUrlConnection.java
```diff
@@ -0,0 +1,339 @@
...
+        }
+        return new String[] { "ICMP", "SNMP" };
+    }
+
+    /**
+     * Determine source for computing hash for foreignId from URL
+     * 
+     * @return a String of "ipAddress" or "nodeLabel"
+     */
+    private int getForeignIdHashSource() {
```
mvrueden: Can we have this return an enum instead of an integer?

PR #1386, opened 2017-03-17
NMS-8587: Add a DisplayStringIndexStorageStrategy

roskens:
NMS-8587: RRD file name very long

Add a DisplayStringIndexStorageStrategy that will take an OID subindex, and convert it to its text string representation.

* JIRA: http://issues.opennms.org/browse/NMS-8587

### DIFF
--- a/opennms-dao/src/main/java/org/opennms/netmgt/dao/support/SnmpOIDStorageStrategy.java
+++ b/opennms-dao/src/main/java/org/opennms/netmgt/dao/support/SnmpOIDStorageStrategy.java
@@ -1,0 +1,62 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2017 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2017 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License,
+ * or (at your option) any later version.
+ *
+ * OpenNMS(R) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with OpenNMS(R).  If not, see:
+ *      http://www.gnu.org/licenses/
+ *
+ * For more information contact:
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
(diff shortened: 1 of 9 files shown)


roskens:
Should I also update the datacollection/f5.xml to use this new StorageStrategy class?

Should there be a upgrade task to rename directories if we update the StorageStrategy class?

opennms-dao/src/main/java/org/opennms/netmgt/dao/support/SnmpOIDStorageStrategy.java
```diff
@@ -0,0 +1,62 @@
...
+ * along with OpenNMS(R).  If not, see:
+ *      http://www.gnu.org/licenses/
+ *
+ * For more information contact:
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.netmgt.dao.support;
```
j-white: Do we not put this in the `org.opennms.netmgt.collection.support` package like the other storage strategies because we need access to the `snmp4j` libraries?

roskens: snmp4j libraries are available to opennms-dao. I didn't have to change the pom to include them at all.

If this should be in `org.opennms.netmgt.collection.support`, then I can move it there.

jeffgdotorg: We need a better name for this strategy. May I suggest `DisplayableInstanceStorageStrategy`?

roskens: `DisplayableInstanceStorageStrategy` doesn't imply that this is only for SNMP OID values. Is there a way to include that in the name?

The net-snmp commands take an argument of `-OX` that will display an extended index format. Would `SnmpExtendedIndexStorageStrategy` work?

roskens: Could also name it `BerEncodedIndexStorageStrategy` or `OctetStringIndexStorageStrategy`.

roskens: Was just looking at the MIB for SNMPv2-TC, and saw that they call it DisplayString in there. So perhaps `DisplayStringStorageStrategy` or `DisplayStringIndexStorageStrategy`?

roskens: The F5's allow for 0-1024 characters in their index strings. They define a custom entry called LongDisplayString:

```
LongDisplayString ::= TEXTUAL-CONVENTION
    DISPLAY-HINT "1024a"
    STATUS       current
    DESCRIPTION
            "A longer version of SNMPv2-TC.DisplayString"
    SYNTAX       OCTET STRING (SIZE (0..1024))
```

jeffgdotorg:
I have no further nits to pick here.

j-white:
Reverted this since it broke the build.

PR #1387, opened 2017-03-20
NMS-9135: Remove all castor references (other than docs and opennms-tools)

RangerRick:
This PR is an additional set of changes on top of the other NMS-9135 work to remove basically all references to castor from the codebase, other than in `opennms-tools` and some old release notes mentions.

* JIRA: http://issues.opennms.org/browse/NMS-9135


soleger:
Looks good, let's merge when it passes Bamboo.

PR #1388, opened 2017-03-21
NMS-6584: Upgrade to Quartz 2.2.3

soleger:
Upgrade to the latest version of Quartz. This is a prerequisite for upgrading to Spring 4.1+.

* JIRA: http://issues.opennms.org/browse/NMS-6584


core/schema/src/main/liquibase/19.1.0/changelog.xml
```diff
@@ -142,4 +142,331 @@
...
+
+  <!-- NMS-6584: Upgrade to Quartz 2 schema -->
+  <!-- http://www.quartz-scheduler.org/documentation/quartz-2.x/migration-guide.html -->
+  <changeSet author="seth" id="19.1.0-upgrade-to-quartz-2-schema">
+
+    <!-- Drop tables that are no longer used -->
+
+    <dropTable tableName="qrtz_job_listeners"/>
+    <dropTable tableName="qrtz_trigger_listeners"/>
+
```
mvrueden: Don't we need the tables anymore? What happens if these tables are populated? Do we need some kind of upgrade task?

core/schema/src/main/liquibase/19.1.0/changelog.xml
```diff
@@ -142,4 +142,331 @@
...
+    </addColumn>
+    <sql>UPDATE qrtz_fired_triggers SET is_nonconcurrent = is_stateful;</sql>
+    <!--
+    <addNotNullConstraint columnName="is_nonconcurrent" tableName="qrtz_fired_triggers"/>
+    <addColumn tableName="qrtz_fired_triggers">
+      <column name="is_update_data" type="boolean"/>
+    </addColumn>
+    <sql>UPDATE qrtz_fired_triggers SET is_update_data = is_stateful;</sql>
+    <addNotNullConstraint columnName="is_update_data" tableName="qrtz_fired_triggers"/>
+    -->
```
mvrueden: Is this important? Can we remove this?

features/poller/remote/src/main/java/org/opennms/netmgt/poller/remote/Poller.java
```diff
@@ -147,13 +151,20 @@ private void schedulePolls() throws Exception {
...
             m_pollerFrontEnd.setInitialPollTime(polledService.getServiceId(), initialPollTime);
 
-            Trigger pollTrigger = new PolledServiceTrigger(polledService);
-            pollTrigger.setStartTime(initialPollTime);
+            /*
+            JobBuilder.newJob(PollJob.class)
+                .withIdentity(jobName)
+                .usingJobData(PollJobDetail.JOB_DATA_MAP_KEY_POLLEDSERVICE, polledService)
+                .usingJobData(PollJobDetail.JOB_DATA_MAP_KEY_POLLERFRONTEND, m_pollerFrontEnd);
+            */
```
mvrueden: Why is this here? Can we remove this?

soleger: This is an example of using the new Quartz 2 JobBuilder API. I didn't have a way to test whether it was correct so I left it commented.

core/schema/src/main/liquibase/19.1.0/changelog.xml
soleger: These are Quartz 1 tables that are gone in Quartz 2. We are only storing database report job details and cron triggers in the database.

core/schema/src/main/liquibase/19.1.0/changelog.xml
soleger: I'll remove this.

core/schema/src/main/liquibase/19.1.0/changelog.xml
```diff
@@ -142,4 +142,331 @@
...
+    <dropColumn columnName="is_stateful" tableName="qrtz_fired_triggers"/>
+
+
+    <!-- Add new 'sched_name' column to all tables -->
+
+    <addColumn tableName="qrtz_blob_triggers">
+      <column name="sched_name" type="varchar(120)" value="TestScheduler"/>
```
soleger: This is a problem: we need to replace "TestScheduler" with our durable scheduler name, "databaseReport".

soleger:
I don't think it's too much for 19.1.0. Despite the long Liquibase changelog, the changes are straightforward and the only data that will be in the database tables is job details and cron schedules (if created by the user) for the database reports. In other places that we use Quartz, we do not use durable jobs (Provisiond, Remote Poller) so nothing is stored in the database.

mvrueden:
But if the user stored data in the database, what is going to happen? 
Does the user have to migrate hat data or is it not required anymore?
If the data needs to be migrated we either should provide a upgrade task or not do it in `19.1.0` if the user must manually migrate the data.

jeffgdotorg:
If memory serves, these jobs are used only for reports scheduled  via the webapp. They're also super gross to deal with, as the contents of the rows are opaque binary data. Their presence and nature have made it difficult and complicated to migrate installations between servers, so we would be well rid of them if we're able to identify a way to help users export and import them (whether automatic or manual).

pbrane:
> On Mar 22, 2017, at 12:34 PM, Jeff Gehlbach <notifications@github.com> wrote:
> 
> If memory serves, these jobs are used only for reports scheduled via the webapp. They're also super gross to deal with, as the contents of the rows are opaque binary data. Their presence and nature have made it difficult and complicated to migrate installations between servers, so we would be well rid of them if we're able to identify a way to help users export and import them (whether automatic or manual).

And better yet, homogenize the schedules for the reporting daemon and the webapp scheduler.  Remove the scheduler running in the WebApp.

pbrane:
> On Mar 22, 2017, at 11:41 AM, Seth Leger <notifications@github.com> wrote:
> 
> In other places that we use Quartz, we do not use durable jobs (Provisiond, Remote Poller) so nothing is stored in the database.

Reportd’s durability is in the reports-configuration.


:David

core/schema/src/main/liquibase/19.1.0/changelog.xml
agalue: If those tables are gone in Quartz 2, what happen if you have "content" made with Quartz 1 that spawn to those and other tables ? Do they require migration ?

soleger:
@mvrueden I do not understand your concerns. My Liquibase migration is all that is necessary. There is no need to write an upgrade script. Quartz 2 just added a new column to all tables.

core/schema/src/main/liquibase/19.1.0/changelog.xml
soleger: We never used "job listeners" or "trigger listeners" in Quartz 1 and whatever they are, they no longer exist in Quartz 2. We only store jobs and cron triggers for database reports in the DB afaik.

features/poller/remote/src/main/java/org/opennms/netmgt/poller/remote/Poller.java
j-white: Can we mention this in a comment then?

j-white:
This worked well for me: I installed OpenNMS from `foundation-2017`, setup a scheduled database report via the WebUI, upgraded to this branch, and the report continued to fire on schedule.

soleger:
I just performed more testing and added more details to the JIRA issue. I'm going to go ahead and merge the PR because the tables that we're using (qrtz_triggers and qrtz_cron_triggers) are migrated successfully on upgrade. Also, if any of the Quartz DB tables doesn't match what your Quartz version is expecting, chances are it will throw nasty exceptions that prevent the web UI from initializing and I haven't seen problems like that in this branch.

PR #1391, opened 2017-03-21
Use java.nio.charset.StandardCharsets

soleger:
Instead of using names for character encodings, switch over to use StandardCharsets. This means that we need fewer try/catch blocks for UnsupportedEncodingException.


j-white:
It may also be worthwhile adding a Checkstyle rule that prevents string literals such as "UTF-8", referring developers to use these constants instead. The  [IllegalTokenTextCheck](http://checkstyle.sourceforge.net/apidocs/com/puppycrawl/tools/checkstyle/checks/coding/IllegalTokenTextCheck.html) rule should do the trick.

soleger:
Added checkstyle. Will merge after Bamboo success.

j-white:
Nice. :+1:

PR #1393, opened 2017-03-22
HZN-699: Added search filter for BSM admin page

christianpape:
* JIRA: https://issues.opennms.org/browse/HZN-699


mvrueden:
I tried this and the user experience is a bit lacking:
 * filtering should be done automatically when the user input is done 
 * The filter icon on the "clear filter" is misleading. I would expect the filter icon to enable filtering, but not clearing it. We should find another icon for this
 * When searching for an element in an hierarchy, e.g. searching for C in A > B > C the user has no idea what C's parents are. Maybe expanding the tree and highlighting the search result in a neutral color (no color already used by a severity) may make this more clear.

mvrueden:
Looks good to me 👍

PR #1395, opened 2017-03-23
NMS-9140: Check permissions on ACKing through ReST

fooker:
* JIRA: http://issues.opennms.org/browse/NMS-9140

### DIFF
--- a/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/AcknowledgmentRestService.java
+++ b/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/AcknowledgmentRestService.java
@@ -39,6 +39,7 @@ import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.MultivaluedMap;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
+import javax.ws.rs.core.SecurityContext;
 import javax.ws.rs.core.UriInfo;
 
 import org.opennms.core.criteria.CriteriaBuilder;
@@ -135,13 +136,22 @@ public class AcknowledgmentRestService extends OnmsRestService {
     @POST
     @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
     @Transactional
-    public Response acknowledge(MultivaluedMap<String, String> formParams) {
+    public Response acknowledge(@Context final SecurityContext securityContext, MultivaluedMap<String, String> formParams) {
         String alarmId = formParams.getFirst("alarmId");
         String notifId = formParams.getFirst("notifId");
         String action = formParams.getFirst("action");
+        String ackUser = formParams.getFirst("ackUser");
+
         if (action == null) {
             action = "ack";
         }
+
+        if (ackUser == null) {
+            ackUser = securityContext.getUserPrincipal().getName();
+        }
+
+        AlarmRestService.assertUserEditCredentials(securityContext, ackUser);
+
         OnmsAcknowledgment ack = null;
         if (alarmId == null && notifId == null) {
             return getBadRequestResponse("You must supply either an alarmId or notifId");
@@ -156,7 +166,7 @@ public class AcknowledgmentRestService extends OnmsRestService {
             if (alarm == null) {
                 return Response.notModified().build();
             }
-            ack = new OnmsAcknowledgment(alarm);
+            ack = new OnmsAcknowledgment(alarm, ackUser);
         } else if (notifId != null) {
             final Integer numericNotifId = getNumericValue(notifId);
             if (numericNotifId == null) {
@@ -166,7 +176,7 @@ public class AcknowledgmentRestService extends OnmsRestService {
             if (notification == null) {
                 return Response.notModified().build();
             }
-            ack = new OnmsAcknowledgment(notification);
+            ack = new OnmsAcknowledgment(notification, ackUser);
         }
         
         if ("ack".equals(action)) {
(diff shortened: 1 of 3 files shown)


mvrueden:
Why is it tagged as `backport` but the base branch is `develop`? It should probably be merged into `foundation-2017` instead.

agalue:
> On Mar 23, 2017, at 7:09 AM, mvrueden <notifications@github.com> wrote:
> 
> Why is it tagged as backport but the base branch is develop? It should probably be merged into foundation-2017 instead.

I totally agree.

And as a side note, the acks end-point is not the only one that suffers the same permission problem.

j-white:
@agalue can you elaborate on the other REST endpoints you know of that have the same problem?

agalue:
> On Mar 23, 2017, at 8:26 AM, Jesse White <notifications@github.com> wrote:
> 
> @agalue <https://github.com/agalue> can you elaborate on the other REST endpoints you know of that have the same problem?


Absolutely, but as this is a PR for the ACK end-point, I think we should discuss this on Jira (of course, after creating an issue for each end-point with problem, or creating a common Jira issue for all end-points). Let me know what way do you prefer.

j-white:
Okay. Let's focus on the `/acks` endpoint here then.

I'd opt for creating a single issue with the remaining endpoints, and we can then assess if we need to break them out into multiple issues or not.

agalue:
> On Mar 23, 2017, at 8:34 AM, Jesse White <notifications@github.com> wrote:
> 
> Okay. Let's focus on the /acks endpoint here then.
> 
> I'd opt for creating a single issue with the remaining endpoints, and we can then assess if we need to break them out into multiple issues or not.
> 
Perfect!

fooker:
The code has been written against foundation-2017, only the PR was wrong. I changed the PR accordingly.

mvrueden:
You are probably going to kill me, but if this is a security fix, shouldn't this be targeted for `foundation-2016` instead? Or even `foundation`?

agalue:
> On Mar 28, 2017, at 9:51 AM, mvrueden <notifications@github.com> wrote:
> 
> You are probably going to kill me, but if this is a security fix, shouldn't this be targeted for foundation-2016 instead? Or even foundation?


I have to agree with Markus here, as this is indeed a security fix.

In fact, I’m wondering if there is a way to either use the @Secured annotation that Spring Security provides, or even better, enforce security through applicationContext-spring-security.xml for any PUT/POST/DELETE request to make sure that a user should not have ROLE_READONLY, and should have either ROLE_ADMIN or ROLE_REST (or ROLE_PROVISION in terms of the requisitions) in order to perform these queries. That way we can fix ALL the end-points at once, instead of having separate PR like this one for each end-point, because a read-only users should not ack alarms, but it should also be rejected when modifying nodes, delete interfaces, change snmp credentials, modify requisitions, etc.

mvrueden:
If spring does not support what we need, it is very easy to implement our own aspect to enforce our rules.

agalue:
> On Mar 28, 2017, at 9:58 AM, mvrueden <notifications@github.com> wrote:
> 
> If spring does not support what we need, it is very easy to implement our own aspect to enforce our rules
> 
It does, we’re just not using the full capabilities of Spring Security. We can definitively enforce our rules, but if Spring Security was designed for this, why not using it ? In fact, I learnt how to use @Secured with SPEL in Spring 3 years ago.

fooker:
I changed the base to `foundation-2016`. If this should be backported to `foundation`, this needs to be a separate commit / PR.

About the `@Security` annotation: I would prefer to do this in some kind of _Rework ReST  Security_ task.

agalue:
> On Mar 28, 2017, at 10:22 AM, Dustin Frisch <notifications@github.com> wrote:
> I changed the base to foundation-2016. If this should be backported to foundation, this needs to be a separate commit / PR.
> 
Sounds good, as I know that something made for `foundation` cannot be automatically merged to `foundation-2016` or vice versa.
> About the @Security annotation: I would prefer to do this in some kind of Rework ReST Security task.
> 
I’m just trying to save time here, as I was planning to create a Jira issue for each end point that requires a similar treatment but it turns out that any end-point that allows DELETE, POST, PUT will required what you just did with the acks end-point.

So, do you think it worth patching every single end-point individually ? Or, make a solution that covers all of them considering that all share the same access rules ?

I’m good with both, but the second option sounds better.

fooker:
> I’m good with both, but the second option sounds better.
Indeed. Especially, when it comes to security having a good framework and avoiding cumbersome repetition is crucial.

I just want to move the discussion to the overall story issue (can you provide a link to the story for reference?) and proceeding with this PR as it is for now.

agalue:
> On Mar 28, 2017, at 4:54 PM, Dustin Frisch <notifications@github.com> wrote:
> 
> I just want to move the discussion to the overall story issue (can you provide a link to the story for reference?) and proceeding with this PR as it is for now.

You mean, discuss on Jira ?

Here’s what I’m talking about:
http://docs.spring.io/spring-security/site/docs/current/reference/html/el-access.html

The @Secured is an alternative but in code. Although, I prefer doing it on the Spring Security context, because that way the user can alter it according to their needs. Our current solution (and check AlarmRestService as a reference), or @Secured are examples of "hard-coded" solution that will require code changes in order to alter the access.

fooker:
> You mean, discuss on Jira ?
Yup. I want to get the PR closed / merged and would like to avoid to document the discussion about a general solution somewhere in a PR only related to a single endpoint.

agalue:
> On Mar 29, 2017, at 4:08 AM, Dustin Frisch <notifications@github.com> wrote:
> 
> Yup. I want to get the PR closed / merged and would like to avoid to document the discussion about a general solution somewhere in a PR only related to a single endpoint.

Sounds good. As this problem affects all the end-points with POST/PUT/DELETE, I would say close this PR, even close the Jira issue, and then we can create a "generic" issue to cover all the cases at once and discuss the solution there.

PR #1398, opened 2017-03-27
NMS-9244: Remove files from repository which are in .gitignore

indigo423:
Remove .gitignore files which are excluded from .gitignore in the root directory but checked in into the source code repository.

* JIRA: http://issues.opennms.org/browse/NMS-9244


core/icmp-jna/c-ping/.gitignore
```diff
@@ -1,21 +0,0 @@
-*.cache
```
soleger: Don't these exclusions need to stay in place to exclude the output files of the C build from getting checked in?

features/opennms-gwt-theme/.gitignore
```diff
@@ -1 +0,0 @@
-src/main/webapp/
```
soleger: Are these files regenerated during each build or was ```src/main/webapp``` added to this .gitignore by mistake?

features/system-report/src/.gitignore
```diff
@@ -1 +0,0 @@
-META-INF
```
soleger: Are these files regenerated during each build or was ```src/META-INF``` added to this .gitignore by mistake?

tools/packages/fink/build.properties
```diff
@@ -1,45 +0,0 @@
-###############################################################################
```
soleger: It seems like removing this will break the (unmaintained) fink build.

tools/packages/solaris/build.properties
```diff
@@ -1,120 +0,0 @@
-###############################################################################
```
soleger: It seems like removing this will break the (unmaintained) Solaris build.

soleger:
Also, I would revert any changes inside opennms-tools since it will just bit-rot those projects (since they're not build/tested as part of the main build). Some of the projects in opennms-tools with .gitignore problems have already been removed in foundation-2017 and later anyway.

indigo423:
@soleger did it on develop to get the repo cleaned up.

RangerRick:
FYI, I just rebased this against latest develop and set it back up in bamboo to make sure it still works.

tools/packages/fink/build.properties
RangerRick: we should delete the fink build altogether, I doubt it works, and it's definitely unmaintained

tools/packages/solaris/build.properties
RangerRick: we should delete the solaris build altogether, I doubt it works, and it's definitely unmaintained

PR #1401, opened 2017-03-27
NMS-8725, NMS-8726: Parser that can parse all syslog message formats

soleger:
This PR introduces a new SyslogParser called the RadixTreeSyslogParser that can process syslog messages using a set of pattens instead of a single regex pattern. The patterns are specified using a simplified 'grok' syntax. The parser uses a radix tree data structure internally to store the parsing tree.

Because many patterns can be processed at once, the parser can parse syslog messages in many different formats (RFC 3164, RFC 5424, plus many variants of datestamp format and vendor-specific formatting). 

Advantages:
- Agents that send messages in different formats do not have to all be tuned to send in a uniform format.
- You don't have to use a relay like syslog-ng to normalize the formats before sending to OpenNMS.
- All fields in the syslog header will be captured reliably including message ID, time zone, fractional-second values, etc.

To switch to the new parser, just change the ```parser``` parameter in syslogd-configuration.xml to ```org.opennms.netmgt.syslogd.RadixTreeSyslogParser```.

* JIRA: http://issues.opennms.org/browse/NMS-8725


j-white:
Docs?

soleger:
There aren't really any docs necessary for this pass, all you can do is switch the parser implementation over like you would to the old Rfc5424SyslogParser.

pbrane:
> On Mar 30, 2017, at 10:24 AM, Seth Leger <notifications@github.com> wrote:
> 
> There aren't really any docs necessary for this pass, all you can do is switch the parser implementation over like you would to the old Rfc5424SyslogParser.

From the technical perspective, it would be nice to document the details of the engineering decisions, design, and/or patterns used by the new parser.  I think it’s cool and significant enough that it’s warranted.

:David

PR #1402, opened 2017-03-28
NMS-8135: Make deletion great again

fooker:
This reworks deletion of nodes in the Web UI and the ReST API by just sending events and let providiond do the work. This also changes the way node, interface and service deletion is handled and make it the same for all:
* Elements are deleted independent of the state of discovery and if requisition is enabled
* If the parent element will become empty it will be deleted, too
* Events will be send only once for each element

This also removes the need of the `enableDeletionOfRequisitionedEntities` property.

* JIRA: http://issues.opennms.org/browse/NMS-8135


indigo423:
@fooker make your tests great again

mvrueden:
@fooker requested a `use case` review on this yesterday. @agalue @jeffgdotorg feel free to jump in any time (-:

opennms-base-assembly/src/main/filtered/etc/opennms.properties
```diff
@@ -444,14 +444,6 @@ smslib.serial.polling=true
 # setting is true (See org.opennms.netmgt.provision.service.ProvisionService)
 #org.opennms.provisiond.enableDiscovery=true
 
-# Prior to 1.10 it was possible to delete entities that have been provisioned as
-# part of a provisioning group.  In 1.10 we have disabled this so that in order to
-# delete these entities you have to go back to the provisioning group and delete them
-# from there.  To reenable this deletion you can set this to true.
-# NOTE: if you do this then the object will be recreated when the provisioning group is
-# next imported/synchronized
-org.opennms.provisiond.enableDeletionOfRequisitionedEntities=true
```
agalue: What would happen on those users that have the "old" default ? (i.e. enableDeletionOfRequisitionedEntities=false).

agalue:
Have you tried if the solution works regardless how the node has been added ? I mean, does it works for auto-discovered nodes, in the same way it works for requisitioned nodes ? Remember that auto-discovered nodes are not part of a requisition (i.e. foreignSource == foreignId == null)

opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/DefaultProvisionService.java
```diff
@@ -306,218 +301,75 @@ private void updateNodeHostname(final OnmsNode node) {
...
-
-        if (isDiscoveryEnabled()) {
-            LOG.debug("deleteService: discovery is enabled");
-            final String foreignSource = monSvc.getForeignSource();
-            final String foreignId = monSvc.getForeignId();
-            OnmsNode requisitionedNode = null;
-            if (foreignSource != null && foreignId != null) {
-                requisitionedNode = getRequisitionedNode(foreignSource, foreignId);
-            }
+    public void deleteService(final Integer nodeId, final InetAddress addr, final String svcName) {
```
agalue: What about all those useful log messages that are not there anymore ?

opennms-base-assembly/src/main/filtered/etc/opennms.properties
fooker: It will be ignored completely.

fooker:
I tried requisitioned nodes and nodes added using quick-add. I tested also for discovered and provisioned interfaces and services. If you can provide a test-matrix I'm happy to do some more testing.

j-white:
I tested this on my end, it worked well against both provisioned and discovered nodes.

I'm wondering what the original motivation was behind the `enableDeletionOfRequisitionedEntities` flag though: https://github.com/OpenNMS/opennms/commit/1ad48021a1d62c0dcd8e5b98a39908692c8f293d

Maybe it was still relevant when `capsd` was still around?

j-white:
After discussing this, we can't think of any good reason to keep the `enableDeletionOfRequisitionedEntities` flag. The only caveat is that if a service belonging to a requisition node is manually deleted, it may reappear when the node is re-synchronized unless it is also removed from the corresponding requisition and is no longer detected.

PR #1403, opened 2017-03-28
NMS-9248: Make the location field on the requisition UI to be a "select" instead of an "input" field.

agalue:
Make the location field on the requisition UI to be a "select" instead of an "input" field.

* JIRA: http://issues.opennms.org/browse/NMS-9248

Instead of a select, I've used a non-editable typeahead (because it looks nicer). Putting free text will make the field invalid, and you won't be able to save the node. Removing the content is equivalent to set the value to null (because no selection has been made). The options will display all the available locations, excluding "Default", and the field is disabled if there are no locations, to avoid confusions, and a small message will be shown below the field.


opennms-webapp/src/main/webapp/js/onms-requisitions/scripts/services/Requisitions.js
```diff
@@ -1184,6 +1185,36 @@
...
+    */
+    requisitionsService.getAvailableLocations = function() {
+      var deferred = $q.defer();
+      var url = requisitionsService.internal.monitoringLocationsUrl;
+      $log.debug('getAvailableLocations: getting available locations');
+      $http.get(url)
+      .success(function(data) {
+        $log.debug('getAvailableLocations: got available locations');
+        var locations = [];
+        angular.forEach(data.location, function(loc) { if (loc['location-name'] != 'Default') locations.push(loc['location-name']); });
```
soleger: If we exclude "Default" then it is impossible to assign a node to the "Default" location isn't it? The value needs to either be "Default" or ```null```. So it seems like we need to keep "Default" here (and make sure that it is the first entry in the list, followed by the rest in alphabetical order).

agalue:
> On Apr 5, 2017, at 5:06 PM, Seth Leger <notifications@github.com> wrote:
> 
> If we exclude "Default" then it is impossible to assign a node to the "Default" location isn't it? The value needs to either be "Default" or null. So it seems like we need to keep "Default" here (and make sure that it is the first entry in the list, followed by the rest in alphabetical order).


I understand that if it is null, it is equivalent to physically assign Default on the requisition (in terms of how OpenNMS will behave after synchronize.In fact the OnmsNode entities will have Default when you have null on the requisition). So, do we need to allow the ability to manually set Default ?

j-white:
The `Default` location doesn't *need* be explicitly set, the nodes will be associated with the `Default` location when this field is null or empty.

j-white:
When editing a node, we renamed 'Location' to 'Minion Location', should we do the same when look at the list of nodes in the requisition?

agalue:
> On Apr 6, 2017, at 11:56 AM, Jesse White <notifications@github.com> wrote:
> 
> The Default location doesn't need be explicitly set, the nodes will be associated with the Default location when this field is null or empty.

Exactly, that’s why I don’t show Default as an option, because by default, if you "clear" the content of a typeahead field that is not editable (like this one), its value should be "null", and will be invalid if the text doesn’t match one of the options. But, if there is value on making what Seth suggested, I can do it. Let me know how you guys want to proceed.

agalue:
> On Apr 6, 2017, at 11:57 AM, Jesse White <notifications@github.com> wrote:
> 
> When editing a node, we renamed 'Location' to 'Minion Location', should we do the same when look at the list of nodes in the requisition?

I’ve already done that. Please take a look at node-basic.html

j-white:
I was referring to: https://github.com/OpenNMS/opennms/blob/907fa3f6d02581a937c5a0e1c1c4c017178b1e0c/opennms-webapp/src/main/webapp/js/onms-requisitions/views/requisition.html#L70

agalue:
> On Apr 6, 2017, at 12:11 PM, Jesse White <notifications@github.com> wrote:
> 
> I was referring to: https://github.com/OpenNMS/opennms/blob/907fa3f6d02581a937c5a0e1c1c4c017178b1e0c/opennms-webapp/src/main/webapp/js/onms-requisitions/views/requisition.html#L70 <https://github.com/OpenNMS/opennms/blob/907fa3f6d02581a937c5a0e1c1c4c017178b1e0c/opennms-webapp/src/main/webapp/js/onms-requisitions/views/requisition.html#L70>
Done!

soleger:
Sorry, my suggestion from above re: "Default" in the list assumed that you were using a select instead of the typeahead (I didn't read the description first). Is there a way to easily see the list of all locations without typing? If not, I think a select is more usable.

agalue:
> On Apr 6, 2017, at 8:59 PM, Seth Leger <notifications@github.com> wrote:
> 
> Is there a way to easily see the list of all locations without typing? If not, I think a select is more usable.


Yes, as soon as the element is on focus, all the available locations will be displayed. That’s because the field has typeahead-min-length="0":
https://angular-ui.github.io/bootstrap/#!#typeahead <https://angular-ui.github.io/bootstrap/#!%23typeahead>

soleger:
typeahead-min-length=0 sounds good to me, thanks @agalue!

PR #1404, opened 2017-03-29
HZN-1021: Update the Vmware requisition handler to implement the RequisitionHandler interface

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1021

Here we update the Vmware requisition handler to support the new `requisition://` protocol, while maintaining backward compatibility with the existing `vmware://` protocol style URLs.

With this in place, it should now be possible to provision and monitor Vmware servers using Minion.


j-white:
Tested against 2 separate vCenter instances.

PR #1406, opened 2017-03-30
NMS-9259: Upgrade to Spring 4.1.6

soleger:
Bump Spring to version 4.1.6. There are no code changes required for this upgrade, only OSGi metadata bumps, Spring schema version bumps, and the version change in pom.xml.

JIRA: https://issues.opennms.org/browse/NMS-9259

### DIFF
--- a/container/features/src/main/resources/features.xml
+++ b/container/features/src/main/resources/features.xml
@@ -363,7 +363,7 @@
 
     <feature name="spring-security32" description="Spring :: Security" version="${springSecurityVersion}">
       <!-- spring-web is necessary for HTTP remoting -->
-      <feature version="[4.0,4.1)">spring-web</feature>
+      <feature version="[4.1,4.2)">spring-web</feature>
       <feature>commons-codec</feature>
 
       <bundle>wrap:mvn:org.springframework.ldap/spring-ldap-core/1.3.2.RELEASE$Export-Package=org.springframework.ldap*;version="1.3.2"</bundle>
@@ -409,8 +409,8 @@
     
     
     <feature name="opennms-core" description="OpenNMS :: Core" version="${project.version}">
-      <feature version="[4.0,4.1)">spring</feature>
-      <feature version="[4.0,4.1)">spring-orm</feature>
+      <feature version="[4.1,4.2)">spring</feature>
+      <feature version="[4.1,4.2)">spring-orm</feature>
 
       <feature>commons-io</feature>
       <feature>dnsjava</feature>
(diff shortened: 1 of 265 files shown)


pbrane:
Now *that's* a lot of XML!  ;)

David

indigo423:
We do not track this change in JIRA with versioning?

soleger:
My rationale for not opening a JIRA issue:
* The change has no user impact.
* It's not addressing a bug.
* I reviewed the Spring release notes and didn't see anything that introduced extra risk for our project.
* It required zero lines of code changes.

The only benefit of upgrading is to get our version of Spring to line up better with newer versions of Apache ActiveMQ and Apache Camel.

j-white:
Haven't we always tracked any non-trivial change in JIRA?

indigo423:
> The change has no user impact

A user IMHO is someone who is installing, running and contributing to the project. As soon we have to deal with bugs, issues we should track more detailed this type of changes. Especially with the dependency complexity of our project.

> It's not addressing a bug.
Should we do not use JIRA everything else then "bugs"?

> I reviewed the Spring release notes and didn't see anything that introduced extra risk for our project.
I'm ok with that which is probably the reason why it goes into a minor release

> It required zero lines of code changes.
As all dependency bumps :)

pbrane:
Ronny, I'm sure it's just a language thing but I'm unsure whether your for or against whether or not an issue should be created. 

David

soleger:
My assessment was that this was a trivial change that didn't require tracking. Ronny's assessment is that it is not trivial and needs to be tracked. Ronny, can you create an issue that satisfies your requirements?

j-white:
A trivial change affecting a core framework and touching 265 files?

mvrueden:
From my point of view this is far from a trivial fix.
We touch a lot of files, and even if the spring guys say, it should not have any effects, and even if no code changes required, in the end there may be issues which we have to support. 

This beeing said, I do not understand why all of the sudden everything is targeted for `foundation-2017`. This may be a different discussion, but I don't like our approach of upgrading versions of major/core features (spring, hibernate, etc) in a minor release. But opinions differ (-:

indigo423:
@soleger done

pbrane:
> On Apr 4, 2017, at 9:42 AM, mvrueden <notifications@github.com> wrote:
> 
> This beeing said, I do not understand why all of the sudden everything is targeted for foundation-2017. This may be a different discussion, but I don't like our approach of upgrading versions of major/core features (spring, hibernate, etc) in a minor release. But opinions differ (-:

Foundation 2017 is a major release.

PR #1410, opened 2017-04-03
NMS-9126: Allow string attribute to be persisted over the TCP stream

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9126

- Create a new persistence strategy that only send data to the TCP stream, without any persisting of data to the RRD or other physical database.  This allows us to get rid of persistence code related to DB writing.
- Allow for the new TCP persistence strategy to persist both numeric and string values to the TCP stream.
- Update the performance data protobuf message to hold a lit of double values and a list of string messages.


tdefilip:
Thanks!

PR #1411, opened 2017-04-04
NMS-9252: Upgrade Newts to 1.4.3

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9252

This adds support for globally changing the behavior of counter wraps using a system property. By default, we maintain the existing behavior, but we now have the option to ignore counter wraps and replace these with NaNs, by setting a system property.

Example screenshots are available in the JIRA issue.


agalue:
On a VM on which I have some spikes, I've upgraded Newts, and added `org.opennms.newts.nan_on_counter_wrap=true` to `opennms.properties`. After a restart, the spike is not rendered on the graph.

PR #1412, opened 2017-04-04
NMS-9095: Move the poller-test command to the Karaf shell.

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9095

Here we convert the `poller-test` command in `$OPENNMS_HOME/bin` to a Karaf shell command. Aside from being much faster, this also helps prevent problems such as those listed in the bug.


agalue:
Looks good to me. I'll compile the branch and test it.

agalue:
Works for me! As soon as Bamboo goes green, this can be merged.

PR #1413, opened 2017-04-04
NMS-9250: clean up JAXB classes

RangerRick:
This is a massive patch to clean up a huge number of issues in the JAXB classes and "modernize" them.

### Changes

It does a number of things to the `opennms-config-model` package:

* Change a few XSDs that did not match our default configurations (mostly making some things optional or non-ordered)
* Make return values that are optional return Java 8 `Optional` when possible
* Remove TONS of extraneous Castor code (multiple ways to set/get arrays, `hasXXX` and `isXXX` methods that are duplicates of `getXXX` methods, etc.)
   This included getting rid of `enumerateXXX` and `iterateXXX` which were often used for no discernible reason.
* Get rid of extraneous plural objects generated by Castor.  For example:
  ```xml
   <foo>
      <filters>
         <filter></filter>
      </filters>
   </foo>
   ```
   ...would create a `Filters` object that contains a single field with a list of `Filter` objects, but instead the `Foo` object could use `@XmlElementWrapper(name="filters")` and  `@XmlElementWrapper(name="filter")` to directly reference the list of `Filter` objects without the intermediate object. ([example](https://github.com/OpenNMS/opennms/blob/features/jaxb-optionals/opennms-config-model/src/main/java/org/opennms/netmgt/config/ackd/AckdConfiguration.java#L88))
* Changed a number of `String` properties that should really be `Boolean`, ie, pattern-match required a string of "true" or "false" in the XSD 😐 ([example](https://github.com/OpenNMS/opennms/blob/features/jaxb-optionals/opennms-config-model/src/main/java/org/opennms/netmgt/config/notificationCommands/Argument.java#L51))
* Changed a number of `String` properties that should really be `enum`, ie, pattern-match required a limited number of values ([example](https://github.com/OpenNMS/opennms/blob/features/jaxb-optionals/opennms-config-model/src/main/java/org/opennms/netmgt/config/threshd/ThresholdType.java#L35))

### Concerns

While I'm pretty confident in the *idea* of these changes, I'm super uncomfortable with the sheer amount of code that was modified.  The JSPs compile and the smoke tests pass, but there are a lot of holes in our coverage. We can't rely on the compiler to catch places where, say, an `Optional<String>` is being compared using `Object.equals()` against a string object, because that return value used to be a straight string.  I attempted to audit for these things and did a lot of "find references in workspace" kind of stuff, but I'd bet there are still places I missed.

On the other hand, it would be great to have all these cleanups make it into Meridian 2017. I lean towards going ahead and merging these changes to `foundation-2017`, but then calling the next release `20.0.0` and bumping develop to `21.0.0`, rather than making a `19.1.0` release.

Thoughts?

### JIRA

The associated JIRA issue is here: http://issues.opennms.org/browse/NMS-9250


RangerRick:
Yeah, I kinda... just kept going.  :D

PR #1414, opened 2017-04-04
NMS-9203: Discard expired messages instead of using the DLQ

soleger:
To avoid filling up the kahadb storage limit with expired messages, we should just discard them. Users can reenable the DLQ easily by commenting out the section of ```opennms-activemq.xml```.

* JIRA: http://issues.opennms.org/browse/NMS-9203


j-white:
I've confirmed this work too. Waiting on Bamboo.

PR #1416, opened 2017-04-04
NMS-9261: Make org.opennms.newts.nan_on_counter_wrap=true by default

agalue:
Make org.opennms.newts.nan_on_counter_wrap=true by default

* JIRA: http://issues.opennms.org/browse/NMS-9261

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/opennms.properties
+++ b/opennms-base-assembly/src/main/filtered/etc/opennms.properties
@@ -676,6 +676,9 @@ excludeServiceMonitorsFromRemotePoller=DHCP,NSClient,RadiusAuth,XMP
 #org.opennms.newts.config.cache.redis_hostname=localhost
 #org.opennms.newts.config.cache.redis_port=6379
 
+#Disables the processing of counter wraps, replacing these with NaNs instead.
+org.opennms.newts.nan_on_counter_wrap=true
+
 ###### Evaluate #####
 # Use these properties to configure persistence using Evaluate
 # Note that Evaluate must be enabled using the 'org.opennms.timeseries.strategy' property


j-white:
Just so we're on the same page, this will cause a `.rpmnew` file for `opennms.properties` in 2016.1.5.

agalue:
Correct!
That being said, I received a request to have this as the default behavior, hence this PR.

pbrane:
> On Apr 4, 2017, at 2:34 PM, Alejandro Galue <notifications@github.com> wrote:
> 
> Correct!
> That being said, I received a request to have this as the default behavior, hence this PR.

Hmm…  Interesting point.  I guess that’s something I didn’t consider.


David

j-white:
What's the plan here?

agalue:
> On May 12, 2017, at 4:06 PM, Jesse White <notifications@github.com> wrote:
> 
> What's the plan here?

It is up to you guys, but in my personal opinion if the property is not true by default, then we’re not fixing NMS-9261, as the "current" behavior is incorrect, even if that means having .rpmnew files.

j-white:
Isn't the current behavior the same as rrdtool's? A spike is generated when a counter is reset.

agalue:
> On May 12, 2017, at 4:17 PM, Jesse White <notifications@github.com> wrote:
> 
> Isn't the current behavior the same as rrdtool's? A spike is generated when a counter is reset.

In RRD that is not a problem because of the heartbeat. Once a sysUpTime change is detected, that sample is not inserted by Collectd, so when the next one (i.e. next collection attempt) comes in, the counters are rearranged (i.e. the last value is updated), so no spike will be added. In other words, RRDtool/JRobin doesn’t suffer of this, it is just a Newts problem.

j-white:
Right, when using Newts the sample won't be inserted either (since collectd behaves the same way), but the heartbeat handling is slightly different.

We could also look at changing the default without touching the configuration files. Once the .properties files are loaded, we can manually set the system property if it was not already set. This would be the right spot to do that:
    https://github.com/OpenNMS/opennms/blob/foundation-2016/core/daemon/src/main/java/org/opennms/netmgt/vmmgr/Starter.java#L171

In {{foundation-2017}} we can then revert that change, and update the configs instead.

agalue:
Sounds good to me.

pbrane:
+1

j-white:
Closing in favor of https://github.com/OpenNMS/opennms/pull/1495

PR #1421, opened 2017-04-05
NMS-9205: Reload resource types if datacollection config changes

fooker:
* JIRA: http://issues.opennms.org/browse/NMS-9205

This fixes the reloading, but one must still trigger the reload of the datacollection config by touching `datacollection-config.xml`.


j-white:
Looks good, this should be fixed in `foundation-2016` though.

j-white:
I've rebased this on top of `foundation-2016`, waiting for Bamboo.

PR #1424, opened 2017-04-05
NMS-9132: Upgrade log4j2 to latest release 2.8.2

roskens:
This PR updates log4j2 to 2.8.1. Also, change the dependency management to use the log4j-bom which pins all log4j2 dependencies to the same version. (From https://logging.apache.org/log4j/2.x/maven-artifacts.html)

The big list of changes since 2.5 can be seen here: https://logging.apache.org/log4j/2.x/changes-report.html

* JIRA: http://issues.opennms.org/browse/NMS-9132


soleger:
Created a local branch rebased against foundation-2017. Will merge if it passes Bamboo.

soleger:
The branch doesn't build:
```
features/poller/remote/src/main/java/org/opennms/netmgt/poller/remote/support/Log4j2StringAppender.java:[117,70] unreported exception java.io.IOException; must be caught or declared to be thrown
```

roskens:
Add a fix for the IOException compile error. I've rebased against develop and force pushed that to jira/NMS-9132.

PR #1425, opened 2017-04-06
NMS-7548: Assure that resourceIds are only decoded once

christianpape:
https://issues.opennms.org/browse/NMS-7548


j-white:
Can we extend the [GraphRestServiceIT](https://github.com/OpenNMS/opennms/blob/foundation-2016/opennms-webapp-rest/src/test/java/org/opennms/web/rest/v1/GraphRestServiceIT.java) to make sure that resources special characters, such at the `+` reported in the bug, continue to work properly?

christianpape:
I believed this was only a Webapp issue but it seems that it is a more general problem since it does not work in the ReST services either. I'll update the PR as soon I've a valid solution...

christianpape:
I updated the PR. Let's wait for Bamboo :-)

christianpape:
Closing this PR, since the fix will not work in all cases. So, I start to refactor the encode/decode resourceId handling. As soon as I finish I'll open a new PR...

PR #1427, opened 2017-04-06
NMS-8908: Added null check when NRTG metrics cannot be found

soleger:
This will avoid crashing the NRTG graphs when we cannot find metrics that we expect inside the DefaultCollectionJob.

* JIRA: http://issues.opennms.org/browse/NMS-8908


j-white:
What's the affect on the graph when these series are missing? Is there any way to identify the missing metrics in the log? (Perhaps we should log a warning when a null is detected here). Also, this would probably be a good candidate for an integration test.

soleger:
This is only a partial bugfix that will at least prevent the page from failing in the ```foundation``` branch. Please update the JIRA issue with further discussion about the issue.

PR #1428, opened 2017-04-07
NMS-9223: Optimize SNMP4J session creation.

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9223

Optimize the way SNMP4J sessions are created, avoid unnecessary calls to `SecurityProtocols.getInstance().addDefaultProtocols()`.


j-white:
I tested this patch by attempting to collect metrics using both SNMPv2 and SNMPv3, which worked properly, but I'd appreciate a second pair of eyes.

PR #1436, opened 2017-04-13
HZN-1047: Added ability to override patterns for RadixTreeSyslogParser

soleger:
A customized set of patterns can be put into ```$OPENNMS_HOME/etc/syslogd-grok-patterns.txt```.

* JIRA: http://issues.opennms.org/browse/HZN-1047


j-white:
Should we not provide some type of reference documentation for the Grok statements supported in `syslogd-grok-patterns.txt` and their behavior?

PR #1445, opened 2017-04-18
NMS-7548: Encode/decode resourceId only for the web layer

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-7548


j-white:
We may also need to update the routines that parse the resource id, since square brackets will no longer be encoded. i.e. `node[FS:funky[foreignId+]].nodeSnmp[]`

opennms-model/src/main/java/org/opennms/netmgt/model/ResourceId.java
```diff
@@ -0,0 +1,127 @@
+package org.opennms.netmgt.model;
```
j-white: Missing copy-left header.

j-white:
I gave this a go, and it works well.

We do risk breaking existing queries to the Measurements API though. I verified that the Grafana plugin is not affected, unless there are special characters in the FS/FID.

I'd suggest we only make these changes in the next major release i.e. 20.0.0.

PR #1446, opened 2017-04-18
NMS-9225: Prevent Kafka Heartbeat consumer from failing on startup

soleger:
Catch exceptions when receiving messages from Kafka so that the consumer thread isn't stopped. Added system tests that restart Minion and OpenNMS and verify that heartbeats resume.

* JIRA: http://issues.opennms.org/browse/NMS-9225

### DIFF
--- a/features/minion/heartbeat/consumer/src/main/java/org/opennms/minion/heartbeat/consumer/HeartbeatConsumer.java
+++ b/features/minion/heartbeat/consumer/src/main/java/org/opennms/minion/heartbeat/consumer/HeartbeatConsumer.java
@@ -28,7 +28,10 @@
 
 package org.opennms.minion.heartbeat.consumer;
 
-import com.google.common.collect.Sets;
+import java.util.Collections;
+import java.util.Date;
+import java.util.Objects;
+import java.util.Set;
 
 import org.opennms.core.ipc.sink.api.MessageConsumer;
 import org.opennms.core.ipc.sink.api.MessageConsumerManager;
@@ -57,10 +60,7 @@ import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.dao.DataAccessResourceFailureException;
 import org.springframework.transaction.annotation.Transactional;
 
-import java.util.Collections;
-import java.util.Date;
-import java.util.Objects;
-import java.util.Set;
+import com.google.common.collect.Sets;
 
 public class HeartbeatConsumer implements MessageConsumer<MinionIdentityDTO, MinionIdentityDTO>, InitializingBean {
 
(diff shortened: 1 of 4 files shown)


j-white:
Nice catch.

features/minion/heartbeat/consumer/src/main/java/org/opennms/minion/heartbeat/consumer/HeartbeatConsumer.java
```diff
@@ -154,7 +154,7 @@ private void provision(final OnmsMinion minion,
             return;
         }
 
-        // Return fast until the provisioner is running to pick up the events send below
+        // Return fast until the provisioner is running to pick up the events sent below
         if (!this.eventSubscriptionService.hasEventListener(EventConstants.RELOAD_IMPORT_UEI)) {
```
j-white: Should we also attempt to catch the *IllegalStateException* here, and prevent this known case from appearing in the logs?

soleger: I think we should really change the ```component-dao.xml``` into a context that is loaded later (like a ```component-service.xml```) since it relies on the EventIpcManager. BTW, I grepped and I think that this is the only ```component-dao.xml``` where we're injecting/autowiring the EventIpcManager interfaces so I don't there are similar problems in any other code.

j-white: Can we make that change then?

smoke-test/src/test/java/org/opennms/smoketest/minion/MinionHeartbeatOutageTest.java
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    /**
+     * Install the Kafka features on Minion.
+     * 
+     * @param minionSshAddr
+     * @param kafkaAddress
+     * @throws Exception
+     */
+    protected static void installFeaturesOnMinion(InetSocketAddress minionSshAddr) throws Exception {
```
j-white: Do we need this? We don't isntall anything any we already print the list of features installed after the Minion is spawned. Same for the OpenNMS instance.

soleger: This list of installed features/bundles is useful when debugging test failures so I'm going to leave it in.

features/minion/heartbeat/consumer/src/main/java/org/opennms/minion/heartbeat/consumer/HeartbeatConsumer.java
soleger: I'll address the startup ordering in a separate PR.

soleger:
I'm going to turn continuous builds on for this in Bamboo and if the new test doesn't flap, I'll merge this before Monday.

PR #1447, opened 2017-04-18
NMS-9143: Persist Drools working memory to disk when stopping and starting OpenNMS

agalue:
Persist Drools working memory to disk when stopping and starting OpenNMS

* JIRA: http://issues.opennms.org/browse/NMS-9143

### DIFF
--- a/opennms-correlation/drools-correlation-engine/src/main/java/org/opennms/netmgt/correlation/drools/Cause.java
+++ b/opennms-correlation/drools-correlation-engine/src/main/java/org/opennms/netmgt/correlation/drools/Cause.java
@@ -28,6 +28,7 @@
 
 package org.opennms.netmgt.correlation.drools;
 
+import java.io.Serializable;
 import java.util.HashSet;
 import java.util.Set;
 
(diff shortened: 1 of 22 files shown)


opennms-correlation/drools-correlation-engine/src/main/java/org/opennms/netmgt/correlation/drools/Cause.java
```diff
@@ -40,7 +41,8 @@
  * @author <a href="mailto:brozow@opennms.org">Mathew Brozowski</a>
  * @version $Id: $
  */
-public class Cause {
+@SuppressWarnings("serial")
```
j-white: Can add the serialVersionUID instead of ignoring the warnings? :)

agalue:
> On Apr 18, 2017, at 5:17 PM, Jesse White <notifications@github.com> wrote:
> Can add the serialVersionUID instead of ignoring the warnings? :)
> 
Done!

One interesting thing is that, I'm not the only one. I executed the following from the root of my branch after fixing what you’ve suggested, and …

$ git grep 'SuppressWarnings[(]"serial"[)]' . | wc -l
     101

j-white:
Thanks for the update. Can we also include the new `persistState` flag in the associated XSD?

agalue:
> On Apr 19, 2017, at 3:43 PM, Jesse White <notifications@github.com> wrote:
> 
> Thanks for the update.

My pleasure!

> Can we also include the new persistState flag in the associated XSD?

Done! Certainly I forgot about that, thanks!

PR #1451, opened 2017-04-24
NMS-9113, NMS-9116: Upgrade to Camel 2.14.4 to fix Netty4 support, RPC fixes

soleger:
The syslog receiver code still uses Netty 3 and cannot switch to Netty 4 without this minor Camel upgrade from 2.14.1 to 2.14.4.

This branch also contains additional fixes for the RPC code:
- Avoid starting the context manually CamelRpcServerRouteManager.
- Added tests to make sure that overriding ```CamelLogDebugBodyMaxChars``` is working.
- Fixed ConcurrentModificationException in MockLogAppender.

This code has been running in Bamboo continuously without failure.

* JIRA: http://issues.opennms.org/browse/NMS-9113
* JIRA: http://issues.opennms.org/browse/NMS-9116

### DIFF
--- a/core/ipc/rpc/camel-impl/src/main/java/org/opennms/core/rpc/camel/CamelRpcServerRouteManager.java
+++ b/core/ipc/rpc/camel-impl/src/main/java/org/opennms/core/rpc/camel/CamelRpcServerRouteManager.java
@@ -97,14 +97,14 @@ public class CamelRpcServerRouteManager {
         if (module != null) {
             final RpcModule<RpcRequest,RpcResponse> rpcModule = (RpcModule<RpcRequest,RpcResponse>)module;
             if (routeIdsByModule.containsKey(rpcModule)) {
-                LOG.warn("RpcModule {} was already registered.", rpcModule.getId());
-                return;
-            }
-            final DynamicRpcRouteBuilder routeBuilder = new DynamicRpcRouteBuilder(context, identity, rpcModule);
-            context.addRoutes(routeBuilder);
-            routeIdsByModule.put(rpcModule, routeBuilder.getRouteId());
+                LOG.warn("RpcModule {} ({}) was already registered", rpcModule.getId(), rpcModule.hashCode());
+            } else {
+                final DynamicRpcRouteBuilder routeBuilder = new DynamicRpcRouteBuilder(context, identity, rpcModule);
+                context.addRoutes(routeBuilder);
+                routeIdsByModule.put(rpcModule, routeBuilder.getRouteId());
 
-            LOG.info("Registered RpcModule {} on queue {}", rpcModule.getId(), routeBuilder.getQueueName());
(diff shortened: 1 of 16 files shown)


core/ipc/rpc/camel-impl/src/main/java/org/opennms/core/rpc/camel/CamelRpcServerRouteManager.java
```diff
@@ -77,7 +76,11 @@ private DynamicRpcRouteBuilder(CamelContext context, MinionIdentity identity, Rp
         }
 
         public String getRouteId() {
-            return "RPC.Server." + module.getId();
+            return "RPC.Server." + module.getId() + "-" + Math.abs(module.hashCode());
```
j-white: Do we need the hashCode here if we only create one route per module?

core/ipc/rpc/camel-impl/src/main/java/org/opennms/core/rpc/camel/CamelRpcServerRouteManager.java
```diff
@@ -94,22 +97,29 @@ public void bind(RpcModule module) throws Exception {
...
-                LOG.warn("RPC module {} was already registered.", rpcModule.getId());
-                return;
+                LOG.warn("RpcModule {} ({}) was already registered", rpcModule.getId(), rpcModule.hashCode());
+            } else {
+                final DynamicRpcRouteBuilder routeBuilder = new DynamicRpcRouteBuilder(context, identity, rpcModule);
+                context.addRoutes(routeBuilder);
+                routeIdsByModule.put(rpcModule, routeBuilder.getRouteId());
+
+                LOG.info("Registered RpcModule {} ({}) on queue {}", rpcModule.getId(), rpcModule.hashCode(), routeBuilder.getQueueName());
+                new Exception().printStackTrace();
```
j-white: I'm guessing you forgot to remove this.

core/ipc/rpc/camel-impl/src/main/java/org/opennms/core/rpc/camel/CamelRpcServerRouteManager.java
soleger: I put this in because the DNS RPC service was being added twice in some of the tests causing tests to flap (depending on whether or not the second service initialized before the Camel context started). By making the routes unique, you can see that two modules are trying to register for the same RPC service and trace that back to the log messages that I added to bind/unbind.

j-white: Won't this make it difficult to collect route specific statistics though? Since the MBeans are named using the route id.

j-white: For that reason, I feel that we should revert the change on this line. Loading the same modules multiple times is not a problem during the normal run-time, and if it does reoccur in the tests, we can still isolate it using the log message you added.

PR #1452, opened 2017-04-25
NMS-9265: Don't include iplike query when it is an actual result

pdzaccone:
When forming iplike queries in Topology, duplicate entries occur when typing an actual result - fixed


mvrueden:
Looks good so far. I edited the description and added the link to the jira issue.

j-white:
Shouldn't we use a `LinkedHashSet` instead of a `HashSet` so that we don't lose the ordering?

mvrueden:
@j-white, yes that is probably true (-:

mvrueden:
There should also be a test covering the issue described so we do not run into it again in the future.

mvrueden:
We decided to not change the API in general and have the `SearchProvider`s not add duplicate entries instead. 

I pulled the changes in to `jira/NMS-9265` to see what bamboo thinks about it.

PR #1457, opened 2017-04-26
Get rid of some deprecated Spring JdbcTemplate calls

soleger:


roskens:
Why does testGetEventHostWithOneMatch() have nodeId as a long, while most of the other tests have it as an int?

indigo423:
Do we need to track those changes in JIRA for released versions for Meridian/Horizon and possible regression?

soleger:
I don't think we need to track in JIRA since this is routine refactoring.

PR #1461, opened 2017-05-02
NMS-7813: Expose statistics of provisiond via jmx

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-7813

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/service-configuration.xml
+++ b/opennms-base-assembly/src/main/filtered/etc/service-configuration.xml
@@ -172,13 +172,13 @@ maintained
   </service>
   <service>
     <name>OpenNMS:Name=Provisiond</name>
-    <class-name>org.opennms.netmgt.daemon.SimpleSpringContextJmxServiceDaemon</class-name>
+    <class-name>org.opennms.netmgt.provision.service.jmx.Provisiond</class-name>
     <attribute>
-      <name>LoggingPrefix</name>
+      <name>loggingPrefix</name>
       <value type="java.lang.String">provisiond</value>
     </attribute>
     <attribute>
-      <name>SpringContext</name>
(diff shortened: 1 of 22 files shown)


opennms-base-assembly/src/main/filtered/etc/service-configuration.xml
```diff
@@ -172,15 +172,7 @@ maintained
...
-    <class-name>org.opennms.netmgt.daemon.SimpleSpringContextJmxServiceDaemon</class-name>
-    <attribute>
-      <name>LoggingPrefix</name>
-      <value type="java.lang.String">provisiond</value>
-    </attribute>
-    <attribute>
-      <name>SpringContext</name>
-      <value type="java.lang.String">provisiondContext</value>
-    </attribute>
+    <class-name>org.opennms.netmgt.provision.service.jmx.Provisiond</class-name>
```
j-white: We want to avoid changing configuration files in `foundation-2016`, so if this change is required, we may want to consider re-targeting this towards `foundation-2017` instead.

jeffgdotorg: @j-white these stats are valuable enough, IMO, to *consider* a rare exception to the intra-*foundation* .rpmnew policy. Can we solve it with an upgrader task? Is this change even a hard requirement? (I'm admittedly foggy on the reasons for the different wiring styles, so apologies if that latter question is objectively dumb)

mvrueden: It should work without the configuration change, but spring did not initialize the beans then and I was not able to figure out why. So yeah, it is required (-:

mvrueden:
Closing for now, as it is not doing what it is supposed to do.

PR #1462, opened 2017-05-02
NMS-9291: Logging for TCP output queue

tdefilip:
- Add logging of queue checks/offers/drains and queue size/capacity statistics.

JIRA: http://issues.opennms.org/browse/NMS-9291

Todo Review:
- [ ] Typo and grammar
- [ ] Formatting and conventions
- [ ] Content

### DIFF
--- a/features/collection/persistence-tcp/pom.xml
+++ b/features/collection/persistence-tcp/pom.xml
@@ -27,6 +27,12 @@
     </plugins>
   </build>
   <dependencies>
+    <!-- Dependencies used -->
+    <dependency>
+      <groupId>io.dropwizard.metrics</groupId>
+      <artifactId>metrics-core</artifactId>
+      <version>${dropwizardMetricsVersion}</version>
+    </dependency>
     <dependency>
       <groupId>org.opennms.features.collection</groupId>
       <artifactId>org.opennms.features.collection.api</artifactId>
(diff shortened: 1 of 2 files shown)


j-white:
This looks like a good use case for using a `MetricRegistry` along with a `Slf4jReporter`:
    http://metrics.dropwizard.io/3.1.0/manual/core/#slf4j

tdefilip:
If I'm understand how to use the metric registry, I still need to maintain the various counter members.  But instead of having a bunch of get methods, I can just register all such values with the registry (which I would pass in to various constructors) and then dump the registry using Slf4jReporter?

j-white:
Yeah, that's the idea. This makes it easier to manage and gives more detailed statistics in most cases.

tdefilip:
I've added use of the MetricRegistry.  I have not included the Slf4jReporter since it looks like resetting the counters would be much more difficult that way; additionally, the queue size and remaining capacity are not part of the registry, so it looks like it would be difficult to get the reporter to include them.

Let me know if they are a must-have.

j-white:
We don't really gain much by using the MetricRegistry this way.

I'll go ahead and merge the patch before we made the MetricRegistry changes.

j-white:
I've pushed [a377d](https://github.com/OpenNMS/opennms/pull/1462/commits/a377d807c08c49742dcb2e4da0a20d45ae3d6136) to `pr/jira/NMS-9291`. Waiting on Bamboo.

j-white:
Manually merged into `foundation-2016`.

tdefilip:
Thanks!

PR #1464, opened 2017-05-02
NMS-9294: Add initialSleepTime to IPC sink code

soleger:
This PR adds a ```org.opennms.core.ipc.sink.initialSleepTime``` system property that can be used to delay sink consumers from registering for messages to give the rest of the event system (ie. Elasticsearch forwarding) time to start up.

* JIRA: http://issues.opennms.org/browse/NMS-9294

### DIFF
--- a/smoke-test/src/test/resources/org.apache.karaf.features.cfg
+++ b/smoke-test/src/test/resources/org.apache.karaf.features.cfg
@@ -1,7 +1,7 @@
 #
 # Comma separated list of features repositories to register by default
 #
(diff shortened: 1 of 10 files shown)


smoke-test/src/test/resources/org.apache.karaf.features.cfg
```diff
@@ -0,0 +1,52 @@
+#
+# Comma separated list of features repositories to register by default
+#
+featuresRepositories=mvn:org.opennms.container/org.opennms.container.karaf/19.0.0-SNAPSHOT/xml/features,mvn:org.opennms.karaf/opennms/19.0.0-SNAPSHOT/xml/features
```
j-white: I suspect that the fixed version string here will cause problems.

PR #1466, opened 2017-05-04
NMS-9127: bail if OpenNMS is currently running

RangerRick:
It turns out it's a TOTAL PAIN IN THE ASS to handle this properly because of the way RPM and Debian transactions happen.  Depending on what version you're starting from, and how upgrades go, OpenNMS can be already stopped during the process of upgrading (Debian packages already did this).  So I end up having to track, manually, whether an "upgrade" is already in-process.

Here's what I ended up doing:

1. wrote a script called `upgrade-precheck.sh` that does the following:
    * check for `${OPENNMS_HOME}/etc/upgradewhilerunning` -- if the user touches that file, then the upgrade will continue on no matter what; otherwise...
    * read a lockfile created by previous runs of `upgrade-precheck.sh` which contains a datestamp and version number
    * check if OpenNMS is running, if it is, update the lockfile and fail; otherwise...
    * if the lockfile's version number matches the current run of `upgrade-precheck.sh`, then check the timestamp; if the timestamp is within 60 seconds of the start of the current run of `upgrade-precheck.sh`, then it assumes that a previous run of `upgrade-precheck.sh` from another package in the transaction has already failed, so it updates the lockfile and fails too
2. create a new package, `opennms-upgrade`, which is depended upon by all other OpenNMS packages, which contains this `upgrade-precheck.sh` script, so it will always be installed first
3. if OpenNMS is *not* running, then the `opennms-core` package (RPM) or `opennms-db` package (DEB) will rsync the current contents of `${OPENNMS_HOME}/etc/` to `${OPENNMS_HOME}/share/etc-pre-upgrade/` so you will always have a backup of your pre-upgrade configs. (Not part of the issue, but seemed like it'd be useful while I was at it.)

In short, if OpenNMS *was* running at the beginning of the upgrade, this lockfile will be created and make sure all the packages fail consistently (otherwise there will be cases in which the RPMs or Debian packages will be in a half-installed state).  If OpenNMS was *not* running at the beginning of the upgrade, back up `etc` and continue on.

* JIRA: http://issues.opennms.org/browse/NMS-9127


RangerRick:
Marking as too scary.  Merge the alternate PR instead: https://github.com/OpenNMS/opennms/pull/1494

PR #1474, opened 2017-05-08
HZN-1054: Some (other) events have Log Messages that are more verbose than Description

mfuhrmann:
I found another events with that problem. I hope I got them all...

https://issues.opennms.org/browse/HZN-1054


opennms-base-assembly/src/main/filtered/etc/events/opennms.config.events.xml
```diff
@@ -5,7 +5,7 @@
         <uei>uei.opennms.org/internal/reloadScriptConfig</uei>
         <event-label>OpenNMS-defined internal event: reloadScriptConfig</event-label>
         <descr>
-            &lt;p&gt;The administrator has changed the ScriptD
+            &lt;p&gt;The administrator has changed the ScriptDuei.opennms.org/internal/reloadDaemonConfigFaileduei.opennms.org/internal/reloadDaemonConfigFaile
```
j-white: Hmmmm, this doesn't look right.

PR #1476, opened 2017-05-09
NMS-5646: Replaced weird antlr.StringUtils call with regex

soleger:
Includes a test to verify that the behavior of the new function matches the antlr.StringUtils behavior.

* JIRA: http://issues.opennms.org/browse/NMS-5646

### DIFF
--- a/features/poller/monitors/core/src/main/java/org/opennms/netmgt/poller/monitors/HostResourceSwRunMonitor.java
+++ b/features/poller/monitors/core/src/main/java/org/opennms/netmgt/poller/monitors/HostResourceSwRunMonitor.java
@@ -35,6 +35,7 @@ import java.util.Map;
 
 import org.opennms.core.utils.InetAddressUtils;
 import org.opennms.core.utils.ParameterMap;
+import org.opennms.core.utils.StringUtils;
 import org.opennms.core.utils.TimeoutTracker;
 import org.opennms.netmgt.poller.Distributable;
 import org.opennms.netmgt.poller.DistributionContext;
@@ -51,8 +52,6 @@ import org.opennms.netmgt.snmp.SnmpWalker;
 import org.opennms.netmgt.snmp.TableTracker;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import antlr.StringUtils;
 
 /**
  * <P>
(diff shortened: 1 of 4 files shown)


features/poller/monitors/core/src/main/java/org/opennms/netmgt/poller/monitors/HostResourceSwRunMonitor.java
```diff
@@ -198,7 +197,7 @@ public void rowCompleted(SnmpRowResult result) {
                 final SnmpValue name = nameResults.get(nameInstance);
                 final SnmpValue value = statusResults.get(nameInstance);
                 // See if the service name is in the list of running services
-                if (name != null && value != null && match(serviceName, stripExtraQuotes(name.toString()))) {
+                if (name != null && value != null && match(serviceName, StringUtils.stripExtraQuotes(name.toString()))) {
```
j-white: `stripExtraQuotes` and `StringUtils.stripExtraQuotes` aren't equivalent. This refactor changes the behavior.

This source appears to match what the `stripFrontBack` function actually does:
    https://github.com/xtuml/pt_antlr/blob/master/pt_antlr/antlr/StringUtils.java#L79

soleger: The only difference is that antlr.StringUtils.stripFrontBack() throws a StringIndexOutOfBoundsException if you pass it ```"```, and we were only handling this in one of the methods.

j-white: The result of both functions against  `String str = "a\"b\"\"c"` don't match.

soleger: I guess this regex correctly fixes NMS-3882 then, since strings with internal quotes would have been hacked up by the antlr.StringUtils call. In any case, the regex is the behavior that is intended.

j-white: Cool, just making sure that was intentional.

PR #1477, opened 2017-05-09
NMS-7904: Allow root arrays in json datacollection

fooker:
If a json datacollection returns an arrays as to level element, it is
warpped in an object exposing the array using an 'elements' property.
This allows to query such data using the existing xpath queries.

* JIRA: http://issues.opennms.org/browse/NMS-7904


roskens:
I didn't see any documentation for this. How would a user know to use an /elements tag for this?

fooker:
I've added documentation and therefore rebased to development.

PR #1478, opened 2017-05-09
NMS-9089: grafanaBox needs a count limiter/pager

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-9089


opennms-webapp/src/main/webapp/includes/grafana-box.jsp
```diff
@@ -128,9 +136,14 @@
                     }
                 }
                 if (showDashboard) {
-                    $('#dashboardlist').append('<a href="<%=grafanaProtocol%>://<%=grafanaHostname%>:<%=grafanaPort%>/dashboard/' + val['uri'] + '"><span class="glyphicon glyphicon-signal" aria-hidden="true"></span>&nbsp;' + val['title'] + "</a><br/>");
+                    if (limit < 1 || count++ < limit) {
+                        $('#dashboardlist').append('<a href="<%=grafanaProtocol%>://<%=grafanaHostname%>:<%=grafanaPort%>/dashboard/' + val['uri'] + '"><span class="glyphicon glyphicon-signal" aria-hidden="true"></span>&nbsp;' + val['title'] + "</a><br/>");
+                    }
                 }
             };
+            if (limit > 0 && count > limit) {
```
roskens: Did I leave a logic bug in here?  Should this be `limit > 0 && count >= limit` ?

christianpape: No, it's correct since count is increased in the preceding if-statement. So, count will match limit in the case limit equals the number of available dashboards. In this case we do not need the "View list of all Dashboards" link.

PR #1479, opened 2017-05-09
NMS-9330: Update a lot of minor dependencies

soleger:
There's nothing really risky here except maybe upgrading ecj since it's used by Drools. This fixes two deps that were dep-managed downward (ehcache and cglib-nodep).

* JIRA: http://issues.opennms.org/browse/NMS-9330

### DIFF
--- a/container/features/src/main/resources/features.xml
+++ b/container/features/src/main/resources/features.xml
@@ -96,26 +96,26 @@
       <bundle>wrap:mvn:com.atomikos/transactions-jdbc/${atomikosVersion}</bundle>
     </feature>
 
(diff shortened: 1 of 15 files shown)


soleger:
I ended up reverting the ecj upgrade anyway since it was causing some minor problems on a Spring MVC page. :)

j-white:
Are we updating these to address any specific issues?

soleger:
The ehcache, cglib, and ognl changes fix mismatches between JARs we're packaging and what code has been compiled against. Several are projects that have moved to new artifacts and the rest is just housekeeping.

soleger:
Closing until I have time to review the problems in this branch.

PR #1480, opened 2017-05-09
NMS-9325: Calculation of KSC reports timespans "Last Week" and "This Week" broken

agalue:
Calculation of KSC reports timespans "Last Week" and "This Week" broken due to Locale.

* JIRA: http://issues.opennms.org/browse/NMS-9325

### DIFF
--- a/opennms-config/src/main/java/org/opennms/netmgt/config/KSC_PerformanceReportFactory.java
+++ b/opennms-config/src/main/java/org/opennms/netmgt/config/KSC_PerformanceReportFactory.java
@@ -397,7 +397,7 @@ public class KSC_PerformanceReportFactory {
                 end_time.add(Calendar.DATE, -1);
                 end_time.set(Calendar.HOUR_OF_DAY, 22);
             } else if (interval.equals("This Week") || interval.equals("Last Week")) {
-                begin_time.set(Calendar.DAY_OF_WEEK, 1);
+                begin_time.set(Calendar.DAY_OF_WEEK, begin_time.getFirstDayOfWeek());
                 end_time.set(Calendar.DAY_OF_WEEK, 7);
                 end_time.set(Calendar.HOUR_OF_DAY, 23);
                 end_time.set(Calendar.MINUTE, 59);
--- a/opennms-config/src/main/java/org/opennms/netmgt/config/KSC_PerformanceReportFactory.java
+++ b/opennms-config/src/main/java/org/opennms/netmgt/config/KSC_PerformanceReportFactory.java
@@ -398,7 +398,8 @@ public class KSC_PerformanceReportFactory {
                 end_time.set(Calendar.HOUR_OF_DAY, 22);
             } else if (interval.equals("This Week") || interval.equals("Last Week")) {
                 begin_time.set(Calendar.DAY_OF_WEEK, begin_time.getFirstDayOfWeek());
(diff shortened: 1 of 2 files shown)


opennms-config/src/main/java/org/opennms/netmgt/config/KSC_PerformanceReportFactory.java
```diff
@@ -397,7 +397,7 @@ public static synchronized void getBeginEndTime(String interval, Calendar begin_
                 end_time.add(Calendar.DATE, -1);
                 end_time.set(Calendar.HOUR_OF_DAY, 22);
             } else if (interval.equals("This Week") || interval.equals("Last Week")) {
-                begin_time.set(Calendar.DAY_OF_WEEK, 1);
+                begin_time.set(Calendar.DAY_OF_WEEK, begin_time.getFirstDayOfWeek());
                 end_time.set(Calendar.DAY_OF_WEEK, 7);
```
soleger: I just noticed that this needs to change as well. You need to calculate the end of the week day based on the beginning of the week day, otherwise you'll only have 6 days in your week.

agalue:
> On May 10, 2017, at 11:24 AM, Seth Leger <notifications@github.com> wrote:
> 
> I just noticed that this needs to change as well. You need to calculate the end of the week day based on the beginning of the week day, otherwise you'll only have 6 days in your week.

Good point!

You mean something like this ?
end_time.set(Calendar.DAY_OF_WEEK, end_time.getFirstDayOfWeek() + 6);

soleger:
Unfortunately, that is not good enough because if the DAY_OF_WEEK is MONDAY, that is integer 2 so you'd get DAY_OF_WEEK == 8 which (I assume?) is invalid.

I think you'll have to write a test for this, and maybe do some different Calendar math. :)

agalue:
> On May 10, 2017, at 11:35 AM, Seth Leger <notifications@github.com> wrote:
> 
> I think you'll have to write a test for this, and maybe do some different Calendar math. :)

I would say it depends on what’s the meaning of "this" or "last" week.

I have another change in mind that does the following (tested on Meridian 2016 using the KSC View UI).

For LANG=de_DE.utf8, this week goes from Mon May 01 00:00:00 UTC 2017, to Sun May 07 23:59:00 UTC 2017
For LANG=en_US.utf8, this week goes from Sun Apr 30 00:00:00 UTC 2017, to Sat May 06 23:59:00 UTC 2017

The range is correctly 7 days :)

If I choose "Last Week" the currently logic works (i.e. the same dates shifted 7 days to the left).

Sounds good?

soleger:
If you're saying that ```end_time.set(Calendar.DAY_OF_WEEK, end_time.getFirstDayOfWeek() + 6);``` works as expected if the value overflows ```7```, then that's fine. I'd definitely put a comment in the code so that people looking at it aren't confused.

agalue:
> On May 10, 2017, at 12:02 PM, Seth Leger <notifications@github.com> wrote:
> 
> If you're saying that end_time.set(Calendar.DAY_OF_WEEK, end_time.getFirstDayOfWeek() + 6); works as expected if the value overflows 7, then that's fine


No. I’m pretty sure I said, I have ANOTHER change in mind; which BTW I’ve already committed it.

soleger:
OK looks good! :)

PR #1481, opened 2017-05-09
NMS-9333: Fix 'mvn install clean' builds

soleger:
This avoids cleaning the ```remote-poller``` project to avoid problems in the ```remote-poller-jnlp``` build when using ```mvn install clean```.

* JIRA: http://issues.opennms.org/browse/NMS-9333

### DIFF
--- a/features/remote-poller/pom.xml
+++ b/features/remote-poller/pom.xml
@@ -124,6 +124,16 @@ groovy.compiler.output.path=target/classes
           </execution>
         </executions>
       </plugin>
+      <!--
+        NMS-9333: Skip the clean phase so that the artifact can be resolved by the 
+        plugins in the remote-poller-jnlp project
+      -->
+      <plugin> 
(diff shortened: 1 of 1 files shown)


mvrueden:
I read `fix mvr builds` and slightly pannicked (-:

I rather not skip the clean phase, as this is a crucial workflow of maven.
Can we not have it deploy the artifacts to the maven repository instead and use them in the jnlp project?

soleger:
The artifact does deploy to ~/.m2/repository however, the mojohaus/webstart maven plugin uses an odd API call to resolve bundles that doesn't fall back to using the .m2/repository so it fails if both projects (```remote-poller``` and ```remote-poller-jnlp```) are invoked in the same build.

I think in this case, it's OK to avoid the clean phase for this one artifact.

indigo423:
The reason breaking a maven default behavior is just to save some disk space?

soleger:
Closing until I can revisit this.

PR #1483, opened 2017-05-11
NMS-9328: Add status charts to indicate overall status

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-9328


j-white:
There's a lot to digest here, but here are my initial thoughts/findings:
* If you don't have any business services / applications, but have nodes, the two doughnuts aren't centered properly
* Clicking on "Alarms" and getting a list of nodes was confusing at first, I realize that this is the "nodes grouped by alarm status", but this was not immediately evident
* Nit: After clicking on "Alarms", I need to press back twice on my browser to get to the home screen
* It looks like there's a lot of code duplication on the different status pages. (If this were Java code, I'm sure you'd treat it differently :) )
* Should we also provide alternate links to the status pages? Say under the "Status" drop-down.

mvrueden:
> If you don't have any business services / applications, but have nodes, the two doughnuts aren't centered properly

I can spent some more time to try to have it center properly, but so far I was not able to achieve that. When I add two normal div's instead of the charts, they are centered accordingly. So something in c3 is interfering.

>  Clicking on "Alarms" and getting a list of nodes was confusing at first, I realize that this is the "nodes grouped by alarm status", but this was not immediately evident

Yes, but there is not enough space to label it "Node status calculated based on alarms". 
We can add a tooltip indicating this.

> Nit: After clicking on "Alarms", I need to press back twice on my browser to get to the home screen

Hmm, I will have a look at it.

> It looks like there's a lot of code duplication on the different status pages. (If this were Java code, I'm sure you'd treat it differently :) )

Yes, but I wanted to get a discussion started before having everything layed out perfectly.

> Should we also provide alternate links to the status pages? Say under the "Status" drop-down.

I have no strong opinion about this.

mvrueden:
Besides rebasing it against foundation-2017 to resolve merge conflicts, I was able to center the charts a bit more, but when only 2 charts are visible, they are centered by row, but not 100% by column. If this is not "centered enough" I need someone to give me a hand. I am no css guru.

In addition I added a tooltip for each graph, to indicate what it shows, e.g. 'Nodes grouped by acknowleded alarms' or ' Nodes grouped by current Outages'. 

If links in the status drop down are desired, I can add them. Please let me know.

If the feature is accepted in general, I'd like to redo the ui code-side, which so far I waited to do.
While doing that the code-duplication and the "go back twice issue" will be addressed.

Thank you.

mfuhrmann:
> If the feature is accepted in general,
Really? Is there a chance that it could be rejected? ;-)

> If links in the status drop down are desired, I can add them. Please let me know.
Do you want a single page with the donut graphs on it? I usually don't need these single pages. I'm just using the starting page for work. But it doesn't hurt...
There is a other problem. Pages like this are neat to have for presentation monitors. Like the geo map (if you have a lot of locations). But you always run in some problems, I've documented here once; https://issues.opennms.org/browse/NMS-8750
But this is an other construction site..

It would be very cool, if you could provide a live version with this feature in it. So others could easily test the usability. Similar to the demo.opennms.org. dev.opennms.org or something similar.

mvrueden:
I refactored the client-side code to re-use common components and also introduced general components, such as pagination or filter-controls. Those are at the moment bound to the `onms.status` module, but can be easily extracted if needed.

Having a automated way to run/test certain branches would be nice, but is out of scope in this issue. There is a way to build a specific branch using docker. Ask @indigo423 for more details if you want to try it out.

No strong statement was made about the adding links to the status drop down. I will leave them for now. If they are required in the future, it is easy to add them.

The `go back` issue should be resolved now.

@indigo423 mentioned an issue on the `business-service status page` with pagination. Can you verify if this issue still exists?

Thank you.

mvrueden:
@indigo423 reported some problems. Will investigate.

mvrueden:
With the current architecture of opennms this is not doable in a performant way. We either need some kind of "statusd" which already knows the status and one can do queries based on severities or some other mechanisms need to be established, e.g. views.

As this feature was intended to be a small enhancement and already ate a lot of time, I am burrying it now.

Closing PR.

mvrueden:
@christianpape and I spent some more time on this and were able to address the performance issues. We solved the problems by introducing two status views: 
 * node_alarm_status (status calculated based on alarms)
 * node_outage_status (status calculated based on outages)

christianpape:
Tested it and seems to work pretty good. 
+1

PR #1484, opened 2017-05-11
NMS-9342: Make geomap a requirejs module to avoid requirejs conflicts

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-9342


j-white:
I can confirm that this fixes the requirejs error that used to show up in the console.

agneal:
I'm still having this issue in Debian even after applying these changes. What might be the best course of action? I'm currently trying to fix it on my own- I think it might be a difference between rhel and debian file structure. but I'm uncertain. Console brings up a 404 not found error for geomap.js, which leads to a require.js error.

Apologies in advance if it's bad etiquette to respond to an old bug like this. I'm just not sure where else to post.

PR #1489, opened 2017-05-16
NMS-9347: Replace Lock with ReentrantLock

soleger:
Replaces a homegrown Lock class with java.util.concurrent.locks.ReentrantLock.

* JIRA: http://issues.opennms.org/browse/NMS-9347


j-white:
+1 to built-ins vs homegrown code.

PR #1491, opened 2017-05-17
HZN-705, HZN-1058: Upgrade to Camel 2.16, ActiveMQ 5.13

soleger:
Now that we're using Spring 4.1, this PR upgrades Camel and ActiveMQ to versions that roughly align with each other.

* JIRA: http://issues.opennms.org/browse/HZN-705
* JIRA: http://issues.opennms.org/browse/HZN-1058

### DIFF
--- a/core/daemon/src/main/resources/META-INF/opennms/applicationContext-daemon.xml
+++ b/core/daemon/src/main/resources/META-INF/opennms/applicationContext-daemon.xml
@@ -29,6 +29,17 @@
   <onmsgi:reference id="eventProxy" interface="org.opennms.netmgt.events.api.EventProxy"/>
   <onmsgi:reference id="eventSubscriptionService" interface="org.opennms.netmgt.events.api.EventSubscriptionService"/>
 
+  <!-- TODO: Remove this bean after we stop using serialized object messages -->
+  <bean id="enableActiveMqObjectMessages" class="org.springframework.beans.factory.config.MethodInvokingFactoryBean">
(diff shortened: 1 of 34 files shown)


core/daemon/src/main/resources/META-INF/opennms/applicationContext-daemon.xml
```diff
@@ -29,13 +29,26 @@
   <onmsgi:reference id="eventProxy" interface="org.opennms.netmgt.events.api.EventProxy"/>
   <onmsgi:reference id="eventSubscriptionService" interface="org.opennms.netmgt.events.api.EventSubscriptionService"/>
 
+  <!-- TODO: Remove this bean after we stop using serialized object messages -->
```
j-white: I thought we got rid of the remaining serialized object messages. Where do we still use them?

soleger: I don't think that we do now but the original commits in this branch pre-dated that work. I'll track removing the ```org.apache.activemq.SERIALIZABLE_PACKAGES``` system properties in a new issue.

j-white: Was this tracked anywhere?

soleger: https://issues.opennms.org/browse/HZN-1071

PR #1493, opened 2017-05-17
HZN-1055: Added scvcli utility for setting, listing and retrieving cr…

christianpape:
* JIRA: https://issues.opennms.org/browse/HZN-1055

### DIFF
--- a/features/scv/scvcli/src/main/java/org/opennms/features/scv/cli/ScvCli.java
+++ b/features/scv/scvcli/src/main/java/org/opennms/features/scv/cli/ScvCli.java
@@ -1,0 +1,126 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2017 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2017 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License,
+ * or (at your option) any later version.
+ *
+ * OpenNMS(R) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with OpenNMS(R).  If not, see:
+ *      http://www.gnu.org/licenses/
+ *
+ * For more information contact:
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.features.scv.cli;
+
+import java.io.IOException;
+import java.util.Properties;
+import java.util.function.Function;
+
+import org.kohsuke.args4j.Argument;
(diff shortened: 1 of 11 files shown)


opennms-doc/guide-install/src/asciidoc/text/minion/configure-minion.adoc
```diff
@@ -30,6 +30,15 @@ admin@minion()> scv:set opennms.http minion minion
 admin@minion()> scv:set opennms.broker minion minion
 ----
 
+Another way to configure credentials is to use the `scvcli` utility in your _Minion_ `bin` directory.
+
+.Configure the credentials by using the command line utility `scvcli`
+[source]
+----
+[root@localhost /usr/share/minion]# $ ./bin/scvcli --keystore etc/scv.jce set opennms.http minion minion
```
j-white: Can we make it so these work without specifying the keystore path? i.e. setting the default relative to the binary or $MINION_HOME

features/scv/scvcli/src/main/java/org/opennms/features/scv/cli/ScvCli.java
```diff
@@ -0,0 +1,126 @@
...
+
+    private final static String DEFAULT_PASSWORD_PROPERTY = "org.opennms.features.scv.cli.password";
+
+    @Argument(required = true,
+            index = 0,
+            metaVar = "ACTION",
+            handler = SubCommandHandler.class)
+    @SubCommands({@SubCommand(name = "list", impl = ListCommand.class),
+            @SubCommand(name = "set", impl = SetCommand.class),
+            @SubCommand(name = "get", impl = GetCommand.class)})
```
j-white: I think we may want to omit the `get` command here, making it more difficult to "peek at" the credentials.

opennms-doc/guide-install/src/asciidoc/text/minion/configure-minion.adoc
```diff
@@ -30,6 +30,15 @@ admin@minion()> scv:set opennms.http minion minion
 admin@minion()> scv:set opennms.broker minion minion
 ----
 
+Another way to configure credentials is to use the `scvcli` utility in your _Minion_ `bin` directory.
+
+.Configure the credentials by using the command line utility `scvcli`
+[source]
+----
+[root@localhost /usr/share/minion]# $ ./bin/scvcli set opennms.http minion minion
```
j-white: In the .rpms the correct path is `/opt/minion` vs `/usr/share/minion` in the .debs.

Maybe we should add a note to avoid confusion?

christianpape: The same path/prompt is used in the other minion shell examples.

features/scv/scvcli/src/main/java/org/opennms/features/scv/cli/ScvCli.java
christianpape: The password will not be displayed anymore...

opennms-doc/guide-install/src/asciidoc/text/minion/configure-minion.adoc
j-white: I don't see `/usr/share/minion` being referenced anywhere else. What example are you referring to exactly?

The ones I do see connect via the Karaf shell, where the working directory doesn't matter.

christianpape: Ok. You're right. It was the Debian-related stuff. I point the path to /opt/minion, right?

j-white: The Debian packages don't create `/opt/minion`, so this won't work with the Debian packages now.

Maybe we should append something like:
```
NOTE: If you are using the Debian packages, use `/usr/share/minion` instead of `/opt/minion`
```

Sorry for the confusion :)

PR #1497, opened 2017-05-19
HZN-950: Add SHA1 checksums to the Karaf repositories

soleger:
To prevent a bunch of warning messages during Karaf startup.

* JIRA: http://issues.opennms.org/browse/HZN-950


j-white:
We'll need a similar update for the Debian packages that were added in `develop`.

I just saw your comment in HZN-950. Wouldn't adding the checksums in the Maven assembly be a better approach? That way all packages, including the tarball distribution (HZN-1052) would contain the checksums.

j-white:
Closing in favor of https://github.com/OpenNMS/opennms/pull/1517

PR #1501, opened 2017-05-23
NMS-9364: "Add Edge" in BSM configuration interface is slow

rssntn67:
NMS-9346: BSM Admin add Edge BSM service is slow

Check:
* JIRA: http://issues.opennms.org/browse/NMS-9346


rssntn67:
Nice , thanks

PR #1503, opened 2017-05-23
NMS-8093: Never create sessions in REST-APIs

fooker:
This avoids creation of sessions, and thus setting a session cockie,
when using the REST APIs using basic authentication. It still uses the
session for authentication if one already exists.

* JIRA: http://issues.opennms.org/browse/NMS-8093


fooker:
Should we backport this?

j-white:
Can we add also a simple test that verifies no "Set-Cookie" header is present?

As for backporting, I'd opt to leaving it targeted for develop to avoid inadvertently breaking any integrations that may rely on those sessions.

PR #1505, opened 2017-05-24
NMS-9370: Colors on trend graph differ from OpenNMS styling

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-9370


mvrueden:
I tried this out and it works.
I think it is more consistent with the rest of the UI as before.
When bamboo goes green, this can be merged from my point of view.

Please squash merge and add the issue number to the squashed commit.

indigo423:
I would suggest avoiding colors which are used to indicate status or severities when there is no status or a severity is indicated. This is the case when we show a green line which can lead to a false indication as "Normal".

The trend line is not able to represent a status nor severity and should not have a color which can lead to misinterpretation. This type of issues was reported from customers and NOC agents. We tried to fix them, e.g. red title description in the dashboard and we do same mistakes again.

PR #1506, opened 2017-05-25
NMS-9369: Update business service states properly when alarms are deleted

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9369

Here we introduce a new 'alarmDelete' lifecycle event and update the default automations responsible for deleting alarms to generate these. We also add support to bsmd for handling these, and making sure that deleted alarms are properly accounted for when performing the scheduled poll of the alarm table.

With these changes, if a business service references an alarm that is deleted by a vacuumd automation, it's state will be immediately updated. Similarly, if a business service references an alarm, and that alarm is deleted by some other means, the business service will be updated on the next poll.

### DIFF
--- a/features/bsm/daemon/pom.xml
+++ b/features/bsm/daemon/pom.xml
@@ -112,5 +112,15 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.jayway.awaitility</groupId>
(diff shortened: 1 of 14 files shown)


agalue:
Looks good to me. I haven't tested it, but a customer that heavily relies on BSMD will be able to do it very soon.

PR #1510, opened 2017-05-27
NMS-9373: Add a lookup parameter for DNSResolutionMonitor

roskens:
Add a lookup parameter to DNSResolutionMonitor. Its value replaces the node label as the DNS record to lookup.

JIRA: 
* http://issues.opennms.org/browse/NMS-9249
* http://issues.opennms.org/browse/NMS-9373
* http://issues.opennms.org/browse/NMS-9378
* http://issues.opennms.org/browse/NMS-9383


features/poller/monitors/core/src/test/java/org/opennms/netmgt/poller/monitors/DNSResolutionMonitorTest.java
```diff
@@ -51,10 +56,20 @@
  * @author brozow
  */
 public class DNSResolutionMonitorTest {
+
+    @Rule
+    public TestName m_test = new TestName();
     
     @Before
     public void setUp() {
+        System.out.println("------------ begin test " + m_test.getMethodName() + " ------------");
```
indigo423: Should we use a Logger here instead of System.out.println()?

roskens: I can change that.

roskens:
This commit ends up containing pieces for 4 different JIRA issues that referenced DNSResolutionMonitor.

indigo423:
> This commit ends up containing pieces for 4 different JIRA issues

You can call this PR bloating :) In general, I would recommend to just implement the things related to the issue. The things for itself are small changes, but combined them all together makes it more complicated. The main problem is:

1.) changes address different topics which don't depend on each other but now can block each other to get merged
2.) code changes get bigger, harder to review and again can block unnecessarily another issue which could be easily merged
3.) The release notes with PR's is cleaner and it is easier to backport specific issues to other releases when they are separate

The problem with NMS-9249, we have to talk about what make sense here and with the additional parameter for the `lookup` and possibility to specify a DNS server, we probably don't need the DNS Monitor anymore?

I would recommend making just the changes in the PR here for NMS-9373 so we get this changes quickly out in "develop" and in the next release.

pbrane:
> On May 30, 2017, at 3:08 AM, Ronny Trommer <notifications@github.com> wrote:
> 
> The problem with NMS-9249, we have to talk about what make sense here and with the additional parameter for the lookup and possibility to specify a DNS server, we probably don't need the DNS Monitor anymore?

Just sounding in here a bit.  The original intent of the DNS Resolution Monitor was to use the system's resolver libraries for the Name Resolution and not to monitor a DNS.  Basically, this was fairly targeted at the Remote Poller where on outage can occur when the system's configuration doesn't support the *lookup* of a host name.  Since the system resolver libraries can use the systems "hosts" file as well as other protocols, such as NIS for example, the idea was to not constraint the the call to resolve a host name as this is probably not how the the other monitors, Page Sequence for example, will handle resolving a URL in it's service configuration.

David

roskens:
With my changes, I think there is too much overlap now between DnsResolutionMonitor and DnsMonitor, and functionality for both should be merged into one of them.

pbrane:
I think that there is a conceptual difference between monitoring a DNS service (DNSMonitor) and testing the resolution of a host name (DnsResolutionMonitor) on a system somewhere in the network using its resolver configuration (/etc/resolv.conf, /etc/hosts, /etc/networks, /etc/named.conf).

Perhaps renaming the DnsResolutionMonitor class to something more accurate such as NameResolutionMonitor and merge you changes into the DnsMonitor class.


David

roskens:
Maybe there should be an AbstractDnsMonitor class that implements the core pieces for both of them, and then DnsServerMonitor (DnsMonitor) implements things from the server perspective, and DnsClientMonitor (DnsResolutionMonitor) implements things from the client perspective?

pbrane:
Well, I still see two use cases.  One is about just DNS monitoring and the other is about name resolution, as it is configured on the system, that may (probably will) eventually end up with one or more DNS queries to get a host name.  So these, IMHO, are two related but different use cases. That doesn't mean that they shouldn't share inherentable core functionality if that makes sense.

indigo423:
For the following reasons I would move support for other protocols like NIS out of scope:

* It does not exist
* I don't see a big need to support monitoring NIS name resolution capabilities
  
IMHO, this use cases should OpenNMS cover nicely for monitoring Network Name resolution and DNS:

1. Test if a "Name" can be resolved using the local systems name service configuration following the rules for `/etc/hosts` in `nsswitch.conf` limited network names which is equivalent to A/AAAA and PTR records and can be achieved with something like `InetAddress.getByName("network-name-to-lookukp")`.
2. The Test if an A/AAAA/MX/CNAME/TXT record from {node label} or custom lookup string can be  
** resolved at all
** resolved to a user defined specific value
** evaluated against a return code of the response
** PTR or PTR6 record from IP address of the assigned monitor exist

For the reason you have to configure different monitors anyway, it could make sense to create smaller monitor classes for the the specific use cases, e.g. 

* for use case 1) and 2) just for the A/AAAA records
* for MX/CNAME/TXT records
* for the reverse lookups PTRv4/PTRv6 records

pbrane:
> On May 31, 2017, at 5:41 AM, Ronny Trommer <notifications@github.com> wrote:
> 
> For the following reasons I would move support for other protocols like NIS out of scope:

I don't think this conversation is going the right direction.  The concept of the resolution monitor is not to monitor any specific protocols.  It simply makes a system call to resolve a hostname.  That's it.  It either resolves or it doesn't.  There is no knowledge of any protocol being used by the *system* to resolve the name.

I'm not sure how this has developed into a discussion about protocols :(

David

indigo423:
> Since the system resolver libraries can use the system "hosts" file as well as other protocols, such as NIS for example ...

That's why I wrote explicitly to move other protocols like NIS out of scope in this discussion.

pbrane:
> On May 31, 2017, at 07:59, Ronny Trommer <notifications@github.com> wrote:
> 
> That's why I wrote explicitly to move other protocols like NIS out of scope in this discussion.

That was used only to simply point out the fact that this particular monitor wasn't protocol specific. Sorry for mentioning a protocol to point out that the monitor wasn't a protocol monitor but I couldn't find another way to explain its intent.

roskens:
Ah, thats why you suggested having a NameResolutionMonitor to monitor that name resolution can be done. The current DNSResolutionMonitor is strictly DNS resolution against a remote DNS. A "new" NameResolutionMonitor would be one that would do as Ronny is saying "InetAddress.getByName()" whcih would return the lookup via system.

pbrane:
> On May 31, 2017, at 2:30 PM, Ron Roskens <notifications@github.com> wrote:
> 
> The current DNSResolutionMonitor is strictly DNS resolution against a remote DNS.

I didn't realize that.  I thought it worked as you described, below.


> A "new" NameResolutionMonitor would be one that would do as Ronny is saying "InetAddress.getByName()" whcih would return the lookup via system.

I agree.


David

soleger:
Now that the confusion has been cleared up about DNSResolutionMonitor, is there any remaining reason not to merge these changes? I've rebased the branch and pushed it to our repo so that it can run through Bamboo but after that it seems worth committing to get these bugfixes.

soleger:
Merged branch, closing this PR.

PR #1516, opened 2017-05-30
NMS-9349: Fixed NPE in VMware Provisioning

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-9349


j-white:
Cherry-picked into `foundation-2017`.

PR #1517, opened 2017-05-30
HZN-950: create sha1 files for Karaf repositories

RangerRick:
This is an alternate PR to https://github.com/OpenNMS/opennms/pull/1497 which does the sha1 generation in maven rather than generating after the fact in the debian/rpm packages.

* JIRA: http://issues.opennms.org/browse/HZN-950

### DIFF
--- a/container/karaf/pom.xml
+++ b/container/karaf/pom.xml
@@ -90,7 +90,7 @@
                             <goal>copy-resources</goal>
                         </goals>
                         <configuration>
-                            <outputDirectory>target/filtered-resources</outputDirectory>
+                            <outputDirectory>${project.build.directory}/filtered-resources</outputDirectory>
                             <resources>
                                 <resource>
                                     <directory>src/main/filtered-resources</directory>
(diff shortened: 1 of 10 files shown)


container/karaf/pom.xml
```diff
@@ -102,6 +102,31 @@
                 </executions>
             </plugin>
             <plugin>
+                <groupId>net.nicoulaj.maven.plugins</groupId>
+                <artifactId>checksum-maven-plugin</artifactId>
+                <version>1.5</version>
```
j-white: Can we set the version in the top-level pom instead in each of these?

opennms-full-assembly/pom.xml
```diff
@@ -301,6 +301,33 @@
...
+                <include>**/*.jar</include>
+                <include>**/*.pom</include>
+                <include>**/*.war</include>
+                <include>**/*.xml</include>
+              </includes>
+            </fileSet>
+          </fileSets>
+        </configuration>
+      </plugin>
+      <!--
```
j-white: If we need to disable the `maven-clean-plugin` here I'd suggest adding a note as to why, or removing the block completely.

RangerRick: Oops, good catch... I had only turned it off while fixing things up so I could inspect the generated repo.  I'll push a fixed rebase.

container/karaf/pom.xml
RangerRick: Alright, did that too.  New build should be happening shortly.

j-white:
Thanks for the updates. Will merge once Bamboo says it's OK.

PR #1518, opened 2017-05-30
NMS-9386: Breadcrumb for requisitions has extra "/app" in path

jeffgdotorg:
* JIRA: http://issues.opennms.org/browse/NMS-9386


agalue:
Looks good to me, good catch!
I'll merge it as soon as Bamboo goes green.

PR #1520, opened 2017-05-31
NMS-9356: Allow provisiond to perform reverse lookups without requiring an A record

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9356

Support reverse lookups without requiring an A record in provisiond. See JIRA for details.


j-white:
Merged to foundation-2017, will evaluate back-porting this to foundation-2016 with NMS-9406.

PR #1525, opened 2017-06-02
HZN-1050: Add support for Jetty request logging (NSCA format)

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1050


RangerRick:
Just tried this out and it works fine.  👍

PR #1536, opened 2017-06-12
NMS-7940: Remove guide-doc and guide-concept

indigo423:
Remove guide-doc and guide-develop and fixed copyright date.

* JIRA: http://issues.opennms.org/browse/NMS-7940


mvrueden:
This issue is supposed to remove the two modules `guide-doc` and `guide-concept` but also includes copyright header changes. I'd split the commits and handle the copyright header changes in a different pull request/issue. 

Side node: I am not sure what the proper copyright header would be. From what we do in the source code it whould be `<creation date of file> - <current year>`

PR #1539, opened 2017-06-13
NMS-9421: Escape favorite tooltip correctly

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-9421


mvrueden:
Please DO NOT MERGE, the favorite filter is also used on the events page. I'd like to take a look and verify that this is not a problem. Thank you.

mvrueden:
Also considered event's index page. @j-white the html-code is escaped correctly and will be rendered accordingly.

j-white:
If one of the filter elements contains a single quote, the full tool-tip is not displayed.

I managed to get around this by removing the call to `WebSecurityUtils.sanitizeString()` and replacing, `title='${favTitle}'` with `title="<c:out value='${favTitle}'/>"`.

mvrueden:
Yeah, I just checked that the tooltip is rendered correctly, but oversaw that the description is not. Should work now.

PR #1541, opened 2017-06-13
HZN-1089: Added missing include statement

christianpape:
* JIRA: https://issues.opennms.org/browse/HZN-1089


mvrueden:
Looks good to me. Can be merged if bamboo goes green.

mvrueden:
The tests are failing. This is probably due to the fact that the branch `release-20.0.1` was not created correctly. Can you please rebase against the actual `release-20.0.1` and see if that fixes the problems?

PR #1543, opened 2017-06-13
NMS-9422: Event Configuration Screen Requires Alarm Type

agalue:
* JIRA: http://issues.opennms.org/browse/NMS-9422

The solution has been manually verified on my test environment.

The solution will require manual merge to foundation-2017, as the Vaadin API is slightly different.

In case this is needed in foundation for Meridian 2015, the solution has to be reworked as well (same thing, because of the Vaadin API).

### DIFF
--- a/features/vaadin-snmp-events-and-metrics/src/main/java/org/opennms/features/vaadin/events/EventForm.java
+++ b/features/vaadin-snmp-events-and-metrics/src/main/java/org/opennms/features/vaadin/events/EventForm.java
@@ -29,10 +29,10 @@
 package org.opennms.features.vaadin.events;
 
 import org.opennms.netmgt.model.OnmsSeverity;
-import org.opennms.netmgt.xml.eventconf.AlarmData;
 import org.opennms.netmgt.xml.eventconf.Logmsg;
 import org.opennms.netmgt.xml.eventconf.Mask;
 
+import com.vaadin.data.Property;
 import com.vaadin.data.fieldgroup.BeanFieldGroup;
 import com.vaadin.data.fieldgroup.FieldGroup.CommitException;
 import com.vaadin.ui.CheckBox;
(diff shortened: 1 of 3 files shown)


mvrueden:
I tried this and the original issue is fixed, but a new one exists.
I tried to edit an event which does not have any alarm data assigned (the checkbox is empty).
If I now add it by enabling the checkbox and set:
 * Alarm Type = 1
 * Auto Clean = true (checkbox checked)
And save, I get the same error message as the original issue indicated.

In addition I think we should add a test which captures this in the future.

agalue:
@mvrueden thanks for testing the branch. I'll work on this and see if I can add some tests. It is Vaadin, so I might need some help with the tests.

agalue:
Is there anything else that require more changes ?

mvrueden:
That was all I found, I haven't given it a try since your fix though. I'll give it a try soon. Assigning to me. Thank you

mvrueden:
I tried this again and it still does not work. I get a "COMMIT FAILED" exception.
The log does not indicate what is going on:
`2017-07-03 17:38:11,951 | ERROR | s/UIDL/?v-uiId=0 | SimpleLogger                     | ?                                   ? | 151 - org.opennms.features.vaadin-snmp-events-and-metrics - 17.0.1.SNAPSHOT | Can't save the changes: Commit failed`

agalue:
> On Jul 3, 2017, at 11:38 AM, mvrueden <notifications@github.com> wrote:
> 
> I tried this again and it still does not work. I get a "COMMIT FAILED" exception.

What exactly you tried? It works for me so it seems I missed a use case. Could you please give me the step by step procedure to reproduce the failure ?

mvrueden:
1) Go to the Admin > Configure Events page.
1) Select 1st file.
1) Select 1st event
1) Hit edit
1) Check "Add AlarmType Data" box
1) Select Type 1
1) Hit checkbox 
1) Hit save
-> error

agalue:
> On Jul 4, 2017, at 3:01 AM, mvrueden <notifications@github.com> wrote:
> 
> Select Type 1
> Hit checkbox
> Hit save
> -> error

It is doing the right thing because the reduction key is mandatory, not optional, so the UI is working.

mvrueden:
Yeah now it works. I did not see the reduction key field. Sorry (-:

PR #1544, opened 2017-06-13
NMS-9413: Validate the agent prior to scheduling the interfaces

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9413

Validate the agent prior to scheduling the interface in order to ensure that the SNMP collector only gets scheduled to run against the primary interfaces.


agalue:
I'll try the branch and let you know later how it goes.

agalue:
I tested the fix with and without Minions. Everything is working as expected.

PR #1547, opened 2017-06-15
HZN-1086, HZN-1087: Upgrade to ActiveMQ 5.14.5 and Camel 2.18.4

soleger:
This updates to the latest versions of ActiveMQ 5.14 and Camel 2.18. No code changes were required, just Maven build issues.

* JIRA: http://issues.opennms.org/browse/HZN-1086
* JIRA: http://issues.opennms.org/browse/HZN-1087


j-white:
We should also remove our custom ConnectionPool implementation as part of the upgrade. We won't need this anymore now that we're on ActiveMQ 5.14.x:

https://github.com/OpenNMS/opennms/blob/40cde003ea22c3f041fad526152db6396dfc6da1/features/activemq/pool/src/main/java/org/opennms/features/activemq/ConnectionPool.java#L40

core/daemon/src/main/resources/META-INF/opennms/applicationContext-daemon.xml
```diff
@@ -55,6 +55,7 @@
     init-method="start" destroy-method="stop">
     <property name="maxConnections" value="#{T(java.lang.System).getProperty('org.opennms.activemq.client.max-connections', '8')}" />
     <property name="idleTimeout" value="#{T(java.lang.System).getProperty('org.opennms.activemq.client.idle-timeout', '30000')}" />
+    <property name="reconnectOnException" value="#{T(java.lang.System).getProperty('org.opennms.activemq.client.reconnect-on-exception', 'true')}" />
```
j-white: If we add a new option here, we should also add it in the Blueprint:
 https://github.com/OpenNMS/opennms/blob/ee96be8496af43550188622abcc0934c25d95bcf/features/activemq/component/src/main/resources/OSGI-INF/blueprint/blueprint.xml#L41

This pool definition is only used by OpenNMS, and the other is used by Minion.

PR #1548, opened 2017-06-15
NMS-9426: Fix thresholding editor

RangerRick:
The thresholding editor was broken by the refactor of JAXB code, and because it uses the old servlet-with-model-objects style, it wasn't caught at compile time.

This fixes all the issues I could find in both rendering and editing of thresholds and expressions.

* JIRA: http://issues.opennms.org/browse/NMS-9426

### DIFF
--- a/opennms-webapp/src/main/webapp/WEB-INF/jsp/admin/thresholds/editExpression.jsp
+++ b/opennms-webapp/src/main/webapp/WEB-INF/jsp/admin/thresholds/editExpression.jsp
@@ -2,8 +2,8 @@
 /*******************************************************************************
  * This file is part of OpenNMS(R).
  *
- * Copyright (C) 2007-2016 The OpenNMS Group, Inc.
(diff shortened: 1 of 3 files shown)


mvrueden:
Can we add tests to catch this in the future?!

RangerRick:
Probably... should they block 20.0.1?  :)

mvrueden:
When trying this I ran into the following issues:
 1. If you hit save, it takes forever until the changes get persisted, but the ui is not locked and one can use it. When the save returns, the ui is closed. This was weird to me.
2. If I create a new Threshold Expression (http://localhost:8980/opennms/admin/thresholds/index.htm?groupName=cisco&newThreshold) and hit save afterwards, I get the "An unknown error occurred" error page.

RangerRick:
Hrm, OK, that worked at one point in my testing of the fixes.  I guess maybe I'll just close this for now and roll it into the other JAXB work I'm doing as part of NMS-9447 and test it all.

PR #1557, opened 2017-06-22
NMS-9427: Fixed NPE in Vaadin dashboard

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-9427


mvrueden:
Shouldn't we provide a test for this?!

mvrueden:
Looks good to me, if bamboo passes, this can be merged.

mvrueden:
Does this also occurr in foundation-2017? If can we retarget it?

christianpape:
The ResourceId changes are not in foundation-2017. So, this can't be applied for foundation-2017.

PR #1561, opened 2017-06-26
 NMS-9365: Path Outage Provider for Topology UI

mvrueden:
A new feature, allowing user to visualize path outages and the status in the topology framework - ready and tested

* JIRA: https://issues.opennms.org/browse/NMS-9365


mvrueden:
The provided (original) requisition was flawed. The node-parent-id was node-parent-label and therefore it could not work. However an issue was found in the requisition ui. See https://issues.opennms.org/browse/NMS-9482 for more details.

I addressed some other issues:
 * topology info was not set 
 * null pointer exception in some cases
 * failing tests

From my point of view this now works as expected. Thank you @pdzaccone for your contribution.

PR #1562, opened 2017-06-26
NMS-9452: Improve logging for EventIpcManagerDefaultImpl

agalue:
* JIRA: http://issues.opennms.org/browse/NMS-9452

The new format is going to be an automatically generated version of all the members of the event related classes using JSON , skipping the null fields and empty collections.

### DIFF
--- a/features/events/api/pom.xml
+++ b/features/events/api/pom.xml
@@ -33,8 +33,8 @@
   </build>
   <dependencies>
     <dependency>
-      <groupId>commons-lang</groupId>
-      <artifactId>commons-lang</artifactId>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
     </dependency>
     <dependency>
       <groupId>org.opennms.core</groupId>
--- a/features/events/daemon/src/main/java/org/opennms/netmgt/eventd/EventIpcManagerDefaultImpl.java
+++ b/features/events/daemon/src/main/java/org/opennms/netmgt/eventd/EventIpcManagerDefaultImpl.java
@@ -175,7 +175,7 @@ public class EventIpcManagerDefaultImpl implements EventIpcManager, EventIpcBroa
                 @Override
                 public void run() {
                     try {
-                        LOG.debug("run: calling onEvent on {} for event {} dbid {} with time {}", m_listener.getName(), event.getUei(), event.getDbid(), event.getTime());
+                        LOG.debug("run: calling onEvent on {} for event {}", m_listener.getName(), event);
 
                         // Make sure we restore our log4j logging prefix after onEvent is called
                         Map<String,String> mdc = Logging.getCopyOfContextMap();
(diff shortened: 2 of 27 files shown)


features/events/daemon/src/main/java/org/opennms/netmgt/eventd/EventIpcManagerDefaultImpl.java
```diff
@@ -175,7 +175,7 @@ public void rejectedExecution(Runnable r, ThreadPoolExecutor executor) {
                 @Override
                 public void run() {
                     try {
-                        LOG.debug("run: calling onEvent on {} for event {} dbid {} with time {}", m_listener.getName(), event.getUei(), event.getDbid(), event.getTime());
+                        LOG.debug("run: calling onEvent on {} for event {}", m_listener.getName(), event.toStringSimple());
```
j-white: This statement should be wrapped in a `if (LOG.isDebugEnabled()) { }` block to avoid calling `Event#toStringSimple()` unnecessarily.

features/events/api/pom.xml
```diff
@@ -33,8 +33,8 @@
   </build>
   <dependencies>
     <dependency>
-      <groupId>commons-lang</groupId>
-      <artifactId>commons-lang</artifactId>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
```
j-white: You'll also need to update the feature definition to depend on commons-lang3:
    https://github.com/OpenNMS/opennms/blob/foundation-2017/container/features/src/main/resources/features.xml#L612

(This is likely why some of the current smoke tests are failing)

agalue:
> On Jun 26, 2017, at 3:10 PM, Jesse White <notifications@github.com> wrote:
> 
> This statement should be wrapped in a if (LOG.isDebugEnabled()) { } block to avoid calling Event#toStringSimple() unnecessarily.


No problem, but calling the standard `toString()` on the Event class is even more expensive, so maybe I should do it everywhere ?

j-white:
`LOG.debug("{}", e)` will only call `e.toString()` if the logger is set to DEBUG, whereas LOG.debug("{}", e.toStringSimple()) will always invoke `e.toStringSimple()`.

features/events/daemon/src/main/java/org/opennms/netmgt/eventd/EventIpcManagerDefaultImpl.java
```diff
@@ -252,7 +252,7 @@ public void sendNow(Event event) {
     public void sendNow(Log eventLog) {
         Assert.notNull(eventLog, "eventLog argument cannot be null");
 
-        LOG.debug("sending: {}", eventLog);
+        if (LOG.isDebugEnabled()) LOG.debug("sending: {}", eventLog);
```
j-white: This one isn't necessary, but it shouldn't do any harm :)

agalue:
> On Jun 26, 2017, at 3:17 PM, Jesse White <notifications@github.com> wrote:
> 
> LOG.debug("{}", e) will only call e.toString() if the logger is set to DEBUG, whereas LOG.debug("{}", e.toStringSimple()) will always invoke e.toStringSimple().

Got it. Either way, I’ve added the check in both statements on EventIpcManagerDefaultImpl.

PR #1563, opened 2017-06-26
NMS-5105: Fix handling of delete related events in collectd

j-white:
JIRA: https://issues.opennms.org/browse/NMS-5105

This fixes a long standing bug where node, interface and service deleted events weren't being handled properly by collectd, leaving the services scheduled after deletion.


RangerRick:
Looks reasonable to me.

PR #1568, opened 2017-06-27
NMS-7526: Support Wifi Connections Link

rssntn67:
NMS-7526: Support Wifi Connections Link

* JIRA: http://issues.opennms.org/browse/NMS-7526


opennms-services/src/main/java/org/opennms/netmgt/enlinkd/NodeDiscoveryWifi.java
```diff
@@ -0,0 +1,100 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class is designed to collect the necessary SNMP information from the
+ * target address and store the collected information. When the class is
+ * initially constructed no information is collected. The SNMP Session
+ * creating and collection occurs in the main run method of the instance. This
+ * allows the collection to occur in a thread if necessary.
+ */
+public final class NodeDiscoveryWifi extends NodeDiscovery {
```
j-white: The whitespace in this file is inconsistent.

opennms-services/src/main/java/org/opennms/netmgt/enlinkd/snmp/MtxrWlRtabTableTracker.java
```diff
@@ -0,0 +1,109 @@
...
+import org.opennms.netmgt.snmp.SnmpInstId;
+import org.opennms.netmgt.snmp.SnmpObjId;
+import org.opennms.netmgt.snmp.SnmpRowResult;
+import org.opennms.netmgt.snmp.TableTracker;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class MtxrWlRtabTableTracker extends TableTracker {
+
+	private final static Logger LOG = LoggerFactory.getLogger(MtxrWlRtabTableTracker.class);
```
j-white: The whitespace in this file is inconsistent.

opennms-services/src/test/java/org/opennms/netmgt/enlinkd/EnLinkdSnmpIT.java
```diff
@@ -143,6 +144,64 @@ public void testInSameNetwork() throws Exception {
     
     @Test
     @JUnitSnmpAgents(value={
+            @JUnitSnmpAgent(host=MIKROTIK_IP, port=161, resource=MIKROTIK_SNMP_RESOURCE)
```
j-white: More inconsistent whitespace.

opennms-services/src/main/java/org/opennms/netmgt/enlinkd/NodeDiscoveryWifi.java
```diff
@@ -0,0 +1,100 @@
...
+    public NodeDiscoveryWifi(final EnhancedLinkd linkd, final Node node) {
+    	super(linkd, node);
+    }
+
+    protected void runCollection() {
+
+        SnmpAgentConfig peer = m_linkd.getSnmpAgentConfig(getPrimaryIpAddress(), getLocation());
+
+        final List<MtrxWlRTabRow> links = new ArrayList<MtrxWlRTabRow>();
+        MtxrWlRtabTableTracker ospfNbrTableTracker = new MtxrWlRtabTableTracker() {
```
j-white: ospfNbrTableTracker?

opennms-services/src/main/java/org/opennms/netmgt/enlinkd/NodeDiscoveryWifi.java
```diff
@@ -0,0 +1,100 @@
...
+    
+            public void processMtrxWlRTabRow(final MtrxWlRTabRow row) {
+    		links.add(row);
```
j-white: I might be missing something, but what do we do with the links?

rssntn67:
Jesse, I just started working on this. Actually we do nothing with links. 
There are two different strategy, the first just create wifi link as have been created  cdp, ospf isis lldp and bridge links.

Another option should be integrate wifi links into BridgeLink. 

I guess the first option is the best and we move to rendering all the problems related to have a coherent layout of the network. 

Also for example you can have routing enabled on wifi links and then you are not able to stay on the same broadcast domain. In this situation the integration with bridge topology will fail.

On the other side all the problems will come back later wehn we try to produce a map. 
But we can also decide to have some smart consolidation function. 

At the end discovering a link (a connection is a relation among more objects) is a little bit different to layout a map!

indigo423:
Is there anything we have to add regarding documentation? I haven't seen something in the changed files. How can I configure it and what information is used to discover a link? A good orientation is what we have right now for the supported protocols in the section: http://docs.opennms.org/opennms/branches/develop/guide-admin/guide-admin.html#ga-enlinkd

If there is anything I can help with, please let me know.

PR #1573, opened 2017-06-28
NMS-9480: Added support for parameterized SQL to CriteriaBuilder.sql()

soleger:
Add a safe, parameterized version of CriteriaBuilder.sql().

* JIRA: http://issues.opennms.org/browse/NMS-9480

### DIFF
--- a/core/criteria/src/main/java/org/opennms/core/criteria/CriteriaBuilder.java
+++ b/core/criteria/src/main/java/org/opennms/core/criteria/CriteriaBuilder.java
@@ -37,6 +37,7 @@ import org.opennms.core.criteria.Alias.JoinType;
 import org.opennms.core.criteria.Fetch.FetchType;
 import org.opennms.core.criteria.restrictions.Restriction;
 import org.opennms.core.criteria.restrictions.Restrictions;
+import org.opennms.core.criteria.restrictions.SqlRestriction.Type;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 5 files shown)


core/criteria/src/main/java/org/opennms/core/criteria/CriteriaBuilder.java
```diff
@@ -282,6 +283,24 @@ public CriteriaBuilder sql(final Object sql) {
         return this;
     }
 
+    public CriteriaBuilder sql(final Object sql, Object parameter, Type type) {
```
j-white: Why not make the `sql` parameter a `String` instead of an object here?

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/HibernateCriteriaConverter.java
```diff
@@ -427,7 +433,28 @@ public void visitBetween(final BetweenRestriction restriction) {
...
+                    case LONG:
+                        return new LongType();
+                    case STRING:
+                        return new StringType();
+                    case TIMESTAMP:
+                        return new TimestampType();
+                    default: 
+                        throw new UnsupportedOperationException("Unsupported type specified in SqlRestriction");
+                    }
+                }).collect(Collectors.toList()).toArray(new org.hibernate.type.Type[0]);
```
j-white: We could use `restriction.getTypes().length` instead of `0` here to avoid re-allocating the array.

opennms-dao/src/test/java/org/opennms/netmgt/dao/AlarmDaoIT.java
```diff
@@ -321,4 +323,18 @@ public void testSortOnNodeLabel() {
...
 	}
+
+	/**
+	 * @see https://issues.opennms.org/browse/NMS-9480
+	 */
+	@Test
+	public void testParameterizedSql() {
+		CriteriaBuilder cb = new CriteriaBuilder(OnmsAlarm.class);
+		cb.sql("{alias}.alarmid in (?)", 1, Type.INTEGER);
+		m_alarmDao.findMatching(cb.toCriteria());
```
j-white: Can we verify the result set as well? Right now all we know is that no exceptions were thrown.

core/criteria/src/main/java/org/opennms/core/criteria/CriteriaBuilder.java
soleger: This was a cut-and-paste from the original method... and I'm not sure why it was ever an ```Object```. I'll switch it to ```String```.

soleger:
I'll address these issues in a separate PR. I'm going to merge this as-is so that I can rebase the RESTv2 branch with these changes ASAP.

PR #1575, opened 2017-06-29
Removed a bunch of unnecessary Maven repositories from POMs

soleger:
This should speed up builds and make them more reliable since fewer repos will be hit when trying to look up Maven artifacts.

Most of these repos are completely unused (JBoss, Spring.io, Servicemix, Codehaus, Apache snapshots, Eclipse). We do use the Sonatype repo to fetch ```dnsjava``` but this PR removes extra references to that repo.

### DIFF
--- a/container/pom.xml
+++ b/container/pom.xml
@@ -18,15 +18,4 @@
     <module>bridge</module>
     <module>jaas-login-module</module>
   </modules>
-
-  <repositories>
-    <repository>
-      <id>servicemix</id>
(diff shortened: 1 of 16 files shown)


RangerRick:
I'm gonna start a dependency monkey on this branch first just so we can be sure it doesn't break things.

mvrueden:
Waiting for @RangerRick response. Assigning to him.

RangerRick:
Dependency monkey passed, looks like it's safe to merge.

PR #1578, opened 2017-06-29
NMS-9293: JDBCDataCollector : do not change custom resourcetype when query returns only one row

jm-k:


features/jdbc-collector/src/main/java/org/opennms/netmgt/collectd/JdbcCollector.java
```diff
@@ -241,8 +241,11 @@ public CollectionSet collect(CollectionAgent agent, EventProxy eproxy, Map<Strin
                             JdbcCollectionResource resource = null;
                             
                             // Create the appropriate resource container.
-                            if(singleInstance) {
-                                resource = nodeResource;
+                            if(null==query.getResourceType()) {
```
j-white: By convention, we typically allow a resource type of `"node"` to refer to node level resources i.e. https://github.com/OpenNMS/opennms/blob/develop/integrations/opennms-vmware/src/main/java/org/opennms/netmgt/collectd/VmwareCollector.java#L247

We should also consider this in addition to checking for `null`.

features/jdbc-collector/src/main/java/org/opennms/netmgt/collectd/JdbcCollector.java
```diff
@@ -241,8 +241,11 @@ public CollectionSet collect(CollectionAgent agent, EventProxy eproxy, Map<Strin
...
+                                if (!singleInstance){
+                                    LOG.warn("Multiple row results but there is no resourceType defined : only first row will be persisted");
+                                    resource = nodeResource;
+                                }
```
j-white: Won't resource be `null` if resource type is defined, and this is a single instance?

j-white:
Closing due to lack of activity. PR is still pending changes.

jb4415:
How do we get this change pushed into production?

PR #1579, opened 2017-06-30
NMS-9481: Clone foreign source requisition overwrites config of other requisition without any hint or warning

agalue:
Clone foreign source requisition overwrites config of other requisition without any hint or warning.

* JIRA: http://issues.opennms.org/browse/NMS-9481

Because a customer opened a support ticket about this for Meridian, I'm applying the solution for `foundation-2016`.

The JavaScript files have been moved as part of the WebUI rework for H19, the solution has to be manually ported to `foundation-2017`. I can do that without issues as soon as this PR is merged.


agalue:
I merged `foundation-2016` into `foundation-2017`, so everything should be ready now.

PR #1581, opened 2017-06-30
NMS-8857: Make an automation to clear pathOutage alarms

agalue:
Make an automation to clear pathOutage alarms

* JIRA: http://issues.opennms.org/browse/NMS-8857


opennms-base-assembly/src/main/filtered/etc/vacuumd-configuration.xml
```diff
@@ -416,7 +428,16 @@
            AND statustime &lt; now() - interval '24 hours'
       </statement>
     </action>
-    
+
+    <action name="clearPathOutages" >
+      <statement>
+        UPDATE alarms
+           SET severity = 2, lastautomationtime = ${_ts}
+         WHERE eventuei = 'uei.opennms.org/nodes/pathOutage'
```
j-white: Should we not also factor in the time of the `nodeUp` alarm here?

I could imaging a scenario, where a node is flapping up and down, and existing `nodeUp` alarm clears a newly created `pathOutage` alarm.

agalue:
> On Jul 4, 2017, at 9:44 AM, Jesse White <notifications@github.com> wrote:
> 
> Should we not also factor in the time of the nodeUp alarm here?
> 
> I could imaging a scenario, where a node is flapping up and down, and existing nodeUp alarm clears a newly created pathOutage alarm.
> 
Sounds reasonable. If you have suggestions, they will be welcome, otherwise, I’ll try to see how can I improve the query.

opennms-base-assembly/src/main/filtered/etc/vacuumd-configuration.xml
agalue: what about tweak the trigger to have the following ?

AND GREATEST(lastautomationtime, lasteventtime) &lt; now() - interval '1 minute'

j-white: I don't see how that statement would guarantee that the `nodeUp` alarm occurred after the `pathOutage` alarm.

If I were to take a go at it, I'd try updating the trigger to pull the `lasteventtime` from the `nodeUp` alarm, and then add this clause to the action:
`AND lasteventtime <= ${_lasteventtime_from_node_up}`

This should guarantee that we received a `nodeUp` event, since the `pathOutage` alarm was last updated.

I might be wrong though, getting the timing right in these statements is tricky.

agalue:
> On Jul 5, 2017, at 4:35 PM, Jesse White <notifications@github.com> wrote:
> I don't see how that statement would guarantee that the nodeUp alarm occurred after the pathOutage alarm.
> 
And I’m not sure how Pollerd (and maybe Alarmd) can guarantee your expectations (i.e. the nodeUp alarm appears AFTER a pathOutage alarm).
> If I were to take a go at it, I'd try updating the trigger to pull the lasteventtime from the nodeUp alarm, and then add this clause to the action:
> AND lasteventtime <= ${_lasteventtime_from_node_up}
> 
> This should guarantee that we received a nodeUp event, since the pathOutage alarm was last updated.
> 
> I might be wrong though, getting the timing right in these statements is tricky.
> 
I’ll give it a try on a "real" test environment to see how that goes.

agalue:
@j-white I tried your suggestion on a VM with Meridian 2016, and it works.

agalue:
I also verified the solution on a VM with 19.1.0

j-white:
:+1:  from me, waiting on Bamboo

PR #1586, opened 2017-07-09
NMS-8484: Implements ColumnPropertyExtender

jmguzmanc:
This class allows using a column value, as an index to get a value from another table.
See: https://issues.opennms.org/browse/NMS-8484


agalue:
Hello, the code looks good, thanks for your contribution.
Would you mind create a JUnit test to ensure that the extender works as expected ?
For references, take a look at https://github.com/OpenNMS/opennms/blob/develop/opennms-services/src/test/java/org/opennms/netmgt/collectd/SnmpCollectorWithMibPropertiesIT.java

j-white:
Closing due to lack activity. PR is still pending changes.

PR #1587, opened 2017-07-10
NMS-8919: Add foreignSource and foreignId parameters when sending events through ReST

agalue:
Add foreignSource and foreignId parameters when sending events through ReST

* JIRA: http://issues.opennms.org/browse/NMS-8919

### DIFF
--- a/core/criteria/src/main/java/org/opennms/core/criteria/Criteria.java
+++ b/core/criteria/src/main/java/org/opennms/core/criteria/Criteria.java
@@ -67,7 +67,7 @@ public class Criteria implements Cloneable {
 	}
 
     public static interface CriteriaVisitor {
-        public void visitClass(final Class<?> clazz);
+        public void visitClassAndRootAlias(final Class<?> clazz, final String rootAlias);
 
         public void visitOrder(final Order order);
 
@@ -95,7 +95,7 @@ public class Criteria implements Cloneable {
     }
 
     public void visit(final CriteriaVisitor visitor) {
-        visitor.visitClass(getCriteriaClass());
+        visitor.visitClassAndRootAlias(getCriteriaClass(), getRootAlias());
 
         for (final Order order : getOrders()) {
             visitor.visitOrder(order);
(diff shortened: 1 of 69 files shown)


j-white:
Can we move this logic into eventd, instead of at the REST endpoint, so that other source of events can leverage this too?

agalue:
> On Jul 11, 2017, at 9:28 AM, Jesse White <notifications@github.com> wrote:
> 
> Can we move this logic into eventd, instead of at the REST endpoint, so that other source of events can leverage this too?

The reason why I didn’t do that was due to the amount of code changes required for that functionality. As you can see on the PR, doing that for ReST was extremely easy, but that’s not the case of Eventd.

j-white:
If we take the same approach of using optional event parameters to resolve the node id, the code shouldn't be much different within eventd itself.

I vote a :-1: on this one it stands since I don't believe that events coming in through REST should get special treatment compared to those received from other sources.

agalue:
> On Jul 13, 2017, at 9:16 AM, Jesse White <notifications@github.com> wrote:
> 
> I vote a 👎 on this one it stands since I don't believe that events coming in through REST should get special treatment compared to those received from other sources.

Please put that comment on the Jira issue so the supported customer who has opened the issue with the expectation to have this in place on an upcoming version, will know about it so they can plan according to it. Their idea was to drastically simplify their complex workflow to manage their inventory, so I think they deserve an explanation.

Thanks

agalue:
Where is the Eventd's code responsible for parsing an event prior adding them to the bus, so I can move the logic there ?

agalue:
Would DefaultEventHandlerImpl.java be the answer to my question ? I'm sorry but it is not obvious (as it is on ReST) what part of Eventd should be responsible for this special parsing which would require access to NodeDao.

agalue:
DefaultEventHandlerImpl seems to be a good place. I've tested it on a branch (I've moved the exact same code from ReST to that class, and I can see the nodeId is properly associated with nodes). I'll create a new PR soon.

agalue:
Here is the new PR: https://github.com/OpenNMS/opennms/pull/1621

PR #1589, opened 2017-07-12
HZN-1092: Upgrade to Karaf 4.1.1

soleger:
This will upgrade us to the latest Apache Karaf. Some minor caveats:

* Pax Web has been downgraded from 6.0 (that comes with Karaf 4.1) to 4.3 (which comes with Karaf 4.0) to work with our HTTP bridge code.
* Bundle auto-refreshing has been disabled inside the KarafExtender because the feature installs were causing an auto-refresh of many of the system bundles which halted the feature installs themselves. This shouldn't be an issue because when the KarafExtender runs, it is running in an empty container so no existing bundles need refreshing.

JIRA: http://issues.opennms.org/browse/HZN-1092
JIRA: http://issues.opennms.org/browse/NMS-9363


j-white:
Here's a couple things I found while testing this:
* The karaf.log file has been moved back to `data/log/karaf.log` (vs `logs/karaf.log`)
* A pid file at `$OPENNMS_HOME/karaf.pid` gets created, I think this is already the case with 4.0, but perhaps we should co-locate this with the `opennms.pid` in `$OPENNMS_HOME/logs/`
* Autocomplete no longer works when using our custom shell commands i.e. `collection:collect`

soleger:
I'm going to close this until the Karaf people address this issue. Otherwise, we'll likely have to change this patch anyway to work with their changes.

https://issues.apache.org/jira/browse/KARAF-5251

PR #1590, opened 2017-07-13
NMS-9478: Remove query parameter from OnmsRestService

soleger:
* JIRA: http://issues.opennms.org/browse/NMS-9478


j-white:
My concern with making these change in foundation is that we are removing functionality in a patch release without providing a suitable replacement.

Perhaps would should introduce a flag (off by default) that can be used to toggle this functionality, and then remove it completely in a major release.

soleger:
I addressed the review suggestions, recommend merging so that it percolates through Bamboo.

PR #1592, opened 2017-07-13
NMS-9133: Added RESTv2 API URL to CORS filter configuration

soleger:
Pretty straightforward. :)

* JIRA: http://issues.opennms.org/browse/NMS-9133


j-white:
It is necessary to add this to foundation-2016? The v2 API is quite limited there, and this may introduce .rpmnew files.

PR #1594, opened 2017-07-13
HZN-1110: Replace Eventd listeners with camel-netty4 endpoints

soleger:
Replaces our TCP/UDP event listener code with 2 simple camel-netty4 endpoints. It uses the XmlFrameDecoder that was added in Netty 4.1 to break the streams into event records.

* JIRA: http://issues.opennms.org/browse/HZN-1110


soleger:
Bump.

features/events/daemon/src/main/java/org/opennms/netmgt/eventd/XmlEventProcessor.java
```diff
@@ -0,0 +1,182 @@
...
+        final InputStream stream = exchange.getIn().getBody(InputStream.class);
+
+        // Unmarshal the XML document
+        Log eventLog = null;
+        try {
+            eventLog = JaxbUtils.unmarshal(Log.class, new InputSource(stream));
+            LOG.debug("Event record converted");
+        } catch (final Throwable e) {
+            LOG.error("Could not unmarshal the XML record", e);
+            // TODO: Throw an exception or something?
```
j-white: Shouldn't we re-throw the exception here? We can the use Camel to keep track of the count of failed exchanges.

features/events/daemon/src/main/java/org/opennms/netmgt/eventd/XmlEventProcessor.java
```diff
@@ -0,0 +1,182 @@
...
+            }
+
+            if (hasReceipt) {
+                // Transform it to XML and send it to the socket in one call
+                final StringWriter writer = new StringWriter();
+                JaxbUtils.marshal(receipt, writer);
+                writer.flush();
+                exchange.getOut().setBody(writer.toString());
+
+                if (LOG.isDebugEnabled()) {
```
j-white: It looks like this whole block can be replaced with: `LOG.debug("Sent Event Receipt: ", writer);`

features/events/daemon/src/main/resources/META-INF/opennms/applicationContext-eventDaemon.xml
```diff
@@ -154,4 +120,62 @@
...
+      <exception>java.lang.Throwable</exception>
+      <handled>
+        <constant>true</constant>
+      </handled>
+      <log loggingLevel="ERROR" message="Unexpected exception while processing event in route ID ${routeId}; not processing with any later processors"/>
+      <log loggingLevel="ERROR" message="${exception.stacktrace}"/>
+    </onException>
+
+    <route id="tcpEventReceiver">
+        <from uri="netty4:tcp://{{tcpIpAddress}}:{{tcpPort}}?receiveBufferSize=2147483647&amp;sync=true&amp;allowDefaultCodec=false&amp;decoders=#tcpXmlFrameDecoder"/>
```
j-white: Does the TCP socket behave in the same way as the previous implementation, or are there any differences?

soleger: It should behave the same as the old TCP listener. In particular, the Syslogd tests send events over Eventd's TCP connection for load testing and the tests work the same as before.

j-white:
LGTM. There are some merge conflicts we need to fix before we can merge though.

soleger:
I'll rebase it, Bamboo, then merge.

pbrane:
> On Aug 9, 2017, at 16:12, Seth Leger <notifications@github.com> wrote:
> 
> I'll rebase it, Bamboo, then merge.

I think we can morph Bamboo into a verb, now:

I'll rebase it, bamboo it, then merge it.  And as a transitive verb: "I bambooed it!"



David Hustace
The OpenNMS Group, Inc.

PR #1595, opened 2017-07-13
NMS-9507: Using varbind to match event definition doesn't work

agalue:
Using varbind to match event definition doesn't work

* JIRA: http://issues.opennms.org/browse/NMS-9507

A customer is expecting a patched JAR for this based on 19.1.0, so I've created the PR to see if there are repercussions about the fix.

In terms of the fix itself, it works on my test environment.

I don't know how to make a JUnit test for this, so someone wants to have one, someone else should do it.


agalue:
Unfortunately, even if the fix solves the problem, there are several trapd related tests that are not passing on Bamboo (verified on my own machine), so I'll let the person who broke this feature to work on the fix.

j-white:
Closing in favor of https://github.com/OpenNMS/opennms/pull/1617.

PR #1597, opened 2017-07-17
NMS-9511: ConcurrentModificationException thrown by Statsd

agalue:
ConcurrentModificationException thrown by Statsd

* JIRA: http://issues.opennms.org/browse/NMS-9511

The solution suggested by @j-white was validated by the customer who reported the problem.


j-white:
While this fix serves as a good workaround, it will also make enumerating resources slower, particularly in the case of Newts.

I think it's worth investigating other ways of isolating the transactions before merging this.

j-white:
I spent some more time looking into this and can't find an easy way to isolate the transaction contexts, while still performing the resource lookups in parallel.

Let's go ahead and and merge this solution and revisit it if there are any performance issues - if there are any, they can also be mitigated by disabling unused resource types.

PR #1601, opened 2017-07-18
Copyright bump in docs

jeffgdotorg:
* JIRA: http://issues.opennms.org/browse/NMS-9516

### DIFF
--- a/opennms-doc/guide-admin/src/asciidoc/index.adoc
+++ b/opennms-doc/guide-admin/src/asciidoc/index.adoc
@@ -8,7 +8,7 @@
 :numbered:
 
 = Administrators Guide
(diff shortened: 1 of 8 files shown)


mvrueden:
Okay, this seems like a regular thing, can we change it to `${year}` and have maven replace it automatically? [Here](https://stackoverflow.com/questions/39307133/maven-property-for-current-year) is described how it is done.

soleger:
It doesn't look like the Asciidoc files undergo Maven filtering yet at all so making this automated is more work than just adding the one extra Maven plugin. I'm going to go ahead and merge this PR, we can address automating it in a later issue.

indigo423:
Documented here: https://issues.opennms.org/browse/HZN-1126

PR #1602, opened 2017-07-18
NMS-9423: An empty category is not shown when in focus

pdzaccone:
Given a a topology provider which supports searching for categories (e.g. Enhanced Linkd), the following issue is present:

 * Clear the focus (No focus defined window appears)
 * Search for a category which does not have any vertices (nodes) assigned
 * The vertex is in focus but the "no focus defined window" is still present
 * collapse the vertex -> The "no focus defined window" is still present

JIRA: https://issues.opennms.org/browse/NMS-9423


smoke-test/src/test/java/org/opennms/smoketest/TopologyIT.java
```diff
@@ -879,6 +884,23 @@ public void verifyPathOutageTopologyInfo() {
...
+     * </p>
+     */
+    @Test(expected = NoSuchElementException.class)
+    public void verifyCollapsibleCriteriaNoDefaultFocusWindow() throws IOException, InterruptedException {
+        topologyUiPage.clearFocus();
+
+        String categoryName = "Servers";
+
+        topologyUiPage.search(categoryName).selectItemThatContains(categoryName);
+        topologyUiPage.testCase.findElementByXpath("//div[@id='gwt-uid-24']");
```
mvrueden: This will fail. The uid is generated and will vary. Probably use the `topologyUiPage.getVisibleVertices()` method and search for a vertex with the label `Servers`

RangerRick:
@mvrueden It looks like you made your own version of a test for this and pushed it to opennms/jira/NMS-9423 so I've gone ahead and pushed @pdzaccone's version as jira/NMS-9423-pdzaccone so we can let Bamboo run through the branch.

mvrueden:
@RangerRick, yeah I pulled this PR in and added some things, as the PR here does not fully fix the problem. However, my changes did not work properly either.

mvrueden:
Closing in favour for https://github.com/OpenNMS/opennms/pull/1684

PR #1603, opened 2017-07-18
Replaced a couple of jquery plugin files with bower-managed versions

soleger:
These have been replaced by bower-managed versions.


mvrueden:
AFAIK both are still used. The treemap.js file is used by the heatmap and the sparkline.js file is used by the trend graph. Am I wrong?!

soleger:
Thanks for the feedback, I didn't check the require.js imports. I'll see about refactoring these scripts over to bower deps.

soleger:
OK, bower deps have been created. The hash version that I used for `jquery.ui.treemap` is identical to what was checked in and it is the terminal version of the library.

PR #1604, opened 2017-07-18
NMS-9505: EnLinkd blocks service startup

rssntn67:
NMS-9505: Enlinkd blocks servce startup

* JIRA: http://issues.opennms.org/browse/NMS-9505


rssntn67:
while not even every problem is fixed with bridge discovery we can safe merge this that the init method works fine. Tested over a 7000 node bridge topology.

PR #1605, opened 2017-07-19
NMS-9512 : Update the OpenNMS spec file to prune empty directories fr…

roskens:
…om the documentation tree.

* JIRA: http://issues.opennms.org/browse/NMS-9512


RangerRick:
Don't see why not...

PR #1607, opened 2017-07-19
HZN-1112: Fix formatting

indigo423:
To make maintenance easier and to avoid unnecessary diffs I have applied the 1 line per sentence rule.


christianpape:
Already fixed.

PR #1609, opened 2017-07-25
NMS-9290: Reload Drools configuration without restart OpenNMS

agalue:
Reload Drools configuration without restart OpenNMS

* JIRA: http://issues.opennms.org/browse/NMS-9290

I ended up for making the reload process extremely flexible and powerful. There are going to be 2 ways to reload the configuration:

1) Send a `reloadDaemonConfig` event, using `DroolsCorrelatorEngine` for the `daemonName` parameter; to add new engines, or remove existing engines.

2) Send a `reloadDaemonConfig` event, using `DroolsCorrelatorEngine-XXX` for the `daemonName` parameter, where `XXX` is the name of the engine (or the `RuleSet` on the configuration files); to reload the configuration of an existing engine, without affecting the execution of all the rest of the running engines.

If there is a mistake on the configuration files, the engine will be removed from the running environment, to avoid having inconsistent code consuming resources. When the configuration is fixed, it can be added as a "new" engine (i.e. using method [1]).

By default, each engine has its own log file (the name of the file would be `DroolsCorrelatorEngine-XXX` as you may expect), so it would be easy to know if the configuration was successfully reloaded or not.


opennms-correlation/drools-correlation-engine/src/main/java/org/opennms/netmgt/correlation/drools/DroolsCorrelationEngineBuilder.java
```diff
@@ -238,16 +270,103 @@ public boolean accept(File file) {
...
+            final String daemonName = getDaemonNameFromReloadDaemonEvent(e);
+            if (daemonName != null && daemonName.equals(getName())) {
+                doAddAndRemoveEngines();
+                return;
             }
         }
-        
+        if (e.getUei().equals(EventConstants.RELOAD_DAEMON_CONFIG_FAILED_UEI)) {
+            final String daemonName = getDaemonNameFromReloadDaemonEvent(e);
+            if (daemonName.startsWith(getName() + '-')) {
```
j-white: This seems redundant if we use the matcher on the next line.

opennms-correlation/drools-correlation-engine/src/main/java/org/opennms/netmgt/correlation/drools/DroolsCorrelationEngine.java
```diff
@@ -85,17 +92,31 @@
...
     private final Meter m_eventsMeter;
+    private MetricRegistry m_metricRegistry;
     private Boolean m_persistState;
+    private Resource m_configPath;
+    private ApplicationContext m_configContext;
     
-    public DroolsCorrelationEngine(final String name, final MetricRegistry metricRegistry) {
+    public DroolsCorrelationEngine(final String name, final MetricRegistry metricRegistry, final Resource configPath, final ApplicationContext configContext) {
         this.m_name = name;
+        this.m_configPath = configPath;
```
agalue: I know, I was just following the convention used on hundreds of files on our code base, but I can remove it needed.

opennms-correlation/drools-correlation-engine/src/main/java/org/opennms/netmgt/correlation/drools/DroolsCorrelationEngine.java
```diff
@@ -311,4 +349,28 @@ public String toString() {
...
+                initialize();
+            } else {
+                ebldr.setUei(EventConstants.RELOAD_DAEMON_CONFIG_FAILED_UEI);
+                ebldr.addParam(EventConstants.PARM_REASON, "RuleSet not found on " + m_configPath);
+            }
+        } catch (Exception e) {
+            ebldr.setUei(EventConstants.RELOAD_DAEMON_CONFIG_FAILED_UEI);
+            ebldr.addParam(EventConstants.PARM_REASON, e.getMessage());
+        } finally {
+            if (ebldr != null) sendEvent(ebldr.getEvent());
```
agalue: good catch, fixed!

opennms-correlation/drools-correlation-engine/src/main/java/org/opennms/netmgt/correlation/drools/DroolsCorrelationEngineBuilder.java
agalue: Indeed it is. I forgot to remove it. Fixed now.

agalue:
I heard that the lack of JUnit tests to verify the reload operation is the reason why this PR has not been merged. These expected tests would be ready by now, if those requirements would have been mentioned either here or on Jira, which is not the case. Either way, I'll adding some tests once I have time.

agalue:
Tests have been added!

PR #1610, opened 2017-07-25
HZN-1114: Add first version of quick search page

soleger:
Dev-Jam 2017 project! Some things are not implemented yet:

- ~~Node search is not done: this is dependent on the RESTv2 node interface being completed~~
- Searching by IP address or IPLIKE expression
- Searching on integer fields
- Searching on enum fields (ie. alarm/event severity, node status, etc.)
- Searching on timestamp fields (which is a weak use case for this style of search page anyway)
- Performance/responsiveness tuning
- The page isn't linked into the UI yet

However, even with these things being incomplete, it is probably useful enough to link into the UI.

JIRA: http://issues.opennms.org/browse/HZN-1114

### DIFF
--- a/opennms-webapp/src/main/webapp/js/angular-onms-quicksearch.js
+++ b/opennms-webapp/src/main/webapp/js/angular-onms-quicksearch.js
@@ -1,10 +1,4 @@
 /**
- * ISO-8601 date format string.
- */
-var ISO_8601_DATE_FORMAT = 'yyyy-MM-ddTHH:mm:ss.sssZ';
-var ISO_8601_DATE_FORMAT_WITHOUT_MILLIS = 'yyyy-MM-ddTHH:mm:ssZ';
-
-/**
  * Convert from a clause into a FIQL query string.
  */
 function toFiql(clauses) {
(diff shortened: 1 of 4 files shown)


indigo423:
Haven't tested it yet but what I've seen at DevJam the UI pieces seem to me as a proof-of-concept to show possibilities with the technology underneath. IMHO, we should have a story and a plan how we move forward with "Search" for this feature before merging to 21.

mvrueden:
From my point of view we should no longer merge proof of concepts or prototypes into the base code line of opennms. The reason for this is, that with OpenNMS development we do not always have the time to finish features we started. Therefore it will probably never leave or for a long time not leave the proof-of-concept state.

I would prefer an approach were we actually define the use case and implement it properly, using this as the base line and go from there.

soleger:
I content that substring typeahead searching of alarms, events, and notifications is worthwhile enough to not be considered a "proof-of-concept" or "prototype" and that this branch should be merged and linked into the "Search" tab in the UI. I listed the possible enhancements to set expectations for people who want to test the page.

soleger:
I'm pulling some of the refactoring into another branch. I'll rebase this branch with those changes after they are merged to `develop`.

soleger:
The rebase is complete. The diff should be much easier to read now.

opennms-webapp/src/main/webapp/js/angular-onms-quicksearch.js
```diff
@@ -0,0 +1,532 @@
...
+ * <li>+</li>
+ * <li>,</li>
+ * <li>;</li>
+ * <li>=</li>
+ * </ul>
+ * @param value
+ * @returns String with reserved characters URL-encoded
+ */
+function escapeSearchValue(value) {
+	if (typeof value === 'string') {
```
RangerRick: Is there any reason you're not just doing the builtin JS `encodeURIComponent()` here instead?  Pretty sure this will be slower *and* we already know CXF is URL-decoding it on their side anyways...

indigo423:
Why I'm not in favor merging this feature as it is for the following reason:

* It is just is a Proof-of-concept about a technical implementation to search for several things and is just another search for things page as we already have
* There no clear concept about how this technical implementation should be used and embedded in the Web UI.
* A feature like this should replace (delete)old search pages

soleger:
It sounds like I still need to rationalize the basic need for a feature like this. Let me try and go point-by-point:

> It is just is a Proof-of-concept about a technical implementation to search for several things and is just another search for things page as we already have

The use case is different. I think somebody at Dev-Jam phrased it as "Spotlight search for OpenNMS" and I think that's accurate. All desktop environments (Mac OS, Windows, GNOME, KDE, etc) support typeahead searching like this that uses metadata to search across different categories of information. I think that the need for a similar feature in OpenNMS is clear.

> There no clear concept about how this technical implementation should be used and embedded in the Web UI.

I think the concept is clear: typeahead search for OpenNMS business objects. I think that "embedding" it by creating a menu item is reasonable. There are probably other more creative things we could do (embedding it in the header, etc) but none of that work is a blocker for merging a feature like this.

> A feature like this should replace (delete) old search pages

I think that the use case is separate from any existing search pages which search for values in specific fields of object instead of globally searching key fields in all objects. Can you point out examples of overlap?

mvrueden:
To have a _Spotlight Search_ within OpenNMS in general is reasonable and makes sense, in my opinion. A _Spotlight Search_ requires a central place and should be accessible very easy (on every page and context). 

However, the current proposal lacks the "general search concept" and proposes yet another search page. In this state I would not recommend merging this, because of the following reasons:
 * We are not good at getting back to once merged features, especially if they are not done yet
 * Having said that, if we ever going to need a _Spotlight Search_, we already have a PoC implementation in the code base and then need to support/merge that (then old) code, or rewrite it completely.

At this point I'd rather put more resources into this feature and give it a more _Spotlighty_ look and feel.

pbrane:
> On Nov 20, 2017, at 08:22, mvrueden <notifications@github.com> wrote:
> 
> At this point I'd rather put more resources into this feature and give it a more Spotlighty look and feel.

+1

indigo423:
Totally agree with @mvrueden. With something like "Spotlight" we have a story to talk about. This technology proof-of-concept enables it but we need some iterations here.

mvrueden:
Closing PR for now, as the common tenor is to put more resources in this feature.

PR #1611, opened 2017-07-25
NMS-9513: Programmable Northbounder Interfaces

agalue:
Programmable Northbounder Interfaces

* JIRA: http://issues.opennms.org/browse/NMS-9513


agalue:
Is there something that prevents merging this PR ?

PR #1612, opened 2017-07-28
NMS-9542: Microsoft IE and Edge doesn't like relative path on Angular apps

agalue:
Microsoft IE and Edge doesn't like relative path on Angular apps.

The solution is always use full URLs, which works on any browser (including IE and Edge). Tested against H20.

* JIRA: http://issues.opennms.org/browse/NMS-9542

### DIFF
--- a/opennms-webapp/karma.conf.js
+++ b/opennms-webapp/karma.conf.js
@@ -29,6 +29,8 @@ module.exports = function(config) {
       'src/main/webapp/lib/angular-bootstrap/ui-bootstrap-tpls.js',
       'src/main/webapp/lib/angular-loading-bar/build/loading-bar.js',
       'src/main/webapp/lib/angular-growl-v2/build/angular-growl.js',
+      // Global Variables/Methods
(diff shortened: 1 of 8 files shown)


soleger:
Do some of these changes also need to be backported to `foundation-2016`? Like the provisioning UI, quick-add node page, etc.

agalue:
> On Aug 1, 2017, at 1:14 PM, Seth Leger <notifications@github.com> wrote:
> 
> Do some of these changes also need to be backported to foundation-2016? Like the provisioning UI, quick-add node page, etc.

No. The Requisitions UI on foundation-2016 uses the old style (prior the JavaScript refactoring made for H19). Old style means, always use full URLs. After the refactoring, everything went back to relative paths which unfortunately doesn’t work on IE/Edge, but works on all the other serious browsers. Now, the problem with relative paths appears to affect only window.location.href, it doesn’t affect Angular’s $http, but just in case, now everything uses full URLs.

PR #1613, opened 2017-07-31
NMS-9548: initial-delay="null" when adding a new destination path.

agalue:
initial-delay="null" when adding a new destination path.

The default should be "0s" like in older versions of OpenNMS (including Meridian 2016).

* JIRA: http://issues.opennms.org/browse/NMS-9548


RangerRick:
Yup, good catch.

PR #1621, opened 2017-08-02
NMS-8919: Add foreignSource and foreignId parameters when sending events as an alternative to nodeId

agalue:
Add foreignSource and foreignId parameters when sending events as an alternative to nodeId

Please use the [JIRA](https://issues.opennms.org) issue number and create a link in the JIRA issue back to this pull request so we have a quick reference from the issue to the pull request.

* JIRA: http://issues.opennms.org/browse/NMS-8919

### DIFF
--- a/features/events/daemon/src/main/java/org/opennms/netmgt/eventd/DefaultEventHandlerImpl.java
+++ b/features/events/daemon/src/main/java/org/opennms/netmgt/eventd/DefaultEventHandlerImpl.java
@@ -35,6 +35,8 @@ import java.util.Objects;
 import org.opennms.netmgt.events.api.EventHandler;
 import org.opennms.netmgt.events.api.EventProcessor;
 import org.opennms.netmgt.events.api.EventProcessorException;
+import org.opennms.netmgt.dao.api.NodeDao;
+import org.opennms.netmgt.model.OnmsNode;
 import org.opennms.netmgt.xml.event.Event;
 import org.opennms.netmgt.xml.event.Events;
 import org.opennms.netmgt.xml.event.Log;
@@ -70,6 +72,8 @@ public final class DefaultEventHandlerImpl implements InitializingBean, EventHan
     private final Timer processTimer;
 
     private final Histogram logSizes;
+
+    private NodeDao m_nodeDao;
 
     /**
      * <p>Constructor for DefaultEventHandlerImpl.</p>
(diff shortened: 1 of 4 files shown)


agalue:
Here are my thoughts ... Even if at first glance this makes sense, here is why I did it at ReST level in the first place, on my first attempt:

Now, for every single event that comes in through Eventd, the nodeId will be checked, as well as getting the foreignSource and foreignId from the parameters list if exist regardless the nature of the event. Fortunately, the NodeDao call will be executed only when the conditions met. Maybe this won't add any harm in terms of performance, but through ReST this would have done only when it is used (in other words, when necessary, instead of every single time).

features/events/daemon/src/main/java/org/opennms/netmgt/eventd/DefaultEventHandlerImpl.java
```diff
@@ -119,6 +123,17 @@ public void run() {
             }
 
             for (final Event event : events.getEventCollection()) {
+                if (event.getNodeid() == 0) {
+                    final Parm foreignSource = event.getParm("_foreignSource");
+                    final Parm foreignId = event.getParm("_foreignId");
```
j-white: If `foreignSource` was not found, or has a null value, we don't need to continue search for the the `foreignId`.

features/events/daemon/src/main/java/org/opennms/netmgt/eventd/DefaultEventHandlerImpl.java
```diff
@@ -119,6 +123,17 @@ public void run() {
...
+                    if (foreignSource != null && foreignSource.getValue() != null && foreignId != null && foreignId.getValue() != null) {
+                        final OnmsNode node = getNodeDao().findByForeignId(foreignSource.getValue().getContent(), foreignId.getValue().getContent());
+                        if (node != null) {
+                            event.setNodeid(Long.parseLong(node.getId().toString()));
```
j-white: We can avoid the cast to a string by using: `node.getId().longValue()`

features/events/daemon/src/main/java/org/opennms/netmgt/eventd/DefaultEventHandlerImpl.java
```diff
@@ -119,6 +123,17 @@ public void run() {
             }
 
             for (final Event event : events.getEventCollection()) {
+                if (event.getNodeid() == 0) {
+                    final Parm foreignSource = event.getParm("_foreignSource");
+                    final Parm foreignId = event.getParm("_foreignId");
+                    if (foreignSource != null && foreignSource.getValue() != null && foreignId != null && foreignId.getValue() != null) {
+                        final OnmsNode node = getNodeDao().findByForeignId(foreignSource.getValue().getContent(), foreignId.getValue().getContent());
+                        if (node != null) {
```
j-white: Should we log a warning if/when the node is not found?

agalue:
@j-white thanks for all your suggestions. All have been applied.

agalue:
> On Aug 8, 2017, at 12:07 PM, Jesse White <notifications@github.com> wrote:
> 
> The previous PR also included some updates the to the documentation for the REST endpoint, should we include this here too?


On the previous PR it was obvious where to put that documentation. Unfortunately, it is not clear to me where a similar documentation can be added for Eventd. Once I get a clear suggestion about it, I’ll be more than happy to include it.

j-white:
Can't we just use the same paragraph from the previous PR. Those notes are still relevant here.

agalue:
> Can't we just use the same paragraph from the previous PR. Those notes are still relevant here.

I can certainly do it, but as I mentioned on my last reply, "where" should I put that information ? It is not obvious. It looks like section 6 (i.e.  "Events") from the Administrators Guide is a good candidate, but inside of it, it is not clear where should I that. At the end ? Somewhere else ?

agalue:
> It looks like section 6 (i.e.  "Events") from the Administrators Guide is a good candidate, but inside of it, it is not clear where should I that. At the end ? Somewhere else ?

I’ve added a sub-section to the "Event Bus" with some information. Keep in mind that I've asked twice where to add the information, and I've decided to place it there in absence of suggestions. I hope this is enough to merge the PR.

j-white:
Thanks for the updates.

PR #1624, opened 2017-08-03
NMS-9538: Document supported products and versions for Remedy integration

jeffgdotorg:
* JIRA: http://issues.opennms.org/browse/NMS-9538

### DIFF
--- a/opennms-doc/guide-admin/src/asciidoc/index.adoc
+++ b/opennms-doc/guide-admin/src/asciidoc/index.adoc
@@ -193,6 +193,7 @@ include::text/operation/daemon-config-files/pollerd.adoc[]
 == Ticketing
 include::text/ticketing/introduction.adoc[]
 include::text/ticketing/jira.adoc[]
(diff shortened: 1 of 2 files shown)


opennms-doc/guide-admin/src/asciidoc/index.adoc
```diff
@@ -213,4 +214,4 @@ include::text/plugins/opennms-es-rest.adoc[]
 
 [[ga-special-cases-and-workarounds]]
 == Special Cases and Workarounds
-include::text/workarounds/snmp.adoc[]
\ No newline at end of file
+include::text/workarounds/snmp.adoc[]
```
RangerRick: I think previously it was missing a carriage return at the EOF and his edit added one, or vice-versa.

jeffgdotorg: I think @RangerRick is right, it's a net no-op that I didn't intend to make.

PR #1630, opened 2017-08-08
NMS-9562: Add support for targeting a specific Minion with a RPC:

j-white:
Here we add support for targeting a specific Minion with a remote procedure call, and leverage this to:
1) Automatically perform JMX data collection against the Minions
2) Monitor directed RPCs and graph response time with the MinionRpcMonitor


core/ipc/rpc/utils/src/main/java/org/opennms/core/rpc/utils/RpcTargetHelper.java
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * Utility class used to compute the target (location, system-id) for some
+ * service oriented (i.e. monitor, collector) request.
+ */
+public class RpcTargetHelper {
+
+    protected static final String LOCATION_KEY = "location";
+    protected static final String SYSTEM_ID_KEY = "system-id";
+    protected static final String USE_FOREIGN_IS_AS_SYSTEM_ID_KEY = "use-foreign-id-as-system-id";
```
RangerRick: Should this be `USE_FOREIGN_ID_AS_SYSTEM_ID_KEY`?

RangerRick:
Haven't tried setting up an environment to run it, but I scanned through the code and it looks pretty sane to me.

PR #1638, opened 2017-08-23
HZN-1098: Add typeahead search endpoints for RESTv2 property values

soleger:
This branch adds a Hibernate-based typeahead query to the RESTv2 services. Values are returned using a `JaxbListWrapper` of primitive, `Date` or `String` values.

This PR also adds an `iplike` field to the metadata to indicate if an `iplike` expression can be used as a query value.

TODO: Add `limit` support to the autocomplete endpoints. I'll do this ASAP but this PR could be merged as-is.

* JIRA: http://issues.opennms.org/browse/HZN-1098


soleger:
I need to split this branch into foundation-2017 bits (for RESTv2) and develop (for notification elementList improvements).

mvrueden:
I tried this with the Performance Management Datasource and this is what I observed:
 * Result is not sorted
 * Result may not be limited
 * Queriing for `uei` or `ipAddr` are limited to alarms. This may work for `uei`, but we should return all existing `ipAddr` instead.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/AbstractDaoRestService.java
```diff
@@ -275,21 +297,115 @@ public Response getProperties(@QueryParam("q") final String query) {
...
+        public List<T> doInHibernate(Session session) throws HibernateException, SQLException {
+            final Query hql;
+            // TODO: Add limit?
+            // TODO: Sort by count?
+
+            // If there is a query string...
+            if (query != null && query.length() > 0) {
+                hql = session.createQuery(
+                    String.format(
+                        "select distinct %s from %s where lower(%s) like :query order by %s",
```
soleger: The results appear sorted in the REST results (JSON and XML) so the lack of ordering must be in the JS client or higher.

soleger:
> Queriing for uei or ipAddr are limited to alarms. This may work for uei, but we should return all existing ipAddr instead.

Can you clarify what you mean here? Which endpoint are you querying and how do the results differ from what you expect?

mvrueden:
`select distinct ipaddr from ipinterface` (expected) vs `select distinct ipaddr from alarms` (actual)

soleger:
You can reach that list by querying:
`http://127.0.0.1:8980/opennms/api/v2/alarms/properties/ipInterface.ipAddress`
instead of:
`http://127.0.0.1:8980/opennms/api/v2/alarms/properties/ipAddr`

PR #1640, opened 2017-08-24
HZN-1143: Use DTOs in REST v2 endpoints for events and alarms

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1143

Here we update the alarm and event REST endpoints in the v2 API to use DTOs.

The DTOs are built from the corresponding entities using mappers, built with the help of MapStruct.

The DTOs are maintained in a separate project, with minimal dependencies on other artifacts so that these can be easily imported by other projects.


soleger:
This should be in foundation-2017 with the rest of the RESTv2 enhancements.

j-white:
Right. I'll re-open this once the rebase to `foundation-2017` is complete.

PR #1642, opened 2017-08-24
NMS-8856: Resource graph forecast page lacks header

agalue:
Resource graph forecast page lacks header

* JIRA: http://issues.opennms.org/browse/NMS-8856

### DIFF
--- a/opennms-webapp/src/main/webapp/WEB-INF/jsp/graph/results.jsp
+++ b/opennms-webapp/src/main/webapp/WEB-INF/jsp/graph/results.jsp
@@ -197,8 +197,8 @@
      <div class="panel-body">
         <!-- NRTG Starter script 'window'+resourceId+report -->
         <script type="text/javascript">
-            function nrtgPopUp(resourceId, report) {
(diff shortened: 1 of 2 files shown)


j-white:
These were intentionally omitted because the original implementation opened as a pop-up window, similar to the NRTG button. Personally, I'd prefer to see this open as a pop-up rather than adding those headers.

Also, the current breadcrumbs on this page only show the link to 'Home', there is nothing to return the user back to the 'Resource Graphs' or the 'Results' page.

agalue:
That makes sense, and would explain why the headers were omitted, but it is surprising to find that it was implemented this way instead of launching a pop-up window (which is not hard to do, and I agree it is better). This is just another example of the consequences of not having things well documented. So, should I close this one, and create another that actually opens a new pop-up window (even knowing that some browsers block this ability by default) ?

agalue:
@j-white I have another solution (tested on my environment), based on the current NRTG implementation that will open a popUp for the Forecast page (similar to NRTG) which will use the `graph/forecast.jsp` unchanged, but will require a small change on `WEB-INF/jsp/graph/results.jsp`. Let me know if that's more aligned for what you think we should have as a solution so I can go ahead and close this PR and create another one (or re-use the branch by reverting the change if that's easier).

j-white:
Yeah, I'd go with the pop-up instead.

agalue:
> Yeah, I'd go with the pop-up instead.

Done!

PR #1648, opened 2017-08-29
NMS-9587: Expose troubleTicketerLink via rest

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-9587

### DIFF
--- a/core/upgrade/src/test/resources/etc2/opennms.properties
+++ b/core/upgrade/src/test/resources/etc2/opennms.properties
@@ -188,7 +188,7 @@ opennms.ticketer.plugin=org.opennms.netmgt.ticketd.NullTicketerPlugin
 
 #opennms.ticketer.plugin=org.opennms.netmgt.ticketer.otrs.OtrsTicketerPlugin
 #opennms.alarmTroubleTicketEnabled = true
-#opennms.alarmTroubleTicketLinkTemplate = <a href="http://localhost/otrs/index.pl?Action=AgentTicketZoom&TicketNumber=\${id}">\${id}</a>
+#opennms.alarmTroubleTicketLinkTemplate = http://localhost/otrs/index.pl?Action=AgentTicketZoom&TicketNumber=\${id}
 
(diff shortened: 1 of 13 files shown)


agalue:
I would say that besides updating the documentation, this should be mentioned on the "breaking changes" for the target version, as there are users already using this feature and they should be aware that they have to update opennms.properties after the upgrade.

Or, as part of this solution, an upgrade task can be created to remove the anchor tag and leave the plain URL inside opennms.properties if the template is used (or if it is active).

mvrueden:
I don't think change the template url is not that big of an issue. An update will create a merge conflict, won't it?! 

In addition it is easy to fix if anyone encounters it.

Having a "breaking change" (or "update notes", as I would call it) section in the documentation would be nice.

agalue:
> I don't think change the template url is not that big of an issue.

I think it is, as you have to update opennms.properties after an upgrade to continue using the feature.

> An update will create a merge conflict, won't it?!

I’m not sure what you mean here.
> In addition it is easy to fix if anyone encounters it.
> 
It is easy to fix, but not necessarily obvious that you have to do it, and you might will encounter that you have to do it, when it is too late, or when an operator, who might not have access to OpenNMS, see this and requires this functionality. That’s why at least mention it is the minimum thing I would to in order to have this in place.
> Having a "breaking change" (or "update notes", as I would call it) section in the documentation would be nice.
> 
Indeed, and as I said, this is the minimum required thing to do. An update task would be wonderful, but if this is just too much (even if this can be done with less than 10 lines of code), document this as breaking change should be sufficient for the users and customers.

mvrueden:
Added an hint to the releasenotes "Breaking Changes" section.

soleger:
It might be a good idea to post about the `<a>` removal on the devel mailing list. It seems like the only reason that the template was made that way would have been so that somebody could make something other than a simple link, ie. a link with an image, a button, or something like that.

j-white:
We've opted to go ahead with this change as is.

If users need more control over how the link is displayed, we'll introduce additional properties.

PR #1654, opened 2017-09-01
NMS-9497: fix alarm-type on BSM events to be 3 rather than 1

RangerRick:
This PR sets the alarm-type on the BSM events to `3` (no possible resolution) rather than `1` (has a possible resolution) and introduces liquibase code to update existing BSM alarms to have the right type in the database.

* JIRA: http://issues.opennms.org/browse/NMS-9497

### DIFF
--- a/core/schema/src/main/liquibase/changelog.xml
+++ b/core/schema/src/main/liquibase/changelog.xml
@@ -79,6 +79,7 @@
 	<include file="19.0.0/changelog.xml"/>
 	<include file="19.1.0/changelog.xml"/>
 	<include file="19.1.1/changelog.xml"/>
+	<include file="foundation-2017/changelog.xml"/>
 
 	<include file="stored-procedures/getManagePercentAvailIntfWindow.xml" />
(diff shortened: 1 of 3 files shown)


pbrane:
This seems strange to me that BSM alarms wouldn't have resolutions.

agalue:
> This seems strange to me that BSM alarms wouldn't have resolutions.

It is to me as well, but that’s how the logic was defined. In fact, if you see the definition, the exact same alarm "is updated" when a serviceProblemResolved event comes in, which is the main reason why the alarm-type should be 3. Why that logic is that way ? The people whom created this feature should know.

pbrane:
> On Sep 1, 2017, at 12:02, Alejandro Galue <notifications@github.com> wrote:
> 
> It is to me as well, but that’s how the logic was defined. In fact, if you see the definition, the exact same alarm "is updated" when a serviceProblemResolved event comes in, which is the main reason why the alarm-type should be 3. Why that logic is that way ? The people whom created this feature should know.

I'll review this, again, today.

PR #1656, opened 2017-09-05
NMS-9605: Fixed wrong event column name

fooker:
* JIRA: http://issues.opennms.org/browse/NMS-9605


fooker:
I can't reproduce this by myself. But this seems to be a no-brainer.

PR #1657, opened 2017-09-05
HZN-1146: Added eventparm search to RESTv2 endpoints

soleger:
Now that the event parameters are in a separate database table, we can filter based on parameters in our Criteria API and, subsequently, in the RESTv2 services.

This PR adds eventparm filtering support to the alarm, event, and notification endpoints.

* JIRA: http://issues.opennms.org/browse/HZN-1146

### DIFF
--- a/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/README.adoc
+++ b/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/README.adoc
@@ -242,6 +242,16 @@ org.opennms.web.rest.support.SearchPropertiesToAsciidocTest
 | event.ipAddr | IP Address | IP Address
 |===
 
+[[eventParameterProperties]]
+.Event Parameter Properties
+[options="header",width="99%",cols="2m,1,3"]
+|===
+| Name | Type | Description
+| eventParameter.name | String | Name
+| eventParameter.type | String | Type
+| eventParameter.value | String | Value
+|===
+
 [[ipInterfaceProperties]]
 .IP Interface Properties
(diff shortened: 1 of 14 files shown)


opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/README.adoc
```diff
@@ -242,6 +242,16 @@ org.opennms.web.rest.support.SearchPropertiesToAsciidocTest
 | event.ipAddr | IP Address | IP Address
 |===
 
+[[eventParameterProperties]]
+.Event Parameter Properties
+[options="header",width="99%",cols="2m,1,3"]
+|===
+| Name | Type | Description
+| eventParameter.name | String | IP Address
```
j-white: IP Address?

soleger: Ah, I fixed that in the code but forgot to re-run the doc generator. Good catch!

j-white:
After taking another look, I did find a problem.

When searching with:
```
_s=eventParameter.name==category;eventParameter.value==ENVIRONMENTAL
```

we actually return all events that have some event parameter name 'category' and some event parameter with value 'ENVIRONMENTAL'. 

We don't actually search for a parameter named 'category' with value 'ENVIRONMENTAL'.

Would it be possible to pair these tokens?

j-white:
This works better now, when searching for both a name and value, only entries containing the name/value pair are returned.

The only limitation I found is that searching for multiple key/value pairs is not supported (the result set is empty), but this is something we can live with for the time being.

PR #1658, opened 2017-09-05
HZN-1150: Allow users with ROLE_DELEGATE to acknowledge alarms on behalf of other users

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1150

Here we add a new role, ROLE_DELEGATE, that allows alarms to be acknowledged on behalf of other users, without required ROLE_ADMIN.


RangerRick:
This looks pretty straightforward. 👍

pbrane:
So is the the role that should be assigned to a user used in via the ReST AP? Would this user require both roles?

David

j-white:
This role does not add any additional Spring security grants (i.e. access to the REST API), so yes the users will need multiple roles.

PR #1662, opened 2017-09-07
NMS-9608: Restored update() function in AbstractDaoRestServiceWithDTO

soleger:
The `update()` function was mistakenly erased from `AbstractDaoRestServiceWithDTO` (previously known as `AbstractDaoRestService`) during the refactoring of the RESTv2 code that took place in preparation for the Helm enhancements. This was causing edits to the Minions, monitoring locations, and scan reports to fail via REST.

* JIRA: http://issues.opennms.org/browse/NMS-9608


RangerRick:
Looks pretty straightforward.

PR #1663, opened 2017-09-12
NMS-9558: BridgeTopologyDao: refatoring code

rssntn67:
the BridgeTopologyInMemoryDao is not properly defined.
Move all persisting stuff to BridgeTopologyDaoHibernate and remove any reference to BridgeBridgeLinkDao and BridgeMacLinkDao from enlinkd service to the BridgeTopologyDao.
 
Al in memory stuff must be moved to EnlinkdService.  Reorder the persisting of BroadcastDomain topology using:
 * a Hibernate class to persist to database; 
 * enlinkd service to manage in memory objects

JIRA: http://issues.opennms.org/browse/NMS-9558


RangerRick:
Went ahead and switched this to target develop since it's not fixing a particular bug and we're too close to Meridian 2017 and a new Horizon release based on `foundation-2017` to merge it there.

PR #1666, opened 2017-09-13
NMS-9618: Fix the 'Quick Add Node' modal in the provisioning UI

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9618

Adds a missing dependency. Fixes a regression from NMS-9353.


RangerRick:
Hah, seems easy enough.  :)

PR #1669, opened 2017-09-14
NMS-9108: Fix effective location when handling newSuspect events

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9108

The logic for determining the effective location when handling `newSuspect` events is now:
* Use the value of the location parameter if set
* Otherwise, use the source location of the event, if it is associated
  with a particular system
* Fallback to the using the default location


RangerRick:
Confirmed, works as expected.

PR #1672, opened 2017-09-14
HZN-1151: Replace outage list with Angular implementation

soleger:
This PR replaces the legacy outage list with a new Angular/RESTv2 implementation.

Advantages:
- Metadata-based typeahead filtering for all fields that can be joined to an outage (ie. filter outages by locations, asset fields, SNMP fields, etc)
- The list is dynamically resizable without hacking the URL
- Jumps straight into the list without the top-level `outages/index.jsp` entry page
- Removes 3,000 lines of ancient JSP/Spring MVC code

Disadvantages:
- Uses a new URL format so bookmarked links will break. However, the new searching is so easy to use that you could recreate your bookmarks easily.

**JIRA:** https://issues.opennms.org/browse/HZN-1151

Also addresses parts of many other JIRA issues:
* https://issues.opennms.org/browse/NMS-832
* https://issues.opennms.org/browse/NMS-2439
* https://issues.opennms.org/browse/NMS-7147
* https://issues.opennms.org/browse/HZN-427


indigo423:
I've noticed some weird behavior when you switch between current, resolved and both, page is not updated correctly. I've noticed @mvrueden had similar issues when he implemented the alarm list, outage list pages for the Alarm Donut Charts. Do we use the same things here? IMHO we have two implementations for the outage list, one from the Alarm Donut Charts and the rewrite here. IMHO it would make sense to just use one. Probably @mvrueden can give some input here, cause he spent already some time and did some work which is published in develop.

indigo423:
FYI: https://demo.opennms.eu runs the branch If you want to play with it.

indigo423:
In the current stable release, we have two versions of the outage list, one JSP and an AngularJS version. IMHO, when we spend the time replacing the JSP version, I would prefer to have in result one outage list written in AngularJS and not two AngularJS versions.

mvrueden:
I am bit confused here. If this replaces the existing jsp version, why is the `outages/main.html` file not part of this PR?

soleger:
The Angular version has been in the codebase for several releases but was never linked into the menus or links in the UI. This PR has 2 commits:
1. Remove the legacy JSP-based outage list
2. Rename the Angular-based outage list from `indexNew.jsp` to `index.jsp` so that it takes the place of the old JSP-based list

WRT to comments from @indigo423, that's exactly what the result of this PR is: no more JSP implementation and 1 Angular implementation.

indigo423:
@soleger This is the URL when I click on "Resolved"

https://demo.opennms.eu/opennms/outage/index.jsp?_s=ifRegainedService!%3D1970-01-01T01:00:00.000%2B0100

mvrueden:
I just tried this locally and it works so far.

However as soon as I switch between "Both", "Current" or "Resolved" the ui breaks.
The request made returns a 500.
The URL is as follows:
`http://localhost:8980/opennms/api/v2/outages?_s=ifRegainedService!=1970-01-01T01:00:00.000+0100&limit=20&offset=0&order=asc&orderBy=id`
Probably the `+0100` in the timerange is causing the 500 (probably similar to https://issues.opennms.org/browse/NMS-9578).

All the (new) angular pages contain duplicated code. For example:
* No centralized filter component to allow the re-usage of filters on other pages.
* No centralized pagination component to allow the re-usage of pagination on other pages

smoke-test/src/test/javascript/tests/outages.js
```diff
@@ -12,38 +12,32 @@ casper.test.begin('Outage Page', {
...
-		casper.then(function() {
-			casper.waitForAlert(function(response) {
-				casper.test.assertEquals(response.data, 'Please enter a valid outage ID.', 'Alert box should show when clicking "Get details" without an outage ID.');
-				return true;
-			}, function() {
-				casper.test.fail('No alert box showed when clicking "Get details".');
-			});
-			casper.clickLabel('Get details');
-		});
+		// TODO: Test filters, ordering
```
indigo423: Merge with TODO?

soleger:
@indigo423 and @mvrueden, what browsers are you guys using? Because it appears that you encountered different behavior on the same page (Ronny's browser is encoding the timestamp as `1970-01-01T01:00:00.000%2B0100` and Markus' is displaying `1970-01-01T01:00:00.000+0100`). Maybe it is just a difference in how the URL is displayed... in any case, it's obvious that the +XX:XX timezone is causing a problem and I will look into this.

soleger:
@indigo423 @mvrueden The date issues that affected the _Current_ and _Resolved_ buttons should be resolved now if you'd like to retest the branch.

j-white:
Given the large Javascript refactor we did in 22.0.0, there are now a large number of conflicts here.

Closing pending updates.

PR #1674, opened 2017-09-14
LTS-215: make sure an empty ncs directory is not created in core

RangerRick:
This PR fixes the RPMs to not create the drools ncs directory in the main (core) RPM, but to leave it to the NCS package.

* JIRA: http://issues.opennms.org/browse/LTS-215


RangerRick:
FYI, don't merge this quite yet, I have a bit of verification to do.

RangerRick:
OK, confirmed this is all good and ready to merge.

PR #1678, opened 2017-09-20
NMS-9495: Migrate docs to AsciiBinder framework

indigo423:
Migrated install-, admin-, user-, development-guide and release notes to AsciiBinder framework. 
Created a basic OpenNMS stylesheet. The maven module opennms-docs are removed. Refactored the how to write docs and contribution part to reflect now the AsciiBinder framework workflow.

Note: The yum and debian packaging is a part which is not refactored.

* JIRA: http://issues.opennms.org/browse/NMS-9495


j-white:
Generally speaking, I like the new structure and layout, but I'm a 👎 here until we can get a better solution for search.

indigo423:
Agree search needs to be solved.

indigo423:
Work preserved here: https://github.com/opennms-forge/opennms-doc

PR #1687, opened 2017-09-26
NMS-9468: vacuumd config for service cleanup

mfuhrmann:
A cleanup job to remove unassigned services. See https://issues.opennms.org/browse/NMS-9468

In my system it works fine and I don't get any issues because of this config. But please make sure, that the entries the automation deletes, aren't required for anything else.

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/vacuumd-configuration.xml
+++ b/opennms-base-assembly/src/main/filtered/etc/vacuumd-configuration.xml
@@ -74,6 +74,9 @@
                 trigger-name="selectPathOutagesNodes"
                 action-name="clearPathOutages" />
 
+    <automation name="deleteUnassignedServices" interval="3600000" active="true"
+                action-name="deleteUnassignedServices" />
+
     <!-- Enable these automation for automatic Trouble Ticketing -->
     <automation name="createTickets" interval="30000" active="false" 
                 trigger-name="selectNullTicketStateAlarms" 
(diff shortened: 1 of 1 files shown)


pbrane:
Instead of a subselect within the action, this should use a trigger which would also allow you to send a deletion event for each row that is deleted.

indigo423:
@pbrane is this adding unnecessary load on the DB? Just running one query which deletes everything you want to delete vs. selecting first what you want to delete and run a delete statement for each element found in the selectstatement?

pbrane:
I don't think that there will be that much impact on the DB.  There generally are not that many services and since this is cleanup, then there shouldn't be much overhead in the removal of the service.  Also, iterating over each service allows you to send a any events that need to be sent to inform the system during the clean up.

indigo423:
@pbrane agree, 99% there is nothing to do a and just a select is cheap. Forget about my comment :)

mfuhrmann:
But we don't need an action event, aye?
In my opinion it's not necessary to have an event which tells us, that service ABC which isn't used anymore was deleted from the drop down box on the starting page.
Wdyt?

pbrane:
> On Nov 15, 2017, at 09:23, Marcel Fuhrmann <notifications@github.com> wrote:
> 
> But we don't need an action event, aye?
> In my opinion it's not necessary to have an event which tells us, that service ABC which isn't used anymore was deleted from the drop down box on the starting page.
> Wdyt?

Anytime something is deleted from the system, the corresponding informational event should be sent.  You never know what other components rely on that information.  Sometimes, components need to know so that they can update their cache of services, etc.

This is the reason we ask users to not use DB calls directly and use the ReST API.  For example, if you used the ReST API to delete something, all the proper events would be sent.

This also why I would like to get rid of automations (Vaccumd) and use something like Drools that can make calls to a service layer that is also used by the ReST API so that deleting a service via an automation or via the ReST API or directly within Provisiond code, it is all handled with the same code.

So, in the meantime we should have an action-event from the automation to simulate that there is actually a service for deleting things. ;)


David

indigo423:
It is not possible to easily delete a service in OpenNMS. When you delete the service manually, events with the associated service id stay in the database and causes exceptions in the Web UI (.e.g. Events List).

One possible solution would be a delete on cascade from services to events which deletes all events associated with this event.

indigo423:
👎 We should close this PR unmerged, we need to change the code to allow deletion of services without side effects. Issues are collected here: https://issues.opennms.org/browse/NMS-9822

mfuhrmann:
So let's close it :-)

PR #1688, opened 2017-09-26
NMS-9649: Align WS-Management collection definitions with WMI ones

jeffgdotorg:
Also made a few small enhancements and fixes along the way.

* JIRA: http://issues.opennms.org/browse/NMS-9649


j-white:
Some of these resources types referenced here used the `IndexStorageStrategy`.

From the [docs](http://docs.opennms.org/opennms/releases/20.1.0/guide-admin/guide-admin.html#ga-performance-mgmt-collectors-wsman-collector):

When using a generic resource type, the IndexStorageStrategy cannot be used since records have no implicit index. Instead, you must use an alternative such as the SiblingColumnStorageStrategy.

jeffgdotorg:
Good catch, thanks @j-white . Will fix those up.

jeffgdotorg:
Resource-type problem addressed. I left the node-level attributes unchanged.

j-white:
Cool. Looks better now. :+1:

PR #1691, opened 2017-10-03
NMS-9660: trapd may silently discard invalid traps

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9660

Log any errors that occur from asynchronously sending the traps via the Sink dispatcher and update the error counters accordingly.

Also, refactored the counters to use atomics to avoid unnecessary locks.


RangerRick:
This looks pretty straightforward, and passed bamboo.  Merging.

PR #1692, opened 2017-10-04
NMS-9385: move netmask to the ipinterface table

RangerRick:
This PR moves the `snmpinterface.snmpipadentnetmask` column to `ipinterface.netmask` and changes all of the places that we reference netmask from the `OnmsSnmpInterface` object to its equivalent in `OnmsIpInterface`.

* JIRA: http://issues.opennms.org/browse/NMS-9385


j-white:
Looks good to me too. (No more FIXMEs either :+1:)

PR #1694, opened 2017-10-04
NMS-6287: convert config files to match JAXB output

RangerRick:
This PR integrates @agalue's config normalizer over from the Meridian tree, and includes a commit with the output of the normalizer, with some non-trivial comments added back in.

* JIRA: http://issues.opennms.org/browse/NMS-6287


j-white:
How do we make sure that these files stay normalized going forward?

agalue:
> How do we make sure that these files stay normalized going forward?

That’s a very good question. Are we doing something on Meridian ? If not, we should.

RangerRick:
Good question. Making something automated would require actually handling comments properly, which I would imagine is a pretty large change to make.

j-white:
Perhaps we can use xmlunit to compare the tags, white-space and ordering, but ignore the comments.

RangerRick:
Alright, updated to do xmlunit stuff, and guess what? things be broke...

1. `jmx-config.xml` is annotated with `@XmlAccessorType(XmlAccessType.FIELD)` but then the getter on the `mbean-server` method is annotated as an element, so it was writing everything out twice, with different tags
2. `snmptrap-northbounder-config.xml` varbind `max` field was an `int`, so it was rendering to 0 when written out, instead of skipping the element
3. `ifttt-config.xml` had an empty category field that was just not written out when re-marshalled
4. `jms-northbounder-configuration.xml` requires the entries to be in a certain order but the config file is actually in a different order

Force-pushed a new branch with `WillItUnmarshalIT` now doing xmldiff tests, if this goes green it should be ready for merge.

The config file bugs I've gone ahead and committed to `foundation-2017` or `develop` as appropriate, and rebased this branch to include the fixes.

j-white:
With the xmlunit changes in, I'm assuming we now expect the `WillItUnmarshalIT` to fail if the files are not formatted properly.

If someone is contributing configuration, and this test fails, what's the easiest way for them to reformat the file appropriately?

RangerRick:
Add it to the normalizer code in `opennms-tools` and run that on the directory the config is in.  Might be worth looking into making a "jaxbutils" cli tool for convenience's sake, but that feels kinda out of scope for this.

j-white:
Good to know, we can address it separately.

It doesn't seem reasonable to require users that contribute configuration to change/compile Java source to get the files formatted properly.

RangerRick:
Alright, after finding *more* busted things, I've fixed all errors that I've found and cleaned up more JAXB stuff, and I think this is ready to merge.

mvrueden:
@RangerRick asked me to verify the JMX Configuration Generator UI changes. Tried them out and everything worked. 👍  for removing the duplicated classes. I was not able to use the `jmx-config-generator` command line tool, as I was not able to connect to the existing opennms. I think this is a different issue and not caused by your changes.

PR #1698, opened 2017-10-08
HZN-1092: Upgrade to Karaf 4.1.2

soleger:
This will upgrade us to the latest Apache Karaf 4.1.2. Some minor caveats:

* Pax Web has been downgraded from 6.0 (that comes with Karaf 4.1) to 4.3 (which comes with Karaf 4.0) to work with our HTTP bridge code.
* Bundle auto-refreshing has been disabled inside the KarafExtender because the feature installs were causing an auto-refresh of many of the system bundles which halted the feature installs themselves. This shouldn't be an issue because when the KarafExtender runs, it is running in an empty container so no existing bundles need refreshing.
* Autocomplete is broken in our custom Karaf commands because we are using the Karaf 2.4 shell API. I'll upgrade the commands in a separate PR because adding those changes here touches a lot of files and would make this PR harder to review.

JIRA: http://issues.opennms.org/browse/HZN-1092
JIRA: http://issues.opennms.org/browse/NMS-9363


soleger:
This PR replaces #1589.

smoke-test/src/test/java/org/opennms/smoketest/minion/DetectorsCommandIT.java
```diff
@@ -178,13 +178,11 @@ public void canLoadDetectorsOnOpenNMS() throws Exception {
...
 
             // Verify
-            for (String detectorName : expectedDetectors.keySet()) {
-                if (!detectorMap.containsKey(detectorName)) {
-                    unmatchedDetectors.add(detectorName);
-                }
-            }
+            return expectedDetectors.keySet().stream().filter(e -> {
+                // If no line matches the expected value, return it 
+                return detectorMap.keySet().stream().noneMatch(s -> s.contains(e));
```
j-white: Using contains isn't quite reliable since: `"HTTPS".contains("HTTP") == true`

I'm guessing there's extra white-space in the output?

container/standalone/src/main/distribution/etc/system.properties
```diff
@@ -158,3 +153,14 @@ karaf.secured.services = (&(osgi.command.scope=*)(osgi.command.function=*))
...
+# OPENNMS: Disable the Bouncy Castle security provider for Apache MINA SSHD
+# because it silently fails if Bouncy Castle is in the system classpath
+#
+# @see https://github.com/apache/mina-sshd/blob/master/sshd-core/src/main/java/org/apache/sshd/common/util/security/SecurityUtils.java
+#
+#org.apache.sshd.security.registrars = \
+#  org.apache.sshd.common.util.security.bouncycastle.BouncyCastleSecurityProviderRegistrar,\
+#  org.apache.sshd.common.util.security.eddsa.EdDSASecurityProviderRegistrar
+#
+org.apache.sshd.security.registrars = org.apache.sshd.common.util.security.eddsa.EdDSASecurityProviderRegistrar
```
j-white: I think we'll just need to remove BouncyCastle instead. We've had some issues with it in the past: https://issues.opennms.org/browse/NMS-7959

core/test-api/karaf/src/main/java/org/opennms/core/test/karaf/KarafTestCase.java
```diff
@@ -249,10 +259,27 @@ private static int getAvailablePort(int min, int max) {
...
             //editConfigurationFilePut("etc/system.properties", "org.ops4j.pax.exam.rbc.rmi.port", paxExamRmiRegistryPort),
+
+            // Work around bug KARAF-5223, should be unnecessary once we upgrade past Karaf 4.1.1
+            replaceConfigurationFile("etc/shell.init.script", emptyFile),
         };
 
+        // Work around bug KARAF-5384
+        // If there is a MINA JAR available, then add MINA to the startup classpath
+        // TODO: Don't hardcode the version number here
+        File minaJar = new File("target/dependency/mina-core-2.0.16.jar");
```
j-white: What happens if we upgrade Mina and forget to update this block? I'm guessing that the Karaf integration tests will run unreliably.

j-white:
I encountered a couple of other issues while testing this, but these also affect the current Karaf 4.0.x container:
* [HZN-1175](https://issues.opennms.org/browse/HZN-1175): Karaf shell bundle command fails with "Insufficient credentials" error
* [HZN-1174](https://issues.opennms.org/browse/HZN-1174): features:install vs feature:install in admin guide

j-white:
I think that most of these issues can be fixed once this is merged. I'm concerned about the updates to `MonitorsListCommandIT` and `DetectorsCommandIT` though.

core/test-api/karaf/src/main/java/org/opennms/core/test/karaf/KarafTestCase.java
soleger: They should fail 100% without that.

container/standalone/src/main/distribution/etc/system.properties
soleger: BC is required for the following enhancement because it is a JRadius dependency and JRadius uses it to do EAP-TTLS (afaik from looking at the imports in JRadius):

https://issues.opennms.org/browse/NMS-9416

Maybe we should shade JRadius ourselves to include BC?

smoke-test/src/test/java/org/opennms/smoketest/minion/DetectorsCommandIT.java
soleger: The output has ANSI color codes provided by JLine.

container/standalone/src/main/distribution/etc/system.properties
j-white: As discussed, avoiding issues like those in NMS-7959 is more important then supporting the feature in NMS-9416. We should probably just exclude BC from JRadius, and users can drop it in lib/ if they need it.

PR #1700, opened 2017-10-10
NMS-9448: Verify that database and system time match

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-9448

### DIFF
--- a/core/db-install/src/main/java/org/opennms/core/db/install/InstallerDb.java
+++ b/core/db-install/src/main/java/org/opennms/core/db/install/InstallerDb.java
@@ -43,6 +43,7 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.sql.Timestamp;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


core/db-install/src/main/java/org/opennms/core/db/install/InstallerDb.java
```diff
@@ -2322,4 +2323,25 @@ public void vacuumDatabase(final boolean full) throws SQLException {
...
+
+        try (Statement st = getConnection().createStatement()) {
+            try (ResultSet rs = st.executeQuery("SELECT NOW()")) {
+                if (rs.next()) {
+                    final Timestamp currentDatabaseTime = rs.getTimestamp(1);
+                    final long currentSystemTime = System.currentTimeMillis();
+                    final long diff = currentDatabaseTime.getTime() - currentSystemTime;
+                    if (Math.abs(diff) > 1000) {
+                        m_out.println("NOT OK");
+                        throw new Exception("Database time and system time differ. Please update either the database time or the system time");
```
j-white: It might be a good idea to include the (formatted) timestamps and the delta in the exception message.

core/db-install/src/main/java/org/opennms/core/db/install/InstallerDb.java
```diff
@@ -2322,4 +2323,25 @@ public void vacuumDatabase(final boolean full) throws SQLException {
...
+    public void checkTime() throws Exception {
+        m_out.print("- checking if time of database \"" + getDatabaseName() + "\" is matching system time... ");
+
+        try (Statement st = getConnection().createStatement()) {
+            try (ResultSet rs = st.executeQuery("SELECT NOW()")) {
+                if (rs.next()) {
+                    final Timestamp currentDatabaseTime = rs.getTimestamp(1);
+                    final long currentSystemTime = System.currentTimeMillis();
+                    final long diff = currentDatabaseTime.getTime() - currentSystemTime;
+                    if (Math.abs(diff) > 1000) {
```
j-white: We may also want to include the time taken to execute the query as part of the calculation i.e. if a GC were to occur after the executing the SELECT, but before calling `System.currentTimeMillis()`, the result may be inaccurate.

mvrueden: The result may always be inaccurate. If I were to take the execution time of the calculation in consideration, I would have the same effect. GC can always occur. 
The only thing I could do, to minimize the issue is to trigger a GC before running the check. Otherwise I don't know what you mean.

j-white: I guess what I'm trying to say is that there's a lot that can happen between the time we execute the SELECT statement and when control is passed back to this block.

If we take this time into account using something like:
```
final long beforeQueryTime = System.currentTimeMillis();
try (ResultSet rs = st.executeQuery("SELECT NOW()")) {
    if (rs.next()) {
        final Timestamp currentDatabaseTime = rs.getTimestamp(1);
        final long currentSystemTime = System.currentTimeMillis();
        final long diff = currentDatabaseTime.getTime() - currentSystemTime;
        final long queryExecuteDelta = Math.abs(currentSystemTime - beforeQueryTime);
        if (Math.abs(diff) > 1000 + queryExecuteDelta) {
	...
```
then we'll have more confidence in identifying a true mismatch.

PR #1703, opened 2017-10-13
NMS-9684: Refactor compatibility for current version

indigo423:
Remove version compatibility for old versions and give the user the information which software is compatible with the version the documentation is released with.

* JIRA: http://issues.opennms.org/browse/NMS-9684


j-white:
Makes sense to me. We're already missing a row in the H20 docs.

soleger:
If we're only displaying information for a single version, it probably makes sense to use a list instead of table for the data. 

Is there anywhere that we can still have a compatibility matrix, like on the .org site? I feel like we really need one because of the increasingly large number of systems that we integrate closely with. People need this information to make informed decisions about which version of OpenNMS they need to install or upgrade to.

indigo423:
no problem can change it to a list. I would prefer to generate more of this type of information instead of manual editing. In the website we can query JIRA and versions and generate links like here: https://www.opennms.org/en/opennms/versions

One idea could be, add an additional column named compatibility and just link it in the docs to correct page.

This is one of the reasons I would prefer to drive releases from JIRA and follow a clear consistent version schema and project management, e.g.

* NMS vs. HZN in JIRA: where NMS is used for releases and release notes and HZN needs manual love
* 20.1.0 vs. 20.1.0-1: appending versions for packaging

soleger:
Links to the docs may give you the information, but it would require many clicks and trial-and-error to see, for instance, what versions of OpenNMS support your preferred (or already-installed) version of Elasticsearch.

A tabular matrix view is the standard for showing compatibility.

indigo423:
https://www.opennms.org --> OpenNMS/Versions --> Version you are interested in

This is IMHO an improvement to the current solution which requires more clicks to come to the current compatibility list with a high chance having no reliable information.

mhuot:
Let’s try that again. I agree with Seth a matrix is usually the way this kind of data is presented.

Mike

indigo423:
All I want to say, is we should not build this "compatibility matrix" in the individual release of the documentation for the following reason. We duplicate for every n+1 release information and publish it. If you have to backport for some reasons things to meridian you have not just to fix it for the releases, you have to touch everything in between. I would recommend to make a compatibility list for every version and document it in each release. If you want a compatibility list over all versions, create it externally make it manual or find a way to generate it using the information from individual docs. I would prefer a generated version.

soleger:
It will have to be created manually in my opinion because the two axes of the table:

- What we're compatible with
- Versions that are compatible

are loosely-defined and also could quickly become unwieldy or contain redundant information (for instance, if we had a row for every micro release of Horizon and Meridian). However, i don't think that this is a very intensive task. We only need to update the matrix following major releases.

And just to make sure everyone's on the same page, the doc that you are replacing was intended to be an "installation" compatibility matrix. As such, it was designed to only contain the systems that you may be integrating with at installation time (Cassandra, Kafka, etc).

@indigo423 it sounds like your main concern is removing inconsistent information from the docs and that's fine but I'd prefer not to delete the matrix view without replacing it somewhere, ie. opennms.org.

indigo423:
Created compatibility list here: https://wiki.opennms.org/wiki/Installation_and_Upgrades

Reachable through:

* https://wiki.opennms.org -> Installation & Upgrades
* https://www.opennms.org -> Installation -> Link Installation & Upgrades

This allows fixing wrong information immediately. Maintained in the source of OpenNMS means, it requires doing a release for OpenNMS.

j-white:
+1, the current matrix is already out of date and has been for a while: https://issues.opennms.org/browse/NMS-9804

indigo423:
Maintained branch and updated version information based on the wiki.

indigo423:
Changed base to 21.0.3 instead of 22.0.0

indigo423:
Any news here? Merge or don't apply?

j-white:
I'm still a 👍 here, the fact that we still don't have an entry for 21.x months after the release reinforces that we don't have a good system in place for keeping this up to date.

indigo423:
@mhuot / @soleger any additions 2 vs. 2 😄

soleger:
If it changes to a list view I'm +1 otherwise I'm +0 (now that the compatibility matrix is on the wiki).

indigo423:
@soleger no problem, I'll change it to a list view in the existing branch. But it seems like Ben -1 and closed it :)

PR #1704, opened 2017-10-13
NMS-9671: Fixing cross-site vulnerability in heatmap

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-9671

### DIFF
--- a/opennms-webapp/src/main/webapp/heatmap/heatmap-box.jsp
+++ b/opennms-webapp/src/main/webapp/heatmap/heatmap-box.jsp
@@ -1,4 +1,4 @@
-<%--
+<%@ page import="org.opennms.core.utils.WebSecurityUtils" %><%--
 /*******************************************************************************
  * This file is part of OpenNMS(R).
  *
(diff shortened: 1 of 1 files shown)


j-white:
If we apply `WebSecurityUtils.sanitizeString` to the parameters and use the results to build URLs we risk breaking legitimate cases when the parameters may contain special characters (i.e. a `<`).

I think we should URL encode the parameters when using them to build URLs and use `WebSecurityUtils.sanitizeString` exclusively for rendering the values in HTML.

(We may not expect all or any of these variables to include special characters, but I think this is good practice to follow.)

j-white:
Can we re-target this for `foundation-2016`, it appears that the same code there is also affected.

christianpape:
Rebase for foundation-2016 will not works since there were a lot of JS changes like requirejs. Should I create another branch/fix for foundation-2016?

j-white:
Yeah, let's fix foundaiton-2016 as well.

j-white:
The page is still vulnerable since the parameters are appended directly to the `url` variable without being encoded: https://github.com/OpenNMS/opennms/blob/72b5e48ee7e0ab6f1dd96303f59c9c4b5413866f/opennms-webapp/src/main/webapp/heatmap/heatmap-box.jsp#L230

christianpape:
Sorry, missed that one.

PR #1705, opened 2017-10-16
HZN-1169: Provide Netflow5 Parser and integrate with existing telemetryd

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1169

Code has still some TODOs. It needs to be decided weather this must be done now, or can be investigated  further down the road.

Please squash merge.


mvrueden:
I just noticed, that the documentation is missing. Will add.

features/telemetry/adapters/netflow/src/main/java/org/opennms/netmgt/telemetry/adapters/netflow/v5/NetflowHeader.java
```diff
@@ -0,0 +1,96 @@
...
+    private static final int BODY_SIZE = 48;
+
+    private final ByteBuffer data;
+
+    public NetflowHeader(ByteBuffer data) {
+        this.data = data;
+    }
+
+    public int getVersion() {
+        return new IntField(0, 1, data, getOffset()).getValue();
```
j-white: Why do we call `getOffset()` if it's always 0?

features/telemetry/adapters/netflow/src/main/java/org/opennms/netmgt/telemetry/adapters/netflow/v5/NetflowPackage.java
```diff
@@ -0,0 +1,90 @@
...
+    public int getVersion() {
+        return getHeader().getVersion();
+    }
+
+    public boolean isValid() {
+        // ensure fields can be set
+        boolean valid = getVersion() == VERSION
+                && getHeader().getCount() >= MIN_COUNT && getHeader().getCount() <= MAX_COUNT
+                && getRecords().size() == getHeader().getCount();
+            //TODO MVR I noticed some packages are longer than they should be. May worth investigating
```
j-white: Perhaps we could add a "lenient vs strict" option to the adapter that allows us to control whether or not these cases are accepted.

features/telemetry/adapters/netflow/src/main/java/org/opennms/netmgt/telemetry/adapters/netflow/v5/NetflowHeader.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    public int getCount() {
+        return new IntField(2, 3, data, getOffset()).getValue();
```
j-white: The value may be cached within the `AbstractField` class, but we create a new instance of the field on every get, so we still end up rebuilding the value.

This is fine if we access the value once, but will add unnecessary overhead otherwise.

features/telemetry/adapters/netflow/src/main/java/org/opennms/netmgt/telemetry/adapters/netflow/v5/fields/InetAddressField.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Override
+    protected String extractValue() {
+        byte[] bytes = getFieldBytes();
+        if (bytes == null ||bytes.length != 4) {
+            throw new IllegalStateException("Could not parse field to InetAddress");
+        }
+        try {
+            return InetAddress.getByAddress(bytes).toString().replaceFirst("/", "");
```
j-white: Can't we just use `InetAddressUtils` here?

features/telemetry/adapters/netflow/src/main/java/org/opennms/netmgt/telemetry/adapters/netflow/Netflow5Adapter.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Override
+    public void handleMessageLog(TelemetryMessageLogDTO messageLog) {
+        LOG.debug("Received {} telemetry messages", messageLog.getMessages().size());
+
+        for (TelemetryMessageDTO eachMessage : messageLog.getMessages()) {
+            LOG.debug("Parse log message {}", eachMessage);
+
+            // Create NetflowPackage which delegates all calls to the byte array
+            final NetflowPackage flowPackage = new NetflowPackage(eachMessage.getBytes());
```
j-white: `Package` or `Packet`?

mvrueden: In Netflow speak it would be `Packet` instead of `Package`. For me `Package` sounds better, but for me as a non-native speaker there is no difference, so feedback is welcome (-:

features/telemetry/adapters/netflow/src/main/java/org/opennms/netmgt/telemetry/adapters/netflow/v5/NetflowHeader.java
mvrueden: No particular reason. I can make it a constant or just use 0 instead.

features/telemetry/adapters/netflow/src/main/java/org/opennms/netmgt/telemetry/adapters/netflow/v5/NetflowHeader.java
mvrueden: The original prototype worked a bit different, while refactoring it out I missed to not re-instantiate the fields every time. This makes no sense. LoL.

features/telemetry/adapters/netflow/src/main/java/org/opennms/netmgt/telemetry/adapters/netflow/v5/NetflowPackage.java
mvrueden: That makes sense. I'd like to learn a bit more about this before doing this, though.

features/telemetry/adapters/netflow/src/main/java/org/opennms/netmgt/telemetry/adapters/netflow/v5/fields/InetAddressField.java
mvrueden: Originally the idea was to not have any dependencies to a opennms library. That way we could release the "flow package parsing" as it's own project and have others benefit from the work. If we don't want to do that, in fact we should use `InetAddressUtils` instead of replicating that logic here.

opennms-base-assembly/src/main/filtered/etc/telemetryd-configuration.xml
```diff
@@ -1,2 +1,11 @@
 <?xml version="1.0"?>
-<telemetryd-config />
+<telemetryd-config>
+  <protocol name="Netflow-5" description="Listener for Netflow 5 UDP packages" enabled="false">
```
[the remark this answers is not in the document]
mvrueden: Yep, thank you.

mvrueden:
I made some changes. I squashed the earlier work into one single commit and added more commits: 
1) makes the implementation simpler
1) package vs packet fix

j-white:
Looks good. I think the only things left here now are:
1) Handling of "invalid" packets
2) Documentation

As long as we have separate issues to track these, I think we're good to merge this.

mvrueden:
I rebased against `features/drift` to actually make it mergeable. I also created 2 follow up issues. Can be merged when green.

PR #1706, opened 2017-10-16
NMS-9668: Use OWASP HTML sanitizer

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9668
JIRA: https://issues.opennms.org/browse/NMS-9669

Here we update WebSecurityUtils to use the OWASP HTML Sanitizer for safely rendering strings that we allow to contain HTML.

Our current protection is quite limited since we only filter out a few known patterns.

### DIFF
--- a/container/features/src/main/resources/features.xml
+++ b/container/features/src/main/resources/features.xml
@@ -257,8 +257,13 @@
       <bundle>wrap:mvn:org.json/json/20140107$Export-Package=org.json</bundle>
     </feature>
 
-    <feature name="owasp-encoder" description="owasp-encoder" version="${owaspEncoderVersion}">
+    <feature name="owasp-encoder" description="OWASP :: Encoder" version="${owaspEncoderVersion}">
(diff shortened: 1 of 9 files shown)


container/features/src/main/resources/features.xml
```diff
@@ -257,10 +257,15 @@
       <bundle>wrap:mvn:org.json/json/20140107$Export-Package=org.json</bundle>
     </feature>
 
-    <feature name="owasp-encoder" description="owasp-encoder" version="${owaspEncoderVersion}">
+    <feature name="owasp-encoder" description="OWASP :: Encoder" version="${owaspEncoderVersion}">
       <bundle>wrap:mvn:org.owasp.encoder/encoder/${owaspEncoderVersion}$Export-Package=org.owasp.encoder</bundle>
     </feature>
 
+    <feature name="owasp-html-sanitizer" description="OWASP :: HTML Sanitizer" version="${waspHtmlSanitizerVersion}">
```
RangerRick: Shouldn't this be "${owaspHtmlSanitizerVersion}" ?

j-white: Why, yes, yes it should. Nice catch.

PR #1711, opened 2017-10-17
HZN-1164: System tests for handling JTI payloads

cgorantla:
Add system tests for handling JTI payloads both at OpenNMS and Minion side.

* JIRA: https://issues.opennms.org/browse/HZN-1164

### DIFF
--- a/smoke-test/pom.xml
+++ b/smoke-test/pom.xml
@@ -356,6 +356,11 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.opennms.features.telemetry.adapters</groupId>
+      <artifactId>org.opennms.features.telemetry.adapters.jti</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
       <groupId>org.apache.camel</groupId>
       <artifactId>camel-elasticsearch</artifactId>
       <version>${camelVersion}</version>
--- a/smoke-test/src/test/java/org/opennms/smoketest/minion/TelemetryIT.java
+++ b/smoke-test/src/test/java/org/opennms/smoketest/minion/TelemetryIT.java
@@ -1,0 +1,314 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2016-2017 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2017 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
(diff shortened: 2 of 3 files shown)


smoke-test/pom.xml
```diff
@@ -356,6 +356,11 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.opennms.features.telemetry.adapters</groupId>
+  	  <artifactId>org.opennms.features.telemetry.adapters.jti</artifactId>
```
j-white: Nit: whitespace is off

smoke-test/src/test/java/org/opennms/smoketest/minion/TelemetryIT.java
```diff
@@ -0,0 +1,340 @@
...
+    public static final TestEnvironment getTestEnvironment() {
+        if (!OpenNMSSeleniumTestCase.isDockerEnabled()) {
+            return new NullTestEnvironment();
+        }
+        try {
+            final TestEnvironmentBuilder builder = TestEnvironment.builder().all();
+            builder.withOpenNMSEnvironment()
+                    .addFile(TelemetryIT.class.getResource("/telemetryd-configuration.xml"),
+                            "etc/telemetryd-configuration.xml")
+                    .addFile(TelemetryIT.class.getResource("/telemetryd-adapters/junos-telemetry-interface.groovy"),
```
j-white: We should be able to use the existing Groovy script instead of re-bundling it here.

smoke-test/src/test/java/org/opennms/smoketest/minion/TelemetryIT.java
```diff
@@ -0,0 +1,340 @@
...
+        }
+    }
+
+    @Before
+    public void checkForDocker() {
+        Assume.assumeTrue(OpenNMSSeleniumTestCase.isDockerEnabled());
+    }
+
+    @Test
+    public void verifyJtiListenerOnOpenNMS() throws IOException, InterruptedException, JAXBException {
```
j-white: Most of the code in the Minion and OpenNMS tests is the same. Can we refactor this to share the common blocks?

smoke-test/src/test/java/org/opennms/smoketest/minion/TelemetryIT.java
```diff
@@ -0,0 +1,340 @@
...
+                }
+
+            }
+        };
+    }
+
+    private static TelemetryTop.TelemetryStream buildJtiMessage(String ipAddress, String ifName, long ifInOctets,
+            long ifOutOctets) {
+        final Port.GPort port = Port.GPort.newBuilder()
+                .addInterfaceStats(
```
j-white: The indentation in this block make this unreadable. Compare this to https://github.com/OpenNMS/opennms/blob/f1d252b5bc0b9a6bcf58ed9dbaa89bc3ec09aa4a/features/telemetry/adapters/jti/src/main/java/org/opennms/netmgt/telemetry/adapters/jti/JtiClient.java#L54.

PR #1712, opened 2017-10-18
NMS-7967: Update blueprint namespaces and schemaLocations

soleger:
Clean up the OSGi blueprint namespaces and schemaLocations by upgrading them to the versions from OSGi 6 and Karaf 4.1. This cleans up some remnants from NMS-7967 and should also make schema validation faster since it won't have to follow a HTTP redirect to get to the OSGi 6 schema.

https://issues.opennms.org/browse/NMS-7967


indigo423:
These undocumented changes will be untrackable in case they introduce side effects. Just to ensure if this is really intended?

PR #1713, opened 2017-10-18
Clean up osgi.bnd files

soleger:
This PR removes unused `osgi.bnd` files and for projects that have a `maven-bundle-plugin` configuration, it moves properties from `osgi.bnd` into the `pom.xml` (which is a more intuitive place to look for the configuration properties).

I `diff`ed all of the `MANIFEST.MF` files in the projects and the output is almost identical. The only difference is that the `Vaadin-Package-Version` attributes were being ignored when placed inside the `osgi.bnd` file, so by moving them to the POM, they're actually making it into the metadata now.


indigo423:
These undocumented changes will be untrackable in case they introduce side effects. Just to ensure if this is really intended?

PR #1718, opened 2017-10-19
 NMS-9673: Fixed XSS in userGroupView

fooker:
* JIRA: http://issues.opennms.org/browse/NMS-9645

This needs https://github.com/OpenNMS/opennms/pull/1717 in before.

### DIFF
--- a/opennms-webapp/src/main/webapp/admin/userGroupView/roles/view.jsp
+++ b/opennms-webapp/src/main/webapp/admin/userGroupView/roles/view.jsp
@@ -149,7 +149,7 @@
 
     <tr>
       <th>Description</th>
-  	<td colspan="3"><%= request.getAttribute("role") == null ? "" : WebSecurityUtils.sanitizeString(((WebRole) request.getAttribute("role")).getDescription()) %></td>
+  	<td colspan="3"><%= WebSecurityUtils.sanitizeString(((WebRole) request.getAttribute("role")).getDescription()) %></td>
     </tr>
   </table>
 </div> <!-- panel -->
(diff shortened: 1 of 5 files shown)


j-white:
@fooker can you rebase this against `foundation-2016`? https://github.com/OpenNMS/opennms/pull/1717 was merged but the commits still show up here.

fooker:
Rebased

opennms-webapp/src/main/webapp/admin/userGroupView/roles/view.jsp
```diff
@@ -148,14 +149,14 @@
 
     <tr>
       <th>Description</th>
-  	<td colspan="3">${role.description}</td>
+  	<td colspan="3"><%= request.getAttribute("role") == null ? "" : WebSecurityUtils.sanitizeString(((WebRole) request.getAttribute("role")).getDescription()) %></td>
```
j-white: Is there any reason why this one contains a null check, but the others on this page don't?

RangerRick:
I imagine it's impossible to not have a role, but guarding against nulls doesn't hurt either.  One way or the other, can we merge this before Thursday? :)

j-white:
I think the `WebSecurityUtils.sanitizeString` calls can be simplified by using `<c:out>` tags instead. If we don't get around to this before Thursday, I'd say we merge it as it though.

j-white:
Closing in favor of https://github.com/OpenNMS/opennms/pull/1763, using tags makes this more manageable and avoid having to case the types.

PR #1719, opened 2017-10-19
NMS-9691: Implement the Sink pattern using AWS SQS

agalue:
This requires AWS Credentials when used outside a VPC.

The JUnit test requires AWS Credentials, so it is disabled until the AwsUtils class is re-worked to allow mocks for tests (I'm open for options here).

* JIRA: http://issues.opennms.org/browse/NMS-9691


container/features/src/main/resources/features.xml
```diff
@@ -551,6 +551,15 @@
       <bundle dependency="true">mvn:org.apache.servicemix.bundles/org.apache.servicemix.bundles.kafka-clients/0.10.1.1_1</bundle>
     </feature>
 
+    <feature name='opennms-aws-sqs' description="OpenNMS :: Features :: AWS SQS" version="${project.version}">
+      <bundle dependency="true">wrap:mvn:com.amazonaws/aws-java-sdk-core/1.11.209$Bundle-Version=2.2.0&amp;Export-Package=*;-noimport:=true;version="1.11.209"</bundle>
+      <bundle dependency="true">wrap:mvn:com.amazonaws/aws-java-sdk-sqs/1.11.209$Bundle-Version=2.2.0&amp;Export-Package=*;-noimport:=true;version="1.11.209"</bundle>
+      <bundle dependency="true">wrap:mvn:com.amazonaws/amazon-sqs-java-messaging-lib/1.0.4$Bundle-Version=2.2.0&amp;Export-Package=*;-noimport:=true;version="1.0.4"</bundle>
+      <bundle dependency="true">mvn:com.fasterxml.jackson.core/jackson-databind/2.6.6</bundle>
```
j-white: Can we reference the jackson2Version propery there instead?

container/features/src/main/resources/features.xml
```diff
@@ -551,6 +551,15 @@
       <bundle dependency="true">mvn:org.apache.servicemix.bundles/org.apache.servicemix.bundles.kafka-clients/0.10.1.1_1</bundle>
     </feature>
 
+    <feature name='opennms-aws-sqs' description="OpenNMS :: Features :: AWS SQS" version="${project.version}">
+      <bundle dependency="true">wrap:mvn:com.amazonaws/aws-java-sdk-core/1.11.209$Bundle-Version=2.2.0&amp;Export-Package=*;-noimport:=true;version="1.11.209"</bundle>
```
j-white: Can we make the aws-java-sdk version a property in the root pom.xml instead of referencing the specific version here?

core/ipc/sink/aws-sqs-impl/pom.xml
```diff
@@ -0,0 +1,230 @@
...
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <extensions>true</extensions>
+        <configuration>
+          <instructions>
+            <Bundle-RequiredExecutionEnvironment>JavaSE-1.8</Bundle-RequiredExecutionEnvironment>
+            <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+            <Bundle-Version>${project.version}</Bundle-Version>
+            <Import-Package>
```
j-white: This statement is not necessary if we import *.

core/ipc/sink/aws-sqs-impl/pom.xml
```diff
@@ -0,0 +1,230 @@
...
+    </dependency>
+    <dependency>
+      <!-- This is required for the JmsQueueNameFactory -->
+      <groupId>org.opennms.core</groupId>
+      <artifactId>org.opennms.core.camel</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.amazonaws</groupId>
+      <artifactId>aws-java-sdk-core</artifactId>
+      <version>1.11.209</version>
```
j-white: Again, let's reference a version property instead.

core/ipc/sink/aws-sqs-impl/src/main/java/org/opennms/core/ipc/sink/aws/sqs/AwsConditionalSinkContext.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * The Class AwsConditionalSinkContext.
+ * 
+ * @author <a href="mailto:agalue@opennms.org">Alejandro Galue</a>
+ */
+@Configuration
+@Conditional(AwsConditionalSinkContext.Condition.class)
+@ImportResource("/META-INF/opennms/applicationContext-ipc-sink-server-aws.xml")
+public class AwsConditionalSinkContext {
```
j-white: Can we prefix these classes with "Sqs" instead of "Aws"?

core/ipc/sink/aws-sqs-impl/src/main/java/org/opennms/core/ipc/sink/aws/sqs/AwsMessageConsumerManager.java
```diff
@@ -0,0 +1,214 @@
...
+                            dispatch(module, module.unmarshal(txtMessage.getText()));
+                        } catch (RuntimeException e) {
+                            LOG.warn("Unexpected exception while dispatching message", e);
+                        }
+                    } else {
+                        LOG.warn("Received an unsupported message {}", message.getJMSMessageID());
+                    }
+                }
+            } catch (JMSException ex) {
+                LOG.error("Can't receive message", ex);
```
j-white: How do we recover from these errors? If the consumer fails for some reason, possibly a network interruption, we likely don't want to just exit the thread and stop consuming altogether.

core/ipc/sink/aws-sqs-impl/src/main/java/org/opennms/core/ipc/sink/aws/sqs/AwsRemoteMessageDispatcherFactory.java
```diff
@@ -0,0 +1,199 @@
...
+    /* (non-Javadoc)
+     * @see org.opennms.core.ipc.sink.common.AbstractMessageDispatcherFactory#dispatch(org.opennms.core.ipc.sink.api.SinkModule, java.lang.Object, org.opennms.core.ipc.sink.api.Message)
+     */
+    @Override
+    public <S extends Message, T extends Message> void dispatch(SinkModule<S, T> module, String topic, T message) {
+        try (MDCCloseable mdc = Logging.withPrefixCloseable(MessageConsumerManager.LOG_PREFIX)) {
+            LOG.trace("dispatch({}): sending message {}", topic, message);
+
+            try {
+                TextMessage msg = session.createTextMessage(module.marshal((T)message));
```
j-white: This method should be thread safe. I don't think that either the `session` or the `producer` can be safely used among threads.

core/ipc/sink/aws-sqs-impl/src/main/java/org/opennms/core/ipc/sink/aws/sqs/AwsRemoteMessageDispatcherFactory.java
```diff
@@ -0,0 +1,199 @@
...
+                getProducer(getModuleMetadata(module)).send(msg);
+            } catch (JMSException e) {
+                LOG.error("Error occured while sending message to topic {}.", topic, e);
```
j-white: If the Minion has no internet connectivity, I suspect we'll end up hitting this block and dropping all of the messages.

agalue:
@j-white Thanks a lot for all your comments. As soon as I have time, I'll apply all your suggestions. Keep in mind that this was a POC, as it was you who suggested to have something working quickly, and then polish it, so for sure there is a lot work to do as I never expected this to be ready to merge.

container/features/src/main/resources/features.xml
agalue: Absolutely! Interestingly, the Kafka version doesn't have it, which is what I've used as a reference. To be consistent, someone should fix Kafka, I think.

core/ipc/sink/aws-sqs-impl/src/main/java/org/opennms/core/ipc/sink/aws/sqs/AwsConditionalSinkContext.java
agalue: Just "Sqs", or maybe "AwsSqs" ? I mean, if just "Aws" doesn't look right, just "Sqs" is meaningless as well.

core/ipc/sink/aws-sqs-impl/src/main/java/org/opennms/core/ipc/sink/aws/sqs/AwsMessageConsumerManager.java
agalue: I understand. The reason why I've added the "try" at that level is because `connection.start();` can throw a JMSException, which can be for example due to bad credentials, on those extremely rare occasions where SQS is not available on the chosen region. What do you suggest to do there ? Maybe just do the try/catch for `connection.start`, and if it goes well, enter the receive loop, if not leave the thread ?

Here is what the documentation says about the `connect` method:

```
You can call the start method on the connection at any time.
However, the consumer does not begin to receive messages until you call it.
```

core/ipc/sink/aws-sqs-impl/src/main/java/org/opennms/core/ipc/sink/aws/sqs/AwsRemoteMessageDispatcherFactory.java
agalue: You're right, the Session is not thread safe. I'll rework that part.
I'm wondering, where is the documentation that says that everything used inside the `dispatch` has to be thread-safe ? Are there other methods with these kind of restrictions in this or another required class ?

core/ipc/sink/aws-sqs-impl/src/main/java/org/opennms/core/ipc/sink/aws/sqs/AwsRemoteMessageDispatcherFactory.java
agalue: You're right! Internet connectivity is mandatory is you want to do something with Amazon Web Services outside a VPC. That being said, how do you suggest to do to handle this fact ?

agalue:
Thinking loud about my initial approach, I think I don't need the JMS bridge for SQS, I should be able to use the AWS-SQS Client directly, as the Sink implementation fully control how to send and receive messages. Maybe that's required for RPC but not here. Thoughts ?

pbrane:
What’s the license on the client library?

David

agalue:
> On Oct 19, 2017, at 4:23 PM, David Hustace <notifications@github.com> wrote:
> 
> What’s the license on the client library?

https://github.com/aws/aws-sdk-java/blob/master/LICENSE.txt <https://github.com/aws/aws-sdk-java/blob/master/LICENSE.txt>

core/ipc/sink/aws-sqs-impl/src/main/java/org/opennms/core/ipc/sink/aws/sqs/AwsConditionalSinkContext.java
soleger: "AwsSqs" is pretty unreadable to most people, I'd prefer "AmazonSQS" for clarity.

container/features/src/main/resources/features.xml
soleger: I would strip off all of the metadata after (and including) the `$`, I don't think that any of that should normally be necessary. I see that you were copying the kafka feature but I'm not sure why it's there either. Do that for every bundle in this feature.

agalue:
All the suggestions have been applied.

I've decided to remove the SQS JMS usage promoting the pure AWS SDK for SQS, which simplifies the code a lot, and eliminate any potential thread-safe issue, as the `AmazonSQS` object is thread-safe.

agalue:
I've merged develop, recompiled everything and made a real life test against AWS, and I'm receiving heartbeats and traps from the Minion through SQS, using the new simplified implementation.

I'll be waiting for more suggestions. I'd like to hear opinions about how can I refactor `AmazonSQSUtils` in order to be able to mock the AWS-SDK calls and remove the `@Ignore` from the JUnit tests. As I've mentioned, the JUnit tests are ignored as they require physical access to AWS and also require valid credentials to execute them.

soleger:
All of the current updates look great, nice work!! Let's use the Minion chat channel to talk about the JUnit tests.

core/ipc/sink/aws-sqs-impl/src/main/java/org/opennms/core/ipc/sink/aws/sqs/AmazonSQSMessageConsumerManager.java
```diff
@@ -0,0 +1,192 @@
...
+
+        /* (non-Javadoc)
+         * @see java.lang.Runnable#run()
+         */
+        @Override
+        public void run() {
+            Logging.putPrefix(MessageConsumerManager.LOG_PREFIX);
+            while (!closed.get()) {
+                final String queueUrl = sqs.getQueueUrl(AmazonSQSUtils.getQueueName(module)).getQueueUrl();
+                sqs.receiveMessage(queueUrl).getMessages().forEach(m -> {
```
j-white: Can we add a comment about the blocking behavior of `receiveMessage` here? From what I can gather it should block for `ReceiveMessageWaitTimeSeconds` and return an empty list if no messages were available.

core/ipc/sink/aws-sqs-impl/src/main/java/org/opennms/core/ipc/sink/aws/sqs/AmazonSQSRemoteMessageDispatcherFactory.java
```diff
@@ -0,0 +1,183 @@
...
+            if (properties != null) {
+                final Enumeration<String> keys = properties.keys();
+                while (keys.hasMoreElements()) {
+                    final String key = keys.nextElement();
+                    awsConfig.put(key, properties.get(key));
+                }
+            }
+
+            LOG.info("AwsRemoteMessageDispatcherFactory: initializing the AWS SQS producer with: {}", awsConfig);
+            final ClassLoader currentClassLoader = Thread.currentThread().getContextClassLoader();
```
j-white: I believe this was a class-loader hack that we needed for Kafka, it may not be necessary here.

core/ipc/sink/aws-sqs-impl/src/main/java/org/opennms/core/ipc/sink/aws/sqs/AmazonSQSRemoteMessageDispatcherFactory.java
```diff
@@ -0,0 +1,183 @@
...
+    @Override
+    public <S extends Message, T extends Message> void dispatch(SinkModule<S, T> module, String topic, T message) {
+        try (MDCCloseable mdc = Logging.withPrefixCloseable(MessageConsumerManager.LOG_PREFIX)) {
+            LOG.trace("dispatch({}): sending message {}", topic, message);
+            final String queueName = getModuleMetadata(module);
+            final String queueUrl = getQueueUrl(queueName);
+            if (queueUrl == null) {
+                LOG.warn("Cannot obtain URL for queue {}. The message cannot be sent.", queueName);
+            } else {
+                sqs.sendMessage(queueUrl, module.marshal((T)message));
```
j-white: How does this behave if SQS is not reachable?

agalue: The API doesn't say anything about it, and there is no sign that the method will thrown an exception, so maybe silently ignored ? I can put a try/catch around it, catching `Exception` maybe ?

core/ipc/sink/aws-sqs-impl/src/main/java/org/opennms/core/ipc/sink/aws/sqs/AmazonSQSRemoteMessageDispatcherFactory.java
agalue: ok

core/ipc/sink/aws-sqs-impl/src/main/java/org/opennms/core/ipc/sink/aws/sqs/AmazonSQSMessageConsumerManager.java
agalue: Will do it. The `MessageWaitTimeSeconds` is configured at queue level with my current design. I'll add that as well on the comment.

agalue:
@j-white  thanks again for all your comments.

For the JUnit tests, it is not clear to me where should I declare the bean behind the new interface. The blueprint (blueprint-ipc-client.xml) ? the spring context (applicationContext-ipc-sink-server-aws.xml) ? That's why I deferred this on my initial POC.

j-white:
Declare the bean in both: the blueprint is used on the Minion, and the application context is used in OpenNMS.

core/ipc/sink/aws-sqs-impl/src/main/java/org/opennms/core/ipc/sink/aws/sqs/AmazonSQSRemoteMessageDispatcherFactory.java
j-white: Generally speaking, the implementation of the dispatch method should block until the message is successfully delivered.

I think we need to validate how the SQS client behaves if for example the Minion were to lose internet connectivity and continue to receive SNMP traps.

agalue:
> Declare the bean in both: the blueprint is used on the Minion, and the application context is used in OpenNMS.

Done!

agalue:
> Generally speaking, the implementation of the dispatch method should block until the message is successfully delivered.

In theory, if the "result object" contains a message ID, the message is considered "delivered". I can check that, but what should I do if this is not the case ? Should I retry ? If so, how many times ?
> I think we need to validate how the SQS client behaves if for example the Minion were to lose internet connectivity and continue to receive SNMP traps.
> 
How do you suggest we do this ?

Remember that internally, using any method of the AmazonSQS Object is translated into an HTTPS Query against the AWS API. SQS is NOT like ActiveMQ.

j-white:
> How do you suggest we do this ?

Start the Minion, disable the Internet access, send a few traps, see what happens.

agalue:
> Start the Minion, disable the Internet access, send a few traps, see what happens.

I can certainly do it, but that is not what I meant with my question.

As I said, we’re talking about HTTPS calls. So, if there’s no internet, you’ll get the same as trying to reach Google from a browser on a computer with no internet connection:

http://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessage.html

We should provide a way to “hold” messages that has not been sent, and try again later when AWS is reachable, as loosing traps is unacceptable for the use case we’re trying to solve with SQS.

So, what would be the best approach? An internal queue? A buffer? Something else ? And then, some intelligence to retry sending the messages again ?

j-white:
> In theory, if the "result object" contains a message ID, the message is considered "delivered". I can check that, but what should I do if this is not the case ? Should I retry ? If so, how many times ?

I would block and keep retrying indefinitely (or until interrupted) with a small retry interval, say every 15 seconds.

Also we need to make sure that we're only retrying if it's a connectivity problem. If it's a problem with the message for example, say it's too large, then there's no point in blocking up a thread and continuing to retry.

agalue:
> I would block and keep retrying indefinitely (or until interrupted) with a small retry interval, say every 15 seconds.

Sounds good, but what happens if there are new incoming traps when we have this blocked code running ?
> Also we need to make sure that we're only retrying if it's a connectivity problem. If it's a problem with the message for example, say it's too large, then there's no point in blocking up a thread and continuing to retry.
> 
Makes sense, and fortunately there are ways to do it exactly as you’ve mentioned.

j-white:
> Sounds good, but what happens if there are new incoming traps when we have this blocked code running ?

If all of the dispatcher threads are blocked, the traps will queue up in memory until the [configured](https://github.com/OpenNMS/opennms/blob/d1d9c0f3bd7e6561e0f2e8d63942629da5c14dc6/opennms-config-jaxb/src/main/resources/xsds/trapd-configuration.xsd#L85) limit is reached. This is managed by the Sink API.

agalue:
> If all of the dispatcher threads are blocked, the traps will queue up in memory until the configured <https://github.com/OpenNMS/opennms/blob/d1d9c0f3bd7e6561e0f2e8d63942629da5c14dc6/opennms-config-jaxb/src/main/resources/xsds/trapd-configuration.xsd#L85> limit is reached. This is managed by the Sink API.

Excellent, thanks!

agalue:
The first attempt to implement the infinite loop until access is regained or another exception is thrown is in place. I'll continue the tests tomorrow.

agalue:
Is there anything else that requires my attention here ?

core/ipc/sink/aws-sqs-impl/src/main/java/org/opennms/core/ipc/sink/aws/sqs/AmazonSQSManager.java
```diff
@@ -0,0 +1,123 @@
...
+    public static final String AWS_SECRET_ACCESS_KEY = "aws_secret_access_key";
+
+    /** The Constant AWS_QUEUE_NAME_PREFIX. */
+    public static final String AWS_QUEUE_NAME_PREFIX = "aws_queue_name_prefix";
+
+    /** The Constant AWS_USE_FIFO_QUEUE. */
+    public static final String AWS_USE_FIFO_QUEUE = "aws_use_fifo_queue";
+
+    /** The Constant QUEUE_ATTRIBUTES. */
+    public static final Map<String, String> QUEUE_ATTRIBUTES = new HashMap<>();
```
j-white: It seems odd to have the map declaration in the interface and the initialization in the implementation.

I'd opt for moving the declaration into the implementation, or using a static initializer like: https://google.github.io/guava/releases/21.0/api/docs/com/google/common/collect/ImmutableMap.Builder.html

core/ipc/sink/common/src/main/java/org/opennms/core/ipc/sink/common/SinkStrategy.java
```diff
@@ -36,9 +36,12 @@
 
     private static final String KAFKA_SINK_STRATEGY_NAME = "kafka";
 
+    private static final String AWS_SINK_STRATEGY_NAME = "aws";
+
     public static enum Strategy {
         CAMEL(CAMEL_SINK_STRATEGY_NAME, "JMS implementation using Camel"),
-        KAFKA(KAFKA_SINK_STRATEGY_NAME, "Kafka implementation using the Kafka consumer/producer APIs");
+        KAFKA(KAFKA_SINK_STRATEGY_NAME, "Kafka implementation using the Kafka consumer/producer APIs"),
+        AWS(AWS_SINK_STRATEGY_NAME, "AWS SQS implementation using the AWS JMS consumer/producer APIs");
```
j-white: Can we rename this enum and the related constants to `SQS` rather than `AWS`? We can also remove the reference to JMS from the description.

agalue: No problem! should I also rename the actual value (currently, "aws") ? I'm asking as that also means I'd have to tweak the documentation.

agalue: Speaking about docs changes, what about `AmazonSQSSinkConstants.AWS_CONFIG_PID` ? currently it defaults to "org.opennms.core.ipc.sink.aws". Should I rename it to "org.opennms.core.ipc.sink.sqs" ?

j-white: Yeah, I'd rename the "aws" value too.

As for the pid, I'd opt for `org.opennms.core.ipc.sink.sqs` or `org.opennms.core.ipc.sink.awq.sqs`.

agalue: ok!

core/ipc/sink/aws-sqs-impl/src/main/java/org/opennms/core/ipc/sink/aws/sqs/AmazonSQSManager.java
agalue: Sounds good! should I update the AWS feature and/or POMs to reflect the dependency against Guava ?

core/ipc/sink/common/src/main/java/org/opennms/core/ipc/sink/common/SinkStrategy.java
agalue: Final question, the feature itself to be installed on Minion is called `opennms-core-ipc-sink-aws`, should I also rename that to be `opennms-core-ipc-sink-sqs` ?

agalue: Disregard the last part, it is called `opennms-core-ipc-sink-aws-sqs`, so we're good.

agalue:
All suggestions have been applied, and I verified on a life environment that the solution works against AWS.

PR #1723, opened 2017-10-23
NMS-9524: Northbounders implementation are not sending feedback events for reloadDaemonConfig.

agalue:
Northbounders implementation are not sending feedback events for reloadDaemonConfig.

* JIRA: http://issues.opennms.org/browse/NMS-9524

I've created the PR against `foundation-2017` due to the Drools NBI and BSF NBI. For `foundation-2016` I can cherry-pick the solution and manually remove the changes on the NBIs that don't exist there, which is easier than creating 2 separate PRs, one for each foundation branch.

### DIFF
--- a/opennms-alarms/api/src/main/java/org/opennms/netmgt/alarmd/api/Northbounder.java
+++ b/opennms-alarms/api/src/main/java/org/opennms/netmgt/alarmd/api/Northbounder.java
@@ -67,6 +67,6 @@ public interface Northbounder {
     /**
      * Reloads configuration.
      */
-    public void reloadConfig();
+    public void reloadConfig() throws NorthbounderException;
 
 }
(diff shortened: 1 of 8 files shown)


j-white:
Looks good. I added this manually to Bamboo, not sure why it wasn't picked up.

Generally speaking I find it easier to make changes to the oldest foundation branches first and them merge them forward rather than starting with the newest and then cherry picking the changes back.

agalue:
> Generally speaking I find it easier to make changes to the oldest foundation branches first and them merge them forward rather than starting with the newest and then cherry picking the changes back.

I agree with you, but in this particular case, it will be faster this way, as newer versions contain classes that don’t exist on older versions, which will ends up on 2 PR about the same size.

agalue:
I can do it the other way around, if that accelerates the review and merging process. Just let me know what do you prefer.

PR #1724, opened 2017-10-24
HZN-1170: Implement Elastic Flow Repository

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1170

When testing this, please install the feature via the karaf console first: `feature:install flows`
An elastic instance should be running as well. Default setting is `localhost:9200`. 

I will create appropriate follow up issues to tackle the open TODOs.


features/flows/api/src/main/java/org/opennms/netmgt/flows/api/FlowType.java
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * For more information contact:
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.netmgt.flows.api;
+
+public interface FlowType {
```
j-white: Should this be an enum instead?

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/ClientFactory.java
```diff
@@ -0,0 +1,144 @@
...
+
+        // Ensure urls is set
+        if (urls.isEmpty()) {
+            throw new IllegalArgumentException("No urls have been provided");
+        }
+
+        // TODO MVR Fix the following hack:
+        // This is a hack, due to our osgi-jetty-bridge, because the annotation at the NetflowDocument
+        // living in the api module (on the jetty side) are not accessible directly (only as proxy) by the Gson lirary
+        // living in the osgi container (osgi-jest-complete module) and therefore @SerializedName annotations won't work.
```
j-white: If the `@SerializedName` annotations don't work, then why do we need the Gson dependency in the flow api module?

smoke-test/src/test/java/org/opennms/smoketest/flow/FlowStackIT.java
```diff
@@ -0,0 +1,217 @@
...
+                JestResult result = jestClient.execute(new GetTemplate.Builder(TEMPLATE_NAME).build());
+                if (result.isSucceeded() && result.getJsonObject().get(TEMPLATE_NAME) != null) {
+                    return true;
+                }
+                return false;
+            });
+            // Verify that the flows have been created
+            verify(client, jestClient -> {
+                SearchResult response = jestClient.execute(new Search.Builder("").addIndex("flow-*").build());
+                if (response.isSucceeded() && response.getTotal() == 2) {
```
j-white: Nice. I'm assuming the packet contains two flows?

opennms-services/src/main/java/org/opennms/netmgt/notifd/ServiceRegistryExecutor.java
```diff
@@ -28,25 +28,28 @@
...
-import java.lang.management.ManagementFactory;
-import java.util.List;
-import java.util.concurrent.TimeUnit;
-
 public class ServiceRegistryExecutor implements ExecutorStrategy {
     private static final Logger LOG = LoggerFactory.getLogger(ServiceRegistryExecutor.class);
 
-    private static final DefaultServiceRegistry s_registry = DefaultServiceRegistry.INSTANCE;
-
     private static final String GRACE_PERIOD_MS_SYS_PROP = "org.opennms.netmgt.notifd.notificationStrategyGracePeriodMs";
```
j-white: Can we move these properties and default into the service lookup classes? I think we'd be better off having a global properties instead of having to tune these for every interface.

I'm thinking something along the lines of  `org.opennms.core.soa.lookup.gracePeriodMs` and `org.opennms.core.soa.lookup.lookupDelayMs`.

features/telemetry/adapters/netflow/src/main/java/org/opennms/netmgt/telemetry/adapters/netflow/Netflow5Adapter.java
```diff
@@ -53,27 +61,86 @@ public void handleMessageLog(TelemetryMessageLogDTO messageLog) {
...
+    private List<NetflowDocument> convert(NetflowPacket netflowPacket, TelemetryMessageLogDTO messageLog) {
+        return netflowPacket.getRecords().stream()
+                .map(record -> {
+                    final NetflowDocument document = new NetflowDocument();
+
+                    // META
+                    document.setExporterAddress(InetAddressUtils.toIpAddrString(messageLog.getSourceAddress()));
+                    document.setLocation(messageLog.getLocation());
+                    document.setFlowType(FlowType.NETFLOW_5);
+                    document.setTimestamp(netflowPacket.getUnixSecs() * 1000L + netflowPacket.getUnixNSecs() / 1000L / 1000L);
```
j-white: Can we also include some basic meta data related to the node? Perhaps we can include the Node ID, FS/FID if available and the list categories associated with the node.

We could also address this in a separate issue.

opennms-services/src/main/java/org/opennms/netmgt/notifd/ServiceRegistryExecutor.java
```diff
@@ -28,25 +28,28 @@
...
     private static final int GRACE_PERIOD_MS = Integer.getInteger(GRACE_PERIOD_MS_SYS_PROP, 3*60*1000);
     private static final int LOOKUP_DELAY_MS = 5*1000;
 
+    private static final ServiceLookup SERVICE_LOOKUP = new ServiceLookupBuilder(DefaultServiceRegistry.INSTANCE)
+            .blocking(GRACE_PERIOD_MS, LOOKUP_DELAY_MS, () -> ManagementFactory.getRuntimeMXBean().getUptime())
```
j-white: Can we make `ManagementFactory.getRuntimeMXBean().getUptime()` the default uptime provider instead of having to pass this each time?

With the properties moved into the class, the a default uptime provider, most of the calls could just end up looking like:
```
new ServiceLookupBuilder(DefaultServiceRegistry.INSTANCE)
  .blocking()
  .build();
```

features/flows/api/src/main/java/org/opennms/netmgt/flows/api/FlowType.java
mvrueden: Not necessarily. At the moment it is just a String. If we need more information an enum or a type might be required. I prefer this constant variant for now.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/ClientFactory.java
mvrueden: The idea was to get gson to work, in order to define custom mappings. If that is not going to work in the end, we do not need this dependency, true.

features/telemetry/adapters/netflow/src/main/java/org/opennms/netmgt/telemetry/adapters/netflow/Netflow5Adapter.java
mvrueden: Yes, the idea is to enrich each flow with such data. This however was left out on purpose in this issue.

opennms-services/src/main/java/org/opennms/netmgt/notifd/ServiceRegistryExecutor.java
mvrueden: Yep, makes sense.

smoke-test/src/test/java/org/opennms/smoketest/flow/FlowStackIT.java
mvrueden: Yes

opennms-services/src/main/java/org/opennms/netmgt/notifd/ServiceRegistryExecutor.java
mvrueden: Yep, makes also sense.

j-white:
I gave this a spin and it worked well: flow data sent to a Minion was successfully indexed in ES.

PR #1726, opened 2017-10-25
HZN-1171: Implement Flow ReST API to query for flows

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1171


j-white:
Looks good :+1:

PR #1728, opened 2017-10-25
HZN-1163: Use protobuf messages for the Telemetryd Sink

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1163

Here we convert telemetryd to use protobuf based messages via the Sink API instead of XML. This makes the messages smaller, and cheaper to encode/decode.

We have also updated the Sink API to use byte arrays instead of strings and refactored part of telemetryd to avoid having the adapters depend on the protobuf implementation.


pbrane:
> On Oct 25, 2017, at 15:23, Jesse White <notifications@github.com> wrote:
> 
> We have also updated the Sink API to use byte arrays instead of strings

Nice!

PR #1729, opened 2017-10-26
NMS-9704: Manually refresh all theme related bundles

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-9704

It seems to be an issue with Karaf 4.1.2. 
Even in a pure container having only the bundles started, they are not resolved fully after a restart.
However the behaviour is a bit different than the OpenNMS' Karaf Container. In OpenNMS a refresh worked reliably, while only a re-installation of the bundle worked in a plain container.

This solution is not ideal, but gets the issue resolved and allows us to investigate further and may provide a better solution.


j-white:
Nice hack. I think we may also be able to solve this cleaning Karaf's cache on startup by calling `System.setProperty("karaf.clean.cache", "true")` before the container is started.

mvrueden:
Yeah, but that has also some other side-effects, like clearing the history and so on in the console, doesn't it?. If we use the system property we must ensure, that the topology map history is still persisted and does not get cleaned.

j-white:
I've confirmed that `karaf.clean.cache` does work, but it does end up clearing the topology map history.

I think this solution is probably the simplest for the time being, pending the proper fix in Karaf.

pbrane:
> On Nov 1, 2017, at 12:45, Jesse White <notifications@github.com> wrote:
> 
> but it does end up clearing the topology map history.

What about saved layouts? Does it delete those, also?

j-white:
Saved layouts don't get lost, those are stored in the database.

j-white:
The refreshed bundle appears to be breaking the smoke tests. I suspect that the refresh occurs while we've already starting browsing through the WebUI, and then things go missing while they are reloaded.

features/vaadin-components/bundle-refresher/src/main/java/org/opennms/features/vaadin/components/bundlerefresher/Activator.java
```diff
@@ -0,0 +1,58 @@
...
+import org.osgi.framework.BundleActivator;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.wiring.FrameworkWiring;
+
+
+public class Activator implements BundleActivator {
+
+    @Override
+    public void start(BundleContext context) throws Exception {
+        final Set<Bundle> bundlesToRefresh = Arrays.stream(context.getBundles()).filter(b -> b.getSymbolicName().contains("theme")).collect(Collectors.toSet());
```
j-white: We might be able to fix the smoke tests if we filter for resolved bundles and avoid refreshing these if they are already resolved.

mvrueden:
The bundle-refresher bundle was not in the system directory and therefore could not be installed, which resulted in the container not beeing able to start correctly, which means some (or all) bundles were missing which had of course some affects on the UI, which let the smoke tests fail.

Should pass now.

features/vaadin-components/bundle-refresher/src/main/java/org/opennms/features/vaadin/components/bundlerefresher/Activator.java
```diff
@@ -0,0 +1,82 @@
...
+import org.osgi.framework.BundleActivator;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.wiring.FrameworkWiring;
+
+
+public class Activator implements BundleActivator {
+
+    @Override
+    public void start(BundleContext context) throws Exception {
+        // Determine if any vaadin theme fragments are unresoled
```
indigo423: unresoled vs. unresolved

PR #1731, opened 2017-10-27
NMS-9718: Unexpected errors while generating event definitions from TRAP-TYPE without a DESCRIPTION field

agalue:
Unexpected errors while generating event definitions from TRAP-TYPE without a DESCRIPTION field

* JIRA: http://issues.opennms.org/browse/NMS-9718

### DIFF
--- a/features/mib-compiler/src/main/java/org/opennms/features/mibcompiler/services/JsmiMibParser.java
+++ b/features/mib-compiler/src/main/java/org/opennms/features/mibcompiler/services/JsmiMibParser.java
@@ -157,7 +157,7 @@ public class JsmiMibParser implements MibParser, Serializable {
             try {
                 mib = parser.parse();
             } catch (Exception e) {
-                LOG.error("Can't compile " + mibFile, e);
+                LOG.error("Can't compile {}", mibFile, e);
                 errorHandler.addError(e.getMessage());
                 return false;
             }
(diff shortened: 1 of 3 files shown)


features/mib-compiler/src/main/java/org/opennms/features/mibcompiler/services/JsmiMibParser.java
```diff
@@ -157,7 +157,7 @@ public boolean parseMib(File mibFile) {
             try {
                 mib = parser.parse();
             } catch (Exception e) {
-                LOG.error("Can't compile {}", mibFile, e);
+                LOG.error("Can't compile " + mibFile, e);
```
j-white: Why change these? Both of these end up printing the same message.

```
Exception e = new Exception();
String param = "p1";
LOG.error("error: {}", param, e);
LOG.error("error: " + param, e);
```

```
2017-10-30 09:54:19,207 ERROR [main] org.opennms.netmgt.snmp.LogTest - error: p1
java.lang.Exception
	at org.opennms.netmgt.snmp.LogTest.doIt(LogTest.java:40)
	...
2017-10-30 09:54:19,213 ERROR [main] org.opennms.netmgt.snmp.LogTest - error: p1
java.lang.Exception
	at org.opennms.netmgt.snmp.LogTest.doIt(LogTest.java:40)
	...
```

agalue: I was unsure that if I add placeholders, and then the throwable at the end, I can have the full exception trace besides my custom message (based on the docs). So, are you saying that will happen ? If so, I can rollback those particular changes.

j-white: https://www.slf4j.org/faq.html#paramException

agalue:
Done! let me know if there is something else I should change.

PR #1732, opened 2017-10-30
NMS-9709: clean up duty schedule parsing

RangerRick:
This PR makes duty schedule (and other group) parsing more forgiving, trimming strings down.

* JIRA: http://issues.opennms.org/browse/NMS-9709

### DIFF
--- a/core/xml/src/main/java/org/opennms/core/xml/StringListTrimAdapter.java
+++ b/core/xml/src/main/java/org/opennms/core/xml/StringListTrimAdapter.java
@@ -1,0 +1,60 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2017 The OpenNMS Group, Inc.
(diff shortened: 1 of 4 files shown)


core/xml/src/main/java/org/opennms/core/xml/StringListTrimAdapter.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.stream.Collectors;
+
+import javax.xml.bind.annotation.adapters.XmlAdapter;
+
+public class StringListTrimAdapter extends XmlAdapter<String[],List<String>> {
+    private static final String[] TYPE = new String[0];
+
+    @Override
+    public String[] marshal(final List<String> value) throws Exception {
+        System.err.println("unmarshal: " + value);
```
j-white: `System.err` -> `LOG.trace`?

RangerRick:
OK, the `System.err` logs should be gone now, and I've fixed this to also include cleaning up duty schedules on the `users.xml` also.

PR #1734, opened 2017-10-30
NMS-9620: Ensure that the snmp4j session is closed

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9620

This fixes a snmp4j session leak caused by an exception thrown during the call to send().


RangerRick:
Do we know for certain we can never get any `RuntimeException`s inside the `onResponse`?  Is it worth changing that too?

j-white:
Not that I'm aware of, but it we do, the session will still be closed in the `finally` clause.

j-white:
Right. We should probably change that just to be safe :)

If we don't close the future there, the caller will be left dangling.

PR #1738, opened 2017-11-01
NMS-9722: fix NPE in string handling

RangerRick:
Since the newer JAXB code cleans out empty strings properly, the editor was NPE'ing trying to call `.trim()` on it.  This PR fixes it.

* JIRA: http://issues.opennms.org/browse/NMS-9722


j-white:
Pretty straightforward.

PR #1739, opened 2017-11-02
NMS-9725: improve validation of requisitions

RangerRick:
This PR adds validation for sub-objects of the requisition, and normalizes the way validation is performed in the ReST interface.

* JIRA: http://issues.opennms.org/browse/NMS-9725


agalue:
I was able to verify that the API returns HTTP 400 when there are missing required fields, and ignore invalid fields (i.e. fields that don't exist on the XSD). As soon as this goes green, I'll merge it

RangerRick:
Yeah, looks like some tests are failing that used to pass, but it might actually because of problems with the requisitions.  :D  I'll look into it this morning.

RangerRick:
Turns out the test had wrong data for the last... oh... 11 years.  ;)  Everything else is good, I'm gonna go ahead and merge this.  Thanks for looking at it.

PR #1740, opened 2017-11-02
NMS-9723: Refactoring in preparation for SQS RPC implementation

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9723

Here we refactor the AmazonSQSManager and related configuration of the Sink implementation and into a separate module where it can be shared and leveraged by the RPC implementation.

This is a partial solution for NMS-9723 - I'm going to break it out into several PRs to make it easier to review.


agalue:
Verified against AWS. Both types of queues (standard and FIFO) are working for the Sink pattern (for heartbeats and SNMP traps).

PR #1741, opened 2017-11-06
NMS-9723: Add support for using AWS SQS for RPCs.

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9723

Here we generalize the Camel RPC implementation, allowing the routes to be controlled by different strategies, and add support for using AWS SQS instead of ActiveMQ.


agalue:
It looks very good!
Unfortunately, I haven't had time to test it (I hope to have time next week).

The main reason of this comment is that I think we should add a section on the documentation about the size and characters used on the location names.

That's because for standard queues, the queue name cannot be greater than 85 characters (80 when using FIFO queues). Also, there are restrictions about the allowed characters (already mentioned on the docs if I'm not mistaken). So, considering the prefix for the environment that can be configurable (which is affected by this restriction), and the other forced fields on the queue names, the users should be aware of this when using SQS, as this will impose a restriction about the length and content of the location name.

j-white:
Good point. I added a section to the limitations that covers this.

Given the current queue names and formatting, we can support up to 32 characters for the prefix + location names.

agalue:
I've verified that everything is working as expected. I've installed the RPMs from the branch on an EC2 instance, using an IAM Role to provide SQS Access, and a Minion running on a VM on my machine (using the RPMs from the branch), to monitor my local environment.

I discover that on this particular deployment, it is crucial to have everything in sync in terms of time, so everything went well after I've updated the timezones and activated NTP everywhere.

PR #1743, opened 2017-11-06
HZN-1185: Add support for OSGi based Telemetryd adapters

cgorantla:
HZN-1185: Add support for OSGi based Telemetryd adapters

Add jti adapter osgi feature. 
Move OSGiScript related files to common project

* JIRA: http://issues.opennms.org/browse/hzn-1185


checkstyle/src/main/resources/license-suppressions.xml
```diff
@@ -13,6 +13,7 @@
     <suppress files="org/opennms/features/topology/plugins/topo/graphml/internal/scripting/.*" checks=".*"/>
     <suppress files="org/opennms/poller/remote/Mac.*" checks=".*"/>
     <suppress files="org/opennms/container/web/.*" checks=".*"/>
+    <suppress files="org/opennms/features/osgi/script/.*" checks=".*"/>
```
j-white: We can remove the previous entry for these since we changed the package name.

container/features/src/main/resources/features.xml
```diff
@@ -1077,4 +1077,12 @@
       <bundle>mvn:org.opennms.features.poller/org.opennms.features.poller.shell/${project.version}</bundle>
     </feature>
 
+    <feature name="opennms-telemetry-jti" description="OpenNMS :: Telemetry :: JTI" version="${project.version}">
+      <feature>opennms-collection-api</feature>
+      <feature version="${guavaVersion}">guava</feature>
+      <bundle>mvn:com.google.protobuf/protobuf-java/${protobufVersion}</bundle>
```
j-white: Do we actually depend on protobuf?

container/karaf/src/main/filtered-resources/etc/custom.properties
```diff
@@ -795,5 +795,10 @@ org.osgi.framework.system.packages.extra=org.apache.karaf.branding,\
         org.opennms.web.svclayer.model;version=${opennms.osgi.version},\
         org.opennms.web.svclayer.support;version=${opennms.osgi.version},\
         org.opennms.netmgt.icmp;version=${opennms.osgi.version},\
-        org.opennms.netmgt.icmp.proxy;version=${opennms.osgi.version}
+        org.opennms.netmgt.icmp.proxy;version=${opennms.osgi.version},\
+        org.opennms.netmgt.filter.api;version=${opennms.osgi.version},\
```
j-white: These should be in alphabetical order instead of just appending them to the end.

We could also fix up the the `org.opennms.netmgt.icmp` and `org.opennms.netmgt.icmp.proxy` entries while were at it.

features/telemetry/adapters/api/src/main/java/org/opennms/netmgt/telemetry/adapters/api/Adapter.java
```diff
@@ -38,7 +38,7 @@
  *
  * @author jwhite
  */
-public interface Adapter {
+public interface Adapter{
```
j-white: Unnecessary change.

container/features/src/main/resources/features.xml
```diff
@@ -1077,4 +1077,12 @@
...
+      <bundle>mvn:org.opennms.features.telemetry.adapters/org.opennms.features.telemetry.adapters.registry/${project.version}</bundle>
```
j-white: I don't think this feature depends on the registry either.

features/telemetry/adapters/jti/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -0,0 +1,36 @@
...
+		http://www.osgi.org/xmlns/blueprint/v1.0.0 
+		http://www.osgi.org/xmlns/blueprint/v1.0.0/blueprint.xsd
+		http://aries.apache.org/blueprint/xmlns/blueprint-cm/v1.1.0
+		http://aries.apache.org/schemas/blueprint-cm/blueprint-cm-1.1.0.xsd
+		http://aries.apache.org/blueprint/xmlns/blueprint-ext/v1.1.0
+		http://aries.apache.org/schemas/blueprint-ext/blueprint-ext-1.1.xsd
+">
+
+	<bean id="jtiFactory" class="org.opennms.netmgt.telemetry.adapters.jti.JtiAdapterFactory">
+	    <argument ref="blueprintBundleContext" />
```
j-white: Nit: This one isn't lined up with the others :)

features/telemetry/adapters/jti/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -0,0 +1,36 @@
...
+			<entry key="type" value="org.opennms.netmgt.telemetry.adapters.jti.JtiGpbAdapter" />
+		</service-properties>
+	</service>
+
+	<reference id="collectionAgentFactory" interface="org.opennms.netmgt.collection.api.CollectionAgentFactory"/>
+	<reference id="interfaceToNodeCache" interface="org.opennms.netmgt.dao.api.InterfaceToNodeCache"/>
+	<reference id="nodeDao" interface="org.opennms.netmgt.dao.api.NodeDao"/>
+	<reference id="filterDao" interface="org.opennms.netmgt.filter.api.FilterDao"/>
+	<reference id="transactionTemplate" interface="org.springframework.transaction.support.TransactionOperations"/>
+    <reference id="persisterFactory" interface="org.opennms.netmgt.collection.api.PersisterFactory"/>
```
j-white: Nit: This one either :)

features/telemetry/adapters/registry/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+      <groupId>org.opennms.features.telemetry.adapters</groupId>
+      <artifactId>org.opennms.features.telemetry.adapters.api</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.opennms.features.telemetry.adapters</groupId>
+      <artifactId>org.opennms.features.telemetry.adapters.factory</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+	<dependency>
```
j-white: Nit: White space is off here.

features/telemetry/daemon/src/main/java/org/opennms/netmgt/telemetry/daemon/TelemetryMessageConsumer.java
```diff
@@ -99,34 +106,59 @@ public void handleMessage(TelemetryProtos.TelemetryMessageLog messageLog) {
     }
 
     private Adapter buildAdapter(org.opennms.netmgt.telemetry.config.model.Adapter adapterDef) throws Exception {
```
j-white: I think we should move all the logic about constructing an adapter into the registry instead of having it here.

features/telemetry/daemon/src/main/java/org/opennms/netmgt/telemetry/daemon/TelemetryMessageConsumer.java
```diff
@@ -99,34 +106,59 @@ public void handleMessage(TelemetryProtos.TelemetryMessageLog messageLog) {
...
-        } catch (Exception e) {
-            throw new RuntimeException(String.format("Failed to instantiate adapter with class name '%s'.",
-                    adapterDef.getClassName(), e));
-        }
 
-        // Cast
-        if (!(adapterInstance instanceof Adapter)) {
-            throw new IllegalArgumentException(String.format("%s must implement %s", adapterDef.getClassName(), Adapter.class.getCanonicalName()));
+        AdapterFactory adapterFactory = null;
+        while (ManagementFactory.getRuntimeMXBean().getUptime() < GRACE_PERIOD_MS) {
```
j-white: If the daemon is reloaded after the system has been up, we won't even attempt a lookup in the registry.

container/features/src/main/resources/features.xml
cgorantla: We do depend on protobuf in jti adapter.

features/telemetry/daemon/src/main/java/org/opennms/netmgt/telemetry/daemon/TelemetryMessageConsumer.java
cgorantla: Took care of all the review comments.

container/features/src/main/resources/features.xml
j-white: Duh, of course. I was confusing this with the protobuf messages used between OpenNMS and Minion.

j-white:
Any input on this @mvrueden?  We should be able to migrate the Netflow 5 adapter in the `features/drift` branch to OSGi with these changes.

container/features/src/main/resources/features.xml
```diff
@@ -1077,4 +1077,11 @@
       <bundle>mvn:org.opennms.features.poller/org.opennms.features.poller.shell/${project.version}</bundle>
     </feature>
 
+    <feature name="opennms-telemetry-jti" description="OpenNMS :: Telemetry :: JTI" version="${project.version}">
```
mvrueden: Usually we let the `maven-features-plugin` take care of the feature generation. Is there a particular reason why the feature is defined manually?

features/telemetry/adapters/collection/src/main/java/org/opennms/netmgt/telemetry/adapters/collection/ScriptedCollectionSetBuilder.java
```diff
@@ -56,13 +59,19 @@
...
 
         final String ext = Files.getFileExtension(script.getAbsolutePath());
-        final ScriptEngineManager manager = new ScriptEngineManager();
+
+        ScriptEngineManager manager;
+        if (bundleContext != null) {
+            manager = new OSGiScriptEngineManager(bundleContext);
+        } else {
+            manager = new ScriptEngineManager();
+        }
```
mvrueden: Can we move this code to the constructor and have something like:

```java
public ScriptedcollectionSetBuilder(File script) {
    this(script, new ScriptEngineManager())
}

public ScriptedCollectionSetBuilder(File script, BundleContext context) {
    this(script, new OSGiScriptEngineManager(context);
}

private ScriptCollectionSetBuilder(File script, ScriptEngineManager manager) {
     ...
}
```

features/telemetry/adapters/registry/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+    <groupId>org.opennms.features.telemetry</groupId>
+    <artifactId>org.opennms.features.telemetry.adapters</artifactId>
+    <version>22.0.0-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>org.opennms.features.telemetry.adapters</groupId>
+  <artifactId>org.opennms.features.telemetry.adapters.registry</artifactId>
+  <name>OpenNMS :: Features :: Telemetry :: Adapters :: Registry</name>
+  <packaging>bundle</packaging>
+  <build>
```
mvrueden: The `maven-bundle-plugin` setting is duplicated a lot. We may want to define it in a parent module instead.

features/topology-map/plugins/org.opennms.features.topology.plugins.topo.graphml/pom.xml
```diff
@@ -103,6 +103,11 @@
       <artifactId>org.osgi.core</artifactId>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.opennms.features</groupId>
+      <artifactId>org.opennms.features.osgi-jsr223</artifactId>
+      <version>${project.version}</version>
```
mvrueden: scope should be provided

container/features/src/main/resources/features.xml
j-white: Personally I like to avoid using our `maven-features-plugin`: it adds unnecessary bundles, and doesn't reference other features that already exist.

mvrueden: If it is setup correctly it will not add unnecessary bundles. It only adds transitive dependencies if the scope is not set to provided. Also features are supported. Only thing I see which makes it hard, is that you cannot define a version on a feature.

cgorantla: I always had features defined here except for some of the minion features.  Unless there is specific advantage, I will keep it here.

features/telemetry/adapters/registry/pom.xml
cgorantla: parent module is adapters which just includes other modules. There may be other adapters which may not be osgi bundles. I will keep  it that way.  Also good to have bundle info in it's own pom.xml I feel.

features/topology-map/plugins/org.opennms.features.topology.plugins.topo.graphml/pom.xml
cgorantla: I updated scope here. But in case of adapters-collection module,  I have it as default scope. We need to have it work on Spring as well if feature doesn't exist, so OSGiScriptEngine related files needs to be in default classpath.

features/telemetry/adapters/registry/pom.xml
mvrueden: The definition here, only extends maven, which allows defining bundles. It does not make all modules bundles. The type is defined by the `<type>bundle</type>` statement. Therefore I am still for moving it to a parent module.

container/features/src/main/resources/features.xml
mvrueden: The specific advantage is, that the `maven-features-plugin` will add/remove a bundle to the definition every time the pom of the module changes.

PR #1744, opened 2017-11-07
NMS-9699: Enable preemptive auth for first host

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-9699

@gallenc, @jeffgdotorg, @agalue this PR enables preemptive authentication when username and password is provided. Please give it a try and let me know if it works as expected.

### DIFF
--- a/features/opennms-es-rest/main-module/src/main/java/org/opennms/plugins/elasticsearch/rest/RestClientFactory.java
+++ b/features/opennms-es-rest/main-module/src/main/java/org/opennms/plugins/elasticsearch/rest/RestClientFactory.java
@@ -140,8 +140,8 @@ public class RestClientFactory {
 			final URL targetUrl = new URL(targetHosts.get(0));
 
 			if (targetHosts.size() > 1) {
-				LOG.warn("Username/Password have been defined, but multiple target hosts were found. Each host will use the same credentials.");
-				LOG.warn("Username/Password have been defined, but multiple target hosts were found. Preemptive auth is only enabled for host {}",  targetHosts.get(0));
+				LOG.warn("Credentials have been defined, but multiple target hosts were found. " +
+						"Each host will use the same credentials. Preemptive auth is only enabled for host {}", targetHosts.get(0));
 			}
 
 			configBuilder.defaultCredentials(esusername, espassword);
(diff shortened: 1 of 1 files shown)


features/opennms-es-rest/main-module/src/main/java/org/opennms/plugins/elasticsearch/rest/RestClientFactory.java
```diff
@@ -122,13 +131,23 @@ public void setServers(Set<String> servers) {
...
 
 		// If multiple URLs are specified in a comma-separated string, split them up
-		config = new HttpClientConfig.Builder(Arrays.asList(elasticSearchURL.split(",")))
-				.multiThreaded(true)
-				.defaultCredentials(esusername, espassword)
-				.build();
+		final List<String> targetHosts = Arrays.asList(elasticSearchURL.split(","));
+		final HttpClientConfig.Builder configBuilder = new HttpClientConfig.Builder(targetHosts).multiThreaded(true);
+		if (!Strings.isNullOrEmpty(esusername) && !Strings.isNullOrEmpty(espassword)) {
+			final URL targetUrl = new URL(targetHosts.get(0));
```
jeffgdotorg: Why not set preemptive auth for all targets, using a for() loop here? Or am I missing something?
I'm thinking about the case where we are happily forwarding to the first target, but that cluster member goes away and we switch to the second target, and now performance might suffer because we're no longer doing preemptive auth now. This would be difficult to troubleshoot at best.

mvrueden: The API only supports setting it for one target. I could loop over all targets, but that would enable it only for the last target. That is just a limitation of the Jest client library in version 2.0.3

features/opennms-es-rest/main-module/src/main/java/org/opennms/plugins/elasticsearch/rest/RestClientFactory.java
```diff
@@ -122,13 +131,23 @@ public void setServers(Set<String> servers) {
...
+
+			if (targetHosts.size() > 1) {
+				LOG.warn("Username/Password have been defined, but multiple target hosts were found. Each host will use the same credentials.");
+				LOG.warn("Username/Password have been defined, but multiple target hosts were found. Preemptive auth is only enabled for host {}",  targetHosts.get(0));
+			}
```
RangerRick: Is there a reason you have 2 almost identical `LOG.warn` calls one right after the other?

mvrueden: Yes, because I like it this way (-:

mvrueden:
The preemptive auth thing is only a temporary fix here and will be replaced in develop with a newer version of the jest library soon, which will remove the limitations here.

mvrueden:
Adjusted logging.

PR #1745, opened 2017-11-07
HZN-1181: Merge ClientFactory and JestClientFactory

mvrueden:
The first implementation of Project Drift introduced another factory to instantiate `JestClients`s. Some of that functionality was already provided by the `JestClientFactory` living inside `features/opennms-res-rest/main-module`. This PR merges both implementations and moves it into a common module called `client`.

Jira: https://issues.opennms.org/browse/HZN-1181


mvrueden:
@gallenc, @soleger feedback on this is appreciated.

j-white:
:+1:

j-white:
I'm not sure that we'll get any feedback from either @gallenc or @soleger in the near term. It's still going to be a while before me merge `features/drift` back into develop, so we could always make any adjustments afterwards.

PR #1746, opened 2017-11-07
NMS-9724: add validation of the event before attempting to send it

RangerRick:
This PR adds some basic validation of the event object(s) when `POST`ed to the event rest service.

* JIRA: http://issues.opennms.org/browse/NMS-9724

### DIFF
--- a/core/xml/src/main/java/org/opennms/core/xml/JaxbUtils.java
+++ b/core/xml/src/main/java/org/opennms/core/xml/JaxbUtils.java
@@ -432,7 +432,10 @@ public abstract class JaxbUtils {
                     };
                 }
                 if (schemaInputStream == null) {
-                    final URL schemaResource = Thread.currentThread().getContextClassLoader().getResource("xsds/" + schemaFileName);
+                    URL schemaResource = Thread.currentThread().getContextClassLoader().getResource("xsds/" + schemaFileName);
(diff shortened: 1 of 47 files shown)


j-white:
Shouldn't we be using the XSD to perform these validations instead of writing extra code for null and count checks?

RangerRick:
All that work was done in {{foundation-2017}} and backporting it is way beyond the scope of fixing a few high-value targets IMHO.

j-white:
I agree it's not worth back-porting. This approach of adding boilerplate validation code feels wrong to me though - it's error prone, requires an explicit call to validate, and difficult to maintain.

RangerRick:
I believe the reason the event objects were not marked with `@ValidateUsing` was for throughput/performance reasons, but I honestly don't know how much of an impact JAXB validation is on that, maybe it's worth just turning it back on.

RangerRick:
I've updated it to use JPA validation tags instead and call a validator on POST.

j-white:
The updates to use the validation API look good.

Since that seems to address the underlying issue, we may want to skip enabling the ValidateUsing tags unless we understand the impacts. Particularly because this targets the `foundation-2016` branch.

RangerRick:
Makes sense, I'll update the branch.

PR #1747, opened 2017-11-08
HZN-1188: Cache NodeInfo objects instead of re-fetching them all the time

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1188

### DIFF
--- a/features/telemetry/adapters/netflow/pom.xml
+++ b/features/telemetry/adapters/netflow/pom.xml
@@ -73,7 +73,6 @@
     <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-all</artifactId>
-      <version>1.10.19</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/features/telemetry/adapters/netflow/pom.xml
+++ b/features/telemetry/adapters/netflow/pom.xml
@@ -60,5 +60,21 @@
       <artifactId>hamcrest-library</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.opennms.core.test-api</groupId>
+      <artifactId>org.opennms.core.test-api.lib</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.opennms</groupId>
+      <artifactId>opennms-dao-mock</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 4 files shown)


features/telemetry/adapters/netflow/pom.xml
```diff
@@ -60,5 +60,21 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.opennms</groupId>
+      <artifactId>opennms-dao-mock</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-all</artifactId>
+      <version>1.10.19</version>
```
j-white: Unnecessary, the version for this is defined in the root pom.

features/telemetry/adapters/netflow/src/main/java/org/opennms/netmgt/telemetry/adapters/netflow/Netflow5Adapter.java
```diff
@@ -79,9 +115,23 @@
...
+    private LoadingCache<NodeInfoKey, Optional<NodeInfo>> nodeInfoCache;
+
     @PostConstruct
     public void init() {
         meter = metricRegistry.meter("persistence");
+
+        // TODO MVR make this configurable, when it is actually an osgi-module
+        nodeInfoCache = CacheBuilder.newBuilder()
+                .maximumSize(1000)
+                .expireAfterAccess(300, TimeUnit.SECONDS) // 5 Minutes
```
j-white: I think we should use `expireAfterWrite` instead of `expireAfterAccess` here to ensure that we don't cache these objects for any longer than the configured time - the node details may change.

features/telemetry/adapters/netflow/pom.xml
j-white: Version is already defined in the root pom.

mvrueden: Yeah, intellij added it, I forgot to remove it afterwards. Thanks for catching this.

features/telemetry/adapters/netflow/src/main/java/org/opennms/netmgt/telemetry/adapters/netflow/Netflow5Adapter.java
mvrueden: Yep. Makes sense.

PR #1748, opened 2017-11-08
HZN-1184: Defer initialization until needed and try indefinately

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1184


j-white:
We also need to consider thread safety here. If many threads are accessing the repository before it's initialized, each thread will end up it's in own initialization loop.

PR #1752, opened 2017-11-12
HZN-1190: Allow the Jest connection pool size to be configured

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1190

Update the default connection limits and allow them to be configured.

### DIFF
--- a/features/flows/elastic/src/main/resources/OSGI-INF/blueprint/blueprint.xml
+++ b/features/flows/elastic/src/main/resources/OSGI-INF/blueprint/blueprint.xml
@@ -18,6 +18,8 @@
             <cm:property name="elasticUser" value="" />
             <cm:property name="elasticPassword" value="" />
             <cm:property name="elasticIndexStrategy" value="monthly" />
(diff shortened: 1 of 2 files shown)


features/flows/elastic/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -18,6 +18,8 @@
             <cm:property name="elasticUser" value="" />
             <cm:property name="elasticPassword" value="" />
             <cm:property name="elasticIndexStrategy" value="monthly" />
+            <cm:property name="defaultMaxTotalConnectionPerRoute" value="-1" />
+            <cm:property name="maxTotalConnection" value="-1" />
```
indigo423: I can add those in the docs in HZN-1178. Does -1 mean, there is no upper limit?

j-white: See https://github.com/OpenNMS/opennms/pull/1752/files#diff-89d0a529573b1b0f14a653bc351e56e4R159

PR #1754, opened 2017-11-13
NMS-9670: Prevent XSS attacks when creating a filter

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-9670

This may also be worth back porting to `foundation-2016` as well.


j-white:
Could this affect existing favorites in any way? i.e. Is it possible that existing, valid favorites no longer work after this change?

mvrueden:
Existing favorites should work.
I tried it with some variations, but there is as always no 100% insurance (-:

j-white:
Given that there *might* be some effect on existing filters, let's keep this in foundation-2017 instead of back-porting all the way to foundation-2016.

PR #1755, opened 2017-11-14
NMS-9509: Cleanup default SNMP datacollection

indigo423:
Add MIB-2 HighSpeed counters to mib2 default MIB. Split the MIB2 into different files to allow users to select which OIDs need to be collected. Split configuration files and cleaned up overlapping include groups. The cleanup and separation give users better control which OIDs get collected.

* Microsoft SNMP Informant MIB into separate files to remove overhead in case several SNMP Informant extensions are not licensed.
* Removed duplicated entries for already overlapping MIB-2 system definitions from device-specific configurations like HighSpeed counters and packets.
* Removed duplicated entries for Host Resources support and created a dedicated file
* Removed duplicated entries for MIB2 UPS support
* Removed Makelsan UPS definition file, cause it is covered with default MIB2-UPS MIB
* Separated the Coffee MIB support and commented in the include to not load it by default
* MIB2 MIB is separated files
* Renamed reference OID files with prefix REF_ for very generic IETF, Router, and Printers. They are not assigned to a system definition and are included in device specific system definitions.
* Cleaned up the Dell Open Manage device support which where added Net-SNMP and Microsoft Windows configurations

* JIRA: http://issues.opennms.org/browse/NMS-9509


agalue:
Here is my first comment, NMS-9509 is related with event definitions, not data collection changes :)

indigo423:
Major release 22 develop, other than that, delete default configuration and deliver Horizon empty. The config is buggy and needs to be fixed.

agalue:
Guess I haven't explained myself correctly, so I'm going to try again:

On this PR there is a bunch of changes associated with data collection files. Now, if I follow the link to the Jira issue, I see something related with event definitions, which has nothing to do with data collection. In other words, where is the correct Jira issue associated with your data collection changes ?

indigo423:
Uh yes you are right, copy & paste issue. Thank you for the hint

PR #1756, opened 2017-11-14
NMS-9737: Cleanup default SNMP data collection

indigo423:
Add MIB-2 HighSpeed counters to mib2 default MIB. Split the MIB2 into different files to allow users to select which OIDs need to be collected. Split configuration files and cleaned up overlapping include groups. The cleanup and separation give users better control which OIDs get collected.

* Microsoft SNMP Informant MIB into separate files to remove overhead in case several SNMP Informant extensions are not licensed.
* Removed duplicated entries for already overlapping MIB-2 system definitions from device-specific configurations like HighSpeed counters and packets.
* Removed duplicated entries for Host Resources support and created a dedicated file
* Removed duplicated entries for MIB2 UPS support
* Removed Makelsan UPS definition file, cause it is covered with default MIB2-UPS MIB
* Separated the Coffee MIB support
* MIB2 MIB is separated files
* Renamed reference OID files with prefix REF_ for very generic IETF, Router, and Printers. They are not assigned to a system definition and are included in device specific system definitions.
* Cleaned up the Dell Open Manage device support which where added Net-SNMP and Microsoft Windows configurations

* JIRA: http://issues.opennms.org/browse/NMS-9737


indigo423:
@jeffgdotorg @agalue got bamboo happy

indigo423:
tests went green, anything else we need to do for this PR to apply any veto?

indigo423:
Anything to do here? +1/-1 feedback would be nice.

j-white:
+0 from me. I'd look for input from @agalue or @jeffgdotorg on this one.

opennms-base-assembly/src/main/filtered/etc/datacollection/mib2-interfaces.xml
```diff
@@ -0,0 +1,38 @@
...
+      <mibObj oid=".1.3.6.1.2.1.31.1.1.1.8" instance="ifIndex" alias="ifHCInMulticastPkts" type="Counter64"/>
+      <mibObj oid=".1.3.6.1.2.1.31.1.1.1.9" instance="ifIndex" alias="ifHCInBroadcastPkts" type="Counter64"/>
+      <mibObj oid=".1.3.6.1.2.1.31.1.1.1.11" instance="ifIndex" alias="ifHCOutUcastPkts" type="Counter64"/>
+      <mibObj oid=".1.3.6.1.2.1.31.1.1.1.12" instance="ifIndex" alias="ifHCOutMulticastPkt" type="Counter64"/>
+      <mibObj oid=".1.3.6.1.2.1.31.1.1.1.13" instance="ifIndex" alias="ifHCOutBroadcastPkt" type="Counter64"/>
+   </group>
+   <systemDef name="Enterprise MIB2 Interfaces">
+      <sysoidMask>.1.3.6.1.4.1.</sysoidMask>
+      <collect>
+         <includeGroup>mib2-interfaces</includeGroup>
```
agalue: I would not include `mib2-interfaces` by default, as we're in 2018, I seriously doubt that it makes sense to duplicate the metrics assuming that there are "slow interfaces" out there.

As a reference, during a consulting with a customer, I saw they were storing 3.3 million metrics every 5 minutes. After some discussion, we were able to validate that all their monitored devices have fast interfaces (i.e. all of them support IF-MIB::ifXTable; in other words, all the interfaces are either FastEthernet or faster), so we disable `mib2-interfaces`, because everything covered by this group is also covered by `mib2-X-interfaces`; which means, it is redundant. In fact, after this simple change, we reduced from 3.3 to 2.1 million metrics to be stored every 5 minutes; and when you're using Newts, this is a tremendous improvement.

My $.02

agalue:
For the record, even if having multiple `datacollection` groups is a valid way to easily enable/disable certain metric groups, the `include-collection` has an internal tag called `exclude-filter`, which expects a regular expression to be aplied over systemDef's names (and can be used multiple times). So, instead of dozens of XML, you can have one or a few, and then use this feature to remove systemDefs at run time. Remember, when OpenNMS is starting, it will create a thread per XML inside `etc/datacollection` to load all the files at once. These threads have a short life, but worth mention it.

That being said, the change looks good, but I would add a warning somewhere as this will end up on lots of `.rpmnew` on an upgrade and it is important to advice users about this.

opennms-base-assembly/src/main/filtered/etc/datacollection/mib2-interfaces.xml
indigo423: @agalue make sense to me, I'll change it.

indigo423:
@agalue I've removed the mib2-interfaces from the default collection and just kept the mib2-X-interfaces with the 64bit counter in default.

opennms-base-assembly/src/main/filtered/etc/datacollection/mib2-host-resources.xml
```diff
@@ -0,0 +1,60 @@
...
+      <mibObj oid=".1.3.6.1.2.1.25.2.2" instance="0" alias="hrMemorySize" type="integer"/>
+   </group>
+   <group name="mib2-host-resources-processor" ifType="all">
+      <mibObj oid=".1.3.6.1.2.1.25.3.2.1.1" instance="hrDeviceEntry" alias="hrDeviceIndex" type="string"/>
+      <mibObj oid=".1.3.6.1.2.1.25.3.2.1.2" instance="hrDeviceEntry" alias="hrDeviceType" type="string"/>
+      <mibObj oid=".1.3.6.1.2.1.25.3.2.1.3" instance="hrDeviceEntry" alias="hrDeviceDescr" type="string"/>
+      <mibObj oid=".1.3.6.1.2.1.25.3.3.1.2" instance="hrDeviceEntry" alias="hrProcessorLoad" type="Gauge32"/>
+   </group>
+   <systemDef name="Enterprise MIB2-Host-Resources">
+      <sysoidMask>.1.3.6.1.4.1.</sysoidMask>
```
agalue: I think that assuming that every single device supports HOST-RESOURCES-MIB is not correct. There is a high chance that servers support it, but for sure almost all network devices wont, that's why only a few vendor systemDef had references to them.

indigo423:
@agalue @jeffgdotorg @RangerRick  I've added also a section for this issue in the _Breaking Changes_ section. A native speaker review would be nice https://github.com/OpenNMS/opennms/blob/33065c9b0fc0792f18cc07ad7183f77e2779ead8/opennms-doc/releasenotes/src/asciidoc/releasenotes/whatsnew-22.adoc#nms-9737-cleaning-up-snmp-data-collection

indigo423:
@agalue thanks for pointing out the issues with Host Resources MIB, I've added them back to the related devices and removed them from the default enterprise OID. The Power over Ethernet MIB is removed as well. Added a hint in the release notes sees link above.

opennms-base-assembly/src/main/filtered/etc/datacollection/mib2-coffee.xml
```diff
@@ -0,0 +1,13 @@
+<datacollection-group xmlns="http://xmlns.opennms.org/xsd/config/datacollection" name="MIB2-Coffee">
+   <group name="mib2-coffee-rfc2325" ifType="ignore">
+      <mibObj oid=".1.3.6.1.2.1.10.132.2" instance="0" alias="coffeePotCapacity" type="integer"/>
+      <mibObj oid=".1.3.6.1.2.1.10.132.4.1.2" instance="0" alias="coffeePotLevel" type="integer"/>
+      <mibObj oid=".1.3.6.1.2.1.10.132.4.1.6" instance="0" alias="coffeePotTemp" type="integer"/>
+   </group>
+   <systemDef name="Enterprise MIB2-Coffee">
+      <sysoidMask>.1.3.6.1.4.1.</sysoidMask>
+      <collect>
+         <includeGroup>mib2-coffee-rfc2325</includeGroup>
```
agalue: Does it make sense to include this by default ? Doesn't look like the kind of things that ALL devices should support.

opennms-base-assembly/src/main/filtered/etc/datacollection/ref_mib2-host-resources.xml
```diff
@@ -0,0 +1,60 @@
...
+      <mibObj oid=".1.3.6.1.2.1.25.2.2" instance="0" alias="hrMemorySize" type="integer"/>
+   </group>
+   <group name="mib2-host-resources-processor" ifType="all">
+      <mibObj oid=".1.3.6.1.2.1.25.3.2.1.1" instance="hrDeviceEntry" alias="hrDeviceIndex" type="string"/>
+      <mibObj oid=".1.3.6.1.2.1.25.3.2.1.2" instance="hrDeviceEntry" alias="hrDeviceType" type="string"/>
+      <mibObj oid=".1.3.6.1.2.1.25.3.2.1.3" instance="hrDeviceEntry" alias="hrDeviceDescr" type="string"/>
+      <mibObj oid=".1.3.6.1.2.1.25.3.3.1.2" instance="hrDeviceEntry" alias="hrProcessorLoad" type="Gauge32"/>
+   </group>
+   <systemDef name="Enterprise MIB2-Host-Resources">
+      <sysoidMask>.1.3.6.1.4.1.</sysoidMask>
```
agalue: I don't think the above is relevant for network devices. HOST-RESOURCES-MIB is relevant for certain (not necessarily all) SNMP agents you can find on server operating systems only.

opennms-base-assembly/src/main/filtered/etc/datacollection/snmp-informant-exchange.xml
```diff
@@ -332,103 +233,11 @@
...
-      <mibObj oid=".1.3.6.1.4.1.9600.1.2.52.1.17" instance="ainfNetInstance" alias="ainPktsSentPerSec" type="Gauge32"/>
-      <mibObj oid=".1.3.6.1.4.1.9600.1.2.52.1.18" instance="ainfNetInstance" alias="ainPktsPerSec" type="Gauge32"/>
-   </group>
-   <!-- This group requires the pay-for SNMP Informant Advanced agent -->
-   <group name="snmpinformant-adv-termsvc-scalars" ifType="ignore">
-      <mibObj oid=".1.3.6.1.4.1.9600.1.2.81.1" instance="0" alias="sinfTsActiveSess" type="gauge"/>
-      <mibObj oid=".1.3.6.1.4.1.9600.1.2.81.2" instance="0" alias="sinfTsInactiveSess" type="gauge"/>
-      <mibObj oid=".1.3.6.1.4.1.9600.1.2.81.3" instance="0" alias="sinfTsTotalSess" type="gauge"/>
-   </group>
+   <systemDef name="Microsoft Windows SNMP Informant">
```
agalue: I would rename the `systemDef` to be `Microsoft Windows SNMP Informant Exchange`. Remember that it is mandatory that all `systemDefs` have unique names.

agalue:
What is the problem of having "cross references" ? The current implementation of the `DataCollectionConfigDAO` was designed precisely to provide the ability to "reuse" group and resource type definitions defined in one `datacollection-group` on another `datacollection-group` .

Due to the systemDefs (which is a concept that doesn't exist on any other collector implementation, and only applies to the SNMP Collector), it doesn't make sense to reference `datacollection-group` directly by file name, as the idea is to provide a list of `systemDefs` that will be associated with each `snmp-collection`, assuming that the `groups` and the `resourceTypes` not necessarily belong to the same `datacollection-group`.

For instance, you're assigning to ALL devices the `mibObjs` associated with `HOST-RESOURCES-MIB`. I'm sure that enterprise switches and routers won't be very happy to reject queries all the time, considering all the load they already have, as that MIB doesn't apply to them.

Because there is NO OID shared for all kind of "servers", that's why each server kind had references to the groups  associated with `HOST-RESOURCES-MIB` when it makes sense, as not all the agents properly implement this particular MIB.

Also, this specific use case can be applied to another "generic" MIB that should be "shared" across multiple devices, but not necessarily all those devices support.

Makes sense ?

agalue:
That's why I'm not 100% sure now about merging this. The idea makes sense, but contradicts the current design and implementation; enforcing unnecessary OIDs lots of devices, while the idea of this (if I'm not mistaken) was the opposite.

indigo423:
> What is the problem of having "cross references" ? The current implementation of the DataCollectionConfigDAO was designed precisely to provide the ability to "reuse" group and resource type definitions defined in one datacollection-group on another datacollection-group .

This is exactly was happening with generic SNMP MIB support which are called "REF_.." These are MIB groups which can be reused cause they are so generic they are supported by devices from multiple vendors.

The main issue this type of changes is addressing:

* modularize the configuration so you can add and remove easily configuration files for device support as you need. For example, you don't have any Dell servers in your network and you kick out those configs, your OpenNMS fails because the openmanage-dell groups are somewhere else in systemDefs included. 

This makes it pretty hard to maintain data collection configurations and makes it also hard to extend or remove them. Does it make sense?

indigo423:
> For instance, you're assigning to ALL devices the mibObjs associated with HOST-RESOURCES-MIB. I'm sure that enterprise switches and routers won't be very happy to reject queries all the time, considering all the load they already have, as that MIB doesn't apply to them.

This is fixed.

indigo423:
> Because there is NO OID shared for all kind of "servers", that's why each server kind had references to the groups associated with HOST-RESOURCES-MIB when it makes sense, as not all the agents properly implement this particular MIB.

> Also, this specific use case can be applied to another "generic" MIB that should be "shared" across multiple devices, but not necessarily all those devices support.

This is especially worse with the configuration we had before and existed anyway. The way it is right now, it is just harder to figure out what data is collected. This is something in the concept how OpenNMS implements data collection for SNMP and nothing I can address with PR. The only thing is I can make it more transparent why OpenNMS tries to collect things.

agalue:
I can't see why it is harder considering that the division was based by vendor, so only the vendor XML was required to be edited to add either vendor specific or generic groups from other groups; and the generic groups never had systemDefs.

To be honest, removing cross dependencies requires revisit the whole implementation of the data collection config. What would be the approach ? well, similar to the rework for having `resource-types.d` and put files there with just resourceTypes, we can create `snmp-mib-groups.d` and put group definitions there. Then, the `datacollection-group` will just have a list of `systemDefs` referencing groups from `snmp-mib-groups.d`, and these groups will be referencing resource types from `resource-types.d`.

The only way I would approve merging this is by removing ALL the "generic" systemDefs, unless you can guarantee that at least 80% of all device kinds always support the OIDs in question. Because, as a system administrator, I rather prefer to request less data that I know the devices provide, instead of retrieving a lot of either useless data or data that will never exist on the target devices, because I prefer that my devices focus on their work, and not answering unnecessary SNMP requests.

indigo423:
Sorry can't follow anymore, this changes address two topics:

* make it easier to maintain custom and vendor specific configurations
* reusable generic OIDs like IETF, Printer MIB-II oids can now be easier be identified and added in the vendor specific devices when necessary
* reduce unnecessary data collections, here are two examples what our current does and shouldn't do:
**  trying to collect on any Net-SNMP agent Dell Open Manage metrics, 
** trying to collect unnecessarily `mib2-powerethernet` for each ` <sysoidMask>.1.3.6.1.4.1.</sysoidMask>`

Just to make sure nobody merges this PR, please remove your "approved tag" and vote -1. I'll close then the PR unmerged.

indigo423:
To reduce discussion, this PR is splitted and NMS-9737 is now a Epic. See separate PRs for Epic issues in NMS-9737.

agalue:
> To reduce discussion, this PR is splitted and NMS-9737 is now a Epic. See separate PRs for Epic issues in NMS-9737.

Excellent, thanks! That helps, as there are some cases that are very straight forward and easy to approve compared to others.

PR #1759, opened 2017-11-16
HZN-1193: Improve Netflow performance and metrics

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1193

Here we improve the performance of the Netflow adapter by working on logs instead of individual packets and expose additional metrics via JMX.

The JMX metrics are collected via the opennms-jvm collection and include associated graphs.


mvrueden:
This looks pretty straight forward to me. I think it makes sense to persist all flows from the bucket, vs all flows from a packet. 👍 

OT: If we keep going with all the words, we need a glossary (-:

PR #1760, opened 2017-11-16
HZN-1194: Consolidate existing JMX Detector/Monitor/Collector code

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1194


j-white:
I always wondered why we had different classes for these - it's nice to seem them go.

If an existing instance was using say the `JBoss` detector/monitor/collector, how should they migrate to this new approach?

mvrueden:
For a collector, they should use the Jsr160Collector instead.
For the monitor, they should use the Jsr160Monitor instead.
For the detector, they should use the Jsr160Detector instead.

Configuration wise, the `url` property must be defined instead the deprecated properties.
Optionally `username` and `password` can be provided.
If the protocol is not `jmxrmi` or `jmxmp` a jar containing the protocol e.g. `http-remoting` must be placed in the opennms/lib directory. See https://docs.opennms.org/opennms/branches/develop/guide-admin/guide-admin.html#_jmxcollector -> 3rd party jmx services for more details.

PR #1761, opened 2017-11-16
HZN-1192: Make netflow-adapter run inside karaf

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1192


j-white:
Looks good. Now that the flows feature is installed by default, we should update the setup instructions in the docs.

j-white:
Looks good. There's a conflict with my work from HZN-1193 now though :)

PR #1764, opened 2017-11-17
NMS-8905: Create fresh MockLogger instances in MockLoggerFactory

soleger:
This PR changes MockLoggerFactory so that it doesn't cache MockLogger instances. This means that MockLoggers will use the log level that is specified by calling static methods on MockLogAppender (as it was intended to work).

This reduces the log output of the build on Bamboo from about 3.2 million log messages down to about 1.8 million since DEBUG messages are no longer leaking into the output (from tests where DEBUG logging is not enabled).

There's also thread safety/API cleanup here for MockLogAppender. 

* JIRA: http://issues.opennms.org/browse/NMS-8905


j-white:
:+1:

PR #1766, opened 2017-11-20
NMS-9745: Karaf snmp:show-config command lists value of timeout for retries

jeffgdotorg:
* JIRA: http://issues.opennms.org/browse/NMS-9745


j-white:
Oops.

PR #1767, opened 2017-11-21
HZN-1182: Add more configuration possibilities to elastic jest client factory

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1182


j-white:
Looks good - just need to fix those tests.

opennms-base-assembly/src/main/filtered/etc/elastic-flow-persistence-credentials.xml
```diff
@@ -0,0 +1,2 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
j-white: Can we rename this to something more generic like  `elastic-credentials.xml`? This would allow us to reuse in other contexts, since there isn't anything specific to flows here.

mvrueden: Yes, sure, I just have to place it somewhere else then.

mvrueden:
I changed the name to `elastic-credentials.xml` and also moved the credentials files to the generic `jest.client` module. This looks now much simpler. :)

mvrueden:
I rebased against `features/drift` and made the tests pass.

PR #1768, opened 2017-11-22
NMS-9749: The content of the clone foreign source definition dialog is misleading

agalue:
The content of the clone foreign source definition dialog is misleading

* JIRA: http://issues.opennms.org/browse/NMS-9749

I verified the solution with the customer who reported the problem.


j-white:
Works as expected.

PR #1769, opened 2017-11-22
NMS-9739: allow pre-authentication through HTTP headers

RangerRick:
This commit creates a new pre-authentication filter for servlets
(RequestHeaderPreAuthenticationProcessingFilter) and configures it
in applicationContext-spring-security.xml.

It also refactors the existing servlet attribute filter
(RequestAttributePreAuthenticationProcessingFilter) to look more
like the new header-based filter.

Both now require a property "enabled" to be set to "true" or else
they will silently pass control to the next filter in the chain
without authenticating.  This allows us to leave them always
configured in the spring security .xml without having to make
large/unusual config changes besides enabling the filter.

* JIRA: http://issues.opennms.org/browse/NMS-9739


j-white:
Works as expected.

Is it worth adding a section to the docs for this? I don't think we have much of anything in terms of authentication yet though.

RangerRick:
Yeah, probably should.  I'll look into that Monday.

PR #1770, opened 2017-11-27
NMS-9743: Jexl Measurements API enhancements

gallenc:
* JIRA: https://issues.opennms.org/browse/NMS-9743

* BAMBOO: https://bamboo.opennms.org/browse/OPENNMS-ONMS1923

This pull request is against foundation-2017 because it is a minimal impact enhancement which will be immediately useful for Meridian users. There should be no impact on current users using expressions (unless they have overlapping name spaces which is unlikely). A new set of tests have been added which test the new feature. The old tests still work as expected.

This is a small enhancement to the expression evaluation code in the measurements api. It helps creation of more complex expressions which reference previous samples in the time series. This this allows for instance  FIR / IIR filters to be defined e.g. y = a * f(n) + b * f(n-1) + c * f(n-2)
 It also allows expressions to be defined and stored as string variables which means that formulae which apply corrections to measurements can be applied on a per measurement basis ( useful for IoT). 
The patch adds additional references into the jexl context to objects which are already created in the measurement code but not previously accessible from an expression. e.g.
__i is the index to the current sample being evaluated
__attribute-name is a reference to the array containing all the samples for a given attribute.
Thus for an expression attribute icmp, the value icmp can also be retrieved using  __icmp[__i]

Additional predefined functions are also included which retrieve previous values of an attribute, substituting NaN, 0 or a given constant for references before the start of the sample range e.g. 
fn:arrayNaN("sampleName", n) returns the nth previous sample or NaN if before the start of the array. i.e. fn:arrayNaN("icmp", 5) returns the 5th previous sample for icmp.

Finally  to allow string formula to be evaluated on the fly ;
__context references the whole context
__jexl references the jexel engine
which allows the following expression to reference a  string variable called __formula
__jexl.createExpression(__formula).evaluate(__context)

Full documentation is included in a More Advanced Expressions section at the end of the Measurements Api description  in the development guide 
https://docs.opennms.org/opennms/branches/jira-NMS-9743/guide-development/guide-development.html#_measurements_api


indigo423:
Just in general:

> This pull request is against foundation-2017 because it is a minimal impact enhancement which will be immediately useful for Meridian users. 

MERIDIAN means the code has more time in operation than Horizon code, which ends up in a product which can be supported at less cost with other service level agreements.

In your case developer and support have to wage risk/benefit ratio related to your code changes.

features/measurements/impl/src/main/java/org/opennms/netmgt/measurements/impl/JEXLExpressionEngine.java
```diff
@@ -122,6 +126,12 @@ public void applyExpressions(final QueryRequest request, final FetchResults resu
         jexlValues.put("__inf", Double.POSITIVE_INFINITY);
         jexlValues.put("__neg_inf", Double.NEGATIVE_INFINITY);
         jexlValues.put("NaN", Double.NaN);
+        
+        // add context as a reference to be picked up by complex formulae
```
j-white: Can we introduce a new `eval` function (or similar) instead of exposing the engine and context as variables?

We shouldn't allow scripts to interface with the engine directly.

features/measurements/impl/src/test/java/org/opennms/netmgt/measurements/impl/JEXLExpressionEngineEnhancedTest.java
```diff
@@ -0,0 +1,307 @@
...
+			success = 
+					Double.isNaN(result[0]) &&
+					Double.isNaN(result[1]) &&
+					Double.isNaN(result[2]) &&
+					Double.isNaN(result[3]) &&
+					Double.isNaN(result[4]) &&
+					(result[5] == 5) &&
+					(result[6] == 6);
+
+		} catch (Exception e){
```
j-white: Why not just re-throw the exception instead of catching it here? The test will fail the and the stack will be printed automatically.

features/measurements/impl/src/test/java/org/opennms/netmgt/measurements/impl/JEXLExpressionEngineEnhancedTest.java
```diff
@@ -0,0 +1,307 @@
...
+		try{
+
+			double[] result = performExpression(expression);
+
+			for (double value: result){
+				sb.append(value+",");
+			}
+
+			success = 
+					Double.isNaN(result[0]) &&
```
j-white: If the test were to fail, the output gives no indication as to what the expected values are. For this reason I prefer using asserts on the values were checking i.e. `assertEquals(Double.NaN, result[0], delta);` instead of tracking them via boolean variables.

opennms-doc/guide-development/src/asciidoc/text/rest/measurements.adoc
```diff
@@ -160,3 +160,99 @@ curl -X POST  -H "Accept: application/json" -H "Content-Type: application/json"
     ]
 }
 ----
+===== More Advanced Expressions
+
+The jexl 2.1.x library is used to parse the expression string and this also allows java objects and predefined functions to be included in the expression.
```
j-white: jexl vs JEXL
java vs Java

opennms-doc/guide-development/src/asciidoc/text/rest/measurements.adoc
```diff
@@ -160,3 +160,99 @@ curl -X POST  -H "Accept: application/json" -H "Content-Type: application/json"
...
+Jexl uses a context which is pre-populated by OpenNMS with the results of the query.
```
j-white: Jexl vs JEXL

features/measurements/impl/src/main/java/org/opennms/netmgt/measurements/impl/SampleArrayFunctions.java
```diff
@@ -0,0 +1,181 @@
...
+			if(sample==null) throw new RuntimeException("not in jexl context: __"+sampleName);
+			i = (Integer) m_context.get("__i");
+			if(i==null) throw new RuntimeException("not in jexl context: __i");
+			
+			if(i-n < 0) return Double.valueOf(0);
+			
+			return sample[i-n];
+		}
+		catch (Exception ex){
+			LOG.error("problem retrieving prior sample "+sampleName+" (i - n) ("+i+"-"+n+")",ex);
```
j-white: We don't need to print the full stack trace if we're catching exception that we threw in this very same function.

I'd prefer to use blocks like:
```
if(sample==null) {
  LOG.warn("{} is not in the JEXL context. A NaN value will be used.", sampleName);
  return Double.NaN;
}
```
instead of throwing/catching new exceptions.

gallenc:
new pull request 1775 created targeting release 21.0.2

https://github.com/OpenNMS/opennms/pull/1775

PR #1771, opened 2017-11-27
NMS-9754: RTC subscription events should not be persisted to DB

jeffgdotorg:
* JIRA: http://issues.opennms.org/browse/NMS-9754

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/events/opennms.internal.events.xml
+++ b/opennms-base-assembly/src/main/filtered/etc/events/opennms.internal.events.xml
@@ -81,7 +81,7 @@
       <event-label>OpenNMS-defined internal event: rtc subscribe</event-label>
       <descr>&lt;p>This event is generated to RTC by any process that
             wishes to receive POSTs of RTC data.&lt;/p></descr>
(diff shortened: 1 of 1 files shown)


soleger:
This is a config change so it can't be targeted for `foundation` without impacting upgrades.

indigo423:
Can we use a section in the "Whats new" section which comes from a JIRA component "Configuration Files"? This way people have one spot to look when they do updates.

j-white:
Can we just re-target this to develop? May as well have it in new releases, then not at all.

mvrueden:
Retargeted to `develop`. It is now conflicting @jeffgdotorg can you take a look?

jeffgdotorg:
As @j-white says, better to get this in the next Horizon than never. Targeted the issue to Horizon 26, requested review.

jeffgdotorg:
(Fixed conflict)

indigo423:
@jeffgdotorg If we want this to have in 26.0.0 we have to retarget to release-26.0.0 branch instead of develop. Develop is going to be 27.0.0-SNAPSHOT.

PR #1773, opened 2017-11-28
NMS-9742: Prevent duplicate reduction key entries and fix already wrong memos

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-9742

Might be worth to target to `foundation-2017`


mvrueden:
Wrong branch name. Sorry. Closing.

PR #1774, opened 2017-11-28
NMS-9742: Prevent duplicate reduction key entries and fix already wrong memos

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-9742

### DIFF
--- a/features/topology-map/org.opennms.features.topology.api/src/main/java/org/opennms/features/topology/api/browsers/OnmsVaadinContainer.java
+++ b/features/topology-map/org.opennms.features.topology.api/src/main/java/org/opennms/features/topology/api/browsers/OnmsVaadinContainer.java
@@ -184,7 +184,7 @@ public abstract class OnmsVaadinContainer<T,K extends Serializable> implements C
             }
             // Ensure that the number of items expected matches with the actual ones. See issue NMS-8079 fore more details.
             if (beans.size() != rowMap.size()) {
-                throw new IllegalStateException("The cache is supposed to carry " + beans.size() + " but only contains " + rowMap.size() + " items.");
+                throw new IllegalStateException("The cache is supposed to carry " + beans.size() + " but contains " + rowMap.size() + " items.");
             }
         }
     }
@@ -462,6 +462,8 @@ public abstract class OnmsVaadinContainer<T,K extends Serializable> implements C
         }
         // Ensure that the number of items expected matches with the actual ones. See issue NMS-8079 fore more details.
         if (itemIds.size() != numberOfItems) {
+            LOG.warn("The container is supposed to carry {} but contains {} item. Expected itemIds: {}. Actual items: {}. Executed Criteria: {}. Bailing.",
+                    numberOfItems, itemIds.size(), itemIds, getItemsForCache(m_datasource, getPage()), getCriteria(getPage(), true));
             throw new IllegalStateException("The container is supposed to carry " + numberOfItems + " but only contains " + itemIds.size() + " items.");
         }
         return itemIds;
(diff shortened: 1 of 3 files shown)


features/topology-map/org.opennms.features.topology.api/src/main/java/org/opennms/features/topology/api/browsers/OnmsVaadinContainer.java
```diff
@@ -184,7 +184,7 @@ public void reload(Page page) {
             }
             // Ensure that the number of items expected matches with the actual ones. See issue NMS-8079 fore more details.
             if (beans.size() != rowMap.size()) {
-                throw new IllegalStateException("The cache is supposed to carry " + beans.size() + " but only contains " + rowMap.size() + " items.");
+                throw new IllegalStateException("The cache is supposed to carry " + beans.size() + " but contains " + rowMap.size() + " items.");
```
j-white: In this case, it would of been useful if the exception also carried extra details about the contents of the bean and the resulting map.

core/schema/src/main/liquibase/21.0.0/changelog.xml
```diff
@@ -111,5 +111,46 @@
...
+      UPDATE memos
+      SET body = subquery.body
+      FROM subquery
+      WHERE memos.reductionkey = subquery.reductionkey
+      AND memos.type = subquery.type
+      AND memos.updated = subquery.updated;
+    </sql>
+    <!-- Delete duplicate entries except the newest -->
+    <sql>
+      DELETE FROM memos WHERE id IN (
```
j-white: I think there's a problem with the delete here, if I run:
```
delete from memos;
insert into memos (id,created,updated,author,body,reductionkey,type) VALUES (1,NOW(),NOW(),'me','body1','rkey1','note');
insert into memos (id,created,updated,author,body,reductionkey,type) VALUES (2,NOW(),NOW(),'me','body2','rkey1','note');
insert into memos (id,created,updated,author,body,reductionkey,type) VALUES (3,NOW(),NOW(),'me','body3','rkey2','note');
insert into memos (id,created,updated,author,body,reductionkey,type) VALUES (4,NOW(),NOW(),'me','body4','rkey2','note');
insert into memos (id,created,updated,author,body,reductionkey,type) VALUES (5,NOW(),NOW(),'me','body5','rkey3','note');
```

and then execute these two SQL statements, I'm left with 2 entries in the memos table, whereas I would expect there to be 3.

mvrueden: Yes, you are right. The SQL only works if you have only one duplicate. Good thing you caught that. It could have created a huge mess without anyone noticing. I changed the SQL to work properly.

features/topology-map/org.opennms.features.topology.api/src/main/java/org/opennms/features/topology/api/browsers/OnmsVaadinContainer.java
mvrueden: In this case we would have get 2 lists. One containing 23, and the other one containing 22 items. A manual debugging/investigation of the problem is still needed. In my opinion it is not very helpful to show that information. This exception is purely to raise awareness that something is wrong.

mvrueden:
I added some more log details.

j-white:
Cool, those logs should be helpful it happens again.

Is `release-21.0.2` the right target for this? Seems a like foundation branches may also be affected.

mvrueden:
Yeah, that would make sense (-:

mvrueden:
I rebased against `foundation-2017`.

PR #1777, opened 2017-11-29
Sonar cleanup: Replace lambdas with function references

soleger:
Improve efficiency a microscopic amount by using function references directly where possible.


features/ticketing/jira-client/dependency-reduced-pom.xml
```diff
@@ -1,77 +1,78 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
j-white: This seems out of place - this file should probably be deleted and added to .gitignore instead.

soleger: Yeah, that file snuck into the commit after I did a build. I'll clean it up.

PR #1778, opened 2017-11-29
HZN-1195: Implement retry strategy for ElasticFlowRepository

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1195


features/flows/elastic/log4j2.properties
```diff
@@ -0,0 +1,74 @@
+status = error
```
j-white: I'm guessing this doesn't belong here.

features/flows/elastic/src/test/java/org/opennms/netmgt/flows/elastic/utils/ElasticSearchRule.java
```diff
@@ -0,0 +1,97 @@
...
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+import com.google.common.base.Throwables;
+
+/**
+ * A JUnit rule which starts an embedded elastic-search instance.
+ */
+public class ElasticSearchRule implements TestRule {
```
j-white: Perhaps we should add to the to `core/test-api/elasticsearch` module instead of defining it here.

features/flows/elastic/log4j2.properties
mvrueden: I am not sure. I just copied over the official elastic `log4j2.properties` file and modified it. I kept this line. So I would keep it for now. 

Source: https://github.com/elastic/elasticsearch/blob/master/distribution/src/main/resources/config/log4j2.properties

mvrueden:
I moved the logic to determine failed items to the general jest module. If we need more specific error handling or the "inifinte retry strategy" in other modules as well it is very easy to move the according classes as well. For now I'd say let's wait until we consolidate the other modules.

core/test-api/elasticsearch/src/main/java/org/opennms/core/test/elastic/ElasticSearchRule.java
```diff
@@ -0,0 +1,95 @@
...
+        return new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                if (!config.isManualStartup()) {
+                    startServer();
+                }
+                try {
+                    base.evaluate(); // execute the unit test
+                } finally {
+                    eserver.shutdown();
```
j-white: If `isManualStartup` is set, and the server wasn't actually started, this will throw an NPE. I ran into this while rying to use the rule in another test.

core/test-api/elasticsearch/src/main/java/org/opennms/core/test/elastic/ElasticSearchServerConfig.java
```diff
@@ -0,0 +1,124 @@
...
+
+import org.elasticsearch.cluster.ClusterName;
+import org.elasticsearch.common.network.NetworkModule;
+import org.elasticsearch.common.settings.Settings;
+import org.elasticsearch.env.Environment;
+import org.elasticsearch.node.Node;
+
+public class ElasticSearchServerConfig {
+
+    private static final String DEFAULT_HOME_DIRECTORY = "target/elasticsearch-home";
```
j-white: Can we change the default behavior so that either A) a temporary directory is used, or B) this directory is cleared after a successful test so that multiple tests in the same module can use this rule without worrying the state of the previous test.

features/flows/elastic/pom.xml
```diff
@@ -72,8 +80,48 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.hamcrest</groupId>
+      <artifactId>hamcrest-library</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.github.tomakehurst</groupId>
+      <artifactId>wiremock</artifactId>
+      <version>2.11.0</version>
```
j-white: We already use wiremock in a few other modules. Is it worth consolidating the version in the root pom?

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/FlowRequestExecutor.java
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+
+    public <T extends JestResult> T retry(JestClient client, Action<T> clientRequest) {
+        LOG.debug("Retry request");
+        if (cooldownInMs > 0) {
+            LOG.debug("Sleep " + cooldownInMs + " before retrying");
+            try {
+                Thread.sleep(cooldownInMs);
+            } catch (InterruptedException e) {
```
j-white: If we get interrupted while retrying i.e. on application stop, we should probably abort instead of swallowing the exception and continuing to retry.

features/flows/rest/src/main/java/org/opennms/web/rest/v1/XXXXDTO.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * For more information contact:
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.web.rest.v1;
+
+public class XXXXDTO {
```
j-white: XXXX

core/test-api/elasticsearch/src/main/java/org/opennms/core/test/elastic/ElasticSearchServerConfig.java
mvrueden: I had some issues when doing that, as the Node is doing some magic when instantiating (even before the node is started) which caused some side-effects. I then decided to make it a non-dynamic directory. I can give it another try, though.

features/flows/elastic/pom.xml
mvrueden: Sure.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/FlowRequestExecutor.java
mvrueden: Yeah, makes sense too.

core/test-api/elasticsearch/src/main/java/org/opennms/core/test/elastic/ElasticSearchRule.java
mvrueden: Good point, I didn't try that.

mvrueden:
I had to rebase due to merge conflicts with `features/drift`. Everything since commit `7979ef6` is new.

j-white:
👍

PR #1779, opened 2017-11-29
HZN-1204: Upgrade Maven to version 3.5.2

soleger:
There's a new maintenance version of Maven. It seems like the biggest new feature is color console output (which does make things more readable).

* JIRA: http://issues.opennms.org/browse/HZN-1204


mvrueden:
Funny, I thought about doing this myself yesterday :D

soleger:
I only got the notion to do it after I noticed that the old Maven was using some of the deprecated debug flags that I fixed in #1780. :)

PR #1787, opened 2017-12-07
HZN-1210: Implement flow application classification

mvrueden:
Proposal for a ClassificationEngine implementation.
This contains only the engine, no persistence. This means, it will be part of the flow adapter, but not do anything yet. Persistence is addressed in a different issue.

Some aspects here need to be reworked while working on other related [issues](https://issues.opennms.org/browse/HZN-1206).

* JIRA: https://issues.opennms.org/browse/HZN-1210


features/flows/classification/src/main/java/org/opennms/netmgt/flows/classification/CachingClassificationEngine.java
```diff
@@ -0,0 +1,85 @@
...
+public class CachingClassificationEngine implements ClassificationEngine {
+
+    // The cache. Must return Optional<String> as values retrieved from the cache loader MUST NOT be null.
+    private final LoadingCache<ClassificationRequest, Optional<String>> cache;
+    private final ClassificationEngine delegate;
+
+    public CachingClassificationEngine(ClassificationEngine delegate) {
+        this.delegate = Objects.requireNonNull(delegate);
+        this.cache = CacheBuilder.newBuilder()
+            .maximumSize(5000) // TODO MVR make it configurable
```
j-white: Can we make these configurable now instead of deferring this?

features/flows/classification/src/main/java/org/opennms/netmgt/flows/classification/persistence/Rule.java
```diff
@@ -0,0 +1,134 @@
...
+     * 80,8980,8000-9000
+     * Must always be provided.
+     */
+    private String port;
+
+    /**
+     * The protocol to map.
+     * May define ranges, e.g. 17-20; 2,7,17
+     */
+    private String protocol;
```
j-white: We don't need/want to match protocol ranges. A specific protocol will suffice.

features/flows/classification/src/main/java/org/opennms/netmgt/flows/classification/DefaultClassificationEngine.java
```diff
@@ -0,0 +1,117 @@
...
+        return first.orElse(null);
+    }
+
+    protected static ClassificationRequest createClassificationRequest(NetflowDocument document) {
+        final ClassificationRequest request = new ClassificationRequest();
+        request.setProtocol(ProtocolType.createFrom(document.getIpProtocol()));
+        request.setLocation(document.getLocation());
+
+        // Decide whether to use source or dest address/port to determine application mapping
+        if (isInitiator(document)) {
```
j-white: We should also store the result of this function within the document (as a boolean value), so that we can filter ingress/egress traffic at query time without having to reapply this logic.

features/flows/classification/src/main/java/org/opennms/netmgt/flows/classification/DefaultClassificationEngine.java
```diff
@@ -0,0 +1,117 @@
...
+        }
+        return request;
+    }
+
+    protected static boolean isInitiator(NetflowDocument document) {
+        if (document.getSourcePort()  > document.getDestPort()) {
+            return true;
+        } else if (document.getSourcePort() == document.getDestPort()) {
+            // Tie breaker
+            return document.getIpv4SourceAddress().compareTo(document.getIpv4DestAddress()) > 0;
```
j-white: Comparing the numeric representation of the IP address instead of their strings would make more sense here.

features/flows/classification/src/main/java/org/opennms/netmgt/flows/classification/DefaultClassificationEngine.java
j-white: It may also be worthwhile to perform this logic only if there is a cache miss as opposed to doing it every-time there is a cache lookup.

features/flows/classification/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -0,0 +1,31 @@
...
+
+    <!-- Classification Engine -->
+    <bean id="defaultClassificationEngine" class="org.opennms.netmgt.flows.classification.DefaultClassificationEngine">
+        <argument ref="classificationRuleDao" />
+    </bean>
+    <bean id="cachingClassificationEngine" class="org.opennms.netmgt.flows.classification.CachingClassificationEngine">
+        <argument ref="defaultClassificationEngine" />
+    </bean>
+    <bean id="threadSafeClassificationEngine" class="org.opennms.netmgt.flows.classification.ThreadSafeClassificationEngine">
+        <argument ref="cachingClassificationEngine" />
```
j-white: The cache is already thread safe, shouldn't we wrap the DefaultClassificationEngine instead?

features/flows/classification/src/main/java/org/opennms/netmgt/flows/classification/DefaultClassificationEngine.java
mvrueden: Yeah, I forgot about that (-:

features/flows/classification/src/main/java/org/opennms/netmgt/flows/classification/DefaultClassificationEngine.java
mvrueden: Yeah, true, I forgot about that.

features/flows/classification/src/main/java/org/opennms/netmgt/flows/classification/persistence/Rule.java
mvrueden: Yes, the implementation doesn't support that. The comment is in-accurate.

features/flows/classification/src/main/resources/OSGI-INF/blueprint/blueprint.xml
mvrueden: Yes, true.

j-white:
Could we also include an initial static rule-set here? Port mappings from the IANA would be a good start: https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.txt

PR #1788, opened 2017-12-07
NMS-9767: Consider opennmsHome when adding version.properties

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-9767


j-white:
Should this be targeted to foundation-2017?

mvrueden:
I don't have a strong opinion about this. It is not a very critical bug.

mvrueden:
Rebased against `foundation-2017`.

RangerRick:
But `version.properties` isn't in `$OPENNMS_HOME` anyways, it lives in the webapp's `WEB-INF` directory, so what exactly is this accomplishing?

mvrueden:
It does not live directly in opennms_home, but in `$OPENNMS_HOME/jetty_webapps/opennms/WEB-INF`. At the moment the `version.properties` location is relative to the `working.directory` of the starting script. It assumes `$OPENMS_HOME`, so I think this change makes sense.

RangerRick:
Oh, I see, `VERSION_PROPERTIES` isn't just the filename like the others, it's a relative path.  Never mind.  :)

PR #1789, opened 2017-12-07
HZN-1212: Make FlowRestService run in osgi-container

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1212


features/flows/rest/src/main/java/org/opennms/netmgt/flows/rest/FlowRestService.java
```diff
@@ -0,0 +1,56 @@
...
+import javax.ws.rs.Consumes;
+import javax.ws.rs.GET;
+import javax.ws.rs.POST;
+import javax.ws.rs.PUT;
+import javax.ws.rs.Path;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
+@Path("flows")
+public interface FlowRestService {
```
j-white: Can we split the interface and implementation into different bundles? This will make it easier to write client code that doesn't depend on the impl. deps.

features/flows/rest/pom.xml
```diff
@@ -8,36 +8,58 @@
...
     </dependency>
     <dependency>
       <groupId>org.opennms</groupId>
-      <artifactId>opennms-util</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.opennms</groupId>
-      <artifactId>opennms-dao</artifactId>
+      <artifactId>opennms-dao-api</artifactId>
+      <scope>provided</scope>
```
j-white: This is why I don't like using the Maven plugin that generates the Karaf feature files. By marking these as provided, we omit them from the feature dependencies, but then need to re-declare them if we add a dependency on this (Maven) module.

PR #1791, opened 2017-12-12
HZN-1215: Consolidate flow configuration

mvrueden:
In order to have a centralized flow configuration we need some kind of configuration object, otherwise the configuration would be scattered all over the modules.

* JIRA: https://issues.opennms.org/browse/HZN-1215


features/ticketing/jira-client/dependency-reduced-pom.xml
```diff
@@ -1,77 +1,78 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
mvrueden: This file was added by accident. I will remove it.

PR #1793, opened 2017-12-14
HZN-1207: Implement DAOs to persist rules permanently

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1207


features/flows/classification/engine/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -12,11 +12,16 @@
 ">
 
     <!-- DAO -->
-    <bean id="classificationRuleDao" class="org.opennms.netmgt.flows.classification.persistence.StaticRuleClassificationDAO"/>
+    <reference id="classificationRuleDao" interface="org.opennms.netmgt.flows.classification.persistence.api.ClassificationRuleDao" availability="mandatory"/>
+
+    <!-- Rule Provicer -->
+    <bean id="classificationRuleProvider" class="org.opennms.netmgt.flows.classification.provider.DaoClassificationRuleProvider">
+        <argument ref="classificationRuleDao"/>
```
j-white: Are we losing the built-in static rules here?

mvrueden: Yes we are loosing the static rules here.
I thought having a "csv import" in the rest api would allow to just push the official IANA CSV to the endpoint and have it import the rules.
Alternatively we could always return the static rules if there are no rules in the database.
We can even make that configurable. Let's say via a property `org.opennms.flows.classification.rules.ianadefaultfallback` for example.

j-white: Here's what I'm thinking the behavior could be:
* On a fresh install, the default rules are automatically applied, i.e. if you start sending flows, they are classified using the default rule set and no further action is necessary
* Users can add additional rules, to be used in conjunction with the default rule set
* If the default rules are updated in a subsequent release, they are automatically made available to users unless they have forked these (see the next point)
* Users may optionally choose to "fork" the default rule set, in which case they can modify these and are then responsible for any further updates

indigo423: I would definitely prefer to have a way to disable the IANA import on a fresh install. I don't like to populate a system I'm in control of with tons of garbage which makes it then hard to get rid of or fiddle around ways to why it is not behaving as I thought it would be. This makes so many things intransparent. The first thing when I would go in prod is trying to get rid of this thing and populate the system with information I'm aware of and I can control. If there is a property somebody can set to say, yeah I'm lazy give me the IANA stuff, ok - but I'm against it as a default.

mvrueden: I can see that we need some kind of rules at the moment as there is no you yet.
This allows us to see the classification engine in action and also improve it even without a UI.

However, I don't like the idea of having the IANA rules beeing enabled by default.
There are thousands of rules and even the simplest example, ports 80 and 8080 would map to `http` and `http-alt`, which in my opinion is sub-optimal. 

At the moment I tend to enable the REST API to accept CSV formatted files and also allow the official IANA CSV file to be imported (See https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.csv). 

In order to be able to have flows classified for now, I can enable the static rules and disable them afterwards. This way we have the maximum transparency. 

If we absolutely have to apply the IANA rules by default (configurable) I would not implement a forking mechanism, as we already have some similar functionality with the foreign sources and from my point of view that sucks.

mfuhrmann: From my point of view I don't like the idea to have even more configurations in OpenNMS. We already have so many files with so many many lines of configurations.
I can still remember the merge hell commit! :-P

j-white: My main point from above is that we should provide some default rule-set that provides some level of classification out-of-the-box, whether it be IANA or not. The IANA rules are what we have for now.

Also, we need to consider how we will make updates to the default rule set in the future, and how we will allow users to modify or updates these rules.

We don't need to sort all of this out now, so I agree that for the time being we should go ahead and re-enable the static rule-set.

PR #1794, opened 2017-12-14
HZN-1180: Implement IPFIX and Netflow v9 collection

fooker:
JIRA: http://issues.opennms.org/browse/HZN-1180

Remarks:
- This needs https://github.com/OpenNMS/opennms-system-test-api/pull/10 for smoke tests
- The adapter side is still a complete mess:
  - A lot of duplicated code between IPFIX/v9 and v5
  - Mapping is completely unclear - see TODOs
- Using BSON for transport is not optimal
- The smoke test is incomplete as the flow REST API does not return all results - see TODOs


fooker:
... and this MUST be squash-merged!

j-white:
Can we break this up into smaller PRs? I find the sheer size of this one makes it difficult to review.

Perhaps we can start with the protocol decoding and handling, followed by the integration with telemetryd and then persistence.

fooker:
Splitted out to #1800.

fooker:
And #1834

PR #1800, opened 2017-12-20
HZN-1180: Added parser/listener for NetFlowV9 and IPFIX

fooker:
* JIRA: http://issues.opennms.org/browse/HZN-1180

This is a partial PR (splitted and squashed #1794) for implementing NetflowV9 and IPFIX for telemetryd.


fooker:
Still huge, but only the parsers for the protocols and the transmitters.

features/telemetry/listeners/flow/src/main/java/org/opennms/netmgt/telemetry/listeners/flow/ipfix/UdpListener.java
```diff
@@ -0,0 +1,154 @@
...
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.channel.ChannelInitializer;
+import io.netty.channel.ChannelOption;
+import io.netty.channel.EventLoopGroup;
+import io.netty.channel.nio.NioEventLoopGroup;
+import io.netty.channel.socket.DatagramChannel;
+import io.netty.channel.socket.nio.NioDatagramChannel;
+import io.netty.util.concurrent.ScheduledFuture;
+
+public class UdpListener implements Listener {
```
j-white: Most of this is identical to the Netflow 9 UDP listener. Is it worth sharing this code?

features/telemetry/listeners/flow/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+      <groupId>io.netty</groupId>
+      <artifactId>netty-all</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>net.sf.opencsv</groupId>
+      <artifactId>opencsv</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java</artifactId>
```
j-white: Do we still need protobuf if we're using bson?

PR #1803, opened 2017-12-21
NMS-9766: Limit the number of threads that can be used to send notifications

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9766

Here we introduce a thread pool in notifd that is used to invoke the notification tasks, and use this pool to limit the number of threads that are active at any given time. Additional notifications tasks that a waiting for threads will be kept in an unbounded queue.

By default we limit the number of threads to 100, and allow it to be configured via the notifid-configuration.xml.


pbrane:
I can finally stop thinking about this problem; what’s it been, about a 100 years?

PR #1807, opened 2018-01-02
Fix SyslogMessageTest after year wrap

soleger:
Oops, my smart year parsing inside Syslogd was too smart for one of the unit tests which still assumed that a `March` timestamp must be `March, ${currentYear}`.


RangerRick:
Looks like it passed, I'll merge.

PR #1808, opened 2018-01-02
NMS-9807: Remove duplicated release notes history

indigo423:
Removed duplicated release notes and changelog carried for older versions. The release notes are versioned along with the source code and the history is maintained with publishing releases on docs.opennms.org.

* JIRA: http://issues.opennms.org/browse/NMS-9807


indigo423:
The test fail is unrelated to the changes see: https://bamboo.opennms.org/browse/OPENNMS-ONMS1988-UNIT-1/test/case/207211069

RangerRick:
...ok?  That's nice?  It has nothing to do with me.  :)

j-white:
Those tests should no longer fail if you rebase this over develop.

indigo423:
tests went green, anything else we need to do for this PR to apply any veto?

j-white:
I'm a 👍 for this, and generally keeping the information in documentation tree tied to the current version.

mvrueden:
@indigo423 can you rebase this against `develop` to make the tests pass?

indigo423:
@mvrueden thanks for the hint, rebased today and we will see how it goes.

indigo423:
yeah it's green :)

indigo423:
Need to maintain this PR again because of release. Anything we need to do here?

j-white:
I think we need to +1 from @RangerRick to merge this, since he is the one that typically updates these.

RangerRick:
My only issue is that we currently have links that point to "latest" and then the #whatever-hash for the version to link to old release notes.  If everyone is on-board with breaking those that's fine I guess, but I don't have a super strong opinion about it, on the Horizon side.

PR #1812, opened 2018-01-04
NMS-9789: add diskIO util metrics and graphs

mfuhrmann:
* JIRA: https://issues.opennms.org/browse/NMS-9789

I added the disk IO utilization metrics and graphs because they were missing and really useful and netsnmp provides it per default.
Also I didn't include this definition because the existing disk IO stats are also inactive in default setting.
The graphs are copied from CPU LOAD so they have the same look. Is that ok?

Hopefully you can still merge it manually...


mfuhrmann:
For example: 
image

j-white:
Isn't there some overlap with https://github.com/OpenNMS/opennms/pull/1815 here?

mfuhrmann:
It seems so. I talked some days before with @indigo423 about this topic. I didn't know that @indigo423 wanted to work on it. But it seems his version is better than mine. So feel free to close this PR.

indigo423:
Picked the changes to #1815

PR #1813, opened 2018-01-09
HZN-1202: Move non-SNMP-specific CollectionAgent classes to new project

soleger:
This PR moves `DefaultCollectionAgentFactory`, `DefaultCollectionAgentService`, and `DefaultCollectionAgent` into a `collection/core` project. It also eliminates the dependency on `opennms-services` in the collection projects (so that `opennms-services` depends on the collection projects instead). This dependency change meant that `HTTPCollectorIT` had to move back to `opennms-services`.

* JIRA: http://issues.opennms.org/browse/HZN-1202


j-white:
Can we also remove the reference to `opennms-services` in `opennms/features/telemetry/daemon/pom.xml` with the changes?

soleger:
I just checked and I've deleted that reference in later changes in my own repo but it's not in this changeset. That will be done in some later commits.

PR #1815, opened 2018-01-09
NMS-9816: Add Disk IO monitoring capabilities with Net-SNMP

indigo423:
Enable disk IO metrics for read/writes in bytes and operations in default system definition to get collected by default. Added metrics for disk load in percent for 1, 5 and 15-minute averages and created RRD graph definitions for each value as percentage graph and an aggregated graph with all three values similar to a load average graph.

Added 64-bit counters for reading and write bytes and if they are supported, suppress the graph with the 32-bit counter.

* JIRA: http://issues.opennms.org/browse/NMS-9816

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/datacollection/netsnmp.xml
+++ b/opennms-base-assembly/src/main/filtered/etc/datacollection/netsnmp.xml
@@ -1,219 +1,223 @@
 <datacollection-group xmlns="http://xmlns.opennms.org/xsd/config/datacollection" name="Net-SNMP">
-   <resourceType name="diskIOIndex" label="Disk IO (UCD-SNMP MIB)" resourceLabel="${diskIODevice} (index ${index})">
-      <persistenceSelectorStrategy class="org.opennms.netmgt.collectd.PersistRegexSelectorStrategy">
-         <parameter key="match-expression" value="not(#diskIODevice matches '^(loop|ram).*')"/>
-      </persistenceSelectorStrategy>
-      <storageStrategy class="org.opennms.netmgt.dao.support.SiblingColumnStorageStrategy">
-         <parameter key="sibling-column-name" value="diskIODevice"/>
-         <parameter key="replace-all" value="s/^-//"/>
-         <parameter key="replace-all" value="s/\s//"/>
-         <parameter key="replace-all" value="s/:\\.*//"/>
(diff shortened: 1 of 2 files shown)


pbrane:
These are really nice configurations!  Do these metrics require changes to NetSNMP configuration?

indigo423:
@pbrane no changes required, they are provided by default.

j-white:
Looks good to me, but I'd opt for targeting develop instead of foundation-2017 since:
A) We're modifying an existing group
B) There's the potential to start collecting *alot* more metrics

indigo423:
@j-white no problem with retargeting to develop, just picked foundation with improving support in mind. I'll fix this.

indigo423:
Changed base to develop, I also noticed the whole file has a 3 space indentation. If we decide to not not fix this problem I can revert the separate commit https://github.com/OpenNMS/opennms/pull/1815/commits/7743342e8f2b900da1b83bd757d895d394e39312

indigo423:
@RangerRick do you know why Bamboo is not running on this branch?

indigo423:
@j-white Just looked at the issue NMS-9789 was a little bit confused, cause the PR in the issue points to something with vaccumd cleanup config.

This PR makes #1812 obsolete.

indigo423:
@mfuhrmann @j-white for sanity reasons I suggest two options:

* Close NMS-9816 and move everything to NMS-9789
* Do the other way around

I can fix this just let me know which way you prefer.

indigo423:
A good point raised in the other PR, the new metrics are added to a new group to prevent breaking `storeByGroup` configs, what do you think?

indigo423:
I have picked the changes from @mfuhrmann and added them to this PR which was easier. We can close #1812 and the related issue.

indigo423:
The build failed with building some smoke test docker images :/ https://bamboo.opennms.org/browse/OPENNMS-ONMS1993-SMOKE-2

indigo423:
@RangerRick for some reasons the docker build in the smoke tests fail :/

RangerRick:
Trying re-running it now.  Might have just hit a temporary yum outage or something.

j-white:
The formatting changes seem out of place here. If there's an issue with the 3 space indentation then let's address that separately since many other files are now formatted this way.

j-white:
👍

PR #1816, opened 2018-01-11
NMS-9777: Geographical maps refresh tweaks

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9777

Here we:
1) Update the default refresh rate for the geo maps from 5s to 30s and allow it to be configured via the `org.opennms.features.nodemaps.refresh` system property.
2) Avoid updating the client state after a refresh triggered if the list of nodes was not changed, avoiding unnecessary re-rendering on the client side.


RangerRick:
Looks good to me!  👍

PR #1817, opened 2018-01-11
HZN-1196: Enable Karaf IT that bootstraps our modified container

soleger:
This enables a Karaf IT that bootstraps the `container/karaf` container with our modifications instead of a vanilla Karaf container. A lot of the bundles end up in `Failed` or `GracePeriod` blueprint state because the underlaying OpenNMS services are not started but this simple test will still verify that the system classpath and OSGi classpath dependencies are correct in our modified container.

* JIRA: http://issues.opennms.org/browse/HZN-1196


j-white:
Closing since the test that is added by this PR doesn't currently pass.

PR #1820, opened 2018-01-15
HZN-1223: Include the Drift plugin when running Elasticsearch in our integration tests.

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1223

Here we enhance the Elasticsearch server used in our integration tests to support loading plugins specified by the test, and update the flow module to use the Drift plugin.

We currently use snapshot artifacts for the plugin, but we'll need to issue a tagged release before merging this to develop.


soleger:
Looks good to me.

mvrueden:
LGTM as well. The only thing I'd like to point out is, that we here switch to elastic version 6.x, but when running the smoke tests we use 5.x, which we might want to adapt as well.

PR #1824, opened 2018-01-17
HZN-1161: Improve check of isValid(String, String)

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1161

### DIFF
--- a/features/telemetry/adapters/collection/src/main/java/org/opennms/netmgt/telemetry/adapters/collection/AbstractPersistingAdapter.java
+++ b/features/telemetry/adapters/collection/src/main/java/org/opennms/netmgt/telemetry/adapters/collection/AbstractPersistingAdapter.java
@@ -28,6 +28,13 @@
 
 package org.opennms.netmgt.telemetry.adapters.collection;
 
+import java.io.File;
+import java.util.Collections;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+
 import org.opennms.netmgt.collection.api.CollectionAgent;
 import org.opennms.netmgt.collection.api.CollectionSet;
 import org.opennms.netmgt.collection.api.Persister;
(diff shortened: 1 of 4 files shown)


features/telemetry/adapters/collection/src/main/java/org/opennms/netmgt/telemetry/adapters/collection/AbstractPersistingAdapter.java
```diff
@@ -126,17 +144,23 @@ private Package getPackageFor(Protocol protocol, CollectionAgent agent) {
...
+            if (isValid(agent.getHostAddress(), filterRule)) {
                 return pkg;
             }
         }
         return null;
     }
 
+    private boolean isValid(String hostAddress, String filterRule) {
+        try {
+            Optional<Boolean> validOptional = cache.get(new CacheKey(hostAddress, filterRule));
```
j-white: Maybe we should cache the effective package used for the agent instead of the results of the filter match? This would save us from having to iterate over all of the packages and should reduce the cache size.

opennms-dao/src/test/java/org/opennms/netmgt/dao/support/JdbcFilterDaoIT.java
```diff
@@ -353,4 +362,36 @@ public void testVariousWaysToMatchServiceNames() {
...
+        m_dao.getActiveIPAddressList("IPADDR != '0.0.0.0'");
+        long diff = System.currentTimeMillis() - start;
+
+        // measure
+        start = System.currentTimeMillis();
+        assertThat(m_dao.isValid("10.10.0.1", "IPADDR != '0.0.0.0'"), is(true));
+        long diff2 = System.currentTimeMillis() - start;
+
+        // verify
+        assertThat(diff, greaterThan(diff2));
```
j-white: Comparing timing in this fashion may cause flaps i.e. if the JVM GC pauses during the 2nd measure, then diff2 might be > diff.

I would just print out the different time stamps and include assertion for the functional bits.

features/telemetry/adapters/collection/src/main/java/org/opennms/netmgt/telemetry/adapters/collection/AbstractPersistingAdapter.java
mvrueden: Yeah, that makes sense. Also the CacheKey is missing a `equals()` and `hashCode()` implementation. Ooops.

opennms-dao/src/test/java/org/opennms/netmgt/dao/support/JdbcFilterDaoIT.java
mvrueden: Then we can get rid of the test as it does not make sense to have it (:

PR #1825, opened 2018-01-17
Cleanup: Remove unused javascript test files

mvrueden:
Here we remove all javascript smoke-tests.


j-white:
RIP Casper

indigo423:
+1 for every commit which deletes more then it adds 💃

PR #1827, opened 2018-01-19
HZN-1203: Split experimental features out into separate repo

soleger:
The PR creates a `features-experimental.xml` file for Karaf features that are not used in the product because they are not yet needed or do not load correctly due to OSGi dependency issues. These features may be useful at some point, for instance for running daemon processes or the webapp inside Karaf. I'm splitting them out so that people won't be confused by unsupported features inside the main `features.xml` file.

Please do not delete the branch, I'm using it for other work that's in progress.

* JIRA: http://issues.opennms.org/browse/HZN-1203


j-white:
+1

PR #1828, opened 2018-01-19
HZN-1224: Flow REST API for deep dive tooling

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1224

Here we beef up the Flow REST API so that it support improved filtering capabilities and generates more accurate results.


RangerRick:
It's a lot of changes but I looked through everything and nothing jumps out at me, I'll go ahead and merge.

PR #1829, opened 2018-01-20
NMS-9684: Refactor compatibility for current version

indigo423:
Remove version compatibility for old versions and give the user the information which software is compatible with the version the documentation is released with.

* JIRA: http://issues.opennms.org/browse/NMS-9684


opennms-doc/guide-install/src/asciidoc/text/opennms/compatibility.adoc
```diff
@@ -0,0 +1,13 @@
+[[compatibility]]
+== Compatibility
+
+This version of {opennms-product-name} {opennms-version} is compatible with
+
+* Java: 8
+* PostgreSQL: 9.3+
+* Newt: 1.4.3
```
j-white: Newt vs Newts

j-white:
That was quick :)

PR #1831, opened 2018-01-22
HZN-1202: Fixed CollectCommand by always using SnmpCollectionAgents

soleger:
CollectCommand will use a SnmpCollectionAgentFactory to construct
SnmpCollectionAgents so that the agent that is created is compatible
with all collectors.

* JIRA: http://issues.opennms.org/browse/HZN-1202

### DIFF
--- a/features/collection/shell-commands/pom.xml
+++ b/features/collection/shell-commands/pom.xml
@@ -47,6 +47,10 @@
       <artifactId>opennms-dao-api</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.opennms</groupId>
+      <artifactId>opennms-services</artifactId>
+    </dependency>
+    <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
     </dependency>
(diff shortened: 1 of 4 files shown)


j-white:
Can we keep this logic outside of the shell command? Seems out of place to have to depend on opennms-services and bring all this logic in here to create the collection agents.

Can we just have opennms-services export the implementation of the `CollectionAgentFactory`?

soleger:
This is the only place where we use the `CollectionAgentFactory` API and it is required to return `SnmpCollectionAgent` instances. It _would_ be required inside Collectd's SnmpCollector, but it doesn't use the `CollectionAgentFactory` API. I think this is the right way to do things for now, with the long-term solution being to refactor the SnmpCollector to be more like the rest of the collectors. It should use the `CollectionAgentFactory` API and decouple its caching of SNMP attributes away from the specialized `SnmpCollectionAgent` impl.

j-white:
Even if this is the only place it is used, I don't think the `CollectCommand` should be responsible for  building the `CollectionAgent` implementations (hence the `CollectionAgentFactory`) or have to depend directly on `opennms-services`.

soleger:
You have to have SnmpCollectionAgents to work with SnmpCollector so you either need a) an explicit dependency on opennms-services or b) a side-effect-y dependency on a specific implementation of CollectionAgentFactory that is only satisfied by opennms-services anyway. I feel like (a) follows the principal of least surprise for now, until there's a chance to refactor SnmpCollector to behave more like the other collectors.

j-white:
Prior to the refactor, the `CollectCommand` used the implementation of the `CollectionAgentFactory` pulled from the service registry to create the `CollectionAgent`s when some node criteria was specified. No surprises, and clean separation of concerns.

Now, we're pulling the DAOs and logic previously implemented and abstracted by the  `CollectionAgentFactory` into this specific Karaf shell command. I don't see how this is better.

j-white:
I created a separate issue for this regression. Following up in https://issues.opennms.org/browse/NMS-9885.

PR #1840, opened 2018-01-29
NMS-9835: Avoid unecessary AlarmDao calls when no business rules are defined.

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9835

This will reduce overhead on systems that have bsmd enabled (by default) but do not have any business rules defined.


mvrueden:
Imagine the following scenarios
 * No business services during startup. Adding a new business service. Polling starts.
 * Business services during startup. Polling starts. Deleting all business services. Polling does not stop.

We should also ensure that it stops polling when we deleted the last business service, just to be consistent.

PR #1842, opened 2018-02-01
NMS-9833: fix startup to make sure $MINION_HOME/data exists

RangerRick:
This PR fixes the Minion init to make sure `$MINION_HOME/data` exists.

It also changes `find -print` calls to use `-print0` to make sure there aren't issues with spaces in filenames.  Shouldn't affect anything, but it's best practice when shell scripting.  :)

* JIRA: http://issues.opennms.org/browse/NMS-9833


mvrueden:
@RangerRick a rebase against `foundation-2017` is required in order to have the tests pass.

PR #1843, opened 2018-02-01
NMS-9702: Sanitize resource name inside storage strategy, rather than…

roskens:
… on resource.

NMS-9702: GenericResourceType is altering index names on class object initialization

* JIRA: http://issues.opennms.org/browse/NMS-9702

### DIFF
--- a/features/collection/api/src/main/java/org/opennms/netmgt/collection/support/IndexStorageStrategy.java
+++ b/features/collection/api/src/main/java/org/opennms/netmgt/collection/support/IndexStorageStrategy.java
@@ -1,8 +1,8 @@
 /*******************************************************************************
  * This file is part of OpenNMS(R).
  *
- * Copyright (C) 2006-2014 The OpenNMS Group, Inc.
- * OpenNMS(R) is Copyright (C) 1999-2014 The OpenNMS Group, Inc.
+ * Copyright (C) 2006-2018 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2018 The OpenNMS Group, Inc.
  *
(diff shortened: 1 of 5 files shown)


j-white:
Can we add a test that verifies that the `ObjectNameStorageStrategy` functions properly when initiated via the `GenericResourceType`? This will help capture the requirement and make sure that similar problems don't arise again.

features/collection/api/src/main/java/org/opennms/netmgt/collection/support/IndexStorageStrategy.java
```diff
@@ -65,7 +66,7 @@ public final String getResourceTypeName() {
     @Override
     public String getResourceNameFromIndex(CollectionResource resource) {
         // Use the instance value as the name of the resource
-        return resource.getInstance();
+        return GenericTypeResource.sanitizeInstance(resource.getInstance());
```
j-white: One concern here is that the various strategies now need to explicitly sanitize the instance, making it easy to forget or omit when creating new strategies.

Are there any other places where we need to add calls to `sanitizeInstance`?

j-white: Yeah, it's not easy to do without verifying all of the strategies individually. I agree that it's not ideal but moving that to the persistence layer is a larger effort.

I think a less intrusive approach to solving this would be to make the unsanitized instance available as a new getter and update to the `ObjectNameStorageStrategy` to use that. This would leave the existing code path untouched and maintain the current behavior.

mvrueden:
This PR did not get updated for almost a year. I am going to close this without merging. Please feel free to re-open if you need further assistance or still wish to contribute this. Thank you.

PR #1845, opened 2018-02-02
NMS-9699: Fix basic authentication with Elastisearch.

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9699

Here we update the `RestClientFactory` to preserve basic authentication credentials when set.


mvrueden:
The provided test is testing if basic auth is working in general, but it does not verify if pre-emptive auth works. Can we add a test for that as well?

j-white:
Sure does, the test will fail if the client does not perform pre-emptive auth.

PR #1846, opened 2018-02-02
NMS-9846: Improve event processing rate in alarmd

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9846

This PR contains a number of fixes aimed at improving the processing speed of alarmd:
 * Process events in parallel while use a fair striped lock on both the reduction and clear keys to continue processing events in order.
 * Avoid further processing required by northbounder (NBIs) if none are enabled.

On my system, I was able to double the rate at which events were processed with these changes. The `events:stress` command has been update with an example showing how `alarmd` can be benchmarked.


RangerRick:
This looks good to me, but I don't know this code as well as @soleger so maybe he should take a look as well.  Otherwise, +1 to merging.

PR #1849, opened 2018-02-04
HZN-1237: Improve classification and gather more statistics

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1237

### DIFF
--- a/features/flows/classification/engine/src/main/java/org/opennms/netmgt/flows/classification/DefaultClassificationEngine.java
+++ b/features/flows/classification/engine/src/main/java/org/opennms/netmgt/flows/classification/DefaultClassificationEngine.java
@@ -31,9 +31,7 @@ package org.opennms.netmgt.flows.classification;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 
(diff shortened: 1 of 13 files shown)


features/flows/classification/engine/src/main/java/org/opennms/netmgt/flows/classification/DefaultClassificationEngine.java
```diff
@@ -106,18 +96,32 @@ public String classify(ClassificationRequest classificationRequest) {
...
+                classifierPortMap.get(null).add(classifier);
+            }
         }
-        return new CombinedClassifier(rule.getName(), matchers);
     }
 
+    private List<Classifier> getClassifiers(ClassificationRequest request) {
+        // Get classifiers for the given port
+        final List<Classifier> classifiers = classifierPortMap.getOrDefault(request.getPort(), new ArrayList<>());
+        // Add all classifiers, which match to ANY port
```
j-white: Can we not add these rules to the map and sort them ahead of time? This will minimize the work we need to perform on lookup at the cost of requiring memory to hold the additional references.

features/flows/classification/engine/src/main/java/org/opennms/netmgt/flows/classification/DefaultClassificationEngine.java
```diff
@@ -31,27 +31,26 @@
...
+import org.opennms.netmgt.flows.classification.value.PortValue;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Strings;
 
 public class DefaultClassificationEngine implements ClassificationEngine {
 
     private final ClassificationRuleProvider ruleProvider;
-    private final List<Classifier> classifierList = new ArrayList<>();
+    private final Map<Integer, List<Classifier>> classifierPortMap = new HashMap<>();
```
j-white: Since our keys are integers and there are a fixed number of ports (2^16), we could also use an ArrayList here - this could help reduce the lookup time.

PR #1853, opened 2018-02-06
NMS-9761: rework of JavaScript and CSS Assets

RangerRick:
This is a complete rework of all of our JavaScript and CSS assets to make it easier to do new development while still handling our old creaky random JavaScript.

The project is capable of handling ES6 JavaScript as well as TypeScript.

I have touched *everything* having to do with JavaScript, and it's pretty darn messy, so I would love some folks to try it out and see if they can break anything.  It is extremely likely I missed something.

* JIRA: http://issues.opennms.org/browse/NMS-9761


j-white:
This is a HUUUGE patch. I can't think of any logical way to stage it though - I think it's going to have to stay one huge chunk.

I haven't found any problems with any of the UI components yet. Everything loads properly and feels much more responsive then it was previously.

I did notice:
* CollectableService.java was updated with what appears to be an unrelated change
* `features/themes/onms-default-theme/src/main/java/` shows up with untracked files when doing a `git status` after a clean build

j-white:
Since the commits is so large, could you please avoid squashing from now on in order to make it easier to track the changes.

RangerRick:
@j-white Oop, sorry, didn't see your comment until after I already pushed this.  I'll do individual commits going forward.

The CollectableService thing was from back when Seth had broken SNMP tests.

I can't reproduce the `onms-default-theme` thing. Are you sure it wasn't there from before you switched branches? `develop` currently does weird things with changing `src/main/java/` during the build, but the branch version doesn't even contain that directory anymore.

j-white:
Turns out the `features/themes/onms-default-theme/src/main/java/` output in `git status` was there before I ran a clean and compile in this branch. All is good.

RangerRick:
Any objections from anyone to merging?  I'd love to see it get out just to have some people banging on it.  Also if @mvrueden wants help porting his drift JS bits to the new structure it'll have to be this week.

PR #1855, opened 2018-02-07
HZN-1177: Add some helper methods to get metrics for gpb

cgorantla:
Given a metric name and Telemetry message, get it's value by parsing through message.

* JIRA: http://issues.opennms.org/browse/hzn-1177


agalue:
+1 Looks good to me, at least from Groovy's perspective. I'll trigger my AWS lab against my virtual NXOS Switch to see how the data looks like with several different sensor-path sources.

agalue:
Speaking of things that helps, what do you think about adding a DEBUG or TRACE message on `NxosGpbAdapter` with the "text" version of the telemetry data after parsing the GPB content (i.e. `TelemetryBis.Telemetry.toString()`), to help administrators see that's inside of it, so they can write the Groovy script ?

cgorantla:
@agalue Thanks Alejandro. Yeah, would like to see if some variations of data for which this parser fails. We may want to fix that as well.

pbrane:
I was looking at this:
... findFieldWithName(field, name);

Wouldn't it be better to have something more like:
   subfield = field.get("abc");

as well as abstracting the the message into a interface:

Interface org.opennms.telemetry.TelemetryMessage {}

That way whatever message format is passed into the adapter (gpb or json) the code can just reference the API and parse the fields:

TelemetryBis.Telemetry telemetryMsg = msg;
   Field msg.getField("123");

or

   List<Field> msg.getFields();

and we would just need specific implementations each encoding.  Then, only one adapter could handle both encodings.  I think putting the parsing algorithms in the class and not in the adapter would better.

cgorantla:
@pbrane  TelemetryBis.Telemetry is GPB generated class. So we can't modify it.  So all the methods this utility exposes use  GPB generated classes/methods so we can't easily generalize for all different encodings. 

So we may need to embed both messages ( json/gpb)  into another class and depend on the encoding type to determine which one to use.  It adds another layer for messages/parsing.

@agalue  will give me json data.  I will explore more on this.

agalue:
I'll be testing the solution today on my AWS Lab and my Virtual Nexus 9K Switch. In the meantime, I'd like to share some thoughts, to improve the content of the example Groovy script:

I think we should include more details on the Groovy script to improve performance and be more specific about the meaning of all that code, in order to provide a better background to the final users about how to use this feature.

First of all, I would put a comment at the top of the code section (before the node-level plus generic-index resource table) about what is required at Nexus Switch side in order to use the code in question, as the example heavily relies on that config, and only works with that config. The comment should be something like this:

```
The following code is required in order to process telemetry data from a Nexus switch configured like this:

telemetry
  destination-group 100
    ip address 192.168.205.253 port 50001 protocol UDP encoding GPB 
  sensor-group 200
    data-source NX-API
    path "show system resources" depth 0
  subscription 300
    dst-grp 100
    snsr-grp 200 sample-interval 300000
```

The above comment also reveals what we have to do in terms of performance improvements, which is, validate the sensor-group path, as due to UDP limitations, the switch will have hundreds if not thousands of sensor-groups and/or paths inside a sensor-group to achieve customer's expectations; for example, retrieve the statistics from each physical interface (and a Nexus 9K could have hundreds of physical interfaces with multiple metrics per interface, which requires partition the data into thousands of sensor-grou/path combinations in order to fit the UDP maximum packet limit).

In GPB world, that means checking `TelemetryBis.Telemetry.getEncodingPath()`, which will contains the path of the sensor-group (in this case "show system resources"), otherwise the helper class will be doing its nested loop checking across all the sensor-group/path combinations to find the correct data which is not an optimal way to process the data.

So, the code should look like this:

```JAVA
    static generate(agent, builder, telemetryMsg) {
        log.debug("Generating collection set for message: {}", telemetryMsg)

        # The following code is required in order to process telemetry data from a Nexus switch configured like this:
        # 
        # telemetry
        #   destination-group 100
        #     ip address 192.168.205.253 port 50001 protocol UDP encoding GPB 
        #   sensor-group 200
        #     data-source NX-API
        #     path "show system resources" depth 0
        #   subscription 300
        #     dst-grp 100
        #     snsr-grp 200 sample-interval 300000
        # 
        if (telemetryMsg.getEncodingPath().equals("show system resources")) {
            NodeLevelResource nodeLevelResource = new NodeLevelResource(agent.getNodeId())
            builder.withNumericAttribute(nodeLevelResource, "stats", "load_avg_1min",
                    NxosGpbParserUtil.getValueAsDouble(telemetryMsg, "load_avg_1min"), AttributeType.GAUGE)

           for(TelemetryBis.TelemetryField row : NxosGpbParserUtil.getRowsFromTable(telemetryMsg, "cpu_usage")) {
                String cpuId = NxosGpbParserUtil.getValueFromRowAsString(row, "cpuid");
                DeferredGenericTypeResource genericTypeResource = new DeferredGenericTypeResource(nodeLevelResource, "nxosCpu", cpuId);
                builder.withNumericAttribute(genericTypeResource, "stats", "kernel",
                        NxosGpbParserUtil.getValueFromRowAsDouble(row, "kernel"), AttributeType.GAUGE)
            }
        }
    }
```

That way, the above code will be executed only when it makes sense to do it, and clearly indicates its meaning and intention.

Naturally, NX-API sources are not the only way to get data. In fact, Cisco do not recommend to use NX-API unless it is strictly necessary due to the performance penalties for doing it on the switch. They recommend a non intuitive way which involves using DME DNs (which requires knowledge about the Visore Tool to figure out what paths are required). I'll work on test scenarios for this soon and share them here.

agalue:
I tested it and it works!

Well, partially ;)

I can see the node level metric, but I cannot see the tabular data. I'll review the content of the Groovy script, and compare it with my working JSON example to see what's missing there.

agalue:
I found the problem with generic resources, the following returns an empty string:
```
NxosGpbParserUtil.getValueFromRowAsString(row, "cpuid")
```
So, that method somehow is not working. Because it returns an empty string, the generic resource is not properly built, but an empty string seems to be valid as a resource identifier in terms of persistence which is incorrect in my opinion. Either way, that's why we don't have data.

agalue:
Because BGP enforce value types (which is not the case when using JSON), you cannot get that data as string, you have to get it as it is (numeric), and then convert it; for example:

```
Integer.toString(NxosGpbParserUtil.getValueFromRowAsDouble(row, "cpuid").intValue()))
```

I know, it is horrible, but that's how GPB provides ;)

agalue:
This is what I ended up having on Groovy, to have the solution working (and see data on Cassandra):

```GROOVY
    static generate(agent, builder, telemetryMsg) {
        log.debug("Generating collection set for node {} from message: {}", agent.getNodeId(), telemetryMsg)
        // build the node-level resource

        def nodeLevelResource = new NodeLevelResource(agent.getNodeId())

        /*
         * The following code is required in order to process telemetry data from a Nexus switch configured like this:
         * 
         * telemetry
         *   destination-group 100
         *     ip address 192.168.205.253 port 50001 protocol UDP encoding GPB 
         *   sensor-group 200
         *     data-source NX-API
         *     path "show system resources" depth 0
         *   subscription 300
         *     dst-grp 100
         *     snsr-grp 200 sample-interval 300000
         */
        if (telemetryMsg.getEncodingPath().equals("show system resources")) {
          builder.withNumericAttribute(nodeLevelResource, "nxos-stats", "load_avg_1min",
            NxosGpbParserUtil.getValueAsDouble(telemetryMsg, "load_avg_1min"), AttributeType.GAUGE)
          builder.withNumericAttribute(nodeLevelResource, "nxos-stats", "memory_usage_used",
            NxosGpbParserUtil.getValueAsDouble(telemetryMsg, "memory_usage_used"), AttributeType.GAUGE)

          NxosGpbParserUtil.getRowsFromTable(telemetryMsg, "cpu_usage").each { row ->
            def cpuId = Integer.toString(NxosGpbParserUtil.getValueFromRowAsDouble(row, "cpuid").intValue())
            def genericTypeResource = new DeferredGenericTypeResource(nodeLevelResource, "nxosCpu", cpuId)
            ["idle", "kernel", "user"].each { metric ->
                builder.withNumericAttribute(genericTypeResource, "nxos-cpu-stats", metric,
                  NxosGpbParserUtil.getValueFromRowAsDouble(row, metric), AttributeType.GAUGE)
            }
          }
        } else {
          log.warn("Received data is not 'show system resources' it is {}", telemetryMsg.getEncodingPath())
        }
    }
```

I've enhanced it a little be to be more "Groovy" like, instead of more "Java" like, and added more metrics. I'll work on the DME example now.

features/telemetry/adapters/nxos/src/main/java/org/opennms/netmgt/telemetry/adapters/nxos/NxosGpbParserUtil.java
```diff
@@ -0,0 +1,192 @@
...
+     * @return  value field value
+     */
+    public static String getValueAsString(TelemetryBis.Telemetry msg, String name) {
+        for (TelemetryBis.TelemetryField field : msg.getDataGpbkvList()) {
+            TelemetryBis.TelemetryField subField = findFieldWithName(field, name);
+            if (subField != null) {
+                return subField.getStringValue();
+            }
+        }
+        return "";
```
j-white: We may be better off to return `null` instead of an empty string here.

features/telemetry/adapters/nxos/src/main/java/org/opennms/netmgt/telemetry/adapters/nxos/NxosGpbParserUtil.java
```diff
@@ -0,0 +1,192 @@
...
+     * 
+     * @param   msg  {@link org.opennms.netmgt.telemetry.adapters.nxos.proto.TelemetryBis.Telemetry} message
+     * @param   name  field name
+     * @return  value field value
+     */
+    public static String getValueAsString(TelemetryBis.Telemetry msg, String name) {
+        for (TelemetryBis.TelemetryField field : msg.getDataGpbkvList()) {
+            TelemetryBis.TelemetryField subField = findFieldWithName(field, name);
+            if (subField != null) {
+                return subField.getStringValue();
```
j-white: We should cast the actual value to a string here instead - similar to what we did the the double above.

features/telemetry/adapters/nxos/src/main/java/org/opennms/netmgt/telemetry/adapters/nxos/NxosGpbParserUtil.java
```diff
@@ -0,0 +1,192 @@
...
+        Objects.requireNonNull(msg);
+        for (TelemetryBis.TelemetryField field : msg.getDataGpbkvList()) {
+            // ROWS are arrays for a set of metrics, Assumption is that they
+            // start with ROW_
+            TelemetryBis.TelemetryField subField = findFieldWithName(field, "ROW_" + name);
+            if (subField != null) {
+                return subField.getFieldsList();
+            }
+        }
+        return null;
```
j-white: Maybe return an empty list here instead of null?

features/telemetry/adapters/nxos/src/main/java/org/opennms/netmgt/telemetry/adapters/nxos/NxosGpbParserUtil.java
```diff
@@ -0,0 +1,192 @@
...
+     * @return  value field value
+     */
+    public static String getValueFromRowAsString(TelemetryBis.TelemetryField row, String name) {
+        for (TelemetryBis.TelemetryField nestedField : row.getFieldsList()) {
+            TelemetryBis.TelemetryField subField = findFieldWithName(nestedField, name);
+            if (subField != null) {
+                return subField.getStringValue();
+            }
+        }
+        return "";
```
j-white: If we change the default value of `getValueAsString` we should change this one too.

features/telemetry/adapters/nxos/src/main/java/org/opennms/netmgt/telemetry/adapters/nxos/NxosGpbParserUtil.java
```diff
@@ -0,0 +1,213 @@
...
+
+    /**
+     * Given a Telemetry Message and row name, get the list of fields
+     * 
+     * @param   Telemetry message {@link org.opennms.netmgt.telemetry.adapters.nxos.proto.TelemetryBis.Telemetry}
+     * @param   name of field
+     * @return  list of Telemetry fields
+     */
+    public static List<TelemetryBis.TelemetryField> getRowsFromTable(Telemetry msg, String name) {
+        Objects.requireNonNull(msg);
```
j-white: Is there any reason we check for `null` here, but nowhere else?

features/telemetry/adapters/nxos/src/main/java/org/opennms/netmgt/telemetry/adapters/nxos/NxosGpbParserUtil.java
```diff
@@ -0,0 +1,213 @@
...
+        for (TelemetryBis.TelemetryField field : msg.getDataGpbkvList()) {
+            // ROWS are arrays for a set of metrics, Assumption is that they
+            // start with ROW_
+            TelemetryBis.TelemetryField subField = findFieldWithName(field, "ROW_" + name);
+            if (subField != null) {
+                return subField.getFieldsList();
+            }
+        }
+        return new ArrayList<>();
```
j-white: Using `Collections.emptyList()` would be more efficient here - since we won't actually be adding any objects to the list.

agalue:
Here is a very simple example for DME data (unfortunately I haven't found a good example with tabular data that fits a single UDP packet):

```GROOVY
        /*
         * The following code is required in order to process telemetry data from a Nexus switch configured like this:
         * 
         * telemetry
         *   destination-group 100
         *     ip address 192.168.205.253 port 50001 protocol UDP encoding GPB 
         *   sensor-group 210
         *     path sys/intf/phys-[eth1/1]/dbgIfHCIn depth 0
         *     path sys/intf/phys-[eth1/1]/dbgIfHCOut depth 0
         *     path sys/intf/phys-[eth1/2]/dbgIfHCIn depth 0
         *     path sys/intf/phys-[eth1/2]/dbgIfHCOut depth 0
         *   subscription 300
         *     dst-grp 100
         *     snsr-grp 210 sample-interval 300000
         */
        def m;
        if ((m = telemetryMsg.getEncodingPath() =~ /sys\/intf\/phys-\[(.+)\]\/dbgIfHC(In|Out)/)) {
            def intfId = m.group(1).replaceAll(/\//,"-")
            def statsType = m.group(2)
            def genericTypeResource = new DeferredGenericTypeResource(nodeLevelResource, "nxosIntf", intfId)
            ["ucastPkts", "multicastPkts", "broadcastPkts", "octets"].each { metric ->
                builder.withNumericAttribute(genericTypeResource, "nxos-intfHC$statsType", "$metric$statsType",
                    NxosGpbParserUtil.getValueAsDouble(telemetryMsg, metric), AttributeType.GAUGE)
            }
        }
```

Of course, I've created `/opt/opennms/etc/resource-types.d/nxos-intf-resource.xml` with a definition for the resource type called `nxosIntf `.

The reason why I'm sharing this here is to explain the hack I've used when choosing the resource-group and the metric-name, in order to avoid issues on the persistence layer, and guarrantee that all the data is persisted properly. Without those tricks, the data was not persisted properly.

On the standard collectors implemented by Collectd, there is going to be a single CollectionSet per node with the whole content of all the tables (and node level data). Now, we’re going to have hundreds (if not thousands) of CollectionSets with partial set of metrics from a single row of a table, due to the restrictions imposed by using UDP as the transport protocol for telemetry data on NXOS.

Because of these restrictions, considering how storeByGroup works, and based on my tests on real-world tests, the generic index resources should be built like that.

That's why I'm basically having a group of metrics per GPB packet, an then I have to make sure the metric names are also different; otherwise I was not able to see the data correctly stored on Cassandra.

Thanks to these tricks, here is how it looks like:

```
[root@opennms ~]# /opt/opennms/bin/resourcecli list
node[Minions:a013fb36-02c9-11e8-a005-525400cae48b]
  node[Minions:a013fb36-02c9-11e8-a005-525400cae48b].interfaceSnmp[jmx-minion]
  node[Minions:a013fb36-02c9-11e8-a005-525400cae48b].responseTime[127.0.0.1]
node[Cisco:nexus9k]
  node[Cisco:nexus9k].nodeSnmp[]
  node[Cisco:nexus9k].interfaceSnmp[Ethernet1_2-00a3db0e5709]
  node[Cisco:nexus9k].interfaceSnmp[Ethernet1_1-00a3db0e5708]
  node[Cisco:nexus9k].responseTime[192.168.205.254]
  node[Cisco:nexus9k].nxosCpu[0]
  node[Cisco:nexus9k].nxosCpu[1]
  node[Cisco:nexus9k].nxosIntf[eth1-1]
  node[Cisco:nexus9k].nxosIntf[eth1-2]
```

Thanks to the trick I've explained above, I now have:

```
[root@opennms ~]# /opt/opennms/bin/resourcecli show node[Cisco:nexus9k].nxosIntf[eth1-1]
ID:         node[Cisco:nexus9k].nxosIntf[eth1-1]
Name:       eth1-1
Label:      eth1-1
Type:       Nxos Interface
Link:       null
Parent ID:  node[Cisco:nexus9k]
Children:
Attributes:
  External:
  Metrics:
    multicastPktsOut = 'snmp:fs:Cisco:nexus9k:nxosIntf:eth1-1:nxos-intfHCOut'
    ucastPktsIn = 'snmp:fs:Cisco:nexus9k:nxosIntf:eth1-1:nxos-intfHCIn'
    broadcastPktsIn = 'snmp:fs:Cisco:nexus9k:nxosIntf:eth1-1:nxos-intfHCIn'
    octetsIn = 'snmp:fs:Cisco:nexus9k:nxosIntf:eth1-1:nxos-intfHCIn'
    ucastPktsOut = 'snmp:fs:Cisco:nexus9k:nxosIntf:eth1-1:nxos-intfHCOut'
    multicastPktsIn = 'snmp:fs:Cisco:nexus9k:nxosIntf:eth1-1:nxos-intfHCIn'
    broadcastPktsOut = 'snmp:fs:Cisco:nexus9k:nxosIntf:eth1-1:nxos-intfHCOut'
    octetsOut = 'snmp:fs:Cisco:nexus9k:nxosIntf:eth1-1:nxos-intfHCOut'
  Strings:
```

The above was not the case if I not enforce the groups/metrics to be unique.

I guess that's the best we can do when using UDP, as the only way to get the whole content of tabular resources is by using either HTTP or gRPC for the transport protocol (regardless the encoding protocol).

features/telemetry/adapters/nxos/src/main/java/org/opennms/netmgt/telemetry/adapters/nxos/NxosGpbParserUtil.java
cgorantla: I will remove and will just return null if any of params are null instead of throwing exception

j-white:
I'm going to go ahead and merge this since the issues and comments related to the helper functions have been addressed.

To address @pbrane's point, we could look into further abstractions if we find ourselves needing/wanting to add JSON support, thereby allowing users to use the same scripts in either case.

Enhancements and updates to the Groovy scripts themselves can be handled separately.

PR #1821, opened 2018-01-16
HZN-1208: Introduce Classification Rest API

mvrueden:
* Jira: https://issues.opennms.org/browse/HZN-1208

PR #1857, opened 2018-02-08
HZN-1236: Provide documentation for the Classification Rest Endpoint

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1236

Should be merged after PR https://github.com/OpenNMS/opennms/pull/1821.

Contains a very rough documentation of the Classification Rest Endpoint.


j-white:
The content looks good.

Can we add a note here that makes it clear we're referring to classification rules for flows (and not some other component in OpenNMS)?

PR #1858, opened 2018-02-08
NMS-9783: Include X-Requested-With header for all  requests

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-9783

The browser popup is now omitted, but no further actions are made.
We may need to add some kind of information/modal/popup or forward to the login.jsp in addition to this change to have a smooth user experience.


j-white:
I guess we should wait for https://github.com/OpenNMS/opennms/pull/1853 before merging this.

mvrueden:
Yes, that makes sense. We should also decide what to do in a `401 Unauthorized`. I'd say we forward to the login.jsp, as of now, the only reason jetty would return a 401 is when the session expired (or is unknown, e.g. due to server restart)

j-white:
Closing pending merge conflict changes.

PR #1859, opened 2018-02-09
NMS-9829: Allow any external jdbc drivers to load in jdbc detector/poller/collector

cgorantla:
Adding dynamic import for the bundles allow any other drivers to load for that bundle. 
Also make postgres as default in DBTools.

* JIRA: http://issues.opennms.org/browse/NMS-9829

### DIFF
--- a/core/lib/src/main/java/org/opennms/core/utils/DBTools.java
+++ b/core/lib/src/main/java/org/opennms/core/utils/DBTools.java
@@ -46,7 +46,7 @@ public abstract class DBTools {
      * 
      * @see #constructUrl
      */
-    private static final String JDBC_HOST = "OPENNMS_JDBC_HOSTNAME";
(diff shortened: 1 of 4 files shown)


j-white:
Thanks for addressing this. Can we split these changes up? The fix for the OSGi imports should target `release-21.0.4`, or `foundation-2017`, whereas the change of default values should target `develop`.

There are also a number of references to the current defaults in the Administrators Guide which will need to be updated as well.

j-white:
We don't need to create PRs for multiple branches with the same fix, since they propagate forward (changes made to `foundation-2017` get merged in `release-21.0.4`, and changes made in `release-21.0.4` get merged to `develop`).

j-white:
Can we rebase this over `develop` now that the other changes have merged forward?

PR #1863, opened 2018-02-14
Update the Minion ./runInPlace.sh script to support many Minions

j-white:
Here we update the `./runInPlace.sh` script to support running many Minions which is useful for testing features like the directed RPCs, or experiment with Minion clustering support.

The script contains logic for setting up the various ports so that they do no clash.


indigo423:
Just to make sure, we don't need issue and release tracking for this enhancement?

j-white:
Nope - I don't see a need - It's a dev script that's only part of the source tree.

PR #1872, opened 2018-02-21
HZN-1249: Consolidate modules which communicate to elastic search (flows, opennms-es-rest)

mvrueden:
Here we consolidate the `flows` and the `opennms-es-rest` modules in the way they talk to elastic search.

The only thing, where both implementations are not equal at the moment is, that:
 * the `opennms-es-rest` module retries any failed item in the bulk operation manually after the bulk operation failed
 * the `flows` module has a cooldown between a retry

We should have a discussion about what should be the ideal behaviour for both modules and implement it accordingly.

Also not done, but may be done in a future work is re-working the configuration, as of now each module must be configured individually in its according cfg file.

Related Jira issue: https://issues.opennms.org/browse/HZN-1249


j-white:
Can we look at reverting https://github.com/OpenNMS/opennms/commit/b29dde431618e105d47dc01c9037f420dd0dc745 here too? I think some of the original changes we made to the client broke that test.

mvrueden:
Please note, while reworking the code, I decided to drop support for Elasticsearch version 2.4 (only `opennms-es-rest` plugin may be affected). 

The reason for this is, that an UPDATE action (bulk operation) did not create the index. A separate call must be made, to create the index in order to update it. Later versions of elastic simply create the index first.

mvrueden:
The *flows*  and *opennms-es-rest* module no longer behave differently.

mvrueden:
It finally went green *yay*

mvrueden:
* The elastic properties are in general renamed, not only the url property, also the name, the timeouts, etc. 
* Besides that, the default batch size is now > 1, before it was 1, meaning no bulk insert. 
* If the bulk operation fails, we retry, but with a cooldown, before it just retried.
* In addition elastic in version 2.4 is no longer supported.

mvrueden:
I added some notes to the releasenotes to reflect the changes we made. I also resolved the merge conflict.

j-white:
👍 from me. Thanks for updating the release notes too.

PR #1873, opened 2018-02-21
NMS-9855: Fix asset record not having a node id when created via REST

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-9855


RangerRick:
\o/

PR #1874, opened 2018-02-21
HZN-1227:  Drift: Flow indicator icon on MIB-2 traffic utilization graphs

cgorantla:
Add icon whenever we have flows for the specific interface.

* JIRA: http://issues.opennms.org/browse/hzn-1227

### DIFF
--- a/core/web-assets/src/main/assets/js/apps/add-to-ksc/index.js
+++ b/core/web-assets/src/main/assets/js/apps/add-to-ksc/index.js
@@ -114,4 +114,22 @@ angular.module('onms-ksc', [
     });
   };
 
+}])
+.controller('checkFlowsCtrl', ['$scope', '$http', '$filter', function($scope, $http, $filter) {
+
+$scope.flowCount = 0;
+$scope.hasFlows = false;
+
+  $scope.getFlowCount = function(nodeId, ifIndex , start, end) {
+    $http({
+      url: 'rest/flows/count?exporterNode='+nodeId+'&ifIndex='+ifIndex+'&start='+start+'&end='+end,
+      method: 'GET',
+      params: { limit: 0 }
+    }).success(function(data) {
+      $scope.flowCount = data;
+      if ( data > 0) {
+        $scope.hasFlows = true;
+      }
+    });
+  };
 }]);
--- a/core/web-assets/src/main/assets/js/apps/add-to-ksc/index.js
+++ b/core/web-assets/src/main/assets/js/apps/add-to-ksc/index.js
@@ -115,28 +115,12 @@ angular.module('onms-ksc', [
   };
 }])
 .factory('flowsRestFactory', function ($http, $q) {
-
   var resources = {};
-  resources.retrieveFlowCount = function(nodeId, ifIndex , start, end) {
-      var deferred = $q.defer();
-      $http({
-        url: 'rest/flows/count',
-        method: 'GET',
-        params: { exporterNode : nodeId,
-                  ifIndex : ifIndex,
-                  start : start,
-                  end : end,
-                  limit: 0 }
-      }).success(function(data) {
-        deferred.resolve(data);
-      });
-      return deferred.promise;
-  };
 
-  resources.getDeepDiveToolUrl = function(nodeId, ifIndex , start, end) {
+  resources.getFlowGraphUrl = function(nodeId, ifIndex , start, end) {
       var deferred = $q.defer();
       $http({
-        url: 'rest/flows/deepDiveUrl',
+        url: 'rest/flows/flowGraphUrl',
         method: 'GET',
         params: { exporterNode : nodeId,
                   ifIndex : ifIndex,
@@ -155,25 +139,18 @@ angular.module('onms-ksc', [
 
   $scope.flowCount = 0;
   $scope.hasFlows = false;
-  $scope.deepDiveUrl = '';
-  $scope.getFlowCount = function(nodeId, ifIndex , start, end) {
+  $scope.flowGraphUrl = '';
+  $scope.getFlowInfo = function(nodeId, ifIndex , start, end) {
     if (nodeId == 0 || ifIndex == 0) {
       return;
     }
-    flowsRestFactory.retrieveFlowCount(nodeId, ifIndex, start, end)
+    flowsRestFactory.getFlowGraphUrl(nodeId, ifIndex, start, end)
       .then(function (data) {
-        $scope.flowCount = data;
+        $scope.flowGraphUrl = data.flowGraphUrl;
+        $scope.flowCount = data.flowCount;
         if ( $scope.flowCount > 0) {
-            $scope.hasFlows = true;
-            $scope.getDeepDiveToolUrl(nodeId, ifIndex, start, end);
+                 $scope.hasFlows = true;
         }
-    });
-  };
-
-  $scope.getDeepDiveToolUrl = function(nodeId, ifIndex , start, end) {
-    flowsRestFactory.getDeepDiveToolUrl(nodeId, ifIndex, start, end)
-      .then(function (data) {
-        $scope.deepDiveUrl = data;
       });
   };
 }]);
--- a/core/web-assets/src/main/assets/js/apps/add-to-ksc/index.js
+++ b/core/web-assets/src/main/assets/js/apps/add-to-ksc/index.js
@@ -122,9 +122,13 @@ $scope.hasFlows = false;
 
   $scope.getFlowCount = function(nodeId, ifIndex , start, end) {
     $http({
-      url: 'rest/flows/count?exporterNode='+nodeId+'&ifIndex='+ifIndex+'&start='+start+'&end='+end,
+      url: 'rest/flows/count',
       method: 'GET',
-      params: { limit: 0 }
+      params: { exporterNode : nodeId,
+                ifIndex : ifIndex,
+                start : start,
+                end : end,
+                limit: 0 }
     }).success(function(data) {
       $scope.flowCount = data;
       if ( data > 0) {
--- a/core/web-assets/src/main/assets/js/apps/add-to-ksc/index.js
+++ b/core/web-assets/src/main/assets/js/apps/add-to-ksc/index.js
@@ -113,27 +113,67 @@ angular.module('onms-ksc', [
       growl.error(msg);
     });
   };
-
 }])
-.controller('checkFlowsCtrl', ['$scope', '$http', '$filter', function($scope, $http, $filter) {
+.factory('flowsRestFactory', function ($http, $q) {
 
-$scope.flowCount = 0;
-$scope.hasFlows = false;
+  var resources = {};
+  resources.retrieveFlowCount = function(nodeId, ifIndex , start, end) {
+      var deferred = $q.defer();
+      $http({
+        url: 'rest/flows/count',
+        method: 'GET',
+        params: { exporterNode : nodeId,
+                  ifIndex : ifIndex,
+                  start : start,
+                  end : end,
+                  limit: 0 }
+      }).success(function(data) {
+        deferred.resolve(data);
+      });
+      return deferred.promise;
+  };
 
+  resources.getDeepDiveToolUrl = function(nodeId, ifIndex , start, end) {
+      var deferred = $q.defer();
+      $http({
+        url: 'rest/flows/deepDiveUrl',
+        method: 'GET',
+        params: { exporterNode : nodeId,
+                  ifIndex : ifIndex,
+                  start : start,
+                  end : end,
+                  limit: 0 }
+      }).success(function(data) {
+        deferred.resolve(data);
+      });
+      return deferred.promise;
+  };
+
+  return resources;
+})
+.controller('checkFlowsCtrl', ['$scope', '$http', '$filter', 'flowsRestFactory', function($scope, $http, $filter, flowsRestFactory) {
+
+  $scope.flowCount = 0;
+  $scope.hasFlows = false;
+  $scope.deepDiveUrl = '';
   $scope.getFlowCount = function(nodeId, ifIndex , start, end) {
-    $http({
-      url: 'rest/flows/count',
-      method: 'GET',
-      params: { exporterNode : nodeId,
-                ifIndex : ifIndex,
-                start : start,
-                end : end,
-                limit: 0 }
-    }).success(function(data) {
-      $scope.flowCount = data;
-      if ( data > 0) {
-        $scope.hasFlows = true;
-      }
+    if (nodeId == 0 || ifIndex == 0) {
+      return;
+    }
+    flowsRestFactory.retrieveFlowCount(nodeId, ifIndex, start, end)
+      .then(function (data) {
+        $scope.flowCount = data;
+        if ( $scope.flowCount > 0) {
+            $scope.hasFlows = true;
+            $scope.getDeepDiveToolUrl(nodeId, ifIndex, start, end);
+        }
     });
+  };
+
+  $scope.getDeepDiveToolUrl = function(nodeId, ifIndex , start, end) {
+    flowsRestFactory.getDeepDiveToolUrl(nodeId, ifIndex, start, end)
+      .then(function (data) {
+        $scope.deepDiveUrl = data;
+      });
   };
 }]);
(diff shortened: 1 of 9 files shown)


core/web-assets/src/main/assets/js/apps/add-to-ksc/index.js
```diff
@@ -114,4 +114,22 @@ angular.module('onms-ksc', [
...
 
+}])
+.controller('checkFlowsCtrl', ['$scope', '$http', '$filter', function($scope, $http, $filter) {
+
+$scope.flowCount = 0;
+$scope.hasFlows = false;
+
+  $scope.getFlowCount = function(nodeId, ifIndex , start, end) {
+    $http({
+      url: 'rest/flows/count?exporterNode='+nodeId+'&ifIndex='+ifIndex+'&start='+start+'&end='+end,
```
mvrueden: Can we use the `params` object for the parameters instead having a mix of both?

opennms-webapp/src/main/webapp/WEB-INF/jsp/graph/results.jsp
```diff
@@ -198,6 +199,29 @@
...
+             </c:if>
+             <c:if test="${fn:contains(attribute.name, 'nodeId')}">
+               <c:set var="nodeId" value="${attribute.value}"/>
+             </c:if>
+           </c:forEach>
+           <%
+           String grafanaProtocol = System.getProperty("org.opennms.grafanaBox.protocol", "http");
+           String grafanaHostname = System.getProperty("org.opennms.grafanaBox.hostname", "localhost");
+           int grafanaPort = Integer.parseInt(System.getProperty("org.opennms.grafanaBox.port", "3000"));
+
```
mvrueden: Can we put these properties in the response from the controller/command instead of manually defining them here in the jsp?

opennms-webapp/src/main/webapp/WEB-INF/jsp/graph/results.jsp
```diff
@@ -198,6 +199,29 @@
...
+           %>
+           <div ng-app="onms-ksc" ng-controller="checkFlowsCtrl" ng-init="getFlowCount(${nodeId}, ${ifIndex}, ${results.start.time}, ${results.end.time})">
+               <div ng-show="hasFlows">
+                 <a href="<%=grafanaProtocol%>://<%=grafanaHostname%>:<%=grafanaPort%>/dashboard/flows?orgId=1&amp;var-node=${nodeId}&amp;var-interface=${ifIndex}" style="padding-right: 3px" title="Open Flows for ${resultSet.resource.name}"><button type="button" class="btn btn-default btn-xs"><i class="fa fa-bar-chart" aria-hidden="true"></i></span></button></a>
```
j-white: We make a lot of assumptions here in the URL. I think we'd be better of to provide a URL template with placeholders and replace the nodeId, and ifIndex in the string appropriately.

opennms-webapp/src/main/webapp/WEB-INF/jsp/graph/results.jsp
```diff
@@ -198,6 +199,23 @@
...
+             <c:if test="${fn:contains(attribute.name, 'ifIndex')}">
+               <c:set var="ifIndex" value="${attribute.value}"/>
+             </c:if>
+             <c:if test="${fn:contains(attribute.name, 'nodeId')}">
+               <c:set var="nodeId" value="${attribute.value}"/>
+             </c:if>
+           </c:forEach>
+           <div ng-app="onms-ksc" ng-controller="checkFlowsCtrl" ng-init="getFlowCount(${nodeId}, ${ifIndex}, ${results.start.time}, ${results.end.time})">
+               <div ng-show="hasFlows">
+                 <a target=_blank" href="http://grafana:3000/dashboard/var-node=${nodeId}&amp;var-interface=${ifIndex}" style="padding-right: 3px" title="Open Flows for ${resultSet.resource.label}"><button type="button" class="btn btn-default btn-xs"><i class="fa fa-bar-chart" aria-hidden="true"></i></span></button></a>
```
j-white: We should make this URL configurable, allowing the user to provide some string with placeholders such as $nodeId, $ifIndex, $start and $end, and replace those tokens as necessary.

Ideally, we would do this server side, instead of in the JSP here. Perhaps introduce a new endpoint in the flow API?

j-white: Even if this link is a *resource level* link, I think it might be better to add the icon alongside the existing icons at the top of the graphs. We can do this for each graph in the resource.

opennms-webapp/src/main/webapp/WEB-INF/jsp/graph/results.jsp
```diff
@@ -198,6 +199,23 @@
...
+         <c:if test="${fn:contains(resultSet.resource.label,'(*)') != true}">
+           <c:forEach var="attribute" items="${resultSet.resource.attributes}">
+             <c:if test="${fn:contains(attribute.name, 'ifIndex')}">
+               <c:set var="ifIndex" value="${attribute.value}"/>
+             </c:if>
+             <c:if test="${fn:contains(attribute.name, 'nodeId')}">
+               <c:set var="nodeId" value="${attribute.value}"/>
+             </c:if>
+           </c:forEach>
+           <div ng-app="onms-ksc" ng-controller="checkFlowsCtrl" ng-init="getFlowCount(${nodeId}, ${ifIndex}, ${results.start.time}, ${results.end.time})">
```
j-white: Right now we end up triggering a REST call, which in turns make a call to Elasticsearch every time a user loads  a page with SNMP interfaces. We should find a way to avoid the queries to Elasticsearch if 1) the system is not processing flows or 2) we already know there are flows available.

If this is not feasible, perhaps would could just make it possible to enable/disable the icon i.e. we require some URL to be configured before we attempt to show the icon.

features/flows/rest/impl/src/main/java/org/opennms/netmgt/flows/rest/internal/FlowRestServiceImpl.java
```diff
@@ -48,6 +48,7 @@
 import javax.ws.rs.core.MultivaluedMap;
 import javax.ws.rs.core.UriInfo;
 
+import org.apache.commons.lang3.StringUtils;
```
mvrueden: Can we please not use commons-lang stuff. Try to use google guava or plain java instead.

features/flows/rest/api/src/main/java/org/opennms/netmgt/flows/rest/FlowRestService.java
```diff
@@ -138,4 +138,15 @@ FlowSeriesResponse getTopNConversationsSeries(
             @Context final UriInfo uriInfo
     );
 
+    /**
+     * Retrieves the deep dive tool url from configured properties.
+     *
+     *
+     * @param uriInfo JAX-RS context
+     * @return url to deep dive tool
+     */
```
mvrueden: The method name and `@Context` annotation is pretty straight forward. I'd opt for not repeating that with a comment.

features/flows/rest/impl/src/main/java/org/opennms/netmgt/flows/rest/internal/FlowRestServiceImpl.java
```diff
@@ -327,4 +329,26 @@ private static void populateResponseFromTable(Table<?, Long, Double> table, Flow
...
 
+    @Override
+    public String getDeepDiveToolUrl(UriInfo uriInfo) {
+
+        long count = waitForFuture(
+                flowRepository.getFlowCount(getFiltersFromQueryString(uriInfo.getQueryParameters())));
+
+        if (count > 0) {
+            // If there are valid flows then generate URL from template.
+            String deepDiveToolUrl = System.getProperty("org.opennms.netmgt.flows.deepDiveUrl");
```
mvrueden: Can we set the property when instantiating the repository. 
We should probably provide some defaults, as otherwise the deep dive url would not work by default, but OpenNMS usually provides some behaviour if a certain feature is enabled.

features/flows/rest/impl/src/main/java/org/opennms/netmgt/flows/rest/internal/FlowRestServiceImpl.java
```diff
@@ -327,4 +329,26 @@ private static void populateResponseFromTable(Table<?, Long, Double> table, Flow
         response.setValues(values);
     }
 
+    @Override
+    public String getDeepDiveToolUrl(UriInfo uriInfo) {
```
mvrueden: We may want to return a `Response` instead of a `String`, as `null` would be converted to a `404`.
I'd prefer a `204` in this case.

opennms-doc/guide-admin/src/asciidoc/text/telemetryd/protocols/netflow5/interface.adoc
```diff
@@ -163,6 +163,19 @@ Credentials for each node must be stored in `${OPENNMS_HOME}/etc/elastic-credent
 </elastic-credentials>
 ```
 
+===== Configure Deep Dive Tool  URL
+
+In order to directly access flows from OpenNMS Web UI, deep dive tool url needs to be configured in opennms.properties.
```
mvrueden: `{opennms-product-name}` should be used instead of `OpenNMS` to be consistent with the rest of the documentation.

opennms-doc/guide-admin/src/asciidoc/text/telemetryd/protocols/netflow5/interface.adoc
```diff
@@ -163,6 +163,19 @@ Credentials for each node must be stored in `${OPENNMS_HOME}/etc/elastic-credent
...
+
+Enable `org.opennms.netmgt.flows.deepDiveUrl` property and specify placeholders for OpenNMS related fields like `nodeId` (Node Id), `ifIndex` (SNMP Interface Index), `start` (Start time), `end` (End time).
```
mvrueden: `{opennms-product-name}` should be used instead of `OpenNMS` to be consistent with the rest of the documentation.

features/flows/rest/impl/src/main/java/org/opennms/netmgt/flows/rest/internal/FlowRestServiceImpl.java
mvrueden: Actually, this is not true. `null` will return a `204`, so everything is fine.

mvrueden:
I gave this is a spin and in general this works. 

Some things I noticed though:

To determine if the flow indicator should be shown, we actually perform 2 GET requests. The first to indicate if the icon is visible (`flowCount > 0`) and the second to determine the url. Whereas the last also does a check for `flowCount > 0` and performs the same query again. I vote for something like `/info` which returns an object, e.g. : `{ flowCount: 0, flowGraphUrl: "..."}`. 

The "Open Flow Indicator" is only shown, when there is some flow data. 
I'd prefer to have the button always visible, but disable it, if no data is available and tell the user via a tooltip.

The "Open Flow Indicator" is also visible if there is flow data (`flowCount > 0`) but the `deepDiveUrl` is not set (`500` was returned)

indigo423:
I definitely like the idea to always show the flow indicator. Have it gray and disabled for interfaces where flows are not available and a tooltip with an explanation why it is not available. It is more transparent and not a magic behavior.

cgorantla:
Since it is an extrenal tool that we link and it may not be needed for anybody who don't use flows, I prefer it to display when they are actually interested in flows. i.e. they configured the system to have flows and we have those flow graphs link here.

indigo423:
@cgorantla, I think we have two things to consider:

a) What is the behavior when "Flows" is configured or not configured
b) What is the behavior when you have flows configured but the interfaces do not have flow data

a) is more complicated, cause by default opennms has everything kind-a-turned-on. Is there a way to figure this out by a property? In this case what you are saying make sense.
b) IMHO @mvrueden is more about talking this problem and this makes more sense to me :)

j-white:
With NRTG, we already take the approach of not displaying the icon if it is not supported for the resource. We should probably stick to the same approach here.

We should also have better handling for the case in which flows are not being used - we currently attempt to query ES every time an interface graph is displayed.

indigo423:
@j-white this is exactly one problem of NRTG. It just works with SNMP. It is not visible on other collected metrics and nobody knows why. How does NRTG knows the metric is collected with SNMP? So easy, just look for SNMP in the name of the resource. Ooops, there are resources which are collected with SNMP but don't have the name in it?! We shouldn't copy bad behavior and fix it :)

indigo423:
Btw.: Trendline is also shown even R is not installed :)

pbrane:
> On Mar 14, 2018, at 08:33, Ronny Trommer <notifications@github.com> wrote:
> 
> nobody knows why

I know why. I have a good memory.  ;)

David

mvrueden:
We agreed to implement the following behaviour:
 * Don't provide a default for the `url` property
 * If the `url` property is defined, the feature is considered `enabled`
 * If the `url` property is undefined, the feature is considered `disabled`
 * If the feature is enabled, an icon is shown. If flows exist, it is enabled, otherwise it is disabled
 * If the feature is disabled, an icon is not shown.

pbrane:
Does this mean a ReST call for each graph?

j-white:
Yeah, there will be a REST call for every interface resource on the page. These are done asynchronously though, so they will not prevent the page/graphs from loading.

features/flows/rest/impl/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -15,11 +15,19 @@
     <reference id="transactionOperations" interface="org.springframework.transaction.support.TransactionOperations" availability="mandatory"/>
     <reference id="classificationService" interface="org.opennms.netmgt.flows.classification.ClassificationService" availability="mandatory"/>
 
+  <!-- Configuration properties -->
+	<cm:property-placeholder id="flowGraphUrlProperties" persistent-id="org.opennms.netmgt.flows.rest.internal" update-strategy="reload">
```
j-white: Can we drop the '.internal' here and change the pid to `org.opennms.netmgt.flows.rest` instead?

opennms-webapp/src/main/webapp/WEB-INF/jsp/graph/results.jsp
```diff
@@ -236,14 +248,23 @@
...
 		                    <c:if test="${fn:contains(resultSet.resource.resourceType.label, 'SNMP') || fn:contains(resultSet.resource.resourceType.label, 'TCA') }">
 		                        <c:if test="${fn:contains(resultSet.resource.label,'(*)') != true}">
-		                            <a href="javascript:popUp('${nrtgGraphUrl}')" title="Start NRT-Graphing for ${graph.title}"><button type="button" class="btn btn-default btn-xs" aria-label="Start NRT-Graphing for ${graph.title}"><span class="glyphicon glyphicon-flash" aria-hidden="true"></span></button></a><br/>
+		                            <a href="javascript:popUp('${nrtgGraphUrl}')" title="Start NRT-Graphing for ${graph.title}"><button type="button" class="btn btn-default btn-xs" aria-label="Start NRT-Graphing for ${graph.title}"><span class="glyphicon glyphicon-flash" aria-hidden="true"></span></button></a>
 		                        </c:if>
 		                    </c:if>
+                              <div style="display: inline" ng-if="flowsEnabled">
+                                <a ng-href="{{flowGraphUrl}}" target="_blank" style="padding-right: 3px" title="{{ hasFlows ? 'Open flow graphs' : 'No flows exists'}}">
```
j-white: Can we change 'No flows exist.' to 'No flows were found in the current time range.'? This should help make it clearer that the time range has an impact.

opennms-webapp/src/main/webapp/WEB-INF/jsp/graph/results.jsp
```diff
@@ -236,14 +248,23 @@
...
+                                <span> <button type="button" ng-disabled="!hasFlows" class="btn btn-default btn-xs">
+                                  <i class="fa fa-bar-chart" aria-hidden="true"></i>
```
j-white: Maybe we could use `fa-exchange` instead of `fa-bar-chart` to make it more indicative of "flows".

PR #1875, opened 2018-02-21
NMS-9887: Update Cisco ASA systemDefs with Cisco remote-access group

jeffgdotorg:
* JIRA: http://issues.opennms.org/browse/NMS-9887


jeffgdotorg:
Requesting review from @Sortova since this is config changes in foundation, and I don't want to run afoul of the rules about that sort of thing.

PR #1878, opened 2018-02-27
NMS-9763: Added MD4 provider for Radius

fooker:
* JIRA: http://issues.opennms.org/browse/NMS-9763

This is an ugly hack to get the radius stuff working again: It uses the MD4 implementation from jcifs and provides is to the security service interface, so jradius can pick it up.

The _clean_ solution would be to use BouncyCastle, but it breaks the karaf SSH shell.

Note: Having a OSGI bundle for each monitor implementation would be neat.

### DIFF
--- a/protocols/radius/src/main/java/org/opennms/protocols/radius/detector/RadiusAuthDetector.java
+++ b/protocols/radius/src/main/java/org/opennms/protocols/radius/detector/RadiusAuthDetector.java
@@ -128,7 +128,11 @@ public class RadiusAuthDetector extends BasicDetector<CompositeAttributeLists, R
     private String m_password = DEFAULT_PASSWORD;
     private String m_ttlsInnerAuthType = DEFAULT_TTLS_INNER_AUTH_TYPE;
 	private String m_InnerIdentity = DEFAULT_INNER_IDENTITY;
-    
+
+    static {
+        RadiusUtils.loadSecurityProvider();
+    }
+
     /**
      * Default constructor
      */
--- a/protocols/radius/src/main/java/org/opennms/protocols/radius/monitor/RadiusAuthMonitor.java
+++ b/protocols/radius/src/main/java/org/opennms/protocols/radius/monitor/RadiusAuthMonitor.java
@@ -133,6 +133,10 @@ public final class RadiusAuthMonitor extends AbstractServiceMonitor {
      */
     public static final String DEFAULT_TTLS_INNER_AUTH_TYPE= "pap";
 
+    static {
+        RadiusUtils.loadSecurityProvider();
+    }
+
     /**
      * Class constructor.
      *
--- a/protocols/radius/pom.xml
+++ b/protocols/radius/pom.xml
@@ -44,6 +44,12 @@
       <type>pom</type>
     </dependency>
     <dependency>
+      <!-- This adds support for MD4 digest used by mschapv2 - NMS-9763 -->
+      <groupId>org.opennms.dependencies</groupId>
+      <artifactId>jcifs-dependencies</artifactId>
+      <type>pom</type>
+    </dependency>
+    <dependency>
       <groupId>org.opennms.core.test-api</groupId>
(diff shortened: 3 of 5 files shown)


RangerRick:
Shouldn't this be targeted to `foundation-2017`?

mvrueden:
Can we at least not duplicate the same code three times? 
Maybe something like "HackUtils" or "SecurityUtils" in the module can reference it in the `static {}` blocks.

j-white:
I don't think we need to backport this to foundation-2017, I believe the regression was caused by the Karaf upgrade in 21.0.0.

protocols/radius/src/main/java/org/opennms/protocols/radius/detector/RadiusAuthDetector.java
```diff
@@ -128,7 +131,16 @@
     private String m_password = DEFAULT_PASSWORD;
     private String m_ttlsInnerAuthType = DEFAULT_TTLS_INNER_AUTH_TYPE;
 	private String m_InnerIdentity = DEFAULT_INNER_IDENTITY;
-    
+
+    static {
+        // This adds support for MD4 digest used by mschapv2 - NMS-9763
+        Security.addProvider(new Provider("MD4", 0.0D, "MD4 for Radius") {
+            {
+                this.put("MessageDigest.MD4", "jcifs.util.MD4");
```
j-white: Do we need to add an additional bundle import for these to work on Minion too?

fooker:
I've moved the duplicated code to the already existing `RadiusUtils`. I also switched from the class name as string to a real import to make the import come up in the manifest.

protocols/radius/src/main/java/org/opennms/protocols/radius/monitor/RadiusAuthMonitor.java
```diff
@@ -29,6 +29,8 @@
 package org.opennms.protocols.radius.monitor;
 
 import java.net.InetAddress;
+import java.security.Provider;
+import java.security.Security;
```
RangerRick: it's a small thing, but are these imports needed anymore now that they're in the static `RadiusUtils` call instead?
Otherwise this looks much cleaner, thanks!

fooker:
Ok, cherry-picked to `release-21.0.6`. Must it be merged to development afterwards or is it done automatically?

mvrueden:
That is done automatically.

mvrueden:
Closing PR, as the commit was cherry-picked to `release-21.0.6`

fooker:
This is not merged into the target yet. I've just retarget the PR by cherry-picking the old stuff on top of `release-21.0.6`. Must still be merged.

reissmann:
Finally! Thank you!

fooker:
Looks like this was merged to `release-21.0.6` which never got released. But it's still missing in `release-21.1.0`.

@RangerRick Can you confirm?

RangerRick:
Oh dangit, thought I caught all the "21.0.6 got renamed to 21.1.0" issues.  Looks like you're right.

RangerRick:
Oh wait, no, that made it in.  `git log opennms-21.1.0-1` gives me:

```
commit 01f9b1c404ee8a139489dcd5b16a511f72d66c33
Merge: cc68d9dc8e3 fe4a1ed9fc3
Author: Benjamin Reed <ranger@opennms.org>
Date:   Thu Mar 22 14:56:09 2018 -0400

    Merge remote-tracking branch 'origin/foundation-2017' into release-21.1.0

commit cc68d9dc8e36c647375db3d66777a1697ab66331
Author: Dustin Frisch <fooker@lab.sh>
Date:   Tue Feb 27 14:01:09 2018 +0100

    NMS-9763: Added MD4 provider for Radius

commit c279647ea629f4ac7ecdb2d38b4b76447a146418
Author: Markus von Rüden <mvr@opennms.com>
Date:   Mon Mar 19 17:28:55 2018 +0100

    NMS-9667: Replace camel with own netty-implementation to fix netty exceptions
```

PR #1879, opened 2018-03-01
HZN-1253: Fix error setting up IPFIX / Netflow 9 listeners

fooker:
* JIRA: http://issues.opennms.org/browse/HZN-1253


RangerRick:
@agalue needs to be able to test this next week on-site, so I'm going to go ahead and merge this, but I'll leave `HZN-1253` open with a note that it's not really finished yet.

PR #1880, opened 2018-03-01
Upgrade to Webpack 4 + associated fixes

RangerRick:
This PR introduces a number of performance improvements to speed up the asset build, including upgrading to webpack4 and refactoring the `webpack.config.js` to get rid of extra work the original config was doing.


RangerRick:
Looks like this is ready to merge, but I'd love it if someone could double-check it before I do so.  @mvrueden mind taking a look?

mvrueden:
This improves build time for me. It takes now roughly 5 - 8 minutes longer than before. But that is better than twice as long.

Using `yarn watch` with `org.opennms.web.assets.minified=false` also works.

PR #1881, opened 2018-03-02
NMS-9867 Inconsistent handling of keyboard navigation on top menu

patrick-schweizer:
Fix for https://issues.opennms.org/browse/NMS-9867

### DIFF
--- a/opennms-webapp/src/main/resources/org/opennms/web/controller/navbar.ftl
+++ b/opennms-webapp/src/main/resources/org/opennms/web/controller/navbar.ftl
@@ -50,7 +50,7 @@
 		        <#-- has sub-entries, draw menu drop-downs -->
 		        <li class="dropdown">
 		          <#if item.url?has_content && item.url != "#">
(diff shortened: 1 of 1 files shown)


mvrueden:
LGTM (-:

pbrane:
Shouldn’t this be committed to foundation branch?

mvrueden:
It depends where it should go. It is targeted for Meridian-2017.X. If it should go to Meridian-2015.X releases as well, we must cherry-pick it backwards (or should have merged it to foundation in the first place)

pbrane:
It was a PoweredBy customer that reported the issue and the fix should be targeted at a branch that they will be able to use; hence Foundation vs. Meridian.  And, shouldn't this make it to Horizon as well?

mvrueden:
`foundation` -> `Meridian-2015.X`, `Meridian-2016.X` and `Meridian-2017.X`
`foundation-2016` -> `Meridian-2016.X` and `Meridian 2017.X`
`foundation-2017` -> `Meridian-2017.X`

everything merged to any `foundation` branch will end up in the next Horizon release (22.0.0 -> `develop`) and the next current Horizon release (21.0.5 -> `release-21.0.5`)

As mentioned above, it all comes down to which Meridian version this patch should make it. I assumed `Meridian-2017`. If that is not the case, we have to put it in the correct foundation-branch.

PR #1883, opened 2018-03-05
NMS-9869: build all projects in the minion sub-project

RangerRick:
This is a version of the NMS-9869 patch for Foundation 2017 (where the list of projects was not auto-generated).

* JIRA: http://issues.opennms.org/browse/NMS-9869


mvrueden:
How do we ensure that we don't forget to add new modules here in the future?

RangerRick:
Because the version in develop actually checks all the poms programmatically.  This is just for 2017 where we didn't have that broken out. (It was done as part of the minion packaging refactor post-foundation-2017).

j-white:
We don't actually build Minion packages for Meridian 2017, so I don't think we actually need to backport this.

RangerRick:
Hm, suppose so, doesn't hurt to be there anyways, though.  :)

PR #1884, opened 2018-03-07
NMS-9861: Add jstl library 

cgorantla:
Statsd requires  ExtremeComponents jar which in turn depend on ExpressionEvaluatorManager class for some of the tags.
This class is not available in jstl library that jetty provides by default. So we need to have jstl  1.2 

* JIRA: http://issues.opennms.org/browse/NMS-9861


cgorantla:
closing this as jstl library is introducing conflicts. will update issue with another PR

PR #1885, opened 2018-03-13
HZN-9861: Add missing jstl packages needed by extremecomponents

cgorantla:
Jetty has apache-jstl as jstl dependency but it misses org.apache.taglibs.standard.lang.* package.
 Adding this required fixes the issues with Statsd

* JIRA: http://issues.opennms.org/browse/hzn-9861


pom.xml
```diff
@@ -3435,7 +3446,7 @@
       <dependency>
         <groupId>javax.servlet.jsp</groupId>
         <artifactId>jsp-api</artifactId>
-        <version>99.99.99-exclude-and-use-opennms-servlet-instead</version>
+        <version>2.2</version>
```
RangerRick: Shouldn't we *still* not be including jsp-api? Since everything is as it was but we just included an extra jar.

cgorantla: I have enabled here to avoid confusion in future when we are juggling these libraries.  `jsp-api` is going to end up  OPENNMS\lib anyways. So there is no meaning to exclude it in pom.
We also have it in our karaf features.
https://github.com/OpenNMS/opennms/blob/c7ff214afafd28d126d84917e3ae5f0d9779fb63/container/features/src/main/resources/features.xml#L175

RangerRick: Sure, but in Karaf it's only classloaded in the one bundle that needs it.  It shouldn't end up in lib.  It should be provided by `apache-jsp-*.jar` instead. (Which I actually just noticed we have 2 of, which is probably a problem also...)

RangerRick: Oop, apache-jsp is a false alarm, they're technically 2 related jars:

```
[INFO] org.opennms.dependencies:servlet-dependencies:pom:22.0.0-SNAPSHOT
[INFO] \- org.eclipse.jetty:apache-jsp:jar:9.4.2.v20170220:compile
[INFO]    +- org.eclipse.jetty:jetty-util:jar:9.4.2.v20170220:compile
[INFO]    +- org.eclipse.jetty.toolchain:jetty-schemas:jar:3.1:compile
[INFO]    +- javax.servlet:javax.servlet-api:jar:3.1.0:compile
[INFO]    +- org.mortbay.jasper:apache-jsp:jar:8.5.9.1:compile
[INFO]    |  \- org.mortbay.jasper:apache-el:jar:8.5.9.1:compile
[INFO]    \- org.eclipse.jdt.core.compiler:ecj:jar:4.4.2:compile
```

j-white:
Pulling jstl into our tree?

RangerRick:
Just a small part of it that for some reason isn't in the apache standard taglib jars. (but is in their source)

PR #1896, opened 2018-03-21
NMS-9667: Replace camel with own netty-implementation to fix netty exceptions

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-9667


j-white:
👍 from me. Tested both the UDP and TCP listeners. Also confirmed we can send multiple event logs over a single TCP session.

Can we retarget this for `release-21.0.6`?

mvrueden:
Yes, we can. I originally branched off `release-21.0.6` but forgot to set the target branch accordingly.

PR #1897, opened 2018-03-21
HZN-1274: Do not expose FlowRepository to jetty classpath

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1274


j-white:
Makes sense.

PR #1902, opened 2018-03-22
HZN-1216: sFlow support

christianpape:
* JIRA: https://issues.opennms.org/browse/HZN-1216


features/telemetry/listeners/flow/src/main/java/org/opennms/netmgt/telemetry/listeners/flow/ipfix/TcpListener.java
```diff
@@ -85,6 +85,8 @@ protected void initChannel(final SocketChannel ch) {
                                 .addLast(new ChannelInboundHandlerAdapter() {
                                     @Override
                                     public void exceptionCaught(final ChannelHandlerContext ctx, final Throwable cause) throws Exception {
+                                        cause.printStackTrace();
```
j-white: Oops.

features/telemetry/listeners/flow/src/main/java/org/opennms/netmgt/telemetry/listeners/flow/ipfix/TcpListener.java
```diff
@@ -134,4 +136,14 @@ public void setPort(final int port) {
     public void setDispatcher(final AsyncDispatcher<TelemetryMessage> dispatcher) {
         this.dispatcher = dispatcher;
     }
+
+    public static void main(final String... args) throws Exception {
```
j-white: This still needed?

features/telemetry/listeners/sflow/src/main/java/org/opennms/netmgt/telemetry/listeners/sflow/Listener.java
```diff
@@ -0,0 +1,147 @@
...
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.channel.ChannelInitializer;
+import io.netty.channel.ChannelOption;
+import io.netty.channel.EventLoopGroup;
+import io.netty.channel.nio.NioEventLoopGroup;
+import io.netty.channel.socket.DatagramChannel;
+import io.netty.channel.socket.nio.NioDatagramChannel;
+import io.netty.util.internal.SocketUtils;
+
+public class Listener implements org.opennms.netmgt.telemetry.listeners.api.Listener {
```
j-white: Don't we already have a base listener we can inherit from?

j-white:
Is there any reason we use a custom listener and parse the packets in the listener, as opposed to just forwarding the bytes and handling them in the adapter?

I know we took this approach for NF9 and IPFIX due to the template handling, but there doesn't seem to be any "state" that's kept in the listener here.

features/telemetry/listeners/sflow/src/main/java/org/opennms/netmgt/telemetry/listeners/sflow/Listener.java
christianpape: Yes, but the AbstractUdpListener include the session management stuff and we think it's a good idea to seperate Sflow from NF9/IPFix regarding a possible NSEL-implementation for NMS-9919.

features/telemetry/listeners/flow/src/main/java/org/opennms/netmgt/telemetry/listeners/flow/ipfix/TcpListener.java
christianpape: Oh, this wasn't in my original branch. This was added during the rebase. I'm checking this...

features/telemetry/listeners/flow/src/main/java/org/opennms/netmgt/telemetry/listeners/flow/ipfix/TcpListener.java
christianpape: Same here...

christianpape:
We thought it's a good idea to parse on the minion side. For all data reported from L2 devices the sampled IP layer and transport layer header must be parsed.

features/telemetry/listeners/sflow/src/main/java/org/opennms/netmgt/telemetry/listeners/sflow/Listener.java
j-white: Got it. Thanks.

features/telemetry/listeners/sflow/src/main/java/org/opennms/netmgt/telemetry/listeners/sflow/Listener.java
```diff
@@ -0,0 +1,135 @@
...
+        this.bossGroup = new NioEventLoopGroup();
+
+        final InetSocketAddress address = this.host != null
+                                ? SocketUtils.socketAddress(this.host, this.port)
+                                : new InetSocketAddress(this.port);
+
+        this.socketFuture = new Bootstrap()
+                .group(this.bossGroup)
+                .channel(NioDatagramChannel.class)
+                .option(ChannelOption.SO_REUSEADDR, true)
```
j-white: Can we setup the listener using the same options as the `org.opennms.netmgt.telemetry.listeners.udp.UdpListener`? (We should probably update `org.opennms.netmgt.telemetry.listeners.flow.AbstractUdpListener` with those as well.)

features/telemetry/listeners/sflow/src/main/java/org/opennms/netmgt/telemetry/listeners/sflow/PacketHandler.java
```diff
@@ -0,0 +1,88 @@
...
+            bsonWriter.writeName("data");
+            packet.content().version.datagram.writeBson(bsonWriter);
+
+            bsonWriter.writeEndDocument();
+        }
+
+        // Build the message to dispatch
+        final TelemetryMessage msg = new TelemetryMessage(packet.sender(), output.getByteBuffers().get(0).asNIO());
+
+        // Dispatch and retain a reference to the packet
```
j-white: If we continue to reference data in the packet after this function returns, then we should call `packet.retain()`, and `packet.release()` subsequently when the future is complete.

If this is not the case, we may want to remove this comment to avoid confusion.

j-white:
Looking good so far - I have data getting indexed into ES sent to a Minion from hsflowd.

I am seeing odd values for the `input_snmp` field though. Any pointers for what I should look at?

```json
{
	"@timestamp": 1521729681727,
	"@version": 1,
	"host": "127.0.0.1",
	"location": "MINION",
	"netflow.application": "https",
	"netflow.bytes": 127,
	"netflow.convo_key": "[\"MINION\",17,\"192.168.238.128\",57916,\"172.217.14.161\",443]",
	"netflow.direction": "ingress",
	"netflow.dst_addr": "172.217.14.161",
	"netflow.dst_locality": "public",
	"netflow.dst_port": 443,
	"netflow.first_switched": 1521729681727,
	"netflow.flow_locality": "public",
	"netflow.flow_records": 1,
	"netflow.flow_seq_num": 0,
	"netflow.initiator": true,
	"netflow.input_snmp": 1073741823,
	"netflow.last_switched": 1521729681727,
	"netflow.output_snmp": 2,
	"netflow.packets": 1,
	"netflow.protocol": 17,
	"netflow.sampling_interval": 1000,
	"netflow.src_addr": "192.168.238.128",
	"netflow.src_locality": "private",
	"netflow.src_port": 57916,
	"netflow.tos": 0,
	"netflow.version": "SFLOW",
	"node_exporter": {
		"foreign_source": "Minions",
		"foreign_id": "00000000-0000-0000-0000-000000ddba11",
		"node_id": 2,
		"categories": []
	}
}
```

christianpape:
Hmm..., only the two most significant bits are not set. Due to the specification this means  that there is no input interface and the traffic originates on the device itself. Should I map this to null?

j-white:
Mapping to `null` makes sense to me.

PR #1903, opened 2018-03-22
NMS-9831,HZN-1272: Handle OID event parameter keys and JSON event parameter values correctly

mvrueden:
Here we address some issues with the opennms-es-rest and opennms-alarm-change-notifier modules. It was easier to address it in one branch, than have the branches separated:
* JIRA: https://issues.opennms.org/browse/NMS-9831
* JIRA: https://issues.opennms.org/browse/HZN-1272

Please also note, that this change should also be merged to `release-21.x`.
As we changed the underlying structure for drift and `22.0.0` I will manually backport these changes to `21.x` as soon as this is considered okay.

### DIFF
--- a/features/opennms-es-rest/src/main/java/org/opennms/plugins/elasticsearch/rest/EventToIndex.java
+++ b/features/opennms-es-rest/src/main/java/org/opennms/plugins/elasticsearch/rest/EventToIndex.java
@@ -382,9 +382,14 @@ public class EventToIndex implements AutoCloseable {
 
 		body.put("host",event.getHost());
 
-		//get params from event
 		for(Parm parm : event.getParmCollection()) {
-			body.put("p_" + parm.getParmName(), parm.getValue().getContent());
+			// We have to handle oid parameters differently, as elastic would consider a containing "." as a sub
+			// field, e.g. ".1.2" results in a mapping of { 1: properties: { 2 } } which may exceed the maximum allowed
+			// fields for a mapping, see NMS-9831. Initially an array was used, to store the oids in the fashion:
+			// { "oid": "1.1.1", "value": "dummy value" } but elastic has some limitations on accessing objects in an
+			// array. Therefore we replace all existing . with an _
+			final String parmName = parm.getParmName().replaceAll("\\.", "_");
+			body.put("p_" + parmName, parm.getValue().getContent());
 		}
 
 		// remove old and new alarm values parms if not needed
--- a/features/opennms-es-rest/src/main/java/org/opennms/plugins/elasticsearch/rest/EventToIndex.java
+++ b/features/opennms-es-rest/src/main/java/org/opennms/plugins/elasticsearch/rest/EventToIndex.java
@@ -352,7 +352,7 @@ public class EventToIndex implements AutoCloseable {
 	 * Utility method to populate a Map with the most import event attributes
 	 */
 	private Index createEventIndexFromEvent(final Event event, final IndexAndType indexAndType) {
-		final Map<String,String> body = new HashMap<String,String>();
+		final JSONObject body = new JSONObject();
 
 		final Integer id = event.getDbid();
 		body.put("id", Integer.toString(id));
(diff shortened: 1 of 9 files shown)


features/opennms-es-rest/src/main/java/org/opennms/plugins/elasticsearch/rest/EventToIndex.java
```diff
@@ -384,9 +384,14 @@ private Index createEventIndexFromEvent(final Event event, final IndexAndType in
 
 		body.put("host",event.getHost());
 
-		//get params from event
 		for(Parm parm : event.getParmCollection()) {
-			body.put("p_" + parm.getParmName(), parm.getValue().getContent());
+			// We have to handle oid parameters differently, as elastic would consider a containing "." as a sub
+			// field, e.g. ".1.2" results in a mapping of { 1: properties: { 2 } } which may exceed the maximum allowed
+			// fields for a mapping, see NMS-9831. Initially an array was used, to store the oids in the fashion:
+			// { "oid": "1.1.1", "value": "dummy value" } but elastic has some limitations on accessing objects in an
```
Naicisum: Could you explain the limitations you are referring to?

Replacing periods with underscores will address the default depth issue of a parameter within elastic. As flat parameters, the 1000 default limit will still be hit, the array addresses both the depth and parameter limit.

A corresponding updated mapping template would allow you to "keyword" the "oid" and "value" values in addition can also be used to create a corresponding ".raw" sub-field that is type "text" for full text searches.

mvrueden: > Could you explain the limitations you are referring to?

This is what the official elastic documentation says about arrays: https://www.elastic.co/guide/en/elasticsearch/reference/current/array.html

> Replacing periods with underscores will address the default depth issue of a parameter within elastic. As flat parameters, the 1000 default limit will still be hit, the array addresses both the depth and parameter limit. [...]

That is true and not true. Replacing periods with underscores will flatten the parameters as you said and will reduce the likelihood of hitting the limit. As the event parameters can be configured and therefore are also dynamic only putting all event parameters in an array would fully fix the issue. 

I opted in for a consistent implementation as replacing `.` with `_` should probably always be performed and not only for `oid` parameters. Putting all parameters in a parameter array is not possible due to the limitations mentioned above (HZN-1272 is really the problem here). We can look into the `nested` option for arrays, though. 

Just a side note: The limit is configurable: https://discuss.elastic.co/t/total-fields-limit-setting/53004

Naicisum: How about a middle of the road solution, I would suggest adding a new parameter for the plugin to be able to chose between nested or flat.

In my instance, I would have to deal with over 20K+ flat fields per indices which can influence performance and memory consumption within elasticsearch. With a option available for choosing nested, I would eliminate this field explosion in my situation, while still preserving the option for others not in a similar situation to choose flat.

mvrueden: Yeah, I see your point now.

features/opennms-es-rest/src/main/java/org/opennms/plugins/elasticsearch/rest/EventToIndex.java
```diff
@@ -42,9 +42,11 @@
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
 
 import javax.xml.bind.DatatypeConverter;
 
+import org.json.JSONArray;
```
Naicisum: I found that if you use org.json.JSONArray, it will insert a name tag for the array in the JSON structure. You can correct this issue by using org.json.simple.JSONArray instead and using .add instead of .put for adding members.

mvrueden: Yeah, but I am doing a `body.getAsJsonString()` instead of `body`, which solves the issue as well, but I am in favor of not using 2 different packages. Will change to `org.json.simple`. Thanks for the comment.

PR #1905, opened 2018-03-23
HZN-1276: Updated Elasticsearch Template for Events

Naicisum:
* JIRA: http://issues.opennms.org/browse/HZN-1276

### DIFF
--- a/features/opennms-es-rest/src/main/resources/eventsIndexTemplate.json
+++ b/features/opennms-es-rest/src/main/resources/eventsIndexTemplate.json
@@ -1,100 +1,779 @@
 {
-      "template":"opennms-*",
-      "settings":{
-         "index.refresh_interval":"5s"
-      },
-      "mappings":{
-         "_default_":{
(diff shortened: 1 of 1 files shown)


j-white:
Pushed to `pr/jira/HZN-1276` to run it through Bamboo.

j-white:
The tests for Elasticsearch 5 are failing with these changes. See https://bamboo.opennms.org/browse/OPENNMS-ONMS2122-SMOKE-1

Naicisum:
PR1903 is a dependency, not only from the new p_oids mapping, but also depends on correctly parsing and structuring the JSON. I would recommend updating this branch against current drift and retesting.

j-white:
I merged the latest `features/drift` into `pr/jira/HZN-1276` so that the tests get triggered again.

j-white:
Closing this since the solution needs to be revised - either we need to support templates for multiple versions or drop support for earlier versions of ES.

PR #1907, opened 2018-03-26
NMS-9931: Make node cache configurable

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-9931

Here we provide a general `CacheConfig` bean which can be reused in the future to make caches configurable within `blueprint.xml`. 

The documentation should probably be re-structured and align to the latest features in a separate issue.

### DIFF
--- a/core/cache/pom.xml
+++ b/core/cache/pom.xml
@@ -48,5 +48,10 @@
             <version>${guavaVersion}</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>io.dropwizard.metrics</groupId>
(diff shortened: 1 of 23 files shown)


mvrueden:
Tests should pass now. Some properties are not documented yet. I'd like to address that when reworking the documentation, but can also do that now. They are then loosely in there. What do you think @j-white

j-white:
Looks good - basic properties are document.

I agree about the docs, let's wait until we revisit and restructure the flow related sections.

mvrueden:
I created a follow up issue to tackle the documentation: https://issues.opennms.org/browse/NMS-9939

PR #1911, opened 2018-03-27
NMS-9934: Update docs section of devel guide with section on migratin…

jeffgdotorg:
* JIRA: http://issues.opennms.org/browse/NMS-9934


j-white:
Again, what's the point of the PR here?

PR #1913, opened 2018-03-29
NMS-9831,HZN-1272: Handle OID event parameter keys and JSON event parameter values correctly

mvrueden:
Jira: 
 * https://issues.opennms.org/browse/NMS-9831
 * https://issues.opennms.org/browse/HZN-1272

When this is merged to `release-21.1.0` all changes should be dropped when `develop` is merged to `features/drift` as they have been made there already.


j-white:
Since we're changing the schema here, are there any updates that need to be made to the index template in `opennms-es-rest/main-module/src/main/resources/eventsIndexTemplate.json`?

mvrueden:
Yes, we probably should also backport the changes from PR  https://github.com/OpenNMS/opennms/pull/1905.

j-white:
There are other unrelated changes in https://github.com/OpenNMS/opennms/pull/1905, so we can't use that verbatim.

PR #1917, opened 2018-04-03
NMS-9885: collection:collect command fails with ClassCastException

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9885

Update the default CollectionAgentFactory implementation to create instances of the SnmpCollectionAgent as opposed to the DefaultCollectionAgent so that these work with all of the collectors.

Also improve the error message when the SnmpCollector is invoked via the collection:collection command without specifying a valid node/interface pair.


RangerRick:
Looks pretty straightforward.  I'll go ahead and merge this for the sprint.

PR #1918, opened 2018-04-04
NMS-9929: Updated Karaf to version 4.1.5

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-9929

### DIFF
--- a/features/minion/repository/src/main/resources/features.boot
+++ b/features/minion/repository/src/main/resources/features.boot
@@ -1,5 +1,7 @@
 # Minion Default Features
 pax-war
+spring/4.2.9.RELEASE_1
+spring-jms/4.2.9.RELEASE_1
 opennms-core-ipc-rpc-jms
 opennms-core-ipc-sink-camel
 opennms-syslogd-listener-camel-netty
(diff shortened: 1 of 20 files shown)


christianpape:
The session seems to stay open when using something like "ssh -o ServerAliveInterval=10".

features/minion/repository/src/main/resources/features.boot
```diff
@@ -1,5 +1,7 @@
 # Minion Default Features
 pax-war
+spring/4.2.9.RELEASE_1
```
j-white: It looks like this did the trick for fixing the Minion containers, since the smoke tests are passing now.

Can we add a comment here, something along the lines of:
```
# Install specific versions of the spring and spring-jms features to
# reduce the computations needed by the bundle dependency resolver
```

It would also be nice if we could refer to ${springVersion} as defined in the root pom instead of hard coding the version here.

christianpape:
Seems to work now. Thanks for the fix, Jesse!

features/minion/repository/src/main/resources/features.boot
christianpape: Ok, I'll check.

PR #1920, opened 2018-04-04
HZN-1276: Fix the event and alarm templates for Elasticsearch 6.x

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1276

Here we fix the event and alarm templates for ES 6.x by adding support for version aware templates.

We also include system tests for indexing in ES 6.x similar to those that were already present for ES 5.x.

### DIFF
--- a/features/jest/client/src/main/java/org/opennms/plugins/elasticsearch/rest/template/DefaultTemplateLoader.java
+++ b/features/jest/client/src/main/java/org/opennms/plugins/elasticsearch/rest/template/DefaultTemplateLoader.java
@@ -42,9 +42,9 @@ public class DefaultTemplateLoader implements TemplateLoader {
 
     @Override
     public String load(Version serverVersion, String resource) throws IOException {
-        // Attempt to find a template that is prefixed using the ES# where # is the major version
-        // of the ES server. If no template is found for the current major version, try the previous.
-        // If no template is found then attempt to look it up without the version suffix.
+        // Attempt to find a template that ends in .es#.json where # is the major version
+        // of the ES server. If no template is found for the current major version, try the previous,
+	// and so on. If no template is found then attempt to look it up without the version in the suffix.
         for (int i = serverVersion.getMajor(); i >= 0; i--) {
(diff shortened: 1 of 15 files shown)


features/jest/client/src/main/java/org/opennms/plugins/elasticsearch/rest/template/DefaultTemplateLoader.java
```diff
@@ -31,22 +31,43 @@
...
+            final String versionSuffix = i == 0 ? "" : String.format(".es%d", i);
+            final String resourceWithSuffix = String.format("%s%s.json", resource, versionSuffix);
+            final String template = getResource(resourceWithSuffix);
+            if (template != null) {
+                LOG.info("Using template with resource name: {}", resource);
+                return template;
+            }
+             LOG.debug("No template found with resource name: {}", resource);
+        }
+        throw new NullPointerException(String.format("No template found for server version %s and resource %s.",
```
mvrueden: Why NPE?

features/jest/client/src/main/java/org/opennms/plugins/elasticsearch/rest/template/Version.java
```diff
@@ -0,0 +1,82 @@
...
+    public void setMajor(int major) {
+        this.major = major;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        Version version = (Version) o;
+        return major == version.major &&
```
mvrueden: This may work here, but if at any time the types change from `int` to `Integer` this will not do what it is supposed to. I favour `Object.equals(major, version.major)` over `major == version.major`.

agalue:
I've installed latest RPMs from this branch (` jira/HZN-1276`) on my AWS Lab that was running Drift Build 84, and I was able to verify that the problem is gone. OpenNMS is now sending events to my Elasticsearch cluster version 6.2.2 (I've verified that the `opennms-events-raw-*` index was created, and it is being updated properly), so I believe this is ready to be merged, based on my test.

PR #1923, opened 2018-04-09
NMS-9915: Add missing spring-tx jar and remove duplicate spring-aop

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-9915

### DIFF
--- a/integrations/opennms-jasperstudio-extension/pom.xml
+++ b/integrations/opennms-jasperstudio-extension/pom.xml
@@ -66,11 +66,11 @@
             <artifactId>org.apache.servicemix.bundles.spring-expression</artifactId>
             <version>${springVersion}</version>
         </dependency>
-        <!--<dependency>-->
-            <!--<groupId>org.apache.servicemix.bundles</groupId>-->
-            <!--<artifactId>org.apache.servicemix.bundles.spring-asm</artifactId>-->
-            <!--<version>${springVersion}</version>-->
-        <!--</dependency>-->
+        <dependency>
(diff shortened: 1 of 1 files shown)


j-white:
Did we test loading the resulting .jar in JasperStudio?

mvrueden:
Yeah I tested that, but `JasperStudio` is not always reliably when you change the buildpath. I found it only reliably works (or not works, depending on your test) if you restart `JasperStudio` after you changed the buildpath.

I must have forgotten that when I verified the original PR.

mvrueden:
Waiting for @agalue to confirm that this is now actually working. Adding `triage` label until confirmation is received.

agalue:
In terms of the ClassNotFoundExceptions, those are solved on this PR. Now, using JasperStudio in general is very hard, and I don't have the knowledge to use it. Unless someone else wants to jump in and perform a full test of a given report in JasperStudio, I would say merge the PR.

mvrueden:
I verified the problem mentioned in the issue and this is now working, as the issue was the missing *spring-tx* dependency as stated above. @agalue tried to get a subreport running manually which is hard to achieve and requires a lot of tedious work, even if you know what you are doing. 

I verified that the parent report now runs successfully in JasperStudio, when configured as described in the [docs](https://docs.opennms.org/opennms/branches/develop/guide-admin/guide-admin.html#ga-database-reports-usage-jaspersoft-studio). 

I did not go through the trouble of configuring a subreport and get that to work.

PR #1924, opened 2018-04-10
NMS-9937: Always add group with empty name

fooker:
By adding the empty named group, one can leverage the existing group
maping by adding something like
```
"" => [ROLE_USER, ROLE_DASHBOARD]
```
which will be applied to all users.

* JIRA: http://issues.opennms.org/browse/NMS-9937

### DIFF
--- a/features/springframework-security/src/main/java/org/opennms/web/springframework/security/UserGroupLdapAuthoritiesPopulator.java
+++ b/features/springframework-security/src/main/java/org/opennms/web/springframework/security/UserGroupLdapAuthoritiesPopulator.java
@@ -104,6 +104,9 @@ public class UserGroupLdapAuthoritiesPopulator extends DefaultLdapAuthoritiesPop
 				this.groupRoleAttribute
 		);
 
+		// A Role mapping with an empty name is always applied to all users
+		userRoles.add("");
+
 		for(String group : userRoles) {
 			final List<String> rolesForGroup = this.groupToRoleMap.get(group);
 			LOG.debug("Checking {} for an associated role", group);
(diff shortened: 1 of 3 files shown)


agalue:
It works!

Once the documentation is updated, and maybe after adding a JUnit test, this should be ready to be merged.

Here is what I've added on `WEB-INF/spring-security.d/activeDirectory.xml` for testing purposes:

```
    <beans:property name="groupToRoleMap">
      <beans:map>
        <beans:entry>
          <!-- Name of the AD group for OpenNMS administrators -->
          <beans:key><beans:value>OpenNMS-Admins</beans:value></beans:key>
          <beans:list>
            <beans:value>ROLE_USER</beans:value>
            <beans:value>ROLE_ADMIN</beans:value>
          </beans:list>
        </beans:entry>
        <beans:entry>
          <!-- Name of the AD group for normal (non-admin) OpenNMS users -->
          <beans:key><beans:value></beans:value></beans:key>
          <beans:list>
            <beans:value>ROLE_USER</beans:value>
          </beans:list>
        </beans:entry>
      </beans:map>
    </beans:property>
```

fooker:
@agalue as there is no documentation for this (beside the wiki: https://wiki.opennms.org/wiki/Single_Sign_On) and no testing infrastructure we decided to just add an example the the configs shipped with ONMS.

fooker:
Of course, I will update the wiki article after it's merged.

agalue:
You're right. The official docs don't have the details, but the Admin guide has links to 2 places on the Wiki. One is the SSO article you've mentioned, and the other one is https://wiki.opennms.org/wiki/Spring_Security_and_LDAP, which I think is more appropriate for this particular change, even if that article is not explicit enough about the actual implementation of the external authentication providers (in our case, the usage of UserGroupLdapAuthoritiesPopulator, which is used by multiple implementations like LDAP, ActiveDirectory or Kerberos; check the files inside spring-security.d).

fooker:
* Added the example for active directory, too
* Rebased to `foundation-2017`
* Squashed

RangerRick:
Looks good to me, I'm going to go ahead and merge it.

PR #1926, opened 2018-04-10
HZN-1293: Remove converter from FlowRepository and do conversion before

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1293


j-white:
The reasoning behind the the previous implementation is so that conversion process would be timed by the `logConversionTimer`.

The conversion process is timed by the `logParsingTimer` instead.

PR #1930, opened 2018-04-17
NMS-9958: Set correct sysoid definitions in datacollection

indigo423:
The sysoidMask value can cause unwanted sideffects when the sysoidMask does not end with a "." character. All sysoid defintions which match a full described sysoid are set to sysoid instead of sysoidMask.

* JIRA: http://issues.opennms.org/browse/NMS-9958


indigo423:
@reviewer there was no target version specified in the issue, feedback from support mentioned Meridian which is the reason for merge request against to foundation-2017. Let me know if it should be rebased.

RangerRick:
So in JIRA, @pbrane said it should be targeted to Meridian 2018, so looks like it should rebase against Develop.

indigo423:
Anything else we have to do here?

agalue:
Looks good to me. The changes make sense.

PR #1931, opened 2018-04-17
HZN-1296: Fix thread safety with date formatter used by the IndexStrategy 

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1296


j-white:
@RangerRick pointed out we can use `java.time.DateTimeFormatter`  (which is thread safe) instead  of thread locals here.

mvrueden:
LGTM

PR #1937, opened 2018-04-24
NMS-9932: Fix wrong message displayed on AngularJS based tables.

cgorantla:
If resources are waiting to be loaded specify that instead of showing zero resources which could be misleading

* JIRA: http://issues.opennms.org/browse/NMS-9932

### DIFF
--- a/core/web-assets/src/main/assets/js/apps/onms-resources/index.js
+++ b/core/web-assets/src/main/assets/js/apps/onms-resources/index.js
@@ -39,6 +39,7 @@ angular.module('onms-resources', [
   $scope.maxSize = 5;
   $scope.totalItems = 0;
   $scope.hasResources = false;
+  $scope.loaded = false;
 
   $scope.goTo = function(id) {
     $window.location.href = getBaseHref() + 'graph/chooseresource.jsp?reports=all&parentResourceId=' + id + '&endUrl=' + $scope.endUrl;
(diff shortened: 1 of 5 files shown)


agalue:
Looks good to me. Looking at the code, I found a problem with my original implementation that I think it would be good to fix it as part of the solution.

As you can see, the Angular code is not handling errors after retrieving the information through ReST. The idea would be add a "catch" to the returned Promise and set "loaded" to be "false"; for example:

```JavaScript
 $http.get('rest/resources?depth=0').success(function(data) {
    $scope.loaded = true;
    $scope.hasResources = data.resource.length > 0;
    $scope.resources = data.resource;
    $scope.filteredResources = $scope.resources;
    $scope.update();
  }).error(function(error,status) {
    $scope.loaded = true;
    console.error("There was a problem retrieving the resources through ReST").
  });
```
That way, we nicely handle the case where there was a problem and then show that the table is empty. There are better ways to handle errors (which is how the requisitions UI behaves), but that should be enough for this use case.

Makes sense ?

cgorantla:
Sure, I will fix that

core/web-assets/src/main/assets/js/apps/onms-resources/index.js
```diff
@@ -51,6 +52,7 @@ angular.module('onms-resources', [
   };
 
   $http.get('rest/resources?depth=0').success(function(data) {
+    $scope.loaded = true;
```
agalue: It seems like you've missed the error handler in this case (to set `$scope.loaded=false`).

cgorantla: Thanks @agalue for catching, added it

agalue:
Excellent, thanks! Once Bamboo goes green, feel free to merge it.

PR #1938, opened 2018-04-24
HZN-1285: Initial Karaf assembly for Sentinel

RangerRick:
This PR refactors the project that builds a Minion container into a "base" pom (in `container/shared/`).  The `features/minion/container/karaf/` project has been changed to use this base, and then a `features/sentinel/` tree with a container and repository has been created alongside it.

Finally, an assembly (minus packaging) has been created in `opennms-assemblies/sentinel/` which is a skeleton for a Minion-like `.tar.gz` release along with Debian and RPM packages.

* JIRA: http://issues.opennms.org/browse/HZN-1285


mvrueden:
Here are some thoughts:

We have a `container` module, but `features/minion/container` or `features/sentinel/container` define the actual container.

With `features/sentinel|minion` we have modules which contain container specific bundles.
But with this structure where would we put bundles, which are shared between all containers? Is it enough to just place them in `features/shared` or just in `features`?

Maybe the following structure is more clear?
 * features
 ** container
 *** shared
 ***  minion
 *** sentinel

RangerRick:
What do you mean by "Only thing which actually bothers me, that shared bundles don't have a home" ?

You mean like the `svc` and `karaf-extender` projects?

j-white:
A couple notes:
1. Can we make it so that the Sentinel's port (SSH, RMI) do no conflict with Minion or the OpenNMS JVM, allowing us to run all 3 on a single machine.
2. Can we just put all of the bundles in the `system` folder, instead of using using the `default` repository in the managed by the `karaf-extender` - we don't really need this functionality for Sentinel.

j-white:
Thanks for the updates.

@mvrueden does this look good to merge now?

mvrueden:
Yep, LGTM :)

PR #1939, opened 2018-04-25
HZN-1299: Review and validate JFlow v5 support

christianpape:
* JIRA: https://issues.opennms.org/browse/HZN-1299

See also 
https://github.com/OpenNMS/elasticsearch-drift-plugin/pull/7


features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/SamplingAlgorithm.java
```diff
@@ -0,0 +1,78 @@
...
+    @SerializedName("UniformProbabilisticSampling")
+    UniformProbabilisticSampling,
+    @SerializedName("PropertyMatchFiltering")
+    PropertyMatchFiltering,
+    @SerializedName("HashBasedFiltering")
+    HashBasedFiltering,
+    @SerializedName("FlowStateDependentIntermediateFlowSelectionProcess")
+    FlowStateDependentIntermediateFlowSelectionProcess;
+
+    public static SamplingAlgorithm from(final Flow.SamplingAlgorithm samplingAlgorithm) {
```
mvrueden: Can we please write a test for this, ensuring that in the future we don't forget to map something.

j-white:
What do you think about storing the scaled values in ES instead of performing this action with the plugin?

Some pros:
* Handling of more complex scaling/sampling scenarios
* Faster query execution speed (no need to lookup multiple values and multiply these)

Cons:
* Storage required for another field

christianpape:
Seems reasonable. @fooker Do you have an opinion on that?

fooker:
Another con: we can not correct history.
But as long as we store the value in an additional field, it should be fine and we can fix things by reindexing.

christianpape:
Found additional problem. Please also review and merge PR https://github.com/OpenNMS/elasticsearch-drift-plugin/pull/7

PR #1940, opened 2018-04-25
NMS-9965: Update notification documentation

mmahacek:
NMS-9965: Update Notification documentation with links to OpenNMS 101 videos

* JIRA: https://issues.opennms.org/browse/NMS-9965

Added links to OpenNMS 101 videos and a reference that SMTP settings can be found in javamail-configuration.properties

### DIFF
--- a/opennms-doc/guide-admin/src/asciidoc/text/notifications/concepts.adoc
+++ b/opennms-doc/guide-admin/src/asciidoc/text/notifications/concepts.adoc
@@ -149,4 +149,4 @@ NOTE: Use external binary notification commands sparingly to avoid fork-bombing 
 Originally, all notification commands were external.
 Today only the `numericPage` and `textPage` commands use external programs to do their work.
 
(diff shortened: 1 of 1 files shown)


indigo423:
I would vote against linking YouTube videos in our official docs for the following reasons:

* Probabilities for changing links is pretty high
* Can't be updated by a developer/user when the behavior is changed

IMHO would be better when the content of the video is transcribed to real documentation.

mmahacek:
I get that about the static nature of the video content. Though given how often the videos are referenced in chat, I would at least suggest referencing https://www.youtube.com/OpenNMS to direct users to the repository to more easily find the videos.

indigo423:
@mmahacek Yes, got it agree. What do you think adding a section with a link to the videos here: https://www.opennms.org/en/docs?

RangerRick:
Yeah, I think there should at least be generic links, even if they have a caveat that they may not necessarily apply to the latest version.  The chances of "the content of the video is transcribed to real documentation" happening are exactly 0.  :)

PR #1944, opened 2018-04-28
NMS-9919: Fix first and last switched timestamps for Netflow v9

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9919

Here we fix an issue with timestamp parsing in Netflow v9 - the `sysUpTime` is provided in milliseconds and not in seconds.


christianpape:
Yes, milliseconds - sorry we missed that. PR looks good for me...

PR #1945, opened 2018-04-30
NMS-10067: Enhance AlarmPersisterImpl to support updating acknowledgment values on reduction

jeffgdotorg:
* JIRA: http://issues.opennms.org/browse/NMS-10067

### DIFF
--- a/opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
+++ b/opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
@@ -147,7 +147,7 @@ public class AlarmPersisterImpl implements AlarmPersister {
             if (LOG.isDebugEnabled()) {
                 LOG.debug("addOrReduceEventAsAlarm: reductionKey:{} found, reducing event to existing alarm: {}", reductionKey, alarm.getIpAddr());
             }
-            reduceEvent(e, alarm, event, getEventForwarder());
+            reduceEvent(e, alarm, event);
             m_alarmDao.update(alarm);
             m_eventDao.update(e);
 
@@ -175,7 +175,7 @@ public class AlarmPersisterImpl implements AlarmPersister {
         return new OnmsAlarmAndLifecycleEvent(alarm, ebldr.getEvent());
     }
 
-    private static void reduceEvent(OnmsEvent e, OnmsAlarm alarm, Event event, EventForwarder f) {
+    private static void reduceEvent(OnmsEvent e, OnmsAlarm alarm, Event event) {
         
         //Always set these
         alarm.setLastEvent(e);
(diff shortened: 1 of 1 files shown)


opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
```diff
@@ -63,6 +63,7 @@
  */
 public class AlarmPersisterImpl implements AlarmPersister {
     private static final Logger LOG = LoggerFactory.getLogger(AlarmPersisterImpl.class);
+    public static final String EVENT_SOURCE_NAME = "Alarmd";
```
j-white: Unused constant.

opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
```diff
@@ -174,7 +175,7 @@ private OnmsAlarmAndLifecycleEvent addOrReduceEventAsAlarm(Event event, boolean
         return new OnmsAlarmAndLifecycleEvent(alarm, ebldr.getEvent());
     }
 
-    private static void reduceEvent(OnmsEvent e, OnmsAlarm alarm, Event event) {
+    private static void reduceEvent(OnmsEvent e, OnmsAlarm alarm, Event event, EventForwarder f) {
```
j-white: Unused argument.

opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
```diff
@@ -211,7 +212,8 @@ private static void reduceEvent(OnmsEvent e, OnmsAlarm alarm, Event event) {
                         alarm.setSeverity(OnmsSeverity.valueOf(e.getSeverityLabel()));
                     } else if (fieldName.toLowerCase().contains("descr")) {
                         alarm.setDescription(e.getEventDescr());
-                        alarm.setSeverity(OnmsSeverity.valueOf(e.getSeverityLabel()));
```
j-white: Was this removal intentional?

pbrane:
Those "unused" issues references should be deleted.  This was from the patch I provided.
The update of description when the update of severity was being handled appeared to be a bug IMHO since that removes control of that field from the user's configuration.  Probably something I did decades ago. :(

opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
jeffgdotorg: I agree with @pbrane that this one is likely a copypasta bug.

pbrane:
Let's do that later as another enhancement issue.

j-white:
@jeffgdotorg I'd opt for the JEXL expression which would give more flexibility.

Users using this feature may also want to update the `unclear` automation to un-ack as well.

Once merged, can we also look at updating the wiki page so that this is reference-able somewhere other than in the code: https://wiki.opennms.org/wiki/Configuring_alarms#Update_Field

PR #1947, opened 2018-05-01
HZN-1301: Improve service lookup reliability by adding a relative timeout

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1301

Fix intermittent telemetryd adapter loading when the startup process takes a long time.


cgorantla:
Looks good to me :+1:

PR #1952, opened 2018-05-03
NMS-9064: better handling of thresholding config errors

RangerRick:
This PR does a couple of things:
1. makes Collectd die on startup if the `thresholds.xml` or `threshd-configuration.xml` files are wrong
2. makes Collectd's threshold set reliably fall back to the previous config on a reload (if configs are changed from good to bad at runtime)

I also confirmed that the `config-tester` binary does test the thresholding configs, as per @mfuhrmann's question in the issue.

* JIRA: http://issues.opennms.org/browse/NMS-9064


mfuhrmann:
@RangerRick Thanks!

PR #1953, opened 2018-05-04
 NMS-9882 PageSequenceMonitor broken in Minion

patrick-schweizer:
* JIRA: https://issues.opennms.org/browse/NMS-9882


j-white:
Looks good. Another way we've been fixing these types of issues is simply moving the classes out from `opennms-utils` into `core/lib` since the later does export this package.

patrick-schweizer:
@j-white is opennms-utils deprecated and we want to port things to core/lib or what is the difference between the two?

PR #1954, opened 2018-05-04
NMS-9506: Fix interface deletion from node

cgorantla:
Fix interface deletion from node.
Update Response Status code to 'Accepted' (202) for delete calls which are asynchronous.
Add smoke-test for some of the rest calls.

* JIRA: http://issues.opennms.org/browse/NMS-9506


j-white:
👍  for the changes.


The [developer guide](http://docs.opennms.org/opennms/releases/21.1.0/guide-development/guide-development.html#_http_return_codes) mentions: `All the DELETE requests are going to return a 204 (NO_CONTENT) on success.`

Since this is no longer true, well want to update that as well.

PR #1955, opened 2018-05-04
HZN-1245:  Update the SnmpInterfacePoller to be location aware

cgorantla:
* JIRA: http://issues.opennms.org/browse/HZN-1245
* JIRA: http://issues.opennms.org/browse/NMS-9917


opennms-services/src/main/java/org/opennms/netmgt/snmpinterfacepoller/SnmpPollInterfaceMonitor.java
```diff
@@ -64,6 +67,16 @@
...
     private static final String IF_OPER_STATUS_OID = ".1.3.6.1.2.1.2.2.1.8.";
 
+    private final LocationAwareSnmpClient m_client;
+
+    private String m_location;
+
+    private long m_interval;
+
+    public SnmpPollInterfaceMonitor(LocationAwareSnmpClient locationAwareSnmpClient) {
+        this.m_client = locationAwareSnmpClient;
```
j-white: Nit: `this.` is redundant when using variables prefixed with `m_`

opennms-services/src/main/java/org/opennms/netmgt/snmpinterfacepoller/pollable/PollableSnmpInterface.java
```diff
@@ -284,12 +283,16 @@ public boolean isReady() {
     @Override
     public void run() {
             if (getParent().polling()) {
-                LOG.info("run: polling SNMP interfaces on package/interface {}/{} on primary address: {}", getParent().getPackageName(), getName(), getParent().getIpaddress());
+                String location = getContext().getLocation(getParent().getNodeid());
```
j-white: We could also store the location in the `PollableInterface` class along with the node ID instead of looking this up every time. Other components in the code assume that a node's location doesn't change throughout it's lifetime, so we don't have to worry about the location changing from one run to another.

PR #1958, opened 2018-05-09
NMS-9852: Fixed NPE in debug code

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-9852

### DIFF
--- a/opennms-model/src/main/java/org/opennms/netmgt/model/topology/BroadcastDomain.java
+++ b/opennms-model/src/main/java/org/opennms/netmgt/model/topology/BroadcastDomain.java
@@ -34,13 +34,12 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.locks.ReentrantLock;
 
 import org.opennms.core.utils.InetAddressUtils;
 import org.opennms.netmgt.model.BridgeElement;
 import org.opennms.netmgt.model.BridgeMacLink;
-import org.opennms.netmgt.model.OnmsNode;
 import org.opennms.netmgt.model.BridgeMacLink.BridgeDot1qTpFdbStatus;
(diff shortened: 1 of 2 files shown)


opennms-model/src/main/java/org/opennms/netmgt/model/topology/BroadcastDomain.java
```diff
@@ -504,7 +503,11 @@ public String printTopologyFromLevel(Set<Integer> bridgeIds, int level) {
         strbfr.append(bridgeIds);
         strbfr.append("\n");
         for (Integer bridgeid : bridgeIds) {
-        	strbfr.append(getBridge(bridgeid).printTopology());
+            final Bridge bridge = getBridge(bridgeid);
+            if (bridge == null) {
+                continue;
```
mvrueden: Can we use the Java Stream API instead and filter for the bridges which are actually available instead of using continue here. Something like this:
```java
bridgeIds.stream().map(id -> getBridge(id)).filter(bridge -> bridge != null).forEach(bridge -> {

....

})
```

opennms-services/src/test/java/org/opennms/netmgt/enlinkd/BroadcastDomainTest.java
```diff
@@ -158,6 +160,25 @@ public void testTwoConnectedBridge() throws Exception {
         
     }
 
+    @Test
+    public void testPrintTopologyFromLevel() throws Exception {
+        TwoMergeBridgeTopology topology = new TwoMergeBridgeTopology();
```
mvrueden: 👍 for the test :)

opennms-model/src/main/java/org/opennms/netmgt/model/topology/BroadcastDomain.java
christianpape: Ok

PR #1959, opened 2018-05-10
NMS-9900 Value of ${nodeLabel} for PSM services apparently not eagerly updated

patrick-schweizer:
* JIRA: https://issues.opennms.org/browse/NMS-9900

### DIFF
--- a/opennms-model/src/main/java/org/opennms/netmgt/model/events/NodeLabelChangedEventBuilder.java
+++ b/opennms-model/src/main/java/org/opennms/netmgt/model/events/NodeLabelChangedEventBuilder.java
@@ -1,0 +1,79 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2018 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2018 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License,
+ * or (at your option) any later version.
+ *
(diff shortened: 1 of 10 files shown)


opennms-model/src/main/java/org/opennms/netmgt/model/events/NodeLabelChangedEventBuilder.java
```diff
@@ -0,0 +1,79 @@
...
+import org.opennms.netmgt.events.api.EventConstants;
+import org.opennms.netmgt.xml.event.Event;
+import org.opennms.netmgt.xml.event.Parm;
+
+import com.google.common.base.Strings;
+
+public class NodeLabelChangedEventBuilder extends EventBuilder {
+
+    public NodeLabelChangedEventBuilder(final String source) {
+        super(EventConstants.NODE_LABEL_CHANGED_EVENT_UEI, source, new Date());
```
j-white: Does this not trigger the `checkForIllegalUei()` check defined in the `EventBuilder` class?

patrick-schweizer: you are right. I fixed it by overriding the method in the subclass

patrick-schweizer:
@j-white , it looks to me that the bug exists since 2014, the parameters in the creation of the event were changed (https://github.com/OpenNMS/opennms/commit/444c1c05b7ed9d99718ac21be08c5544b6e3898b) but it looks like the processing of the event was still expecting the old parameter: PollerEventProcessor, line 382

mvrueden:
Looks good to me as well.
This should at least go to `foundation-2017`, but may also be merged to `foundation-2016` or `foundation`. @j-white what do you think?

j-white:
I agree. I'd say we re-target this for `foundation-2017`.

If the issue is raised on an older release we can consider back-porting it further then.

patrick-schweizer:
ok will do

patrick-schweizer:
Replaced by: #1984
=> closing this PR

PR #1960, opened 2018-05-11
NMS-10069: replaced default downtime model config in the doc

mfuhrmann:
* Documentation needed an update because of the changes in: https://issues.opennms.org/browse/NMS-9257
* Jira: https://issues.opennms.org/browse/NMS-10069


mfuhrmann:
So do we still have the delete option and what excactly is it doing? Does it delete or set forced unmanaged the service?
If it's existent I have to add this parameter to the docs, because now there is no information about it.

indigo423:
I would recommend making a new issue about enhancing the docs with your questions.

mfuhrmann:
@indigo423 Ok. I've created an issue for clarification: https://issues.opennms.org/browse/NMS-10133

PR #1961, opened 2018-05-12
NMS-10136: Remove Dell Open-Manage from all Net-SNMP devices

indigo423:
Remove Open Manage OIDs from every Net-SNMP capable device.

* JIRA: http://issues.opennms.org/browse/NMS-10136


agalue:
+1 Makes sense!

PR #1962, opened 2018-05-12
NMS-10137: Split SNMP Informant configurations

indigo423:
To allow users to easily add or remove data collections which support SNMP Informant extension packages, the configurations are splitted in the same way as SNMP provides extension packages.

* JIRA: http://issues.opennms.org/browse/NMS-10137


agalue:
Looks reasonable. Although, not necessarily available on Windows Servers, as people would have to pay and install the SNMP Informant optional plugin to have those OIDs (specially if you're planning to use advanced counters, Exchange, MSSQL, etc.).

In other words, separating the files makes sense to me. But, forcing those stats by adding systemDefs doesn't (even if that's consistent with what we had before).

As I've mentioned, I prefer to only request data that for sure exist on the target device, rather than requesting OIDs with a high chance that won't be present. We can always update the docs to let the small amount of people that use SNMP Informant that they should re-enable them after the upgrade.

agalue:
+1
The changes are easier to understand after reading the proposed docs for the release notes.

indigo423:
@agalue as you mentioned, it makes sense to remove them from the default and tell users to add them to your windows boxes if they really use it.

RangerRick:
When you removed it from default it broke the build.  I reverted the merge for now.

PR #1963, opened 2018-05-12
NMS-10138: Remove Power over Ethernet MIB data collection by default

indigo423:
Remove Power over Ethernet MIB data collection from all SNMP agents supporting Standard MIB-II and from Net-SNMP agents by default. For the reason they can be added to any Vendor who supports the standard PoE MIB it is added and included as a Reference file (REF_...) and can included to any vendor who supports it.

* JIRA: http://issues.opennms.org/browse/NMS-10138

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/datacollection-config.xml
+++ b/opennms-base-assembly/src/main/filtered/etc/datacollection-config.xml
@@ -65,6 +65,7 @@
       <include-collection dataCollectionGroup="Powerware"/>
       <include-collection dataCollectionGroup="Printers"/>
       <include-collection dataCollectionGroup="REF_Compaq-Insight-Manager"/>
+      <include-collection dataCollectionGroup="REF_MIB2-Powerethernet"/>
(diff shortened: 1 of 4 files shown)


agalue:
It seems like you forgot to remove the resource type `pethMainPseGroupIndex` from `mib2.xml`.

opennms-base-assembly/src/main/filtered/etc/resource-types.d/ref_mib2-powerethernet.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0"?>
+<resource-types>
+    <resourceType name="pethMainPseGroupIndex" label="Power Ethernet (MIB-2 powerEthernet)">
```
j-white: I'd opt for keeping the resource type definition for SNMP related data collection bits in the data-collection config rather than moving them here.

The number of resource types loaded has some affect on the performance in enumerating resources, particular when using Newts, and it is more difficult to disable them when they are here.

PR #1964, opened 2018-05-12
NMS-10139: Make 64bit interface counter default

indigo423:
A lot of devices add 64bit counter additional to the 32bit counter. This means the interface metrics are collected twice. Make the 64bit counter default and just allow the user to add 32bit counter on legacy devices and reduce the data collection by default.

The old 32bit MIB-II interface counter is added as a reference file which is not assigned by default to any system definition. They need to be added to legacy devices which support only 32bit counter and not 64bit counter by default.

* JIRA: http://issues.opennms.org/browse/NMS-10139

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/datacollection-config.xml
+++ b/opennms-base-assembly/src/main/filtered/etc/datacollection-config.xml
@@ -65,6 +65,7 @@
       <include-collection dataCollectionGroup="Powerware"/>
       <include-collection dataCollectionGroup="Printers"/>
       <include-collection dataCollectionGroup="REF_Compaq-Insight-Manager"/>
+      <include-collection dataCollectionGroup="REF_MIB2-Interfaces"/>
       <include-collection dataCollectionGroup="REF_MIB2-Powerethernet"/>
       <include-collection dataCollectionGroup="Riverbed"/>
       <include-collection dataCollectionGroup="Routers"/>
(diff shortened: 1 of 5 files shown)


agalue:
This change makes sense, but it also requires documentation changes, to let the users know that they will have to update their respective configurations if they depend on `mib2-interfaces` after the upgrade.

agalue:
+1
The docs looks good, thanks!

indigo423:
@mmahacek @agalue fixed duplicated docs.

jeffgdotorg:
I'm nervous about this one, despite being firmly behind the idea of addressing the problem it's intended to solve. Have we tested to confirm that an upgrading user using JRobin or RRDTool along with store-by-group persisting will not encounter data loss or corruption?

agalue:
Neither the aliasses of the `mibObj` nor the names of the `group` were changed.

The reference to the groups associated with 32-bits counters (a.k.a. `mib2-interfaces) was removed from the default systemDef.

That means, no changes on the existing RRD/JRB files are required or going to be needed. What is going to happen is that the RRD/JRB files associated with 32-bits counters won't be updated, and if you want that to happen (as the new docs say), you just have to enable the new datacollection group (Legacy_MIB2-Interfaces)

jmagudo:
This change introduces a problem because also removes counters for the following:       

<mibObj oid=".1.3.6.1.2.1.2.2.1.13" instance="ifIndex" alias="ifInDiscards" type="counter"/>
<mibObj oid=".1.3.6.1.2.1.2.2.1.14" instance="ifIndex" alias="ifInErrors" type="counter"/>
<mibObj oid=".1.3.6.1.2.1.2.2.1.19" instance="ifIndex" alias="ifOutDiscards" type="counter"/>
<mibObj oid=".1.3.6.1.2.1.2.2.1.20" instance="ifIndex" alias="ifOutErrors" type="counter"/>

Also breaks some of the defaults Thresholds of group mib2 that reference the old counters.

PR #1965, opened 2018-05-12
NMS-10140: Make Standard UPS MIB a reference

indigo423:
Set the Standard UPS MIB as a reference, cause it is required to add it manually to the vendors who support it.

* JIRA: http://issues.opennms.org/browse/NMS-10140


indigo423:
See JIRA

PR #1966, opened 2018-05-12
NMS-10141: Remove overlapping MIB-2 64bit counter configs

indigo423:
Remove overlapping MIB-II overlapping 64bit interface counter from default, cause they are already collected by default from MIB-2 Interface configuration. More details are described in JIRA.

This PR depends on PR #1964 

* JIRA: http://issues.opennms.org/browse/NMS-10141


agalue:
+1 after merging PR 1964

PR #1967, opened 2018-05-12
NMS-10142: Remove checkpoint data collection from all Windows servers

indigo423:
Remove by default Checkpoint firewall SNMP metrics from every Windows server

* JIRA: http://issues.opennms.org/browse/NMS-10142


agalue:
+1 Makes sense!

PR #1968, opened 2018-05-12
NMS-10143: Remove by default to collect Compaq Insight Manager stats

indigo423:
Remove Compaq Insight Manager statistics data collection from all Windows servers.

* JIRA: http://issues.opennms.org/browse/NMS-10143


agalue:
+1 Makes sense!

PR #1969, opened 2018-05-12
NMS-10144: Remove Cisco Call Manager stats from all Windows Server

indigo423:
Remove Cisco Call Manager stats from all Windows Server

* JIRA: http://issues.opennms.org/browse/NMS-10144


agalue:
+1 Makes sense!

PR #1970, opened 2018-05-12
NMS-10145: Classify IETF MIB as reference

indigo423:
Classify IETF MIB as reference, cause nothing is assigned to any device and needs to be added manually to vendors who supports it.

* JIRA: http://issues.opennms.org/browse/NMS-10145


agalue:
+1 Makes sense! Although, if there is some things to consider when expanding the data collection configuration in terms of using the prefix `REF_` when naming a `datacollection-group`, I believe that should be documented somewhere, to avoid changes like this in the future.

indigo423:
See JIRA issue

PR #1971, opened 2018-05-12
NMS-10146: Classify the Printer MIB as reference

indigo423:
Classify the Printer MIB as reference cause it is not assigned by default.

* JIRA: http://issues.opennms.org/browse/NMS-10146


indigo423:
See JIRA issue

PR #1978, opened 2018-05-15
HZN-1302: Include the application name when rendering conversations.

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1302

Here we update the flow conversation related REST endpoint to also include the application name when returning details about the conversations.


RangerRick:
Looks pretty reasonable, and it all passed checks. 👍

PR #1980, opened 2018-05-16
NMS-9117  config-tester doesn't check opennms.properties

patrick-schweizer:
* https://issues.opennms.org/browse/NMS-9117


j-white:
Looks good.

FYI: Only `.properties` files are supported in `opennms.properties.d`, so there's no need verify/validate XML files in this folder.

RangerRick:
FYI, this won't pass because it's failing header validation, you need to add copyright headers to your tests.

mvrueden:
Yeah, usually IntelliJ should do that for you, unless you create the test from the class you are trying to create the tests for. Somehow those classes don't include the opennms license headers, so they must be added manually.

patrick-schweizer:
@RangerRick , @mvrueden thanks, that is what happened. I fixed the license info.
@j-white, ok I will remove it, but before I do that: Is there another directory that needs to be checked or is this the only one?

RangerRick:
I guess there's technically also `snmp-graph.properties.d` but I think the regular `snmp-graph.properties` validation does that already?

patrick-schweizer:
@RangerRick what would the "regular snmp-graph.properties validation does that already?" be? Is there something for me to do? Thanks!

PR #1982, opened 2018-05-16
NMS-10067: Enhance AlarmPersisterImpl to support updating acknowledgment values on reduction

jeffgdotorg:
* JIRA: http://issues.opennms.org/browse/NMS-10067


j-white:
Looks good. Thanks for the adding the `decodeSnmpV2TcDateAndTime` tests.

Can we add a blurb to the wiki page about this functionality https://wiki.opennms.org/wiki/Configuring_alarms#Update_Field?

PR #1983, opened 2018-05-16
NMS-10009: Can't receive SNMP v3 Trap

cgorantla:
There are two instances of Snmp4JStrategy, one from default classLoader and other loaded from karaf.
Second instance is replacing USM in SecurityModels.getInstance() which is causing users to be removed.
Although default and karaf should have different classes for SecurityModels.getInstance() sometimes they are same instance.
To solve this always instantiate classBasedStrategy on OpenNMS and don't register karaf instance.
When classBasedStrategy is not instantiable (on Minion), use the instance from karaf.

* JIRA: http://issues.opennms.org/browse/NMS-10009


j-white:
Merged to `foundation-2018`. Created https://issues.opennms.org/browse/NMS-10153 to track backport.

PR #1985, opened 2018-05-16
NMS-10152: make Backshift the default graphing strategy

RangerRick:
This PR makes Backshift the default.

* JIRA: http://issues.opennms.org/browse/NMS-10152


agalue:
I'm wondering what happens if someone accidentally misspelled the value for `org.opennms.rrd.strategyClass`. I believe that there should be some kind of default when the value is incorrect or unexpected; or enforce valid values by throwing a fatal error.

j-white:
+1 from me.

j-white:
@agalue I don't see how that relates to these changes.

RangerRick:
I mean, that's true, worth opening a new issue, but it's not a change in behavior.

agalue:
Actually it is a change in behavior, because the current implementation actually returns a valid strategy. With your change, if the value of the property is not auto, that will be exactly the returned strategy. I mean, if I set `org.opennms.rrd.strategyClass=blahblah`, with your change the strategy will be `blahblah`; while with the old code, it will be the default. Makes sense ?

j-white:
`org.opennms.rrd.strategyClass` doesn't affect the value of the graphing engine being used.

If `org.opennms.web.graphs.engine=blahblah`, then the resulting strategy will be `blahblah`, but this is the same behavior as it was prior to this patch.

PR #1986, opened 2018-05-17
NMS-9939: Revisit the current documentation of Drift

j-white:
JIRA: https://issues.opennms.org/browse/NMS-9939

Here we update the documentation for Elasticsearch and Flow Support (i.e. Drift).

It's still light on a few details, but it's a start :)


mvrueden:
There are actually a bunch more properties, we have not yet documented but should probably include here in the documentation.

j-white:
Which properties are you referring to? Those listed in `ga-elasticsearch-integration-flow-support`?

mvrueden:
I was refering to https://github.com/OpenNMS/opennms/blob/bdac58683457352b71af11a63c3118eb8b887f6b/features/flows/classification/engine/impl/src/main/resources/OSGI-INF/blueprint/blueprint.xml#L14-L21

fooker:
I've found some more things:
* The configuration section does not specify where/how to set the config
* The credentials properties are `globalElasticUser` and `globalElasticPassword`

j-white:
Merging this for inclusion in 22.0.0. If there's any additional feedback here we can always go back and update the relevant sections.

PR #1988, opened 2018-05-18
NMS-10153: Backport intermittent SNMPv3 failures

cgorantla:
There are two instances of Snmp4JStrategy, one from default classLoader and other loaded from karaf.
Second instance is replacing USM in SecurityModels.getInstance() which is causing users to be removed.
Although default and karaf should have different classes for SecurityModels.getInstance() sometimes they are same instance.
To solve this always instantiate classBasedStrategy on OpenNMS and don't register karaf instance.
When classBasedStrategy is not instantiable (on Minion), use the instance from karaf.

JIRA: http://issues.opennms.org/browse/NMS-10009
JIRA: http://issues.opennms.org/browse/NMS-10153


cgorantla:
This PR is mostly from except for smoke-test which are different for foundation-2016 https://github.com/OpenNMS/opennms/pull/1983

RangerRick:
👍

PR #1989, opened 2018-05-18
NMS-10148: statement correction for serviceDeleted description

mfuhrmann:
* JIRA: https://issues.opennms.org/browse/NMS-10148


mfuhrmann:
based on old fork. will create a new one.

PR #1993, opened 2018-05-24
NMS-10085: The auto-acknowledge-alarm tag with no content doesn't work on notifd-configfuration.xml

cgorantla:
* JIRA: http://issues.opennms.org/browse/NMS-10085


cgorantla:
Since  this is dependent on alarmd and opennms-services, there is a cyclic dependency between POMs so I created a smoke-test in favor of creating seperate project just for integration test.

j-white:
👍

PR #1994, opened 2018-05-24
HZN-1315: Split core/sink/* modules in separate client,server and common modules

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1315


core/ipc/sink/aws-sqs/client/pom.xml
```diff
@@ -0,0 +1,48 @@
...
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <extensions>true</extensions>
+        <configuration>
+          <instructions>
+            <Bundle-RequiredExecutionEnvironment>JavaSE-1.8</Bundle-RequiredExecutionEnvironment>
+            <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+            <Bundle-Version>${project.version}</Bundle-Version>
+            <Import-Package>
+              org.opennms.core.ipc.sink.camel.server;resolution:=optional,
```
j-white: Do we actually need this import?

mvrueden:
Yeah, testing the SQS manually is on my list, as well as verifying the imports.

core/ipc/sink/aws-sqs/client/pom.xml
mvrueden: Yes and no. 
The `Import-Package` statement is set to `resolution=optional` in order to not have it resolved, as this is used only when running inside OpenNMS (Client and server jars with Spring). 

However when running the client bundle in osgi, the import from `org.opennms.core.ipc.sink.camel.server` cannot be resolved, but the package is never used in this scenario anyways. 

In my opinion `!org.opennms.core.ipc.sink.camel.server` would be more fitted.

This is why I decided to use that pattern instead of `org.opennms.core.ipc.sink.camel.server;resolution:=optional`

mvrueden:
I verified the SQS Implementation manually and it seems to work.
The Heartbeat Sink is emptied and the RPC calls worked from opennms side.
Used `poller:poll` and `ping:ping` to verify. 

Let's wait on Bamboo to see how the other implementations are doing :)

j-white:
Looks good. I was going to merge it, but it looks like there's a conflict we need to address first.

mvrueden:
Yeah, thanks for the tip. Last time I checked there was no conflict. Should be mergeable now. Let's wait for Bamboo.

core/ipc/sink/kafka/common/pom.xml
```diff
@@ -0,0 +1,100 @@
...
+        <exclusion>
+          <artifactId>slf4j-log4j12</artifactId>
+          <groupId>org.slf4j</groupId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.scala-lang</groupId>
+      <artifactId>scala-library</artifactId>
+      <version>2.10.6</version>
```
j-white: This is currently at `2.11.11` in develop.

mvrueden: Thank you. Good catch (-:

cgorantla:
Looks good to me from the kafka refactoring and verified versions.

PR #1995, opened 2018-05-24
NMS-10165: Consolidate Kafka client library versions

cgorantla:
* JIRA: http://issues.opennms.org/browse/NMS-10165


j-white:
Since the patch is fairly simple I think it would be worthwhile targeting `foundation-2018` instead of `develop` with this one.

cgorantla:
We still may need this change as we explicitly reverted upgrade of kafka versions only for 22.0 not for 21.0

PR #1996, opened 2018-05-25
NMS-10169: Link to privacy policy from Data Choices UI elements

jeffgdotorg:
* JIRA: http://issues.opennms.org/browse/NMS-10169


j-white:
?

PR #2001, opened 2018-05-25
NMS-10159: new netsnmp sysStat metrics

mfuhrmann:
* Jira Issue: https://issues.opennms.org/browse/NMS-10159
* added 4 new metrics into netsnmp datacollection (described in Jira)
  * created a new datacollection group so StoreByGroup configuration won't damage their rrds.
* added new CPU stat and usage graphs (also used suppress function)
* updated the default load threshold to be more dynamic. Now it's been calculated on the amount of cpu cores and will fire if load > 70%


mfuhrmann:
oops... :-(

PR #2002, opened 2018-05-27
NMS-10159: Extend system statistic metrics from UCD SNMP agents 

mfuhrmann:
* Jira Issue: https://issues.opennms.org/browse/NMS-10159
* added 4 new metrics into netsnmp datacollection (described in Jira)
  * created a new datacollection group so StoreByGroup configuration won't damage their rrds.
* added new CPU stat and usage graphs (also used suppress function)
* updated the default load threshold to be more dynamic. Now it's been calculated on the amount of cpu cores and will fire if load > 70%


opennms-base-assembly/src/main/filtered/etc/snmp-graph.properties.d/netsnmp-graph.properties
```diff
@@ -228,6 +228,87 @@ report.netsnmp.cpuStatsFull.command=--title="CPU Statistics" \
  COMMENT:" " \
  LINE1:float15#cc0000:"15 minutes\\n"
 
+report.netsnmp.cpuStatsFuller.name=CPU Statistics
```
indigo423: The report `netsnmp.cpuStatsFuller` is missing on top of the file in the report section.

opennms-base-assembly/src/main/filtered/etc/snmp-graph.properties.d/netsnmp-graph.properties
```diff
@@ -310,6 +391,67 @@ report.netsnmp.cpuUsageFull.command=--title="CPU Usage" \
  GPRINT:cpuinterrupt:MIN:"Min  \\: %8.2lf %s" \
  GPRINT:cpuinterrupt:MAX:"Max  \\: %8.2lf %s\\n"
 
+report.netsnmp.cpuUsageFuller.name=CPU Usage
```
indigo423: The report `netsnmp.cpuUsageFuller` is missing on top of the file in the report section.

indigo423:
The `storeByGroup` feature in RRD make things ugly. To add new metrics we are forced to create new groups cause the existing ones are now immutable, which leads to this ... more and more things we have for now to live with :) This is a really nice enhancement.

mfuhrmann:
> Why do we create another graph cpuUsageFuller and not enhance the existing one?

@indigo423 
Not every agent (especially older ones) supports all of them. So we might need a new graph for nodes whose agents do support those new objects, and have it suppress the existing one.

It's the same in cpuUsageFull with CpuRawSoftIRQ metric.

Next time we have to call it: report.netsnmp.cpuUsageFullest :-1:

indigo423:
Thanks for the explanation. Do you know what version of the Net-SNMP agent is required?

mfuhrmann:
5.7.3., which is provided by Ubuntu >= 16.04 and Centos >= 7.

opennms-base-assembly/src/main/filtered/etc/snmp-graph.properties.d/netsnmp-graph.properties
mfuhrmann: fixed

opennms-base-assembly/src/main/filtered/etc/snmp-graph.properties.d/netsnmp-graph.properties
mfuhrmann: fixed

indigo423:
@mfuhrmann @RangerRick @agalue to follow the https://semver.org guidelines, this change should be a target for 22.1.0. It is an enhancement which is backward compatible. If we decide to merge into 22.0.1 we have to retarget the PR to release/22.0.1 instead of develop. It makes totally sense for develop which is target 23.0.0. In this case we could be more aggressive with the change and could refactor the ... more more more groups introduced with this change.

mvrueden:
I changed the label to `23.0.0`. If you want to merge it into different versions, please adjust base branch and label accordingly.

pbrane:
Have any screenshots of this new graph?

mfuhrmann:
@pbrane 
grafik
grafik
grafik
grafik

mfuhrmann:
@indigo423 How do we want to proceed?
Shall we remove the older two versions?
Or rename all three to something like: 1.gen, 2.gen, 3.gen?

mvrueden:
Approved by @indigo423. Merging.

PR #2003, opened 2018-05-29
HZN-1308: Move database-schema.xml into jar instead of having it in etc

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1308


j-white:
+1 from me

PR #2005, opened 2018-05-30
HZN-1309: Refactor opennms-server.xml to only reflect Path Outage configurations

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1309


j-white:
Is it even worth keeping the `opennms-server.xml` around? Should we migrate the remaining path outage configuration to the poller?

mvrueden:
Okay, I took another look at this and decided to move everything to the existing `poller-configuration.xml`. I also added a `PathOutageConfig` object, which is used in the  `PathOutageDao`.

PR #2006, opened 2018-06-01
NMS-10164 Support customizing the default time zone when parsing dates in syslog messages

patrick-schweizer:
* JIRA: https://issues.opennms.org/browse/NMS-10164

### DIFF
--- a/features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/Rfc5424SyslogParser.java
+++ b/features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/Rfc5424SyslogParser.java
@@ -114,11 +114,12 @@ public class Rfc5424SyslogParser extends SyslogParser {
         return message;
     }
 
-    protected static Date parseDate(final String dateString) {
+    protected Date parseDate(final String dateString) {
         if (dateString.endsWith("Z")) {
             try {
                 final DateFormat df = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'", Locale.ROOT);
                 df.setTimeZone(TimeZone.getTimeZone("GMT"));
+                adjustTimeZone(df);
                 return df.parse(dateString);
             } catch (final Exception e) {
                 // try again with optional decimals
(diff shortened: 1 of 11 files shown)


features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/Rfc5424SyslogParser.java
```diff
@@ -114,18 +114,20 @@ protected SyslogMessage parse() throws SyslogParserException {
         return message;
     }
 
-    protected static Date parseDate(final String dateString) {
+    protected Date parseDate(final String dateString) {
         if (dateString.endsWith("Z")) {
             try {
                 final DateFormat df = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'", Locale.ROOT);
                 df.setTimeZone(TimeZone.getTimeZone("GMT"));
```
patrick-schweizer: @j-white I see here an inconsistent behavior between Rfc5424SyslogParser.java (defaults to GMT if no default time zone is given) and SyslogParser.java (defaults to TimeZone.getDefault() == the time zone of the JVM). Should we leave this or also change Rfc5424SyslogParser.java to use TimeZone.getDefault()?

j-white: @patrick-schweizer Good catch. I would opt to change the default behavior of the `Rfc5424SyslogParser` as well to keep it consistent among parsers. Would like to hear what others think though.

mvrueden: I am usually for a consistent behaviour, unless it is not possible and then document why that is. However which implication a change of `Rfc5424SyslogParser` has, I cannot say.

pbrane:
I would have one of our support guys, Jeff probably, check with our customer in Savannah as they are probably the biggest user of the syslog feature.  We should also check with our customer in Kansas as they make extensive use of it as well.

Are these changes in the Radix parser?

j-white:
Given the level of uncertainty here, I'd say we keep the default behavior as is. When a specific timezone is set however, we should use that timezone for all of the parsers.

patrick-schweizer:
ok, that is the current behavior of the pull request. Feel free to merge ;-)

To sum the behavior up:

SyslogParser + date string with timezone => time zone of date string
SyslogParser + date string without timezone + no configured time zone => JVM time zone
SyslogParser + date string without timezone + with configured time zone => configured time zone

Rfc5424SyslogParser + date string with timezone => time zone of date string
Rfc5424SyslogParser + date string without timezone + no configured time zone => GMT time zone
Rfc5424SyslogParser + date string without timezone + with configured time zone => configured time zonezone

PR #2007, opened 2018-06-01
NMS-10172: Add karaf shell command for filtering nodes/interfaces

cgorantla:
* JIRA: http://issues.opennms.org/browse/NMS-10172

### DIFF
--- a/features/poller/shell/src/main/java/org/opennms/netmgt/filters/shell/FilterCommand.java
+++ b/features/poller/shell/src/main/java/org/opennms/netmgt/filters/shell/FilterCommand.java
@@ -1,0 +1,117 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2016-2018 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2018 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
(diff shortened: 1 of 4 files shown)


features/poller/shell/src/main/java/org/opennms/netmgt/filters/shell/FilterCommand.java
```diff
@@ -0,0 +1,116 @@
...
+                String foreignSource = (node.getForeignSource() != null) ? "" : "foreignSource=" + node.getForeignSource();
+                String location = nodeDao.getLocationForId(node.getId());
+                System.out.printf("nodeId=%d nodeLabel=%s location=%s %s %s %s \n", node.getId(), node.getLabel(),
+                        location, foreignId, foreignSource, categories);
+
+            }
+            List<String> ipAddresses = new ArrayList<>();
+            if (ipAddressList != null && !ipAddressList.isEmpty()) {
+                ipAddressList.stream().forEach(ipAddress -> ipAddresses.add(ipAddress.getHostAddress()));
+                System.out.printf("---Matching IP Addresses that are active---\n");
```
j-white: What do you think about showing the nodes followed by the associated IP addresses on the node that were matched rather than displaying these separately?

cgorantla: Makes sense.  Updating.

j-white:
Looks good to me. I'd wait until we get feedback from the requesters before merging though.

j-white:
Given that there isn't a better place to put the documentation right now I'll go ahead and merge this as-is.

PR #2008, opened 2018-06-01
HZN-1318: Database model for Alarm Situations

smith-opennms:
* JIRA: http://issues.opennms.org/browse/HZN-1318

Situations allow for roll up or correlation of multiple alarms.

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/create.sql
+++ b/opennms-base-assembly/src/main/filtered/etc/create.sql
@@ -1052,7 +1052,7 @@ ALTER TABLE memos ADD CONSTRAINT reductionkey_type_unique_constraint UNIQUE (red
 
 create table alarms (
     alarmID                 INTEGER, CONSTRAINT pk_alarmID PRIMARY KEY (alarmID),
-    dtype                   TEXT NOT NULL,
+    discriminator           TEXT NOT NULL,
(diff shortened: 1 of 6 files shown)


j-white:
Good start, +1 from me.

opennms-base-assembly/src/main/filtered/etc/create.sql
```diff
@@ -1122,6 +1123,12 @@ CREATE TABLE alarm_attributes (
 CREATE INDEX alarm_attributes_idx ON alarm_attributes(alarmID);
 CREATE UNIQUE INDEX alarm_attributes_aan_idx ON alarm_attributes(alarmID, attributeName);
 
+CREATE TABLE alarm_situations (
+    situation_id    INTEGER,
+    alarms_alarmid  INTEGER
+);
```
mvrueden: The changelog.xml and the create.sql show differences, e.g. some constraints are missing. 
This may result in tests pass, but in live scenario it may not work.
Can we add the missing restrictions here as well?

smith-opennms: Thanks. I've updated the test SQL to be in line with the liquibase schema.

mvrueden: 👍 Thank you

PR #2009, opened 2018-06-04
NMS-10173: Optionally persist the results when calling collectors:collect

cgorantla:
* JIRA: http://issues.opennms.org/browse/NMS-10173

### DIFF
--- a/features/collection/shell-commands/src/main/java/org/opennms/netmgt/collection/commands/CollectCommand.java
+++ b/features/collection/shell-commands/src/main/java/org/opennms/netmgt/collection/commands/CollectCommand.java
@@ -29,6 +29,8 @@
 package org.opennms.netmgt.collection.commands;
 
 import java.net.InetAddress;
+import java.nio.file.Paths;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


features/collection/shell-commands/src/main/java/org/opennms/netmgt/collection/commands/CollectCommand.java
```diff
@@ -77,6 +85,9 @@
     @Option(name = "-n", aliases = "--node", description = "Node ID or FS:FID", required = false, multiValued = false)
     String nodeCriteria;
 
+    @Option(name = "--persist", description = "Persist collection")
```
j-white: Since all the other options have both a name and an alias, should we not do the same here? `-p` seems like a natural fit.

features/collection/shell-commands/src/main/java/org/opennms/netmgt/collection/commands/CollectCommand.java
```diff
@@ -122,12 +141,32 @@ public Void execute() {
...
 
+        Persister persister = null;
+        if (persist) {
+            ServiceParameters params = new ServiceParameters(Collections.emptyMap());
+            RrdRepository repository = new RrdRepository();
+            persister = persisterFactory.createPersister(params, repository);
+            if (rras != null && rras.size() > 0) {
+                repository.setRraList(rras);
+            } else {
+                repository.setRraList(Lists.newArrayList(
```
j-white: Can we defined this list as a constant and share it with https://github.com/OpenNMS/opennms/blob/cdc8610caa2708ca5437daef2e537b0eaf409933/features/collection/shell-commands/src/main/java/org/opennms/netmgt/collection/commands/StressCommand.java#L201 instead of duplicate it here?

features/collection/shell-commands/src/main/java/org/opennms/netmgt/collection/commands/CollectCommand.java
cgorantla: yeah, I mistook that boolean option can only have name. It's not.

j-white:
Looks good to me. I'd wait until we get feedback from the requesters before merging though.

RangerRick:
@agalue @jeffgdotorg do you have feedback on this?  It's ready to merge.

cgorantla:
Thanks @agalue for verifying.

PR #2010, opened 2018-06-05
HZN-1311: Enable InterfaceToNodeCache to refresh itself

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1311

At this point I keep the existing behaviour as is, but enable the cache to refresh itself if so desired.

Shall we make it automatically refresh by default?

### DIFF
--- a/opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/InterfaceToNodeCacheDaoImpl.java
+++ b/opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/InterfaceToNodeCacheDaoImpl.java
@@ -33,9 +33,13 @@ import static org.opennms.core.utils.InetAddressUtils.str;
 import java.net.InetAddress;
 import java.util.Collections;
 import java.util.Objects;
+import java.util.Timer;
+import java.util.TimerTask;
 import java.util.TreeSet;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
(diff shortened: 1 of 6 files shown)


opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/InterfaceToNodeCacheDaoImpl.java
```diff
@@ -176,9 +182,40 @@ public int compareTo(final Value that) {
...
+    @PostConstruct
+    public void init() {
+        // we call this, to ensure the bean-initialization is blocked, until it is initialized
+        dataSourceSync();
+
+        if (refreshRate > 0) {
+            refreshTimer.schedule(new TimerTask() {
+                @Override
+                public void run() {
+                    dataSourceSync();
```
j-white: Should we also catch and log run-time exceptions here so that the timer continues to fire even if there is a temporary blip?

mvrueden: That is a good point, I guess that makes sense.

mvrueden:
What would be the default refresh-time? 30 seconds?

j-white:
In OpenNMS, I would have it default to something closer to 5 or 10 minutes.

mvrueden:
I made the cache refresh automatically every 5 minutes and this can be overriden via properties

j-white:
+1

PR #2011, opened 2018-06-08
HZN-1331: Consume events from Kafka

malateshs49:
This PR includes -

- Ability to consume and process generic events from Kafka inside eventd.
- We can now add threads, batchsize and queue size as part of eventd-configuration.xml now for eventsinkconsumer module(configurable).

Jira Link - https://issues.opennms.org/browse/HZN-1331


pbrane:
Could you please create an issue in our Jira describing this feature enhancement?  It's the only way we have to track the contributions, create release-notes, etc. etc.

Thanks,
David

malateshs49:
Please find the JIRA created for the new feature done -
https://issues.opennms.org/browse/BSHIFT-176
Please let me know if anything else needed.

features/events/api/src/main/java/org/opennms/netmgt/events/api/Events.java
```diff
@@ -0,0 +1,66 @@
...
+import javax.xml.bind.annotation.XmlElement;
+import javax.xml.bind.annotation.XmlRootElement;
+
+import org.opennms.core.ipc.sink.api.Message;
+import org.opennms.netmgt.xml.event.Event;
+
+import com.google.common.primitives.Bytes;
+
+@XmlRootElement(name = "event")
+public class Events implements Message {
```
j-white: Would it be possible to use existing `org.opennms.netmgt.xml.event.Log` and `org.opennms.netmgt.xml.event.Event` objects instead of defining new ones?

opennms-base-assembly/src/main/filtered/etc/eventd-configuration.xml
```diff
@@ -1,2 +1,2 @@
-<EventdConfiguration xmlns="http://xmlns.opennms.org/xsd/config/eventd" TCPAddress="127.0.0.1" TCPPort="5817" UDPAddress="127.0.0.1" UDPPort="5817" receivers="5" getNextEventID="SELECT nextval('eventsNxtId')" socketSoTimeoutRequired="yes" socketSoTimeoutPeriod="3000">
+<EventdConfiguration xmlns="http://xmlns.opennms.org/xsd/config/eventd" TCPAddress="127.0.0.1" TCPPort="5817" UDPAddress="127.0.0.1" UDPPort="5817" receivers="5" getNextEventID="SELECT nextval('eventsNxtId')" socketSoTimeoutRequired="yes" socketSoTimeoutPeriod="3000" threads="300">
```
j-white: Can we avoid modifying this file? Other sink modules default to (2*number of cores) for the number of threads when none are set.

features/events/api/src/main/java/org/opennms/netmgt/events/api/Events.java
malateshs49: Yes, org.opennms.netmgt.xml.event.Log is a very good option. Rather than converting logs at receiving end i would change these accordingly by updating it.Thanks for the info

opennms-base-assembly/src/main/filtered/etc/eventd-configuration.xml
malateshs49: Yes thats the main what we have, this field is like configurable like how syslogd-configuration.xml and trapd-configuration.xml has where we can even make threads configurable.

malateshs49:
@j-white : Sure i will take care of the copyright year changes and will update the code and push it back.I just followed previous copyright.

features/events/api/src/main/java/org/opennms/netmgt/events/api/Events.java
malateshs49: @j-white : Sorry jesse i got carried away.why new Events class was written because org.opennms.netmgt.xml.event.Log or org.opennms.netmgt.xml.event.Event doesn't implement org.opennms.core.ipc.sink.api.Message hence had to write new Events class in order to use MessageConsumer

j-white: Can we just implement those interfaces on the existing objects?

features/events/api/pom.xml
```diff
@@ -83,6 +83,10 @@
       <artifactId>hibernate-validator-annotation-processor</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.opennms.core.ipc.sink</groupId>
+      <artifactId>org.opennms.core.ipc.sink.xml</artifactId>
```
j-white: We should change this to depend on the `org.opennms.core.ipc.sink.api` instead, and add a dependency on `org.opennms.core.ipc.sink.xml` in the `features/events/daemon/pom.xml` where the module is implemented.

features/events/api/src/main/java/org/opennms/netmgt/events/api/EventsWrapper.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.opennms.core.ipc.sink.api.Message;
+import org.opennms.netmgt.xml.event.Log;
+
+public class EventsWrapper implements Message {
+
+    private final Log eventLog;
+
+    public EventsWrapper(Log eventLog)
+            throws IllegalArgumentException {
```
j-white: We never actually throw a `IllegalArgumentException`.

opennms-config-model/src/main/java/org/opennms/netmgt/config/eventd/EventdConfiguration.java
```diff
@@ -127,6 +127,36 @@
 
     @XmlValue
     private String m_contents;
+    
+    
+    /**
+     * Number of threads used for consuming/dispatching messages.
+     * Defaults to 2 x the number of available processors.
+     */
+    @XmlAttribute(name = "threads", required=false)
```
j-white: Can we prefix these new attribute with `sink-` to make it clear that these settings are used for the Sink API consumer?

opennms-config-model/src/main/java/org/opennms/netmgt/config/eventd/EventdConfiguration.java
```diff
@@ -162,6 +192,14 @@ public Integer getUDPPort() {
     public void setUDPPort(final Integer UDPPort) {
         m_udpPort = ConfigUtils.assertNotNull(UDPPort, "UDPPort");
     }
+    
+    public Optional<Integer> getNumThreads() {
+    	return Optional.ofNullable(m_threads);
+    }
+
+    public void setNumThreads(final Integer numThreads) {
+        m_threads = ConfigUtils.assertNotNull(numThreads, "threads");
```
j-white: Why do we not allow this field to be null?

features/events/daemon/src/main/java/org/opennms/netmgt/eventd/sink/EventsModule.java
```diff
@@ -0,0 +1,136 @@
...
+			}
+
+			@Override
+			public Object key(EventsWrapper message) {
+				return message.getEvents();
+			}
+
+			@Override
+			public Log aggregate(Log accumulator, EventsWrapper eventsWrapper) {
+				accumulator = eventsWrapper.getEvents();
```
j-white: The  aggregate method should add the given object to the existing accumulator, and create a new one if necessary. Here we are ignoring the existing accumulator and just using the log in the message.

The `EventsWrapper`  probably be changed to wrap a single event, and this message should updated to roll those up into the `Log`.

opennms-config-model/src/main/java/org/opennms/netmgt/config/eventd/EventdConfiguration.java
malateshs49: we followed the legacy of how it is handled and described in syslogd-configuration.xml and trapd-configuration.xml for example https://github.com/OpenNMS/opennms/blob/develop/opennms-base-assembly/src/main/filtered/etc/trapd-configuration.xml

features/events/daemon/src/main/java/org/opennms/netmgt/eventd/sink/EventsModule.java
malateshs49: this has been addressed.

opennms-config-model/src/main/java/org/opennms/netmgt/config/eventd/EventdConfiguration.java
malateshs49: We have maintained same legacy of opennms i.e syslogd it self which is actually kept as not null  for example https://github.com/OpenNMS/opennms/blob/develop/opennms-config-model/src/main/java/org/opennms/netmgt/config/syslogd/Configuration.java .If we need to change then we are ready to update it.

features/events/api/src/main/java/org/opennms/netmgt/events/api/EventsWrapper.java
malateshs49: our old files had illegalArgument the same came off here.

opennms-config-model/src/main/java/org/opennms/netmgt/config/eventd/EventdConfiguration.java
j-white: In those cases all of the traps/syslogs are sourced from the Sink API, so there's no ambiguity as to what threads we're configuring. In the case of eventd there are a number of existing threads pools, and various sources of events. Prefixing these attributes will help make whats being configured clearer.

opennms-config-model/src/main/java/org/opennms/netmgt/config/eventd/EventdConfiguration.java
j-white: In the case referenced we perform validation on *all* of the attributes, whereas here we only verify that the `threads` field is not null. The behavior is not consistent.

features/events/daemon/src/main/java/org/opennms/netmgt/eventd/sink/EventsModule.java
```diff
@@ -0,0 +1,140 @@
...
+import org.opennms.core.ipc.sink.api.AggregationPolicy;
+import org.opennms.core.ipc.sink.api.AsyncPolicy;
+import org.opennms.core.ipc.sink.xml.AbstractXmlSinkModule;
+import org.opennms.netmgt.config.api.EventdConfig;
+import org.opennms.netmgt.events.api.EventsWrapper;
+import org.opennms.netmgt.xml.event.Log;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class EventsModule extends AbstractXmlSinkModule<EventsWrapper, Log> {
```
j-white: Since the `Log` implements the `Message` interface, it doesn't look like there's any need for the `EventsWrapper` wrapper.

That said, I think we'd be better off using an `Event` as the initial message rather than a `Log` here. Otherwise we need to create a `Log` every time we need to send a single event.

This won't change anything for your specific use case, but will make it easier to use when dispatching messages directly with the Sink API.

opennms-base-assembly/src/main/filtered/etc/eventd-configuration.xml
```diff
@@ -1,2 +1,2 @@
-<EventdConfiguration xmlns="http://xmlns.opennms.org/xsd/config/eventd" TCPAddress="127.0.0.1" TCPPort="5817" UDPAddress="127.0.0.1" UDPPort="5817" receivers="5" getNextEventID="SELECT nextval('eventsNxtId')" socketSoTimeoutRequired="yes" socketSoTimeoutPeriod="3000">
+<EventdConfiguration xmlns="http://xmlns.opennms.org/xsd/config/eventd" TCPAddress="127.0.0.1" TCPPort="5817" UDPAddress="127.0.0.1" UDPPort="5817" receivers="5" getNextEventID="SELECT nextval('eventsNxtId')" socketSoTimeoutRequired="yes" socketSoTimeoutPeriod="3000" sink-threads="300">
```
j-white: Can we omit the entry for `sink-threads` here? We have a default value for this attribute.

opennms-config-model/src/main/java/org/opennms/netmgt/config/eventd/EventdConfiguration.java
```diff
@@ -127,6 +127,36 @@
 
     @XmlValue
     private String m_contents;
+    
+    
+    /**
+     * Number of threads used for consuming/dispatching messages.
+     * Defaults to 2 x the number of available processors.
+     */
+    @XmlAttribute(name = "sink-threads", required=false)
```
j-white: Can we prefix **all** the sink related settings we added with the 'sink-' prefix?

features/events/daemon/src/main/java/org/opennms/netmgt/eventd/sink/EventsModule.java
malateshs49: Yup i would update to event from log rather than doing to log at sender side its better to reuse and do at receiver under sendnowsync as usual.However we need eventswrapper class because MessageConsumer<S extends Message, T extends Message> uses two Message argument and we cant use Events or Log as first one since both return Events and in aggregate we cant iterate over Events and get single event or we cant hardcode to return first event .Hence EventsWrapper is used.

features/events/daemon/src/main/java/org/opennms/netmgt/eventd/sink/EventsModule.java
```diff
@@ -0,0 +1,136 @@
...
+		return MODULE_ID;
+	}
+
+	@Override
+	public int getNumConsumerThreads() {
+		return m_config.getNumThreads();
+	}
+
+	@Override
+	public AggregationPolicy<EventsWrapper, Event, Event> getAggregationPolicy() {
```
j-white: In the latest update we switched from using the `Log` to the `Event` for *both* types - this was not the intent of my last comment.

In the current incarnation the `Event` class implements the `Message` interface so there is no need for the  `EventsWrapper`. The `AggregationPolicy` also does not perform any aggregation, so it could be replaced with an instance of the `IdentityAggregationPolicy`. The batch size / interface configurations are also no longer required.

I suggest that we define the module as:
```
public class EventsModule extends AbstractXmlSinkModule<Event, Log> 
```
and update the aggregation policy to append the events to the log, creating a new one when necessary.

features/events/api/src/main/java/org/opennms/netmgt/xml/event/Event.java
```diff
@@ -56,14 +56,15 @@
...
 import org.opennms.core.network.InetAddressXmlAdapter;
 import org.opennms.core.utils.StringUtils;
 import org.opennms.netmgt.events.api.DateTimeAdapter;
 
 @XmlRootElement(name = "event")
 @XmlAccessorType(XmlAccessType.FIELD)
 //@ValidateUsing("event.xsd")
-public class Event implements Serializable {
+public class Event implements Message,Serializable {
         private static final long serialVersionUID = 6997816158234653400L;
```
RangerRick: If you're changing the interfaces it implements, you need to update the `serialVersionUID`.

features/events/api/src/main/java/org/opennms/netmgt/xml/event/Log.java
```diff
@@ -50,7 +52,7 @@
 @XmlRootElement(name="log")
 @XmlAccessorType(XmlAccessType.FIELD)
 //@ValidateUsing("event.xsd")
-public class Log implements Serializable {
+public class Log implements Message,Serializable {
     private static final long serialVersionUID = 8526177705077223094L;
```
RangerRick: here too

features/events/api/src/main/java/org/opennms/netmgt/events/api/EventsWrapper.java
```diff
@@ -0,0 +1,51 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2016 The OpenNMS Group, Inc.
```
RangerRick: Make sure you update the copyrights in the files you create/change if possible, as well.

features/events/api/src/main/java/org/opennms/netmgt/events/api/EventsWrapper.java
```diff
@@ -0,0 +1,51 @@
...
+ * @author ms043660 (Malatesh.Sudarshan@cerner.com)
+ */
+package org.opennms.netmgt.events.api;
+
+import java.util.Objects;
+
+import org.opennms.core.ipc.sink.api.Message;
+import org.opennms.netmgt.xml.event.Event;
+
+public class EventsWrapper implements Message {
```
j-white: We can remove this class, since it is no longer used.

j-white:
Thanks again for the updates. Pushed to `pr/jira/HZN-1331` for a Bamboo run.

j-white:
It looks like the `EventdConfigurationTest` is failing with these changes. See https://bamboo.opennms.org/browse/OPENNMS-ONMS2283-UNIT-2

j-white:
That test passed, but there are a few other integration tests failing now.

The system tests are also failing, it looks like there is a error is karaf.log that prevents some of the OSGi features from bootstrapping.

malateshs49:
@j-white : I am confused how does this change impact there.Because we built RPM and did certification testing as well.

malateshs49:
@j-white : You can merge changes from this branch to PR-1331 and let me know if bamboo succeeds.
@cgorantla : Thanks for the quick help.Probably we had old develop code from which we had updated and did PR and later on these new things we added to develop which was missed and caused failure and again Thanks for the quick help.

integration-tests/config/src/test/java/org/opennms/netmgt/config/WillItUnmarshalIT.java
```diff
@@ -220,7 +220,7 @@ private static void addFile(final Source source, final String file, final Class<
         addFile(Source.CONFIG, "email-northbounder-configuration.xml", EmailNorthbounderConfig.class, true, null);
         addFile(Source.CONFIG, "enlinkd-configuration.xml", EnlinkdConfiguration.class, false, null);
         addFile(Source.CONFIG, "eventconf.xml", Events.class, true, null);
-        addFile(Source.CONFIG, "eventd-configuration.xml", EventdConfiguration.class, true, null);
+        addFile(Source.CONFIG, "eventd-configuration.xml", EventdConfiguration.class, false, null);
```
j-white: We shouldn't need to disable this check here. If the test is failing, we need to rework how the defaults are handled in the JAXB objects.

cgorantla:
@malateshs49  We should add integration-test and docs for this feature. You can refer to [these](https://github.com/OpenNMS/opennms/tree/develop/core/ipc/sink/kafka/itests/src/test/java/org/opennms/core/ipc/sink/kafka/itests) tests

opennms-config/src/main/java/org/opennms/netmgt/config/EventdConfigManager.java
```diff
@@ -242,4 +242,48 @@ public String getGetNextEventID() {
             getReadLock().unlock();
         }
     }
+    
+    @Override
+    public int getNumThreads() {
+        getReadLock().lock();
+        try {
+            if (m_config.getNumThreads() <= 0) {
```
cgorantla: I think this condition should be opposite.

RangerRick:
I just noticed I had old "requested changes" that never got approved so this was still showing the red check.  @cgorantla you've been managing this PR, right?  I'll leave this to you, I'm just doing some going through of the current queue.  :)

cgorantla:
@malateshs49  I have added few more changes in cg/jira/HZN-1331 branch and bamboo is passing.
Please merge them in your branch to reflect in PR.

cgorantla:
@j-white  I think most of the issues are addressed now. Can you please review one more time

cgorantla:
Closing this PR as we merged code from cg/jira/HZN-1331

PR #2012, opened 2018-06-10
NMS-9202 Event-Analysis Report shows incorrect numbers for big values in Top25 Events

patrick-schweizer:
* JIRA: https://issues.opennms.org/browse/NMS-9202


mvrueden:
From a "code" perspective this looks good to me.
Are the "width changes" necessary?

Let's wait on @indigo423 to verify this before merging.

patrick-schweizer:
@mvrueden thanks for reviewing. The width changes are necessary to display the tally numbers fully, otherwise they are cut off, see also my new comment in the ticket.

PR #2016, opened 2018-06-12
NMS-8524: The ReST API used to return XMLs with namespace, and now it doesn't

patrick-schweizer:
* JIRA: https://issues.opennms.org/browse/NMS-8524


patrick-schweizer:
@agalue I added your suggestion. Could you check again? Thank you!

agalue:
Looks good, thanks.

PR #2017, opened 2018-06-12
NMS-9923: Upgrade Drools to 7.7.0.FINAL

UberPinguin:
JIRA: https://issues.opennms.org/browse/NMS-9923

Add KIE dependencies required by the 7.x release series.
Avoid a ConcurrentModificationException in org.opennms.features.events.api.

### DIFF
--- a/features/events/api/src/main/java/org/opennms/netmgt/xml/event/Event.java
+++ b/features/events/api/src/main/java/org/opennms/netmgt/xml/event/Event.java
@@ -1883,7 +1883,8 @@ public class Event implements Serializable {
 		if (_dbid   != null) builder.append("dbid", _dbid);
 		if (_source != null) builder.append("source", _source);
 		if (_nodeid != null) builder.append("nodeid", _nodeid);
-		if (_parms  != null) builder.append("parms", _parms);
+		// Copy the _parms array instead of referencing it, to avoid Concurrent Modification Exceptions
+		if (_parms  != null) builder.append("parms",  new ArrayList<>(_parms));
 		return builder.toString();
 	}
 
--- a/dependencies/drools/pom.xml
+++ b/dependencies/drools/pom.xml
@@ -15,7 +15,7 @@
     can depend on to use the drools library.
   </description>
   <properties>
-    <droolsVersion>6.5.0.Final</droolsVersion>
+    <droolsVersion>7.7.0.Final</droolsVersion>
   </properties>
   <dependencies>
     <dependency>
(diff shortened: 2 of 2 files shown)


j-white:
In NMS-9923, there's a note that reloading the engines doesn't work with 7.6.0, is this still the case?

features/events/api/src/main/java/org/opennms/netmgt/xml/event/Event.java
```diff
@@ -1883,7 +1883,7 @@ public String toStringSimple() {
 		if (_dbid   != null) builder.append("dbid", _dbid);
 		if (_source != null) builder.append("source", _source);
 		if (_nodeid != null) builder.append("nodeid", _nodeid);
-		if (_parms  != null) builder.append("parms", _parms);
+        if (_parms  != null) builder.append("parms",  new ArrayList<>(_parms));
```
j-white: Do you have a copy of the stack trace for the `ConcurrentModificationException` that this resolves?

UberPinguin: I don't, anymore. It was on paste.opennms.eu, but is no more.

The engine reload integration tests work; I'm getting an install set up to see if it works "live".
The engines at work may not be the best test case for this, since they're not all refactored to use CEP yet.

UberPinguin: For example, in NMS-9907, I reported that persistence doesn't work with 6.5.0, either. But I'm not 100% confident that it isn't my rule-sets to blame.

UberPinguin:
In a live OpenNMS install, I've tested:
* Reload engine with a change to an existing rule
* Reload engine with a new rule
* Reload engine with both a change to an existing rule and adding a new rule

In all 3 cases, Working Memory was persisted across reloads, and the rule changes were activated on reload.

UberPinguin:
Dependency changes introduced by upgrading Drools to 7.x: 
* `antlr-runtime-3.5` -> `antlr-runtime-3.5.2`
* `drools-compiler-6.5.0.Final` -> `drools-compiler-7.7.0.Final`
* `drools-core-6.5.0.Final` -> `drools-core-7.7.0.Final`
* `kie-api-6.5.0.Final` -> `kie-api-7.7.0.Final`
* `kie-internal-6.5.0.Final` -> `kie-internal-7.7.0.Final`
* `mvel2-2.2.8.Final` -> `mvel2-2.4.0.Final`
* `xstream-1.4.9` -> `xstream-1.4.10`

New dependencies:
* `kie-soup-commons-7.7.0.Final`
* `kie-soup-maven-support-7.7.0.Final`
* `kie-soup-project-datamodel-api-7.7.0.Final`
* `kie-soup-project-datamodel-commons-7.7.0.Final`

features/events/api/src/main/java/org/opennms/netmgt/xml/event/Event.java
j-white: Can we add a comment here that explains why we're copying the `_params` array? (since it's not obvious)

j-white:
Thanks for the tests and details on the dependencies. Given the nature of the bug this resolves, I think we should look at targeting `foundation-2018` instead of `develop` with this update.

UberPinguin:
OK. Should I abandon this PR and submit a new one to that branch?

features/events/api/src/main/java/org/opennms/netmgt/xml/event/Event.java
UberPinguin: A comment that explains the `new ArrayList<>(_parms)`, or one that explains why it's even being copied into the `ToStringBuilder` object in the first place?

UberPinguin:
Comment added, and rebased onto foundation-2018.

j-white:
Pushed to `pr/jira/NMS-9923`.

PR #2025, opened 2018-06-19
HZN-1316: fix heatmap when included on the front page

RangerRick:
This PR fixes the heatmap when it's included on the front page (with, eg, `org.opennms.web.console.centerUrl=/heatmap/heatmap-box.jsp`).  It also very likely fixes some other race-conditions in things attempting to use jQuery UI, since there was an issue with the bootstrap js included globally that could step on the initialized jQuery.

* JIRA: http://issues.opennms.org/browse/HZN-1316


mvrueden:
Looks reasonable to me, I haven't tried it, though.

PR #2026, opened 2018-06-20
HZN-1326: Alarmd support for Situations

smith-opennms:
This PR provides support for Alarmd to create and update Situations from events received with the one or more parameters with name 'related-reductionKey'. e.g.: 
     -p 'related-reductionKey uei.value1' 

Each 'related-reductionKey' that resolves to an existing Alarm will be grouped under the Situation.

* JIRA: https://issues.opennms.org/browse/HZN-1326

### DIFF
--- a/opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
+++ b/opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
@@ -33,7 +33,11 @@ import java.util.Arrays;
 import java.util.Calendar;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
 import java.util.concurrent.locks.Lock;
+import java.util.stream.Collectors;
 
 import org.hibernate.Hibernate;
 import org.opennms.netmgt.dao.api.AlarmDao;
@@ -44,8 +48,10 @@ import org.opennms.netmgt.events.api.EventForwarder;
 import org.opennms.netmgt.model.OnmsAlarm;
 import org.opennms.netmgt.model.OnmsEvent;
 import org.opennms.netmgt.model.OnmsSeverity;
+import org.opennms.netmgt.model.Situation;
 import org.opennms.netmgt.model.events.EventBuilder;
 import org.opennms.netmgt.xml.event.Event;
+import org.opennms.netmgt.xml.event.Parm;
 import org.opennms.netmgt.xml.event.UpdateField;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -135,7 +141,7 @@ public class AlarmPersisterImpl implements AlarmPersister {
             if (LOG.isDebugEnabled()) {
                 LOG.debug("addOrReduceEventAsAlarm: reductionKey:{} not found, instantiating new alarm", reductionKey);
             }
-            alarm = createNewAlarm(e, event);
+            alarm = createNewAlarm(e, event, m_alarmDao);
 
             //FIXME: this should be a cascaded save
             m_alarmDao.save(alarm);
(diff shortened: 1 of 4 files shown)


opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
```diff
@@ -263,11 +269,31 @@ private void reduceEvent(OnmsEvent e, OnmsAlarm alarm, Event event) {
...
+                LOG.warn("reduceEvent: Event {} attempts to add alarms to alarm that is not a Situation: {}.", event.getUei(), alarm);
+            }
+        }
+
         e.setAlarm(alarm);
     }
 
-    private static OnmsAlarm createNewAlarm(OnmsEvent e, Event event) {
-        OnmsAlarm alarm = new OnmsAlarm();
+    private static OnmsAlarm createNewAlarm(OnmsEvent e, Event event, AlarmDao alarmDao) {
```
j-white: Can we make this non-static instead of passing the AlarmDao as an argument?

opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
```diff
@@ -294,6 +320,23 @@ private static OnmsAlarm createNewAlarm(OnmsEvent e, Event event) {
         return alarm;
     }
     
+    private static Set<OnmsAlarm> getAlarms(List<Parm> list, AlarmDao alarmDao) {
```
j-white: Similarly comment here. Will save us from having to pass the DAO around.

opennms-alarms/daemon/src/test/java/org/opennms/netmgt/alarmd/AlarmdIT.java
```diff
@@ -278,48 +283,58 @@ public void run() {
...
+        //there should be no alarms in the alarm_situations table
+        assertEmptyAlarmSituationTable();
 
-        Integer alarmId = m_jdbcTemplate.queryForObject("select alarmId from alarms where reductionKey = ?", new Object[] { reductionKey }, Integer.class).intValue();
-        rowCount = m_jdbcTemplate.queryForObject("select count(*) from events where alarmid = ?", new Object[] { alarmId }, Integer.class).intValue();
-        MockUtil.println(String.valueOf(rowCount) + " of events with alarmid: "+alarmId);
-        //      assertEquals(numberOfAlarmsToReduce, rowCount);
+        //create 3 alarms to roll up into situation
+        sendNodeDownEvent("Alarm1", node);
+        Thread.sleep(1000);
```
j-white: I'd opt to use `await().until()` for checking asynchronous updates instead of relying to fixed sleep times in this new test. 1) tests tend to finish faster and 2) they work more reliably

opennms-alarms/daemon/src/test/java/org/opennms/netmgt/alarmd/AlarmdIT.java
```diff
@@ -278,48 +283,58 @@ public void run() {
...
-            m_jdbcTemplate.query("select eventid from events where alarmID is not null", new RowCallbackHandler() {
-                @Override
-                public void processRow(ResultSet rs) throws SQLException {
-                    reducedEvents.add(rs.getInt(1));
-                }
-            });
-            Collections.sort(reducedEvents);
+        int alarmCount = m_alarmDao.findAll().size();
+        assertEquals("there should be 1 alarm", 1, alarmCount);
+        OnmsAlarm alarm = m_alarmDao.findByReductionKey(reductionKey);
```
j-white: Thanks for updating the other tests to use the DAOs too.

opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
```diff
@@ -263,11 +269,31 @@ private void reduceEvent(OnmsEvent e, OnmsAlarm alarm, Event event) {
...
 
+        Set<OnmsAlarm> containedAlarms = getAlarms(event.getParmCollection(), m_alarmDao);
+        if (containedAlarms != null && !containedAlarms.isEmpty()) {
+            if (alarm instanceof Situation) {
+                for(OnmsAlarm related : containedAlarms) {
+                    // Related Alarms are additive for Situations reduced from Events.
+                    ((Situation)alarm).addAlarm(related);
+                }
+            } else {
+                LOG.warn("reduceEvent: Event {} attempts to add alarms to alarm that is not a Situation: {}.", event.getUei(), alarm);
```
j-white: Would it be possible to change the type to a situation in this case? I could foresee cases where we wanted to associate alarms with an existing alarm rather than group these under a new situation.

pbrane: Can alarms be related to more than one situation?

smith-opennms: @pbrane Currently, yes. We do not explicitly prevent or constrain that behavior.

smith-opennms: @j-white Spent some time looking into conversion. It is tricky to convert on the fly to the sub-type. (We probably need to use a native query in a transaction).
If adding Alarms to existing alarms is the behavior we want, I think we would be best to remove the Alarm-Situation object hierarchy and just have relatedAlarms as an optional attribute of OnmsAlarm.

j-white: Make sense.  Given that this is not easy to change, let's just keep the current behavior. If the use case is strong enough, we'll come back and revisit this.

PR #2028, opened 2018-06-23
NMS-10072: Normalize date formats across the WebUI

patrick-schweizer:
* JIRA: https://issues.opennms.org/browse/NMS-10072

This is a proof of concept and base for discussions. Please do not merge yet.

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/opennms.properties
+++ b/opennms-base-assembly/src/main/filtered/etc/opennms.properties
@@ -363,6 +363,10 @@ org.eclipse.jetty.server.Request.maxFormKeys=2000
 #
 #opennms.web.base-url = https://%x%c/
 
+# Choose the format how the date is displayed in the UI. It needs to follow the pattern in SimpleDateFormat:
+# see https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html
+# org.opennms.ui.datettimeformat=yyyy-MM-dd'T'kk:mm:ssxxx
+
 ###### ASTERISK AGI SUPPORT ######
 # If you are using the built-in Asterisk Gateway Interface (AGI) server
 # to integrate with an external Asterisk server, you will need to set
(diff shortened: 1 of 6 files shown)


patrick-schweizer:
@agalue , @j-white please take also a look at the comments in: https://issues.opennms.org/browse/NMS-10072

opennms-base-assembly/src/main/filtered/etc/opennms.properties
```diff
@@ -363,6 +363,10 @@ org.eclipse.jetty.server.Request.maxFormKeys=2000
 #
 #opennms.web.base-url = https://%x%c/
 
+# Choose the format how the date is displayed in the UI. It needs to follow the pattern in SimpleDateFormat:
+# see https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html
+# org.opennms.ui.datettimeformat=yyyy-MM-dd'T'kk:mm:ssxxx
```
j-white: Thinking ahead here... what should we do for Javascript based pages that want to render dates using this format as well?

If we stick to using Java's SimpleDataFormat it may be difficult to render the same date: https://www.beyondjava.net/date-patterns-cheatsheet

mvrueden: For javascript I'd opt for using an UTC timestamp of the date and let the client worry about timezone conversions. This would also work for jsp's as we could probably read the requester's locale and apply a conversion.

patrick-schweizer: +1 on @mvrueden suggestion on the Javascript.
for the JSPs: currently we have different behaviors / implementations (see my comments in the ticket). However the most common use case is the usage of <fmt:formatDate which takes the user locale in consideration. About time zones: user locale !=time zone. I am not clear yet what time zone we want to display: the one of the current user, the one of the server or even UTC/GMT? Locales and time zones are by no means a trivial subject and before changing anything we should be very clear on what we want. I wonder what is the best way to get to a common understanding? I think we should also separate our thinking process into 2 parts: 1. what do we want (from a product owner perspective) 2. how to we get there (technically).
@mvrueden , @j-white , @agalue I will ping you on mattermost...

patrick-schweizer:
closing this PR. It's content was moved to: https://github.com/OpenNMS/opennms/pull/2033
Reason: I broke NMS-10072 up into sub tasks

PR #1972, opened 2018-05-14
HZN-1307: Initial Sentinel work

mvrueden:
This PR implements changes to allow the telemetry daemon to be run within a standalone container (sentinel).

This is still WIP but I wanted to give everyone a chance to provide input early on.

# Stuff left TODO
- Remove TODOs
- provide documentation
- Make stuff nicer (e.g. the opennms-dataschema.xml and opennms-server.xml exposition below)
- Provide some kind of statistics/diagnostics, e.g. how many sentinels are started at the moment (similar to the minion heartbeat maybe)
- is `features/minion/core/repository` AND `features/minion/repository` needed, or can we merge this to `features/minion/repository`. 
- rework features definitions, especially all `opennms-telemetry-*` features to see if they are actually used or must be rewritten/removed

# How to setup a sentinel

## Sentinel Part
Ensure Java is installed: `java -version`.

Ensure JAVA_HOME is set: `echo $JAVA_HOME`, e.g. `export JAVA_HOME=/usr/lib/jvm/jdk-1.8.0/`

Start a raw apache-karaf-4.1.5 container and add the following repositories:
```
repo-add mvn:org.apache.karaf.features/spring/4.1.5/xml/features
repo-add mvn:org.opennms.karaf/opennms/22.0.0-SNAPSHOT/xml/karaf-extensions
repo-add mvn:org.opennms.karaf/opennms/22.0.0-SNAPSHOT/xml/features
repo-add mvn:org.opennms.karaf/opennms/22.0.0-SNAPSHOT/xml/sentinel
```

Afterwards, install the following features:
```
# Scv is required for sentinel-jms
feature:install scv
# is required to read application-context files
feature:install opennms-spring-extender
# Is required for sentinel:ping, the Identity and possibility to configure the sentinel 
feature:install sentinel-core
# Required to load configurations remotely via HTTP from OpenNMS and 
# basically opennms-dao and opennms-dao-api including custom datasource configuration (see below)
feature:install sentinel-persistence

feature:install sentinel-jms
feature:install opennms-core-ipc-sink-camel-server
feature:install sentinel-telemetry
feature:install sentinel-flows
```

Afterwards a custom adapter can be started, like so:
```
admin@minion()> config:edit org.opennms.features.telemetry.adapters-netflow5
admin@minion()> config:property-set name Netflow-5
admin@minion()> config:property-set class-name org.opennms.netmgt.telemetry.adapters.netflow.v5.Netflow5Adapter
admin@minion()> config:update
```
[...]

PR #2030, opened 2018-06-25
HZN-1307: Make flow*-adapters run inside sentinel.

mvrueden:
**NOTE:** This PR is still Work in Progress. 
I'd like to do some more cleanup and testing before merging, but want to give everyone a chance to have a look before merging.

This PR implements changes to allow the telemetry daemon to be run within a standalone _Sentinel_ container and is based on changes from PR https://github.com/OpenNMS/opennms/pull/1972.
In order to make a review easier, in case you already reviewed the old PR, new changes are starting at commit 72b5e72c780c380fdc846c9813c7238733b7404b

Besides the original PR, it addresses the following issues mentioned there:
 * Some changes are now done in `develop` and have been merged forward instead of doing them in `jira/HZN-1307`
 * Basic documentation was added
 * JMS and Kafka system tests are now provided

# Stuff left TODO
- Remove TODOs
- provide documentation
- Provide some kind of statistics/diagnostics, e.g. how many sentinels are started at the moment (similar to the minion heartbeat maybe)
- is `features/minion/core/repository` AND `features/minion/repository` needed, or can we merge this to `features/minion/repository`. 
- rework features definitions, especially all `opennms-telemetry-*` features to see if they are actually used or must be rewritten/removed

# General Structure

**core/spring-extender**
This module is similar to the functionality from the gemini-blueprint project, but does not integrate with karaf that heavily. This module just listens for bundles with “Spring-Context” META-INF definitions and starts an application context from that bundle.
This allows us to re-use the functionality of gemini-blueprint, but are not required to remove aries-blueprint.

Another difference to gemini-blueprint is, that this bundle supports loading an spring application context from itself (required to bootstrap the ServiceRegistry)

Why do we need this?
Almost all modules which are at the moment only visible to the jetty classloader heavily rely on Spring. However those modules will not work out of the box inside a standalone karaf. To allow us to re-use the `applicationContext-*.xml` and also `component-*.xml` files, the `geminini-blueprint` project is used. 

In order to get it to work, some `applicationContext-*.xml` files have to be splitted. The idea is the following:
* e.g. `applicationContext-dao.xml` is loaded by OpenNMS
* e.g. `component-dao.xml` is loaded when running inside OpenNMS AND a distributable container such as sentinel or minion
 * e.g. `applicationContext-osgi.xml` is loaded when running inside a distributable container

A usual `applicationContext-osgi.xml` file would look something like this:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<beans xmlns="http://www.springframework.org/schema/beans"
       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
       xmlns:context="http://www.springframework.org/schema/context"
       xmlns:tx="http://www.springframework.org/schema/tx"
       xmlns:osgi="http://www.springframework.org/schema/osgi"
       xmlns:onmsgi="http://xmlns.opennms.org/xsd/spring/onms-osgi"
       xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-4.2.xsd
          http://www.springframework.org/schema/context http://www.springframework.org/schema/context/spring-context-4.2.xsd
          http://www.springframework.org/schema/tx http://www.springframework.org/schema/tx/spring-tx-4.2.xsd
          http://www.springframework.org/schema/osgi http://www.springframework.org/schema/osgi/spring-osgi.xsd
          http://xmlns.opennms.org/xsd/spring/onms-osgi http://xmlns.opennms.org/xsd/spring/onms-osgi.xsd">

     <!-- NOTE: Contains only Minion/Sentinel-specific beans -->

    <!-- Resolve references, which are usually provided by the OpenNMS Runtime Environment -->
    <osgi:reference id="serviceRegistry" interface="org.opennms.core.soa.ServiceRegistry" />

</beans>
```
The idea is, that it pulls in all required dependencies in order to load the bundle successfully. 
In most cases that is probably only the `ServiceRegistry`.

**features/sentinel**
All implementations specific to sentinel

**features/minion**
All implementations specific to minion

**container/features**
Contains all feature definitions for minion, sentinel, opennms

**features/distributed**
All modules which are shared between distributed containers, such as sentinel and minion.
A sub module is not intended to run inside the opennms’ karaf container.

**features/distributed/dao-api**
A module which merges (with maven-shade-plugin) all DAO APIs into one uber dao-api bundle.
For example the API modules of the persistence from bsm and classification as well as opennms-dao-api are merged here.

**features/distributed/dao-impl**
A module which merges all DAO implementations into one uber dao-impl bundle.
As all DAOs require the sessionFactory to be instantiated.
Until now the SessionFactory class was accessible via the system bundle. 
Therefore we need all DAO implementations to be in one bundle.
(The reason for this is, that the SessionFactory is not an interface).

An alternative to this could be to implement a SessionFactoryProvider interface which is used by the DAOs instead.
However this requires more code changes.

**features/distribution/dao-distributed**
This module contains DAO implementations which need to be slightly different than the ones shipped by OpenNMS.
At the moment only the DistPollerDAO is implemented here, to provide a custom “whoami()” implementation and throws a bunch of UnsupportedOperationExceptions for unsupported operations.

**features/distribution/config**
Contains the ConfigProxyAnnotationBeanPostProcessor which is required to proxy jaxb configuration objects

**features/distributed/datasource**
Exposes possibliity to configure the opennms datasource via an `org.opennms.netmgt.distributed.datasource.cfg` file.

# Additional Changes
Besides the general structure mentioned above, the following changes have been implemented:

A new `Identity` interface was introduced to also add a `SystemType` (`Minion` | `Sentinel`). This `Identity` is only used by the new `DistPollerDaoDistributed` (former `DistPollerDaoMinion`).
The `MinionIdentity` interface is still available. In the long run, the `MinionIdentity` should be replaced with the new Identity concept.

Everything from `features/minion/core` was moved to `features/distributed/core` and is re-used by sentinel

At the moment each distributable container must provide at least the following services:
`Identity` (and optionally the `MinionIdentity` in case of a minion container). The `RestConfig` and if required an implementation of the `MinionPingCommand`/`SentinelPingCommand`. Refer to bundles `features/sentinel/core` or `features/minion/core/core-impl` for more details.


mvrueden:
FYI, There is a docker image you can use, see : 
- http://hub.docker.com/r/opennms/sentinel/
- https://github.com/opennms-forge/docker-sentinel

Documentation is a bit short for now.

opennms-doc/guide-admin/src/asciidoc/text/sentinel/sentinel.adoc
```diff
@@ -0,0 +1,305 @@
...
+NOTE:   Basically the same properties as for the _Minion Controller_ are supported, but must be placed in a different file name.
+        `org.opennms.sentinel.controller.cfg` vs `org.opennms.minion.controller.cfg`.
+
+
+===== Install features
+
+The first step is to make the features known to the container:
+
+```
+repo-add mvn:org.apache.karaf.features/spring/4.1.5/xml/features
```
j-white: Should we not include the feature repositories as part of the default configuration?

The set of features to install will depend on the particular deployment, but the repositories will always be needed.

mvrueden: LoL, yes, you are right. I added these repositories probably 500 times during testing and didn't really question this anymore, until I read your comment and 🤦‍♂️ (-:

mvrueden: Okay, after revisiting this, I now know why it is the way it is: There is no easy AND nice way to address this. 
The container is built as follows:
 * `container/shared` provides the basic distribution and is shared by minion and sentinel
 * `features/sentinel/repository` defines the system repository for the sentinel container
 * `features/container/sentinel` defines overrides of the shared container for sentinel and also builds the final distribution
 * `opennms-assemblies/sentinel` does some stuff on top of `features/container/sentinel`, which is not important here

The problem with this structure is, that the feature repositories are already defined in `container/shared` by the `karaf-maven-plugin`. However those must be overriden in module `features/container/sentinel`. But the `karaf-maven-plugin` does not support this functionality. 

Therefore one way to address this could be to place a `org.apache.karaf.features.cfg` file in `features/container/sentinel` containing the feature repositories for sentinel. But this means, any change to `container/shared` must be manually adapted to `features/container/sentinel`.

Another approach can be to NOT build a custom `system` directory, but the `default`repository and place everything in `repositories/default` instead of `system`. This allows us to use a `features.uri` file and place all custom repositories there. The `karaf-extender` takes care of the rest.

The last approach is similar to the previous one, except, we enhance the `karaf-extender` to load also a `features.uri` and `feature.boot` file from the system directory or from the `features.boot.d` directory. 

At the moment I am leaning forward the last approach, to enhance the `karaf-extender` to make use of the `features.uri` either in the `system` or `etc/features.boot.d` directory.

j-white: I'd opt to just drop in a custom `org.apache.karaf.features.cfg` file and avoid using/enhancing the `karaf-extender`.

mvrueden:
As nothing `severe` came up, I am going forward and merge this. All other things can be addressed in a different PR, as the focus of this was to get procession of flows working in a single karaf container, which this addresses.

PR #2031, opened 2018-06-26
HZN-1336: Fix packet logging issue

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1336


j-white:
Nice catch. Since this affects 22.0.0 and 22.0.1, can we re-target this to `release-22.0.2`?

mvrueden:
I always mix up the release version of `drift` and `sentinel`.
As you said, `release-22.0.2` should be targeted here.

PR #2032, opened 2018-06-27
HZN-1316: fix heatmap for reals

RangerRick:
My last fix for this worked for me, but I think it was just luck/a race condition.  This PR updates webpack to a newer version that contains some fixes for how code splitting happens, and then also lets me put the "runtime chunk" (a bit of webpack loader logic that avoids double-loading resources) in the `vendor.js` file.  This should fix any initialization issues with jQuery bits in the UI.  It also makesa  change to the `AssetLocator` to handle the webpack asset plugin generating asset manifest entries for "anonymous" resources, ie, resources that aren't named explicitly as webpack endpoints, but which end up in the bundle(s).

I also cleaned up some commented out stuff from the `webpack.config.js` file, and fixed a few lint warnings in a few recently-changed js files.

* JIRA: http://issues.opennms.org/browse/HZN-1316


christianpape:
Works!

PR #2034, opened 2018-06-29
NMS-10238: match behavior of all test data from C IPLIKE

RangerRick:
These changes do a number of things to fix the behavior of matching
to more closely follow the C version of IPLIKE.  It also includes
a few "best practices" fixes.

* do simple matching of scope identifiers
* use 'if X is not true' rather than 'if not X' to avoid incorrect matches when NULL is returned
* handle wildcards in hex rule checks properly

* JIRA: http://issues.opennms.org/browse/NMS-10238


j-white:
Is there any test coverage for this function? If so, it might help to add cases for the updated behavior.

RangerRick:
That's why I marked this triage, I'm still working on that part.  :)

RangerRick:
Tests passed so this is ready to merge, I've removed triage from the list.

PR #2048, opened 2018-07-09
HZN-1324: RPC over Kafka

cgorantla:
Use kafka as message bus for RPC.

* JIRA: http://issues.opennms.org/browse/hzn-1324


container/features/src/main/resources/features.xml
```diff
@@ -662,6 +662,12 @@
       <bundle>mvn:org.opennms.core.ipc.rpc/org.opennms.core.ipc.rpc.aws-sqs-impl/${project.version}</bundle>
     </feature>
 
+    <feature name="opennms-core-ipc-rpc-kafka" description="OpenNMS :: Core :: IPC :: RPC :: Kafka." version="${project.version}">
```
j-white: Nit: Name should be `OpenNMS :: Core :: IPC :: RPC :: Kafka Impl.` to follow the naming convention of the other IPC implementations.

core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcClientFactory.java
```diff
@@ -0,0 +1,311 @@
...
+                String marshalledRequest = module.marshalRequest(request);
+                // Generate RPC Id for every request to track request/response.
+                String rpcId = UUID.randomUUID().toString();
+                // Calculate timeout based on ttl and default timeouts.
+                long metadataRefreshRate = Long.parseLong(TOPICS_METADATA_MAX_AGE);
+                long ttl = request.getTimeToLiveMs() != null ? request.getTimeToLiveMs().longValue() : 0;
+                // timeout for RPC > metadataRefreshRate, add default timeout to metadata refresh rate.
+                long timeout = metadataRefreshRate + Long.parseLong(TIMEOUT_FOR_KAFKA_RPC);
+                // ttl should be > default timeout otherwise use default.
+                timeout = ttl < timeout ? timeout : ttl;
```
j-white: Nit: Can be replaced with `timeout = Math.max(timeout,ttl);`

core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcClientFactory.java
```diff
@@ -0,0 +1,311 @@
...
+                    RpcMessageProtos.RpcMessage rpcMessage = RpcMessageProtos.RpcMessage.newBuilder()
+                                                                 .setRpcId(rpcId)
+                                                                 .setSystemId(request.getSystemId())
+                                                                 .setRpcContent(ByteString.copyFromUtf8(marshalledRequest))
+                                                                 .build();
+                    List<PartitionInfo> partitionInfo = producer.partitionsFor(requestTopic);
+                    partitionInfo.stream().forEach(partition -> {
+                        // Use systemId as key.
+                        final ProducerRecord<String, byte[]> record = new ProducerRecord<>(requestTopic,
+                                partition.partition(), request.getSystemId(), rpcMessage.toByteArray());
```
j-white: Is there are specific reason we use the systemId as the key here? Do we just need *some* key in order to be able to specify the partition, or does it have some special meaning?

core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcClientFactory.java
```diff
@@ -0,0 +1,311 @@
...
+        //Start a new thread which handles timeouts from delayQueue and calls response callback.
+        final ThreadFactory threadFactory = new ThreadFactoryBuilder().setNameFormat("rpc-client-timeout-tracker-%d")
+                .build();
+        Executors.newSingleThreadExecutor(threadFactory).execute(() -> {
+            while (true) {
+                try {
+                    ResponseCallback responseCb = delayQueue.take();
+                    responseCb.sendResponse(null);
+                } catch (InterruptedException e) {
+                    // exit, nothing to do
```
j-white: I'd log some info level message here to indicate the thread is stopping.

core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcClientFactory.java
```diff
@@ -0,0 +1,311 @@
...
+                }
```
j-white: I'd also add a generic `catch(Exception e)`  clause here - if an exception is throws while issuing a single callback we don't want to thread to abort and stop processing the queue.

core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcClientFactory.java
```diff
@@ -0,0 +1,311 @@
...
+
+        public KafkaConsumerRunner(KafkaConsumer<String, byte[]> consumer) {
+            this.consumer = consumer;
+        }
+
+        @Override
+        public void run() {
+            try {
+                // subscribe to topics with "rpc-response" as part of topic name.
+                Pattern pattern = Pattern.compile(".*rpc-response.*");
```
j-white: Won't this match topics that have 'rpc-response' *anywhere* in the name?

core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcServerManager.java
```diff
@@ -0,0 +1,280 @@
...
+                        String minionId = getMinionIdentity().getId(); 
+                        if (hasSystemId && !(minionId.equals(rpcMessage.getSystemId()))) {
+                            // directed RPC and not directed at this minion
+                            LOG.info("MinionIdentity {} doesn't match with systemId {}, ignore the request", minionId,
+                                    rpcMessage.getSystemId());
+                            continue;
+                        } 
+                        RpcRequest request = module.unmarshalRequest(rpcMessage.getRpcContent().toStringUtf8());
+                        if (hasSystemId) {
+                            // directed RPC, there may be more than one request with same request Id, cache and allow only one.
```
j-white: What's the purpose of the value in the cache here? The RPC IDs should be unique, so provided there is an entry we shouldn't need to process it again.

We can also perform this logic prior to unmarshaling the request - this will save some cycles if we choose to ignore the request.

core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcClientFactory.java
```diff
@@ -0,0 +1,311 @@
...
+                } else {
+                    RpcMessageProtos.RpcMessage rpcMessage = RpcMessageProtos.RpcMessage.newBuilder()
+                                                                .setRpcId(rpcId)
+                                                                .setRpcContent(ByteString.copyFromUtf8(marshalledRequest))
+                                                                .build();
+                    final ProducerRecord<String, byte[]> record = new ProducerRecord<>(requestTopic,
+                            rpcMessage.toByteArray());
+                    producer.send(record);
+                }
+                LOG.debug("RPC Request {} sent to minion at location {}", request.toString(), request.getLocation());
```
j-white: We can skip the `.toString()` here - it will be called if necessary.

core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcServerManager.java
```diff
@@ -0,0 +1,280 @@
...
+                                final JmsQueueNameFactory topicNameFactory = new JmsQueueNameFactory("rpc-response",
+                                        module.getId());
+                                RpcMessageProtos.RpcMessage rpcResponse = RpcMessageProtos.RpcMessage.newBuilder()
+                                                                              .setRpcId(rpcId)
+                                                                              .setRpcContent(ByteString.copyFromUtf8(responseAsString))
+                                                                              .build();
+                                final ProducerRecord<String, byte[]> producerRecord = new ProducerRecord<>(
+                                        topicNameFactory.getName(), rpcResponse.toByteArray());
+                                producer.send(producerRecord);
+                                LOG.debug("request {} executed, sending response {} ", request.toString(),
```
j-white: We can skip the `.toString()` here too.

core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcClientFactory.java
```diff
@@ -0,0 +1,311 @@
...
+                consumer.subscribe(pattern);
+                while (!closed.get()) {
+                    ConsumerRecords<String, byte[]> records = consumer.poll(Long.MAX_VALUE);
+                    for (ConsumerRecord<String, byte[]> record : records) {
+                        RpcMessageProtos.RpcMessage rpcMessage = RpcMessageProtos.RpcMessage.parseFrom(record.value());
+                        // Get Response callback from rpcId and send rpc content to callback.
+                        ResponseCallback responseCb = rpcResponseMap.get(rpcMessage.getRpcId());
+                        if (responseCb != null) {
+                            responseCb.sendResponse(rpcMessage.getRpcContent().toStringUtf8());
+                        }
```
j-white: We should log something when we receive a response for an entry that can not be found in the map.

core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcClientFactory.java
cgorantla: yeah, ProducerRecord needs some "key" when we specify partition

core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcClientFactory.java
cgorantla: sure, makes sense. I will add that

core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcClientFactory.java
cgorantla: Yeah, if there are multiple instances of OpenNMS and only kafka instance, this may be a problem.
 I will add instance name to regex match

pbrane:
Nice work!

j-white:
I setup a local environment to test this using the following settings:
```
echo "org.opennms.activemq.broker.disable=true
org.opennms.core.ipc.rpc.strategy=kafka
org.opennms.core.ipc.rpc.kafka.bootstrap.servers=127.0.0.1:9092
org.opennms.core.ipc.sink.strategy=kafka
org.opennms.core.ipc.sink.kafka.bootstrap.servers=127.0.0.1:9092" >> $OPENNMS_HOME/etc/opennms.properties.d/kafka.properties
```

I also patched the runInPlace.sh script so that the Minions would spawn up using Kafka instead of JMS:
```
diff --git a/features/minion/runInPlace.sh b/features/minion/runInPlace.sh
index 2d15f028751..fbed51f28fe 100755
--- a/features/minion/runInPlace.sh
+++ b/features/minion/runInPlace.sh
@@ -102,6 +102,16 @@ set_instance_specific_configuration() {
     * ) MINION_ID="test-$idx"
   esac
   echo "id=$MINION_ID" > "$MINION_HOME/etc/org.opennms.minion.controller.cfg"
+
+  echo "!opennms-core-ipc-sink-camel
+!opennms-core-ipc-rpc-jms
+!minion-jms
+opennms-core-ipc-sink-kafka
+opennms-core-ipc-rpc-kafka" >> "$MINION_HOME/etc/featuresBoot.d/kafka.boot"
+  echo "bootstrap.servers=127.0.0.1:9092
+acks=1" >> "$MINION_HOME/etc/org.opennms.core.ipc.rpc.kafka.cfg"
+  echo "bootstrap.servers=127.0.0.1:9092
+acks=1" >> "$MINION_HOME/etc/org.opennms.core.ipc.sink.kafka.cfg"
 }
 
 spawn_minion() {
```

The set of resulting topics in Kafka looked like:
```
./bin/kafka-topics.sh --list --zookeeper 127.0.0.1
OpenNMS.MINION.rpc-request.Collect
OpenNMS.MINION.rpc-request.DNS
OpenNMS.MINION.rpc-request.Detect
OpenNMS.MINION.rpc-request.Echo
OpenNMS.MINION.rpc-request.PING
OpenNMS.MINION.rpc-request.PING-SWEEP
OpenNMS.MINION.rpc-request.Poller
OpenNMS.MINION.rpc-request.Requisition
OpenNMS.MINION.rpc-request.SNMP
OpenNMS.MINION1.rpc-request.Poller
OpenNMS.Sink.Heartbeat
OpenNMS.Sink.Trap
OpenNMS.rpc-response.Collect
OpenNMS.rpc-response.DNS
OpenNMS.rpc-response.Echo
OpenNMS.rpc-response.Poller
```

I increased the number of partitions using:
```
./bin/kafka-topics.sh --alter --topic OpenNMS.* --partition 16 --zookeeper 127.0.0.1
```

It performed well and directed RPCs worked as expected.

j-white:
I did however run into an issue with TTL handling.

If the TTL on a request is set to 30s and the request actually takes 40s to process. The request is completed successfully, whereas I would expect it to have timed out.

This can be reproduced by running:
```
rpc:stress --location MINION --ttl 40000 --delay 50000 --count 100
```

I suspect this may be due to the way we update the TTL to take the metadata refresh interval into account, but haven't confirmed.

core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcServerManager.java
```diff
@@ -0,0 +1,278 @@
...
+
+        @Override
+        public void run() {
+            try {
+                consumer.subscribe(Arrays.asList(topic));
+                LOG.info("subscribed to topic {}", topic);
+                while (!closed.get()) {
+                    ConsumerRecords<String, byte[]> records = consumer.poll(Long.MAX_VALUE);
+                    for (ConsumerRecord<String, byte[]> record : records) {  
+                        RpcMessageProtos.RpcMessage rpcMessage = RpcMessageProtos.RpcMessage.parseFrom(record.value());
```
j-white: We should have a mechanism to validate whether or not the TTL has already expired on the request, and if this is the case to avoid processing it preemptively.

We could do this by including an additional field in the message, something like `expiresAt` with `$now + ttl` and comparing that to the current time here.

core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcClientFactory.java
j-white: Thinking about this some more... I think we should use the message UUID as the partition key for all messages in both the request and response topic. This should have a uniform distribution and will allow external consumers to easily join both streams to match requests and responses.

j-white:
With respect to the TTLs, I think it's important that we respect the TTL in the given request, and ensure that the callback is done in a timely fashion.

As a way around the meta-data timeout issue, we can manually specify the list of topics against which the consumer should subscribe and update these dynamically.
 
For example, we could call`kafkaConsumerRunner.startConsumingForModule(module.getId())` before we publish a request.

The implementation could look something like:
```
public void startConsumingForModule(String moduleId) {
    // Index the topics by module id to avoid having to recompute the topic
	// name on every call
	if (moduleIdsForTopics.contains(moduleId)) {
		// We're already consuming, or have already added this module to the
		// list of topics from which to consume
		return;
	}

	final JmsQueueNameFactory topicNameFactory = new JmsQueueNameFactory("rpc-response",
			moduleId);
	synchronized (topics) {
		if (topics.add(topicNameFactory.getName())) {
			topicAdded.set(true);
			firstTopicAdded.countDown();
			consumer.wakeup();
		}
	}
}
```

And then the run loop could be modified with:
```
Logging.putPrefix(RpcClientFactory.LOG_PREFIX);
// Wait for the first topic to be added to the set
if (topics.isEmpty()) {
	while (!closed.get()) {
		try {
			firstTopicAdded.await(1, TimeUnit.SECONDS);
			LOG.debug("First topic added. Ready to start consuming.");
		} catch (InterruptedException e) {
			LOG.info("Interrupted before first topic was added. Terminating Kafka RPC consumer thread.");
			return;
		}
	}
}

while (!closed.get()) {
	if (topicAdded.get()) {
		synchronized (topics) {
			// Topic subscriptions are not incremental. This list will replace the current assignment (if there is one).
			LOG.info("Subscribing Kafka RPC consumer to topics named: {}", topics);
			consumer.subscribe(topics);
			topicAdded.set(false);
		}
	}
	
	try {
	    final ConsumerRecords<String, byte[]> records = consumer.poll(Long.MAX_VALUE);
...			
```

cgorantla:
@j-white  cool. This approach should allow consumer to subscribe dynamically without overwriting metadata interval. Thanks for detailed explanation

core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcServerManager.java
```diff
@@ -0,0 +1,283 @@
...
+                                String responseAsString = null;
+                                try {
+                                    responseAsString = module.marshalResponse(response);
+                                    final JmsQueueNameFactory topicNameFactory = new JmsQueueNameFactory("rpc-response",
+                                            module.getId());
+                                    RpcMessageProtos.RpcMessage rpcResponse = RpcMessageProtos.RpcMessage.newBuilder()
+                                            .setRpcId(rpcId).setRpcContent(ByteString.copyFromUtf8(responseAsString))
+                                            .build();
+                                    final ProducerRecord<String, byte[]> producerRecord = new ProducerRecord<>(
+                                            topicNameFactory.getName(), rpcResponse.toByteArray());
```
j-white: We should use the RPC ID as the key for the responses too.

core/ipc/rpc/kafka/src/main/proto/rpc.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto2";
+option java_package = "org.opennms.core.ipc.rpc.kafka.model";
+option java_outer_classname = "RpcMessageProtos";
+
+message RpcMessage {
+required string rpc_id = 1;
+required bytes  rpc_content = 2;
+optional uint64 expiresAt = 3;
```
j-white: Can we avoid mixing camelCase and under_scores here?

core/ipc/rpc/kafka/src/test/java/org/opennms/core/ipc/rpc/kafka/RpcKafkaIT.java
```diff
@@ -0,0 +1,308 @@
...
+        request.setLocation(REMOTE_LOCATION_NAME);
+        EchoResponse expectedResponse = new EchoResponse("Kafka-RPC");
+        EchoResponse actualResponse = echoClient.execute(request).get();
+        assertEquals(expectedResponse, actualResponse);
+    }
+
+    @Test(timeout = 30000)
+    public void testKafkaRpcAtRemoteLocation() throws InterruptedException, ExecutionException {
+        assertNotEquals(REMOTE_LOCATION_NAME, identity.getLocation());
+        Thread.sleep(5000);
```
j-white: Why do we need to sleep for 5 seconds here?

core/ipc/rpc/kafka/src/test/java/org/opennms/core/ipc/rpc/kafka/RpcKafkaIT.java
```diff
@@ -0,0 +1,308 @@
...
+        count.set(0);
+        // Send 100 requests.
+        int maxRequests = 100;
+        for (int i = 0; i < maxRequests; i++) {
+            sendRequestAndVerifyResponse(request, 0);
+        }
+        // Try to consume response messages to test single request on minion.
+        kafkaConfig.put(ConsumerConfig.GROUP_ID_CONFIG, "testMinionCache");
+        kafkaConfig.put(ConsumerConfig.METADATA_MAX_AGE_CONFIG, "1000");
+        kafkaConfig.put("key.deserializer", StringDeserializer.class.getCanonicalName());
```
j-white: These can be constants too.

cgorantla:
Thanks @j-white for the extensive review. I  handled all of the review comments. Please have an another look.

j-white:
I gave this another try and everything appears to be working well now. Thanks for your resilience on this one.

PR #2049, opened 2018-07-10
NMS-10255: Updated asciidoctor 1.5.6 -> 1.5.7 and stylesheet

indigo423:
Update the asciidoctor dependency from 1.5.6 to 1.5.7. Requires also to bump the JRuby dependency from 9.0.4.0 to 9.1.16.0. Also updated the asciidoctor default stylesheet using Roboto / Roboto Mono fonts.

* JIRA: http://issues.opennms.org/browse/NMS-10255


j-white:
+1

PR #2052, opened 2018-07-11
NMS-10242: Add Validation for Trapd, Snmp configs.

cgorantla:
* JIRA: http://issues.opennms.org/browse/NMS-10242


j-white:
I'm assuming that the `snmp-trap-port` attribute is defined as required?

Should this PR be targeted for `foundation-2017` instead of `develop`?

cgorantla:
yeah, snmp-trap-port is defined as required.  Don't want to change config files in order to assign a defalut value and make it optional.

PR #2053, opened 2018-07-12
HZN-1352: Allow any Alarm to be a Situation

d5smith:
This PR removes the Situation sub class from the persistence model.
Now any OnmsAlarm can be a Situation simply by having relatedAlarms

* JIRA: https://issues.opennms.org/browse/HZN-1352


features/rest/mapper/src/main/java/org/opennms/web/rest/mapper/v2/AlarmMapper.java
```diff
@@ -89,8 +88,8 @@ protected void fillAlarm(OnmsAlarm alarm, @MappingTarget AlarmDTO alarmDTO) {
         if (alarm.getTTicketId() != null && !alarm.getTTicketId().isEmpty() && ticketUrlTemplate != null) {
             alarmDTO.setTroubleTicketLink(getTicketUrl(alarm.getTTicketId()));
         }
-        if (alarm instanceof Situation) {
-            alarmDTO.setRelatedAlarms(((Situation)alarm).getAlarms().stream()
+        if (alarm.isSituation()) {
```
j-white: Is there any need for this explicit check here? If it's not a situation then `getRelatedAlarms()` must return an empty collection.

opennms-model/src/main/java/org/opennms/netmgt/model/OnmsAlarm.java
```diff
@@ -1153,4 +1151,37 @@ public Date getAckTime() {
...
+        return m_relatedAlarms;
+    }
+
+    /**
+     * <p>setDetails</p>
+     *
+     * @param alarms a {@link java.util.Set} object.
+     */
+    public void setRelatedAlarms(Set<OnmsAlarm> alarms) {
+        this.m_relatedAlarms = alarms;
```
j-white: Nit: `this.` is redundant when used with a variable prefixed with `m_`.

features/rest/mapper/src/main/java/org/opennms/web/rest/mapper/v2/AlarmMapper.java
smith-opennms: minimally. it only matters if the serialization does not include null fields and then the empty list  is not sent over the wire.
I will remove.

opennms-model/src/main/java/org/opennms/netmgt/model/OnmsAlarm.java
smith-opennms: I'm not a fan of either - will remove.

PR #2054, opened 2018-07-12
NMS-10261: Ensure that newSuspect events get processed if the monitoring system cannot be found

j-white:
JIRA: https://issues.opennms.org/browse/NMS-10261


pbrane:
I’m unsure this is the way to handle this.

j-white:
Me too. What would you suggest doing instead?

j-white:
After some discussion we've agreed that this is the best approach to take currently. We can revisit this if any further problems arise.

PR #2055, opened 2018-07-12
NMS-10264: Only increment counter if alarm type is not of type resolution

pbrane:

Please read our [Contribution Guidelines](https://github.com/OpenNMS/opennms/blob/develop/CONTRIBUTING.md) and format the title of the pull request in the format of:

${JIRA-ISSUE-NUMBER}: a-short-subject-title

Please use the [JIRA](https://issues.opennms.org) issue number and create a link in the JIRA issue back to this pull request so we have a quick reference from the issue to the pull request.

### DIFF
--- a/opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
+++ b/opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
@@ -47,6 +47,7 @@ import org.opennms.netmgt.model.OnmsAlarm;
 import org.opennms.netmgt.model.OnmsEvent;
 import org.opennms.netmgt.model.OnmsSeverity;
 import org.opennms.netmgt.model.Situation;
+import org.opennms.netmgt.xml.event.AlarmData;
 import org.opennms.netmgt.xml.event.Event;
 import org.opennms.netmgt.xml.event.Parm;
 import org.opennms.netmgt.xml.event.UpdateField;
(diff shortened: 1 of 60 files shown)


opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
```diff
@@ -154,7 +154,10 @@ private void reduceEvent(OnmsEvent e, OnmsAlarm alarm, Event event) {
         // Always set these
         alarm.setLastEvent(e);
         alarm.setLastEventTime(e.getEventTime());
-        alarm.setCounter(alarm.getCounter() + 1);
+        
+        if (event.getAlarmData().getAlarmType().equals(alarm.getAlarmType())) {
```
j-white: I'd change this to ` if (Objects.equals(event.getAlarmData().getAlarmType(), alarm.getAlarmType())) {` to help guard against NPEs.

pbrane: Sure.

pbrane:
When we change the configuration to have a single alarm vs. the current default way of having a "clearing alarm", the counter increments when the clearing event is reduced to the alarm.  We should only increment the counter if the alarm repeats.  So, if the alarm-type changes, we shouldn't increment that counter.

I also need to verify that the alarm-type gets updated on reduction.  I'll do that with the NPE guards you suggested.

pbrane:
https://issues.opennms.org/browse/NMS-10264

j-white:
Just to validate the current logic:
* If an alarm has type 1, and we send another related event with type 2, we want to change the alarm's type from 1 to 2?
* If an alarm has type 2, and we send multiple events, would we not want to increment the count for each of these?

pbrane:
If we send a clearing event (alarm type 2), we update the alarm type, logmsg, and not the count.  If the problem reoccurs (alarm type 1), we increment the counter, update the logmsg and set the alarm type to back to problem (alarm type 1).

j-white:
With these changes, if we send 2 clear events the count will always be 1.

Can you clarify why we want to change the alarm type on reduction? Shouldn't we use the severity to track whether or not an alarm has "cleared"?

RangerRick:
You're gonna have to rename this branch to something shorter if you want it to pass in bamboo. 

`... /file-----home-bamboo-ci-bamboo-xml-data-build-dir-OPENNMS-ONMS2316-RPM-opennms-source-target-rpm-BUILD-opennms-23.0.0-0.20180716.onms2316.features.drools4alarms.betteralarmcounthandling.2-features-ticketing-otrs-integration-src-main-wsdl-TicketService.wsdl (File name too long)`

pbrane:
We should only be tracking counts of problems.  When we move to having a single alarm vs. the current crappy two alarm situation, this makes more sense.

opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
```diff
@@ -154,74 +155,87 @@ private void reduceEvent(OnmsEvent e, OnmsAlarm alarm, Event event) {
         // Always set these
         alarm.setLastEvent(e);
         alarm.setLastEventTime(e.getEventTime());
-        alarm.setCounter(alarm.getCounter() + 1);
 
-        if (!event.getAlarmData().hasUpdateFields()) {
+        AlarmData alarmData = event.getAlarmData();
+        if (alarmData != null) {
```
j-white: The event will be dropped earlier if this is null, so there's no need to check this again here - see  `checkEventSanityAndDoWeProcess`

opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
```diff
@@ -154,74 +155,87 @@ private void reduceEvent(OnmsEvent e, OnmsAlarm alarm, Event event) {
...
-                if (fieldName.equalsIgnoreCase("LogMsg") && !field.isUpdateOnReduction()) {
-                    continue;
+            Integer newAlarmType = alarmData.getAlarmType();
+            Integer currentAlarmType = alarm.getAlarmType();
+            if (newAlarmType != null && currentAlarmType != null) {
+                if (!newAlarmType.equals(OnmsAlarm.RESOLUTION_TYPE)) {
+                    alarm.setCounter(alarm.getCounter() + 1);
                 } else {
-                    alarm.setLogMsg(e.getEventLogMsg());
+                    alarm.setAlarmType(newAlarmType);
```
j-white: It looks like an alarm can be updated from type 1 to type 2, but it won't get updated back to type 1 if the problem reoccurs.

opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
pbrane: But asked in previous code change to do Null checking on these same objects.

opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
pbrane: Fixed in latest commit.

opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
j-white: I had recommended changing the line in question to `if (Objects.equals(event.getAlarmData().getAlarmType(), alarm.getAlarmType())) {`.

Using `Objects.equals` in this case would only help prevent an NPE if/when the first call to `getAlarmType()` returned a null value.

RangerRick:
FYI, over on [the other version of this branch](https://bamboo.opennms.org/browse/OPENNMS-ONMS2321-UNIT-2/log) (that passes the wsdl length thing) there's currently a test failure:

```
17-Jul-2018 10:31:26 | Failed tests:
17-Jul-2018 10:31:26 | org.opennms.netmgt.alarmd.AlarmdBlackboxIT.canFlapAlarm(org.opennms.netmgt.alarmd.AlarmdBlackboxIT)
17-Jul-2018 10:31:26 | Run 1: AlarmdBlackboxIT.canFlapAlarm:161
17-Jul-2018 10:31:26 | Expected: <2>
17-Jul-2018 10:31:26 | but: was <1>
17-Jul-2018 10:31:26 | Run 2: AlarmdBlackboxIT.canFlapAlarm:161
17-Jul-2018 10:31:26 | Expected: <2>
17-Jul-2018 10:31:26 | but: was <1>
```

PR #2061, opened 2018-07-18
NMS-10266: Remove not used core/doc module

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-10266


mvrueden:
This simply removes an unused empty project. I go ahead and merge it.

PR #2062, opened 2018-07-18
NMS-8891: Fix Ldap error log entries GLPSRV044W triggered by LdapMonitor

indigo423:
* JIRA: http://issues.opennms.org/browse/NMS-8891


RangerRick:
Seems pretty straightforward to me.  Is there any reason to keep the changes commented out, rather than just deleting the redundant code?

indigo423:
@michael-batz probably knows the reason. Agree with you, commented code in the tool is a no go :)

indigo423:
@michael-batz @RangerRick thanks for feedback, cleaned up the PR.

PR #2063, opened 2018-07-18
NMS-8204: Reusing existing VMware session getting data

indigo423:
Allow reusing existing sessions to VMware by getting data instead of login/logout for every request.

* JIRA: http://issues.opennms.org/browse/NMS-8204

### DIFF
--- a/integrations/opennms-vmware/src/main/java/org/opennms/protocols/vmware/VmwareViJavaAccess.java
+++ b/integrations/opennms-vmware/src/main/java/org/opennms/protocols/vmware/VmwareViJavaAccess.java
@@ -63,10 +63,12 @@ import javax.net.ssl.X509TrustManager;
 
 import org.exolab.castor.xml.MarshalException;
 import org.exolab.castor.xml.ValidationException;
+
 import org.opennms.core.spring.BeanUtils;
 import org.opennms.netmgt.collectd.vmware.vijava.VmwarePerformanceValues;
 import org.opennms.netmgt.config.vmware.VmwareServer;
 import org.opennms.netmgt.dao.VmwareConfigDao;
+
 import org.sblim.wbem.cim.CIMException;
 import org.sblim.wbem.cim.CIMNameSpace;
 import org.sblim.wbem.cim.CIMObject;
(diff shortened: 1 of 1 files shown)


mvrueden:
@christianpape can you please take a look.

integrations/opennms-vmware/src/main/java/org/opennms/protocols/vmware/VmwareViJavaAccess.java
```diff
@@ -202,7 +210,26 @@ public VmwareViJavaAccess(String hostname) throws MarshalException, ValidationEx
     public void connect() throws MalformedURLException, RemoteException {
         relax();
 
-        m_serviceInstance = new ServiceInstance(new URL("https://" + m_hostname + "/sdk"), m_username, m_password);
+	//Fix for NMS-8204: use singleton of serverConnection, also make method synchronized
+	String connectionString = new String(m_hostname + m_username + m_password);
+	if(m_userSessions.get(connectionString) == null)
```
j-white: Are the sessions thread safe? What happens if two different threads try and the same session?

integrations/opennms-vmware/src/main/java/org/opennms/protocols/vmware/VmwareViJavaAccess.java
```diff
@@ -233,19 +260,22 @@ public boolean setTimeout(int timeout) {
      * Disconnects from the server.
      */
     public void disconnect() {
-        if (m_serviceInstance == null) {
+	//Fix for NMS-8204: Doing nothing here
```
j-white: So we never close the sessions, and just wait for them to timeout?

integrations/opennms-vmware/src/main/java/org/opennms/protocols/vmware/VmwareViJavaAccess.java
christianpape: Shouldn't this block/method be "synchronized"?

christianpape:
This approach is simple, but we have to assure that this handling is thread-safe and stable concerning configuration changes during runtime. I don't know if we really need a more complicated connection pooling/management. In my point of view these connections will be opened and used/reopened as long as OpenNMS (data collection, provisioning, polling) is running. Stale connections will happen on configuration changes provisiond-configuration.xml or vmware-config.xml. I will try to run this branch using my two vCenter servers (v6.5 and v5.5).

christianpape:
Yes, agree.

christianpape:
Please see #2075

j-white:
Closing in favor of https://github.com/OpenNMS/opennms/pull/2075.

PR #2066, opened 2018-07-19
HZN-1345: Provide health:check command

mvrueden:
Here we introduce a new `health:check` command, which can possibly run in all containers.

* JIRA: https://issues.opennms.org/browse/HZN-1345

Screenshots of the new `health:check` command:
 * https://ibin.co/490rlOr1jTLj.png
 * https://ibin.co/490ryVs0QX6N.png

The documentation could be  a bit more, but decided to wait until the install-guid changes from @indigo423 are merged to avoid merge conflicts.

The introduced TODOs will be addressed when the feature definitions are reworked. See issue [HZN-1339](https://issues.opennms.org/browse/HZN-1339)


j-white:
Looks good and is working well 👍 

Should we add the command to the OpenNMS container too?

mvrueden:
I decided to not add it at the moment by default, as the Nominatim Geocoder Bundle always fails, thus making the `health:check` command also always fail.

However if we want this to be enabled in opennms container, too, it is not much work, and we can address that in a separate issue very easily (-:

j-white:
I noticed that this bundle health check also fails on the Minion container when started via ./runInPlace.sh - we install Hawtio by default, but one of the related bundles fails.

Having a bundle in a failed state isn't necessary a problem though...

mvrueden:
I disagree.

In `runInPlace.sh` we install the hawtio-offline feature, which starts a bundle which requires dependencies, which are not fullfilled, thus failing to start that bundle. Therefore we should not start/install the bundle or have it resolve its dependencies accordingly.

If we don't enforce the bundle check, the health-check deployment gets more complicated, which is fine, but adds more complexity.

j-white:
🥇

PR #2074, opened 2018-07-24
NMS-10257: Only refresh the event listener when handling a reloadDaemonConfig event

j-white:
JIRA: https://issues.opennms.org/browse/NMS-10257


mvrueden:
LGTM

PR #2076, opened 2018-07-26
NMS-10212: Do not reuse RadiusAuthenticator in RadiousAuthenticationProvider

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-10212

This is first targeted for foundation-2018. 
We need to decide where to backport this fix as well.

### DIFF
--- a/opennms-webapp/src/main/webapp/WEB-INF/spring-security.d/radius.xml.disabled
+++ b/opennms-webapp/src/main/webapp/WEB-INF/spring-security.d/radius.xml.disabled
@@ -47,7 +47,7 @@
     <!--<beans:property name="port" value="1812"/>-->
     <!--<beans:property name="timeout" value="5"/>-->
     <!--<beans:property name="retries" value="3"/>-->
-    <!--<beans:property name="authTypeClass"><beans:bean class="net.jradius.client.auth.PAPAuthenticator"/></beans:property>-->
+    <!--<beans:property name="authTypeClass" value="net.jradius.client.auth.PAPAuthenticator"/>
     <!--<beans:property name="defaultRoles" value="ROLE_USER"/>-->
     <!--<beans:property name="rolesAttribute" value="Unknown-VSAttribute(%%% Vendor ID %%%:%%% Attribute Number %%%)"/>-->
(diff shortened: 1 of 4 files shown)


indigo423:
Tried to:

a) compile from source
b) Installed RPMS from Bamboo

When I try to start OpenNMS it got stuck here. I don't know how to test it.

screen shot 2018-07-27 at 18 43 50

indigo423:
Was able to test the branch and works as expected. Thanks to fearless @mvrueden and his debugging skills 🏅

@RangerRick we have to add a changelog, the fix requires to change an existing `radius.xml`. To not reuse an existing class, the configuration needs now to be in the radius.xml:

```
<beans:property name="authTypeClass" value="net.jradius.client.auth.PAPAuthenticator"/>
```

We have to give a hint to all users with Horizon 22.0.2+ and Meridian 2018+ when we merge to foundation-2018 right?

opennms-webapp/src/main/webapp/WEB-INF/spring-security.d/radius.xml.disabled
```diff
@@ -47,7 +47,7 @@
     <!--<beans:property name="port" value="1812"/>-->
     <!--<beans:property name="timeout" value="5"/>-->
     <!--<beans:property name="retries" value="3"/>-->
-    <!--<beans:property name="authTypeClass"><beans:bean class="net.jradius.client.auth.PAPAuthenticator"/></beans:property>-->
+    <!--<beans:property name="authTypeClass" value="net.jradius.client.auth.PAPAuthenticator"/>
```
BugBuster1701: This is missing in the end:  `-->`

mvrueden: Good catch, thanks (-:

mvrueden:
Marking as triage as it is not yet decided on how to address the backwards-compatibility issue.

mvrueden:
After a quick discussion, we are going to backport this to `foundation-2015` and create a `Breaking Changes` section in the changelog.

PR #2079, opened 2018-07-27
NMS-10282: Use ISO8601 date format conversion from StringUtils

cgorantla:
Use ISO8601 date format conversion like [here](https://github.com/OpenNMS/opennms/blob/c1cb2d3777059be9672218491b6374ee4140ded9/core/lib/src/main/java/org/opennms/core/utils/StringUtils.java#L249) instead of specifying it in `yyyy-MM-dd'T'HH:mm:ss.SSS` format.
Remove default date format in xml, it will still default to ISO8601 format.

* JIRA: http://issues.opennms.org/browse/NMS-10282

### DIFF
--- a/opennms-alarms/api/src/main/java/org/opennms/netmgt/alarmd/api/support/AbstractNorthbounder.java
+++ b/opennms-alarms/api/src/main/java/org/opennms/netmgt/alarmd/api/support/AbstractNorthbounder.java
@@ -32,6 +32,7 @@ import java.io.Serializable;
 import java.io.StringWriter;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
+import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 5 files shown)


cgorantla:
@agalue  Thanks for the review. that would help if you can verify it once.

mvrueden:
Marking as `triage` as we are waiting on confirmation from @agalue

agalue:
I tried the latest RPMs on a VM, and it works as expected.

First, I enabled rsyslog to listen on UDP 514 on the VM itself. Then, I've enabled Syslog NBI, and added a date field on the message.

Initially, I've used the default (no custom date format), and send a nodeDown event. The event was received as:

```
Aug  2 14:00:09 horizon.local OpenNMS: 2018-08-02T14:00:08-04:00 - ID:1, NODE:; #012            Node  is down.#012      
```

So, it is indeed using the default.

Now, I've changed the format to be "", and reloaded the Syslog NBI. I sent the event again and it was received as:

```
Aug  2 14:04:03 horizon.local OpenNMS: 08/02/2018 14:04:02 - ID:1, NODE:; #012            Node  is down.#012 
```

PR #2082, opened 2018-07-30
HZN-1348: Allow Drools rule for Alarmd to be customized

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1348

Here we externalize the Drools rules used by alarmd into `etc/alarmd/drools-rules.d/alarmd.drl`, and allow them to be reloaded dynamically by sending a daemon-reload event to alarmd.

This allow the rules to be customized.


mvrueden:
At first glance this looks good to me. As I am not very familiar with Drools, maybe some else with a bit more insights can take a look?

PR #2085, opened 2018-07-30
NMS-11794: Allow partially completed provisioning scans

brynjare:
* JIRA: https://issues.opennms.org/browse/NMS-11794

Basing this on the 'develop' branch, since the commit affects a critical part of OpenNMS, changing the current behaviour.

This commit adds two new events: 
1. uei.opennms.org/internal/provisiond/provisionTaskFailed
2. uei.opennms.org/internal/provisiond/nodeScanPartiallyCompleted

Whenever NodeScan encounters a non-critical problem during the scan, a Warning event (including information about which step failed) will be created, but the scan will continue.

If one or more of these failures were encountered, a nodeScanPartiallyCompleted Warning event will be created, triggering an alarm. This alarm will be auto-cleared by the next nodeScanCompleted event, triggered by a successful scan.

### DIFF
--- a/features/events/api/src/main/java/org/opennms/netmgt/events/api/EventConstants.java
+++ b/features/events/api/src/main/java/org/opennms/netmgt/events/api/EventConstants.java
@@ -437,8 +437,12 @@ public abstract class EventConstants {
 
     /** Constant <code>PROVISION_SCAN_COMPLETE_UEI="uei.opennms.org/internal/provisiond/nod"{trunked}</code> */
     public static final String PROVISION_SCAN_COMPLETE_UEI="uei.opennms.org/internal/provisiond/nodeScanCompleted";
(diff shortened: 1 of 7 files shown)


j-white:
Looks reasonable at first glance, I'll need to take a closer look to assess the impact though.

One thing that's missing are tests to capture the new behavior.

RangerRick:
Agreed.  Once there's some test coverage, we can run it through Bamboo and make sure everything passes.

brynjare:
Thanks for the feedback so far! I just returned from vacation, but I'll look into creating some tests for the new functionality.

mvrueden:
Last state of this is, that test coverage is needed. @brynjare are you still considering working on this?

mvrueden:
@j-white, @RangerRick do we take the time implementing the tests ourselves, or what do we want to do with this PR?

PR #2086, opened 2018-07-30
NMS-10291: DefaultProvisionService logs noisily for monitored service having state "N"

jeffgdotorg:
* JIRA: http://issues.opennms.org/browse/NMS-10291


j-white:
Simple enough. Will merge once green.

PR #2089, opened 2018-08-02
NMS-10304: Refactor timeseries applicationContext xml files with 'timeseries' prefix

cgorantla:
* JIRA: https://issues.opennms.org/browse/NMS-10304


mvrueden:
LGTM, thank you for the quick patch 👍

PR #2092, opened 2018-08-03
 NMS-9899: Add a new WsMan WQL-based detector

dino2gnt:
* Jira: https://issues.opennms.org/browse/NMS-9899

Add a new WsMan detector that can use WQL filters against an
endpoint to detect arbitrary services.

### DIFF
--- a/opennms-doc/guide-admin/src/asciidoc/text/provisioning/detectors/wsmanwql.adoc
+++ b/opennms-doc/guide-admin/src/asciidoc/text/provisioning/detectors/wsmanwql.adoc
@@ -1,0 +1,70 @@
+
+// Allow GitHub image rendering
+:imagesdir: ../../../images
+
+==== WS-Man WQL Detector
+
+The WS-Management WQL detector attempts to connect to the agent defined in `wsman-config.xml` and issues a WQL query.
+If the query successfully returns one or more items, the service is marked as detected. The WS-Man WQL detector can be used to define arbitrary services based on WQL filter results.
+
+===== Detector facts
+
+[options="autowidth"]
+|===
+| Implementation | `org.opennms.netmgt.provision.detector.wsman.WsManWQLDetector`
+|===
+
+===== Configuration and Usage
+
+.Parameters for the <DETECTOR-NAME-HERE>
+[options="header, autowidth"]
+|===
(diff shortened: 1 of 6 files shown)


j-white:
Looks great. The only issue I see is that the factory is not exposed via the Blueprint, so the detector won't be available on Minion.

Try adding an entry similar to existing one for the `WsManDetectorFactory` in: `opennms/features/wsman/src/main/resources/OSGI-INF/blueprint/blueprint.xml`

dino2gnt:
As requested

j-white:
Pushed to `pr/jira/NMS-9899` for a Bamboo run.

j-white:
Build passed.

@dino2gnt can you also add an entry for this detector to: https://github.com/OpenNMS/opennms/blob/develop/smoke-test/src/test/java/org/opennms/smoketest/minion/DetectorsCommandIT.java#L117

This will test to ensure that the detector loads on Minion successfully.

dino2gnt:
Sufficient?

indigo423:
@dino2gnt thank you very much 🤘

j-white:
I updated `DetectorsCommandIT` to expect the name to be `WsManWQLService` as defined in the detector instead of `WsManWQL`, so that the test would pass.

opennms-doc/guide-admin/src/asciidoc/text/provisioning/detectors/wsmanwql.adoc
```diff
@@ -0,0 +1,70 @@
...
+===== Detector facts
+
+[options="autowidth"]
+|===
+| Implementation | `org.opennms.netmgt.provision.detector.wsman.WsManWQLDetector`
+|===
+
+===== Configuration and Usage
+
+.Parameters for the <DETECTOR-NAME-HERE>
```
pioto: "<DETECTOR-NAME-HERE>"?

dino2gnt: Same as opennms-doc/guide-admin/src/asciidoc/text/provisioning/detectors/wsman.adoc, which was the template for the wql detector doc.  if it's wrong, it's wrong there too :)

PR #2094, opened 2018-08-05
HZN-1370: Provide simple health check mechanism for configured adapters

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1370


mvrueden:
Maybe we should add the same for the minion -> `ListenerAdapter` to diagnose if the Minion is starting the Flow* Listeners correctly as well.

j-white:
Good idea, wouldn't be a bad idea to run these checks on Minion - maybe even OpenNMS too.

I'll merge this as is, since it's already useful on the Sentinel.

PR #2098, opened 2018-08-06
Jira/hzn 1338

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1338


mvrueden:
Even if this states  `Test are Pending` bamboo finished successfully: https://bamboo.opennms.org/browse/OPENNMS-ONMS2352-5

PR #2102, opened 2018-08-07
HZN-1369: Display metrics from the MetricRegistry objects in the Karaf shell

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1369

Here we add new `health:metrics-list` and `health:metrics-display` shell commands to enumerate and display metrics from the MetricRegistry objects in the Karaf shell.

Short term this will allow us to view metrics related to the flow adapters in the Sentinel containers shell. Longer term we can work to ensure all of the metric sets are exposed in the service registry for ad-hoc viewing.


core/health/shell/src/main/java/org/opennms/core/health/shell/MetricsListCommand.java
```diff
@@ -0,0 +1,65 @@
...
+            return null;
+        }
+
+        for (NamedMetricSet namedMetricSet : metricSets) {
+            System.out.printf("Metric set: %s\n", namedMetricSet.getName());
+            if (namedMetricSet.hasDescription()) {
+                System.out.printf("\t%s\n", namedMetricSet.getDescription());
+            }
+        }
+        System.out.println();
```
mvrueden: Can we print the metrics in one line instead of splitting them? 
I would imagine something like this:

```
Name       Description
Flows      Description of metric
...        ...
```

PR #2112, opened 2018-08-11
Sentinel

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1284

Most of the changes here were reviewed when merged into the `features/sentinel` branch. I'm creating this PR to track any changes that need to be made prior to merge `features/sentinel` back into `develop`.


core/ipc/sink/kafka/server/src/main/java/org/opennms/core/ipc/sink/kafka/server/KafkaMessageConsumerManager.java
```diff
@@ -152,30 +162,12 @@ protected void stopConsumingForModule(SinkModule<?, Message> module) throws Exce
 
     @Override
     public void afterPropertiesSet() throws Exception {
-        // Set the defaults
         kafkaConfig.clear();
-        kafkaConfig.put("group.id", SystemInfoUtils.getInstanceId());
```
j-white: It looks like we now default to using 'OpenNMS' instead of the value from `SystemInfoUtils.getInstanceId()` as the default `group.id`.

cgorantla: Looks like this still has  `SystemInfoUtils.getInstanceId` as group.id 
  https://github.com/OpenNMS/opennms/pull/2112/files/d9723e6f7ccd8f8c3f2a6fd7326821c78c6c6253#diff-fd3056f4d5972e167c7e54d80f71c92fR42

j-white: Good point. Looks like it only affects the new code in Sentinel. Shouldn't be a problem them.

j-white:
I'm looked over this a few times and haven't found any other issues... going to move ahead and merge :)

core/ipc/sink/kafka/server/src/main/java/org/opennms/core/ipc/sink/kafka/server/KafkaMessageConsumerManager.java
mvrueden: It should be configurable for both systems. For the OpenNMS one it is defined in the system properties as before and for sentinel it can be defined in the `kafka.consumer` properties.

mvrueden:
# 🎈 🎉

PR #2114, opened 2018-08-12
NMS-10312:  Be able to use Proxy for any Monitor or Collector that uses HttpClientWrapper directly

patrick-schweizer:
* JIRA: https://issues.opennms.org/browse/NMS-10312


smoke-test/opennms-home/etc/poller-configuration.xml
```diff
@@ -61,6 +61,7 @@
       <parameter key="rrd-repository" value="/opt/opennms/share/rrd/response" />
       <parameter key="rrd-base-name" value="http" />
       <parameter key="ds-name" value="http" />
+      <parameter key="use-system-proxy" value="false"/>
```
j-white: Is there a need to set these given that the default is `false`?

patrick-schweizer: not really, I thought it might serve as an example - I will revert

agalue:
Is it possible to put something on the logs when this option is enabled ? I've tried and it doesn't seem to work. 

Here is what I did:

* I've started a VM with the RPMs from the branch jira/NMS-10312
* I've configured the VM to use a global proxy (192.168.205.1 is the IP of my machine):
```
[root@horizon-nms10312 ~]# env | grep http
http_proxy=http://192.168.205.1:3128
https_proxy=http://192.168.205.1:3128
```
* On my machine (192.168.205.1), I've started Squid through Docker using a sample basic configuration:
```
docker run --name squid --rm -it --publish 3128:3128 --volume /Users/agalue/Development/opennms/git/squid.conf:/etc/squid/squid.conf --volume /tmp/squid-cache:/var/spool/squid sameersbn/squid:3.5.27
```
* I've used iptables on the OpenNMS server to block outbound traffic for HTTP and HTTPS (to force using the proxy):
```
[root@horizon-nms10312 ~]# iptables -L
Chain INPUT (policy ACCEPT)
target     prot opt source               destination         

Chain FORWARD (policy ACCEPT)
target     prot opt source               destination         

Chain OUTPUT (policy ACCEPT)
target     prot opt source               destination         
DROP       tcp  --  anywhere             anywhere             tcp dpt:http
DROP       tcp  --  anywhere             anywhere             tcp dpt:https
```
* I've verified with curl that I can reach web sites only when the proxy is configured after applying the iptables rule.
* I've configured all the HTTP related services in `poller-configuration.xml` to have:
```
<parameter key="use-system-proxy" value="true"/>
```
* I've created a PSM monitor to also test the proxy.
* I've created a requisition and forced the services on the requisition as the detectors doesn't allow me to enable the proxy (this has to be fixed).
* After synchronizing the requisition, I can see outages for all the services which is not expected.

Let me know if you need more details.

agalue:
The proxy settings have been configured through `/etc/profile.d/proxy.conf`, and I've verified that they are available on all users (including root).

As I'm curious, I've added the following to `opennms.conf` and restarted OpenNMS:
```
ADDITIONAL_MANAGER_OPTIONS="$ADDITIONAL_MANAGER_OPTIONS -Dhttp.proxyHost=192.168.205.1 -Dhttp.proxyPort=3128"
```
That cleared the outages for the PSM based monitor, but not for the basic monitors for HTTP and HTTPS.

Tomorrow, I'll try the HTTPCollector.

patrick-schweizer:
@agalue, based on your comments I added a logging message: https://github.com/OpenNMS/opennms/pull/2114/files#diff-28b0daa680183ddc33de1e24a4e1310aR60

RangerRick:
@agalue do you have any other feedback for this (and https://github.com/OpenNMS/opennms/pull/2115) or are these mergeable?

agalue:
I haven't had time to fully test the solution. So, I don't know if it works or not.

mvrueden:
Waiting for feedback from @agalue

agalue:
Prior resuming the tests ...

This PR seems to affects the following:
* HttpCollector
* HttpsPostMonitor
* PageSequenceMonitor
* WebMonitor

Indeed there are documentation, but I think it worth emphasizing the fact that the environment variables for Proxy on Linux are not enough, and you have to use the Java System properties on opennms.conf in order to use the System Proxy (which is not obvious).

I can see that the XML Collector is covered by the PR 2115.

But, I cannot see where are the detectors covered:

* WebDetector, or to be more precise, the detectors that use `WebClient` from `opennms-provision/opennms-detector-web`

There are other classes that appear to be using HTTP like:

* UsageStatisticsReporter
* NominatimGeocoderService
* NCSNorthbounder
* RequestTracker
* HttpRequisitionProvider
* AbstractSlackCompatibleNotificationStrategy
* HttpNotificationStrategy

It is OK if these other features are not covered, but we should warn on the docs that they won't work when a proxy is present.

I'll try to find a time slot to test what's covered here, and looking forward to hear opinions about what is not covered.

agalue:
So far, HttpCollector, WebMonitor and PageSequenceMonitor are working.

agalue:
@patrick-schweizer, thanks for adding support for the `WebDetector`, the code looks good. Unfortunately, out Bamboo instance didn't trigger the execution of the test and the RPM generation for your branch. I'll ask internally, so I can test it on my lab.

patrick-schweizer:
@agalue although I added the support for the WebDetector I am not sure if we shouldn't rename the parameter so that it is consistent?
* in WebDetector: "useSystemProxySettings"
* for the rest: "use-system-proxy"

In WebDetector it can't be "use-system-proxy" since it is not a proper Java method name and it seems that the parameters on WebDetector are set via reflection on setters.

So my suggestion is to use everywhere  "useSystemProxySettings" instead of "use-system-proxy". Are you ok with that change?

On the other classes that use HTTP I will create a new sub ticket so we can merge this PR independently.

agalue:
I believe all the poller and collector parameters use a "dash" as a separator instead of camel case names. Certainly detectors use reflections to get the attribute names from the getters, and force to have different naming conventions.

I would keep `use-system-proxy` for everywhere, and for the detector, use the camel case format (`useSystemProxy` or `useSystemProxySettings`), to be consistent with the existing parameters.

And, totally agree for having separate JIRA Issues and PR for the other HTTP stuff.

pbrane:
> On Sep 28, 2018, at 14:56, Alejandro Galue <notifications@github.com> wrote:
> 
> I would keep use-system-proxy for everywhere, and for the detector, use the camel case format (useSystemProxy or useSystemProxySettings), to be consistent with the existing parameters.

It's very unfortunate that the camel case syntax leaked into the XML configurations.  That violates it's design semantics.  The W3C states that the elements and attributes should be lower case and all one word or use "-" to separate words.

You may not recall, but tools such as Castor (bad o'le days) would automatically convert the dashed entities into proper Java POJOs.


David

agalue:
David,

I understand, but here is how the detector configuration looks like (to show where the camel case will be used):

```xml
<detector name="WebServer-HTTPS" class="org.opennms.netmgt.provision.detector.web.WebDetector">
         <parameter key="virtualHost" value="www.google.com"/>
         <parameter key="port" value="443"/>
         <parameter key="schema" value="https"/>
         <parameter key="useSystemProxySettings" value="true"/>
</detector>
```

As you can see, we're talking about the name of the parameter. In terms of the detector implementation that requires a getter/setter (i.e. getUseSystemProxySettings() / setUseSystemProxySettings(String))

Makes sense ?

patrick-schweizer:
@agalue I renamed the methods from "setUseSystemProxySettings" to "setUseSystemProxy" to be consistent with "use-system-proxy"

patrick-schweizer:
@agalue did you have a chance to take another look? Are we good to merge or is there something still open? Thanks!

agalue:
@patrick-schweizer sorry, I've been very busy lately, but I believe this PR is ready to be merged. Thanks again for your hard work.

patrick-schweizer:
@j-white are you happy with the documentation (and the PR in general) or are there changes you want me to make? Otherwise I believe we are good to merge.

PR #2115, opened 2018-08-12
NMS-10313 : Be able to use Proxy for any Monitor or Collector that uses HttpClient via UrlFactory

patrick-schweizer:
* JIRA: https://issues.opennms.org/browse/NMS-10313

### DIFF
--- a/opennms-doc/guide-admin/src/asciidoc/text/performance-data-collection/collectors/xml.adoc
+++ b/opennms-doc/guide-admin/src/asciidoc/text/performance-data-collection/collectors/xml.adoc
@@ -31,7 +31,6 @@ The following handlers are not currently supported on _Minion_:
 | Parameter              | Description                              | Required | Default value
 | `collection`           | The name of the _XML Collection_ to use  | required | `-`
 | `handler-class`        | Class used to perform the collection     | optional | `org.opennms.protocols.xml.collector.DefaultXmlCollectionHandler`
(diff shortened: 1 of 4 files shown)


j-white:
Also looks good. Again, we should update the relevant documentation.

patrick-schweizer:
@j-white I added documentation

agalue:
This appear to cover only the XML Collector. Is this correct ? Is that the only feature that has to be tested ?

j-white:
Looks fine to me. Having it as part of the request object seems more natural since that's where the other HTTP settings are set.

agalue:
Of course, maybe tuning the docs to be more precise about where to put the flag would be useful.

patrick-schweizer:
@agalue I updated the documentation. Could you take a look? Thanks!

agalue:
Looks good, but I would additionally add a small note to say something like: 

The request parameters are optional. On the above example, `use-system-proxy` is one of the available parameters that can be used. This one is to let the XML Collector know that it should send the requests through the system Proxy.

Maybe not with those words, but you get the idea.

This is because the other valid parameters are not documented ;)

patrick-schweizer:
@agalue I added documentation for all parameters. That should make it clearer. Could you take another look? Thanks!

patrick-schweizer:
@agalue , @j-white are we good to merge? Or is there still something open? Cheers Patrick

PR #2119, opened 2018-08-14
NMS-10273: Add support for Cisco Syslog Message formats to the RadixTreeSyslogParser

mattixtech:
* JIRA: http://issues.opennms.org/browse/NMS-10273

### DIFF
--- a/features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/GrokParserStageSequenceBuilder.java
+++ b/features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/GrokParserStageSequenceBuilder.java
@@ -28,7 +28,11 @@
 
 package org.opennms.netmgt.syslogd;
 
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.time.ZoneId;
 import java.util.List;
+import java.util.Locale;
 import java.util.function.BiConsumer;
 
 import org.opennms.core.time.ZonedDateTimeBuilder;
@@ -42,7 +46,10 @@ import org.slf4j.LoggerFactory;
  * assign parsed values to a syslog message object. It supports the following
  * pattern types:</p>
  * <ul>
+ * <li>HOSTNAME: any valid hostname character.</li>   
+ * <li>HOSTNAMEORIP: any valid hostname or IP address character.</li>  
  * <li>INT: Positive integer.</li>
+ * <li>IPADDRESS: any valid IP address character.</li> 
  * <li>MONTH: 3-character English month abbreviation.</li>
  * <li>NOSPACE: String that contains no whitespace.</li>
  * <li>STRING: String. Because this matches any character, it must be followed by a delimiter in the pattern string.</li>
@@ -63,8 +70,11 @@ public abstract class GrokParserStageSequenceBuilder {
 		END_PATTERN
 	}
 
-	private static enum GrokPattern {
+	enum GrokPattern {
+		HOSTNAME,
+		HOSTNAMEORIP,
 		INT,
+		IPADDRESS,
 		MONTH,
 		NOSPACE,
 		STRING
@@ -202,6 +212,11 @@ public abstract class GrokParserStageSequenceBuilder {
 	}
 
 	/**
+	 * The prefix for a generic parameter.
+	 */
+	private static final String PARAMETER_PREFIX = "parm";
+
+	/**
 	 * This function maps {@link SyslogSemanticType} values of type int to fields in the parser
 	 * state.
 	 * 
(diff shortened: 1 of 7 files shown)


features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/ParserStageSequenceBuilder.java
```diff
@@ -715,6 +822,46 @@ public boolean equals(Object o) {
...
+		}
+
+		@Override
+		public AcceptResult acceptChar(ParserStageState state, char c) {
+			for (char end : getEnd()) {
+				if (end == c || ! m_charMatcher.test(c)) {
+					return AcceptResult.COMPLETE_WITHOUT_CONSUMING;
+				}
+			}
+			// TODO: Make this more efficient?
```
j-white: Is there a way to make this more efficient? `c` is a single character, so comparing it against a set of known white-space characters (which would be truncated by `trim()`) may be faster.

features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/ParserStageSequenceBuilder.java
```diff
@@ -152,6 +172,38 @@ private void addStage(ParserStage stage) {
...
+
+		switch (pattern) {
+			case HOSTNAME:
+				return hostNameMatcher;
+			case HOSTNAMEORIP:
+				return hostnameOrIPMatcher;
+			case IPADDRESS:
+				return ipV4OrV6AddressMatcher;
+			default:
+				throw new IllegalArgumentException("Unsupported pattern.");
```
j-white: I would include the pattern in the exception message.

features/events/syslog/src/main/resources/org/opennms/netmgt/syslogd/grok-patterns.txt
```diff
@@ -78,3 +81,15 @@
...
+<%{INT:facilityPriority}>%{INT:parmSequenceNum}: %{NOSPACE:parmComponentId}:%{MONTH:month} %{INT:day} %{INT:hour}:%{INT:minute}:%{INT:second}.%{INT:secondFraction} : %{NOSPACE:processName}[%{INT:processId}]: %{STRING:message}
+<%{INT:facilityPriority}>: %{INT:year} %{MONTH:month} %{INT:day} %{INT:hour}:%{INT:minute}:%{INT:second} %{STRING:timezone}: %{STRING:message}
+<%{INT:facilityPriority}>%{INT:parmSequenceNum}: %{STRING:month} %{INT:day} %{INT:hour}:%{INT:minute}:%{INT:second}.%{INT:secondFraction}: %{STRING:message}
+<%{INT:facilityPriority}> %{INT:year} %{MONTH:month} %{INT:day} %{INT:hour}:%{INT:minute}:%{INT:second} %{STRING:timezone} *%{HOSTNAME:hostname}* %{STRING:message}
+<%{INT:facilityPriority}> %{INT:year} %{MONTH:month} %{INT:day} %{INT:hour}:%{INT:minute}:%{INT:second} %{STRING:timezone} %{STRING:message}
+
+# Note: This pattern '%{STRING:message}' should always come last as it will greedily match anything. If a pattern is
+# placed after this one it will never be matched since this pattern matches anything.
+# TODO: Update this to take any string beginning with '%' i.e. %{BEGINSWITH_%:message}
```
j-white: It may not be a bad thing to leave this here without matching `%` first. This way the message won't be dropped, and given the new behavior we've introduced it will be associated with the current time, and the source host.

mattixtech:
Addressed your feedback. Also included a fix for a corner case I didn't originally notice (where the hostname is the last pattern in a grok string, that is handled in its own section).

features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/GrokParserStageSequenceBuilder.java
```diff
@@ -198,7 +208,17 @@
 		/**
 		 * Remaining string message.
 		 */
-		message
+		message,
+
+		/**
+		 * Sequence number.
+		 */
+		parmSequenceNum,
```
j-white: Is there an easy way to make the parameter naming generic so that we don't have to hard-code the parameter names?

mattixtech: Could use the text form the grok pattern as the key for the map. That's easy for the map.put(). For the get() you would need to know the literal string for the parameter you want to get since you could no longer reference it here. Probably safe to assume those pattern names will never change and if so then we can drop the grok pattern -> key mapping and get rid of this enum. There's also no getters at the moment to worry about. So yes, I'll change this.

j-white:
Thanks for the update 👍

PR #2120, opened 2018-08-15
NMS-10306: Add Karaf shell command for Daemon Reload

cgorantla:
* JIRA: http://issues.opennms.org/browse/NMS-10306


features/events/api/src/main/java/org/opennms/netmgt/events/api/DaemonReloadEnum.java
```diff
@@ -0,0 +1,49 @@
...
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.netmgt.events.api;
+
+public enum DaemonReloadEnum {
+
+    EVENTD("Eventd"),
```
j-white: Is there a better place to put these constants outside out the event api? Perhaps in `core/daemon`, `core/lib` or `core/config`.

PR #2128, opened 2018-08-22
NMS-9267: fix some issues in users.xml handling in the migrator

RangerRick:
1. It was possible for upgrade to fail if `read-only="false"` tags existed without any `true` read-only user.
2. There are cases when the `read-only` flag could fail to be parsed out on upgrade.

This version implements simple DOM parsing instead of direct string manipulation to avoid any unmarshalling issues.

* JIRA: http://issues.opennms.org/browse/NMS-9267


j-white:
Should this be targeted to {{foundation-2017}}?

agalue:
I'll give it a try.

Also, I agree on having this PR targeting `foundation-2017`

j-white:
Looks good to wait. I'll wait for confirmation from @agalue to merge though.

PR #2132, opened 2018-08-28
HZN-1380: HA redundancy via ZooKeeper leader election

mattixtech:
* JIRA: http://issues.opennms.org/browse/HZN-1380


features/distributed/coordination/zookeeper/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -0,0 +1,20 @@
+<blueprint xmlns="http://www.osgi.org/xmlns/blueprint/v1.0.0"
+           xmlns:cm="http://aries.apache.org/blueprint/xmlns/blueprint-cm/v1.1.0">
+    <!-- Configuration properties -->
+    <cm:property-placeholder id="zookeeperProperties"
+                             persistent-id="org.opennms.features.distributed.coordination.zookeeper"
+                             update-strategy="reload">
+        <cm:default-properties>
+            <cm:property name="connectString" value="127.0.0.1:2181"/>
+            <cm:property name="namespace" value="coordination"/>
+            <!--<cm:property name="auths" value=""/>-->
```
j-white: If we're not including any authentication options, I'd omit it completely rather than leaving it commented.

features/distributed/coordination/zookeeper/src/test/java/org/opennms/features/distributed/coordination/zookeeper/TestZookeeperDomainManager.java
```diff
@@ -0,0 +1,145 @@
...
+
+        manager = ZookeeperDomainManager.withMocks("test.domain", mockBuilder, mockLeaderSelector);
+        LeaderSelectorListener leaderSelectorListener = manager.getLeaderSelectorListener();
+
+        int numToRegister = 100;
+        IntStream.range(0, numToRegister).parallel().forEach(this::register);
+
+        // We want to wait until the connection from above actually occurred
+        await().atMost(10, TimeUnit.SECONDS).until(manager::isConnected);
+        // TODO: This has potential to false positive since we could theoretically have another thread connecting
```
j-white: Is this a TODO, or just a note?

features/distributed/coordination/zookeeper/src/main/java/org/opennms/features/distributed/coordination/zookeeper/AlwaysZookeeperDomainManagerFactory.java
```diff
@@ -0,0 +1,104 @@
...
+import org.apache.curator.framework.CuratorFrameworkFactory;
+import org.apache.curator.retry.ExponentialBackoffRetry;
+import org.opennms.features.distributed.coordination.api.DomainManager;
+import org.opennms.features.distributed.coordination.api.DomainManagerFactory;
+import org.opennms.features.distributed.coordination.base.AbstractDomainManagerFactory;
+
+/**
+ * A {@link DomainManagerFactory} that uses Apache ZooKeeper to manage all domains.
+ */
+public class AlwaysZookeeperDomainManagerFactory extends AbstractDomainManagerFactory {
```
j-white: Can we drop the `Always` prefix in the name here? It's normally implicit for these types of factories.

features/distributed/coordination/zookeeper/src/test/java/org/opennms/features/distributed/coordination/zookeeper/TestZookeeperDomainManager.java
mattixtech: More of a note. I wasn't sure how to resolve this in a clean way. Basically I need to sleep for some time before I do the assert to make sure there really wasn't any other thread attempting a connection.

The easiest thing to do would be just add a Thread.sleep(1000) before the assert and that should take care of it but I wasn't sure if that was appropriate since bare sleeps are typically discouraged in unit tests.

features/distributed/coordination/zookeeper/src/main/java/org/opennms/features/distributed/coordination/zookeeper/ZookeeperDomainManager.java
```diff
@@ -0,0 +1,224 @@
...
+        try {
+            LOG.info("Connecting to ZooKeeper with client: ", client);
+            client.blockUntilConnected();
+            LOG.info("Connected to ZooKeeper");
+        } catch (InterruptedException e) {
+            LOG.warn("Interrupted while connecting to ZooKeeper", e);
+        }
+
+        if (leaderSelector == null) {
+            // TODO: Not sure if the leading slash is necessary
```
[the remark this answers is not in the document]
mattixtech: Forgot to verify that, yes it did end up being necessary.

features/distributed/coordination/zookeeper/src/test/java/org/opennms/features/distributed/coordination/zookeeper/TestZookeeperDomainManager.java
mattixtech: I'm just going to add the 1 second sleeps in and remove those comments.

j-white:
👍 thanks for the updates

features/distributed/coordination/api/src/main/java/org/opennms/features/distributed/coordination/api/Role.java
```diff
@@ -0,0 +1,38 @@
...
+
+package org.opennms.features.distributed.coordination.api;
+
+/**
+ * The valid roles when participating in a redundant cluster.
+ */
+public enum Role {
+    ACTIVE,
+    STANDBY,
+    UNKNOWN
```
mvrueden: Can we define UNKNOWN first? Otherwise Active may be the default which may not be the intention.

features/distributed/coordination/api/src/main/java/org/opennms/features/distributed/coordination/api/RoleChangeHandler.java
```diff
@@ -0,0 +1,52 @@
...
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.features.distributed.coordination.api;
+
+/**
+ * Defines the callbacks invoked by a {@link DomainManager} when a registrant becomes ACTIVE or STANDBY.
+ */
+public interface RoleChangeHandler {
```
mvrueden: Just a thought: Is it possible to change this to a `@FunctionalInterface` by having something like `roleChanged(Role newRole, String domain)`?

features/distributed/coordination/base/src/main/java/org/opennms/features/distributed/coordination/base/AbstractDomainManager.java
```diff
@@ -0,0 +1,222 @@
...
+    protected synchronized final void becomeActive() {
+        currentRole = Role.ACTIVE;
+        LOG.debug("Notifying all registrants of ", currentRole, " status");
+
+        // Handlers must not block so that this happens in a timely manner
+        for (Map.Entry<String, RoleChangeHandler> handler : roleChangeHandlers.entrySet()) {
+            try {
+                handler.getValue().becomeActive(getDomain());
+            } catch (Exception e) {
+                LOG.warn("Got exception while notifying handler '", handler.getKey(), "' of role '",
```
mvrueden: Just a note: Usually we use the `{}` syntax in log messages.

features/distributed/coordination/base/src/main/java/org/opennms/features/distributed/coordination/base/AbstractDomainManager.java
```diff
@@ -0,0 +1,222 @@
...
+            public void becomeStandby(String domain) {
+                onStandby.accept(domain);
+            }
+        });
+    }
+
+    @Override
+    public final synchronized void deregister(String id) {
+        if (!this.roleChangeHandlers.containsKey(Objects.requireNonNull(id))) {
+            throw new IllegalArgumentException(id + " is not registered");
```
mvrueden: Just a thought: Maybe `NoSuchElementException` is better suited here?

features/distributed/coordination/base/src/test/java/org/opennms/features/distributed/coordination/base/TestAbstractDomainManager.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.stream.IntStream;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.opennms.features.distributed.coordination.api.Role;
+
+/**
+ * Tests for {@link AbstractDomainManager}.
+ */
+public class TestAbstractDomainManager {
```
mvrueden: We suffix tests with `Test`. If it is prefixed it is very likely not picked up while building. This should be named `AbstractDomainManagerTest` or `DomainManagerTest`.

features/distributed/coordination/base/src/test/java/org/opennms/features/distributed/coordination/base/TestAbstractDomainManagerFactory.java
```diff
@@ -0,0 +1,83 @@
...
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.opennms.features.distributed.coordination.api.DomainManager;
+import org.opennms.features.distributed.coordination.api.DomainManagerFactory;
+
+/**
+ * Tests for {@link AbstractDomainManagerFactory}.
+ */
+public class TestAbstractDomainManagerFactory {
```
mvrueden: See above.

features/distributed/coordination/base/src/test/java/org/opennms/features/distributed/coordination/base/TestConnectionBasedDomainManager.java
```diff
@@ -0,0 +1,148 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * Tests for {@link ConnectionBasedDomainManager}.
+ */
+public class TestConnectionBasedDomainManager {
```
mvrueden: See above.

features/distributed/coordination/zookeeper/src/main/java/org/opennms/features/distributed/coordination/zookeeper/ZookeeperDomainManager.java
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        client.start();
+
+        try {
+            LOG.info("Connecting to ZooKeeper with client: ", client);
+            client.blockUntilConnected();
+            LOG.info("Connected to ZooKeeper");
+        } catch (InterruptedException e) {
+            LOG.warn("Interrupted while connecting to ZooKeeper", e);
```
mvrueden: Can we safely continue here or do we have to `return`?

features/distributed/coordination/zookeeper/src/main/java/org/opennms/features/distributed/coordination/zookeeper/ZookeeperDomainManager.java
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        if (leaderSelector == null) {
+            leaderSelector = new LeaderSelector(client, "/" + getDomain(),
```
mvrueden: Devils Advocat here: What if `getDomain()` starts with `/`?

features/distributed/coordination/zookeeper/src/test/java/org/opennms/features/distributed/coordination/zookeeper/TestZookeeperDomainManager.java
```diff
@@ -0,0 +1,148 @@
...
+import org.apache.curator.framework.recipes.leader.CancelLeadershipException;
+import org.apache.curator.framework.recipes.leader.LeaderSelector;
+import org.apache.curator.framework.recipes.leader.LeaderSelectorListener;
+import org.apache.curator.framework.state.ConnectionState;
+import org.junit.Test;
+
+/**
+ * Tests for {@link ZookeeperDomainManager}.
+ */
+public class TestZookeeperDomainManager {
```
mvrueden: See above related to test naming.

features/distributed/coordination/zookeeper/src/test/java/org/opennms/features/distributed/coordination/zookeeper/TestZookeeperDomainManager.java
```diff
@@ -0,0 +1,148 @@
...
+
+        if (leadershipFuture.isDone() || leadershipFuture.isCompletedExceptionally()) {
+            fail("Leadership thread did not block");
+        }
+
+        try {
+            leaderSelectorListener.stateChanged(null, ConnectionState.LOST);
+            fail("No cancel leadership exception was caught");
+        } catch (Exception e) {
+            assertTrue(e instanceof CancelLeadershipException);
```
mvrueden: Can we please not use `assertTrue`?
I find it very hard to investigate failing `assertTrue/False` tests as they not clearly state what the problem is. 
I'd opt for either using `assertThat(...)`, `assertEquals(...)` or if you absolutely must use `assertTrue/False`, add the expected condition: `assertTrue("Exception is not a CancelLeadershipException but was expected", e instanceof CancelLeadershipException)`

features/distributed/coordination/zookeeper/src/test/java/org/opennms/features/distributed/coordination/zookeeper/TestZookeeperDomainManager.java
```diff
@@ -0,0 +1,148 @@
...
+        CompletableFuture.allOf(standbyFutures.toArray(new CompletableFuture[numToRegister])).get(10,
+                TimeUnit.SECONDS);
+
+        manager.getRoleChangeHandlers().keySet().parallelStream().forEach(manager::deregister);
+        await().atMost(10, TimeUnit.SECONDS).until(() -> !manager.isConnected());
+        // Wait some extra time to avoid a race condition where a thread was disconnecting right after we became 
+        // disconnected
+        Thread.sleep(1000);
+        assertEquals(1, disconnectionAttempts.get());
+        assertFalse(manager.isAnythingRegistered());
```
mvrueden: See above.

container/features/src/main/resources/features-sentinel.xml
```diff
@@ -109,4 +109,39 @@
...
+        <bundle>mvn:org.opennms.features.distributed/coordination-zookeeper-impl/${project.version}</bundle>
+    </feature>
+
+    <feature name="sentinel-coordination-shell"
+             description="OpenNMS :: Features :: Distributed :: Coordination :: Shell" version="${project.version}">
+        <feature version="${project.version}">sentinel-coordination-api</feature>
+        <bundle>mvn:org.opennms.features.distributed/coordination-shell/${project.version}</bundle>
+    </feature>
+
+    <feature name="zookeeper-dependencies" version="${project.version}" description="ZooKeeper Dependencies">
```
mvrueden: Can we put the `zookeeper.version` here instead?

features/distributed/coordination/zookeeper/src/main/java/org/opennms/features/distributed/coordination/zookeeper/ZookeeperDomainManager.java
mattixtech: Bad things happen! I'll add validation on the domain to restrict the character set.

features/distributed/coordination/base/src/main/java/org/opennms/features/distributed/coordination/base/AbstractDomainManager.java
mattixtech: Yes, that makes more sense.

features/distributed/coordination/api/src/main/java/org/opennms/features/distributed/coordination/api/Role.java
mattixtech: Sure I can change it, I just had them in alphabetical order. What do you mean by the default? As far as I know this isn't being used in such a way that it could have a default value.

features/distributed/coordination/api/src/main/java/org/opennms/features/distributed/coordination/api/RoleChangeHandler.java
mattixtech: Yes, good suggestion. This also helps in case we were to add additional roles since we will have already forced the existing role handling logic to be in one place for every client.

features/distributed/coordination/zookeeper/src/main/java/org/opennms/features/distributed/coordination/zookeeper/ZookeeperDomainManager.java
mattixtech: I don't think we can safely continue. I've modified this so that we will continue trying to blockUntilConnected() until it succeeds. This way if we get interrupted we will just go right back to trying.

mattixtech:
Thanks all for the feedback. Latest commit addresses all feedback up to now.

container/features/src/main/resources/features-sentinel.xml
```diff
@@ -109,4 +109,39 @@
...
+        <bundle>mvn:org.opennms.features.distributed/coordination-zookeeper/${project.version}</bundle>
+    </feature>
+
+    <feature name="sentinel-coordination-shell"
+             description="OpenNMS :: Features :: Distributed :: Coordination :: Shell" version="${project.version}">
+        <feature version="${project.version}">sentinel-coordination-api</feature>
+        <bundle>mvn:org.opennms.features.distributed/coordination-shell/${project.version}</bundle>
+    </feature>
+
+    <feature name="zookeeper-dependencies" version="${zookeeper.version}" description="ZooKeeper Dependencies">
```
j-white: We've updated the feature definition and the reference to use `zookeeper.version`, but where is zookeeper.version defined? This module doesn't see the one defined in `features/distributed/coordination/zookeeper/pom.xml`.

Given that we don't necessarily want to use this version of zookeeper globally (since it may conflict with what version the Kafka client libraries depend on), we may be better off just hard-coding the version to `3.4.7` here instead.

PR #2134, opened 2018-08-29
HZN-1377: CDP topology REST Service

rssntn67:
HZN-1377: CDP topology REST Service

* JIRA: http://issues.opennms.org/browse/HZN-1377

Create a REST service that return a graphml in provided format supported by topology app.
Only added two properties for supporting edge status on edge sourceifindex and targetifindex


opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/TopologyRestService.java
```diff
@@ -0,0 +1,142 @@
...
+import org.opennms.features.graphml.model.InvalidGraphException;
+import org.opennms.netmgt.dao.api.TopologyDao;
+import org.opennms.netmgt.model.OnmsTopology;
+import org.opennms.netmgt.model.topology.Topology;
+import org.opennms.netmgt.model.topology.Topology.ProtocolSupported;
+import org.springframework.stereotype.Component;
+
+@Component
+@Path("topology")
+public class TopologyRestService {
```
mvrueden: Can we write this a bit more generic?
At the moment this only allows us to export topologies defined by linkd, but I'd prefer a version which would also simply allow us to export any other existing topologies, e.g. bsm, graphml, etc. 

For this, I'd create a `TopologyProvider` (maybe a better name) which for now returns a `GraphMLType` allowing us to export more topologies.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/TopologyRestService.java
```diff
@@ -0,0 +1,142 @@
...
+    private final static String SOURCE_IFINDEX= "sourceifindex";
+    private final static String TARGET_IFINDEX= "targetifindex";
+    
+    private TopologyDao m_topologyDao;
+
+
+    @GET
+    @Path("{supported-protocol}")
+    public Response getGraph(@PathParam("supported-protocol") String supportedprotocol) throws InvalidGraphException {
+        if (supportedprotocol.equals(Topology.ProtocolSupported.CDP.name())) {
```
mvrueden: This is not very generic. 
Depending on how generic the solution is, we either need to iterate over the above mentioned `TopologyProvider` and determine if the provider can provide the protocol. 
Otherwise I'd prefer something like

```java
ProtocolSupported protocol = ProtocolSupported.valueOf(supportedProtocol);
GraphMLType graph = getGraph(protocol);
return Response.ok(graph).build();
```

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/TopologyRestService.java
```diff
@@ -0,0 +1,142 @@
...
+//    private final static String TARGET_PORT= "targetport";
+    private final static String SOURCE_IFINDEX= "sourceifindex";
+    private final static String TARGET_IFINDEX= "targetifindex";
+    
+    private TopologyDao m_topologyDao;
+
+
+    @GET
+    @Path("{supported-protocol}")
+    public Response getGraph(@PathParam("supported-protocol") String supportedprotocol) throws InvalidGraphException {
```
mvrueden: We should add `@Produces` statements to indicate if we support json, xml or both.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/TopologyRestService.java
```diff
@@ -0,0 +1,142 @@
...
+        if (supportedprotocol.equals(Topology.ProtocolSupported.BRIDGE.name())) {
+            return Response.ok(getGraph(ProtocolSupported.BRIDGE)).build();
+        }
+        if (supportedprotocol.equals(Topology.ProtocolSupported.OSPF.name())) {
+            return Response.ok(getGraph(ProtocolSupported.OSPF)).build();
+        }
+        if (supportedprotocol.equals(Topology.ProtocolSupported.ISIS.name())) {
+            return Response.ok(getGraph(ProtocolSupported.ISIS)).build();
+        }
+        return Response
```
mvrueden: If not supported, we should return `Response.Status.Not_FOUND`. The `SERVICE_UNAVAILABLE` does not work in this case.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/TopologyRestService.java
mvrueden: It might also be nice to know the supported protocols in general, to fine grane select those the consumer is interested in.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/TopologyRestService.java
```diff
@@ -0,0 +1,142 @@
...
+            GraphMLEdge gedge = new GraphMLEdge();
+            gedge.setId(edge.getId());
+            gedge.setSource(graph.getNodeById(edge.getSource().getId()));
+            gedge.setTarget(graph.getNodeById(edge.getTarget().getId()));
+            gedge.setProperty(TOOLTIP_TEXT, edge.getTooltipText());
+            gedge.setProperty(SOURCE_IFINDEX, edge.getSourceIfIndex());
+            gedge.setProperty(TARGET_IFINDEX, edge.getTargetIfIndex());
+            graph.addEdge(gedge);
+        });
+        graphml.addGraph(graph);
```
mvrueden: The conversion here is very specific to enlinkd and should probably go somewhere else to also allow converting other topologies besides those discovered by enlinkd.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/TopologyRestService.java
rssntn67: Yup

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/TopologyRestService.java
rssntn67: Markus, this is nice...we can have a property to identify edge to let set the edge status....whatever topology we have. We can call in different name...but should be there. Also we need to support ports....as graphml specifies

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/TopologyRestService.java
rssntn67: Markus I agree but consider that the issue is create a CDP topology REST Service.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/TopologyRestService.java
rssntn67: Ok

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/TopologyRestService.java
rssntn67: OK

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/TopologyRestService.java
rssntn67: This is general as we can...because supportedProtocol is an enum....and can be extended...as our needs

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/TopologyRestService.java
rssntn67: I moved the path to topologies....eheh
Also consider that for some reason I do not understant the path rest/topology/CDP will be intercepted by topology app and the topology map is displayed....ehehe

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/TopologyRestService.java
mvrueden: I meant something like:

```Java
// Pseudo-Code not tested
@GET
@Path("/supported-protocols")
public Response getSupportedProtocols() {
    return Response.ok(Lists.newArrayList(SupportedProtocols.values())).build();
}
```

j-white:
GraphML works well for rendering graphs in XML, but what should we do for JSON? Just mapping the objects from the XML representation to JSON doesn't normally give nice objects which are easy to consume.

Perhaps we should use GraphSON when working with JSON? https://github.com/tinkerpop/blueprints/wiki/GraphSON-Reader-and-Writer-Library

j-white:
Can we add a test and example of the output XML? i.e. given a particular network and CDP topology, what do we expect the resulting output to look like?

rssntn67:
So this is just a testing start.
I'll do the necessary change in next step....follow NMS-10375....

So when you are OK with this POc, I'll close the PR without merging....
This is a good place to see the changes.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/TopologyRestService.java
rssntn67: The Service is quite abstract.
You can also registrer a graphML updaters...and do updates via that service

The name is a convention.

So for example your graphml REST service will provide CRUD fir graphml topologies.

You can get a new graphml say myspecialtopology.
Then you subscribe the TopologyService for the name GrapPHML/myspecialTopology

Then the TopologyProvider will ask the TopologyService to get the GrapPHML/myspecialtopology and get the right graph

rssntn67:
I'm Going to close this.
The discussion will as been reported as new issues in the following epic link

https://issues.opennms.org/browse/HZN-990

rssntn67:
The branch can be deleted too!

rssntn67:
We do not need anymore the branch

PR #2135, opened 2018-08-30
HZN-1383: Kafka Producer: Don't forward alarms for every reduce

mattixtech:
HZN-1383: Added new configuration to the OpennmsKafkaProducer to suppress sending alarms that differ from an existing alarm only by the count or last event time. Also added equivalent functionality in the KafkaAlarmDataSync.

* JIRA: http://issues.opennms.org/browse/HZN-1383

### DIFF
--- a/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
+++ b/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
@@ -35,10 +35,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
+import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
 
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerRecord;
@@ -214,6 +216,9 @@ public class OpennmsKafkaProducer implements AlarmLifecycleListener, EventListen
     private void updateAlarm(String reductionKey, OnmsAlarm alarm) {
         // Always push null records, no good way to perform filtering on these
         if (alarm == null) {
+            // The alarm has been deleted so we shouldn't track it in the map of outstanding alarms any longer
+            outstandingAlarms.remove(reductionKey);
+            
             // The alarm was deleted, push a null record to the reduction key
             sendRecord(() -> {
                 LOG.debug("Deleting alarm with reduction key: {}", reductionKey);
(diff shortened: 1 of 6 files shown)


features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
```diff
@@ -191,7 +199,25 @@ public boolean shouldForwardAlarm(OnmsAlarm alarm) {
...
+
+    private void recordIncrementalAlarm(String reductionKey, OnmsAlarm alarm) {
+        // Apply the excluded fields when putting to the map so we do not have to perform this calculation
+        // on each equality check
+        outstandingAlarms.put(reductionKey,
+                AlarmEqualityChecker.Exclusions.defaultExclusions(protobufMapper.toAlarm(alarm)).build());
+    }
+
     private void updateAlarm(String reductionKey, OnmsAlarm alarm) {
+        if (suppressIncrementalAlarms && isIncrementalAlarm(reductionKey, alarm)) {
```
j-white: The `alarm` may be null here. I'd defer this check until after the filtering call to `shouldForwardAlarm`.

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
```diff
@@ -191,7 +199,25 @@ public boolean shouldForwardAlarm(OnmsAlarm alarm) {
...
+            suppressedAlarm.countDown();
```
j-white: Do we need this latch for anything?

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
```diff
@@ -191,7 +199,25 @@ public boolean shouldForwardAlarm(OnmsAlarm alarm) {
         return true;
     }
 
+    private boolean isIncrementalAlarm(String reductionKey, OnmsAlarm alarm) {
+        return outstandingAlarms.containsKey(reductionKey) &&
```
j-white: Multiple threads can call `updateAlarm` at the same time, so we need to make sure that the usage of the `outstandingAlarms` map is thread safe.

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
mattixtech: I was using it in the integration test but looks like I don't really need it since the test is already verifying via the consumer.

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
j-white: It's possible for the call to `equalsExcludingOnFirst` to throw an NPE since the map may change between the call to `containsKey` and `get`.

mattixtech: In this case the map may change, but not in such a way that the key no longer exists since I'm not aging entries out. So I don't think it could throw an NPE. However I will change it so that if we decide to age entries out in the future this won't break.

j-white: Right. We should be deleting alarms from the map then, otherwise the map will just continue to grow.

I would update the `updateAlarm` function to just delete the entry from the map when invoked with a `null` alarm.

PR #2136, opened 2018-08-30
NMS-10324: make the default relative time period 1 day again

RangerRick:
This PR fixes a long-standing regression and make the default relative time period 1 day again, rather than 1 hour.

I also add a property (`org.opennms.web.defaultGraphPeriod`) for setting the default, if you prefer the old behavior of 1 hour.

* JIRA: http://issues.opennms.org/browse/NMS-10324


cgorantla:
:+1:

PR #2138, opened 2018-08-31
HZN-1384: Extend /api/v2/alarms to provide better support for situations

smith-opennms:
JIRA: https://issues.opennms.org/browse/HZN-1384

This PR adds 2 attributes to Alarm (affectedNodeCount, inSituation) and 2 attributes to AlarmSummary (label, logMessage) to extend Correlation/Situation information.

AffectedNodeCount and inSituation can be used to filter rest requests.
Additionally, Alarms can now also be filtered on 'situationAlarmCount' i.e. the number of related alarms in the Situation.


opennms-model/src/main/java/org/opennms/netmgt/model/OnmsAlarm.java
```diff
@@ -1188,6 +1190,32 @@ public boolean isSituation() {
         return ! m_relatedAlarms.isEmpty();
     }
 
+    @Transient
+    @XmlTransient
+    public Integer getAffectedNodeCount() {
+        if (m_relatedAlarms == null || m_relatedAlarms.isEmpty()) {
+            return 1;
```
j-white: This should be 1 iff `m_node  != null`, 0 otherwise.

j-white:
+1

PR #2140, opened 2018-09-04
NMS-10331: Optionally create new alarms when a problem reoccurs

j-white:
JIRA: https://issues.opennms.org/browse/NMS-10331

Here we add a new flag that allows us to force alarmd to create a new alarm instead of reducing onto the existing alarm when it is already cleared.

This can be used to make sure there is a single `OnmsAlarm` entity per problem.


pbrane:
Tested locally.  Merge when ready.

PR #2143, opened 2018-09-07
NMS-10296: link to the node page if the minion has an associated node

RangerRick:
This PR adds a query to the minion status page to retrieve node information for the listed minions, and it turns the ID into a link to the node page if the minion has an associated node.

* JIRA: http://issues.opennms.org/browse/NMS-10296

### DIFF
--- a/opennms-webapp/src/main/webapp/js/angular-onms-elementList-minion.js
+++ b/opennms-webapp/src/main/webapp/js/angular-onms-elementList-minion.js
@@ -42,8 +42,10 @@
 	/**
 	 * Minion list controller
 	 */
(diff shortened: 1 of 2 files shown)


opennms-webapp/src/main/webapp/js/angular-onms-elementList-minion.js
```diff
@@ -64,6 +66,29 @@
 				function(value, headers) {
 					$scope.$parent.items = value;
 
+					if (value && value.length > 0) {
+						var query = '(' + value.map(function(minion) {
```
j-white: Should we also be matching on the `foreignSource` name here? Since it is technically possible for a different node in a another requisition to have the same id.

RangerRick: Yeah, I guess just assume the 'Minions' foreign source hardcoded?

PR #2144, opened 2018-09-07
NMS-10342: fix navbar rendering in node maps

RangerRick:
The node maps code, when ported to Vaadin 7, was given a servlet facade.  This facade is currently only used for the purposes of making a fake request that could be passed to get the header HTML, so I just removed it and changed it to work like the topology UI.

* JIRA: http://issues.opennms.org/browse/NMS-10342


features/vaadin-node-maps/src/main/java/org/opennms/features/vaadin/nodemaps/internal/HttpServletRequestVaadinImpl.java
```diff
@@ -449,11 +418,26 @@ public long getContentLengthLong() {
...
-        throw new RuntimeException("Cannot upgrade.");
+    public <T extends HttpUpgradeHandler> T upgrade(final Class<T> handlerClass) throws IOException, ServletException {
+        return getServletRequest().get().upgrade(handlerClass);
+    }
+
+    private Optional<VaadinServletRequest> getServletRequest() {
+        if (m_request instanceof VaadinServletRequest) {
+            return Optional.ofNullable((VaadinServletRequest)m_request);
+        }
+        return Optional.empty();
```
mvrueden: I am not sure if this is reasonable. Looking at the code as is, `getServletRequest().get()....` may throw a `NullPointerException`. I think in this case we should just throw an Exception if the request is not a `VaadinServletRequest`?

RangerRick: I just figured if something really weird happens, we want to just let it throw an exception anyways, and there isn't much point in trying to roll it up into some other exception.

j-white:
Can we add a simple smoke test that would have caught the problem?

features/vaadin-node-maps/src/main/java/org/opennms/features/vaadin/nodemaps/internal/HttpServletRequestVaadinImpl.java
j-white: I also think we would be better off throwing an exception if the we cannot cast the request to a `VaadinServletRequest` - we currently return an optional, but none of the callers treat it as optional.

Maybe just throw a IllegalStateException?

mvrueden: Actually we don't need the wrapper at all. See last commit from @RangerRick (https://github.com/OpenNMS/opennms/pull/2144/commits/b5ee9ca277db35be12a55e55a83c18350b247a07). We can just use the same code as the `TopologyUI` is doing :)

mvrueden:
Nice, more deletes than additions 👍

PR #2147, opened 2018-09-12
NMS-10345: Added feature to report vSphere alarms in VmwareMonitor

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-10345

### DIFF
--- a/integrations/opennms-vmware/src/main/java/org/opennms/netmgt/poller/monitors/VmwareMonitor.java
+++ b/integrations/opennms-vmware/src/main/java/org/opennms/netmgt/poller/monitors/VmwareMonitor.java
@@ -30,8 +30,14 @@ package org.opennms.netmgt.poller.monitors;
 
 import java.net.MalformedURLException;
 import java.rmi.RemoteException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
 
+import org.opennms.core.utils.PropertiesUtils;
 import org.opennms.core.utils.TimeoutTracker;
 import org.opennms.netmgt.poller.Distributable;
 import org.opennms.netmgt.poller.DistributionContext;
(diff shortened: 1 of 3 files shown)


opennms-base-assembly/src/main/filtered/etc/poller-configuration.xml
```diff
@@ -268,6 +268,7 @@
          <parameter key="retry" value="2"/>
          <parameter key="timeout" value="3000"/>
          <parameter key="ignoreStandBy" value="false"/>
+         <parameter key="reportAlarms" value=""/>
```
indigo423: Is it required to force a merge for updating users with an empty new introduced parameter?

christianpape: no, empty string is the default

indigo423: You could avoid touching the config-tester, by making reportAlarms="" when the whole parameter is missing and make it backward compatible.

christianpape: I think I understood you correctly, the default value "" is also applied in code even the reportAlarms entry does not exist. Of course we can also remove the line here if you feel better :-)

integrations/opennms-vmware/src/main/java/org/opennms/netmgt/poller/monitors/VmwareMonitor.java
```diff
@@ -83,6 +98,16 @@
...
         final boolean ignoreStandBy = getKeyedBoolean(parameters, "ignoreStandBy", false);
+
+        final List<String> severitiesToReport =
+                Splitter.on(",")
+                    .trimResults()
+                    .omitEmptyStrings()
+                    .splitToList(getKeyedString(parameters, "reportAlarms", ""))
+                    .stream()
+                    .filter(e -> VALID_VSPHERE_ALARM_STATES.contains(e))
+                    .collect(Collectors.toList());
```
christianpape: Here it is: getKeyedString(parameters, "reportAlarms", "") defaults to an empty string

indigo423:
I have tears in my eyes for this PR which includes documentation :)
@christianpape🥇

PR #2148, opened 2018-09-12
HZN-1333: "Change Icon" In Topology Map Fails

patrick-schweizer:
* JIRA: https://issues.opennms.org/browse/HZN-1333

### DIFF
--- a/smoke-test/src/test/java/org/opennms/smoketest/topo/GraphMLTopologyIT.java
+++ b/smoke-test/src/test/java/org/opennms/smoketest/topo/GraphMLTopologyIT.java
@@ -238,9 +238,25 @@ public class GraphMLTopologyIT extends OpenNMSSeleniumTestCase {
         Assert.assertEquals("Servers", topologyUIPage.getVisibleVertices().get(1).getLabel());
     }
 
-    /**
-     * Creates and publishes a requisition with 2 dummy nodes with predefined parameters
-     */
+    @Test
+    public void verifyCanChangeIcon() throws IOException, InterruptedException {
+        topologyUIPage.selectTopologyProvider(() -> LABEL);
+        topologyUIPage.findVertex("North Region").contextMenu().click("Navigate To", "Markets (North Region)");
+        String label = "North 1";
+
+        // open icon select popup
+        topologyUIPage.findVertex(label).select();
+        topologyUIPage.findVertex(label).contextMenu().click("Change Icon");
+
+        // choose new icon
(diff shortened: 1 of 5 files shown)


mvrueden:
At first glance this looks good to me. However this has an impact on all Vaadin applications not only the Topology Map, I'd like to test it out manually before merging, though. I'll come back to you, as soon as I had the chance.

mvrueden:
I tested it and LGTM as well. 
We should add a test for changing the icon though.

patrick-schweizer:
@mvruden I added a test that verifies that the icon can be changed.

smoke-test/src/test/java/org/opennms/smoketest/topo/GraphMLTopologyIT.java
```diff
@@ -238,9 +238,25 @@ public void verifyCanFilterByCategory() throws IOException, InterruptedException
...
+        topologyUIPage.findVertex("North Region").contextMenu().click("Navigate To", "Markets (North Region)");
+        String label = "North 1";
+
+        // open icon select popup
+        topologyUIPage.findVertex(label).select();
+        topologyUIPage.findVertex(label).contextMenu().click("Change Icon");
+
+        // choose new icon
+        String newIconName = "microwave_backhaul_1";
+        findElementByXpath("//*[name()='title' and text()='"+newIconName+"']/../*[name()='rect']").click();
```
mvrueden: Can we move this to a `setIcon(...)` method?

patrick-schweizer: to make sure: you want me to put the whole process of choosing an icon in the method?
That would be:
* click on an Element
* open the context menu via right click
* open the change icon dialog
* select an icon
* close the icon dialog

mvrueden: Hmm sorry, the comment was in state `pending`, but I wrote it 4 days ago, is it visible to you now?

patrick-schweizer: now it is :-) 
the code snipped you posted does steps 1-3 but I assume we also want step 4. I'll wrap it in a method.

mvrueden: Yeah, makes sense (-:

RangerRick:
Yay!  Looks good!

PR #2149, opened 2018-09-12
NMS-10332: fix minion init to honor sysconfig

RangerRick:
Karaf's `bin/inc` only initializes `JAVA_OPTS` if it isn't already set, so in most cases our default init scripts were ignoring most of the values set in `/etc/sysconfig/minion`.

This PR fixes that.

* JIRA: http://issues.opennms.org/browse/NMS-10332


j-white:
Installed the packages, uncommented some fields in `/etc/sysyconfig/minion` and restarted the service - the properties were applied.

PR #2150, opened 2018-09-13
HZN-1382: Persist collection sets from telemetry adapters in Newts on Sentinel

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1382


opennms-doc/guide-admin/src/asciidoc/text/sentinel/sentinel.adoc
```diff
@@ -374,3 +387,132 @@ Simply copy it to `${SENTINEL_HOME}/deploy/`.
     </feature>
 </features>
 -----
+
+=== Persisting Collection Sets to Newts
+
+In the previous chapter it is described on how to setup _{opennms-product-name}_, _Minion_ and _Sentinel_ in order to distribute the procession of flows.
```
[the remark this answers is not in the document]
j-white: (There are a few other references in this document to `procession` that should be updated too)

opennms-doc/guide-admin/src/asciidoc/text/sentinel/sentinel.adoc
```diff
@@ -374,3 +387,132 @@ Simply copy it to `${SENTINEL_HOME}/deploy/`.
...
+
+Please note, that in both cases the file `/opt/sentinel/etc/sflow-host.groofy` must be provided manually, e.g. by manually copying it over from _{opennms-product-name}_.
+
+===== NxosGpbAdapter
+
+In order to use this adapter, the feature `sentinel-telemetry-nxos` and `sentinel-newts` must be installed.
+In addition either `sentinel-jms` or `sentinel-kafka` should be installed and configured properly.
+See the previous _Flow Procession_ chapter for more details.
+
+Besides this, configuration files from _{opennms-product-name}_ must be copied to _Sentinel_ to `/opt/sentinel/etc`.
```
j-white: It looks like the `OpenNMS :: Features :: Distributed :: Collection` bundle requires these datacollection files to be in place, irrespective of which adapter is being used. Perhaps we should restructure the docs to reflect this?

mvrueden: The config files are only required by the `NxosGpbAdapter` and `JtiGpbAdapter`. At the moment the `SFlowTelemetryAdapter` does not require them, thus the structure of the documentation.

However we can make it always require those files, and can re-structure the docs a bit.

j-white: The bundle doesn't load without those configuration files in place. Doesn't the `SFlowTelemetryAdapter` require that bundle to be loaded in order to persist the collection sets?

mvrueden: Yes, the bundle requires those configuration files in place, however the `SFlowTelemetryAdapter` does not require that bundle.

The system test is also reflecting that, as it does not load the bundle, but the tests still succeed.

PR #2151, opened 2018-09-13
NMS-10352: Fix type for some Cassandra JMX values

brynjare:
* JIRA: https://issues.opennms.org/browse/NMS-10352

Changes type from 'gauge' to 'counter' for MemtableSwitchCount and
the DroppedMessage attributes, to match the JMX types.


mvrueden:
I pulled this into bamboo

j-white:
Thanks for fixing this.

PR #2152, opened 2018-09-13
NMS-10353: Add support for suppressAdminDownEvent

brynjare:
* JIRA: https://issues.opennms.org/browse/NMS-10353

These commits add support for suppressing *almost* all events related to changes in ifAdminStatus.

If the status change is because of ifAdminStatus:
- ifOperStatus DOWN will *only* be sent when ifAdminStatus goes UP
  while ifOperStatus remains DOWN.
- All ifAdminStatus events will be suppressed when suppressAdminDownEvent
  is enabled. The UP events are fairly useless without corresponding
  DOWN events.

If the status change is *not* related to a change in ifAdminStatus:
- ifOperStatus events will be sent as normal, unaffected by any
  of the changes related to re-adding suppressAdminDownEvent.

The default setting of suppressAdminDownEvent is also changed from `true` to `false`, since the option has been deprecated (and had no effect at all) since 2010.


mvrueden:
I pulled this into bamboo.

j-white:
@rssntn67 any input here?

mvrueden:
Tests passed, no feedback so far (from us).
What are we going to do about this?

PR #2153, opened 2018-09-13
NMS-10354: Add an option to suppress ifOperStatus events during initialization

brynjare:
* JIRA: https://issues.opennms.org/browse/NMS-10354

SNMP Interface Poller would generate duplicate operStatus Down events
after every restart for interfaces that was already down.

The new config option, 'suppressInitializationEvent', will ignore all
interfaces that were already marked as down in the snmpinterface table
before the restart.

This also suppresses all events while importing new devices.

Interfaces that change from Up to Down while OpenNMS is stopped will
be handled by doPoll(), which will create the expected events.

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/snmp-interface-poller-configuration.xml
+++ b/opennms-base-assembly/src/main/filtered/etc/snmp-interface-poller-configuration.xml
@@ -1,4 +1,4 @@
-<snmp-interface-poller-configuration xmlns="http://xmlns.opennms.org/xsd/config/snmpinterfacepoller" threads="30" service="SNMP">
+<snmp-interface-poller-configuration xmlns="http://xmlns.opennms.org/xsd/config/snmpinterfacepoller" threads="30" service="SNMP" suppressInitializationEvent="false">
    <node-outage>
       <critical-service name="ICMP"/>
       <critical-service name="SNMP"/>
--- a/opennms-base-assembly/src/main/filtered/etc/snmp-interface-poller-configuration.xml
+++ b/opennms-base-assembly/src/main/filtered/etc/snmp-interface-poller-configuration.xml
@@ -1,4 +1,4 @@
-<snmp-interface-poller-configuration xmlns="http://xmlns.opennms.org/xsd/config/snmpinterfacepoller" threads="30" service="SNMP" suppressInitializationEvent="false">
+<snmp-interface-poller-configuration xmlns="http://xmlns.opennms.org/xsd/config/snmpinterfacepoller" threads="30" service="SNMP">
    <node-outage>
       <critical-service name="ICMP"/>
       <critical-service name="SNMP"/>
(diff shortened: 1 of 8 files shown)


brynjare:
This commit will cause conflicts with [NMS-10353](https://issues.opennms.org/browse/NMS-10353) (#2152), but should both pull requests be considered for merging, I'd be happy to rebase or merge as required.

mvrueden:
I pulled this into bamboo.

brynjare:
> 
> 
> I pulled this into bamboo.

Thanks! Looks like the build system ran out of memory, however. Is there a problem with the code, or does it just need to be rerun?

j-white:
@rssntn67 any input here?

mvrueden:
> > I pulled this into bamboo.
> 
> Thanks! Looks like the build system ran out of memory, however. Is there a problem with the code, or does it just need to be rerun?

Probably some bamboo problems. I re-pulled it in to bamboo.

opennms-base-assembly/src/main/filtered/etc/snmp-interface-poller-configuration.xml
```diff
@@ -1,4 +1,4 @@
-<snmp-interface-poller-configuration xmlns="http://xmlns.opennms.org/xsd/config/snmpinterfacepoller" threads="30" service="SNMP">
+<snmp-interface-poller-configuration xmlns="http://xmlns.opennms.org/xsd/config/snmpinterfacepoller" threads="30" service="SNMP" suppressInitializationEvent="false">
```
j-white: We can omit this change, since the attribute is optional.

opennms-services/src/main/java/org/opennms/netmgt/snmpinterfacepoller/pollable/PollableSnmpInterface.java
```diff
@@ -165,18 +167,25 @@ public void setSnmpinterfaces(List<OnmsSnmpInterface> snmpinterfaces) {
...
+                // is enabled, which is likely to result in duplicate events for certain interfaces.
+                // (If an interface goes down while OpenNMS is stopped, doPoll() will handle the events at next run.)
+                if (iface.getIfAdminStatus() != null &&
         				iface.getIfAdminStatus().equals(SnmpMinimalPollInterface.IF_UP) && 
         				iface.getIfOperStatus() != null &&
         				iface.getIfOperStatus().equals(SnmpMinimalPollInterface.IF_DOWN) &&
         				(oldStatus == null || (iface.getIfOperStatus().intValue() != oldStatus.intValue()))) {
-        			sendOperDownEvent(iface);
+                    if (!getSuppressInitializationEvent()) {
+                        sendOperDownEvent(iface);
```
j-white: Can we add test coverage for this?

syepes:
+1 Hope to see this included in 23.0.2

opennms-base-assembly/src/main/filtered/etc/snmp-interface-poller-configuration.xml
brynjare: Config option removed in 8350b4e

mvrueden:
No feedback from @rssntn67. @jeffgdotorg, @indigo423 can you provide feedback here?

PR #2154, opened 2018-09-14
HZN-1383: Fix for an issue where kafka producer alarm suppression would not work for situation alarms

mattixtech:
* JIRA: http://issues.opennms.org/browse/HZN-1383


mattixtech:
@j-white Hi Jesse, I haven't had a chance to test this yet but I think this should address the issue you were seeing where situations were getting sent via the Kafka producer on every sync.

j-white:
Looks like that should do it. Can we add a test case for this one?

mattixtech:
Added test case to cover the fix.

PR #2155, opened 2018-09-17
NMS-10357: Fix Collectd and Pollerd's ThreadPool graphs

brynjare:
* JIRA: https://issues.opennms.org/browse/NMS-10357

Adds missing pool size attribute ('numPoolThreads') to Pollerd's
MBean, and adds all the PoolThreads attributes (num, max, peak) to
the JMX datacollection config for Pollerd.

Also makes it slightly more obvious what numPoolThreads actually is,
by changing the @return comment to match the official javadocs.

Finally, the graph properties for the ThreadPool attributes have been
changed to graph the active threads, peak threads and max pool size.

The original graph included both numPoolThreads and maxPoolThreads,
which seem to be identical when using fixed thread pools.
numPoolThreads was replaced by active threads, which made the
standalone graphs redundant.

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/jmx-datacollection-config.xml
+++ b/opennms-base-assembly/src/main/filtered/etc/jmx-datacollection-config.xml
@@ -23,6 +23,9 @@
          </mbean>
          <mbean name="OpenNMS Pollerd" objectname="OpenNMS:Name=Pollerd">
             <attrib name="NumPolls" alias="ONMSPollCount" type="counter"/>
+            <attrib name="NumPoolThreads" alias="ONMSPollerPoolThrd" type="gauge"/>
+            <attrib name="MaxPoolThreads" alias="ONMSPollerPoolMax" type="gauge"/>
+            <attrib name="PeakPoolThreads" alias="ONMSPollerPoolPeak" type="gauge"/>
             <attrib name="ActiveThreads" alias="ONMSPollerThreadAct" type="gauge"/>
             <attrib name="TasksTotal" alias="ONMSPollerTasksTot" type="counter"/>
             <attrib name="TasksCompleted" alias="ONMSPollerTasksCpt" type="counter"/>
--- a/opennms-base-assembly/src/main/filtered/etc/snmp-graph.properties.d/opennms-graph.properties
+++ b/opennms-base-assembly/src/main/filtered/etc/snmp-graph.properties.d/opennms-graph.properties
@@ -1,7 +1,7 @@
 reports=onms.manager.uptime, onms.queued.updates, onms.queued.pending, \
-onms.pollerd.activeThreads, onms.pollerd.completedRatio, onms.pollerd.polls, \
+onms.pollerd.threadpool, onms.pollerd.completedRatio, onms.pollerd.polls, \
 onms.pollerd.taskqueue, \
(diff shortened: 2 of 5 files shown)


mvrueden:
I pulled this into bamboo.

syepes:
Would be great to see this one on the next bleeding image ;-)

indigo423:
I've had to re-run bamboo to see if it pass all tests. Meanwhile I just test the configs in a local setup. Thank you very much for your contribution

pioto:
Will this change have negative implications for people with existing .rrd files who are using `org.opennms.rrd.storeByGroup=true`? (I seem to recall having an issue when I added other metrics before).

indigo423:
@pioto yes, when you have `storeByGroup` enabled, you have to deal with it. a) recreate the RRD file new and loose old data, b) don't apply the update when you update c) migrate the changed metrics in a new group.

roskens:
Since both PollerD and CollectD use a fixed size thread pool are these additions useful?

pbrane:
> On Sep 27, 2018, at 16:06, Ron Roskens <notifications@github.com> wrote:
> 
> Since both PollerD and CollectD use a fixed size thread pool are these additions useful?
> 
Ironically, I Jesse and I discussed making these pools unbounded, today. :)


David



David Hustace
The OpenNMS Group, Inc.

roskens:
> On Sep 27, 2018, at 16:06, Ron Roskens ***@***.***> wrote: Since both PollerD and CollectD use a fixed size thread pool are these additions useful?
> Ironically, I Jesse and I discussed making these pools unbounded, today.

That would be nice to see too. Was just wanting to note that for a fixed pool size, peak threads == core threads so the graph with both would just be lines on top of each other. Although you'll see it fluctuate due to the occasional thread exiting on an exception.

brynjare:
> Since both PollerD and CollectD use a fixed size thread pool are these additions useful?

Having pool size information for Pollerd was the main reason I started looking into this in the first place. Pollerd was missing all pool size metrics entirely, as they weren't even exposed through JMX, but I agree that PeakPoolThreads and MaxPoolThreads are pretty much useless with the current implementation. I wasn't aware of this until I started working on this code, but I decided to keep them anyway since they were already present in Collectd. :)

The main benefit for Collectd would be to have the active threads graphed along with the actual pool size. Just a minor improvement (imho) since I was already fixing Pollerd.

j-white:
+1. I think we should retarget this enhancement to `develop` instead of `foundation-2018` though, that way we don't alter the `jmx-datacollection-config.xml` which is commonly edited.

brynjare:
> +1. I think we should retarget this enhancement to `develop` instead of `foundation-2018` though, that way we don't alter the `jmx-datacollection-config.xml` which is commonly edited.

Makes sense! I've switched the base to develop now. (Please let me know if merging in develop isn't the right way to do it, and I'll see if I can find some other way to rebase while keeping the same pull request.)

indigo423:
@j-white is this something we should consider against 23.0.0 or should we merge it into develop for 24.0.0?

j-white:
Pushed to `pr/jira/NMS-10357`.

@indigo423 given that 23.0.0 is right around the corner, I'd defer this to 24.

mvrueden:
Pulled it into bamboo again.

PR #2156, opened 2018-09-21
Lombok examples

patrick-schweizer:
This PR is a showcase of how Projekt Lombok could be used in OpenNMS. It is not intended to be merged but as a base for discussion.

Projekt Lombok reduces boilerplate code significantly (especially getters, setters, toString, builders) and helps to avoid common hard to detect problems ( especially equals() and hashCode() implementation). Read more about Lombok here: https://projectlombok.org
Feature List: https://projectlombok.org/features/all

Please look at my comments in the changed files for additional information.

In order to open this branch in InteliJ and play around with it it, it is recommended that you enable the Lombok Plugin (File -> Settings -> Plugins -> Lombok Plugin). The plugin enables a seamless integration into InteliJ.


opennms-webapp/src/main/java/org/opennms/web/services/FilterFavoriteService.java
```diff
@@ -193,11 +193,8 @@ protected boolean deleteFavorite(OnmsFilterFavorite favorite) {
         return false;
     }
 
+    @SneakyThrows(UnsupportedEncodingException.class)
```
patrick-schweizer: We replace here the kind of useless try / catch (since every Java platform has to support UTF-8) by this annotation. See more details here: https://projectlombok.org/features/SneakyThrows

smoke-test/src/test/java/org/opennms/smoketest/TopologyIT.java
```diff
@@ -70,10 +72,9 @@
  *
  * @author jwhite
  */
+@Slf4j
```
patrick-schweizer: Logger construction is here replaced with an annotation (small improvement). Read more: https://projectlombok.org/features/log

smoke-test/src/test/java/org/opennms/smoketest/TopologyIT.java
```diff
@@ -82,24 +82,15 @@
      * Represents a vertex that is focused using
      * criteria listed bellow the search bar.
      */
+    @Data
```
patrick-schweizer: Here it gets interesting: '@Data' generates Getters, Setters, equals() and hashCode() methods,  '@NonNull' creates a null check.
42 lines of code become 4 lines :-)

Read more here:
- https://projectlombok.org/features/Data
- https://projectlombok.org/features/NonNull

smoke-test/pom.xml
```diff
@@ -377,6 +377,12 @@
       <version>3.0.3</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
patrick-schweizer: Lombok needs to present only during compile time, see als: https://projectlombok.org/setup/maven

mvrueden:
I am not a big fan of code generation. Yes it reduces the amount of source code, but generates it afterwards anyways. We have so much complexity already, that adding lombok to our stack in my opinion we do not gain much. Setters and Getters can be generated with any reasonable IDE in no time. The same goes for hashCode and equals.

patrick-schweizer:
@mvrueden I see it a bit differently: Think of Lombok as a language extension and not as a code generator. Yes it does generate code under the hood but you don't see it and therefor you reduce visible boilerplate code. I don't think an IDE does the same: It generates visible source code and only once. E.g. if you add a field to a class your equals and hash code methods are sort of outdated. You need to make sure you take that field in those methods as well into consideration.
Lombok brings to some extend the idea of convention over configuration to the generation of boilerplate methods. E.g. you define a getter only if it doesn't do the standard thing.

To me Java without Lombok feels to some extend like it was with Java 1.4. Yes you can live without generics but you don't want to ;-)

Having said that OpenNMS is of course a very big project and I would be very hesitant in bringing another technology into it as well. But maybe there is a part of OpenNMS or a tool where we could try it. It is also no problem to "delombok" if we decide it is not worth it or it makes problems. in InteliJ: right click on project -> refactor -> delombok. See also: https://projectlombok.org/features/delombok

mvrueden:
I worked on some projects using lombok and yes I get your point, but in the end it comes down to my personal preference when working on code. And in contraire to your personal preference, I just don't like it (-:

j-white:
I don't have a strong opinion on this one. On one hand it helps save some boilerplate code, but on the other hand it adds more complexity to the build.

I remember seeing a thread on the ODL mailing lists about this, for reference:
  https://lists.opendaylight.org/pipermail/dev/2016-October/002877.html

mvrueden:
If we want to introduce this, we should definetly wait until we are on the Java 9, if not even the Java >= 11 train in order to ensure that we are not introducing problems not allowing us to migrate. 
Until then I propose to close this PR and reopen the discussion if required sometimes later. Feel free to reopen this PR if you feel the need to do it now.

PR #2160, opened 2018-09-25
NMS-10350: Added a reloadable attribute to the service-configura…

Zottel92:
jira: https://issues.opennms.org/browse/NMS-10350

### DIFF
--- a/opennms-config-model/src/main/resources/xsds/service-configuration.xsd
+++ b/opennms-config-model/src/main/resources/xsds/service-configuration.xsd
@@ -42,6 +42,7 @@
       </sequence>
 
       <attribute name="enabled" type="boolean" use="optional" default="true" />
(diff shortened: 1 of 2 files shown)


opennms-config-model/src/main/resources/xsds/service-configuration.xsd
```diff
@@ -42,6 +42,7 @@
       </sequence>
 
       <attribute name="enabled" type="boolean" use="optional" default="true" />
+      <attribute name="reloadable" type="boolean" use="optional" default="true" />
```
mvrueden: This should be either mandatory or `default="false"`

j-white:
LGTM. Sure, we can add a line on the release notes explaining why it's there.

mvrueden:
Following up issue was created: https://issues.opennms.org/browse/NMS-10411

PR #2161, opened 2018-09-25
NMS-10370: fix misleading INFO log message

brynjare:
* JIRA: https://issues.opennms.org/browse/NMS-10370

`2018-09-25 20:46:27,735 INFO  [scanExecutor-8] o.o.n.p.s.ForceRescanScan: The node with ID OP does not have a primary interface` will no longer be printed for nodes with valid primary SNMP interfaces.

Should the node lack a primary SNMP interface, for any reason, the log message will be printed with the proper node ID:
`2018-09-25 21:04:48,939 INFO  [scanExecutor-3] o.o.n.p.s.ForceRescanScan: The node with ID 40 does not have a primary interface`

### DIFF
--- a/opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/ForceRescanScan.java
+++ b/opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/ForceRescanScan.java
@@ -92,8 +92,9 @@ public class ForceRescanScan implements Scan {
             OnmsIpInterface iface = m_provisionService.getPrimaryInterfaceForNode(node);
             if (iface == null) { // NMS-6380, a discovered node added with wrong SNMP settings doesn't have a primary interface yet.
                 iface = node.getIpInterfaces().isEmpty() ? null : node.getIpInterfaces().iterator().next();
-            } else {
-                LOG.info("The node with ID OP does not have a primary interface", m_nodeId);
+                if (iface != null) {
+                    LOG.info("The node with ID {} does not have a primary interface", m_nodeId);
+                }
             }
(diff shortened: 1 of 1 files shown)


opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/ForceRescanScan.java
```diff
@@ -92,8 +92,9 @@ protected void scanExistingNode(final BatchTask phase) {
             OnmsIpInterface iface = m_provisionService.getPrimaryInterfaceForNode(node);
             if (iface == null) { // NMS-6380, a discovered node added with wrong SNMP settings doesn't have a primary interface yet.
                 iface = node.getIpInterfaces().isEmpty() ? null : node.getIpInterfaces().iterator().next();
-            } else {
-                LOG.info("The node with ID OP does not have a primary interface", m_nodeId);
+                if (iface != null) {
```
mvrueden: This is just a very nit comment, but why change the else to an if, if you only want to adjust the log message?

brynjare: It wasn't just changed from "else" to "if"; I also moved the log statement inside the "if (iface == null) { }" block.
(The "OP" typo was a less critical issue, and I probably wouldn't have bothered to do a pull request for that alone.)

The log message used to trigger when iface actually existed, which was the opposite of what I presume the original author intended. Now it will correctly trigger *if* there is an interface, but it's not the primary SNMP interface, which is more aligned with the actual log message.

PR #2162, opened 2018-09-26
NMS-10367: Fix ttl issue with collection:collect karaf command

cgorantla:
* JIRA: https://issues.opennms.org/browse/NMS-10367


j-white:
Is the SnmpMonitor affected in a similar fashion when invoked via the `poller:poll` command?

agalue:
No. On my tests, `poller:poll` respects the `-t` or `--ttl`.

cgorantla:
Looks like proxy-rpc is getting used by collector and snmpinterface poller. So `poller:poll` is not affected

PR #2164, opened 2018-09-27
NMS-10372: Update MBeans to work with Horizon 22

brynjare:
* JIRA: https://issues.opennms.org/browse/NMS-10372

This pull requests updates the MBeans in minion.xml to work with minions running on 22.0.4.

(NB: I have used foundation-2018 as the base, but I'm not sure if this could create problems for Meridian, if they still use the old object names.)


j-white:
LGTM. Pushed to `pr/jira/NMS-10372`. M2018 uses the same code in this case, so it's worth updating the configuration files.

PR #2166, opened 2018-09-28
NMS-10377: Added timeout property for VMware provisiong/polling/data collection

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-10377


christianpape:
I'll rework this one to pass the timeout parameters via the provisioning URLs.

PR #2167, opened 2018-09-30
NMS-10379: Other classes that use Http

patrick-schweizer:
* JIRA: https://issues.opennms.org/browse/NMS-10379


features/ncs/ncs-northbounder/src/main/java/org/opennms/netmgt/ncs/northbounder/NCSNorthbounderConfig.java
```diff
@@ -115,10 +115,12 @@ HttpEntityEnclosingRequestBase getRequestMethod(URI uri) {
     @XmlAttribute(name="fragment", required=false)
     private String m_fragment;
 
+    @XmlAttribute(name="use-system-proxy")
```
patrick-schweizer: I didn't find any documentation for this configuration. @agalue did I miss it or is there none?

features/datachoices/src/main/java/org/opennms/features/datachoices/internal/UsageStatisticsReporter.java
```diff
@@ -74,6 +74,8 @@
 
     private AlarmDao m_alarmDao;
 
+    private boolean m_useSystemProxy = true; // true == legacy behaviour
```
agalue: I believe it should be `m_useSystemProxy = false; // false == legacy behaviour`.

features/geocoder/nominatim/src/main/java/org/opennms/features/geocoder/nominatim/NominatimGeocoderService.java
```diff
@@ -55,12 +55,16 @@
     private String m_emailAddress;
     private String m_referer;
 
+    private boolean m_useSystemProxy = true; // true == legacy behaviour
```
agalue: Similar situation here: I believe it should be `m_useSystemProxy = false; // false == legacy behaviour.`

features/datachoices/src/main/java/org/opennms/features/datachoices/internal/UsageStatisticsReporter.java
patrick-schweizer: @agalue if you look at line 131 (code on left side - before my change) you can see that clientWrapper.useSystemProxySettings(); was always called - therefor I think this is the legacy behavior. We could set `private boolean m_useSystemProxy = false;` this would make the behavior consistent with the other classes that use HttpClientWrapper.  That might be the better long term solution. What you you think? Shall I change it?

features/geocoder/nominatim/src/main/java/org/opennms/features/geocoder/nominatim/NominatimGeocoderService.java
patrick-schweizer: see my comment above. Here it would be line 63. Please let me know if I should change it to false. Thanks!

agalue:
Ohh, that is definitely not expected and counterintuitive, but if that is the "current" behavior for these classes I believe we can live with it (even if that is not the case for the rest of them).

I don't have a strong opinion, so I'd guess we can keep the changes to be consistent with the "current" behavior, instead of the "desired" behavior. I'll leave that to you, unless someone else wants to jump in and suggest something different.

patrick-schweizer:
@agalue if I have the choice I would change it to "false" to have a consistent behavior. Let's ask jesse for his opinion: @j-white  are you ok with that approach?

j-white:
I'd opt to keep the existing defaults. This was purposely turned on by default in the `UsageStatisticsReporter`.

patrick-schweizer:
@j-white, @agalue leaving the defaults as they are that means the PR should be ready to be merged, or is there anything else that needs to be changed? Thanks

container/karaf/src/main/filtered-resources/etc/org.opennms.features.geocoder.nominatim.cfg
```diff
@@ -1,2 +1,3 @@
 emailAddress=
 referer=
+useSystemProxy=false
```
j-white: Can we add the default in the blueprint.xml instead of editing this configuration file?

core/web/src/main/java/org/opennms/core/web/HttpClientWrapperConfigHelper.java
```diff
@@ -0,0 +1,43 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with OpenNMS(R).  If not, see:
+ *      http://www.gnu.org/licenses/
+ *
+ * For more information contact:
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+package org.opennms.core.web;
```
j-white: FYI, we're getting merge conflicts here now that the other branches have been merged.

PR #2169, opened 2018-10-01
Fixed references and removed orphaned files in our docs

indigo423:
After merging admin and user guide and rebuilding the installation guide. I've notices missing references and orphaned files in our docs.

The orphaned files are not rendered in current install guide and are related to:

* Guide how to install Cassandra on Debian/RHEL/Windows cause this is described in the Cassandra project
* Guide how to install Oracle Java on Debian/RHEL/Windows cause this is described in the Oracle JDK project and we ship a JDK for a basic setup with the packages
* Dedicated RRDTool guide which is not used anymore, cause we ship RRDTool as an package dependency


indigo423:
@RangerRick thanks for rebase

PR #2171, opened 2018-10-01
NMS-10382: SNMP special configs are missing in the docs

mfuhrmann:
In development channel on Mattermost Jesse told me to try the SNMP properties from `opennms.properties` in Minion config to solve my datacollection issue (since it works in OpenNMS too). But we don't have documentation about those settings in general and there is no hint, that it also works in Minions. 

So I've added the content of SNMP section in `opennms.properties` into a adoc file.

* JIRA: http://issues.opennms.org/browse/NMS-10382


mfuhrmann:
I was told to put it into the docs. So, shall I also remove this part from `opennms.properties` file?

mfuhrmann:
@j-white @pbrane You pointed me in Mattermost into this direction to document this part  Since I'm now in the same situation like @indigo423 was 3 years ago ( see #278 ), we need decisions.

pbrane:
I'm not following.

mvrueden:
If you need clarification, it probably would be best if you set up a call with the people you need to discuss any problems, rather than discussion it here @mfuhrmann to move forward from here.

j-white:
The issue here is that the properties in question also apply to Minion, whereas they are currently only documented in `$OPENNMS_HOME/etc/opennms.properties`.

Short term, we could add a note to the Minion section in the admin guide that explains how to set system properties on the Minion (by adding entries to `$MINION_HOME/etc/custom.properties`) and mention that the SNMP4J related properties defined in `$OPENNMS_HOME/etc/opennms.properties` can also be set there.

Longer term, we need to figure out a better strategy for documenting and managing the system properties...

pbrane:
By convention, most applications with properties files have comments in the properties file to give a hint about a setting that isn't completely understood.  IMHO, the properties file should only have hints.  OpenNMS prosperities files have traditionally had more than just these hints because there wasn't a sufficient documentation effort in the beginning.

I believe that we should leave hints in the properties file where it makes sense and fully document each property in the official documentation.

Example:
```
# Some SNMP agents claim to support SNMPv2c, but either completely fail to
# respond to GET-BULK requests or return bad responses to them under some
# conditions. To disable the use of GET-BULK PDUs globally in the SNMP4J
# strategy, set this property to true. This change will adversely affect the
# efficiency of SNMP data collection from agents that properly support
# GET-BULK.
org.opennms.snmp.snmp4j.noGetBulk=false
```
vs.
```
# Disable GET-BULK requests on SNMPv2c agents;
# Default: false
org.opennms.snmp.snmp4j.noGetBulk=false
```
Parenthetically, I see a problem with this configuration in that the default isn't commented.  It should be:
```
# Disable GET-BULK requests on SNMPv2c agents
# Default: false
#org.opennms.snmp.snmp4j.noGetBulk=false
```

indigo423:
Just as an information, the initial work was rejected to migrate this to the versioned docs, you can follow that discussion here: https://github.com/OpenNMS/opennms/pull/278. Seems like we have to revise the decision made it in the first place :)

j-white:
Let's get this merged into the docs and just document in both places for now.

mfuhrmann:
Can someone help me with this broken tests?

indigo423:
Made branch for this PR and updated see #2936

PR #2173, opened 2018-10-01
NMS-10384: Set JVM flag to start Attach listener by default

jeffgdotorg:
* JIRA: http://issues.opennms.org/browse/NMS-10384


j-white:
I'm not aware of any downsides of settings this by default.

PR #2177, opened 2018-10-04
NMS-10377: Added timeout parameters for VMware provisioning

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-10377


mvrueden:
LGTM, can we add a test, that verifies that the timeouts are actually applied?

pbrane:
> LGTM, can we add a test, that verifies that the timeouts are actually applied?

Welcome back!

PR #2178, opened 2018-10-04
NMS-10388: Fix issues with Sentinel karaf.features.cfg.

mattixtech:
Also added placeholders in karaf.features.cfg to make it easier to programmatically add features to the featuresBoot.

* JIRA: http://issues.opennms.org/browse/NMS-10388


mvrueden:
LGTM at first glance. However I would opt for not using the touch command. If we need something like that, a better place may be the system test api.

mattixtech:
@mvrueden I only added the touch in there to try to narrow in on why the tests are failing since I added my changes.

The behaviour I've been seeing is that with the featuresBoot re-ordered the deploy directory is not reliable at starting features as part of initial startup. To solve this in the end to end tests I wrote I'm instead updating the featuresBoot before startup and that has been 100% reliable. I'm not sure how to address the tests that have already been written that are relying on hot-deploy at initial startup. Updating the file via touch or some other mechanism was one thought but it seems that a test still failed (I think there is 7 tests that are now intermittently failing since my changes). It does look like the number of failures went down though.

Ideally looking for a small change that I can make to fix the existing tests that are failing.

mvrueden:
Hmm, that sounds weird. Ideally it should work without touching anything. Karaf only reads the data from the deploy directory after it has been started. That means by definition after ALL features and bundles of the featuresBoot have been resolved and are active. 

If you like, I can take a look or we can take a look at it together and see if there is an alternate solution?

PR #2179, opened 2018-10-05
NMS-10300:  add health:check for kafka connectivity on minion/sentinel

cgorantla:
Add health:check to check kafka connectivity from Sink and RPC modules.
Move common classes to  ipc/common/kafka
Refactor all kafka related classes to use KafkaConfigProvider and Classloader Utils.
Fix kafka class loader issue when using authentication

* JIRA: http://issues.opennms.org/browse/NMS-10300
* JIRA: http://issues.opennms.org/browse/HZN-1387


core/ipc/common/kafka/src/main/java/org/opennms/core/ipc/common/kafka/KafkaHealthCheck.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public KafkaHealthCheck(OsgiKafkaConfigProvider kafkaConfigProvider, String type) {
+        this.osgiKafkaConfigProvider = kafkaConfigProvider;
+        this.type = type;
+    }
+
+    public KafkaHealthCheck(String sysPropPrefix, String type) {
+        this.kafkaConfigProvider = new OnmsKafkaConfigProvider(sysPropPrefix);
+        this.type = type;
+    }
```
mvrueden: Can we not just use the `ConfigProvider` here instead of the concrete implementations? I'd rather see that than 2 different constructors.

core/ipc/common/kafka/src/main/java/org/opennms/core/ipc/common/kafka/KafkaHealthCheck.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @Override
+    public String getDescription() {
+        return "Connecting to Kafka from " + type ;
+    }
+
+    @Override
+    public Response perform() throws Exception {
+        if (osgiKafkaConfigProvider != null) {
```
mvrueden: See above, if there were a single `configProvider` this could be plainly `kafkaConfig = configProvider.getProperties()` instead.

core/ipc/sink/kafka/server/src/main/java/org/opennms/core/ipc/sink/kafka/server/offset/KafkaOffsetProvider.java
```diff
@@ -221,8 +222,12 @@ public void shutdown() {
 
     }
 
+    public KafkaOffsetProvider() {
+        this(new OnmsKafkaConfigProvider(KafkaSinkConstants.KAFKA_CONFIG_SYS_PROP_PREFIX));
+    }
+
     public KafkaOffsetProvider(KafkaConfigProvider configProvider) {
-        this.configProvider = Objects.requireNonNull(configProvider);
+        this.configProvider = configProvider;
```
mvrueden: The `configProvider` is still required, isn't it? If so, we should still have `Objects.requireNonNull(...)` as a NotNull check.

core/ipc/common/kafka/src/main/java/org/opennms/core/ipc/common/kafka/KafkaHealthCheck.java
cgorantla: true, I think I didn't refactor this after making all the changes.

core/ipc/common/kafka/src/main/java/org/opennms/core/ipc/common/kafka/Utils.java
```diff
@@ -47,4 +47,15 @@
             Thread.currentThread().setContextClassLoader(classLoader);
         }
     }
+
+    public static <T> T runWithGivenClassLoader(final Supplier<T> supplier, ClassLoader classLoader) {
+        Objects.requireNonNull(supplier);
```
mvrueden: Can we use `runWithNullcontextClassLoader` use this method as well, instead of duplicating the code (more or less) here?

core/ipc/common/kafka/src/main/java/org/opennms/core/ipc/common/kafka/KafkaHealthCheck.java
```diff
@@ -33,44 +33,35 @@
...
-        if (osgiKafkaConfigProvider != null) {
-            kafkaConfig = osgiKafkaConfigProvider.getProperties();
-        }
-        if (kafkaConfigProvider != null) {
-            kafkaConfig = kafkaConfigProvider.getProperties();
-        }
-        // Default 1 sec should be sufficient, need changes to health check API to make it configurable.
-        kafkaConfig.put("request.timeout.ms", 1000);
+    public Response perform(Context context) throws Exception {
+        kafkaConfig = kafkaConfigProvider.getProperties();
```
mvrueden: The change here, will it create a timeout response? To me it looks like it would create a failure response, which is probably not what is intended. In general a HealthCheck implementation should not worry about time outs, but an implementation of the HealthCheckService should, see https://github.com/OpenNMS/opennms/blob/7840afa2e7802c7520c9347fc35552595a767d7d/core/health/impl/src/main/java/org/opennms/core/health/impl/DefaultHealthCheckService.java#L127.

If a HealthCheck must implement a timeout itself, we should verify if the DefaultHealthCheckService is not enforcing the timeout correctly, before we do that.

cgorantla: It will not create timeout response but here` listTopicsResult.names().get()`  itself is a future, so this timeout is for this future to exit.

core/ipc/common/kafka/src/main/java/org/opennms/core/ipc/common/kafka/Utils.java
cgorantla: yup, makes sense.

core/ipc/common/kafka/src/main/java/org/opennms/core/ipc/common/kafka/KafkaHealthCheck.java
mvrueden: Okay, makes sense, but can we also encounter for timeout responses then?

cgorantla: most likely command will timeout before the internal future timesout. So user will get timeout response.

core/ipc/common/kafka/src/main/java/org/opennms/core/ipc/common/kafka/KafkaHealthCheck.java
```diff
@@ -0,0 +1,73 @@
...
+
+
+    @Override
+    public String getDescription() {
+        return "Connecting to Kafka from " + type ;
+    }
+
+    @Override
+    public Response perform(Context context) throws Exception {
+        kafkaConfig = kafkaConfigProvider.getProperties();
```
mvrueden: I tried this and was never able to get a success response.

In addition I think the `kafkaConfig` should be local to the method and not a member variable of the `HealthCheck`, as it is overwritten on each call anyways.

mvrueden:
LGTM 👍 
One thing I noticed is, that the `HealthCheck` is not available on OpenNMS, but don't we want the check also there?

cgorantla:
> LGTM 
> One thing I noticed is, that the `HealthCheck` is not available on OpenNMS, but don't we want the check also there?

Yeah, with current  `HealthCheck` is in osgi and all the Kafka Impls are in default classloader, it is not loading KafkaHealthCheck with the current structure. we can discuss on how to proceed with this.

mvrueden:
Yeah, we don't have a clear strategy on what we want on OpenNMS in terms of `HealthCheck`s. 
We should discuss this at some point. 
However, that is not that relevant for this issue, as we don't perform a JMS check either (on OpenNMS).

RangerRick:
Sounds like everything is cool with this as-is and more work needs to be done for other issues, so I'll go ahead and merge it.

PR #2181, opened 2018-10-05
HZN-1333:  "Change Icon" In Topology Map Fails

patrick-schweizer:
* JIRA: https://issues.opennms.org/browse/HZN-1333

### DIFF
--- a/smoke-test/src/test/java/org/opennms/smoketest/TopologyIT.java
+++ b/smoke-test/src/test/java/org/opennms/smoketest/TopologyIT.java
@@ -39,7 +39,6 @@ import java.lang.reflect.Field;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
-import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


mvrueden:
Looking at the failed screen shot, it seems that the icon is not selected correctly, as it is showing an empty icon.

mvrueden:
I looked at this and found a couple of issues:
1. If a `WebElement` is retrieved calling `findElement(s)` with a global search pattern like `//*` does not guarantee that it is using the local view, meaning it is not search FROM the first `WebElement`, but it actually means search globally. If a search FROM the returned `WebElement` must be performed, add a `.` in front.
2. The `select()` method was not implemented correctly.
3. Waiting a number of seconds is sometimes required, but in most cases waiting for the actual element to be present is more elegantly.
4. Invoking `select()` before `contextMenu()` is not necessary.

Let's see if my changes make bamboo more happy.

mvrueden:
I looked at this and found a couple of issues:
1. If a `WebElement` is retrieved calling `findElement(s)` with a global search pattern like `//*` does not guarantee that it is using the local view, meaning it is not search FROM the first `WebElement`, but it actually means search globally. If a search FROM the returned `WebElement` must be performed, add a `.` in front.
2. The `select()` method was not implemented correctly.
3. Waiting a number of seconds is sometimes required, but in most cases waiting for the actual element to be present is more elegantly.
4. Invoking `select()` before `contextMenu()` is not necessary.

Let's see if my changes make bamboo more happy.

mvrueden:
Github is fishy, it is not showing my comments anymore and also at the moment not my commits. I am going to wait and see if they show up later.

j-white:
Tests passed twice in a row in Bamboo. Looks much better now.

mvrueden:
Yep, I gave it just another kick. If it passes again, I'd say it is good to be merged (-:

PR #2185, opened 2018-10-11
NMS-10378: Alarm processing is very slow when Kafka producer is enabled and Kafka is unavailable

mattixtech:
* JIRA: http://issues.opennms.org/browse/NMS-10378


mattixtech:
Hi Jesse, Not sure what to set the queue capacity to, does it need to be configurable or is a fixed value ok?

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
```diff
@@ -294,28 +306,53 @@ private void sendRecord(Callable<ProducerRecord<String,byte[]>> callable) {
...
+                            return;
+                        }
+                        if (consumer != null) {
+                            consumer.accept(recordMetadata);
+                        }
+                    });
+                } catch (RuntimeException e) {
+                    LOG.warn("Failed to send record to producer: {}.", producerRecord, e);
+                }
+            } catch (InterruptedException ignore) {
```
j-white: We should exit the thread when interrupted i.e. bundle/system is stopping

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
```diff
@@ -131,6 +140,9 @@ public void init() throws IOException {
         if (forwardEvents) {
             eventSubscriptionService.addEventListener(this);
         }
+        
+        // Start processing records that have been queued for sending
+        kafkaSendQueueExecutor.execute(this::processKafkaSendQueue);
```
j-white: We should also stop this thread in the `destroy()` method.

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
```diff
@@ -96,6 +100,11 @@
     private final AlarmEqualityChecker alarmEqualityChecker =
             AlarmEqualityChecker.with(AlarmEqualityChecker.Exclusions::defaultExclusions);
 
+    // TODO: Determine an appropriate queue capacity
+    private final BlockingQueue<KafkaRecord> kafkaSendQueue =
+            new LinkedBlockingQueue<>(1000);
+    private final Executor kafkaSendQueueExecutor = Executors.newSingleThreadExecutor();
```
j-white: We should use the constructor that takes a ThreadFactory, so that we can set the name of the thread - which is useful for identifying it in a thread dump.

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
```diff
@@ -96,6 +100,11 @@
     private final AlarmEqualityChecker alarmEqualityChecker =
             AlarmEqualityChecker.with(AlarmEqualityChecker.Exclusions::defaultExclusions);
 
+    // TODO: Determine an appropriate queue capacity
+    private final BlockingQueue<KafkaRecord> kafkaSendQueue =
```
j-white: Can we expose the queue size as a property in the blueprint so that it can be configured?

mattixtech:
Added the capacity as a property with default value of 1000. Do we want to document this property?

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
```diff
@@ -128,12 +144,23 @@ public void init() throws IOException {
             Thread.currentThread().setContextClassLoader(currentClassLoader);
         }
 
+        // Start processing records that have been queued for sending
+        if(kafkaSendQueueCapacity <= 0) {
+            kafkaSendQueueCapacity = 1000;
+            Log.info("Defaulted the 'kafkaSendQueueCapacity' to 1000 since no property was set");
```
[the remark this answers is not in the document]
mattixtech: oops!

PR #2188, opened 2018-10-15
HZN-1422: Remove NCS related code

mvrueden:
Here we remove NCS related code. If the PR gets enough 👍 s I will create an issue for it.


j-white:
There are also a series of ncs related tables in the database - should we drop those too?

mvrueden:
Yep, I guess so

PR #2189, opened 2018-10-15
Proposal: Directly land on events/alarms/outages page

mvrueden:
This is a POC where I removed all of the landing pages and the user will get directly to the events/alarms/outages page. Please let me know what you think.


indigo423:
Absolutely please also remove the unnecessary "Change Password" link when you click on "Change Password" :D

mvrueden:
I think we are all on the same page, closing for now, as there is more work to do in order to get a smooth user experience.

PR #2191, opened 2018-10-16
HZN-1394: Added situation UI elements

christianpape:
* JIRA: https://issues.opennms.org/browse/HZN-1394


j-white:
Looking good.

I had a hard time trying to figure out how to filter the alarm list for situations - clicking on the checkmark for an existing situation in the list seems to do the trick. It may also be worth adding a option for filtering situations in the 'Advanced Search' window.

christianpape:
Yes, you're right. I'll add it to "Advanced Search".

PR #2192, opened 2018-10-16
HZN-1270: Support multiple protocols with a single UDP port for supported protocols

mvrueden:
* JIRA: https://issues.opennms.org/browse/HZN-1270


mvrueden:
After some discussion with @j-white we decided to push this off to `24.0.0` instead

j-white:
It's green! Any changes pending here?

mvrueden:
Yes, there are a bunch of changes pending.

pbrane:
Great effort here.  This is fantastic!

opennms-base-assembly/src/main/filtered/etc/resource-types.d/jmx-resource.xml
```diff
@@ -16,4 +17,35 @@
       </storageStrategy>
     </resourceType>
 
+    <!-- Netflow 5 -->
+    <resourceType name="netflow5" label="Netflow 5" resourceLabel="${index}">
```
j-white: We shouldn't need to define a different resource type for every protocol - we should be able to model this using a single resource type using the index to represent the different types.

mvrueden: Done

mvrueden:
From my and @fooker points of view (hmm is it points of view or point of views?) this is now ready. Removing `triage` label. Please feel free to merge.

pbrane:
Change fooker to be possessive and point would only be plural if you had more than on point of view. 

Fooker’s and my point of view is that the code is ready.

Fooker’s and my points of view are that the code is ready and the code is awesome.

mhuot:
David is correct and I’m really happy to see this feature!

Mike

j-white:
Went to try this out, I built from source, enabled the Netflow 5 listener on OpenNMS, sent a daemon relaod event to telemetryd, and it failed to start with an NPE.

I think we should also update the default `telemetryd-configuration.xml` we ship to use a single port for all of the flow protocols instead of continuing to use separate entries like we used to.

fooker:
@j-white I've fixed a problem, where there are no adapters enabled in a queue but this queue is referenced by a listener. Can you try again.

I will add an example with the multi-port listener to the config next.

j-white:
This is working well for me now. I think we're good to merge here - we can work on updating the default configuration in a follow up issue.

mvrueden:
PR approved. I am going to merge. @fooker please create a new issue for the config issue as @j-white suggested.

mvrueden:
# 🎉 🎈

fooker:
Example added.

PR #2198, opened 2018-10-20
NMS-10418: Create views for SQL based reporting

indigo423:
Created four views to simplify SQL based reporting which also denormalise the data:

* node_categories: Nodes with categories
* node_alarms: Alarm status from nodes and allow filtering on categories
* node_outages: Outages of nodes and allow filtering on categories
* node_ip_services: Denormalise Nodes and IP services

JIRA: http://issues.opennms.org/browse/NMS-10418

### DIFF
--- a/core/schema/src/main/liquibase/24.0.0/changelog.xml
+++ b/core/schema/src/main/liquibase/24.0.0/changelog.xml
@@ -24,5 +24,220 @@
     <dropTable tableName="ncscomponent" />
   </changeSet>
 
+  <changeSet author="ronny" id="24.0.0-node-categories-nms-10418">
+    <preConditions onFail="MARK_RAN">
+      <not>
+        <viewExists viewName="node_categories"/>
+      </not>
+    </preConditions>
+    <sql>
+      CREATE VIEW node_categories AS (
+        SELECT
+          n.*,
+          COALESCE(s_cat.categoryname, 'no category') AS categoryname
+        FROM
+          node n
+        LEFT JOIN
+          category_node cn
+        ON
+          n.nodeid = cn.nodeid
+        LEFT JOIN
+          categories s_cat
+        ON
+          cn.categoryid = s_cat.categoryid
+      );
+    </sql>
+  </changeSet>
+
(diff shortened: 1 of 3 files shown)


indigo423:
Added to create.sql thanks 👍

syepes:
+1

pbrane:
Have you tested these with the query planner when further filtering the views?  I attempted to create similar views in the past with PostgreSQL and had performance issues when adding where clauses to restrict the results; unless the filter condition was targeted on the joining condition.

j-white:
Having to duplicate these views in `create.sql` reminds how we really need to update the tests to be driven from the Liquibase change-log.

Can we add some test coverage for these to make sure that they remain functional and not forgotten about if/when we make schema changes?

indigo423:
@pbrane running the a

```sql
SELECT 
  *
FROM
  node_outages
WHERE
  node_outages.nodelabel like '%';
```
has the same cost using EXPLAIN ANALYZE as running it natively using the SELECT statement instead of the node_outages view. If you I don't understand the problem can you please give me an example what I should test?

indigo423:
@j-white do you have any hints for me how to add these tests? Had a look in NMS-9328 where @mvrueden added some views for the donut charts, but haven't found a test which ensures the views are populated.

j-white:
@indigo423 in the case of NMS-9328, we have a service that leverages the views, and there are tests written for that service i.e. https://github.com/OpenNMS/opennms/blob/52044858adce2ad4db16df3e0ec695c6a73cf983/features/status/api/src/test/java/org/opennms/features/status/api/node/strategy/DefaultNodeStatusCalculatorIT.java

Given that nothing in the code uses the views yet, we could resort to using a similar setup - a database setup with the @JUnitTemporaryDatabase and the DatabasePopulator to add some initial data. Results from the views could then be validated by executing SQL statements against them.

If you need something more concrete, we could help you get the outline setup for the test.

indigo423:
@j-white, @mvrueden added a super naive approach to test if the views exist. Please let me know what you guys think.

j-white:
LGTM, this is sufficient to verify that the views are present and don't throw any errors when invoked.

PR #2199, opened 2018-10-23
NMS-10420: Fix data type and attribute name for Cassandra Thread Pool metrics

indigo423:
* JIRA: http://issues.opennms.org/browse/NMS-10420


mvrueden:
Was this broken from the beginning or did cassandra change there metrics?

indigo423:
I haven't followed the Cassandra release cycles to figure out if this changed in some point in time from their major version 2 to 3. I don't know. This one of the reasons to maintain these types of external configs outside the opennms code base and decouple releases.

PR #2206, opened 2018-10-24
HZN-1387: Fix  SASL Authentication for Kafka  for sentinel 

cgorantla:
* JIRA: http://issues.opennms.org/browse/HZN-1387


cgorantla:
Verified that bundle loads properly on sentinel with SASL configured kafka server

agalue:
I'll test it, but I believe it worth to rebase it to `release-23.0.0`, like the original solution. What do you think ?

cgorantla:
Thanks @agalue for verifying

PR #2207, opened 2018-10-25
Proposal: Remove RemotePollerMap

mvrueden:
Here we rip out the RemotePollerMap.


j-white:
Oops. I didn't realize there wasn't a JIRA already created for this.

I created https://issues.opennms.org/browse/NMS-10440 after it was merged.

PR #2210, opened 2018-10-26
HZN-1429: Refactor the alarm test driver so that it can be used by other modules

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1429

Here we move the alarm test driver into a new module and add some new facilities to help make it usable by other modules.


core/test-api/alarms/src/main/java/org/opennms/core/test/alarms/driver/AcknowledgeAlarmAction.java
```diff
@@ -36,6 +36,7 @@
     private final Date ackTime;
     private final String reductionKey;
 
+
```
mattixtech: Is this stray line intentional?

PR #2212, opened 2018-10-29
NMS-10424: exclude flapping tests by default

RangerRick:
This PR adds a few small changes to be able to tag tests as "flapping" (by adding `@Category(FlappingTests.class)` to the test class or individual tests).

The patch looks bigger than you would think it needs to be because anywhere that the `maven-surefire-plugin` or `maven-failsafe-plugin` runs needs to be updated to be able to find the `FlappingTests` class (it's an interface).

There are Bamboo changes that go along with this, but they are backwards compatible (branches that do not have "flapping" support will just run as they already had.)

* JIRA: http://issues.opennms.org/browse/NMS-10424


pom.xml
```diff
@@ -3765,9 +3777,9 @@
         <scope>test</scope>
       </dependency>
       <dependency>
-        <groupId>junit</groupId>
-        <artifactId>junit</artifactId>
-        <version>4.12</version>
+        <groupId>org.opennms.core.test-api</groupId>
+        <artifactId>org.opennms.core.test-api.junit</artifactId>
+        <version>4.12</version> <!-- if you change this, be sure to update core/test-api/junit/pom.xml as well! -->
```
mvrueden: Can't we just define it as a property, e.g. `${junitVersion}` so we don't have to worry about the comment?

RangerRick: Because of how the reactor has to set things up to get the correct order on a clean build, I can't make the top-level pom a parent of `core/test-api/junit` so it can't inherit properties.

RangerRick:
Since there are no objections, I'm gonna go ahead and merge this.  It will take some hand-fixing going forward to other branches so I'd like to do that now.

PR #2215, opened 2018-10-31
HZN-1404: Modify ticketing api for Situations

cgorantla:
* JIRA: http://issues.opennms.org/browse/HZN-1404

### DIFF
--- a/features/ticketing/api/src/main/java/org/opennms/api/integration/ticketing/RelatedAlarmSummary.java
+++ b/features/ticketing/api/src/main/java/org/opennms/api/integration/ticketing/RelatedAlarmSummary.java
@@ -1,0 +1,96 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2018 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2018 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License,
+ * or (at your option) any later version.
+ *
+ * OpenNMS(R) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with OpenNMS(R).  If not, see:
+ *      http://www.gnu.org/licenses/
+ *
+ * For more information contact:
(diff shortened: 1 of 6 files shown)


features/ticketing/api/src/main/java/org/opennms/api/integration/ticketing/RelatedAlarmSummary.java
```diff
@@ -0,0 +1,85 @@
...
+package org.opennms.api.integration.ticketing;
+
+import java.net.InetAddress;
+
+/**
+ * Encapsulates alarm details that are associated with a situation
+ * (primary alarm associated with ticket)
+ **/
+
+public class RelatedAlarmSummary {
```
mattixtech: This might just me being picky...

This class looks like a bean with getters and setters but the usage of it seems to treat it as immutable. I think it would be more convenient if it was made immutable by making the fields final, removing the setters, and then creating a builder class that is the only way to instantiate it (private constructor). That is the preferable pattern to use in place of a new instance creation followed by a set() for several fields.

You can then change your number fields to ints and add error checking in the builder to check for those fields not being set (along with whatever other fields are mandatory):

```
private RelatedAlarmSummary(int alarmId, int nodeId, InetAddress ipAddress, String summary,
                                String details) {
        this.alarmId = alarmId;
        this.nodeId = nodeId;
        this.ipAddress = ipAddress;
        this.summary = summary;
        this.details = details;
    }

    public static class Builder {
        private Integer alarmId;
        private Integer nodeId;
        private InetAddress ipAddress;
        private String summary;
        private String details;

        public Builder() {
        }

        public RelatedAlarmSummary build() {
            return new RelatedAlarmSummary(Objects.requireNonNull(alarmId), Objects.requireNonNull(nodeId), ipAddress,
                    summary, details);
        }

        public Builder withAlarmId(int alarmId) {
            this.alarmId = alarmId;
            return this;
        }

        public Builder withNodeId(int nodeId) {
            this.nodeId = nodeId;
            return this;
        }

        public Builder withIpAddress(InetAddress ipAddress) {
            this.ipAddress = ipAddress;
            return this;
        }

        public Builder withSummary(String summary) {
            this.summary = summary;
            return this;
        }

        public Builder withDetails(String details) {
            this.details = details;
            return this;
        }
    }
```

Then you can refactor instance creation to:
```
return new RelatedAlarmSummary.Builder()
                    .withAlarmId(relatedAlarm.getId());
                    .withIpAddress(relatedAlarm.getIpAddr());
                    .withNodeId(relatedAlarm.getNodeId());
                    .withSummary(relatedAlarm.getDescription());
                    .withDetails(relatedAlarm.getLogMsg());
                    .build();
```

j-white: This differs from the patterns that are already used in the Ticketing API, so it seems out of place to just add this for the RelatedAlarmSummary objects.

mattixtech: Fair enough. For new classes I think it is nicer to pattern them like this when possible but if it introduces a consistency issue due to existing API then it is probably not worth it.

PR #2216, opened 2018-10-31
HZN-1431: Update OpennmsKafkaProducer to handle alarm feedback

mattixtech:
This PR causes all situation feedback received via the rest API to be sent to a new alarm feedback topic in Kafka.

I'm assuming we don't need any mechanism to sync alarm feedback since we aren't persisting to the Opennms DB.

* JIRA: http://issues.opennms.org/browse/HZN-1431

### DIFF
--- a/features/kafka/producer/src/main/proto/opennms-kafka-producer.proto
+++ b/features/kafka/producer/src/main/proto/opennms-kafka-producer.proto
@@ -73,6 +73,22 @@ message Alarm {
   repeated Alarm relatedAlarm = 22;
 }
 
+message AlarmFeedback {
+  string situationKey = 1;
+  string situationFingerprint = 2;
+  string alarmKey = 3;
+  enum FeedbackType {
+    FALSE_POSITIVE = 0;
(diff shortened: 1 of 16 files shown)


j-white:
Looks great. I merged `develop` back into this branch to see if Bamboo will pass now.

mattixtech:
Looks like there is an issue with the karaf feature now since my changes which is likely why the tests failed. Looking into it...

features/kafka/producer/src/main/proto/opennms-kafka-producer.proto
```diff
@@ -73,6 +73,22 @@ message Alarm {
   repeated Alarm relatedAlarm = 22;
 }
 
+message AlarmFeedback {
```
cgorantla: I think we were using underscore in naming fields in `.proto` file instead of camel case.

mattixtech: Yep, you are right. I'll update that proto and regenerate the model file.

PR #2217, opened 2018-10-31
NMS-10409: Default text "Add Outage" in scheduled outage editor should be a placeholder

jeffgdotorg:
* JIRA: http://issues.opennms.org/browse/NMS-10409


mattixtech:
👍

mvrueden:
Tests failed, but the job is already deleted in bamboo. Put it back in and re-started a build.

PR #2223, opened 2018-11-01
NMS-10398 speed up loading of CDP topology

patrick-schweizer:
introduce a cdp link and node cache with lightweight objects without relations, avoid unnecessary layout computations.

* JIRA: https://issues.opennms.org/browse/NMS-10398

### DIFF
--- a/features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/TopologyUI.java
+++ b/features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/TopologyUI.java
@@ -221,7 +221,7 @@ public class TopologyUI extends UI implements MenuUpdateListener, ContextMenuHan
             return false;
         }
 
-        public void handleRequestParameter(VaadinRequest request) {
+        private void handleRequestParameter(VaadinRequest request) {
             boolean updateURL = false;
             for (RequestParameterHandler handler : requestHandlerList) {
                 if (handler.handleRequest(request)) {
@@ -873,7 +873,6 @@ public class TopologyUI extends UI implements MenuUpdateListener, ContextMenuHan
     // may change during layout.
     private void loadUserSettings() {
         applyHistory(m_applicationContext.getUsername(), m_historyManager.getHistoryFragment(m_applicationContext.getUsername()));
-        m_graphContainer.redoLayout();
     }
 
     private void applyHistory(String username, String fragment) {
--- a/features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/jung/D3TopoLayout.java
+++ b/features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/jung/D3TopoLayout.java
@@ -214,6 +214,10 @@ public class D3TopoLayout<V, E> extends AbstractLayout<V, E> implements Iterativ
     @Override
     public void step() {
 
+        if(getGraph().getEdges().isEmpty()){
+            m_alpha = 0; // => makes done() == true. We don't need to iterate over nothing
+        }
+
         double currentForce;
 
         //guass-seidel relaxation for links
@@ -341,13 +345,9 @@ public class D3TopoLayout<V, E> extends AbstractLayout<V, E> implements Iterativ
         private int m_charge = DEFAULT_CHARGE;
         private Point2D m_previous = null;
 
-        protected void offset(double x, double y)
-        {
-            String before = this.toString();
+        protected void offset(double x, double y) {
             this.x += x;
             this.y += y;
-            String after = this.toString();
-            print(before, after);
         }
         
         protected void offsetPrevious(double x, double y) {
@@ -365,23 +365,7 @@ public class D3TopoLayout<V, E> extends AbstractLayout<V, E> implements Iterativ
             m_previous = new Point2D.Double(x, y);
         }
 
-        @Override
-        public void setLocation(double x, double y) {
-            String before = this.toString();
-            super.setLocation(x, y);
-            String after = this.toString();
-            print(before, after);
-        }
-
         private void print(String before, String after) {
-        }
-
-        @Override
-        public void setLocation(Point2D p) {
-            String before = this.toString();
-            super.setLocation(p);
-            String after = this.toString();
-            print(before, after);
         }
 
         protected double norm()
--- a/features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/jung/D3TopoLayout.java
+++ b/features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/jung/D3TopoLayout.java
@@ -110,7 +110,7 @@ public class D3TopoLayout<V, E> extends AbstractLayout<V, E> implements Iterativ
     @Override
     public void step() {
 
-        if(getGraph().getEdges().isEmpty()){
+        if(getGraph().getEdges().isEmpty() && getGraph().getVertices().isEmpty()){
             m_alpha = 0; // => makes done() == true. We don't need to iterate over nothing
         }
 
--- a/features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/jung/D3TopoLayout.java
+++ b/features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/jung/D3TopoLayout.java
@@ -107,110 +107,6 @@ public class D3TopoLayout<V, E> extends AbstractLayout<V, E> implements Iterativ
 
     }
 
-    public void stepOld() {
-
-        double currentForce;
-
-        //guass-seidel relaxation for links
-        for (E e : getGraph().getEdges()) {
-            Pair<V> endPoints = getGraph().getEndpoints(e);
-            VertexData srcVertexData = getVertexData(endPoints.getFirst());
-            VertexData targetVertexData = getVertexData(endPoints.getSecond());
-
-            double xDelta = targetVertexData.getX() - srcVertexData.getX();
-            double yDelta = targetVertexData.getY() - srcVertexData.getY();
-            double l = xDelta * xDelta + yDelta * yDelta;
-            if (l != 0) {
-                EdgeData edgeData = getEdgeData(e);
-                double lSqrt = Math.sqrt(l);
-                double distance = m_alpha * edgeData.getStrength() * (lSqrt - edgeData.getDistance()) / lSqrt;
-
-                xDelta *= distance;
-                yDelta *= distance;
-
-                currentForce = (double)srcVertexData.getWeight() / (double)(targetVertexData.getWeight() + srcVertexData.getWeight());
-                targetVertexData.offset(-(xDelta * currentForce), -(yDelta * currentForce));
-
-                currentForce = 1 - currentForce;
-                srcVertexData.offset(xDelta * currentForce, yDelta * currentForce);
-
-            }
-
-        }
-
-        //Apply gravity forces
-        currentForce = m_alpha * getGravity();
-        if(currentForce != 0){
-            double centerX = getSize().getWidth() / 2;
-            double centerY = getSize().getHeight() / 2;
-
-            for (V v : getGraph().getVertices()) {
-                VertexData vData = getVertexData(v);
-                vData.offset((centerX - vData.getX()) * currentForce, (centerY - vData.getY()) * currentForce);
-            }
-            
-        }
-
-        //Compute quad tree center of mass and apply charge force
-        if(getDefaultCharge() != 0){
-            
-            DblBoundingBox bounds = new DblBoundingBox(0,0,getSize().getWidth(), getSize().getHeight());
-            QuadTree<VertexData> quadTree = new QuadTree<VertexData>(bounds);
-            for(V v : getGraph().getVertices()) {
-                VertexData vData = getVertexData(v);
-                quadTree.insert(vData, vData.getCharge(), vData);
-            }
-
-            for(V v: getGraph().getVertices()) {
-                final VertexData vData = getVertexData(v);
-                quadTree.visit(new Visitor<VertexData>() {
-
-                    @Override
-                    public boolean visitNode(Node<VertexData> n) {
-                        
-                        if (n.isLeaf() && vData == n.getValue()) return true;
-                        
-                        double dx = n.getX() - vData.getX();
-                        double dy = n.getY() - vData.getY();
-                        double dw = n.getWidth();
-                        double dSquared = dx*dx + dy*dy;
-
-                        if (dw*dw/m_thetaSquared < dSquared) {
-                            double force = n.getCharge() / dSquared;
-                            vData.offset(- (dx*force), - (dy*force));
-                            return true;
-                        }
-                        
-                        if (n.isLeaf()) {
-                            if (dSquared == 0) {
-                                vData.offset(1, 1);
-                            } else {
-                                double force = n.getCharge() / dSquared;
-                                vData.offset(- (dx*force), - (dy*force));
-                            }
-                            return true;
-                        }
-                        
-                        return false;
-                        
-                    }
-                    
-                });
-            }
-        }
-        
-        
-        for(V v : getGraph().getVertices()) {
-            VertexData vData = getVertexData(v);
-            Point2D location = transform(v);
-            location.setLocation(vData.getX(), vData.getY());
-        }
-        
-        m_alpha *= 0.998235;
-
-
-    }
-
     @Override
     public void step() {
 
@@ -233,13 +129,11 @@ public class D3TopoLayout<V, E> extends AbstractLayout<V, E> implements Iterativ
                 EdgeData edgeData = getEdgeData(e);
                 double lSqrt = Math.sqrt(l);
                 double distance = m_alpha * edgeData.getStrength() * (lSqrt - edgeData.getDistance()) / lSqrt;
-                //double distance = edgeData.getStrength() * (lSqrt - edgeData.getDistance()) / lSqrt;
 
                 xDelta *= distance;
                 yDelta *= distance;
 
                 currentForce = (double)srcVertexData.getWeight() / (double)(targetVertexData.getWeight() + srcVertexData.getWeight());
-                //currentForce = 0.5;
                 targetVertexData.offset(-(xDelta * currentForce), -(yDelta * currentForce));
 
                 currentForce = 1 - currentForce;
@@ -277,11 +171,9 @@ public class D3TopoLayout<V, E> extends AbstractLayout<V, E> implements Iterativ
                         double k = m_alpha * vData2.getCharge() / d;
                         double px = dx*k;
                         double py = dy*k;
-                        
-                        //vData1.offsetPrevious(px, py);
+
                         vData1.offset(px, py);
                     } else {
-                        //vData1.offsetPrevious(0.5-Math.random(), 0.5-Math.random());
                         vData1.offset(0.5-Math.random(), 0.5-Math.random());
                     }
                     
--- a/features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/jung/D3TopoLayout.java
+++ b/features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/jung/D3TopoLayout.java
@@ -110,10 +110,6 @@ public class D3TopoLayout<V, E> extends AbstractLayout<V, E> implements Iterativ
     @Override
     public void step() {
 
-        if(getGraph().getEdges().isEmpty() && getGraph().getVertices().isEmpty()){
-            m_alpha = 0; // => makes done() == true. We don't need to iterate over nothing
-        }
-
         double currentForce;
 
         //guass-seidel relaxation for links
@@ -210,7 +206,10 @@ public class D3TopoLayout<V, E> extends AbstractLayout<V, E> implements Iterativ
 
     @Override
     public boolean done() {
-        return m_alpha < 0.005;
+        return
+                // if we have no objects in our graph to layout we are done:
+                (getGraph().getEdges().isEmpty() && getGraph().getVertices().isEmpty())
+             || m_alpha < 0.005;
     }
 
     private VertexData getVertexData(V v) {
--- a/features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/service/DefaultGraph.java
+++ b/features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/service/DefaultGraph.java
@@ -32,6 +32,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.TreeSet;
 
@@ -162,5 +163,23 @@ public class DefaultGraph implements Graph {
             }
         }
         LOG.debug("Created a graph with {} vertices and {} edges", m_displayVertices.size(), m_displayEdges.size());
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        DefaultGraph that = (DefaultGraph) o;
+        return Objects.equals(m_displayVertices, that.m_displayVertices) &&
+                Objects.equals(m_displayEdges, that.m_displayEdges) &&
+                Objects.equals(m_layout, that.m_layout) &&
+                Objects.equals(edgeStatus, that.edgeStatus) &&
+                Objects.equals(vertexStatus, that.vertexStatus);
+    }
+
+    @Override
+    public int hashCode() {
+
+        return Objects.hash(m_displayVertices, m_displayEdges, m_layout, edgeStatus, vertexStatus);
     }
 }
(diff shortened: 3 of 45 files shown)


j-white:
The caching layer we added seems very specific to CDP - is it possible to make it more generic such that it can be reused by the other protocols as well? I'd prefer to avoid having to creating new `*TopologyInfoCache` implementations for each protocol.

patrick-schweizer:
@j-white the CdpTopologyInfoCache caches 2 things: Nodes and CdpLinks, potentially we want to add CdpElement as well. The node cache could be reused for all other protocols. I am not sure if we can generalize the *Links - the linking mechanism seems to work differently for different protocols.
In any case we will have to write code for every protocol that accesses the database and maps the data to a *LinkInfo.

rssntn67:
It seems that this changes are very similar to that I have done in features/topologies under the branches jira/NMS-10375 and jira/NMS-9273 where I have moved the topology api stuff under new features projects.

The idea is that if we want topology object you should ask to topologies services.
The topologies services are designed to hold share datat from different part of opennms. 

For example the enlinkd project will host data for topologies CDP, LLDP, ISIS, OSPF, Bridge.
So the code for calculating topology must be moved from topologyprovider to TopologyUpdaters....

I'm just now working on protocol topology updaters where using an in memory strategy.
Topology is updated as it changes. 

I guess we should coordinate our developments we are not so far doing the same think.

patrick-schweizer:
@rssntn67 yes, lets discuss. I'll ping you on Mattermost

patrick-schweizer:
> @j-white the CdpTopologyInfoCache caches 2 things: Nodes and CdpLinks, potentially we want to add CdpElement as well. The node cache could be reused for all other protocols. I am not sure if we can generalize the *Links - the linking mechanism seems to work differently for different protocols.
> In any case we will have to write code for every protocol that accesses the database and maps the data to a *LinkInfo.

As discussed i renamed it to TopologyInfo... so that we have a more general way to cache Topology relevant stuff.

mvrueden:
We should find another name. `TopologyInfo` is already existing and reserved for information related to the Topology and not an element.

patrick-schweizer:
@mvrueden what do you think about: TopologyEntityCache? Or other ideas: TopologyPartsCache, TopologyComponentsCache

mvrueden:
> @mvrueden what do you think about: TopologyEntityCache? Or other ideas: TopologyPartsCache, TopologyComponentsCache

Yeah, much better. I like `TopologyEntityCache`

j-white:
@patrick-schweizer you may need to merge `develop` back into this branch for it to Bamboo to pass

opennms-webapp-rest/src/test/resources/v1/acks.json
```diff
@@ -7,7 +7,7 @@
             "ackTime": 1437073152156, 
             "ackType": "UNSPECIFIED", 
             "ackUser": "admin", 
-            "id": 61, 
+            "id": 62,
```
j-white: Hmm?

opennms-model/src/main/java/org/opennms/netmgt/model/VertexInfo.java
```diff
@@ -0,0 +1,111 @@
...
+ *******************************************************************************/
+
+package org.opennms.netmgt.model;
+
+import java.util.Objects;
+
+import org.opennms.netmgt.model.monitoringLocations.OnmsMonitoringLocation;
+
+/** Simplified OnmsNode. We use it as a view on OnmsNode that has less attributes and no lazy loading */
+public class VertexInfo {
```
j-white: `VertexInfo`  seems pretty abstract since a vertex in the context of the topology can be just about anything.

Maybe `NodeVertexInfo` would be more appropriate here?

opennms-dao/src/main/resources/META-INF/opennms/applicationContext-dao.xml
```diff
@@ -323,6 +323,24 @@
...
+  <onmsgi:service interface="org.opennms.netmgt.dao.api.TopologyEntityDao" ref="topologyEntityDao" />
+
+  <bean id="topologyEntityCache" class="org.opennms.netmgt.dao.hibernate.TopologyEntityCacheImpl">
+    <property name="topologyEntityDao" ref="topologyEntityDao" />
+  </bean>
+
+  <onmsgi:service interface="org.opennms.netmgt.dao.api.TopologyEntityCache" ref="topologyEntityCache" />
+
+  <task:scheduled-tasks scheduler="topologyRefresherScheduler">
+    <task:scheduled ref="topologyEntityCache" method="refresh" fixed-delay="300000"/>
```
j-white: This now means that we'll be loading *all* of the nodes and all of the CDP links from the database every 5 minutes, even if no one is using the topology app.

Perhaps we should change the strategy to using a cache that triggers a load/refresh when the data is accessed instead.

mvrueden: We could even have a "Loading UI Screen" in this case.
So the first time anyuser navigates to the topology map, we show a "Loading or Initializing screen". If the cache is initialized (either via event or cache.isInitiazlized()) we hide the loading screen.

There is a `TopologyService` which may be used for requesting a refresh of the underlying cache(s).

opennms-webapp-rest/src/test/resources/v1/acks.json
patrick-schweizer: this is a hard coded database id (I know - not very nice). It gets generated by https://github.com/OpenNMS/opennms/blob/foundation-2018/opennms-dao/src/main/java/org/opennms/netmgt/dao/DatabasePopulator.java . Since I added the creation of a CdpLink the sequence is one higher afterwards...

opennms-model/src/main/java/org/opennms/netmgt/model/VertexInfo.java
patrick-schweizer: @j-white, agreed. I think we should have a common name for these objects (currently: VertexInfo, CdpLinkInfo, later all [Lldp|Isis|...]LinkInfo). They are lightweight entities that contain no references and have only the necessary  attributes to construct the topology. How do you like: NodeTopologyEntity, CdpTopologyEntity, IsisTopologyEntity,...

That would be consistent with @mvrueden remark (calling the cache: TopologyEntityCache)

opennms-dao/src/main/resources/META-INF/opennms/applicationContext-dao.xml
patrick-schweizer: The task was to make the initial loading of the topology map faster. We did do this mainly by 2 means:

1. load lightweight entities with no references and only the necessary attributes instead of the full entity (like OnmsNode) with lazy loaded references
2. preload the lightweight entities and caching them. The cache is a global one, not bound to a user.

Basically you are saying we shouldn't do the preloading in 2. anymore but switch to a load on demand strategy. Therefor we will loose that time benefit. On my computer this is for 40,000 Nodes and 80,000 CdpLinks: ~1-2,5 seconds. This delay we would add each time the cache is refreshed - not only for the initial loading.

We have in general two ways to fill a cache:
1. synchronously on demand / lazy loading -> slower but no unnecessary loading.
1. Asynchronous / preloading -> fast at access time but potentially unnecessary loading.

In order to determine what is the more appropriate strategy for our use case, I think we should look at the costs and the benefits and put that into the context of a typical OpenNMS usage: how many users do work in parallel, how often are the topology maps accessed, is CPU and RAM a concern, is sparse usage of resources more important than usability / speed?

A very rough indication of how much data we hold in memory: I serialized the cache (40,000 Nodes and 80,000 CdpLinks) which amounted to ~12MB. I don't know what impact it has on the network traffic and the database itself.

=> we are talking about a tradeoff between 1-2 seconds of delay for every cache access that needs a refresh vs. 12 MB RAM + database impact.

@j-white, @mvrueden what is your take on these numbers? Shall we do loading on demand or preloading (as implemented now)?

mvrueden: > @j-white, @mvrueden what is your take on these numbers? Shall we do loading on demand or preloading (as implemented now)?

At the moment I would opt for filling the cache on demand by default.
That way we don't add unnecessary load to the system if the topology is not used at all.

Besides that we could add a system property which enables a periodical reload of the data.
And also make the refresh-interval configurable. Important here is, that the refresh is only triggered n-seconds AFTER the first refresh was completed. Not every n seconds.

j-white: While preloading does give us the best performance, it comes at the cost of loading unnecessarily for many of the installs that do not use the topology app.

I agree with the approach that @mvrueden proposes above.

features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/TopologyUI.java
```diff
@@ -873,7 +873,6 @@ public LayoutManager getLayoutManager() {
     // may change during layout.
     private void loadUserSettings() {
         applyHistory(m_applicationContext.getUsername(), m_historyManager.getHistoryFragment(m_applicationContext.getUsername()));
-        m_graphContainer.redoLayout();
```
mvrueden: If removing this, please ensure, that the topology is still loaded correctly. To do so, go to the Topology Map and get some state. Go to the right and click "Share View" (or similar). Copy the URL.
Now log out with the current user. Best option is to open a new private window in a seperate browser and paste in the URL. The same state should be visible. If that still works, this line can be removed.

features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/service/DefaultGraph.java
```diff
@@ -163,4 +164,22 @@ private void updateLayout(Collection<Vertex> displayVertices, Collection<Edge> d
...
+                Objects.equals(m_displayEdges, that.m_displayEdges) &&
+                Objects.equals(m_layout, that.m_layout) &&
+                Objects.equals(edgeStatus, that.edgeStatus) &&
+                Objects.equals(vertexStatus, that.vertexStatus);
+    }
+
+    @Override
+    public int hashCode() {
+
+        return Objects.hash(m_displayVertices, m_displayEdges, m_layout, edgeStatus, vertexStatus);
```
mvrueden: I am not sure if this will work as expected.

In my opinion we should not include `edgeStatus` and `vertexStatus` for `hashCode` or `equals`.
The layout should be "recalculated" if the displaying vertices or edges have changed.
At the moment if the vertices and edges have not changed, but a status (e.g. one vertex is now red instead of green), the layout would be recalculated, which is not ideal, as the positions of the vertices/edges can basically stay the same, only the UI needs to know about the new state.

features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/jung/D3TopoLayout.java
```diff
@@ -214,6 +214,10 @@ public boolean visitNode(Node<VertexData> n) {
     @Override
     public void step() {
 
+        if(getGraph().getEdges().isEmpty()){
+            m_alpha = 0; // => makes done() == true. We don't need to iterate over nothing
```
mvrueden: We are still iterating at least once. Can't we just do this while initializing the layout and not invoke `step()` at all?
Can we also get rid of the `stepOld()` method if we are here anyways changing code?

opennms-model/src/main/java/org/opennms/netmgt/model/VertexInfo.java
patrick-schweizer: I renamed them to NodeTopologyEntity and CdpLinkTopologyEntity

opennms-dao/src/main/resources/META-INF/opennms/applicationContext-dao.xml
patrick-schweizer: ok, I changed the code accordingly.

features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/TopologyUI.java
patrick-schweizer: I went through your test case and it seems that this still works.
The call hierarchy is the following:
TopologyUI.init() -> loadUserSettings() this is the only time that this private method is called. Later in TopologyUI.init() we call m_graphContainer.redoLayout(); directly and thus the call m_graphContainer.redoLayout() in loadUserSettings() should be unnecessary?

features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/service/DefaultGraph.java
patrick-schweizer: ok, I changed the hashing: https://github.com/OpenNMS/opennms/blob/173a37e706df019d70e68640217f81820efa8044/features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/VEProviderGraphContainer.java#L255

features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/jung/D3TopoLayout.java
patrick-schweizer: `stepOld()`=> removed
I am not sure I understand the rest of your suggestion. Where do you see where we can save time?

features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/TopologyUI.java
mvrueden: Yep, makes sense.

features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/service/DefaultTopologyService.java
```diff
@@ -222,7 +225,14 @@ public void setServiceLocator(ServiceLocator serviceLocator) {
         this.serviceLocator = Objects.requireNonNull(serviceLocator);
     }
 
+    public void setTopologyEntityCache(TopologyEntityCache topologyEntityCache) {
+        this.topologyEntityCache = Objects.requireNonNull(topologyEntityCache);
+    }
+
     public void invalidate(String namespace) {
+        // the topologyEntityCache is namespace agnostic:
+        topologyEntityCache.refresh();
```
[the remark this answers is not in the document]
mvrueden: Data is cached for a 30 seconds by default. This means any changes behind the curtains take also 30 seconds to be visualised. In order to not wait at least 30 + delta seconds I opted in for allowing to invalidating the cache manually, forcing the data to be reloaded when queried afterwards. 

From my point of view this is the correct place. 
But keep in mind, that the cache is only invalidated if an event is send manually to OpenNMS (or the 30 seconds pass of course).

Besides that the`topologyEntityCache` is only used in the `nodes` namespace. 
So we probably should refresh it only if the `namespace.startsWith("nodes")`.

features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/service/DefaultGraph.java
mvrueden: Okay, that is much better. Can you please add a comment, why the hash is build this way and please remove the `equals()` and `hashCode()` methods from the `DefaultGraph` if they are not needed?

mvrueden:
Looks good to me. I like the naming `NodeTopologyEntity/CdpLinkTopologyEntity`. 

Here is something I'd like to point out, though:
The above classes are used in the `TopologyEntityDao` and thus are clearly marked Hibernate/JPA entities. However they are missing the `@Entity` annotation. 
They also all have an `id` property, which is not marked as `@Id`. 
Besides that they manually implement the `equals()` and `hashCode()` method, but the `id` should be the deciding factor if it is the same object.

This can lead to severe problems, however as this is at the moment not critical, as we only use it as a `READ WRAPPER` to get the data from the database in a convenient model. But we should make it clear somehow, that these are not proper Hibernate/JPA entities, even if they are named like such.

features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/jung/D3TopoLayout.java
patrick-schweizer: Ah now I understand. I solved it a bit differently though. I moved that logic directly into the done method which I believe is the right place for it since that method is responsible for determining if the layout is finished or not. @mvrueden do you agree?

features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/service/DefaultTopologyService.java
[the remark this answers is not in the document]
patrick-schweizer: ok I changed that to look for `namespace.startsWith("nodes")`.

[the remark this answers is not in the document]
mvrueden: 👍

features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/jung/D3TopoLayout.java
mvrueden: Actually I think it is not that important. Both `done()` or `initialize()` are equally good. I don't prefer the one over the other 👍

features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/service/DefaultTopologyService.java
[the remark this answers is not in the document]
mvrueden: 👍

patrick-schweizer:
> Looks good to me. I like the naming `NodeTopologyEntity/CdpLinkTopologyEntity`.
> 
> Here is something I'd like to point out, though:
> The above classes are used in the `TopologyEntityDao` and thus are clearly marked Hibernate/JPA entities. However they are missing the `@Entity` annotation.
> They also all have an `id` property, which is not marked as `@Id`.
> Besides that they manually implement the `equals()` and `hashCode()` method, but the `id` should be the deciding factor if it is the same object.
> 
> This can lead to severe problems, however as this is at the moment not critical, as we only use it as a `READ WRAPPER` to get the data from the database in a convenient model. But we should make it clear somehow, that these are not proper Hibernate/JPA entities, even if they are named like such.

@mvrueden I tried to address your comment by changing the methods equals() and hashCode and improving the class descriptions.

mvrueden:
Only concern with this is, that if it were a proper hibernate entity and the id is null, but a different object, they would be equal, which in that case they are not. 

The only real solution to this is to not implement equals and hashCode, but implement a "BusinessKey" which implements the equals/hashCode methods accordingly. 

If we need the equals/hashCode, I would implement it properly (as it was before), as we are not running into any issues, thus the comment on the entity. If we don't need them, just don't implement them.

patrick-schweizer:
> If we need the equals/hashCode, I would implement it properly (as it was before), as we are not running into any issues, thus the comment on the entity. If we don't need them, just don't implement them.
In general I think it is good practice to have equals / hashCode implemented on data objects but I have no strong opinion on this. I removed equals and hashCode since we don't use them currently.

opennms-dao-api/src/main/java/org/opennms/netmgt/dao/api/TopologyEntityDao.java
```diff
@@ -0,0 +1,39 @@
...
+ *******************************************************************************/
+
+package org.opennms.netmgt.dao.api;
+
+import java.util.List;
+
+import org.opennms.netmgt.model.CdpLinkTopologyEntity;
+import org.opennms.netmgt.model.NodeTopologyEntity;
+
+public interface TopologyEntityDao {
```
j-white: Can we add some documentation here describing the purpose of this interface?

opennms-model/src/main/java/org/opennms/netmgt/model/NodeTopologyEntity.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+
+    public String getLocation() {
+        return location;
+    }
+
+    @Override
+    public String toString() {
+        return "VertexInfo{" +
```
j-white: `VertexInfo` vs `NodeTopologyEntity`.

opennms-dao-api/src/main/java/org/opennms/netmgt/dao/api/TopologyEntityCache.java
```diff
@@ -0,0 +1,44 @@
...
+ *******************************************************************************/
+
+package org.opennms.netmgt.dao.api;
+
+import java.util.List;
+
+import org.opennms.netmgt.model.CdpLinkTopologyEntity;
+import org.opennms.netmgt.model.NodeTopologyEntity;
+
+public interface TopologyEntityCache {
```
j-white: Can we add some documentation here describing the purpose of this interface?

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/TopologyEntityCacheImpl.java
```diff
@@ -0,0 +1,109 @@
...
+
+public class TopologyEntityCacheImpl implements TopologyEntityCache {
+    
+    private final static Logger LOG = LoggerFactory.getLogger(TopologyEntityCacheImpl.class);
+    private final static String KEY = "KEY";
+    private final static String SYSTEM_PROPERTY_CACHE_DURATION = "org.opennms.ui.topology-entity-cache-duration";
+
+    private TopologyEntityDao topologyEntityDao;
+
+    LoadingCache<String, List<NodeTopologyEntity>> nodeTopologyEntities = createCache(
```
j-white: Can we make these `private`?

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/TopologyEntityCacheImpl.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    @Override
+    public void refresh(){
+        nodeTopologyEntities.refresh(KEY);
+        cdpLinkTopologyEntities.refresh(KEY);
+    }
+
+    private int getCacheDuration(){
+        String duration = System.getProperty(SYSTEM_PROPERTY_CACHE_DURATION, "300");
```
j-white: Can use `Integer.getInteger` to retrieve an integer value from the sys prop.

opennms-dao/src/main/resources/META-INF/opennms/applicationContext-dao.xml
```diff
@@ -6,13 +6,13 @@
        xmlns:tx="http://www.springframework.org/schema/tx"
        xmlns:util="http://www.springframework.org/schema/util"
        xmlns:onmsgi="http://xmlns.opennms.org/xsd/spring/onms-osgi"
+       xmlns:task="http://www.springframework.org/schema/task"
```
j-white: We don't need this one anymore.

opennms-dao-api/src/main/java/org/opennms/netmgt/dao/api/TopologyEntityDao.java
patrick-schweizer: no problem. I added a description.

opennms-model/src/main/java/org/opennms/netmgt/model/NodeTopologyEntity.java
patrick-schweizer: ha. this is one of those stupid mistakes that just disappear with lombok :-(
fixed

opennms-dao-api/src/main/java/org/opennms/netmgt/dao/api/TopologyEntityCache.java
patrick-schweizer: done.

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/TopologyEntityCacheImpl.java
patrick-schweizer: hups :-( done.

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/TopologyEntityCacheImpl.java
patrick-schweizer: You are right, but I believe with Integer.getInteger() we can't distinguish between a not set property and an invalid value (not parsable number):
`propname=123 # will result in 123`
`# propname=123 # will result in default value`
`propname=not_a_number # will result in default value`
2 and 3 will have the same outcome (default value) but we won't have a warning in the log. If thats ok then we can simplify the code by just using Integer.getInteger().
Another option would be to generalize this behavior so we don't have to repeat it everywhere, something like this:
```
SystemHelper.getAsInteger(String name, int defaultValue){
    try {
        return Integer.parseInt(name);
    } catch(NumberFormatException e) {
        LOG.warn(...);
    }
}
```
@j-white What do you think? Is there a general way how to deal with invalid properties?

j-white: We already use `Integer.getInteger` in a number of places so I'd would just continue to use it here.

You raise a good point though. We could add a utility function like the one you proposed, migrate the existing calls to `Integer.getInteger` to use it, and add a checkstyle rule to prevent new uses of `Integer.getInteger`. Maybe we can tackle this in a separate issue?

patrick-schweizer: @j-white, makes sense to me. I changed the code to use Integer.getInteger()
about the separate issue: I created a ticket: https://issues.opennms.org/browse/NMS-10477. Not sure what labels I need to set. I can tackle the issue in the next days if you want, it shoudn't take long.

mvrueden:
Nope, all good now.

PR #2224, opened 2018-11-06
NMS-7797: Migrate to Vaadin v8.5.2

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-7797

Here we migrate Vaadin to 8.5.2 in order to make it later more easily compile with Java 9.

With Vaadin 8 there is a new API, which should be used instead. 
However Vaadin provides so called `compatibility` modules which contain the old API classes, which you can detect via the import statements, as they start with `com.vaadin.v7` (e.g. `com.vaadin.v7.ui.TextField`). 
Here I migrated only the vaadin applications to make use of the old `com.vaadin.v7` classes and did not upgrade to the Vaadin 8 API.

By default however, the Compatibility classes are not integrated, so a custom Widgetset must be defined. So `org.opennms.vaadin.DefaultWidgetset` must be defined by all Vaadin Applications which use the v7 compatibility classes. 

Besides that the biggest changes are: 
 * `setReadOnly(...)` is no longer available for non Editable fields
 * `setImmediate(...)` can no longer be set manually and is automatically assumed
 * `UI.getCurrent(...).access(...)` is now working as documented and therefore no longer working for us

@agalue, @gallenc, @christianpape I would like to invite you to verify the Vaadin Applications you originally contributed and let me know if they look and behave correctly in Vaadin 8. Changes for the Plugin Manager can be viewed here: https://github.com/OpenNMS/osgi-plugin-manager/compare/1.1-SNAPSHOT...1.2-SNAPSHOT?expand=1

~~This PR requires the `RemotePollerMap` to be removed. See PR https://github.com/OpenNMS/opennms/pull/2207~~

### Open Issues

- [x] Verify Dashboard (including Dashlets) (@christianpape)
- [x] Verify BSM Admin Page (@christianpape)
- [x] Verify Wallboard(@christianpape)
- [x] Verify Surveillance View(@christianpape)
- [x] Verify JMX Configuration UI (@mvrueden)
- [x] Verify Node Maps (@mvrueden)
- [x] Verify Topology Maps (@mvrueden)
- [x] Verify Mibs and Snmp Compiler (@agalue)
- [ ] Verify Plugin Manager (@gallenc)
- [x] Release new Version of OSGI Plugin Manager (1.2.0 (see https://github.com/OpenNMS/osgi-plugin-manager/pull/5)) (@mvrueden)
- [x] Remove TODOs


smith-opennms:
image

christianpape:
The configuration page for surveillance view and the OpsBoard is currently broken. I'll fix...

pbrane:
Nice!

christianpape:
Checked Surveillance View/Dashboard, Ops Board, Dashlets and BSM configuration page.

features/topology-map/org.opennms.features.topology.app/src/main/java/org/opennms/features/topology/app/internal/gwt/client/ui/CustomDisplay.java
```diff
@@ -28,13 +28,13 @@
 
 package org.opennms.features.topology.app.internal.gwt.client.ui;
 
+import java.util.Collection;
```
RangerRick: We really need to get IDEA and Eclipse to agree on where `java.util.*` classes go.  :D

mvrueden: For exactly this reason, there is custom configuration you must apply in order to use IntelliJ accordingly to prevent these re-imports. See https://github.com/opennms/ide-setup and for even more details https://wiki.opennms.org/wiki/IDEA_and_OpenNMS

However, when I first used IntelliJ code were already contributed, not following the eclipse import guides. Therefore occasionally if the contributor is not carefully enough, you get re-imports. 

That is why I always recommend `git add -i -p` instead of `git add .` to check what you are committing. In addition `git checkout -i -p` helps you getting rid of those re-imports.

j-white:
For the ops board, there appears to be a new margin on the left:
image

agalue:
@mvrueden asked me to give it a try to verify the MIB compiler.

I've compiled CISCO-WIRELESS-IF-MIB and its dependencies. Everything went well, I was able to generate data collection and event definitions.

However, the UIs for editing/modifying events is not working. I can see an icon at the top right besides the table, but I cannot click it to see more details about it. To reproduce the problem, just install the branch, start OpenNMS, go to Admin, then "Customize Event Configurations", then choose any file from the list (Select Events Configuration File), and once the table with the event definitions is presented, choose one, and you'll see that nothing is presented below the table (which is the expected behavior). As this is not happening, I don't know if I can actually modify an event through the UI.

Nothing is shown on web.log and/or jetty-server.log, but I'm seeing the following on Karaf.log:

```
2018-11-19T07:54:16,775 ERROR com.vaadin.server:8.5.2(85) [qtp39117586-544] com.vaadin.server.DefaultErrorHandler:
java.lang.IllegalStateException: This component does not support the read-only mode, since state is of type CustomComponentState and does not inherit AbstractFieldState
        at com.vaadin.ui.AbstractComponent.setReadOnly(AbstractComponent.java:1055) [85:com.vaadin.server:8.5.2]
        at org.opennms.features.vaadin.events.EventForm.setReadOnly(EventForm.java:310) [277:org.opennms.features.vaadin-snmp-events-and-metrics:24.0.0.SNAPSHOT]
```

Same thing happens when trying to edit data collection groups (Admin -> Manage SNMP Collections and Data Collection Groups). To reproduce the problem is the same thing, there is no need to compile any MIB, just try to edit an exiting SNMP collection, or Data Collection Group.

As the implementation when clicking on a table to show the details below it is the same everywhere, I believe that after fixing one, the rest of them, can be easily fixed.

Please let me know when this part is fixed, so I can test again, and verify the pending behavior (modify something and make sure it is persisted).

mvrueden:
Thank you @agalue for your quick feedback. I addressed the problems you are having and hopefully all should work now. Please give it a re-spin.

agalue:
Overall, editing events, SNMP collections and data collection groups seems to be working as expected after your change. Although, I made just simple edits. If you want me to perform more comprehensive checks let me know, but my time is limited, so I would do it next week if necessary.

The only weird behavior I found is when scrolling over the tabular data on edit mode. The one that gave me trouble was the RRA list. The current workaround for the RRA table is click on an item and then scroll

mvrueden:
I can confirm the issue @agalue reported. I addressed it accordingly. From my point of view this is enough testing. If any issues arise at a later point we can still address them then. Thank you for your quick feedback and responsiveness.

mvrueden:
From my point of view this is ready to be merged to `develop`. I talked to @gallenc and he said if it compiles it works for him. If more issues pop up, we can address them then.

PR #2225, opened 2018-11-06
NMS-10363: Handle exceptions while running Drools rules.

cgorantla:
For streaming case, we reload engine.

* JIRA: http://issues.opennms.org/browse/NMS-10363

### DIFF
--- a/opennms-correlation/drools-correlation-engine/src/main/java/org/opennms/netmgt/correlation/drools/DroolsCorrelationEngine.java
+++ b/opennms-correlation/drools-correlation-engine/src/main/java/org/opennms/netmgt/correlation/drools/DroolsCorrelationEngine.java
@@ -122,7 +122,11 @@ public class DroolsCorrelationEngine extends AbstractCorrelationEngine {
     public synchronized void correlate(final Event e) {
         LOG.debug("Begin correlation for Event {} uei: {}", e.getDbid(), e.getUei());
         m_kieSession.insert(e);
-        if (!m_isStreaming) m_kieSession.fireAllRules();
+        try {
+            if (!m_isStreaming) m_kieSession.fireAllRules();
+        } catch (Exception e1) {
+            LOG.error("Exception while firing rules ", e1);
+        }
         m_eventsMeter.mark();
         LOG.debug("End correlation for Event {} uei: {}", e.getDbid(), e.getUei());
     }
@@ -133,7 +137,11 @@ public class DroolsCorrelationEngine extends AbstractCorrelationEngine {
         LOG.info("Begin correlation for Timer {}", timerId);
         TimerExpired expiration  = new TimerExpired(timerId);
         m_kieSession.insert(expiration);
-        if (!m_isStreaming) m_kieSession.fireAllRules();
+        try {
+            if (!m_isStreaming) m_kieSession.fireAllRules();
+        } catch (Exception e) {
+            LOG.error("Exception while firing rules ", e);
+        }
         LOG.debug("Begin correlation for Timer {}", timerId);
     }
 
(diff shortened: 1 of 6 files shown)


opennms-correlation/drools-correlation-engine/src/test/opennms-home/etc/drools-engine.d/droolsFusion/DroolsFusion.drl
```diff
@@ -34,6 +36,22 @@ then
...
 
+
+rule "this-rule-should-throw-exception"
+       salience 400
+when    $e : Event(uei matches "uei.opennms.org/triggerTestForFusion")
+
+then
+         System.err.println("running exception rule, this will throw exception");
+         EventBuilder eventBuilder = new EventBuilder("uei.opennms.org/triggerTestForFusion", "Component Correlator");
+         // Code to generate ConcurrentModification Exception. see NMS-10363
```
j-white: I guess we could also just directly throw the CME instead of trying to replicate it.

opennms-correlation/drools-correlation-engine/src/main/java/org/opennms/netmgt/correlation/drools/DroolsCorrelationEngine.java
```diff
@@ -214,7 +222,12 @@ public void initialize() throws Exception {
         if (m_isStreaming) {
             new Thread(() -> {
                 Logging.putPrefix(getClass().getSimpleName() + '-' + getName());
-                m_kieSession.fireUntilHalt();
+                try {
+                    m_kieSession.fireUntilHalt();
+                } catch (Exception e) {
+                    LOG.error("Exception while running rules, reloading engine ", e);
```
j-white: Should we not also send out an even that triggers an alarm when this happen?

opennms-correlation/drools-correlation-engine/src/main/java/org/opennms/netmgt/correlation/drools/DroolsCorrelationEngine.java
```diff
@@ -214,11 +224,29 @@ public void initialize() throws Exception {
...
+                    reloadConfig();
+                }
             }, "FireTask").start();
         }
     }
 
+    private void triggerAlarm(Exception exception) {
+        EventBuilder ebldr = new EventBuilder(EventConstants.DROOLS_ENGINE_ENCOUNTERED_EXCEPTION, getName());
+        ebldr.addParam(EventConstants.PARM_STACK_TRACE_STRING, ExceptionUtils.getStackTrace(exception));
+        // Add alarmData with alarm type as notification.
```
j-white: We should probably define the alarm data within the configuration as to hard-coding it here.

For `foundation-2017`, we could add this to a new file such as  `etc/eventconf/opennms.correlator.events.xml`, but omit to include this in the `eventconf.xml` by default since that will surely cause conflicts.

In `release-23.0.1`, we could then update add the import to the `eventconf.xml`.

opennms-correlation/drools-correlation-engine/src/test/java/org/opennms/netmgt/correlation/drools/FusionRulesIT.java
```diff
@@ -0,0 +1,175 @@
...
+        // This will create an exception and the engine should reload.
+        engine.correlate(eb.getEvent());
+        Thread.sleep(10000); // Wait time for the engine to reload.
+        // The engine should run other rules after reload.
+        engine.correlate(createNodeLostServiceEvent(1, "SSH"));
+        Thread.sleep(4000); // Give time to execute the time-based rules.
+        m_anticipatedMemorySize = 5;
+        engine.getKieSessionObjects().forEach(System.err::println);
+        verify(engine);
+        OnmsAlarm alarm = m_alarmDao.findByReductionKey(EventConstants.DROOLS_ENGINE_ENCOUNTERED_EXCEPTION + ":" + "droolsFusion");
```
j-white: I think it would be sufficient to verify that the event is generated when an exception occurs - we know alarm handling works, so there's no need to verify that explicitly here.

cgorantla:
Thanks @j-white  for the review. Addressed review comments.

j-white:
👍

PR #2227, opened 2018-11-08
HZN-1436: Packaging OpenNMS plugins

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1436

Here we add the `karaf-extender` feature both the opennms and sentinel containers.

We also extend the feature to support waiting on `.kar` files before installing the features. This gets us closer to be able to distributed plugins as .kar files, and allows us to reliably start features in the .kar files at boot.

For example, we can now do the following on either of the opennms or sentinel containers:
```
wget https://360-128952957-gh.circle-artifacts.com/0/dist/opennms-oce-plugin.kar
cp $KARAF_HOME/deploy/
echo 'oce-datasource-api wait-for-kar=opennms-oce-plugin' > $KARAF_HOME/etc/featuresBoot.d/oce.boot
```

And have the 'oce-datasource-api' feature reliably installed on startup - even when `$KARAF_HOME/data`, or `$KARAF_HOME/data/cache` is wiped.

### DIFF
--- a/container/extender/src/main/java/org/opennms/karaf/extender/Feature.java
+++ b/container/extender/src/main/java/org/opennms/karaf/extender/Feature.java
@@ -31,34 +31,70 @@ package org.opennms.karaf.extender;
 import java.util.Objects;
 
 public class Feature {
-    private final String m_name;
-    private final String m_version;
+    private final String name;
+    private final String version;
+    private final String karDependency;
 
-    public Feature(String name) {
-        this(name, null);
+    private Feature(Builder builder) {
+        this.name = builder.name;
+        this.version = builder.version;
+        this.karDependency = builder.karDependency;
     }
 
-    public Feature(String name, String version) {
-        m_name = Objects.requireNonNull(name);
-        m_version = version;
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder {
+        private String name;
+        private String version;
+        private String karDependency;
+
+        public Builder withName(String name) {
+            this.name = name;
+            return this;
+        }
+
+        public Builder withVersion(String version) {
+            this.version = version;
+            return this;
+        }
+
+        public Builder withKarDependency(String karDependency) {
+            this.karDependency = karDependency;
+            return this;
+        }
+
+        public Feature build() {
+            Objects.requireNonNull(name, "name is required");
+            return new Feature(this);
+        }
     }
 
     public String getName() {
-        return m_name;
+        return name;
     }
 
     public String getVersion() {
-        return m_version;
+        return version;
+    }
+
+    public String getKarDependency() {
+        return karDependency;
+    }
+
+    public String toInstallString() {
+        return getVersion() != null ? getName() + "/" + getVersion() : getName();
     }
 
     @Override
     public String toString() {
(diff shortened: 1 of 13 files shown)


container/extender/src/main/java/org/opennms/karaf/extender/KarDependencyHandler.java
```diff
@@ -0,0 +1,166 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(KarDependencyHandler.class);
+
+    private static final int KAR_LIST_SLEEP_MS = 5000;
+
+    private final List<Feature> features;
+    private final KarService karService;
+    private final FeaturesService featuresService;
+
+    private static final String FEATURE_CONFIG_FILE = "features.cfg";
+    private static final String KAR_STORAGE = System.getProperty("karaf.data") + File.separator + "kar";
```
RangerRick: Is this where it ends up getting copied to _after_ `/deploy/` through some secret Karaf voodoo?  Or should this be `deploy` rather than the `karaf.data` directory?

j-white: Yeah. The `kar` service in Karaf will extract/install the `.kar` files here.

container/extender/src/main/java/org/opennms/karaf/extender/KarafExtender.java
```diff
@@ -111,78 +119,109 @@ public void init() throws InterruptedException {
...
+            m_installThread.start();
+        } else {
+            LOG.debug("No features to install.");
+        }
+
+
+        final List<Feature> featuresWithKarDependencies = featuresBoot.stream()
+                .filter(f -> f.getKarDependency() != null)
+                .collect(Collectors.toList());
+        if (featuresWithKarDependencies.size() >= 1) {
```
mattixtech:
```suggestion
        if (!featuresWithKarDependencies.isEmpty()) {
```

container/extender/src/main/java/org/opennms/karaf/extender/Feature.java
```diff
@@ -31,34 +31,70 @@
...
     }
 
-    public Feature(String name, String version) {
-        m_name = Objects.requireNonNull(name);
-        m_version = version;
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder {
```
mattixtech:
```suggestion
    private static class Builder {
```

container/extender/src/main/java/org/opennms/karaf/extender/KarDependencyHandler.java
```diff
@@ -0,0 +1,166 @@
...
+                    break;
+                }
+            } catch (Exception e) {
+                LOG.warn("Enumerating installed .kar files failed. Will retry in {}ms.", KAR_LIST_SLEEP_MS, e);
+            }
+
+            try {
+                Thread.sleep(KAR_LIST_SLEEP_MS);
+            } catch (InterruptedException e) {
+                LOG.info("Interrupted. Stopping thread.");
```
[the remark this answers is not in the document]
j-white: Good catch. Fixed.

PR #2229, opened 2018-11-12
HZN-1426 Make Older Alarms Un-Acknowledge Situations

smith-opennms:
This PR addresses the use case where an existing, un-acknowledged alarm is correlated to a Situation that is acknowledged. Previously, only the acknowledgment time of the alarm would be checked and the Situation would remain Acknowledged. Now, we additionally compare the time the Alarm is correlated to the situation and if that is later than the time at which the Situation was acknowledged, we un-acknowledge the Situation.

* JIRA: https://issues.opennms.org/browse/HZN-1426

### DIFF
--- a/opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/drools/DefaultAlarmService.java
+++ b/opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/drools/DefaultAlarmService.java
@@ -29,6 +29,7 @@
 package org.opennms.netmgt.alarmd.drools;
 
 import java.util.Date;
+import java.util.Set;
 
 import org.opennms.netmgt.dao.api.AcknowledgmentDao;
 import org.opennms.netmgt.dao.api.AlarmDao;
@@ -80,6 +81,12 @@ public class DefaultAlarmService implements AlarmService {
         if (alarmInTrans == null) {
             LOG.warn("Alarm disappeared: {}. Skipping delete.", alarm);
             return;
+        }
+        // If alarm was in Situation, fire notifications for the Situation
+        if (alarmInTrans.isPartOfSituation()) {
+            for (OnmsAlarm situation : alarmInTrans.getRelatedSituations()) {
+                alarmEntityNotifier.didUpdateRelatedAlarms(situation, situation.getRelatedAlarms());
+            }
         }
         alarmDao.delete(alarmInTrans);
         alarmEntityNotifier.didDeleteAlarm(alarmInTrans);
--- a/opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/drools/DefaultAlarmService.java
+++ b/opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/drools/DefaultAlarmService.java
@@ -29,6 +29,10 @@
 package org.opennms.netmgt.alarmd.drools;
 
 import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
 
 import org.opennms.netmgt.dao.api.AcknowledgmentDao;
 import org.opennms.netmgt.dao.api.AlarmDao;
@@ -81,13 +85,18 @@ public class DefaultAlarmService implements AlarmService {
             LOG.warn("Alarm disappeared: {}. Skipping delete.", alarm);
             return;
         }
-        // If alarm was in Situation, fire notifications for the Situation
+        // If alarm was in Situation, calculate notifications for the Situation
+        Map<OnmsAlarm, Set<OnmsAlarm>> priorRelatedAlarms = new HashMap<>();
         if (alarmInTrans.isPartOfSituation()) {
             for (OnmsAlarm situation : alarmInTrans.getRelatedSituations()) {
-                alarmEntityNotifier.didUpdateRelatedAlarms(situation, situation.getRelatedAlarms());
+                priorRelatedAlarms.put(situation, situation.getRelatedAlarms());
             }
         }
         alarmDao.delete(alarmInTrans);
+        // fire notifications after alarm has been deleted
+        for (Entry<OnmsAlarm, Set<OnmsAlarm>> entry : priorRelatedAlarms.entrySet()) {
+            alarmEntityNotifier.didUpdateRelatedAlarms(entry.getKey(), entry.getValue());
+        }
         alarmEntityNotifier.didDeleteAlarm(alarmInTrans);
     }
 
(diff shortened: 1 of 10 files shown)


opennms-model/src/main/java/org/opennms/netmgt/model/AlarmAssociation.java
```diff
@@ -45,6 +45,8 @@
 import javax.persistence.TemporalType;
 import javax.persistence.UniqueConstraint;
 
+import org.springframework.core.style.ToStringCreator;
```
mvrueden: Can we please use google guava's https://google.github.io/guava/releases/19.0/api/docs/com/google/common/base/Objects.ToStringHelper.html here instead?

smith-opennms: fwiw, it's marked as deprecated. Did you want to use MoreObjects.ToStringHelper instead?
also fwiw, i brazenly copied this idiom from one of the other model classes (to be consistent) - there are about a dozen and a half that use ToStringCreator. Shall we (in a separate commit) migrate them all?

mvrueden: Yeah, I meant `MoreObjects.ToStringHelper`, sorry.

My concern was just, to not introduce more Spring ToStringCreators, as it may require the spring core dependency vs the guava dependency, which may leads to problems if we want it to behave more nicely with osgi. 

Besides that we already use ToStringHelper (guava) and the apache commans equivalent. Until now I didn't know that there even is a ToStringCreator and we use it. But who was I kidding that we were not using it (-: Anyways as it is not used that heavily, I would opt for migrating them to guava instead.

opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/drools/DefaultAlarmService.java
```diff
@@ -81,6 +81,12 @@ public void deleteAlarm(OnmsAlarm alarm) {
             LOG.warn("Alarm disappeared: {}. Skipping delete.", alarm);
             return;
         }
+        // If alarm was in Situation, fire notifications for the Situation
+        if (alarmInTrans.isPartOfSituation()) {
+            for (OnmsAlarm situation : alarmInTrans.getRelatedSituations()) {
+                alarmEntityNotifier.didUpdateRelatedAlarms(situation, situation.getRelatedAlarms());
```
j-white: This should probably be issued after the call to `alarmDao#delete`. Won't `situation.getRelatedAlarms()` still contain the given alarm we deleted.

opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/drools/DroolsAlarmContext.java
```diff
@@ -180,19 +184,43 @@ public void handleDeletedAlarm(int alarmId, String reductionKey) {
         }
     }
 
+    private void handleRelatedAlarms(OnmsAlarm situation) {
+        if (!situation.isSituation()) {
+            return;
+        }
+        Map<Integer, AlarmAssociationAndFact> associationFacts = alarmAssociationById.get(situation.getId());
+        if (associationFacts == null) {
```
j-white: Can be replaced with call to `computeIfAbsent`.

opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/drools/DroolsAlarmContext.java
```diff
@@ -180,19 +184,43 @@ public void handleDeletedAlarm(int alarmId, String reductionKey) {
...
+            associationFacts = new HashMap<>();
+            alarmAssociationById.put(situation.getId(), associationFacts);
+        }
+        for (AlarmAssociation association : situation.getAssociatedAlarms()) {
```
j-white: What happens if an associated alarm is removed? It seems that we won't actually remove the "alarm association fact" until the situation itself is deleted.

opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/drools/DroolsAlarmContext.java
```diff
@@ -233,6 +262,20 @@ private void deleteAlarmAcknowledgement(int alarmId) {
...
 
+    private void deleteAlarmAssociations(int alarmId) {
+        Map<Integer, AlarmAssociationAndFact> associationFacts = alarmAssociationById.remove(alarmId);
+        if (associationFacts == null) {
+            return;
+        }
+        for (Integer association : associationFacts.keySet()) {
+            AlarmAssociationAndFact assocationFact = associationFacts.get(association);
+            if (assocationFact != null) {
+                LOG.debug("Deleting ack from session: {}", assocationFact.getAlarmAssociation());
```
j-white: ack vs association

opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/drools/DefaultAlarmService.java
smith-opennms: This is desired behaviour.
Second param of didUpdateRelatedAlarms is 'previousRelatedAlarms'

opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/drools/DroolsAlarmContext.java
smith-opennms: Thanks. will change.
I'm still living in Java 1.2....

opennms-model/src/main/java/org/opennms/netmgt/model/AlarmAssociation.java
smith-opennms: done. and did the rest under JIRA https://issues.opennms.org/browse/NMS-10455

opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/drools/DroolsAlarmContext.java
smith-opennms: done.

opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/drools/DroolsAlarmContext.java
smith-opennms: done. thx for the catch

opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/drools/DefaultAlarmService.java
j-white: To help clarify the expected behavior - let's assume that originally `situation.getRelatedAlarms()` returns {A,B,C}.

If we're deleting B, then we would expect `didUpdateRelatedAlarms` to be called. When called:
1. `situation.getRelatedAlarms()` should return {A,C} and
1. previousRelatedAlarms should be set to {A,B,C}

It looks like this second part is true, but since the alarm isn't deleted until after the callback is made, calling `situation.getRelatedAlarms()` still returns {A,B,C}.

j-white:
Shouldn't we change `EifMessageDecoder#charset` by default as well?

PR #2231, opened 2018-11-12
NMS-10445: Improve performance of node search

patrick-schweizer:
* JIRA: https://issues.opennms.org/browse/NMS-10445


j-white:
Makes sense. 👍

PR #2234, opened 2018-11-14
NMS-8185: Creation of a Daemon Config Page

Zottel92:
* JIRA: https://issues.opennms.org/browse/NMS-8185

Should be merged AFTER PR https://github.com/OpenNMS/opennms/pull/2213


mattixtech:
Not sure if this is the right way to do it, but for review purposes it might be easier to target your PR against the branch it is dependent on so that we see a true diff and the dependent commits are not shown as changes in this PR.

So you could target this PR against the branch jira/NMS-10427 until it is ready to be merged at which point you could retarget it to features/daemon-reload-page.

Your local branch while working on this could then be tracking origin/jira/NMS-10427 so you can easily pull in updates.

mvrueden:
I get your point @mattixtech, but if people don't read the `Should be merged after` disclaimer, which indicates, that a review may not be ideal at this point, what will prevent merging into the wrong branch?
That's why I think we should not adapt that method, even if it would solve the issue we encountered.

I think ultimately the best approach is, to just not create a PR if it depends on another one beeing merged before. That is how I usually (apperently not usual enough) handle it. In addition to that, reviewers should read the creator's comment.

core/web-assets/src/main/assets/js/apps/onms-daemons/index.js
```diff
@@ -0,0 +1,106 @@
...
+        'angular-loading-bar',
+        'ngResource',
+        'ui.router',
+        'ui.bootstrap',
+        'ui.checkbox',
+        'ui.toggle',
+        'onms.http',
+        'onms.elementList',
+        'mwl.confirm',
+        'onms.pagination'
```
mvrueden: Do we need all these modules?
It looks to me, they are not all used, can we only include those which are actually required?

core/web-assets/src/main/assets/js/apps/onms-daemons/index.js
```diff
@@ -0,0 +1,106 @@
...
+    ])
+        .config(['$locationProvider', function ($locationProvider) {
+            $locationProvider.hashPrefix('!');
+            $locationProvider.html5Mode(false);
+        }])
```
mvrueden: A `$locationProvider` is not required, if there is only one view.

core/web-assets/src/main/assets/js/apps/onms-daemons/index.js
```diff
@@ -0,0 +1,106 @@
...
+                });
+        }])
+        .controller('ListController', ['$scope', '$http', '$timeout', function ($scope, $http, $timeout) {
+            $scope.serverUrl = location.protocol + '//' + location.hostname+(location.port ? ':'+location.port: '');
+
+
+            $scope.daemons = [];
+
+
+            $scope.LookupDaemonState = function (daemon) {
```
mvrueden: Is there a particular reason for the new lines?

mvrueden:
```suggestion
            $scope.lookupDaemonState = function (daemon) {
```

core/web-assets/src/main/assets/js/apps/onms-daemons/index.js
```diff
@@ -0,0 +1,106 @@
...
+                        daemon.reloadIconClass = "glyphicon glyphicon glyphicon-remove-sign";
+                    } else {
+                        $timeout($scope.LookupDaemonState, 1000, true, daemon);
+                    }
+                });
+            };
+
+            $scope.doTheCopy = function (daemon) {
+                //MAGIC
+            };
```
mvrueden: What is this supposed to do?

core/web-assets/src/main/assets/js/apps/onms-daemons/index.js
```diff
@@ -0,0 +1,106 @@
...
+            $scope.LookupDaemonState = function (daemon) {
+                if (daemon.reloadCount > 5) {
+                    daemon.isReloading = false;
+                    daemon.reloadNote = "Undefined";
+                    daemon.reloadIconClass = "glyphicon glyphicon-question-sign";
+                    return;
+                }
+                daemon.reloadCount = daemon.reloadCount + 1;
+
+                $http.get('rest/daemons/checkReloadState/' + daemon.name + '/').then(function (response) {
```
mvrueden: Now seeing this URL, we should change it, as it does not feel very rest like.
Something like `daemons/<daemon-name>/state` is more fitting. 
This may be addressed in a separate issue.

core/web-assets/src/main/assets/js/apps/onms-daemons/index.js
```diff
@@ -0,0 +1,106 @@
...
+
+
+            $scope.daemons = [];
+
+
+            $scope.LookupDaemonState = function (daemon) {
+                if (daemon.reloadCount > 5) {
+                    daemon.isReloading = false;
+                    daemon.reloadNote = "Undefined";
+                    daemon.reloadIconClass = "glyphicon glyphicon-question-sign";
```
mvrueden: The reloadIconClasses should be handled in the view instead of the controller.

core/web-assets/src/main/assets/js/apps/onms-daemons/index.js
```diff
@@ -0,0 +1,106 @@
...
+        .config(['$stateProvider', function ($stateProvider) {
+            $stateProvider
+                .state('daemons', {
+                    url: '',
+                    controller: 'ListController',
+                    templateUrl: indexTemplate
+                });
+        }])
+        .controller('ListController', ['$scope', '$http', '$timeout', function ($scope, $http, $timeout) {
+            $scope.serverUrl = location.protocol + '//' + location.hostname+(location.port ? ':'+location.port: '');
```
mvrueden: NIT: A comment may be added to indicate that this is used in the view and not to make actual rest calls.

core/web-assets/src/main/assets/js/apps/onms-daemons/index.js
```diff
@@ -0,0 +1,106 @@
...
+                        $timeout($scope.LookupDaemonState, 1000, true, daemon);
+                    }
+                });
+            };
+
+            $scope.doTheCopy = function (daemon) {
+                //MAGIC
+            };
+
+            $scope.reloadPressed = function (daemon) {
```
mvrueden:
```suggestion
            $scope.reloadDaemon = function (daemon) { // or triggerDaemonReload
```

core/web-assets/src/main/assets/js/apps/onms-daemons/index.js
```diff
@@ -0,0 +1,106 @@
...
+                $http.post('rest/daemons/reload/' + daemon.name + '/')
+                    .success(function (data, status) {
+                        if (status === 204) {
+                            var now = new Date();
+                            daemon.reloadTime = now.getTime() + (now.getTimezoneOffset() * 60000);
+                            daemon.showReloadState = true;
+                            daemon.isReloading = true;
+                            daemon.reloadCount = 0;
+                            daemon.reloadNote = "Reloading...";
+                            daemon.reloadIconClass = "fa fa-spinner fa-spin";
```
mvrueden: If possible, this would be better handled in the view instead of the controller

core/web-assets/src/main/assets/js/apps/onms-daemons/index.js
```diff
@@ -0,0 +1,106 @@
...
+                    }
+                });
+            };
+
+            $scope.doTheCopy = function (daemon) {
+                //MAGIC
+            };
+
+            $scope.reloadPressed = function (daemon) {
+                $http.post('rest/daemons/reload/' + daemon.name + '/')
```
mvrueden: Same as above, this may be better at `rest/daemons/<name>/reload` in order to be more rest like.

smoke-test/src/test/java/org/opennms/smoketest/DaemonConfigPageIT.java
```diff
@@ -0,0 +1,258 @@
...
+    private String getExceptedCurlStringForDaemonName(String name){
+        return String.format("curl --request POST -u USERNAME:PASSWORD -url %sopennms/rest/daemons/reload/%s/", getBaseUrl(), name);
+    }
+
+    private class Daemon {
+        private final DaemonReloadPage page;
+        private final String daemonName;
+
+        public Daemon(DaemonReloadPage page, String daemonName) {
+            this.page = page;
```
mvrueden: This looks weird, as this seems to be an enforced relationship, but there is no need to. Probably better to make the `page.getRowForDaemonName(this.daemonName)` and `page.getPopupPreElementForDaemonName(this.daemonName)` methods utility classes, as they don't require anything from the page

smoke-test/src/test/java/org/opennms/smoketest/DaemonConfigPageIT.java
```diff
@@ -0,0 +1,258 @@
...
+            this.daemonName = daemonName;
+        }
+
+        public String getName() {
+            // Assert That Right Name is shown
+            WebElement nameCell = getDaemonRowElement().findElements(By.xpath("./td[1]")).get(0);
```
mvrueden: Any reason why `findElements` instead of `findElement` is used, if only one element is relevant anyways?
Also ensure that these calls are wrapped in a `setImplicitWait(5, SECONDS)`, otherwise execution takes very long if those elements are not visible due to changes or errors.

smoke-test/src/test/java/org/opennms/smoketest/DaemonConfigPageIT.java
```diff
@@ -0,0 +1,258 @@
...
+        }
+    }
+
+    private class DaemonReloadPage {
+
+        private final int expectedDaemonCount = 30;
+
+        public DaemonReloadPage open() {
+            m_driver.get(baseURI + "opennms/admin/daemons/index.jsp");
+            new WebDriverWait(m_driver, 5).until((Predicate<WebDriver>) (driver) -> getDaemonRows().size() == expectedDaemonCount);
```
mvrueden: This would probably be better manually verified when invoking the open() instead of here.
Otherwise `expectedDaemonCount` should be a parameter, which makes the API strange to use.

core/web-assets/src/main/assets/js/apps/onms-daemons/index.js
Zottel92: I also would pledge for addressing this in another issue.

mvrueden: Yep.

mvrueden:
If tests pass now, this can be merged.

core/web-assets/src/main/assets/js/apps/onms-daemons/index.js
```diff
@@ -0,0 +1,92 @@
...
+                        if (status === 204) {
+                            var now = new Date();
+                            daemon.reloadTime = now.getTime() + (now.getTimezoneOffset() * 60000);
+                            daemon.isReloading = true;
+                            daemon.reloadCount = 0;
+                            daemon.reloadState = "Reloading";
+                            $timeout($scope.lookupDaemonState, 1000, true, daemon);
+                        }
+                    });
+            };
```
mvrueden: Looking at the reload logic here, there is the following issue.
If the rest endpoint cannot be reached or failed, reloading is not failing immediately, instead the whole ui is hanging in the reloading state. 
In my opinion the reload logic should be implemented in a more robust way:
 - click on reload sets the ui in state reloading
 - if reloading due to post issue failed -> State Failure -> no event was sent
 - if reloading is timing out a "Timeout" failure should be posted
 - if reload succeeds or fails within the 5 seconds the result should be accordingly

mvrueden:
I puddle back a bit. The daemon reload logic should be hardened before merging. Adding `triage` label.

core/web-assets/src/main/assets/js/apps/onms-daemons/index.html
```diff
@@ -0,0 +1,83 @@
...
+
+                    <span class="text-warning" ng-if="d.reloadState == 'Undefined'"><i class="fa fa-question-circle"/></span>
+                    <span class="small" ng-if="d.reloadState == 'Undefined'">It is unclear if the reload event was processed successfully</span>
+
+                    <span class="text-muted" ng-if="d.reloadState == 'Reloading'"><i class="fa fa-spinner fa-spin"/></span>
+                    <span class="small" ng-if="d.reloadState == 'Reloading'">Reloading...</span>
+                </td>
+
+                <!-- Modal -->
+                <div class="modal" id="curlModal_{{d.name}}" tabindex="-1" role="dialog" aria-labelledby="curlModalLabel">
```
mvrueden: The modal should be implemented using one template only and a modalcontroller to set the variables accordingly

mvrueden:
I addressed the problems accordingly. If the tests pass, it can finally be merged \o/

PR #2240, opened 2018-11-15
NMS-10454: Update EIF Adapter To Support Additional Protocol Version

UberPinguin:
Compare the 2-byte value at offset 30 from <START>> with the 2-byte value
at offset 34. If they are equal, assume the event body begins at
offset 36 from <START>>.
Otherwise, assume the event body begins at offset 37 from <START>>.


UberPinguin:
I found that the `EifAdapterBlueprintTest#testCanParseEifWith36ByteOffset` test doesn't pass when executed with Maven surefire on Windows.
Setting the character set to `StandardCharsets.UTF_8` in [EifMessageDecoder](https://github.com/OpenNMS/opennms/blob/foundation-2018/features/eif-adapter/src/main/java/org/opennms/features/eifadapter/EifMessageDecoder.java#L47) fixes that, but I'm not sure if this is the correct approach. Should the system default character set be honored here, or should we force it to UTF-8?

j-white:
If the payloads are expected to be encoded in UTF-8, then I don't see any problems with forcing it to UTF-8 instead of using the system default.

UberPinguin:
I opted to set the charset to UTF-8 when reading in the packet capture in the unit test.
I don't know that we can expect the charset on the EIF stream to be UTF-8, or any other particular encoding for that matter.

j-white:
In that case, we should also update charset used for decoding to be UTF-8 here: https://github.com/OpenNMS/opennms/blob/b7b29407c91a991bf7a8dfeb80c785f4ac762040/features/eif-adapter/src/main/java/org/opennms/features/eifadapter/EifMessageDecoder.java#L47

UberPinguin:
Yep, I agree.
Do you want me to squash these into a single commit.

j-white:
@UberPinguin a new commit would be fine, we can squash when we merge the PR.

UberPinguin:
@j-white OK. I pushed commit 96f0001 on the 20th.

j-white:
Updated `pr/jira/NMS-10454` with latest code, should be good to merge once that build passes.

UberPinguin:
I see that `EifTranslatorTest.testCanConnectEifEventToNodeWithHostname` failed in Bamboo.
However, all tests pass on both my Windows laptop and my CentoS 7.3.1611 VM. I don't know why that test is failing in Bamboo.

UberPinguin:
Ah, scratch that:
```
2018-11-26 12:17:31,811 ERROR [main] org.opennms.features.eifadapter.EifParser - UnknownHostException while resolving hostname localhost
--
2018-11-26 12:17:31,811 ERROR [main] org.opennms.features.eifadapter.EifParser - UnknownHostException while resolving hostname localhost
2018-11-26 12:17:31,811 ERROR [main] org.opennms.features.eifadapter.EifParser - UnknownHostException while resolving origin 127.0.0.1
2018-11-26 12:17:31,811 DEBUG [main] org.opennms.features.eifadapter.EifParser - Searching for node by label 127.0.0.1
2018-11-26 12:17:31,811 DEBUG [main] org.opennms.features.eifadapter.EifParser - connectEifEventToNode : No matching nodes found. Defaulting to nodeId 0.
```
Looks like a DNS issue in Bamboo?

UberPinguin:
@j-white I added commit fb54e86 to allow the `EifTranslatorTest.testCanConnectEifEventToNodeWithHostname` to succeed when DNS fails for localhost.

UberPinguin:
This is ready for merge, if there are no objections or critiques.

PR #2242, opened 2018-11-18
NMS-10455: Use Guava MoreObjects.toStringHelper 

smith-opennms:
This PR modifies the toString of model object to use Guava's toStringHelper instead of Spring's toStringCreator as suggested in review of PR #2229 

* JIRA: http://issues.opennms.org/browse/NMS-10455


j-white:
We should also add a check-style rule for this to enforce it going forward. See https://github.com/OpenNMS/opennms/blob/develop/checkstyle/src/main/resources/nms_checks.xml#L11.

PR #2245, opened 2018-11-20
HZN-1391: Added ssh command to bin/opennms script

christianpape:
* JIRA: https://issues.opennms.org/browse/HZN-1391


mvrueden:
Can we savely assume to always use the `admin` user or do we need to add an option for this?

pbrane:
Should we make the script prompt for the user name with admin as default or should we use $USER as the default?

christianpape:
The script is invoked as root, so I think using the value of $USER makes no sense. I've found that system.properties includes "karaf.local.user = admin". So, I'll change the script to lookup the username in this file...

PR #2246, opened 2018-11-20
NMS-10227: Add 'Select All' / 'Unselect All' buttons for Scheduled Ou…

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-10227


indigo423:
Tested the feature, the only thing I've noticed you have to make sure it is the last action in the web site, otherwise the selection gets deleted. For example you add a node "Select All" and add a schedule deletes the whole selection.

christianpape:
Yes, you're right. I noticed this behaviour before I added the selection buttons. So, I think we should create a new issue for this problem.

indigo423:
Created [NMS-10478](https://issues.opennms.org/browse/NMS-10478). Awesome close one crate another one. We don't make progress :D

mvrueden:
LGTM

PR #2249, opened 2018-11-21
NMS-10455: Add checkstyle rule to enforce MoreObjects.toStringHelper

smith-opennms:
This PR adds a Checkstyle rule that will prevent contributors from using the SpringFramework's ToStringCreator

JIRA: http://issues.opennms.org/browse/NMS-10455


mvrueden:
LGTM 👍

PR #2250, opened 2018-11-21
NMS-10464: Mode debug logs to trace

cgorantla:
* JIRA: http://issues.opennms.org/browse/NMS-10464


j-white:
Can we retarget this to `foundation-2018`?

cgorantla:
I was wrong.  Offset provider doesn't exist in foundation-2018. re-targeted to 23.0.2

PR #2251, opened 2018-11-21
NMS-10468: Fix offheap name from .doc to .adoc

cgorantla:
* JIRA: http://issues.opennms.org/browse/NMS-10468


cgorantla:
Thanks @indigo423.  Yes, we no longer need that reference.

PR #2253, opened 2018-11-26
OCE-27: Fix transaction related issues when performing managed object tagging

j-white:
JIRA: https://issues.opennms.org/browse/OCE-27

Here we extend the ` org.opennms.netmgt.dao.api.SessionUtils` interface to include a new `withReadOnlyTransaction` method that performs the call within read-only transaction with auto-flush disabled.

All of the existing calls in the `api-layer` module are updated to use this call, which solves the exception that was reported in the issue.

### DIFF
--- a/opennms-dao-api/src/main/java/org/opennms/netmgt/dao/api/SessionUtils.java
+++ b/opennms-dao-api/src/main/java/org/opennms/netmgt/dao/api/SessionUtils.java
@@ -51,6 +51,15 @@ public interface SessionUtils {
     <V> V withTransaction(Supplier<V> supplier);
 
     /**
+     * Invoked the given supplier within the context of a read-only transaction.
+     *
+     * @param supplier supplier to invoke
+     * @param <V> type returned by the supplier
+     * @return value returned by the supplier
+     */
+    <V> V withReadOnlyTransaction(Supplier<V> supplier);
+
+    /**
      * Converts the flush mode for the current session factory to MANUAL
      * for the duration of the call to the given supplier.
      *
--- a/opennms-dao-api/src/main/java/org/opennms/netmgt/dao/api/SessionUtils.java
+++ b/opennms-dao-api/src/main/java/org/opennms/netmgt/dao/api/SessionUtils.java
@@ -51,7 +51,7 @@ public interface SessionUtils {
     <V> V withTransaction(Supplier<V> supplier);
 
     /**
-     * Invoked the given supplier within the context of a read-only transaction.
+     * Invokes the given supplier within the context of a read-only transaction.
      *
      * @param supplier supplier to invoke
      * @param <V> type returned by the supplier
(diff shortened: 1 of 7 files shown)


opennms-dao-api/src/main/java/org/opennms/netmgt/dao/api/SessionUtils.java
```diff
@@ -50,6 +50,15 @@
      */
     <V> V withTransaction(Supplier<V> supplier);
 
+    /**
+     * Invoked the given supplier within the context of a read-only transaction.
```
mattixtech:
```suggestion
     * Invokes the given supplier within the context of a read-only transaction.
```

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/DefaultSessionUtils.java
```diff
@@ -60,4 +84,60 @@
...
+     * @param ex the thrown application exception or error
+     * @throws TransactionException in case of a rollback error
+     */
+    private void rollbackOnException(TransactionStatus status, Throwable ex) throws TransactionException {
+        LOG.debug("Initiating transaction rollback on application exception", ex);
+        try {
+            this.transactionManager.rollback(status);
+        }
+        catch (TransactionSystemException ex2) {
+            LOG.error("Application exception overridden by rollback exception", ex);
```
mattixtech: I'm assuming you want to print the rollback exception here, should be `ex2`. Same for below and for `Error err`.

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/DefaultSessionUtils.java
```diff
@@ -60,4 +84,60 @@
...
+            // Transactional code threw application exception -> rollback
+            rollbackOnException(status, ex);
+            throw ex;
+        }
+        catch (Error err) {
+            // Transactional code threw error -> rollback
+            rollbackOnException(status, err);
+            throw err;
+        }
+        catch (Throwable ex) {
```
mattixtech: Wouldn't any `Supplier<V>` passed to this method cause a compile time error if it was throwing a non-runtime exception?

Maybe this is just a leftover from copying or I'm missing something. If its just a result of copying then its not a problem anyway, it just looks like it would be dead code to me.

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/DefaultSessionUtils.java
j-white: Not necessarily, those are re-thrown, so the caller can catch/print those.

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/DefaultSessionUtils.java
j-white: Suppliers can't throw checked exceptions AFAIK.

This code was copied verbatim.

PR #2260, opened 2018-11-27
NMS-10463: Show only situations with severity greater NORMAL on landing page

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-10463


indigo423:
Tested by sending a uei.opennms.org/nodes/dataCollectionSucceeded event which creates an Alarm with severity normal and is not shown on the start page.

PR #2262, opened 2018-11-27
HZN-1440: Support syslog message parsing for messages found during audit

mattixtech:
This PR enhances syslog message parsing to handle cases that were failing to parse in the field.

* JIRA: http://issues.opennms.org/browse/HZN-1440

### DIFF
--- a/features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/GrokParserStageSequenceBuilder.java
+++ b/features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/GrokParserStageSequenceBuilder.java
@@ -53,6 +53,7 @@ import org.slf4j.LoggerFactory;
  * <li>MONTH: 3-character English month abbreviation.</li>
  * <li>NOSPACE: String that contains no whitespace.</li>
  * <li>STRING: String. Because this matches any character, it must be followed by a delimiter in the pattern string.</li>
+ * <li>WHITESPACE: String that contains only whitespace (spaces and/or tabs).</li>
  * </ul>
  * 
  * @author Seth
@@ -77,7 +78,8 @@ public abstract class GrokParserStageSequenceBuilder {
 		IPADDRESS,
 		MONTH,
 		NOSPACE,
-		STRING
+		STRING,
+		WHITESPACE
 	}
 
 	/**
@@ -217,6 +219,11 @@ public abstract class GrokParserStageSequenceBuilder {
 	private static final String PARAMETER_PREFIX = "parm";
 
 	/**
+	 * Used to ignore the value for the matched field.
+	 */
+	private static final String IGNORE_FIELD = "ignore";
+
+	/**
 	 * This function maps {@link SyslogSemanticType} values of type int to fields in the parser
 	 * state.
 	 * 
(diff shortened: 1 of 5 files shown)


features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/GrokParserStageSequenceBuilder.java
```diff
@@ -53,6 +53,7 @@
  * <li>MONTH: 3-character English month abbreviation.</li>
  * <li>NOSPACE: String that contains no whitespace.</li>
  * <li>STRING: String. Because this matches any character, it must be followed by a delimiter in the pattern string.</li>
+ * <li>WHITESPACE: String that contains only whitespace (spaces and/or tabs).</li>
```
j-white: Can we also document this new patter in the admin guide>

features/events/syslog/src/main/resources/org/opennms/netmgt/syslogd/grok-patterns.txt
```diff
@@ -1,6 +1,13 @@
 # Note: Ordering is important in this file. Patterns should be ordered from most specific to least specific. The first
 # pattern found to match will be the one used.
 
+# Patterns for syslog messages parsed incorrectly during OCE audit
```
j-white: Can we just say we're adding additional support for Cisco syslog messages, rather than referring to the "audit"?

features/events/syslog/src/test/java/org/opennms/netmgt/syslogd/ConvertToEventTest.java
```diff
@@ -634,4 +637,56 @@ public void testHostnameLastPattern() {
...
+        assertThat(hasParmValue, is(equalTo(false)));
+    }
+
+    /**
+     * Test that all patterns found from auditing are successfully parsed such that their log message starts with a
+     * given mnemonic.
+     */
+    @Test
+    public void testNewAuditPatterns() {
+        // The following are patterns that had been previously seen to fail when auditing syslog messages for the OCE
```
j-white: Similarly here.

mattixtech:
Appears to be working for both header/body cases now. Just need to test it a bit more, found thousands of messages with failed body parsing.

mattixtech:
This PR should be good to go, I think any changes for parsing the message body will be done in the OCE project.

features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/GrokParserStageSequenceBuilder.java
```diff
@@ -512,6 +526,10 @@
 						factory.stringUntilWhitespace(semanticStringToField(semanticString));
 						factory.whitespace();
 						break;
+					case WHITESPACE:
+						LOG.warn("WHITESPACE pattern followed immediately by another pattern will greedily consume until non-whitespace is encountered");
```
j-white: Is this log necessary? Seems like something that should be in the documentation rather than being logged at run-time.

Since this pattern is actually used in the default configuration these warnings will always be printed at startup when using the RadixTreeParser.

PR #2263, opened 2018-11-28
NMS-10479: Fixed broken In-restriction

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-10479


mvrueden:
Can you provide a test verifiying this?

PR #2264, opened 2018-11-28
HZN-1408: Added example for multi-port listener to config

fooker:
* JIRA: http://issues.opennms.org/browse/HZN-1408

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/telemetryd-configuration.xml
+++ b/opennms-base-assembly/src/main/filtered/etc/telemetryd-configuration.xml
@@ -93,7 +93,7 @@
     </queue>
 
     <!-- Multi-port listener for Netflow v5, Netflow v9, IPFIX and SFlow -->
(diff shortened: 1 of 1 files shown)


opennms-base-assembly/src/main/filtered/etc/telemetryd-configuration.xml
```diff
@@ -93,7 +93,7 @@
     </queue>
 
     <!-- Multi-port listener for Netflow v5, Netflow v9, IPFIX and SFlow -->
-    <listener name="SFlow-UDP-6343" class-name="org.opennms.netmgt.telemetry.listeners.UdpListener" enabled="false">
+    <listener name="Multi-UDP-6343" class-name="org.opennms.netmgt.telemetry.listeners.UdpListener" enabled="false">
```
mvrueden: Port is `9999`, should probably be reflected by the name as well.

PR #2265, opened 2018-11-28
Changed Link on the Admin page, to lead to the wiki

Zottel92:
When the link says "Wiki", it should lead there imho :)


j-white:
Shouldn't we backport this? :)

PR #2266, opened 2018-11-28
NMS-10473: Reduce incoming Clears against other clears

smith-opennms:
This PR changes the behaviour so that Incoming clears are checked against other clears with the if no matching problem alarm exists. Prior to this, multiple clears before the matching problem alarm would result in attempting to add the subsequent clears instead off reducing on the existing and the creates would fail at the DB level with duplicate reduction keys.

* JIRA: http://issues.opennms.org/browse/NMS-10473

### DIFF
--- a/opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
+++ b/opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
@@ -138,11 +138,18 @@ public class AlarmPersisterImpl implements AlarmPersister {
         String key = reductionKey;
         String clearKey = event.getAlarmData().getClearKey();
         
+        boolean didSwapReductionKeyWithClearKey = false;
         if (!m_legacyAlarmState && clearKey != null && isResolutionEvent(event)) {
             key = clearKey;
+            didSwapReductionKeyWithClearKey = true;
         }
 
         OnmsAlarm alarm = m_alarmDao.findByReductionKey(key);
+
+        if (alarm == null && didSwapReductionKeyWithClearKey) {
+            // if the clearKey returns null, still need to check the reductionKey
+            alarm = m_alarmDao.findByReductionKey(reductionKey);
+        }
 
         if (alarm == null || (m_createNewAlarmIfClearedAlarmExists && OnmsSeverity.CLEARED.equals(alarm.getSeverity()))) {
             if (LOG.isDebugEnabled()) {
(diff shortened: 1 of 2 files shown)


opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
```diff
@@ -138,12 +138,19 @@ private OnmsAlarm addOrReduceEventAsAlarm(Event event) throws IllegalStateExcept
         String key = reductionKey;
         String clearKey = event.getAlarmData().getClearKey();
         
+        boolean isClearingEvent = false;
```
j-white: The name `isClearingEvent` is ambiguous since it possible that the event is in fact a clearing event, but this is set to false. I'd propose to use something like `doesReduceOnProblemAlarm` or `didSwapReductionKeyWithClearKey` to make it clearer.

opennms-alarms/daemon/src/test/java/org/opennms/netmgt/alarmd/AlarmdIT.java
```diff
@@ -233,6 +235,34 @@ public void testPersistAlarm() throws Exception {
...
     }
     
+    @Test
+    public void canHandleMultipleClearsWithNoCorrrespondingProblemAlarm() throws Exception {
+        final MockNode node = m_mockNetwork.getNode(1);
+        sendNodeUpEvent(node);
+        sendNodeUpEvent(node);
+
+        // We should only have one alarm now and should have a count of 2
+        await().until(() -> m_alarmDao.findAll().size() == 1 && m_alarmDao.findAll().get(0).getCounter() == 2);
```
j-white: This should be broken out into multiple assertions - if it fails which one didn't match?

opennms-alarms/daemon/src/test/java/org/opennms/netmgt/alarmd/AlarmdIT.java
```diff
@@ -233,6 +235,34 @@ public void testPersistAlarm() throws Exception {
...
+        sendNodeUpEvent(node, firstClearTime);
+        // Send Second Clear
+        sendNodeUpEvent(node, new Date());
+        // Send OutOfOrder Problem with a timestamp 2 minutes before the initial Clear
+        Date initialProblemDate = new Date(firstClearTime.getTime() - SECONDS.toMillis(60));
+        sendNodeDownEvent(node, initialProblemDate);
+
+        // We should have 2 alarms now
+        // The problem should be cleared and the clear should be normal
+        await().until(() -> m_alarmDao.findAll().size() == 2
```
j-white: Similarly here.

PR #2267, opened 2018-11-28
HZN-1431: Remove obsoleted test

smith-opennms:
This PR removes a test that was obsoleted by recent changes.
Alarms are no longer directly removed from the Situation by the REST API.

* JIRA: http://issues.opennms.org/browse/HZN-1431


mattixtech:
👍

PR #2273, opened 2018-11-30
NMS-10483: Make favicons great again

indigo423:
Add favicons which are used in pinned tabs and also in application icons for mobile applications. Add support for iOS, Android and add a tile for Microsoft Windows on desktop.
Example images are attached in the JIRA issue.

* JIRA: http://issues.opennms.org/browse/NMS-10483


indigo423:
@RangerRick @mvrueden do you have an idea about this flapping test here which does not make this branch green? https://bamboo.opennms.org/browse/OPENNMS-ONMS2613-UNIT-5/test/case/230532760

RangerRick:
I think it's been flapping elsewhere too.  Maybe we need to add it to the flappers...

mvrueden:
I tested this on some devices and it looks good 👍

PR #2274, opened 2018-11-30
NMS-10477: Use a utility function instead of Integer.getInteger

patrick-schweizer:
* JIRA: https://issues.opennms.org/browse/NMS-10477


mvrueden:
The name `SystemPropertiesUtils` sounds weird to me. `SystemPropertyUtils` kinda sounds better, or maybe `SystemPropertiesUtil`, however the first one seems a better fit. What do native speakers think of the name?

mvrueden:
Comment from @pbrane , which reached me, but didn't make it to github:

TLDR:
Just waking up so this may make no sense at all :)  More specifically to your question, I agree with your naming preference.


Perhaps, we should not fall into the Utility class rabbit hole.  In my opinion, utility classes are a bit of a code smell.  However, I do understand that we live with them because they are convenient.  A better way might be a compromise between convenience and OO design:

```Java
package org.opennms.core.system.properties

abstract public class Property {
    private final static Logger LOG = LoggerFactory.getLogger(Property.class);

    private static <T> T getT(String propertyName, T defaultValue, Function<String, T> transformer) {
        String valueAsString = System.getProperty(propertyName);
        if (valueAsString == null) {
            return defaultValue;
        }
        try {
            return transformer.apply(valueAsString);
        } catch (NumberFormatException e) {
            String message = String.format("cannot parse system property: %s with value=%s, using default=%s instead."
                    , propertyName
                    , valueAsString
                    , defaultValue);
            LOG.warn(message, e);
        }
        return defaultValue;
    }

    abstract public getProperty(String name);
    ...
}
```

Move the business logic to child classes:

class Integer extends Property {}
class Long extends Property {}


https://en.wikipedia.org/wiki/Single_responsibility_principle


David

core/lib/src/main/java/org/opennms/core/utils/SystemPropertiesUtils.java
```diff
@@ -0,0 +1,90 @@
...
+        Integer result = getInteger(propertyName, null);
+        return (result == null) ? Integer.valueOf(defaultValue) : result;
+    }
+
+    public static Integer getInteger(String propertyName, Integer defaultValue) {
+        Function<String, Integer> resolver = (propertyValue) -> (Integer.parseInt(propertyValue));
+        return getT(propertyName, defaultValue, resolver);
+    }
+
+    private static <T> T getT(String propertyName, T defaultValue, Function<String, T> transformer) {
```
mvrueden: I'd prefer `getProperty` instead.

mvrueden:
Besides naming, this looks good to me.

Adressing @pbrane' comment:
I am no fan of Utility, or static classes/methods in general, either. Mainly because they are painful, when they need to be mocked or re-used somewhere else and as already mentioned are usually a code smell.

However, System Properties are already static by default and adding additional complexity is in my opinion not necessary here, but I have no strong opinion about this.

patrick-schweizer:
In general I agree with @mvrueden assessment. But let's take a step back:
* _What do we want to achieve?_ Detect malformed property values and log them (instead of just falling back to the default value).
* _What are the use cases?_ Replace Long.getLong and Integer.getInteger
* _what are the trade offs?_ Simplicity and code readability vs. single responsibility principle and the possibility of extension
* _will it be easy to move from static to oo/subclasses?_ most likely yes since it will be find and replace in the code

=> if we leave it at the mentioned use cases I think we should go for simple.
if we plan to expand we probable should use oo/subclasses. What could be a use case here? maybe something like:

```
EmailProperty extends Property {
    public getProperty(String propertyName, String defaultValue) {
        Function<String, T> transformer = // ... email validation
        return getT(propertyName, defaultValue, transformer)
    }
}

```

I vote for the easy solution for now.
@mvrueden , @pbrane how shall we proceed?

mvrueden:
@patrick-schweizer, yeah I agree as well. Let's keep the simple solution with static access for now. 

However we should rename the class. 
Maybe just `SystemProperties` without any suffix.

patrick-schweizer:
@mvrueden I changed the name, I believe now we are ready to merge?

mvrueden:
Not quite yet, you forgot to change the name in the checkstyle definitions:
```
<property name="message" value="Please don't use Long.getLong but use SystemPropertiesUtils instead, see there for an explanantion"/>
```

patrick-schweizer:
@mvrueden, I missed that. It's fixed now.

mvrueden:
Thank you. LGTM. If bamboo passes, we can merge this 👍

patrick-schweizer:
Replaced by https://github.com/OpenNMS/opennms/pull/2292
(I messed up the git history on this pull request and thus mad a new one)

PR #2279, opened 2018-12-03
HZN-1433: Update integration api to handle alarm feedback without Kafka

mattixtech:
This PR adds support for passing alarm feedback through the integration API.

* JIRA: http://issues.opennms.org/browse/HZN-1433

### DIFF
--- a/features/api-layer/src/main/java/org/opennms/features/apilayer/dao/AlarmFeedbackDaoImpl.java
+++ b/features/api-layer/src/main/java/org/opennms/features/apilayer/dao/AlarmFeedbackDaoImpl.java
@@ -26,18 +26,35 @@
  *     http://www.opennms.com/
  *******************************************************************************/
 
-package org.opennms.features.situationfeedback.api;
+package org.opennms.features.apilayer.dao;
 
 import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
 
-/**
- * Implementations of this interface react to alarm feedback being generated.
- */
-public interface AlarmFeedbackListener {
-    /**
-     * Handle the newly generated collection of alarm feedback.
-     *
-     * @param alarmFeedback the collection of alarm feedback
-     */
-    void handleAlarmFeedback(List<AlarmFeedback> alarmFeedback);
+import org.opennms.features.apilayer.utils.ModelMappers;
+import org.opennms.features.situationfeedback.api.FeedbackException;
+import org.opennms.features.situationfeedback.api.FeedbackRepository;
+import org.opennms.integration.api.v1.dao.AlarmFeedbackDao;
+import org.opennms.integration.api.v1.model.AlarmFeedback;
+
+public class AlarmFeedbackDaoImpl implements AlarmFeedbackDao {
+
+    private final FeedbackRepository feedbackRepository;
+
+    public AlarmFeedbackDaoImpl(FeedbackRepository feedbackRepository) {
+        this.feedbackRepository = Objects.requireNonNull(feedbackRepository);
+    }
+
+    @Override
+    public List<AlarmFeedback> getFeedback() {
+        try {
+            return feedbackRepository.getAllFeedback()
+                    .stream()
+                    .map(ModelMappers::toFeedback)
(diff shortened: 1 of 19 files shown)


features/api-layer/src/main/java/org/opennms/features/apilayer/dao/AlarmFeedbackDaoImpl.java
```diff
@@ -0,0 +1,148 @@
...
+public class AlarmFeedbackDaoImpl implements AlarmFeedbackDao {
+    private static final Logger LOG = LoggerFactory.getLogger(AlarmFeedbackDaoImpl.class);
+
+    /**
+     * Used to look up a reference to a {@link FeedbackRepository feedback repository} at runtime.
+     */
+    @SuppressWarnings("unchecked")
+    private final ServiceLookup<Class<?>, String> SERVICE_LOOKUP =
+            new ServiceLookupBuilder(new ServiceRegistryLookup(DefaultServiceRegistry.INSTANCE))
+                    .blocking(5000, 1000, 1000)
```
j-white: Can we make these constants configurable?

features/api-layer/src/main/java/org/opennms/features/apilayer/dao/AlarmFeedbackDaoImpl.java
```diff
@@ -0,0 +1,148 @@
...
+                .map(ModelMappers::fromFeedback)
+                .collect(Collectors.toList());
+
+        try {
+            feedbackRepository.persist(mappedFeedback);
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to execute query: " + e.getMessage(), e);
+        }
+
+        alarmFeedbackListeners.forEach(listener -> {
```
j-white: Can we move the logic of invoking the listeners into the repository rather than duplicating it here? As is stand, everyone that wants to use the repository is now responsible for also managing and calling the listeners.

features/situation-feedback/api/pom.xml
```diff
@@ -32,5 +32,10 @@
       <artifactId>guava</artifactId>
       <version>${guavaVersion}</version>
     </dependency>
+      <dependency>
```
j-white: Do we need to update the feature definition for this as well?

features/situation-feedback/elastic/src/main/java/org/opennms/features/situationfeedback/elastic/FeedbackDocument.java
```diff
@@ -128,7 +128,7 @@ public void setUser(String user) {
 
     public static FeedbackDocument from(AlarmFeedback feedback) {
         FeedbackDocument doc = new FeedbackDocument();
-        doc.setTimestamp(System.currentTimeMillis());
+        doc.setTimestamp(feedback.getTimestamp());
```
j-white: Should we fallback to using `System.currentTimeMillis()` if the timestamp is `null`?

j-white:
`opennms-api-layer` feature fails with `Unable to resolve opennms-api-layer/24.0.0.SNAPSHOT: missing requirement [opennms-api-layer/24.0.0.SNAPSHOT]` when loading.

mattixtech:
> `opennms-api-layer` feature fails with `Unable to resolve opennms-api-layer/24.0.0.SNAPSHOT: missing requirement [opennms-api-layer/24.0.0.SNAPSHOT]` when loading.

@j-white Is it failing because it is depending on something in health check? I had that happen and then it went away, I was going to add a new dependency to the feature but I figured it was just something screwed up with my install.

features/situation-feedback/elastic/src/main/java/org/opennms/features/situationfeedback/elastic/FeedbackDocument.java
mattixtech: Current behaviour is that you can't construct an `AlarmFeedback` with a null timestamp. If no timestamp is explicitly provided in the builder, it already defaults to using the time the builder was created.

mattixtech:
Working on test failure.

PR #2282, opened 2018-12-07
OIA-1: Map service pollers from integration-api to default pollers.

cgorantla:
* JIRA: http://issues.opennms.org/browse/OIA-1


features/api-layer/src/main/java/org/opennms/features/apilayer/pollers/PollerRequestBuilderImpl.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    @Override
+    public PollerRequestBuilder withTimeToLive(Long ttlInMs) {
+        this.ttlInMs = ttlInMs;
+        return this;
+    }
+
+    @Override
+    public CompletableFuture<PollerResult> execute() {
```
mattixtech: Can the `execute()` method be moved out of the request and into the client? I think that would be better encapsulation and we could avoid having to pass in an instance of the client to the request.

The request builder could then become an actual builder that returns a request object which is accepted as a parameter in the `execute()` method that we have moved (or otherwise if you have a better idea). This will involve updating the api interfaces.

features/api-layer/src/main/java/org/opennms/features/apilayer/pollers/PollerRequestBuilderImpl.java
```diff
@@ -0,0 +1,178 @@
...
+        Map<String, Object> props = new HashMap<>();
```
mattixtech:
```suggestion
        Map<String, Object> props = Collections.unmodifiableMap(attributes);
```

features/api-layer/src/main/java/org/opennms/features/apilayer/pollers/PollerRequestBuilderImpl.java
```diff
@@ -0,0 +1,178 @@
...
+        attributes.forEach(props::put);
```
mattixtech:
```suggestion
```

features/api-layer/src/main/java/org/opennms/features/apilayer/pollers/PollerRequestBuilderImpl.java
```diff
@@ -0,0 +1,178 @@
...
+                .withTimeToLive(ttlInMs)
+                .withAttributes(props)
+                .execute();
+        // convert the response to PollResult.
+        CompletableFuture<PollerResult> result = new CompletableFuture<>();
+        PollStatus pollStatus;
+        try {
+            PollerResponse pollerResponse = future.get();
+            pollStatus = pollerResponse.getPollStatus();
+            switch (Status.valueOf(pollStatus.getStatusName())) {
```
mattixtech: Move this logic in the switch statement into the PollerResultImpl class.

Then you can just do:
```result.complete(new PollerResultImpl(pollStatus));```

Could also use this simplified constructor:
```
private PollerResultImpl(PollStatus pollStatus) {
            status = Enums.getIfPresent(Status.class, pollStatus.getStatusName()).or(Status.Unknown);
            if(status == Status.Up) {
                properties = pollStatus.getProperties();
            } else {
                reason = pollStatus.getReason();
            }
        }
```

File: features/api-layer/src/main/java/org/opennms/features/apilayer/pollers/PollerRequestBuilderImpl.java [hunk 5]
```diff
@@ -0,0 +1,178 @@
...
+                default:
+                    result.complete(new PollerResultImpl(Status.Unknown, pollStatus.getReason()));
+            }
+        } catch (InterruptedException | ExecutionException e) {
+           result.complete(new PollerResultImpl(Status.Unknown, e.getMessage()));
+        }
+        return result;
+    }
+
+    private class PollerResultImpl implements PollerResult {
```
mattixtech: This can move to the client as well.

features/api-layer/src/main/java/org/opennms/features/apilayer/pollers/PollerRequestBuilderImpl.java
cgorantla: yeah, here we are using builder for building and executing as well.  That allows us to do 
`poll().withService().withClassName()...execute()`.    By separating builder we need to build first and then executes the request.  Also it needs to know actual instance of the builder not just the interface.   I guess the current one makes it easier for the clients to call without actually bothering about the implementation.   This follows the pattern that we used for all `LocationAware` builders.

features/api-layer/src/main/java/org/opennms/features/apilayer/pollers/PollerRequestBuilderImpl.java
cgorantla: Thanks for this. really made easier.

features/api-layer/src/main/java/org/opennms/features/apilayer/pollers/PollerRequestBuilderImpl.java
mattixtech: You could still maintain the same call signature if you changed the `execute()` inside the builder to just be a proxy for the client that built it's `execute()`.

So the builder would then take a reference to the `ServicePollerClient` that created it on construction rather than a reference to a `LocationAwarePollerClient`. Inside its execute it would simply pass itself to the execute method of the `ServicePollerClient`. This would move all the execution logic inside the client class which seems like better encapsulation to me.

Perhaps an even better solution would be to move the builder class into the client class as a nested class and then you could avoid passing the reference to the client altogether.

Feel free to disagree as I'm not very familiar with this code.

Example change to `ServicePollerClientImpl`:
```
   @Override
    public PollerRequestBuilder poll() {
        return new PollerRequestBuilderImpl(this);
    }

    CompletableFuture<PollerResult> execute(PollerRequestBuilder builder) {
        ... existing execute logic
    }
```

Example change to `PollerRequestBuilderImpl `: 
```
    public PollerRequestBuilderImpl(ServicePollerClient pollerClient) {
        this.pollerClient = pollerClient;
    }

    @Override
    public CompletableFuture<PollerResult> execute() {
        return pollerClient.execute(this);
    }
```

mattixtech: @cgorantla Not sure if you are still looking at this, but feel free to ignore this suggestion as I went off on a bit of a tangent here, the original impl is fine especially if it is attempting to remain consistent with the existing example.

PR #2284, opened 2018-12-08
HZN-956: Implement browser notification

fooker:
* JIRA: http://issues.opennms.org/browse/HZN-956


fooker:
@indigo423 suggested to use a more modern/flat logo instead of https://github.com/OpenNMS/opennms/blob/5f4ae13ccef1ff185c49595e41ca15449283b6fb/opennms-webapp/src/main/webapp/images/o-192.png. Thoughts?

indigo423:
> @indigo423 suggested to use a more modern/flat logo instead of https://github.com/OpenNMS/opennms/blob/5f4ae13ccef1ff185c49595e41ca15449283b6fb/opennms-webapp/src/main/webapp/images/o-192.png. Thoughts?

Up so it uses the new o-logo

opennms-webapp/src/main/java/org/opennms/web/notification/NotificationStreamServlet.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        @Override
+        public void onWebSocketError(final Throwable cause) {
+            super.onWebSocketError(cause);
+            BrowserNotificationDispatcher.getInstance().unsubscribe(this.handler);
+        }
+
+        private void sendNotification(final BrowserNotificationMessage message) {
+            // TODO fooker: Send some JSON
```
mvrueden: TODO fooker?
https://chat.opennms.com/api/v4/emoji/pw3zdt9o8b8izke1qe3mzrhtfw/image"

fooker:
Mentioned in whatsnew

opennms-webapp/src/main/webapp/includes/bootstrap-footer.jsp
```diff
@@ -80,6 +81,45 @@
                 %>
             </p>
         </footer>
+
+        <c:if test="${!empty pageContext.request.remoteUser}">
```
j-white: I'm wondering if there's a better place to put this logic - seems out of place to include this Javascript in the JSP of the footer.

Can we find a way to move this to a .js file instead?

indigo423:
I did a functional test and got it to work with Safari, Chrome and Firefox. 👍 from me.

pbrane:
I tested as well.  Works for me w/Safari.  How difficult would it be to make the notification a clickable reference back to the notification in the browser such that the user could ack the notification?

pbrane:
Pretty cool! 
image

indigo423:
> I tested as well. Works for me w/Safari. How difficult would it be to make the notification a clickable reference back to the notification in the browser such that the user could ack the notification?

I tried this but <a refs=...> but they get encoded and are not interpreted as HTML links. I can imagine the base href can be unknown to opennms itself, e.g. behind proxy aka the opennms jetty property `opennms.web.base-url = https://%x%c/`

pbrane:
Bummer.

fooker:
> How difficult would it be to make the notification a clickable reference back to the notification in the browser such that the user could ack the notification?

I've already looked into it. There is some support for "actions". But it's still beta and not supported by most stable browsers. It also requires HTTPS and a lot of changes to the authentication handling. Maybe later...

opennms-doc/guide-admin/src/asciidoc/text/notifications/introduction.adoc
```diff
@@ -8,6 +8,7 @@
 _{opennms-product-name}_ uses notifications to make users aware of an event.
 Common notification methods are email and paging, but notification mechanisms also exist for:
 
+* Browser base desktop notifications
```
christianpape: Browser base -> Browser-based

opennms-webapp/src/main/webapp/includes/notifications.jsp
```diff
@@ -0,0 +1,77 @@
...
+
+<%@page language="java"
+        contentType="text/html"
+        session="true"
+        import="org.opennms.web.api.Util"
+%>
+
+<%@ taglib uri="http://java.sun.com/jsp/jstl/core" prefix="c" %>
+
+<c:if test="${!empty pageContext.request.remoteUser}">
```
j-white: We've moved this out of the JSP, but we're still depending on JSP rendered Javascript code that appended to all of the footers. Can we move this to a static Javsascript asset instead?

fooker: Migrated

j-white:
Awesome. Thanks.

indigo423:
@fooker what does means your commit message:

> HZN-956: Migrated js to asserts

😄

PR #2286, opened 2018-12-09
NMS-10497: Use application instead of ports in convo_key

fooker:
* JIRA: http://issues.opennms.org/browse/NMS-10497


fooker:
As this changes the schema of the elasticsearch index, we should provide some method to update existing data. Should be possible to do with a update query. But I don't know where to put these...

fooker:
Added update-script to whatsnew

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/ElasticFlowRepository.java
```diff
@@ -529,46 +525,6 @@ private static void toTable(ImmutableTable.Builder<Directional<String>, Long, Do
                 .thenCompose((topN) -> getTotalBytesFromTopN(topN, groupByTerm, keyForMissingTerm, includeOther, filters));
     }
 
-    /**
-     * Perform a best-effort classification of the application based on
```
j-white: Happy to see this go :)

indigo423:
@fooker we should add a warning on your curl command, this thing burned my CPUs and filled all my writes per second I had for my disks for a pretty long time :)

fooker:
@indigo423 can you provide `web.log` while executing the request?

fooker:
Comment about high stress added to update command

j-white:
We need to figure out the source of the NPE before we merge this. We may need to wait until https://issues.opennms.org/browse/HZN-1458 is fixed to figure it out though.

j-white:
I merged the latest develop into this branch, which means we now have fixes for both HZN-1458 (logging the NPEs if they happen) and NMS-10530 (fixing some NPEs which were recently discovered).

@indigo423 can you give this another whirl once the RPMs are ready? If there are any exceptions related to the REST API they will be logged to karaf.log.

indigo423:
> I merged the latest develop into this branch, which means we now have fixes for both HZN-1458 (logging the NPEs if they happen) and NMS-10530 (fixing some NPEs which were recently discovered).
> 
> @indigo423 can you give this another whirl once the RPMs are ready? If there are any exceptions related to the REST API they will be logged to karaf.log.

on my way

indigo423:
Here some results I got so far during testing:

Newly received sFlow and Netflow v9 packets with the running branch NMS-10497
* [sFlow JSON document](https://gist.github.com/indigo423/231ce30e43af0d5813e7b42d3dd928c0) using NMS-10497
* [Netflow v9 document](https://gist.github.com/indigo423/7c6b5030d258a521353d31843ad5cdea) using NMS-10497*
* [Elasticsearch call](https://gist.github.com/indigo423/14bb855e99f44e9c771a577f98cd930a) logged in karaf.log

I don't get any exceptions anymore in using the deep dive tool. I've noticed in the "Top N panel Traffic by Conversation" the application column is empty whereas the "Traffic by Application" show traffic by applications.

nms-10497

j-white:
Seems like this particular feature is OK then - missing applications are probably due to the fact that the top conversations were not tagged with any.

fooker:
@indigo423 did your test also include flows converted from the old format to the new one?

indigo423:
> @indigo423 did your test also include flows converted from the old format to the new one?

Yes I've ran your script which was mentioned here: https://docs.opennms.org/opennms/branches/develop/releasenotes/releasenotes.html#_breaking_changes

PR #2287, opened 2018-12-12
NMS-10337: Replaced jdhcp with dhcp4java

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-10337


j-white:
👍 

Can someone try and give this a spin?

PR #2288, opened 2018-12-12
NMS-10505: Add opennms logging configuration for karaf containers

cgorantla:
* JIRA: http://issues.opennms.org/browse/NMS-10505


features/container/minion/src/main/filtered-resources/etc/org.ops4j.pax.logging.cfg
```diff
@@ -0,0 +1,98 @@
...
+log4j2.logger.spifly.name = org.apache.aries.spifly
+log4j2.logger.spifly.level = WARN
+
+# Security audit logger
+log4j2.logger.audit.name = org.apache.karaf.jaas.modules.audit
+log4j2.logger.audit.level = INFO
+log4j2.logger.audit.additivity = false
+log4j2.logger.audit.appenderRef.AuditRollingFile.ref = AuditRollingFile
+
+# OPENNMS: Display all DEBUG logs for our code
```
cgorantla: Added these two lines on default  configuration file that comes out of container.

j-white:
👍

features/container/minion/src/main/filtered-resources/etc/org.ops4j.pax.logging.cfg
```diff
@@ -0,0 +1,98 @@
...
+
+# Security audit logger
+log4j2.logger.audit.name = org.apache.karaf.jaas.modules.audit
+log4j2.logger.audit.level = INFO
+log4j2.logger.audit.additivity = false
+log4j2.logger.audit.appenderRef.AuditRollingFile.ref = AuditRollingFile
+
+# OPENNMS: Display all INFO logs for OpenNMS code
+log4j2.logger.opennms.name = org.opennms
+log4j2.logger.opennms.level = INFO
```
j-white: If the root level is already set to INFO, is there any need to set `org.opennms` explicitly to INFO?

pbrane: Can we leave it in but commented so that we know where to change it?

cgorantla: On OpenNMS, we already have [opennms](https://github.com/OpenNMS/opennms/blob/743e2a95fd348c7da1eac9a757e4a6b2694dd794/container/karaf/src/main/filtered-resources/etc/org.ops4j.pax.logging.cfg#L53) logging level to DEBUG, we could follow the same here on Minion.

j-white: We shouldn't leave the OpenNMS one set to DEBUG for release builds.. but that's a different problem.

j-white: I guess the goal here is to add an entry so that it's easy to change the log level and set it to DEBUG then?

j-white:
We could move this into `opennms/container/shared/src/main/filtered-resources/etc/` so that we can reuse those settings for both Minion and Sentinel.

PR #2289, opened 2018-12-14
NMS-10495: Introduced 'default' attribute for settings fall-back values in event translator assignments

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-10495

### DIFF
--- a/opennms-config-model/src/main/resources/xsds/translator-configuration.xsd
+++ b/opennms-config-model/src/main/resources/xsds/translator-configuration.xsd
@@ -35,6 +35,7 @@
 				</sequence>
 				<attribute name="type" type="string" use="required" />
 				<attribute name="name" type="string" use="required" />
+				<attribute name="default" type="string" use="optional" />
 			</complexType>
 		</element>
 
--- a/opennms-services/src/main/java/org/opennms/netmgt/translator/EventTranslator.java
+++ b/opennms-services/src/main/java/org/opennms/netmgt/translator/EventTranslator.java
@@ -156,15 +156,15 @@ public class EventTranslator extends AbstractServiceDaemon implements EventListe
             return;
         }
 
-        if (!m_config.isTranslationEvent(e)) {
-            LOG.debug("onEvent: received event that matches no translations: \n", EventUtils.toString(e));
-            return;
-        }
-
-        LOG.debug("onEvent: received valid registered translation event: \n", EventUtils.toString(e));
-
         List<Event> translated = m_config.translateEvent(e);
         if (translated != null) {
+            if (translated.isEmpty()) {
+                LOG.debug("onEvent: received event that matches no translations: \n", EventUtils.toString(e));
+                return;
+            }
+
+            LOG.debug("onEvent: received valid registered translation event: \n", EventUtils.toString(e));
+
             Log log = new Log();
             Events events = new Events();
             for (Iterator<Event> iter = translated.iterator(); iter.hasNext();) {
--- a/opennms-services/src/main/java/org/opennms/netmgt/translator/EventTranslator.java
+++ b/opennms-services/src/main/java/org/opennms/netmgt/translator/EventTranslator.java
@@ -145,6 +145,7 @@ public class EventTranslator extends AbstractServiceDaemon implements EventListe
     /** {@inheritDoc} */
     @Override
     public void onEvent(Event e) {
+        final String eventString = LOG.isDebugEnabled() ? EventUtils.toString(e) : e.getUuid();
 
         if (isReloadConfigEvent(e)) {
             handleReloadEvent(e);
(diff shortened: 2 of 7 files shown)


opennms-services/src/main/java/org/opennms/netmgt/translator/EventTranslator.java
```diff
@@ -156,15 +156,15 @@ public void onEvent(Event e) {
...
-            LOG.debug("onEvent: received event that matches no translations: \n", EventUtils.toString(e));
-            return;
-        }
-
-        LOG.debug("onEvent: received valid registered translation event: \n", EventUtils.toString(e));
-
         List<Event> translated = m_config.translateEvent(e);
         if (translated != null) {
+            if (translated.isEmpty()) {
+                LOG.debug("onEvent: received event that matches no translations: \n", EventUtils.toString(e));
```
j-white: There's no `{}` placeholder to show the event in this message. We should also avoid calling `EventUtils.toString()` unless DEBUG logging is enabled.

(I realize this was existing code, but we may as well change it while we're here.)

opennms-services/src/main/java/org/opennms/netmgt/translator/EventTranslator.java
```diff
@@ -156,15 +156,15 @@ public void onEvent(Event e) {
...
+                return;
+            }
+
+            LOG.debug("onEvent: received valid registered translation event: \n", EventUtils.toString(e));
```
j-white: Similarly here.

opennms-config-model/src/main/resources/xsds/translator-configuration.xsd
```diff
@@ -35,6 +35,7 @@
 				</sequence>
 				<attribute name="type" type="string" use="required" />
 				<attribute name="name" type="string" use="required" />
+				<attribute name="default" type="string" use="optional" />
```
pbrane: Can we change this to:
```XML
<attribute name="default-value" type="string" use="optional" default="Unknown" />
```

Then we don't have to change the XML configuration file.

j-white: Do we really want all of the assignments to be optional by default now?

j-white: A default value doesn't fit the logic.

pbrane: I just looked at it closer.  You're right.  It really doesn't.  I guess I really hate the original design of this schema. :)  The "type" attribute is the killer.  #guilty

pbrane:
:)

opennms-services/src/main/java/org/opennms/netmgt/translator/EventTranslator.java
```diff
@@ -145,6 +145,7 @@ private void createMessageSelectorAndSubscribe() {
     /** {@inheritDoc} */
     @Override
     public void onEvent(Event e) {
+        final String eventString = LOG.isDebugEnabled() ? EventUtils.toString(e) : e.getUuid();
```
j-white: Storing this in a variable seems a little misleading, it may be the event string or may be a uuid.

opennms-services/src/main/java/org/opennms/netmgt/translator/EventTranslator.java
```diff
@@ -156,21 +157,21 @@ public void onEvent(Event e) {
...
-        if (!m_config.isTranslationEvent(e)) {
-            LOG.debug("onEvent: received event that matches no translations: \n", EventUtils.toString(e));
-            return;
-        }
-
-        LOG.debug("onEvent: received valid registered translation event: \n", EventUtils.toString(e));
-
         List<Event> translated = m_config.translateEvent(e);
         if (translated != null) {
+            if (translated.isEmpty() && LOG.isDebugEnabled()) {
```
j-white: `LOG.isDebugEnabled()` is not necessary here.

opennms-services/src/main/java/org/opennms/netmgt/translator/EventTranslator.java
```diff
@@ -156,21 +157,21 @@ public void onEvent(Event e) {
...
+                LOG.debug("onEvent: received event that matches no translations: {}", eventString);
+                return;
+            }
+            if (LOG.isDebugEnabled())
```
j-white: `LOG.isDebugEnabled()` is also not necessary here.

christianpape:
I just checked the same trap against a current develop branch OpenNMS instance and 
the behavior is the same. I think this has never worked, because the event parameters

`  .1.3.6.1.2.1.2.2.1.1 = Int32(text): 2`
`  .1.3.6.1.2.1.2.2.1.7 = Int32(text): 1`
`  .1.3.6.1.2.1.2.2.1.8 = Int32(text): 1`

will never match the configuration's regular expressions, like

`  <value type="parameter" name="~^\.1\.3\.6\.1\.2\.1\.2\.2\.1\.1\.([0-9]*)$" matches=".*" result="${0}" />`

That's because the event parameters represent a single value and not a table and the 
configuration's regular expression matches a table:

`    .1.3.6.1.2.1.2.2.1.1`
    
will never match

`  ~^.1.3.6.1.2.1.2.2.1.1.([0-9]*)$`

because of the missing last index.

j-white:
Looks like I was using the wrong `snmptrap` command. Testing with variations on `snmptrap -v2c -c public localhost "" .1.3.6.1.6.3.1.1.5.3   ifIndex.1 i 1   ifAdminStatus.1 i 1   ifOperStatus.1 i 1` was successful.

PR #2290, opened 2018-12-14
HZN-1442: Update the SnmpCollector to use the CollectionSetBuilder

patrick-schweizer:
This PR s not ready to be merged but a work in progress refactoring. I made it a PR to be able to discuss things easier.

* JIRA: https://issues.opennms.org/browse/HZN-1442


opennms-services/src/main/java/org/opennms/netmgt/collectd/SnmpCollectionSetRefactorDummy.java
```diff
@@ -0,0 +1,105 @@
...
+    SnmpCollectionSetRefactorDummy(SnmpCollectionAgent agent, OnmsSnmpCollection snmpCollection, LocationAwareSnmpClient client) {
+        super(agent,  snmpCollection, client);
+    }
+
+    CollectionSet createCollectionSet() {
+        CollectionSetBuilder builder = new CollectionSetBuilder(getAgent());
+
+
+        // TODO Patrick Very ugly mapping until refactoring is done. We convert here the snmp world into the CollectionSetBuilder world
+        for (CollectionResource collectionResource : getResources()) {
```
patrick-schweizer: @j-white do these mappings make sense to you?

opennms-services/src/main/java/org/opennms/netmgt/collectd/SnmpCollectionSetRefactorDummy.java
```diff
@@ -0,0 +1,105 @@
...
+
+            if(collectionResource instanceof GenericIndexResource) {
+                GenericIndexResource genericResource = (GenericIndexResource) collectionResource;
+                org.opennms.netmgt.config.datacollection.ResourceType resourceType = new org.opennms.netmgt.config.datacollection.ResourceType();
+                resourceType.setName(genericResource.getResourceTypeName());
+                resourceType.setLabel(genericResource.getResourceTypeName());
+                // TODO Patrick what shall we set here? resourceType.setPersistenceSelectorStrategy();
+                // TODO Patrick what shall we set here?  resourceType.setResourceLabel();
+                // TODO Patrick what shall we set here?  resourceType.setStorageStrategy(nodeInfo.getResourceType().);
+                GenericTypeResource resource = new GenericTypeResource(new NodeLevelResource(genericResource.getCollectionAgent().getNodeId()),resourceType,genericResource.getInstance());
```
j-white: Consider using the `DeferredGenericTypeResource` here if you don't have the actual type definition available.

opennms-services/src/main/java/org/opennms/netmgt/collectd/SnmpCollectionSetRefactorDummy.java
```diff
@@ -0,0 +1,105 @@
...
+
+            if(collectionResource instanceof IfInfo) {
+                IfInfo ifInfo = (IfInfo) collectionResource;
+                InterfaceLevelResource resource = new InterfaceLevelResource(new NodeLevelResource(ifInfo.getNodeId()), ifInfo.getAttributesMap().get("snmpifname"));
+                addGroupsToBuilder(builder, resource, ifInfo.getGroups());
+            }
+
+            if(collectionResource instanceof AliasedResource) {
+                AliasedResource aliasedResource = (AliasedResource) collectionResource;
+                Resource resource = null; // TODO: Patrick there is no AliasedResource in the CollectionSetBuilder world, what shall we do?
```
j-white: If the same behavior can't be achieved with one of the existing types, we'll need to add a new type.

opennms-services/src/main/java/org/opennms/netmgt/collectd/SnmpCollectionSetRefactorDummy.java
patrick-schweizer: :+1:

patrick-schweizer: Sounds good. I tried but then it fails in DeferredGenericTypeResource.resolve since there is no ResourceTypeMapper...

opennms-services/src/test/java/org/opennms/netmgt/collectd/SnmpCollectorMinMaxValIT.java
```diff
@@ -238,9 +240,10 @@ public void testPersist() throws Exception {
 
         // by now the values should be the new values
         assertEquals(Double.valueOf(456.0), m_rrdStrategy.fetchLastValueInRange(rrdFile.getAbsolutePath(), "tcpCurrEstab", stepSizeInMillis, rangeSizeInMillis));
+        // TODO: Patrick: is this correct?
```
patrick-schweizer: @j-white I am not sure that this test is correct? We set the value 7654321 in line 235. But in line 242 we test for  the old value _1234567.0_

I think the reason why this test passed with the old implementation is that the value was never set:
1. Block value = 1234567
1. Block value = 7654321 I don't think this gets set, it seems that we can only increment by one
1. Block value = 1234567 
It could also be that I am not clear on what we are testing here but it seems strange to me. I did a bit more investigation and wrote the following tests:
https://github.com/OpenNMS/opennms/pull/2290/files#diff-a64257aa8c4f16de0a23885e4a4fd2c3R274
https://github.com/OpenNMS/opennms/pull/2290/files#diff-f2c4541f8f65108cc68347ab00d15adeR273

It looks like that we can set any value in the new Implementation but only increment by one in the old implementation. Is the old behavior the desired one? Could you help me to understand what is going on? Thanks!

j-white: We figure this out. The test enforces a max value which explains the observed behavior.

j-white:
👍 I've gone ahead and updated the docs with a little more context around the `SnmpCollectorNG`. I also gave it a spin, and it appears to functioning properly.

Given that there is no impact on the current functionality of the `SnmpCollector`, I think we're good to merge this, unless there are any other comments.

PR #2292, opened 2018-12-15
NMS-10477: Use a utility function instead of Integer.getInteger

patrick-schweizer:
* JIRA: https://issues.opennms.org/browse/NMS-10477

Replaces: https://github.com/OpenNMS/opennms/pull/2274

Our [continuous integration system](http://bamboo.internal.opennms.com:8085) will test and verify your changes.

PR #2295, opened 2018-12-17
Revert "NMS-10477: Use a utility function instead of Integer.getInteger"

mvrueden:
Reverts OpenNMS/opennms#2292


mvrueden:
Reverting the changes of the PR as in theory they should not have any negative affect, but in general those refactorings should not be made in a foundation branch. Please retarget to `develop`.

PR #2298, opened 2018-12-18
HZN-1451: Prevent CME during OSGI service binding

mattixtech:
I noticed a few instances of OSGI service binding where we can be vulnerable to a CME due to either explicitly iterating a non-thread-safe collection or exposing a reference to a non-thread-safe collection that a client could then iterate.

The chance of one of these actually generating a CME is probably near zero but still possible so might as well make it correct.

* JIRA: http://issues.opennms.org/browse/HZN-1451


mattixtech:
Going to retry the test run, it passes locally.

PR #2299, opened 2018-12-18
HZN-1419: Add support for prioritizing Event Definitions

smith-opennms:
This PR add support for prioritizing event definitions.
This will be useful if one wants to override an event definition that is loaded from the registry - i.e. it is brought in by a plugin feature.

* JIRA: https://issues.opennms.org/browse/HZN-1419

### DIFF
--- a/opennms-config-model/src/main/java/org/opennms/netmgt/xml/eventconf/Events.java
+++ b/opennms-config-model/src/main/java/org/opennms/netmgt/xml/eventconf/Events.java
@@ -371,8 +371,7 @@ public class Events implements Serializable {
         indexEventsByUei();
     }
 
-    // Recurse through the configuration and return Event Definitions with
-    // priority > 0
+    // Recurse through the configuration and return Event Definitions with priority > 0
     private List<Event> getPrioritizedEvents() {
         List<Event> prioritizedEvents = new ArrayList<Event>();
         prioritizedEvents.addAll(m_events.stream().filter(e -> e.getPriority() > 0).collect(Collectors.toList()));
--- a/opennms-config-model/src/main/java/org/opennms/netmgt/xml/eventconf/Events.java
+++ b/opennms-config-model/src/main/java/org/opennms/netmgt/xml/eventconf/Events.java
@@ -51,6 +51,7 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
 
 import javax.xml.bind.annotation.XmlAccessType;
 import javax.xml.bind.annotation.XmlAccessorType;
@@ -354,6 +355,10 @@ public class Events implements Serializable {
             event.initialize(m_ordering.next());
         }
 
+        // roll up all prioritized events and sort all events by priority
+        m_events.addAll(getPrioritizedEvents());
+        m_events.sort(Comparator.reverseOrder());
+
         partitionEvents(partition);
 
         for (final Events events : m_loadedEventFiles.values()) {
@@ -361,6 +366,17 @@ public class Events implements Serializable {
         }
 
         indexEventsByUei();
+    }
+
+    // Recurse through the configuration and return Event Definitions with
+    // priority > 0
+    private Collection<Event> getPrioritizedEvents() {
+        List<Event> prioritizedEvents = new ArrayList<Event>();
+        prioritizedEvents.addAll(m_events.stream().filter(e -> e.getPriority() > 0).collect(Collectors.toList()));
+        for (final Events eventsFile : m_loadedEventFiles.values()) {
+            prioritizedEvents.addAll(eventsFile.getPrioritizedEvents());
+        }
+        return prioritizedEvents;
     }
 
     private void indexEventsByUei() {
@@ -377,17 +393,8 @@ public class Events implements Serializable {
             }
 
             if (m_eventsByUei.putIfAbsent(uei, e) != null) {
-                // Keep track of the UEIs that have many unprioritized event definitions
-                int existingPriority = m_eventsByUei.get(uei).getPriority();
-                int currentPriority = e.getPriority();
-                if (existingPriority == 0 && currentPriority == 0) {
-                    ueisWithManyEventDefinitions.add(uei);
-                } else {
-                    // use the event definition with the highest priority
-                    if (currentPriority > existingPriority) {
-                        m_eventsByUei.put(uei, e);
-                    }
-                }
+                // Keep trap of the UEIs that have many event definitions
+                ueisWithManyEventDefinitions.add(uei);
             }
         });
 
--- a/opennms-config-model/src/main/java/org/opennms/netmgt/xml/eventconf/Events.java
+++ b/opennms-config-model/src/main/java/org/opennms/netmgt/xml/eventconf/Events.java
@@ -39,6 +39,7 @@ import java.io.Writer;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -265,9 +266,11 @@ public class Events implements Serializable {
                         m_partitionedEvents.put(key, events);
                     }
                     events.add(event);
+                    events.sort(Comparator.reverseOrder());
                 }
             }
         }
+        m_nullPartitionedEvents.sort(Comparator.reverseOrder());
     }
 
 
@@ -293,14 +296,13 @@ public class Events implements Serializable {
             }
         }
 
-        for(final Event event : potentialMatches) {
+        for (final Event event : potentialMatches) {
             if (event.matches(matchingEvent)) {
                 return event;
             }
         }
 
-        for(Entry<String, Events> loadedEvents : m_loadedEventFiles.entrySet()) {
-            final Events subEvents = loadedEvents.getValue();
+        for (Events subEvents : m_loadedEventFiles.values()) {
             final Event event = subEvents.findFirstMatchingEvent(matchingEvent);
             if (event != null) {
                 return event;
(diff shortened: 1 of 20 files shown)


opennms-config-model/src/main/resources/xsds/eventconf.xsd
```diff
@@ -70,6 +70,12 @@
           </annotation>
         </element>
 
+        <element maxOccurs="1" minOccurs="0" name="priority" type="int">
```
pbrane: 1. Should we have a default value?  (such as 100 as suggested in the issue)
2. Maybe make the element name more clear as to what the priority element does?
```XSD
<element maxOccurs="1" minOccurs="0" name="marshaling-priority" type="int" default="100">
```

pbrane: Perhaps better than positiveInteger would be a restriction of 1 - 100.  I'm no expert, so just a guess at this:
```XSD
<element maxOccurs="1" minOccurs="0" name="marshaling-priority" default="100">
  <simpleType>
    <restriction base="positiveInteger">
      <minInclusive value="1"/>
      <maxInclusive value="100"/>
    </restriction>
  </simpleType>
</element>
```

opennms-config-model/src/main/resources/xsds/eventconf.xsd
```diff
@@ -70,6 +70,12 @@
...
+          <annotation>
+            <documentation>The Priority of the Event Definition.</documentation>
```
pbrane: We should specify if priority 1 is low or high or what the range is that is allowed if not specified in a restriction.

opennms-config/src/main/java/org/opennms/netmgt/config/DefaultEventConfDao.java
```diff
@@ -164,7 +164,8 @@ public void saveCurrent() {
 				accum.add(event);
 				return accum;
 			}
-		});
+            // remove duplicates
```
j-white: Can we elaborate on why we remove the duplicates?

opennms-config/src/test/resources/org/opennms/netmgt/config/eventd/reloaded/BGP4.events.xml
```diff
@@ -15,6 +15,7 @@
    <mevalue>1</mevalue>
   </maskelement>
  </mask>
+ <priority>20</priority>
```
j-white: Are these changes necessary?

opennms-config/src/test/resources/priority/eventconf2.xml
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0"?>
```
j-white: It doesn't look like anything references this file.

opennms-config-model/src/main/java/org/opennms/netmgt/xml/eventconf/Events.java
```diff
@@ -354,14 +355,32 @@ public void initialize(final Partition partition, final EventOrdering eventOrder
...
         partitionEvents(partition);
 
-        for(final Entry<String, Events> loadedEvents : m_loadedEventFiles.entrySet()) {
-            final Events events = loadedEvents.getValue();
+        for (final Events events : m_loadedEventFiles.values()) {
             events.initialize(partition, m_ordering.subsequence());
         }
 
+        // roll up all prioritized events and sort all events by priority
+        // must be done after event.initialize() haas been called to set the event.index
```
mattixtech:
```suggestion
         // must be done after event.initialize() has been called to set the event.index
```

opennms-config-model/src/main/java/org/opennms/netmgt/xml/eventconf/Events.java
```diff
@@ -354,14 +355,32 @@ public void initialize(final Partition partition, final EventOrdering eventOrder
...
+        List<Event> prioritizedEvents = getPrioritizedEvents();
+        m_events.addAll(prioritizedEvents);
+        m_events.sort(Comparator.naturalOrder());
+        // Also add to unpartitioned events for first crack when not using an UEI match
```
mattixtech:
```suggestion
         // Also add to unpartitioned events for first crack when not using a UEI match
```

opennms-config-model/src/main/resources/xsds/eventconf.xsd
smith-opennms: There is a default implemented in the Event object and it is '0'.
Using 0 allows us to only prioritize those Events with a non-zero value which helps to reduce processing overhead when matching events.
If we explicitly declare it in the schema however, it will be marshaled out which is not what we want.
I did change the type to 'nonNegativeInteger' and expanded the description.
Please let me know if that is sufficient. Thanks.

opennms-config-model/src/main/resources/xsds/eventconf.xsd
smith-opennms: Let me know if the updated description is sufficient. Thanks.

opennms-config/src/main/java/org/opennms/netmgt/config/DefaultEventConfDao.java
smith-opennms: I added a comment - basically because we roll up the prioritized events and they still remain in thier original position as well, we would check the event multiple times.
This method is only used for testing at this time so I could revert the change if you prefer.

opennms-config/src/test/resources/org/opennms/netmgt/config/eventd/reloaded/BGP4.events.xml
smith-opennms: removed

opennms-config/src/test/resources/priority/eventconf2.xml
smith-opennms: added missing test that uses this

opennms-config-model/src/main/java/org/opennms/netmgt/xml/eventconf/Events.java
smith-opennms: updated

opennms-config-model/src/main/java/org/opennms/netmgt/xml/eventconf/Events.java
smith-opennms: updated

opennms-config-model/src/main/resources/xsds/eventconf.xsd
smith-opennms: Hi, did you see the updated code?
This comment is on an 'Outdated' snippet.....
I think it's somethign funky in the gitUI....
The update calls out that 
 - the higher the number - the higher the priority
 - it restricts the valid values to nonNegativeIntegers
see: https://github.com/OpenNMS/opennms/pull/2299/commits/a1769e05808fe21bbccc3a77848dac02e5b18f5b

I don't think we should limit the range - if we place a MAX value on priority and a plugin uses that value, then there will be no way to override the event definition on the file system.
I believe this is the core Use Case for this feature.

I'm not sure that I understand the 3rd question. do you mean making the default in the XSD "0" or something else? could you rephrase? Thanks.

pbrane: The higher the number the higher the priority.  Okay, priority in what exactly?

pbrane: I'm missing the logic on the MAX value and plugin behavior.  But then, I still don't understand from the comments what priority actually means.

pbrane: Yes, I'm asking why doesn't make the default "0" in the XSD any different than treating it as "0" in the code?

smith-opennms: Priority in matching an incoming event.
As every event comes in and is matched against an event definition, we take the event definitions with non-zero priority and bump them to the front of the queue.
So all of the event definitions with priority > 0, have a chance to match the incoming event before any of the other partitioned event definitions.
This allows the customer to override an event definition that is loaded via a plugin for instance.
With the introduction of Project Sextant, we now load some event definitions via the osgi registry - i.e. they are included in the plugin. 
Without the ability to declare a 'priority' on the event definition on disk, the customer would not be able to override any of the event definitions loaded by the classloader.
The principle Use Case then is for a customer to define a few event definitions with a priority > zero when they need to override the definition included in a plugin.
In the future, there may be more complicated scenarios, e.g. if multiple plugins define the same matching event definition, hence the desire not to cap the priority that the user can apply to their definition on disk.

smith-opennms: as for making the default "0" in the xsd vs in the code, the difference is that when it is declared in the xsd schema, serializing event definitions will include a 'priority' element even if it wasn't there.
we have many tests that cover the serialization and de-serialization of these objects and they would all break if we changed the schema in this manner.

pbrane: Sorry, one less pang here but I don't want to block the PR.  It seems to me that if the tests would fail then would should fix them since this is a new element in the schema and setting a default is the same has having one specified.  I would hate to make the assumption that a user is only going to specify a priority when it makes sense for a plugin's behavior.  I'm still a fan of putting in a default for any *optional* element or attribute.  My $0.02.

smith-opennms: Sorry, I phrased that poorly.
It's not because tests would fail that I feel we shouldn't do it that way; It's because forcing the default into the XML serialization would pollute the XML with unnecessary information. The tests failing are a symptom of that.
There is no assumption in the underlying code that customer would only add priority in the Use Case it is addressing - however there would be no benefit for them to do so otherwise, as the code only solves that particular Use Case.
What I'm hoping to achieve is to make the changes to the configuration as simple and clean as possible for the end user. I think one of the challenges with a system as large and as flexible as OpenNMS is to keep things simple.
I feel if this is covered in the Admin Guide properly (*note to self) the way it is implemented in this review will ultimately be the best for the end user.

smith-opennms: And perhaps if I describe it as follows:
what we are trying to model is Event definitions that can optionally have a priority
those definitions that haven't expressly been assigned a priority, should not have one.
the difference from having a default value for priority vs even having a priority assigned. HTH. :)

PR #2300, opened 2018-12-19
Proposal: Define fixed import order

mvrueden:
We have the problem that using different IDEs result in various import strategies.
This is a proposal to define a common way of importing packages in OpenNMS.
This is compatible with the IDEA formatting I created a while back, which should be mostly compatible to eclipse. 

Let me know what you guys think and we can provide eclipse formatting rules as well to make this easier for everyone.


checkstyle/src/main/resources/nms_checks.xml
```diff
@@ -25,5 +25,24 @@
             <property name="format" value="^UTF-8$"/>
             <message key="forbid.UTF8.string.literal" value="Use ''java.nio.charset.StandardCharsets.UTF_8.name()'' instead of ''{0}''" />
         </module>
+
+        <!--
+            Ensure the imports are in a defined order, otherwise different IDEs enforce different results, which
+            causes unnecessary git diffs
+        -->
+        <module name="ImportOrder">
+            <property name="groups" value="java,javax,org,com,*" />
```
smith-opennms: do you want a separate group for org.opennms?
i assume idea can do this.

mvrueden: When I originally created the import rules for Idea I refactored them from what was already there in OpenNMS + the default rules from eclipse. 

I don't really have a strong opinion about this, except * imports on non static packages.
I just like to have not to all the time worry about the imports when committing or reviewing code, as it is kinda spam (-:
So any (sane) order is fine for me (-:

j-white:
Failing to compile because of improper import order seems pretty drastic.

We'd need to make sure that the rules are well documented, and that we define and document installing the formatting rules for common IDEs.

mvrueden:
Just to recapture: 
The goal is to prevent changing the import order when making small changes, usually caused by auto-import mechanisms of the IDE.

Failing the build is pretty hard when the order is not correct, I agree.
If there is an easier way of achieving this, I am open for ideas.

For documentation purposes:
I'd propose intelliJ and eclipse formatting rules, which should be used and we document the that these rules must be used.

j-white:
I can't think of a different way either - I guess it's fine since it's always possible to disable checkstyle at compile time.

mvrueden:
Merge conflicts. Waiting for them to be resolved.

mvrueden:
Not going to work on this any time soon. Closing.

PR #2301, opened 2018-12-19
NMS-10465: Event Config matching on Paramter key and value

smith-opennms:
This PR adds the ability to configure parameter keys and their values in the event mask for matching event config.

* JIRA: https://issues.opennms.org/browse/NMS-10465

### DIFF
--- a/opennms-config-model/src/main/java/org/opennms/netmgt/xml/eventconf/EventMatchers.java
+++ b/opennms-config-model/src/main/java/org/opennms/netmgt/xml/eventconf/EventMatchers.java
@@ -193,10 +193,13 @@ public abstract class EventMatchers  {
 	public static Field field(String name) {
 
         if (name.startsWith("parm[") && name.endsWith("]")) {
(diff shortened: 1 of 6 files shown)


opennms-config-model/src/main/java/org/opennms/netmgt/xml/eventconf/EventMatchers.java
```diff
@@ -191,6 +191,19 @@ public String toString() {
 	}
 	
 	public static Field field(String name) {
+
+        if (name.startsWith("parm[") && name.endsWith("]")) {
+            String parmName = name.substring(5, name.length() - 1);
```
mattixtech:
```suggestion
            String parmName = name.substring("parm[".length());
```

mattixtech: Oops. I was mostly trying to get rid of the magic number 5 in the call to substring. But it doesn't really matter. Can ignore that one.

PR #2302, opened 2018-12-19
NMS-9659: change BestMatchPingerFactory to use loopback address for testing

RangerRick:
This PR creates a new method for specifically detecting a loopback address, and the `BestMatchPingerFactory` uses that to determine whether ping works.

* JIRA: http://issues.opennms.org/browse/NMS-9659

### DIFF
--- a/core/lib/src/main/java/org/opennms/core/utils/InetAddressUtils.java
+++ b/core/lib/src/main/java/org/opennms/core/utils/InetAddressUtils.java
@@ -32,9 +32,11 @@ import java.math.BigInteger;
 import java.net.Inet4Address;
 import java.net.Inet6Address;
 import java.net.InetAddress;
+import java.net.NetworkInterface;
 import java.net.UnknownHostException;
 import java.text.DecimalFormat;
 import java.util.Arrays;
+import java.util.Enumeration;
 import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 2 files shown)


opennms-icmp/opennms-icmp-best/src/main/java/org/opennms/netmgt/icmp/best/BestMatchPingerFactory.java
```diff
@@ -85,7 +85,7 @@ private static PingerMatch tryPinger(final Class<? extends Pinger> pingerClass)
 
         try {
             final long timeout = Long.valueOf(System.getProperty("org.opennms.netmgt.icmp.best.timeout", "500"), 10);
-            final Number result = pinger.ping(InetAddressUtils.getLocalHostAddress(), timeout, 0);
+            final Number result = pinger.ping(InetAddressUtils.getLocalLoopbackAddress(), timeout, 0);
```
mattixtech: If we use optionals above, then this can become:

```
pinger.ping(InetAddressUtils.getLocalLoopbackAddress()
                    .orElseThrow(() -> new IllegalStateException("No result pinging localhost.")), timeout, 0);
```

core/lib/src/main/java/org/opennms/core/utils/InetAddressUtils.java
```diff
@@ -120,6 +122,27 @@ public static String getLocalHostAddressAsString() {
         return localhost == null? "127.0.0.1" : localhost;
     }
 
+    public static InetAddress getLocalLoopbackAddress() {
```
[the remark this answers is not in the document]
RangerRick: good idea

opennms-icmp/opennms-icmp-best/src/main/java/org/opennms/netmgt/icmp/best/BestMatchPingerFactory.java
```diff
@@ -85,7 +86,8 @@ private static PingerMatch tryPinger(final Class<? extends Pinger> pingerClass)
 
         try {
             final long timeout = Long.valueOf(System.getProperty("org.opennms.netmgt.icmp.best.timeout", "500"), 10);
-            final Number result = pinger.ping(InetAddressUtils.getLocalHostAddress(), timeout, 0);
+            final InetAddress addr = InetAddressUtils.getLocalLoopbackAddress().orElseThrow(() -> new IllegalStateException("No pingable address found."));
```
j-white: We could move this out of the `tryPing` loop, and pass a reference to the interface to the function instead.

It may also be worth trying some hard-coded address such as `::1` and `127.0.0.1` as a last ditch effort before falling back to the `NullPinger` as well.

PR #2304, opened 2018-12-20
NMS-10518: some modernization/fixes for runjava and opennms

RangerRick:
This is a precursor to work required to get OpenNMS running under Java 9.  It's a major refactor/cleanup of the internals of the `runjava` and `opennms` shell scripts.

* fix all shellcheck errors
* convert all variables to arrays where appropriate (things that will be passed on command lines)
* common legacy variables people include in `opennms.conf` should get auto-upgraded to arrays for internal use
* fix java 9+ support in runjava

* JIRA: http://issues.opennms.org/browse/NMS-10518

### DIFF
--- a/opennms-base-assembly/src/main/filtered/bin/opennms
+++ b/opennms-base-assembly/src/main/filtered/bin/opennms
@@ -37,6 +37,7 @@
 #### ------------> DO NOT CHANGE VARIABLES IN THIS FILE <------------- ####
 
 # Home directory for OpenNMS.
+# shellcheck disable=SC2154
 OPENNMS_HOME="${install.dir}"
 
 # PID file for OpenNMS.
@@ -70,7 +71,7 @@ STOP_TIMEOUT=10
 JAVA_HEAP_SIZE=1024
 
 # Additional options passed to Java when starting OpenNMS.
-# ADDITIONAL_MANAGER_OPTIONS=""
+# ADDITIONAL_MANAGER_OPTIONS=()
 
 # Use incremental garbage collection.
 USE_INCGC=""
@@ -82,7 +83,7 @@ HOTSPOT=""
 VERBOSE_GC=""
 
 # Additional options to pass to runjava.
-RUNJAVA_OPTIONS=""
+RUNJAVA_OPTIONS=()
 
 # JMX URL that this script uses to communicate with a running OpenNMS daemon
 # when it cannot connect using the Attach API (automatically or by PID)
@@ -97,6 +98,14 @@ MAXIMUM_FILE_DESCRIPTORS=20480
 # maximum size of stack segment (in kbytes) to be setted by ulimit -s
 MAXIMUM_SIZE_STACK_SEGMENT=8192
 
+ADDITIONAL_CONTROLLER_OPTIONS=()
+ADDITIONAL_MANAGER_OPTIONS=()
+APP_PARMS_CONTROLLER=()
+APP_VM_PARMS=()
+
+CONTROLLER_OPTIONS=("-Dopennms.home=$OPENNMS_HOME")
+#CONTROLLER_OPTIONS=("${CONTROLLER_OPTIONS[@]}" "-Dlog4j.configuration=log4j.properties")
+
 COMMAND=""
 
 #### ------------> DO NOT CHANGE VARIABLES IN THIS FILE <------------- ####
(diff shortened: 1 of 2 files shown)


opennms-base-assembly/src/main/filtered/bin/opennms
```diff
@@ -821,25 +844,25 @@ case "$COMMAND" in
...
+			doKill "$@"
 		fi
 		;;
 
 	restart)
 		## Stop the service and regardless of whether it was
 		## running or not, start it again.
-		$OPENNMS_HOME/bin/opennms stop
-		$OPENNMS_HOME/bin/opennms start
+		"$OPENNMS_HOME/bin/opennms" stop && \
```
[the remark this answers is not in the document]
RangerRick: Yes, we currently have an issue where OpenNMS will fail to stop, and then we restart, and things get in a *really* weird state as far as half-used DB connections and files being written.  I'd rather it actually fail and make the admin evaluate why.

opennms-base-assembly/src/main/filtered/bin/opennms
```diff
@@ -390,176 +406,187 @@ doStart(){
...
 		fi
 
-		APP_VM_PARMS="$CONTROLLER_OPTIONS"
-		APP_PARMS_BEFORE="$PARM_PID -u $JMX_URL pause $SERVICE"
-		if [ -z "$NOEXECUTE" ]; then
-			$JAVA_CMD $APP_VM_PARMS -jar $BOOTSTRAP $APP_PARMS_CONTROLLER $APP_PARMS_BEFORE "$@" $APP_PARMS_AFTER
+		APP_VM_PARMS=("${CONTROLLER_OPTIONS[@]}")
+		APP_PARMS_BEFORE=("${PARM_PID[@]}" -u "$JMX_URL" pause)
+		if [ -n "$SERVICE" ]; then
+			APP_PARMS_BEFORE=("${APP_PARMS_BEFORE[@]}" "$SERVICE")
```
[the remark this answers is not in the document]
RangerRick: OK I don't know why I thought this was a newer feature, but looks like `+=` has been around since bash 3.1.

opennms-base-assembly/src/main/filtered/bin/opennms
```diff
@@ -305,82 +321,82 @@ doStart(){
...
+	if [ "$OPROFILE" -gt 0 ]; then
 		echo "- enabling OProfile support" >&2
-		JPDA="-agentpath:/usr/lib/oprofile/libjvmti_oprofile.so ${JPDA}"
+		JPDA=("-agentpath:/usr/lib/oprofile/libjvmti_oprofile.so" "${JPDA[@]}")
 	fi
 
 	if [ "$SERVICE" = "" ]; then
-		APP_VM_PARMS="$JPDA $MANAGER_OPTIONS"
-		APP_PARMS_BEFORE="start"
+		APP_VM_PARMS+=("$MANAGER_OPTIONS")
```
mattixtech: This one was copying from the JPDA array before so it can't be simplified to just `+=`
```suggestion
		APP_VM_PARMS=("${JPDA[@]}" "$MANAGER_OPTIONS")
```

opennms-base-assembly/src/main/filtered/bin/opennms
```diff
@@ -390,176 +406,187 @@ doStart(){
 			return 1
 		fi
 		sleep $STATUS_WAIT
-		STATUS_ATTEMPTS=`expr $STATUS_ATTEMPTS + 1`
+		STATUS_ATTEMPTS="$((STATUS_ATTEMPTS++))"
```
mattixtech: I don't think this works anymore since the assignment is not capturing the incremented value. Just increment the value without re-assigning.
```suggestion
		((STATUS_ATTEMPTS++))
```

opennms-base-assembly/src/main/filtered/bin/opennms
```diff
@@ -390,176 +406,187 @@ doStart(){
...
+			APP_PARMS_BEFORE=("${PARM_PID[@]}" -u "$JMX_URL" stop)
+			if [ -n "$SERVICE" ]; then
+				APP_PARMS_BEFORE+=("$SERVICE")
+			fi
+			"${JAVA_CMD[@]}" "${APP_VM_PARMS[@]}" -jar "$BOOTSTRAP" "${APP_PARMS_CONTROLLER[@]}" "${APP_PARMS_BEFORE[@]}" "$@" "${APP_PARMS_AFTER[@]}"
 		fi
 
 		sleep $STATUS_WAIT
-		STATUS_ATTEMPTS=`expr $STATUS_ATTEMPTS + 1`
+		STATUS_ATTEMPTS="$((STATUS_ATTEMPTS++))"
```
mattixtech: This one needs to be fixed too
```suggestion
		((STATUS_ATTEMPTS++))
```

RangerRick: d'oh!  yeah

j-white:
It looks like the `ADDITIONAL_MANAGER_OPTIONS` in the `opennms` script are no longer being properly applied with these changes. For example, after startup, the JVM no longer has `-DisThreadContextMapInheritable=true` set, which leads to log messages ending up in `uncategorized.log`.

PR #2305, opened 2018-12-24
HZN-1393: Send Topologies updates to Kafka

rssntn67:
HZN-1393: send topologies updates to kafka

* JIRA: http://issues.opennms.org/browse/HZN-1393

added topology Message to Kafka producer


features/kafka/producer/src/main/proto/opennms-kafka-producer.proto
```diff
@@ -147,3 +147,68 @@ message Node {
...
+  NEW = 0;
+  UPDATE=1;
+  DELETE=2;
+}
+
+message OnmsTopologyRef {
+  string id = 1;
+}
+
+message OnmsTopologyMessage {
```
j-white: Can we remove the "Onms" prefix and the "Message" suffix from these types? All of the objects are implicitly Onms objects.

j-white:
I think we need to rework the way the objects are persisted in Kafka here to make it easier to consume. 

Instead of requiring the consumer to keep track of the adds/updates/deletes, we can split the topology into two topics - one for vertices and one for edges. Vertices and edges have unique ids, so they should be keyed by these ids in the topic. When pushing updates, we simply push a new record over the same key. When pushing a delete, we push a null record over the key.

In this fashion, we can model the topology as a ktable similar to what we do with alarms.

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
```diff
@@ -169,8 +192,64 @@ public void destroy() {
...
         }
+        
+        if (forwardTopologyMessages) {
+            topologyDao.unsubscribe(this);
+        }
+
     }
 
+    private void forwardTopologyMessage(OnmsTopologyMessage message) {
+        if (message.getMessagebody() instanceof OnmsTopologyVertex) {
```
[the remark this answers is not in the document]
rssntn67: the topology contains vertex shared and ports.... so updates should contain creating or deleting vertex shared and updating such obejcts too.

So I guess I have to use suggested using-oneof...in any case decoding is a quite simple operation....

features/kafka/producer/src/main/proto/opennms-kafka-producer.proto
rssntn67: Sure!


> I think we need to rework the way the objects are persisted in Kafka here to make it easier to consume.
> 
> Instead of requiring the consumer to keep track of the adds/updates/deletes, we can split the topology into two topics - one for vertices and one for edges. Vertices and edges have unique ids, so they should be keyed by these ids in the topic. When pushing updates, we simply push a new record over the same key. When pushing a delete, we push a null record over the key.
> 
> In this fashion, we can model the topology as a ktable similar to what we do with alarms.

I see but I have not a clear understanding of what you are just talking about.
Every topology Object has an unique id so no problem for implementng this logic

the code now creates a topic for each protocol CDP, LLDP, OSPF, BRIDGE and ISIS....then I send the refid....and then I serialize the object.

if the topology object is deleted I can send null.
But consider that a topology is a complex object...and edge (better shared) is a relation among vertices....in any case you have to rebuild topology....

rssntn67: ..we need a Topic for ev ery protocol vertex,edge and segment

a vertex is either a cdp and ospf vertex....then for some reason...ospf is not more supported..then you got a delete for the vertex in cdp....

if you have only one matrix with vertex ref....you will delete the vertex....so you need to distinguish among protocols.

rssntn67:
Latest commit...
commit 14d5b65be5d00e3d8a9c72a570f5c80019eb5831 (HEAD -> jira/HZN-1393, origin/jira/HZN-1393)
Author: Antonio <rssntn67@yahoo.it>
Date:   Tue Jan 8 17:26:36 2019 +0100

    HZN-1377: send topologies updates to kafka

    removed the "Onms" prefix and the "Message" suffix
    pushing a delete, we push a null record
    updates will send the whole object
    created topics for vertices, edges and segments

The point is how many topic we should use. IMO a topic for every protocol and for vertices, edges and segments.

features/enlinkd/adapters/collectors/bridge/src/main/java/org/opennms/netmgt/enlinkd/NodeDiscoveryBridge.java
```diff
@@ -28,9 +28,9 @@
 
 package org.opennms.netmgt.enlinkd;
 
+import static org.opennms.core.utils.InetAddressUtils.getBridgeAddressFromStpBridgeId;
```
j-white: Nothing changed in this file but the import order.

features/enlinkd/adapters/collectors/ipnettomedia/src/main/java/org/opennms/netmgt/enlinkd/NodeDiscoveryIpNetToMedia.java
```diff
@@ -80,47 +80,31 @@ protected void runNodeDiscovery() {
...
-                        macep.getSourceIfIndex() != null &&
-                        (macep.getIpNetToMediaType() == IpNetToMediaType.IPNETTOMEDIA_TYPE_DYNAMIC
-                        || macep.getIpNetToMediaType() == IpNetToMediaType.IPNETTOMEDIA_TYPE_STATIC)) {
-                    if (LOG.isDebugEnabled()) {
-                        LOG.debug("processIpNetToMediaRow: node [{}], mac address {} and ip {} mediatype {}. saving",
+
+                LOG.debug("processIpNetToMediaRow: node [{}], {}:{}:{}. ",
+                          getNodeId(),
+                          macep.getPhysAddress(),
+                          str(macep.getNetAddress()),
```
j-white: This should be wrapped in a `LOG.isDebugEnabled() {` statement to avoid unnecessarily calling `str(macep.getNetAddress())`.

features/enlinkd/adapters/collectors/ipnettomedia/src/main/java/org/opennms/netmgt/enlinkd/NodeDiscoveryIpNetToMedia.java
```diff
@@ -80,47 +80,31 @@ protected void runNodeDiscovery() {
...
+
+                if (macep.getPhysAddress() == null
+                        || macep.getNetAddress() == null
+                        || macep.getIpNetToMediaType() == null
+                        || macep.getIpNetToMediaType() == IpNetToMediaType.IPNETTOMEDIA_TYPE_INVALID) {
+                    LOG.debug("processIpNetToMediaRow: node [{}], {}:{}:{}. skipping",
                               getNodeId(),
-                              macep.getPhysAddress(), 
+                              macep.getPhysAddress(),
                               str(macep.getNetAddress()),
```
j-white: Similarly here.

features/enlinkd/adapters/collectors/ipnettomedia/src/main/java/org/opennms/netmgt/enlinkd/NodeDiscoveryIpNetToMedia.java
```diff
@@ -80,47 +80,31 @@ protected void runNodeDiscovery() {
...
-                             macep.getPhysAddress(), 
-                             str(macep.getNetAddress()),
-                             macep.getIpNetToMediaType(),
-                             macep.getSourceIfIndex());
+                }
+                
+                LOG.debug("processIpNetToMediaRow: node [{}], {}:{}:{}. store",
+                          getNodeId(),
+                          macep.getPhysAddress(),
+                          str(macep.getNetAddress()),
```
j-white: Here too.

features/enlinkd/service/impl/src/main/java/org/opennms/netmgt/enlinkd/service/impl/TopologyServiceImpl.java
```diff
@@ -0,0 +1,81 @@
...
+
+package org.opennms.netmgt.enlinkd.service.impl;
+
+import org.opennms.netmgt.enlinkd.persistence.api.TopologyEntityCache;
+import org.opennms.netmgt.enlinkd.service.api.TopologyService;
+
+public class TopologyServiceImpl implements TopologyService {
+
+    private TopologyEntityCache m_topologyEntityCache;
+    private Boolean m_updates = false;
```
j-white: Can use an AtomicBoolean here instead and not have to worry about locking.

features/enlinkd/service/impl/src/main/java/org/opennms/netmgt/enlinkd/service/impl/TopologyServiceImpl.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Override
+    public  boolean parseUpdates() {
+        synchronized (m_updates) {
+            if (m_updates.booleanValue()) {
```
j-white: No need to call `booleanValue()` here.

features/enlinkd/service/impl/src/main/java/org/opennms/netmgt/enlinkd/service/impl/TopologyServiceImpl.java
```diff
@@ -0,0 +1,81 @@
...
+    @Override
+    public void updatesAvailable() {
+        synchronized (m_updates) {
+            m_updates = true;
+        }
+    }
+
+    @Override
+    public boolean hasUpdates() {
+        synchronized (m_updates) {
```
j-white: Can be simplified to `return m_updates`.

features/enlinkd/tests/src/test/java/org/opennms/netmgt/enlinkd/Nms17216EnIT.java
```diff
@@ -745,10 +748,142 @@ public void testNetwork17216CdpTopology() throws Exception {
...
+            assertTrue(ip.isManaged());
+        });
+
+        CdpOnmsTopologyUpdater cdptopology = m_linkd.getCdpTopologyUpdater();
+        assertNotNull(cdptopology);
+        OnmsTopology topology = cdptopology.buildTopology();
+        assertEquals(2, topology.getVertices().size());
+        assertEquals(4, topology.getEdges().size());
+        
+        OnmsTopologyLogger tl = EnLinkdBuilderITCase.createAndSubscribe(
```
j-white: Can we write a test that asserts that the `OnmsTopologyConsumer` has actually received (or consumed) the expected messages?

features/enlinkd/tests/src/test/java/org/opennms/netmgt/enlinkd/Nms17216EnIT.java
```diff
@@ -745,10 +748,142 @@ public void testNetwork17216CdpTopology() throws Exception {
...
+        assertTrue(m_linkd.runSingleSnmpCollection(switch1.getId()));
+        assertEquals(4, m_lldpLinkDao.countAll());
+        assertTrue(m_lldpTopologyService.hasUpdates());
+        
+        assertTrue(m_linkd.runSingleSnmpCollection(switch2.getId()));
+        assertEquals(10, m_lldpLinkDao.countAll());
+        assertTrue(m_lldpTopologyService.hasUpdates());
+        
+        // Now we support by default configuration: BRIDGE, CDP, ISIS, LLDP and OSPF
+        assertEquals(5, m_topologyDao.getSupportedProtocols().size());
```
j-white: This is out of place - why are we asserting that the set of supported protocols in the topologyDao has these 5 protocols in a test called `testNetwork17216CdpTopology` in the `Nms17216EnIT.java`?

features/kafka/producer/src/main/proto/opennms-kafka-producer.proto
```diff
@@ -147,3 +147,42 @@ message Node {
...
+message TopologyVertexNode {
+  NodeCriteria node_criteria = 1;
+}
+
+message TopologyVertexSegment {
+}
+
+message TopologyPort {
+  string vertex_id = 1;
+  uint64 ifindex = 2;
```
j-white: The `ifIndex` and `ifName` fields are already called `if_index` and `if_name ` in some of the other existing messages - can use this same name format?

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
```diff
@@ -168,6 +199,103 @@ public void destroy() {
...
+            OnmsTopologyEdge edge = (OnmsTopologyEdge) message.getMessagebody();
+            final OpennmsModelProtos.TopologyEdge mappedTopoMsg = 
+                    protobufMapper.toEdgeTopologyMessage(message.getProtocol().getId(),edge).build();
+            forwardTopologyEdgeMessage(mappedTopoMsg.getRef().toByteArray(),
+                                       mappedTopoMsg.toByteArray());
+        } else if (message.getMessagebody() instanceof OnmsTopologySegment) {
+            OnmsTopologySegment edge = (OnmsTopologySegment) message.getMessagebody();
+            final OpennmsModelProtos.TopologySegment mappedTopoMsg = 
+                    protobufMapper.toSegmentTopologyMessage(message.getProtocol().getId(),edge).build();
+            forwardTopologySegmentMessage(mappedTopoMsg.getRef().toByteArray(),
```
j-white: Can we model the segments as vertices and send these to the existing "vertex" topic instead of creating a separate topic for the segments? The segments are nothing but "special" vertices.

features/enlinkd/adapters/collectors/ipnettomedia/src/main/java/org/opennms/netmgt/enlinkd/NodeDiscoveryIpNetToMedia.java
rssntn67: commit c1872f7f58a3616e0b18f95044527c2a1beb6c6d (HEAD -> jira/HZN-1393)
Author: Antonio <rssntn67@yahoo.it>
Date:   Tue Jan 15 10:00:33 2019 +0100

    Added if control on LOG.debug to avoid calculation

features/enlinkd/service/impl/src/main/java/org/opennms/netmgt/enlinkd/service/impl/TopologyServiceImpl.java
rssntn67: commit 934cc508d3e5f2ef2d363db7498cde48d4c8788a (HEAD -> jira/HZN-1393)
Author: Antonio <rssntn67@yahoo.it>
Date:   Tue Jan 15 10:14:55 2019 +0100

    Used AtomicBoolean instaed of Boolean

features/enlinkd/adapters/collectors/ipnettomedia/src/main/java/org/opennms/netmgt/enlinkd/NodeDiscoveryIpNetToMedia.java
rssntn67: well now I have set in this class the if LOG.isDebugEnabled...so no need to revert changes...

features/enlinkd/tests/src/test/java/org/opennms/netmgt/enlinkd/Nms17216EnIT.java
rssntn67: I have updated code su support reload event now.
Added tests..too.

commit 4e6b0023e2e14278cee91dc74c37abdeea40fa29 (HEAD -> jira/HZN-1393)
Author: Antonio <rssntn67@yahoo.it>
Date:   Wed Jan 16 19:34:31 2019 +0100

    Added reload configuration for enlinkd

    Supported reload event for enlinkd.
    Added reloadConfig to update configuration and reload daemon
    Also let spring instantiate Updaters and BridgeDiscovery
    Renamed common classes to NodeCollector and Topologyupdater

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
rssntn67: So if I understand is very similar in the way we layout maps....

so then it we should transform the segment into...
Consider we have Segment with 4 ports connectin V1,V2,V3,V4
then this can be translated into

5 Vertex S, V1,V2,V3,V4 and 4 edges
V1-S
V2-S
V3-S
V4-S

Please confirm

j-white: Correct, if `S` is a shared segment and `V1`, `V2`, `V3`, `V4` are ports attached to this shared segment. Then we can model this as 5 vertices, one for each port and one for the segment, as well as 4 edges connecting the port to the segment.

features/enlinkd/tests/src/test/java/org/opennms/netmgt/enlinkd/Nms17216EnIT.java
rssntn67: commit 251ee68c19266e247bd9d84ea709a67ca2ebae11 (HEAD -> jira/HZN-1393)
Author: Antonio <rssntn67@yahoo.it>
Date:   Wed Jan 16 21:22:48 2019 +0100

    OnmsTopologyLogger has actually received the expected messages

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
rssntn67: Another solution is to avoid using ege an using only segments....
at the en an ege has only two vertcies and a segment has many.....

j-white:
Now that there are no more conflicts, I think we're good to merge this now.

Let's squash merge when the build goes green.

PR #2306, opened 2019-01-02
No Jira Yet: Improve efficiency of incremental alarm handling by comparing fields directly rather than mapping to a common object

mattixtech:
This PR contains changes to the Kafka producer's handling of incremental alarms to make comparing alarms more efficient.

At this point, I'm not sure how significant the performance change would be but I think these changes should be used even if the performance is similar since this approach is easier to understand than the previous approach.

I've made some changes to the RelatedAlarm entity in the hopes that we can get a performance increase by comparing only related alarm Ids when looking for a difference in alarms without loading the full alarm objects from the DB. I'm not sure if I've done this correctly so please take a look.

### DIFF
--- a/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/AlarmEqualityChecker.java
+++ b/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/AlarmEqualityChecker.java
@@ -28,105 +28,153 @@
 
 package org.opennms.features.kafka.producer;
 
-import java.util.List;
 import java.util.Objects;
-import java.util.function.Function;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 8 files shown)


mattixtech:
I made some changes to get tests to pass.

I was hoping that these changes would allow us to check related alarm Ids without having to load the situation alarm and related alarm objects from the associated alarms table. However what I am seeing is that those are still fetched regardless of the lazy fetch annotation. I'm wondering if maybe those need to be fetched as a result of the mapping relationship between OnmsAlarm->Associatedalarm. Because of that, I'm not sure if there will be any meaningful performance gain.

mattixtech:
Right now performance doesn't seem to be a problem for this area of the code anymore. To avoid churning code unnecessarily I'll close this.

PR #2308, opened 2019-01-04
HZN-1459: Use mapstruct for mapping to protobuf objects in the kafka producer

mattixtech:
Following the same style as in https://github.com/OpenNMS/opennms/commit/cb72a6f89f80f42d629b5a53a5aa834bf6d3a33f

Updating our kafka producer protobuf mapping code to use MapStruct.

My aim is to update any instances where we are doing explicit type mapping to use a mapping factory backed by a map struct mapping implementation.

* JIRA: http://issues.opennms.org/browse/HZN-1459


mattixtech:
@j-white It ended up being more churn and complication then I was hoping for when I first looked into making the change. After seeing your feedback I agree that there is no advantage to switching.  I'll close this PR for that reason.

PR #2311, opened 2019-01-10
HZN-1460: New syslogd configuration to include raw syslog message in the generated event

mattixtech:
- Adds new configuration attribute in syslogd-configuration.xml called "includeRawSyslogmessage"
- New behavior is disabled by default

When the new config is enabled, all events generated by syslogd include an additional event parameter called "rawSyslogmessage" the contains the entire message as seen before parsing. This differs from existing message parameter since that only includes the body of the message.

* JIRA: http://issues.opennms.org/browse/HZN-1460


opennms-base-assembly/src/main/filtered/etc/syslogd-configuration.xml
```diff
@@ -14,6 +14,7 @@
             matching-group-message="8"
             discard-uei="DISCARD-MATCHING-MESSAGES"
             timezone=""
+            includeRawSyslogmessage="false"
```
j-white: Can we omit changing this file, and set a default in the code and XSD instead?

(Avoid changing the configuration files for new optional parameters to help reduce conflicts when upgrading.)

PR #2312, opened 2019-01-14
NMS-10486: Configure rule to discard notifications when there is no node associated

cgorantla:
* JIRA: http://issues.opennms.org/browse/NMS-10486

### DIFF
--- a/opennms-config-model/src/main/castor/notifications.xsd
+++ b/opennms-config-model/src/main/castor/notifications.xsd
@@ -66,7 +66,7 @@
 
         <element maxOccurs="1" minOccurs="0" name="description" type="string"/>
 
-        <element maxOccurs="1" minOccurs="1" name="rule" type="string"/>
+        <element maxOccurs="1" minOccurs="1" ref="this:rule"/>
 
         <element maxOccurs="1" minOccurs="0" name="notice-queue" type="string"/>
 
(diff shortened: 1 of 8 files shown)


cgorantla:
@agalue @j-white As I m not aware of all the use-cases for notifications,  not sure if there are cases where we need to save notifications  even if there is no node/interface/service info.

agalue:
What I'm going to say is debatable, but here are my thoughts:

As OpenNMS is intended to monitor a pre-defined set of devices from a given network, I cannot see any usefulness on adding behavior for a node that is not managed by OpenNMS. Also, when this happens, the filter rule on the notification definition will fail (hence the issue).

The FilterDao is IP based, but normally users use it to filter nodes (and the "catinc" prefix, introduces other complexities here).

Validating the service depends on the use case (I wouldn't consider it critical); in fact, like the JIRA issue, you can receive syslog messages from a monitored node, and be able to send notifications only for a few of them, and those nodes don't have to have monitored services associated with them.

Makes sense ?

cgorantla:
Thanks @agalue for the input. Updated code to to disregard notices that are not associated with a node.

opennms-services/src/main/java/org/opennms/netmgt/notifd/BroadcastEventProcessor.java
```diff
@@ -447,7 +447,10 @@ private boolean continueWithNotice(Event event) {
         String service = event.getService();
 
         boolean continueNotice = false;
-
+        // If event is not associated with node, disregard Notice
+        if(nodeID == null) {
```
j-white: What if the interface and/or service is null? Won't we run into the same problem?

cgorantla: As @agalue indicated, service may not be critical. So I updated code to check node and interface. 
So the behavior now is:
If notice doesn't have valid node/interface, it won't be saved in DB.  
If only service is null, it  will still be saved in DB.

j-white:
Syslog events are not associated with a "service", which means that the filter configured in the example will never actually be evaluated, so these notifications will also be sent for nodes that don't fall in the given categories.

j-white:
Modifying notifd to suppress notifications for events without an associated node is a breaking change that shouldn't be made in a LTS branch like `foundation-2016`.

cgorantla:
As discussed with @j-white, modified the behavior to suppress notifications only when rule is strict on foundation-2016.
  `<rule strict="true">IPADDR IPLIKE *.*.*.*</rule>`

cgorantla:
Tested the behavior on foundation-2016 branch.

PR #2317, opened 2019-01-16
NMS-10525: Event parameters table have strong limits for the columns

j-white:
JIRA: https://issues.opennms.org/browse/NMS-10525

Here we convert the type of the event parameter column from `varchar(256)` to `text`.

Since this column is part of the primary key, there still is a limit on size which is up to 2712 characters.


agalue:
The customer who reported the problem has tested the patch and it successfully worked on his environment.

RangerRick:
This looks pretty straightforward, is confirmed by the customer, and passed all tests.  I'll go ahead and merge.

PR #2318, opened 2019-01-17
HELM-84: add metadata to measurements queries

RangerRick:
This PR adds the metadata necessary to implement HELM-84.  It does so by enhancing the query response from measurements requests to include a `metadata` block that contains info about nodes and resources related to the query.

* JIRA: http://issues.opennms.org/browse/HELM-84

### DIFF
--- a/features/newts/src/main/java/org/opennms/netmgt/measurements/impl/NewtsFetchStrategy.java
+++ b/features/newts/src/main/java/org/opennms/netmgt/measurements/impl/NewtsFetchStrategy.java
@@ -1,8 +1,8 @@
 /*******************************************************************************
  * This file is part of OpenNMS(R).
  *
- * Copyright (C) 2010-2015 The OpenNMS Group, Inc.
- * OpenNMS(R) is Copyright (C) 1999-2015 The OpenNMS Group, Inc.
+ * Copyright (C) 2010-2019 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2019 The OpenNMS Group, Inc.
  *
  * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
  *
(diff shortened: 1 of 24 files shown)


features/newts/src/main/java/org/opennms/netmgt/measurements/impl/NewtsFetchStrategy.java
```diff
@@ -411,4 +421,30 @@ protected void setSampleRepository(SampleRepository sampleRepository) {
...
     }
+
+    private OnmsNode getNode(final OnmsResource resource, final Source source) {
+        OnmsNode node = null;
+        try {
+            node = ResourceTypeUtils.getNodeFromResourceRoot(resource);
+        } catch (final ObjectRetrievalFailureException e) {
+        }
+        if (node == null) {
+            final String resourceId = source.getResourceId().split("\\.")[0];
```
j-white: Can we use the `ResourceId` class to parse this instead of tokenizing it ourself here? Something like `ResourceId.fromString(resourceId).getParent()`.

core/ipc/sink/kafka/itests/src/test/java/org/opennms/core/ipc/sink/kafka/itests/HeartbeatSinkIT.java
```diff
@@ -177,6 +177,7 @@ public int getNumConsumerThreads() {
     }
 
     @Test(timeout = 60000)
+    @org.springframework.test.annotation.IfProfileValue(name="runFlappers", value="true")
```
j-white: Should this be here, or should this be done to some earlier branch?

features/newts/src/main/java/org/opennms/netmgt/measurements/impl/NewtsFetchStrategy.java
RangerRick: Ah yeah, that makes sense.

PR #2320, opened 2019-01-20
Bootstrap4: Make the Card layout more compact

indigo423:
The bootstrap 4 card introduce a strong padding in the card components in the Web UI which leads to a blown up UI. This commit reduces the padding around the card / box components to make the UI layout more compact.

Additionally I try to use the SVG logo first and fallback to the PNG. The old version uses PNGs in both cases and add some space on the new login page for the GitHub and docs icon.

A before/after screenshot is attached.

### DIFF
--- a/core/web-assets/src/main/assets/style/opennms-theme.scss
+++ b/core/web-assets/src/main/assets/style/opennms-theme.scss
@@ -362,6 +362,19 @@ div.NLnode:hover span.NLdbid, div.NLnode:hover span.NLfs, div.NLnode:hover span.
   margin-bottom: $spacer * 1.5; // equivalent of mb-4
 }
 
+// The bootstrap4 blows up the whole UI. The padding is reduced to have more compact layout.
+.card-header {
+  padding: 0.25rem;
+}
+
+.card-body {
+  padding: 0.5rem;
+}
+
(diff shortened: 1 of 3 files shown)


mmahacek:
I like the updated design. It helps get a lot of information on the screen to minimize scrolling on the main page while still giving a clean look.

mvrueden:
I am going to verify the logo, as I also had that at some point but it didn't work reliably throughout all browsers, maybe that was just a temporary issue.

The first goal was to provide a more or less default bootstrap 4 version.
The reason for this is to avoid customising the styles, and so again basically creating a new OpenNMS Bootstrap Theme.
I see the value of having this, but we should be very careful of not customising the theme at this point. I was more aiming for an already existing theme, like [Core UI](http://coreui.io) instead and customise that, as there all the spacing and such are already more user friendly and the user experience to me looks/feels better. 
Until now I wanted to postpone that discussion, until we actually have a fully migrated bootstrap 4 branch. Migrating to something like core ui would be simple.
To be clear: I am not saying we should not merge this, but should not customise too much, and maybe look for something like Core UI instead before customising.

mvrueden:
As the discussion is for another day, I am going forward and merge this.

PR #2323, opened 2019-01-24
NMS-10533: Make Topology Creator as Karaf command available

patrick-schweizer:
* JIRA: https://issues.opennms.org/browse/NMS-10533

To try it, type in the Karaf console: 
```
feature:install  org.opennms.features.enlinkd.shell
generate-topology
```


features/enlinkd/org.opennms.features.enlinkd.shell/pom.xml
```diff
@@ -0,0 +1,101 @@
...
+      <artifactId>provision</artifactId>
+      <optional>true</optional>
+    </dependency>
+    -->
+    <dependency>
+      <groupId>org.apache.karaf.shell</groupId>
+      <artifactId>org.apache.karaf.shell.core</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
```
j-white: I don't think we need this one.

features/enlinkd/org.opennms.features.enlinkd.shell/pom.xml
```diff
@@ -0,0 +1,101 @@
...
+    <!-- Test dependencies -->
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.opennms.features.enlinkd</groupId>
+      <artifactId>org.opennms.features.enlinkd.topogen</artifactId>
+      <version>24.0.0-SNAPSHOT</version>
```
j-white: This looks like a duplicate.

features/enlinkd/org.opennms.features.enlinkd.shell/pom.xml
```diff
@@ -0,0 +1,101 @@
...
+    <dependency>
+      <groupId>org.osgi</groupId>
+      <artifactId>org.osgi.enterprise</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.opennms.features.enlinkd</groupId>
+      <artifactId>org.opennms.features.enlinkd.topogen</artifactId>
+    </dependency>
+    <dependency>
```
j-white: Some of these are probably not necessary since they are pulled in by `org.opennms.features.enlinkd.topogen`.

features/enlinkd/org.opennms.features.enlinkd.topogen/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+    <!--
+     | uncomment to add all imported (non-local) bundles to your compilation classpath
+    <dependency>
+      <type>pom</type>
+      <groupId>${parent.groupId}</groupId>
+      <artifactId>provision</artifactId>
+      <optional>true</optional>
+    </dependency>
+    -->
+    <dependency>
```
j-white: I don't think we need this here either.

opennms-dao-api/src/main/java/org/opennms/netmgt/dao/api/OnmsDao.java
```diff
@@ -57,6 +57,13 @@
 
     void delete(K key);
 
+    /**
+     * Deletes all entities of this type - use with care!
+     */
+    default void deleteAll(){
```
j-white: Default implementations don't play nice with OSGi, so we may run into some problems here.

features/enlinkd/org.opennms.features.enlinkd.shell/pom.xml
mvrueden: If not, `${project.version}` should be used instead.

features/enlinkd/org.opennms.features.enlinkd.shell/src/main/java/org/opennms/features/enlinkd/shell/GenerateTopologyCommand.java
```diff
@@ -0,0 +1,155 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2016 The OpenNMS Group, Inc.
```
mvrueden:
```suggestion
 * Copyright (C) 2019-2019 The OpenNMS Group, Inc.
```

features/enlinkd/org.opennms.features.enlinkd.shell/src/main/java/org/opennms/features/enlinkd/shell/GenerateTopologyCommand.java
```diff
@@ -0,0 +1,155 @@
...
+ * OpenNMS(R) is Copyright (C) 1999-2016 The OpenNMS Group, Inc.
```
mvrueden:
```suggestion
 * OpenNMS(R) is Copyright (C) 1999-2019 The OpenNMS Group, Inc.
```

features/enlinkd/pom.xml
```diff
@@ -17,5 +17,7 @@
     <module>adapters</module>
     <module>daemon</module>
     <module>tests</module>
+    <module>org.opennms.features.enlinkd.shell</module>
+    <module>org.opennms.features.enlinkd.topogen</module>
```
mvrueden: I prefer `shell` and `topogen` here to be consistent with the other module names.
Can we find a better naming as `topogen`? Maybe `generator` or `generation`?

features/enlinkd/org.opennms.features.enlinkd.topogen/src/test/java/org/opennms/enlinkd/topogen/topology/UndirectedPairGeneratorTest.java
```diff
@@ -0,0 +1,74 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2018 The OpenNMS Group, Inc.
```
mvrueden:
```suggestion
 * Copyright (C) 2019-2019 The OpenNMS Group, Inc.
```

features/enlinkd/org.opennms.features.enlinkd.topogen/src/test/java/org/opennms/enlinkd/topogen/topology/UndirectedPairGeneratorTest.java
```diff
@@ -0,0 +1,74 @@
...
+ * OpenNMS(R) is Copyright (C) 1999-2018 The OpenNMS Group, Inc.
```
mvrueden:
```suggestion
 * OpenNMS(R) is Copyright (C) 1999-2019 The OpenNMS Group, Inc.
```

features/enlinkd/org.opennms.features.enlinkd.topogen/src/test/java/org/opennms/enlinkd/topogen/Asserts.java
```diff
@@ -0,0 +1,47 @@
...
+  public static <T extends Throwable> void assertThrows(Class<T> expectedType, Runnable executable) {
+    try {
+      executable.run();
+    } catch (Throwable t) {
+      if (expectedType.isInstance(t)) {
+        return; // all good
+      }
+      fail("Expected Exception " + expectedType.getName());
+    }
+  }
```
mvrueden: The formatting here seems off.
Can we also point out from which exact version this was copied? 
Can we actually do that without offending any licenses?

features/enlinkd/org.opennms.features.enlinkd.topogen/src/main/java/org/opennms/enlinkd/topogen/protocol/Protocol.java
```diff
@@ -0,0 +1,184 @@
...
+    public void createAndPersistNetwork() throws SQLException {
+        LOG.info("creating {} {} topology with {} {}, {} {}, {} {}, {} {}, {} {}.",
+                this.topology,
+                this.getProtocol(),
+                this.amountNodes, "Nodes" ,
+                this.amountElements, "Elements",
+                this.amountLinks, "Links",
+                this.amountSnmpInterfaces, "SnmpInterfaces",
+                this.amountIpInterfaces, "IpInterfaces");
+        List<OnmsNode> nodes = createNodes(amountNodes);
```
mvrueden: This looks weird. Can we not just add the Strings "Nodes", "Elements", "links" etc. instead of dynamically adding them if they are constants anyways? The log line reads much easier in that case.

features/enlinkd/org.opennms.features.enlinkd.topogen/src/main/java/org/opennms/enlinkd/topogen/TopologyGenerator.java
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    public String toString() {
+      return "TopologyGenerator.TopologyGeneratorBuilder(persister=" + this.persister
+          + ", amountNodes=" + this.amountNodes + ", amountElements=" + this.amountElements
+          + ", amountLinks=" + this.amountLinks + ", amountSnmpInterfaces="
+          + this.amountSnmpInterfaces + ", amountIpInterfaces=" + this.amountIpInterfaces
+          + ", topology=" + this.topology + ", protocol=" + this.protocol
+          + ", deleteExistingTolology=" + this.deleteExistingTolology + ")";
+    }
```
mvrueden: Please use the Google Guava `MoreObjects.ToStringHelper` instead

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/AbstractDaoHibernate.java
```diff
@@ -433,4 +433,10 @@ private void logExtraSaveOrUpdateExceptionInformation(final T entity, final Data
             cause = cause.getCause();
         }
     }
+
+    @Override
+    public void deleteAll() {
+        String sql = String.format("delete from %s entity", this.m_entityClass.getSimpleName());
```
mvrueden: The Usage of `this` and `m_` variables is kinda contrary.
I'd opt for use `m_entityClass` instead.

features/enlinkd/org.opennms.features.enlinkd.topogen/src/main/java/org/opennms/enlinkd/topogen/TopologyPersister.java
```diff
@@ -0,0 +1,252 @@
...
+        final CdpElementDao cdpElementDao,
+        final IsIsElementDao isIsElementDao,
+        final LldpElementDao lldpElementDao,
+        final CdpLinkDao cdpLinkDao,
+        final IsIsLinkDao isIsLinkDao,
+        final LldpLinkDao lldpLinkDao,
+        final OspfLinkDao ospfLinkDao,
+        final IpInterfaceDao ipInterfaceDao,
+        final SnmpInterfaceDao snmpInterfaceDao
+    ) {
```
mvrueden: This also looks odd to me.
If we are only using the DAOs for persisting, let's add a `save` method to the `GenericPersistenceAccessor` instead, as that the purpose of that interface/class is to provide exactly this generic behaviour.

features/enlinkd/org.opennms.features.enlinkd.topogen/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+            <Karaf-Commands>*</Karaf-Commands>
+          </instructions>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.opennms.maven.plugins</groupId>
+        <artifactId>features-maven-plugin</artifactId>
+        <configuration>
+          <bundles>
+            <bundle>mvn:org.opennms.features.enlinkd/org.opennms.features.enlinkd.topogen/${project.version}</bundle>
```
mvrueden: @j-white what is the standing on this? Do we keep doing this or should we already start replacing this with manual `features.xml` files?

features/enlinkd/org.opennms.features.enlinkd.topogen/pom.xml
patrick-schweizer: removed

features/enlinkd/org.opennms.features.enlinkd.shell/pom.xml
patrick-schweizer: removed

features/enlinkd/org.opennms.features.enlinkd.shell/pom.xml
patrick-schweizer: :+1: done.

features/enlinkd/org.opennms.features.enlinkd.shell/pom.xml
patrick-schweizer: :+1: removed

features/enlinkd/org.opennms.features.enlinkd.topogen/pom.xml
j-white: Yeah, good point, let's avoid adding any new features via the `features-maven-plugin` and add them manually to the `features.xml` instead - will make migration easier.

features/enlinkd/org.opennms.features.enlinkd.shell/src/main/java/org/opennms/features/enlinkd/shell/GenerateTopologyCommand.java
```diff
@@ -0,0 +1,155 @@
...
+    @Option(name="--ipinterfaces", description="generate <N> IpInterfaces but not more than amount snmp interfaces")
+    private Integer amountIpInterfaces;
+
+    @Option(name="--topology", description="type of topology (complete | ring | random)")
+    private String topology;
+
+    @Option(name="--protocol", description="type of protocol (cdp | isis | lldp | ospf)")
+    private String protocol;
+
+    @Option(name="--delete", description = "delete existing toplogogy (all OnmsNodes, CdpElements and CdpLinks)")
```
j-white: When we delete, we should only delete nodes and the related links that were created by this command, and not *all* of the nodes in the database.

To achieve this, we could do something like add a new category to the nodes we generate - called something like `GeneratedNode` - and then limit the deletion to nodes with this category.

features/enlinkd/org.opennms.features.enlinkd.topogen/src/main/java/org/opennms/enlinkd/topogen/protocol/Protocol.java
patrick-schweizer: :+1: This is a typical refactor artifact :-( , fixed

features/enlinkd/org.opennms.features.enlinkd.topogen/src/main/java/org/opennms/enlinkd/topogen/TopologyPersister.java
patrick-schweizer: :+1: makes the code much easier, done.

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/AbstractDaoHibernate.java
patrick-schweizer: obsolete, we use the GenericPersistenceAccessor

features/enlinkd/org.opennms.features.enlinkd.shell/src/main/java/org/opennms/features/enlinkd/shell/GenerateTopologyCommand.java
patrick-schweizer: fixed

features/enlinkd/org.opennms.features.enlinkd.topogen/src/main/java/org/opennms/enlinkd/topogen/TopologyGenerator.java
patrick-schweizer: fixed

features/enlinkd/org.opennms.features.enlinkd.shell/src/main/java/org/opennms/features/enlinkd/shell/GenerateTopologyCommand.java
patrick-schweizer: fixed

features/enlinkd/org.opennms.features.enlinkd.shell/src/main/java/org/opennms/features/enlinkd/shell/GenerateTopologyCommand.java
patrick-schweizer: fixed

features/enlinkd/pom.xml
patrick-schweizer: fixed to generator

features/enlinkd/org.opennms.features.enlinkd.topogen/src/test/java/org/opennms/enlinkd/topogen/topology/UndirectedPairGeneratorTest.java
patrick-schweizer: fixed

features/enlinkd/org.opennms.features.enlinkd.topogen/pom.xml
patrick-schweizer: @mvrueden , @j-white uhm what do you want me to change here?

j-white: Let's manually define the feature definition in `container/features/src/main/resources/features.xml` instead of using this plugin to generate it.

features/enlinkd/org.opennms.features.enlinkd.shell/src/main/java/org/opennms/features/enlinkd/shell/GenerateTopologyCommand.java
j-white: Just to clarify, if there are existing nodes and links in the database, and we run the generation command, followed by the associated delete, the database should ideally have the same state as it did *before* the generation command was ran.

patrick-schweizer: understand, I'm on it...

features/enlinkd/org.opennms.features.enlinkd.topogen/pom.xml
patrick-schweizer: done.

patrick-schweizer:
@j-white, @mvrueden, could you take another look?
I fixed your comments and cleaned up the code.

features/enlinkd/org.opennms.features.enlinkd.topogen/pom.xml
patrick-schweizer: hm but now it doesn't compile anymore... Anything else I need to change?

features/enlinkd/org.opennms.features.enlinkd.generator/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+    <!--
+     | uncomment to add all imported (non-local) bundles to your compilation classpath
+    <dependency>
+      <type>pom</type>
+      <groupId>${parent.groupId}</groupId>
+      <artifactId>provision</artifactId>
+      <optional>true</optional>
+    </dependency>
+    -->
+    <dependency>
```
j-white: This dependency should not be necessary.

features/enlinkd/org.opennms.features.enlinkd.shell/src/main/java/org/opennms/features/enlinkd/shell/GenerateTopologyCommand.java
```diff
@@ -0,0 +1,122 @@
...
+@Service
+public class GenerateTopologyCommand implements Action {
+
+    @Option(name = "--nodes", description = "generate <N> OmnsNodes.")
+    private Integer amountNodes;
+
+    @Option(name = "--elements", description = "generate <N> CdpElements")
+    private Integer amountElements;
+
+    @Option(name = "--links", description = "generate <N> CdpLinks")
```
j-white: Are these always CDP links? They don't seem to be. If so, I'd call it --cdp-links rather than --links.

features/enlinkd/org.opennms.features.enlinkd.shell/src/main/java/org/opennms/features/enlinkd/shell/GenerateTopologyCommand.java
```diff
@@ -0,0 +1,122 @@
...
+ * Install: feature:install opennms-enlinkd-shell
+ */
+@Command(scope = "enlinkd", name = "generate-topology", description = "Creates a linkd topology")
+@Service
+public class GenerateTopologyCommand implements Action {
+
+    @Option(name = "--nodes", description = "generate <N> OmnsNodes.")
+    private Integer amountNodes;
+
+    @Option(name = "--elements", description = "generate <N> CdpElements")
```
j-white: Are these always CDP elements? If so, I'd call it `--cdp-elements` rather than `--elements`.

features/enlinkd/org.opennms.features.enlinkd.shell/src/main/java/org/opennms/features/enlinkd/shell/GenerateTopologyCommand.java
```diff
@@ -0,0 +1,122 @@
...
+    @Option(name = "--ipinterfaces", description = "generate <N> IpInterfaces but not more than amount snmp interfaces")
+    private Integer amountIpInterfaces;
+
+    @Option(name = "--topology", description = "type of topology (complete | ring | random)")
+    private String topology;
+
+    @Option(name = "--protocol", description = "type of protocol (cdp | isis | lldp | ospf)")
+    private String protocol;
+
+    @Option(name = "--delete", description = "delete generated toplogogy (OnmsNodes, XxElements, XxLinks, SnmpInterfaces, IpInterfaces)")
```
j-white: `toplogogy`  vs `topology`

features/enlinkd/pom.xml
```diff
@@ -17,5 +17,7 @@
     <module>adapters</module>
     <module>daemon</module>
     <module>tests</module>
+    <module>org.opennms.features.enlinkd.shell</module>
```
j-white: These new modules use fully qualified names whereas the existing modules use short ones.

features/enlinkd/org.opennms.features.enlinkd.generator/src/main/java/org/opennms/enlinkd/generator/protocol/Protocol.java
```diff
@@ -0,0 +1,176 @@
...
+        ArrayList<OnmsNode> nodes = new ArrayList<>();
+        for (int i = 0; i < amountNodes; i++) {
+            nodes.add(createNode(i, location, category));
+        }
+        return nodes;
+    }
+
+    protected OnmsNode createNode(int count, OnmsMonitoringLocation location, OnmsCategory category) {
+        OnmsNode node = new OnmsNode();
+        node.setId(count); // we assume we have an empty database and can just generate the ids
```
j-white: We can't assume that.

features/enlinkd/org.opennms.features.enlinkd.generator/src/main/java/org/opennms/enlinkd/generator/protocol/Protocol.java
```diff
@@ -0,0 +1,176 @@
...
+        node.setLabel("Node" + count);
```
j-white: The resulting nodes are not visible from the UI when viewing the node list - running the following SQL statemen makes them visible: `update nodes set nodetype = 'A';`.

Can we make sure they are visible by default?

features/enlinkd/pom.xml
mvrueden: Good point. I fixed the feature naming, but oversaw this. 
Probably also want to rename the `generator` module as it is also a fully qualified name.

features/enlinkd/org.opennms.features.enlinkd.generator/pom.xml
patrick-schweizer: fixed

features/enlinkd/org.opennms.features.enlinkd.generator/src/main/java/org/opennms/enlinkd/generator/protocol/Protocol.java
patrick-schweizer: true, not anymore. I changed it - JPA gives us now also the ids

features/enlinkd/org.opennms.features.enlinkd.generator/src/main/java/org/opennms/enlinkd/generator/protocol/Protocol.java
patrick-schweizer: done.

features/enlinkd/org.opennms.features.enlinkd.shell/src/main/java/org/opennms/features/enlinkd/shell/GenerateTopologyCommand.java
```diff
@@ -0,0 +1,122 @@
...
+import org.opennms.enlinkd.generator.TopologyGenerator;
+import org.opennms.enlinkd.generator.TopologyPersister;
+import org.opennms.enlinkd.generator.TopologySettings;
+import org.opennms.netmgt.dao.api.GenericPersistenceAccessor;
+
+/**
+ * Generate a enlinkd topology via karaf command.
+ * Install: feature:install opennms-enlinkd-shell
+ */
+@Command(scope = "enlinkd", name = "generate-topology", description = "Creates a linkd topology")
```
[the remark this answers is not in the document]
patrick-schweizer: :+1: done

features/enlinkd/org.opennms.features.enlinkd.shell/src/main/java/org/opennms/features/enlinkd/shell/GenerateTopologyCommand.java
patrick-schweizer: no they depend on the protocol. I changed the description.

features/enlinkd/org.opennms.features.enlinkd.shell/src/main/java/org/opennms/features/enlinkd/shell/GenerateTopologyCommand.java
patrick-schweizer: no, they depend on the protocol. I changed the description.

features/enlinkd/org.opennms.features.enlinkd.shell/src/main/java/org/opennms/features/enlinkd/shell/GenerateTopologyCommand.java
patrick-schweizer: fixed.

patrick-schweizer:
> I also noticed that when we click on `View Node Link Detailed Info` from the node page on one of these generated nodes, an exception is thrown. Can we take a further look and see if the problem is with the way we generate the links, or possibly something on that page? I would expect that page to render successfully.
It hast to do how we generate the Interfaces. So far they are only connected to the node. But in that page the following is queried: snmpInterfaceDao.findByNodeIdAndIfIndex(nodeid, link.getCdpCacheIfIndex()) => The CdpCacheIfIndex doesn't match and leads the outcome of the query to and thus a Nullpointer. I guess that was ok while we used it only for performance tests but now we have a more general solution => I will see that the matching is correct...

features/enlinkd/pom.xml
patrick-schweizer: fixed.

PR #2325, opened 2019-01-28
HZN-1423: Migrate to Bootstrap 4 and revert custom styling

mvrueden:
Here we migrate the existing bootstrap 3 styles to bootstrap 4.
While doing that we also revert a lot of custom stylings, so this is basically plain bootstrap 4 with some stylings. 

It would be nice if you guys could test this so we may be able to merge it at some point.

JIRA: https://issues.opennms.org/browse/HZN-1423

Special thanks to @Zottel92 who helped push this through while I was occupied somewhere else (-:


indigo423:
Tested the branch and looks good so far, @mvrueden had some issues and reverted #2320 which makes the layout more compact. For the reason it is more like a personal matter of taste I would not make it a showstopper. This would be the only thing I would vote for to re-apply this patch. Feedback is welcome.

mmahacek:
I like the new look, but agree with @indigo that compact headers should be included in the default. I see the point about not overcustomizing the base template, but I believe this setting will benefit users that don’t delve into the customization option.

mvrueden:
The reason why I reverted the styling was that it looked weird (on my system) and needed some more fine-tuning.

I am not against the change, we just have to find a better version of it.

opennms-webapp/src/main/webapp/login.jsp
```diff
@@ -108,8 +108,9 @@
     </div>
   </div>
 
-  <div class="" style="color: white; position: absolute; bottom: 0px; right: 0px; font-size: 3em; padding: 20pt 20pt 5pt 20pt">
-    <a href="https://github.com/OpenNMS/opennms.git" title="Fork us on Github"><i class="fa fa-github"></i></a>
+  <div class="" style="position: absolute; bottom: 0px; right: 0px; font-size: 3em; padding: 20pt 20pt 5pt 20pt">
+    <a href="https://docs.opennms.org" class="text-light" title="Show documentation"><i class="fa fa-book" aria-hidden="true"></i></a>
```
RangerRick: Could you make it point to https://docs.opennms.org/opennms/releases/latest/index.html rather than making them browse around to the latest version?

mvrueden: The reason why I pointed to docs.opennms.org is that pointing to latest will most likely not be the version the user is running. 

The following alternatives came to mind:
 - Point to the actual release. This AFAIK cannot be done
 - Point to a possible newer documentation. @RangerRick suggestion.
 - Point to the landing page. Current implementation.

I opted for the landing page. 
I am not bound to this option however.

RangerRick: Good point.  You could at least point to the *opennms* landing page, I suppose: http://docs.opennms.org/opennms/

The version one should be doable as long as it's not a snapshot version.  You could just generate the link.  Still, that's potentially dangerous if we're not validating things well.

indigo423: We are using a versioned link mechanism already in Admin -> Help, maybe we can reuse something:

screenshot 2019-01-29 at 18 51 29

indigo423:
> The reason why I reverted the styling was that it looked weird (on my system) and needed some more fine-tuning.
> 
> I am not against the change, we just have to find a better version of it.

Can you share some screenshots and mark what was weird, I can look into it? We have the same system so I was wondering :)

RangerRick:
Found a couple more things (sorry).

1. Menu highlights are not being constrained properly. (Also when it turns blue when you click) screen shot 2019-01-31 at 3 07 41 pm
2. On the geomaps the menu is behind other controls. screen shot 2019-01-31 at 3 09 04 pm

mvrueden:
Don't worry. I am at a point, where I just don't see the issues anymore ^^

mvrueden:
@RangerRick I addressed the issues you reported. That was only related to Geographical and Topology Map.

agalue:
event-table-layout

The table layout for the first panel on the event details page, the alarm detail page, the outage detail page, and the notification detail page look like the screenshot. Looks like everything is pushed/compressed on the right, and I believe it should be the opposite, to give more space to the values, or try to see if it is possible to look like the current layout.

mvrueden:
Thanks @agalue, I didn't see that, as this is a safari/chrome only thing. I will let you know, when I addressed it

mvrueden:
Should be fixed now.

fooker:
Too much blue, but I lick it!

mmahacek:
I don't disagree about the blue.  It is a little bright, but if I will be able to inject my own CSS, I will probably make mine closer to #337AB7.

RangerRick:
I've played with this some yesterday and nothing huge jumped out at me as a blocker.  I'm sure we can infinitely tweak things but it would be nice to get the theme change into some people's hands for feedback at this point.

+1

mvrueden:
I talked with Jessica and applied some more changes:
 - Link Color is no `#265A87`
 - Card-Headers are slightly bigger
 - Reworked header

**Notifications enabled**
screenshot 2019-02-12 at 18 37 45

**Notifications disabled**
Notifications disabled

Some examples

**Start Page**
screenshot 2019-02-12 at 19 01 08

**Search**
screenshot 2019-02-12 at 19 01 48

christianpape:
Vaadin apps still work as expected. Nice work! Looks fresh!

mvrueden:
General feedback is positive. There may be some smaller issues we need to attend, but we decided to merge this as soon as the branch goes green. Any follow up work will be done later.

Please Squash merge.

PR #2328, opened 2019-01-29
NMS-10514: Select Default location instead of the first alphabetically

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-10514

### DIFF
--- a/smoke-test/src/test/java/org/opennms/smoketest/AdminSnmpConfigForIpPageIT.java
+++ b/smoke-test/src/test/java/org/opennms/smoketest/AdminSnmpConfigForIpPageIT.java
@@ -70,7 +70,7 @@ public class AdminSnmpConfigForIpPageIT extends OpenNMSSeleniumTestCase {
             Assert.assertEquals(new Select(findElementById("lookup_location")).getOptions().size(), locationCount);
             Assert.assertEquals(new Select(findElementById("location")).getOptions().size(), locationCount);
 
-            // create new location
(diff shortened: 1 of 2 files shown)


smoke-test/src/test/java/org/opennms/smoketest/AdminSnmpConfigForIpPageIT.java
```diff
@@ -71,16 +71,21 @@ public void verifyLocationDropdown() throws IOException, InterruptedException {
...
             Assert.assertEquals(new Select(findElementById("location")).getOptions().size(), locationCount + 1);
+
+            // check that "Default" is still selected even if the location "ABC" is added, see issue NMS-10514
+            Assert.assertEquals(new Select(findElementById("lookup_location")).getFirstSelectedOption().getText(), "Default");
+            Assert.assertEquals(new Select(findElementById("location")).getFirstSelectedOption().getText(), "Default");
+
         } finally {
             if (created) {
-                sendDelete("/api/v2/monitoringLocations/Test", 204);
+                sendDelete("/api/v2/monitoringLocations/ABC", 204);
```
mvrueden: Is there a particular reason why `Test`  was changed to `ABC`? Wouldn't the test be the same with `Test`?

christianpape: Yes, exactly

mvrueden: Wouldn't that make a great comment in the test?

PR #2330, opened 2019-01-30
HZN-1467: Upgrade Elasticsearch test server to use v6.5.x

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1467


features/opennms-es-rest/pom.xml
```diff
@@ -69,56 +69,6 @@
...
-        <dependencies>
-          <dependency>
-            <groupId>org.elasticsearch</groupId>
-            <artifactId>elasticsearch</artifactId>
-            <!-- REPLACE THE FOLLOWING WITH THE ES VERSION USED IN TESTS -->
-            <version>6.2.4</version>
-          </dependency>
-        </dependencies>
-      </plugin>
-
```
mvrueden: I almost cried here. Tears of joy (-:

mvrueden:
LGTM. Should we upgrade the system test api as well? AFAIK they are using a different Elastic 6 version.

j-white:
The system-test-api does use an earlier version of ES 6 - I'd say we keep it as is for now.

Ideally we would control the version that the system test api uses directly from this repository, but the system tests aren't structured that way... yet...

PR #2331, opened 2019-01-30
NMS-10545: Handle empty Name-String correctly while creating a new Destination Path

Zottel92:
* JIRA: https://issues.opennms.org/browse/NMS-10545


mvrueden:
Looking at the code and the issue it is not clear which exception or what problem is beeing fixed here.

Besides this, a test should be provided verifying the patch and problem.

mvrueden:
It does not seem this is actively beeing worked on. Closing for now. Feel free to re-open when the review comments have been addressed.

PR #2336, opened 2019-01-31
NMS-10506: Optimize default event query on node page

j-white:
JIRA: https://issues.opennms.org/browse/NMS-10506

Include the eventtime in the ORDER BY clause to the query planner.

Add an IT that validates the list of returned events.

With these changes, the query for the events on the node page now loads in 0.5 milliseconds
instead of 6 seconds on a database with 16 million+ events.


RangerRick:
Hilarious that that's all it took.  LGTM

PR #2338, opened 2019-02-01
HZN-1449: Optimize implementation of  findLatestAcks() for systems with lots of acks

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1449

Add an index to acks table to speed up grouping by refId.

Also limit the findLatestAcks query by time to avoid loading/returning unnecessary entries.

### DIFF
--- a/opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/drools/DroolsAlarmContext.java
+++ b/opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/drools/DroolsAlarmContext.java
@@ -334,7 +334,7 @@ public class DroolsAlarmContext extends ManagedDroolsContext implements AlarmLif
                         // Let's log an error, and return some date for sanity
                         final LocalDateTime oneMonthAgoLdt = LocalDateTime.now().minusMonths(1);
                         final Date oneMonthAgo = Date.from(oneMonthAgoLdt.atZone(ZoneId.systemDefault()).toInstant());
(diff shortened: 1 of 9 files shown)


opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/drools/DroolsAlarmContext.java
```diff
@@ -319,7 +323,21 @@ public void handleNewOrUpdatedAlarm(OnmsAlarm alarm) {
...
+                    .map(OnmsAlarm::getFirstEventTime)
+                    .filter(Objects::nonNull)
+                    .min(Comparator.naturalOrder())
+                    .orElseGet(() -> {
+                        // We didn't find any dates - either the set is empty (in which case this function
+                        // wouldn't be called) or all the dates are null (which they shouldn't be.)
+                        // Let's log an error, and return some date for sanity
+                        final LocalDateTime oneMonthAgoLdt = LocalDateTime.now().minusMonths(1);
+                        final Date oneMonthAgo = Date.from(oneMonthAgoLdt.atZone(ZoneId.systemDefault()).toInstant());
+                        LOG.error("Could not find minimum alarm creation time for alarms: {}. Using: {}", oneMonthAgo);
```
mattixtech: Looks like this is missing the alarms as a parameter.

j-white: Good catch.

PR #2342, opened 2019-02-04
Update maven-war-plugin to 3.2.2

smith-opennms:
This PR bumps the Maven-War-Plugin version to 3.2.2 (was 2.6)

* https://issues.opennms.org/browse/HZN-1231


mvrueden:
Would be nice if the commits are prefixed with the issue number, in this case `HZN-1232:`.

PR #2344, opened 2019-02-04
Provide a Daemons Reload Page and harmonise Daemon Reload Event

mvrueden:
Here we provide a daemon reload page accessible from the admin page.

This PR sums up all the little pieces required to provide the page.

I am adding the `triage` flag for now to put it out there for testing. 

This is what it looks like
screenshot 2019-01-31 at 16 42 51

Special thanks to @Zottel92 for this.


opennms-ackd/src/main/java/org/opennms/netmgt/ackd/Ackd.java
```diff
@@ -313,49 +311,13 @@ public void handleAckEvent(Event event) {
...
-                }
-
-                LOG.debug("handleReloadConfigEvent: configuration reloaded.");
-
-                return;  //return here because we are done.
-            }
-
-        }
-        LOG.debug("handleReloadConfigEvent: reload event not for this daemon: {}; daemon specified is: {}", getName(), specifiedDaemon);
+        DaemonTools.handleReloadEvent(event, NAME, (e) -> {
```
mvrueden: Reminder to myself: Verify all (not only this) re-work or simplifications of the daemon reload.

smoke-test/src/test/java/org/opennms/smoketest/rest/DaemonRestServiceIT.java
```diff
@@ -0,0 +1,294 @@
...
+                .body("[5].internal", equalTo(false))
+                .body("[5].enabled", equalTo(true))
+                .body("[5].reloadable", equalTo(false))
+
+                .body("[6].name", equalTo("Correlator"))
+                .body("[6].internal", equalTo(false))
+                .body("[6].enabled", equalTo(false))
+                .body("[6].reloadable", equalTo(false))
+
+                // ...
```
mvrueden: What is this supposed to mean?

mvrueden:
There is more work to be done here before this can be merged.
Closing the PR for now.

PR #2345, opened 2019-02-05
OIA-4: Map collectors from integration api

cgorantla:
* JIRA: http://issues.opennms.org/browse/OIA-4

### DIFF
--- a/features/api-layer/src/main/java/org/opennms/features/apilayer/collectors/ServiceCollectorImpl.java
+++ b/features/api-layer/src/main/java/org/opennms/features/apilayer/collectors/ServiceCollectorImpl.java
@@ -1,0 +1,183 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2019 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2019 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License,
+ * or (at your option) any later version.
+ *
+ * OpenNMS(R) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with OpenNMS(R).  If not, see:
+ *      http://www.gnu.org/licenses/
+ *
+ * For more information contact:
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.features.apilayer.collectors;
+
+import java.net.InetAddress;
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+
+import org.opennms.integration.api.v1.collectors.CollectionRequest;
+import org.opennms.integration.api.v1.collectors.CollectionSet;
+import org.opennms.integration.api.v1.collectors.ServiceCollector;
+import org.opennms.integration.api.v1.collectors.ServiceCollectorFactory;
+import org.opennms.integration.api.v1.collectors.resource.CollectionSetResource;
+import org.opennms.integration.api.v1.collectors.resource.GenericTypeResource;
+import org.opennms.integration.api.v1.collectors.resource.IpInterfaceResource;
+import org.opennms.integration.api.v1.collectors.resource.NodeResource;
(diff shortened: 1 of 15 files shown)


features/collection/shell-commands/src/main/java/org/opennms/netmgt/collection/commands/CollectCommand.java
```diff
@@ -146,7 +146,7 @@ public Void execute() {
         final CompletableFuture<CollectionSet> future = locationAwareCollectorClient.collect()
                 .withAgent(agent)
                 .withSystemId(systemId)
-                .withCollector(collector)
+                .withCollectorClassName(className)
```
j-white: Is this change necessary? We already use the `ServiceCollectorRegistry` to lookup the collector above.

j-white: I guess it relates to the fact that the CollectorRequestBuilderImpl ends up with the proxy class name instead, if that's it we should add a comment.

features/collection/client-rpc/src/main/java/org/opennms/netmgt/collection/client/rpc/CollectorRequestBuilderImpl.java
```diff
@@ -120,7 +123,7 @@ public CollectorRequestBuilder withAttributes(Map<String, Object> attributes) {
         CollectorRequestDTO request = new CollectorRequestDTO();
         request.setLocation(target.getLocation());
         request.setSystemId(target.getSystemId());
-        request.setClassName(serviceCollector.getClass().getCanonicalName());
+        request.setClassName(className != null ? className : serviceCollector.getClass().getCanonicalName());
```
j-white: Can we add a comment explaining why we do this?

features/api-layer/src/main/java/org/opennms/features/apilayer/collectors/ServiceCollectorManager.java
```diff
@@ -0,0 +1,69 @@
...
+    public synchronized void onBind(ServiceCollectorFactory extension, Map properties) {
+        LOG.debug("bind called with {}: {}", extension, properties);
+        if (extension != null) {
+            extServiceRegistrationMap.computeIfAbsent(extension, (ext) -> {
+                final ServiceCollector mappedExt = map(ext);
+                final Hashtable<String, Object> props = new Hashtable<>();
+                // Make the service available to any Spring-based listeners
+                props.put("registration.export", Boolean.TRUE.toString());
+                // Registry needs type of collector class.
+                props.put("type", ext.getCollectorClassName());
```
j-white: We should keep the onBind logic in the InterfaceMapper and only extend what we need. Instead of overriding onBind completely I would add something like:
```
    @Override
    public Map<String, Object> getServiceProperties(ServicePollerFactory extension) {
        return ImmutableMap.<String,Object>builder()
                // Registry needs type of collector class.
                .put("type", extension.getCollectorClassName())
                .build();
    }
```

j-white: We can also refactor the `ServicePollerManager` to do the same.

features/api-layer/src/main/java/org/opennms/features/apilayer/collectors/ServiceCollectorImpl.java
```diff
@@ -0,0 +1,183 @@
...
+                    org.opennms.netmgt.collection.support.builder.GenericTypeResource genericResource =
+                            new org.opennms.netmgt.collection.support.builder.GenericTypeResource(nodeLevelResource, null, genericTypeResource.getInstance());
+                    addAttributes(collectionSetResource, builder, genericResource);
+                }
+            }
+            builder.withTimestamp(new Date(collectionSet.getTimeStamp()));
+            return builder.build();
+
+        } catch (InterruptedException | ExecutionException e) {
+            LOG.error("collect failed with ", e);
```
j-white: We should probably rethrow this as a CollectionException instead of just returning `CollectionStatus.UNKNOWN`.

features/api-layer/src/main/java/org/opennms/features/apilayer/collectors/ServiceCollectorImpl.java
```diff
@@ -0,0 +1,183 @@
...
+                    addAttributes(collectionSetResource, builder, nodeLevelResource);
+                } else if (resource.getResourceType().equals(Resource.Type.INTERFACE)) {
+                    IpInterfaceResource ipResource = (IpInterfaceResource) resource;
+                    NodeLevelResource nodeLevelResource = new NodeLevelResource(ipResource.getNodeResource().getNodeId());
+                    InterfaceLevelResource interfaceLevelResource = new InterfaceLevelResource(nodeLevelResource, ipResource.getInstance());
+                    addAttributes(collectionSetResource, builder, interfaceLevelResource);
+                } else if (resource.getResourceType().equals(Resource.Type.GENERIC)) {
+                    GenericTypeResource genericTypeResource = (GenericTypeResource) resource;
+                    NodeLevelResource nodeLevelResource = new NodeLevelResource(genericTypeResource.getNodeResource().getNodeId());
+                    // TODO: Get ResourceType from ResourceTypesDao
```
j-white: Could also try using the `DeferredGenericTypeResource` to avoid having to lookup the types.

cgorantla:
Thanks @j-white  for the review. Moved mapper to different class and added unit test for mapper.

PR #2346, opened 2019-02-05
NMS-10546: Encoding for HTTP request locale

mattixtech:
Added encoding for the HTTP request locale in our JSPs.

* JIRA: http://issues.opennms.org/browse/NMS-10546


mattixtech:
Note that request.getLocale() won't throw an NPE if the client omits the header as the server's default is returned in that case.

PR #2352, opened 2019-02-07
NMS-9633: Add more details to CollectionException

cgorantla:
Add specific exception message to  event "reason" and in turn "description"
* JIRA: http://issues.opennms.org/browse/NMS-9633


RangerRick:
LGTM

PR #2354, opened 2019-02-08
NMS-10499: Check JMS connectivity only when JMS is enabled for Sink/RPC

cgorantla:
Currently we check activemq connectivity on Minion/Sentinel even if kafka is enabled for Sink/RPC.
This will allow  `health:check` to explicitly check JMS connectivity only when ActiveMQ is enabled.

Also fix `sentinel-core` feature definition as I was having issues installing it.

* JIRA: http://issues.opennms.org/browse/NMS-10499

### DIFF
--- a/core/ipc/rpc/jms-impl/src/main/resources/OSGI-INF/blueprint/blueprint-rpc-server.xml
+++ b/core/ipc/rpc/jms-impl/src/main/resources/OSGI-INF/blueprint/blueprint-rpc-server.xml
@@ -48,4 +48,12 @@
     <reference-list id="rpcModulesRef" interface="org.opennms.core.rpc.api.RpcModule" availability="optional">
         <reference-listener bind-method="bind" unbind-method="unbind" ref="jmsRpcServerRouteManager"/>
     </reference-list>
+
+    <!-- JMS HealthCheck -->
+    <service interface="org.opennms.core.health.api.HealthCheck">
+        <bean class="org.opennms.distributed.jms.impl.JmsConnectionHealthCheck" >
+            <argument ref="blueprintBundleContext"/>
+            <argument value="RPC" />
+        </bean>
+    </service>
 </blueprint>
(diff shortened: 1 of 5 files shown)


container/features/src/main/resources/features.xml
```diff
@@ -1304,6 +1304,7 @@
 
         <bundle dependency="true">mvn:com.google.guava/guava/18.0</bundle>
         <bundle dependency="true">mvn:org.opennms.core/org.opennms.core.soa/${project.version}</bundle>
+        <bundle>mvn:org.opennms.core/org.opennms.core.sysprops/${project.version}</bundle>
```
mvrueden: Can we make this also `dependency=true`?

core/ipc/rpc/jms-impl/src/main/resources/OSGI-INF/blueprint/blueprint-rpc-server.xml
```diff
@@ -48,4 +48,12 @@
     <reference-list id="rpcModulesRef" interface="org.opennms.core.rpc.api.RpcModule" availability="optional">
         <reference-listener bind-method="bind" unbind-method="unbind" ref="jmsRpcServerRouteManager"/>
     </reference-list>
+
+    <!-- JMS HealthCheck -->
+    <service interface="org.opennms.core.health.api.HealthCheck">
+        <bean class="org.opennms.distributed.jms.impl.JmsConnectionHealthCheck" >
+            <argument ref="blueprintBundleContext"/>
+            <argument value="RPC" />
```
mvrueden: I don't think the differentiation is required as we only check if a connection to the jms is possible in general. 
This looks like a workaround to me to hide the fact, that we are exposing the same health-check twice.

cgorantla: Note that Kafka/ActiveMQ can be enabled either for Sink/RPC.  When user does health:check they should know exactly what they are doing and which configuration is wrong.  So this is very much needed

cgorantla:
Moved sentinel-core fix to #2355

cgorantla:
Since `minion-jms` feature can be uninstalled and there won't be health check for JMS, I m inclined to close this PR.  We should not have `minion-jms` feature enabled by default on minion, since there is already work going for karaf plugin feature, we can look at this later. Closing.

PR #2363, opened 2019-02-20
OIA-3: Map SNMP datacollection configuration from integration API

cgorantla:
Merge snmp data collection configuration exposed on the container to default data collection configuration.
Also merge resource types.

* JIRA: http://issues.opennms.org/browse/OIA-3


j-white:
:+1+

PR #2364, opened 2019-02-20
HZN-1482: Update the topology generator so that topology is sent to the TopologyDAO

mattixtech:
The purpose of this PR is to update the topology generator command so that the topology it creates is published to Kafka. This will help design/test consuming the link data from Kafka in OCE by providing a convenient way of sending test data end to end.

I've also removed the checked exceptions in the related topology code and replaced them with runtime exceptions as they were just being used for pre-condition violation which I think runtime exceptions are more suited for.

* JIRA: http://issues.opennms.org/browse/HZN-1482


features/enlinkd/adapters/common/src/main/java/org/opennms/netmgt/enlinkd/TopologyUpdater.java
```diff
@@ -103,8 +102,8 @@ public void register() {
             m_topologyDao.register(this);
             m_registered  = true;
             LOG.info("register: protocol:{}", getProtocol().getId());
-        } catch (OnmsTopologyException e) {
-            LOG.error("register: protocol:{} {}", e.getProtocol(),e.getMessage());
+        } catch (Exception e) {
+            LOG.error("register: {}", e.getMessage());
```
patrick-schweizer: is it ok to loose the stack trace? Alternative: LOG.error("register: {}", e);

patrick-schweizer:
Looks good to me!

features/enlinkd/generator/src/test/java/org/opennms/enlinkd/generator/TopologyGeneratorTest.java
```diff
@@ -0,0 +1,127 @@
...
+                .build();
+
+        generator.generateTopology(settings);
+
+        assertThat(numVerticesUpdated, is(equalTo(numNodes)));
+        assertThat(numPortsUpdated, is(equalTo(numPorts)));
+        assertThat(numEdgesUpdated, is(equalTo(numLinks / 2)));
+    }
+
+    public class TestOnmsTopologyDao implements OnmsTopologyDao {
```
mvrueden: If most of the code is not used, maybe consider using a `EasyMock.createNiceMock(...)` as it does the same and is less verbose.

features/topologies/service/api/src/main/java/org/opennms/netmgt/topologies/service/api/OnmsTopologyVertex.java
```diff
@@ -28,14 +28,13 @@
...
 
+public class OnmsTopologyVertex extends OnmsTopologyAbstractRef implements OnmsTopologyRef {
 
-    public static OnmsTopologyVertex create(String id,String label,String address,String iconKey) throws OnmsTopologyException {
-        if (id == null) {
-            throw new OnmsTopologyException("id is null, cannot create vertex");
-        }
-        return new OnmsTopologyVertex(id,label,address,iconKey);
+    public static OnmsTopologyVertex create(String id, String label, String address, String iconKey) {
+        return new OnmsTopologyVertex(Objects.requireNonNull(id, "id is null, cannot create vertex"), label, address,
```
mvrueden: Can we put the `Objects.requireNonNull(...)` at the beginning of the method? This makes it easier to read.

features/topologies/service/impl/src/main/java/org/opennms/netmgt/topologies/service/impl/OnmsTopologyDaoInMemoryImpl.java
```diff
@@ -107,26 +104,23 @@ public void unregister(OnmsTopologyUpdater updater) throws OnmsTopologyException
...
+                            "message status: %s. Updater not registered", message.getMessagebody().getId(), protocol,
+                    message.getMessagestatus()));
         }
         synchronized (m_consumers) {
             m_consumers
-                .stream()
-                .filter(consumer ->  consumer.getProtocols().contains(protocol))
-                .forEach(consumer -> consumer.consume(message));            
+                    .stream()
+                    .filter(consumer -> consumer.getProtocols().contains(protocol))
```
mvrueden: Is there a case, where `consumer.getProtocols()` can be null?

rssntn67:
Well, I have to say that the TopologyUpdates are sent when the topology is loaded by Topologyupdaters....so this maybe does not require any change.

mattixtech:
Will address feedback. Also looking into Antonio's suggestion to see if there is an easier way to do this via sending a reload event. Will move to triage for now.

mattixtech:
Antonio showed me how to use reloadTopology() rather than sending updates individually. I tested that and it seems to yield the same results with less work so I am going to switch to using that method instead.

features/topologies/service/impl/src/main/java/org/opennms/netmgt/topologies/service/impl/OnmsTopologyDaoInMemoryImpl.java
mattixtech: There isn't currently, but the method is not documented to imply null is not a valid return value, therefore I'll add null handling here.

features/topologies/service/api/src/main/java/org/opennms/netmgt/topologies/service/api/OnmsTopologyVertex.java
mattixtech: Will do.

features/enlinkd/adapters/common/src/main/java/org/opennms/netmgt/enlinkd/TopologyUpdater.java
mattixtech: You are right, we shouldn't lose the stack trace, I'll update these to output the trace in the log message.

features/enlinkd/generator/src/test/java/org/opennms/enlinkd/generator/TopologyGeneratorTest.java
mattixtech: This test class has been removed.

mattixtech:
In my latest change I've re-targeted to release-24.0.0. I've also changed to calling the topology reload directly rather than sending individual topology updates.

Instead of triggering the reload via an IPC message I decided to expose the reload functionality via OSGI as that seemed more convenient. The topology generator and topology delete commands will now both trigger a topology reload after they have executed which results in topology consumers (the kafka producer) getting updates for the topology delta.

container/features/src/main/resources/features.xml
```diff
@@ -990,9 +990,10 @@
         <bundle>mvn:org.opennms.features/org.opennms.features.api-layer/${project.version}</bundle>
     </feature>
     <feature name="opennms-enlinkd-shell" version="${project.version}" description="OpenNMS :: Features :: Enlinkd :: Shell">
-        <feature>opennms-dao</feature>
         <bundle>mvn:org.opennms.features.enlinkd/org.opennms.features.enlinkd.shell/${project.version}</bundle>
         <bundle>mvn:org.opennms.features.enlinkd/org.opennms.features.enlinkd.generator/${project.version}</bundle>
+        <feature>opennms-dao</feature>
+        <feature>org.opennms.features.enlinkd.service.api</feature>
```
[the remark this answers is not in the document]
mattixtech: Looks like it isn't needed. Removed.

PR #2365, opened 2019-02-21
NMS-10567: Avoid leaking websockets if server is down

fooker:
* JIRA: http://issues.opennms.org/browse/NMS-10567


indigo423:
@fooker Tested and the problem is solved with this changes.

PR #2370, opened 2019-02-26
HZN-1484: Be more relaxed on missing templates

fooker:
* JIRA: http://issues.opennms.org/browse/HZN-1484


agalue:
I installed the RPMs created for the branch associated with this PR on my Minion test machine, and the solution is working as expected for Netflow9. I even restarted the Minion, and the solution is still working.

Unfortunately, I don't have a way to test the other protocols, but from Netflow9's perspective, this patch works.

fooker:
@agalue Thanks for the feedback. The IPFIX protocol shares the same session handling and has a very similar implementation when it comes to parsing.

agalue:
@fooker you're welcome; and thanks for the clarification (and for working on this). From my point of view, this is ready to be merged.

j-white:
Can we add some form of test coverage for this? i.e. ensure that packet containing both a template-set and data-set is parsed properly, and subsequent packets use are parsed using the given template.

fooker:
> Just a general comment that applies throughout. There is several cases where MissingTemplateException and InvalidPacketException are declared as thrown but aren't actually thrown. I would remove those throws clauses since the interface already enforces the exception checking. This will result in less code churn.
> 
> For instance in `BooleanValue.parser()` and `DateTimeValue.parseWith*()...parse()`.

You're right. I've removed at least some of them.

fooker:
> Can we add some form of test coverage for this? i.e. ensure that packet containing both a template-set and data-set is parsed properly, and subsequent packets use are parsed using the given template.

Tests added.

fooker:
Merged latest `release-24.0.0`.

PR #2372, opened 2019-02-26
HZN-1474: Meta-Data

fooker:
* JIRA: http://issues.opennms.org/browse/HZN-1474

This is initial support for assigning arbitrary meta-data to nodes, interfaces and services.
The meta-data is a simple triple of strings containing a context, a key and the associated value. There can be an arbitrary number of meta-data elements assigned to each node, each interface and each service. The only restriction is, that the tuple of context and key must be unique in the element to which they are associated.
The meta-data is associated to the model during provisioning. Therefore, the provisioning model allows to specify the meta-data for nodes, interfaces and services, too. During provisioning, the meta-data is copied to the model and persisted in the database.

To make this user-visible, this PR also includes additions to the node, interface and service detail pages allowing to view the associated meta-data. The Requisition UI has been changed to edit the meta-data assigned to  a requisition. The edit function in the Requisition UI is limited to only edit the context called `requisition` by intention. All other contexts are reserved for future use by other provisioning-adapters and similar applications  like asset-data.

The first usage of the meta-data is an extension to the configuration of detectors, pollers and data-collections. They all use a small DSL which allows to access the meta-data associated to the elements they are working on. The syntax of this allows the usage of patterns like `${context:key|context_fallback:key_fallback|...|default}` in each parameter. The DSL allows to provide scopes which will determine the resolution order. The last scope will be queried first. If the scope does not contain the queried context-key-tuple, the next one will be queried. There will be additional scopes providing thing like the model details (i.e. the node label, ID, foreign source and ID, IP address, service name...). To test parameter expansion, there is also a karaf command which interpolates a string containing a pattern to the final result.

Docs are still missing but will follow soon.

Feedback to this is appreciated.


mvrueden:
Adding `triage` as docs are missing.

fooker:
> Adding `triage` as docs are missing.

Docs have been added

core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/views/interface.html
```diff
@@ -25,20 +25,49 @@ <h3 class="modal-title">Interface</h3>
...
-        <div class="input-group">
-          <div class="input-group-prepend">
-            <div class="input-group-text">
-              <i class="fa fa-pencil"></i>
-            </div>
+
+    <hr>
+    <div class="clearfix">
+      <button type="button" class="btn btn-sm btn-secondary pull-right" id="add-metadata" ng-click="addMetaData()" ng-disabled="intfForm.$invalid">Add Meta-Data</button>
+      <h4>Meta-Data</h4>
```
j-white: Can we add a way to collapse this? If there 10+ kvps, the modal get prretty long.

core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/views/service.html
```diff
@@ -0,0 +1,50 @@
...
+    <div class="form-group">
+      <label class="col-form-label" for="name">Name</label>
+      <input class="form-control" type="text" id="name" name="name" placeholder="A Service Name" ng-model="service.name" required valid-service
+             ng-class="{ 'is-invalid' : serviceForm.name.$invalid }">
+      <p ng-show="serviceForm.name.$invalid" class="invalid-feedback">A valid Service name is required, and must be unique within the interface.</p>
+    </div>
+
+    <hr>
+    <div class="clearfix">
+      <button type="button" class="btn btn-sm btn-secondary pull-right" id="add-metadata" ng-click="addMetaData()" ng-disabled="serviceForm.$invalid">Add Meta-Data</button>
```
j-white: This is failing for me when creating a new requisition:
```
TypeError: Cannot read property 'push' of undefined
    at e.$scope.addMetaData (Service.js:68)
    at fn (eval at compile (angular.js:15126), <anonymous>:4:224)
```

core/ipc/rpc/shell-commands/src/main/java/org/opennms/core/rpc/commands/MetaCommand.java
```diff
@@ -0,0 +1,120 @@
...
+import org.opennms.core.rpc.utils.mate.ContextKey;
+import org.opennms.core.rpc.utils.mate.EntityScopeProvider;
+import org.opennms.core.rpc.utils.mate.FallbackScope;
+import org.opennms.core.rpc.utils.mate.Interpolator;
+import org.opennms.core.rpc.utils.mate.Scope;
+import org.opennms.core.utils.InetAddressUtils;
+import org.opennms.netmgt.dao.api.NodeDao;
+import org.opennms.netmgt.model.OnmsNode;
+
+@Command(scope = "meta", name = "test", description = "Test Meta-Data replacement")
```
j-white: Can we add some example expression in the `--help` output of this command?

core/ipc/rpc/shell-commands/src/main/java/org/opennms/core/rpc/commands/MetaCommand.java
```diff
@@ -0,0 +1,120 @@
...
+@Service
+public class MetaCommand implements Action {
+    @Reference
+    public NodeDao nodeDao;
+
+    @Reference
+    public EntityScopeProvider entityScopeProvider;
+
+    @Option(name = "-n", aliases = "--node-id", description = "Node Id", required = true, multiValued = false)
```
j-white: Can we support node ID or FS:FID?

core/ipc/rpc/shell-commands/src/main/java/org/opennms/core/rpc/commands/MetaCommand.java
```diff
@@ -0,0 +1,120 @@
...
+    private int nodeId;
+
+    @Option(name = "-i", aliases = "--interface-address", description = "Ip Interface Address", required = false, multiValued = false)
```
j-white: Ip vs IP

core/ipc/rpc/shell-commands/src/main/java/org/opennms/core/rpc/commands/MetaCommand.java
```diff
@@ -0,0 +1,120 @@
...
+    private String interfaceAddress;
+
+    @Option(name = "-s", aliases = "--service-name", description = "Service name", required = false, multiValued = false)
+    private String serviceName;
+
+    @Argument(index = 0, name = "expression", description = "Expression to use", required = true, multiValued = false)
```
j-white: Can we make this optional, and just show the meta-data if no expression is set?

opennms-doc/guide-development/src/asciidoc/text/rest/meta-data.adoc
```diff
@@ -0,0 +1,56 @@
+==== Meta-Data
+
+The actual Meta-Data of nodes, interfaces and services can be queried, added, modified or deleted via ReST endpoints.
+
+===== GETs (Reading Meta-Data)
+
+[options="header", cols="5,10"]
+|===
+| Resource                                                                                   | Description
+| `api/v2/nodes/{id}/metaData`                                                               | Get the list of meta-data associated with this node.
```
j-white: Can we make the paths use lowercase `metadata` instead of `metaData`?

features/themes/onms-default-theme/src/main/java/VAADIN/themes/opennms/styles.css
```diff
@@ -0,0 +1,8515 @@
+@import url(../reindeer/legacy-styles.css);
```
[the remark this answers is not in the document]
fooker: Reverted.

core/schema/src/main/liquibase/24.0.0/changelog.xml
```diff
@@ -341,5 +341,64 @@
     </sql>
   </changeSet>
 
-</databaseChangeLog>
+  <changeSet author="jwhite" id="24.0.0-node-metadata">
```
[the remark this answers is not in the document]
fooker: Jut the fun parts :)

core/ipc/rpc/shell-commands/src/main/java/org/opennms/core/rpc/commands/MetaCommand.java
christianpape: done

core/ipc/rpc/shell-commands/src/main/java/org/opennms/core/rpc/commands/MetaCommand.java
christianpape: done

core/ipc/rpc/shell-commands/src/main/java/org/opennms/core/rpc/commands/MetaCommand.java
christianpape: done

core/ipc/rpc/shell-commands/src/main/java/org/opennms/core/rpc/commands/MetaCommand.java
christianpape: done

opennms-doc/guide-development/src/asciidoc/text/rest/meta-data.adoc
christianpape: done

core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/views/interface.html
fooker: Collapse added to meta-data and service list. I'm not sure about the right place for the custom CSS

core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/views/service.html
fooker: Fixed.

core/ipc/rpc/utils/src/main/java/org/opennms/core/rpc/utils/mate/EntityScopeProviderImpl.java
```diff
@@ -0,0 +1,129 @@
...
+        final Scope metaDataScope = this.transactions.execute((tx) -> {
+            final OnmsNode node = nodeDao.get(nodeId);
+            if (node == null) {
+                return EmptyScope.EMPTY;
+            }
+
+            return new SimpleScope(transform(node.getMetaData()));
+        });
+
+        // TODO: Add context "node" {id, label, ...}
```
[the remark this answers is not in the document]
fooker: Removed. See https://issues.opennms.org/browse/NMS-10598

fooker:
This should go into release-24.0.0. Rebasing is in progress...

mvrueden:
This PR is targeted for `develop`. The version of `develop` is `25.0.0-SNAPSHOT`. However looking at the root pom.xml of `features/node-metadata` shows version `24.0.0-SNAPSHOT`. As Github allows me to hit the merge button, something is fishy here. Can you please ensure, that you are either rebased against latest `develop` or have merged `develop` into this branch?

fooker:
@mvrueden as discussed in last sprin meating, this is still targetet for 24.0.0. I've rebased it accordingly. I've missed to change the base of the PR after the rebasing, which I'VE corrected now.

mvrueden:
Thank you @fooker for the clarification.

I find the integration in the requisition UI a bit confusing.
The node meta data and the ip interface meta data is working, but when it comes to the service meta data it gets confusing and complicated.

Here are some thoughts, what we could try instead. 
Make the general popup bigger. 
There is no need to only cover about 20% of the screen.
Convert the IP Interface details to a `form-row`. 
That way we gain some room below.
Due to the now bigger popup, we split the interface area in two columns. 
The left side is for ip interface details (ip, description, type). 
The right side is for the meta data.

The services for the interface we layout the same.
Clicking on add service adds a service with the name and also an `add metadata` button.
The meta data then is shown left to the service, instead of in a separate window.

This shows all the data at once.
Each section can be made toggleable if needed.
We can also try adding the fontawesome icon `tag` instead of using the `meta data` title and use tooltips instead and see if that works.

I haven't implemented it, but I will provide a small mockup of what I mean.

Besides this, I was able to create a meta data with the same key and value twice.
Should this be possible?

Also a general question, should we make the meta data location aware?
Sure, I can achieve this by context, but maybe, a differentiation between location and context is necessary. Let's say for example I want to overwrite a certain key, just for location `FULDA`?

mvrueden:
Here is a first draft of what I mean.

Screenshot 2019-03-12 at 13 49 11

fooker:
> Also a general question, should we make the meta data location aware?
> Sure, I can achieve this by context, but maybe, a differentiation between location and context is necessary. Let's say for example I want to overwrite a certain key, just for location FULDA?

I'm not sure if I understand what you want to archive... Currently, the meta-data is strictly bound to a node and can be overridden by interfaces and services. The nodes location is determined by the requisition the node is created in. Therefore, a node can only be in a single location at a time and there is only one possible value for a context-key-tuple at time.

On the other hand, it may be useful to extend the concept of hierarchical meta-data to the requisition. By doing so, one can specify meta-data which is shared by all nodes in the same requisition. But this would change the way the meta-data is resolved from usage-time to provision-time as the provision itself is not available in the database model.

fooker:
@mvrueden: Thank you for your proposed changes on the requisition UI. I'm also not very happy with the way these things work right now. IMO, the best way to solve this is to get rid of the modal popup boxes completely and have some kind of master/details view where you can select an node, interface or service on the right column and can changes the details of the selected element on the left side. But this would increase the amount of changes in this PR even more and should not be done so late in the release of 24.0.0 and foundation 2019.

The main problem I was trying to avoid with my changes it the scaleability issue as the UI tends to become clumsy if a node has 10 interfaces with 10 services each. But nevertheless, I will give your concept a try.

fooker:
I've removed the requisition UI changes from this PR and moved them to a new one: #2394. (cc @mvrueden)

PR #2376, opened 2019-03-04
NMS-10593: Add a timeout when trying to acquire the locks in order to prevent deadlocks.

j-white:
JIRA: https://issues.opennms.org/browse/NMS-10593

Rely on the snapshot handling to update the context if acquiring the lock fails.

### DIFF
--- a/opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/drools/DroolsAlarmContext.java
+++ b/opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/drools/DroolsAlarmContext.java
@@ -41,13 +41,18 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.concurrent.Callable;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/drools/DroolsAlarmContext.java
```diff
@@ -301,11 +309,33 @@ public void handleNewOrUpdatedAlarm(OnmsAlarm alarm) {
...
+                    callback.accept(alarmId, reductionKey);
+                } finally {
+                    getLock().unlock();
+                }
+            } else {
+                LOG.warn("Failed to acquire Drools session lock within {}ms. " +
+                                "{} for alarm with id={} and reduction-key={} will not be immediately reflected in the context.",
+                        LOCK_TIMEOUT_MS, action, alarmId, reductionKey);
+            }
+        } catch (InterruptedException e) {
```
mattixtech: Should propagate the interruption here
`Thread.currentThread().interrupt();`

j-white: Good catch. Fixed.

PR #2380, opened 2019-03-05
NMS-10539: Upgrade Karaf to 4.2.3

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-10539

Besides upgrading to Karaf 4.2.3, due to some internal changes, we now can no longer use the`pax-whiteboard` feature. This means I had to remove our forked Apache Felix Http Bridge, which is a good thing. However, in order to get our Request forwarding to work (the way it is working now), I provided a new solution, which is kinda duplicating already existing code (from the Apache Felix Http Whiteboard project). The provided solution will work for us, but is not fully OSGi-compatible. For example direct calls to `httpService.registerServlet(...)` will not be forwarded. I opted for now to leave it as is, as we have to revisit this probably anyways, when looking into OSGi 7 and it's JAX-RS Rest support. So instead of the `pax-whiteboard` project, we are now using the `Apache Felix HTTP Bridge` project. With that we also have to use OSGi specified properties (`servlet.init.widgetset` vs `init.widgetset`, for example). 

While working on this, I encountered also the following issues, which may not be limited to the Karaf Upgrade:
 - The integration-api maven artifact resolutions wasn't downloading the generic `-SNAPSHOT` versions, but a concrete snapshot version. `my-bundle-SNAPSHOT.jar` vs `my-bundle-20190301.0000.jar`.
 - The OsgiUIProvider for Vaadin UIs had to be rewritten.
 - Found an issue with `@PreserveOnRefresh` for Vaadin. See https://issues.opennms.org/browse/NMS-10601.
 - Had to upgrade the PluginManager to work with the `Apache Felix HTTP Bridge`.


j-white:
🥇 

Thanks for pushing this forward @mvrueden.

PR #2384, opened 2019-03-06
NMS-10516: Fix NumberAdapter which limits values to MAX_INT

cgorantla:
* JIRA: http://issues.opennms.org/browse/NMS-10516

### DIFF
--- a/core/xml/src/main/java/org/opennms/core/xml/DoubleAdapter.java
+++ b/core/xml/src/main/java/org/opennms/core/xml/DoubleAdapter.java
@@ -42,9 +42,6 @@ public class DoubleAdapter extends XmlAdapter<String, Double> {
         if (v == null) {
             return null;
         }
-        if (!v.isNaN() && !v.isInfinite() && v == Math.rint(v)) {
-            return String.valueOf(v.intValue());
-        }
         return v.toString();
     }
 
--- a/core/xml/src/test/java/org/opennms/core/xml/NumberAdapterTest.java
+++ b/core/xml/src/test/java/org/opennms/core/xml/NumberAdapterTest.java
@@ -1,0 +1,57 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2019 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2019 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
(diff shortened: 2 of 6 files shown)


core/xml/src/test/java/org/opennms/core/xml/NumberAdapterTest.java
```diff
@@ -0,0 +1,44 @@
...
+import org.junit.Test;
+
+public class NumberAdapterTest {
+
+    @Test
+    public void testNumberAdapter() throws Exception {
+
+        NumberAdapter adapter = new NumberAdapter();
+        String value = adapter.marshal(4.65450971582E11);
+        assertTrue(value.equals("4.65450971582E11"));
```
j-white: We should use `assertEquals` instead of `assertTrue` here.

Can we also verify that unmarshalling the resulting string gives us a similar result?

j-white:
Good find.

core/xml/src/main/java/org/opennms/core/xml/DoubleAdapter.java
```diff
@@ -42,9 +42,6 @@ public String marshal(final Double v) throws Exception {
         if (v == null) {
             return null;
         }
-        if (!v.isNaN() && !v.isInfinite() && v == Math.rint(v)) {
```
cgorantla: Not sure if there is any reason to find and return integer values.  So fixed this as well

j-white: I think there was a reason for doing this - iirc `rrdtool` will complain if we use any decimal point i.e. `1.0f` for counters.

cgorantla: Looks like this is getting used only in [DiscoveryConfiguration](https://github.com/OpenNMS/opennms/blob/develop/opennms-config-model/src/main/java/org/opennms/netmgt/config/discovery/DiscoveryConfiguration.java#L67).  @RangerRick  also thinks it is safe to change this as it was done to mimic castor behavior as we no longer use castor anywhere.

cgorantla: @j-white  Compared resource graphs on a node on Minion and a node on OpenNMS for counters specifically tcp* and ifHc* and they look similar.   So think we don't need to discard fractions like `x.0f` in marshalling.  Also unmarshalling converts numeric `x` -> `x.o` always.

cgorantla: With rrdtool as well, resource graphs looks similar.

PR #2388, opened 2019-03-09
Fix ack event: supporting ackAction

rssntn67:
NMS-10603:  Support AckAction Escalate and Clear

* JIRA: http://issues.opennms.org/browse/NMS-10603

The only supported action for ack event is Acknowledge.

Let ack event support also Escalate, Clear and Unacknowledge

Added tests

### DIFF
--- a/opennms-model/src/main/java/org/opennms/netmgt/model/OnmsAcknowledgment.java
+++ b/opennms-model/src/main/java/org/opennms/netmgt/model/OnmsAcknowledgment.java
@@ -122,7 +122,10 @@ public class OnmsAcknowledgment {
         for (Parm parm : parms) {
             final String parmValue = parm.getValue().getContent();
             if (!"ackAction".equals(parm.getParmName()) && 
-                !"ackType".equals(parm.getParmName()) && !"refId".equals(parm.getParmName()) && !"ackUser".equals(parm.getParmName()) ) {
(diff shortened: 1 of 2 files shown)


opennms-model/src/main/java/org/opennms/netmgt/model/OnmsAcknowledgment.java
```diff
@@ -121,25 +121,36 @@ public OnmsAcknowledgment(final Event e) throws ParseException {
         
         for (Parm parm : parms) {
             final String parmValue = parm.getValue().getContent();
-            
-            if (!"ackType".equals(parm.getParmName()) && !"refId".equals(parm.getParmName()) && !"user".equals(parm.getParmName()) ) {
+            if (!"ackAction".equals(parm.getParmName()) && 
+                !"ackType".equals(parm.getParmName()) && !"refId".equals(parm.getParmName()) && !"ackUser".equals(parm.getParmName()) ) {
```
j-white: Should we also continue to support the `user` parameter as well?

rssntn67: Sure!

rssntn67: Good Point

PR #2389, opened 2019-03-11
HZN-1485: Fix cleaning out  .m2 directory

cgorantla:
* JIRA: http://issues.opennms.org/browse/HZN-1485

### DIFF
--- a/tools/packages/minion/minion.spec
+++ b/tools/packages/minion/minion.spec
@@ -204,6 +204,11 @@ if [ -d "${ROOT_INST}/data" ]; then
     fi
 fi
 
+# Clean out .m2 directory
+if [ -d "${ROOT_INST}/.m2" ]; then
+   rm -rf "${ROOT_INST}/.m2"
+fi
+
 # Generate an SSH key if necessary
 if [ ! -f "${ROOT_INST}/etc/host.key" ]; then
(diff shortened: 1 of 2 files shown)


cgorantla:
Verified by installing minion from .rpm and checking that .m2 directory was removed.

tools/packages/minion/minion.spec
```diff
@@ -204,6 +204,11 @@ if [ -d "${ROOT_INST}/data" ]; then
     fi
 fi
 
+# Clean out .m2 directory
+if [ -d "${ROOT_INST}/.m2" ]; then
```
mvrueden: The Minion container should not install anything from the local maven repository. This is defined in `org.ops4j.pax.url.mvn.cfg`. All other jar's are loaded from a designated `tar.gz` file which has been expanded. 

Generically delete the `.m2` repository doesn't seem to be the right approach.
What are we trying to solve here?

cgorantla: The user who reported the issue seems to be upgrading from  older branch based on `drift` to  `snapshot 25`.  There is  `.m2` folder defined  [here](https://github.com/OpenNMS/opennms/commit/f492bea4c0fa34e260a3a3a09f451856663dd9f2)

mvrueden: Nevermind. I oversaw `${ROOT_INST}` here. I was suspecting `~/.m2`, which we should not delete (-:

j-white:
👍

PR #2368, opened 2019-02-23
HZN-1475: Extend topology generator and test suite to support bridge topology

patrick-schweizer:
* JIRA: https://issues.opennms.org/browse/HZN-1475

Our [continuous integration system](http://bamboo.internal.opennms.com:8085) will test and verify your changes.

@j-white , @rssntn67, could you take another look? Thanks!

What is in this PR?
1. Enhancement of the topology generator by the protocol "bridge". The bridge protocol is a bit different in that it contains multiple node types. The generator builds a tree consisting of hosts, bridges, segments and" Mac/IpAdresses on Segment without node". The tree is repeated depending on the amount of nodes (multiple of 10).
2. The generated tree is tested in LinkdTopologyProviderTestIT.testBridge()
3. I introduced a "BridgeBuilder" that allows for easier generating of such topologies. It is sufficient for my use case but could be enhanced in the future. It tries to abstract the low level stitching together of the different database entries.

Thanks again to @rssntn67 who provided me with a working example and lots of explanation.

PR #2390, opened 2019-03-11
HZN-1475: Extend topology generator and test suite to support bridge topology - release 4.0.0

patrick-schweizer:
*  JIRA: https://issues.opennms.org/browse/HZN-1475

Our continuous integration system will test and verify your changes.

Replaces https://github.com/OpenNMS/opennms/pull/2368

What is in this PR?

* Enhancement of the topology generator by the protocol "bridge". The bridge protocol is a bit different in that it contains multiple node types. The generator builds a tree consisting of hosts, bridges, segments and" Mac/IpAdresses on Segment without node". The tree is repeated depending on the amount of nodes (multiple of 10).
* The generated tree is tested in LinkdTopologyProviderTestIT.testBridge()
* I introduced a "BridgeBuilder" that allows for easier generating of such topologies. It is sufficient for my use case but could be enhanced in the future. It tries to abstract the low level stitching together of the different database entries.

Thanks again to @rssntn67 who provided me with a working example and lots of explanation.


j-white:
Nice work. This will go a long way towards helping us maintain and test the bridge topology code.

image

PR #2396, opened 2019-03-15
HZN-1490: Provide new GraphProvider implementation for the ApplicationTopologyProvider

patrick-schweizer:
* JIRA: https://issues.opennms.org/browse/HZN-1490


features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericEdge.java
```diff
@@ -62,6 +64,13 @@ public void setProperty(String key, Object value) {
         properties.put(key, value);
     }
 
+    @Override
```
patrick-schweizer: @mvrueden I had to add this to not end in an endless recursion for the toString() method

features/graph/provider/application/src/main/java/org/opennms/netmgt/graph/provider/application/ApplicationGraphProvider.java
```diff
@@ -0,0 +1,123 @@
...
+            initialized = true;
+        }
+        return graph;
+    }
+
+    @Override
+    public GraphInfo<?> getGraphInfo() {
+        final DefaultGraphInfo graphInfo = new DefaultGraphInfo(TOPOLOGY_NAMESPACE, SimpleVertex.class);
+        graphInfo.setLabel("Application Graph");
+        graphInfo.setDescription("This Topology Provider displays the hierarchy of TODO."); // TODO Patrick
```
patrick-schweizer: Uhm what does it actually display? Applications with its monitored services?

features/graph/provider/application/src/main/java/org/opennms/netmgt/graph/provider/application/ApplicationGraphProvider.java
```diff
@@ -0,0 +1,123 @@
...
+        return graphInfo;
+    }
+
+    Graph<?, ?> createGraph() {
```
patrick-schweizer: @mvrueden I am not sure about the concept behind the Generics here. I guess we should give it concrete classes...

features/graph/provider/application/src/main/java/org/opennms/netmgt/graph/provider/application/ApplicationGraphProvider.java
```diff
@@ -0,0 +1,123 @@
...
+
+    @Override
+    public String getName() {
+        return getClass().getSimpleName();
+    }
+
+    @Override
+    public void onEvent(Event e) {
+        // BSM has been reloaded, force reload
+        // TODO: patrick find out if we need to listen to an event and for which one?
```
patrick-schweizer: to be discussed with @mvrueden

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericEdge.java
mvrueden: Yeah, that is a known bug. Please add a TODO MVR or TODO PS in order to revert these, when it is fixed.

features/graph/provider/application/src/main/java/org/opennms/netmgt/graph/provider/application/ApplicationGraphProvider.java
mvrueden: > This Topology Provider displays all defined Applications and their calculated states.
Source https://github.com/opennms/opennms/blob/09a8e5126ee824d177b4f8b322072a111384f729/features/topology-map/plugins/org.opennms.features.topology.plugins.topo.application/src/main/resources/OSGI-INF/blueprint/blueprint.xml#L13-L17

Also refer to its implementation https://github.com/opennms/opennms/blob/47a0b591e9dd04d6b3bef3d3765bac13eb12cf35/features/topology-map/plugins/org.opennms.features.topology.plugins.topo.application/src/main/java/org/opennms/features/topology/plugins/topo/application/ApplicationTopologyProvider.java#L55

features/graph/provider/application/src/main/java/org/opennms/netmgt/graph/provider/application/ApplicationGraphProvider.java
mvrueden: Yes.
But keep in mind that only the `TopologyService` must invoke `createGraph()` from a provider. 
Even if this were more concrete, a developer may not be able to leverage that.

features/graph/provider/application/src/main/java/org/opennms/netmgt/graph/provider/application/ApplicationGraphProvider.java
mvrueden: Not necessary. It looks like you were copying the GraphProvider for BSM.
The idea behind the `GraphProvider`/`GraphContainerProvider` is, that it knows how to load a graph/container. On the other hand, the `GraphService` is invoking the loading methods accordingly. Also it should implement caching strategies. However each GraphProvider may need other caching strategies:

 - GraphML: Only load once, never reload
 - BSM: Only load when changed (thus the `onEvent(...)`
 - Application: Reload when cache is evicted (e.g. by default 5 Minutes)
 - Vmware: Maybe listen for updates and update adhoc (no reloading needed, only load once)
 - Linkd: See Vmware
 - and so on

However at this state caching is not implemented.
I decided to implement the "only reload on change" manually in BSM to add according TODOs to implement proper caching strategies at some point.

features/graph/provider/application/src/main/java/org/opennms/netmgt/graph/provider/application/ApplicationGraphProvider.java
```diff
@@ -0,0 +1,123 @@
...
+        Objects.requireNonNull(applicationDao);
+        this.applicationDao = applicationDao;
+        LOG.debug("Creating a new {} with namespace {}", getClass().getSimpleName(), TOPOLOGY_NAMESPACE);
+    }
+
+    // TODO MVR We may need some kind of caching strategy implementation allowing each provider to deal with reloads individually if they so choose
+    @Override
+    public Graph<?, ?> loadGraph() {
+        // TODO MVR this is not thread safe
+        if (!initialized) {
```
mvrueden: See my comment for `onEvent(...)`.

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericEdge.java
patrick-schweizer: done

features/graph/provider/application/src/main/java/org/opennms/netmgt/graph/provider/application/ApplicationGraphProvider.java
patrick-schweizer: fixed.

features/graph/provider/application/src/main/java/org/opennms/netmgt/graph/provider/application/ApplicationGraphProvider.java
patrick-schweizer: ok, I added caching to ApplicationGraphProvider for now which we can remove again later when we have a generic caching in place

features/graph/provider/application/src/main/java/org/opennms/netmgt/graph/provider/application/ApplicationGraphProvider.java
patrick-schweizer: ok, thanks. I removed it.

features/graph/provider/application/src/main/java/org/opennms/netmgt/graph/provider/application/ApplicationGraphProvider.java
mvrueden: Let's not add caching for each provider.
Everything goes into the dedicated feature branch and will only be merged when finished.
As caching strategies are a core concept it needs to be implemented before we can call this feature finished. There is no point in adding additional caching strategies for the providers, as that is dead code anyways.

features/topology-map/plugins/org.opennms.features.topology.plugins.topo.application/src/main/java/org/opennms/features/topology/plugins/topo/application/ApplicationTopologyProvider.java
```diff
@@ -62,6 +66,12 @@ public ApplicationTopologyProvider(ApplicationDao applicationDao) {
 
     private void load() {
         graph.resetContainer();
+        // TODO: Patrick: discuss with mvr: why would we transfer one graph into another? I am missing something here?
+        // my assumption was that we use the graph coming from the graph service directly...
+
+        // TODO: Patrick: will be generic later
```
patrick-schweizer: @mvrueden let's discuss when you are back. I have the feeling I am missing something here...

features/graph/provider/application/src/main/java/org/opennms/netmgt/graph/provider/application/ApplicationVertex.java
```diff
@@ -0,0 +1,127 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2015 The OpenNMS Group, Inc.
```
mvrueden:
```suggestion
 * Copyright (C) 2019-2019 The OpenNMS Group, Inc.
```

features/graph/provider/application/src/main/java/org/opennms/netmgt/graph/provider/application/ApplicationVertex.java
```diff
@@ -0,0 +1,127 @@
...
+ * OpenNMS(R) is Copyright (C) 1999-2015 The OpenNMS Group, Inc.
```
mvrueden:
```suggestion
 * OpenNMS(R) is Copyright (C) 1999-2019 The OpenNMS Group, Inc.
```

features/graph/provider/application/src/test/java/org/opennms/netmgt/graph/provider/application/IdGenerator.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * For more information contact:
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.netmgt.graph.provider.application;
+
+public class IdGenerator {
```
mvrueden: Please, no. 
This is no hard task, and can be much easier achieved by using `id++` or something like this.

features/topology-map/plugins/org.opennms.features.topology.plugins.topo.application/src/main/java/org/opennms/features/topology/plugins/topo/application/ApplicationTopologyProvider.java
```diff
@@ -62,6 +66,12 @@ public ApplicationTopologyProvider(ApplicationDao applicationDao) {
 
     private void load() {
         graph.resetContainer();
+        // TODO: Patrick: discuss with mvr: why would we transfer one graph into another? I am missing something here?
+        // my assumption was that we use the graph coming from the graph service directly...
```
mvrueden: Yeah, we should do that at some point, but we are not there yet.

features/topology-map/plugins/org.opennms.features.topology.plugins.topo.application/src/main/java/org/opennms/features/topology/plugins/topo/application/GuiApplicationVertex.java
```diff
@@ -30,39 +30,56 @@
 
 import org.opennms.features.topology.api.topo.AbstractLevelAwareVertex;
 import org.opennms.features.topology.api.topo.LevelAware;
+import org.opennms.netmgt.graph.provider.application.ApplicationVertex;
 import org.opennms.netmgt.model.OnmsApplication;
 import org.opennms.netmgt.model.OnmsMonitoredService;
 import org.opennms.netmgt.model.OnmsServiceType;
 
-public class ApplicationVertex extends AbstractLevelAwareVertex implements LevelAware {
+public class GuiApplicationVertex extends AbstractLevelAwareVertex implements LevelAware {
```
patrick-schweizer: @mvr we should discuss naming here. This class was previously called ApplicationVertex. But we have now a org.opennms.netmgt.graph.provider.application.ApplicationVertex. To avoid confusion I renamed it to GuiApplicationVertex. Not sure if that is a good name.
Maybe this class should be names BackendApplicationVertex since it belongs to the BackendGraph (although I don't really like this name since it holds Frontend specific attributes such as Tooltips).

mvr: This was meant for @mvrueden 

(I think you've just introduced me to who my competition is for the mvr username on other websites!)

patrick-schweizer: @mvr sorry!

features/graph/provider/application/src/test/java/org/opennms/netmgt/graph/provider/application/IdGenerator.java
patrick-schweizer: fixed

features/topology-map/plugins/org.opennms.features.topology.plugins.topo.application/src/main/java/org/opennms/features/topology/plugins/topo/application/GuiApplicationVertex.java
mvrueden: @mvr everyone needs a villain :)

container/features/src/main/resources/features.xml
```diff
@@ -1216,8 +1216,12 @@
     </feature>
     <feature name="opennms-topology-runtime-application" version="${project.version}" description="OpenNMS :: Features :: Topology :: Features :: Application">
         <details>Generated using Pax-Construct</details>
+        <!-- TODO: patrick , mvr fix dependency problems -->
```
patrick-schweizer: @mvrueden it seems I can't get the dependencies right here. Maybe you want to take a look?

features/topology-map/plugins/org.opennms.features.topology.plugins.topo.application/src/main/java/org/opennms/features/topology/plugins/topo/application/GuiApplicationVertex.java
mvrueden: Yeah, the naming is bad.
The problem here is, that we have various `Graph` classes:
 - `org.opennms.features.topology.api.Graph`. This is actually the UI representation of the Graph
- `org.opennms.features.topology.api.topo.Graph`. The graph provided by the `TopologyProvider`. To not have two classes named `Graph`, I decided to name this `BackendGraph`. Probably better to name the UI part, just `UIGraph` or `GraphSnippet`.
 - `org.opennms.netmgt.graph.api.Graph`. The new `Graph` interface actually representing a Graph.

The naming `GuiApplicationVertex` is misleading and not correct. 
It is an `ApplicationVertex` within the `ApplicationTopologyProvider`. 
We already have an `ApplicationVertex` in `ApplicationGraphProvider`, so I decided to for now name it `OnmsApplicationVertex`. Alternatively we could bane it `TopologyApplicationVertex` or something else.
We should come up with a consistent naming scheme, as we will have the same problem with all the other providers as well. We could also just name them the same and distinguish only via full name `a.A` vs `b.A`

container/features/src/main/resources/features.xml
mvrueden: I fixed that with various other things.

features/topology-map/plugins/org.opennms.features.topology.plugins.topo.application/src/main/java/org/opennms/features/topology/plugins/topo/application/ApplicationSearchProvider.java
```diff
@@ -92,8 +92,9 @@ public boolean supportsPrefix(String searchPrefix) {
         bldr.limit(10);
         Criteria dbQueryCriteria = bldr.toCriteria();
 
+        // TODO: patrick do we want to replace the dao call also with a call to GraphService?
```
mvrueden: You are right, at some point we want to call the `GraphService` for this calculation.
To implement this, we need search capabilities on the `GraphService` side, so we should implement that next.

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericVertex.java
```diff
@@ -50,6 +50,7 @@ public GenericVertex(GenericVertex copyMe){
     }
 
 
+    // TODO MVR this is a duplicat eand should probably be removed, as a vertex is already a vertexref, there is no need to have a `getVertexRef` method
     public GenericVertexRef getVertexRef() {
```
patrick-schweizer: @mvrueden we should discuss this.
I change that: The GenericVertexRef is a immutable final object - it is a composite key at the end and not the mother class of GenericVertex.
The advantage is that we can pass around that key without having the problem that some one casts it to a GenericVertex or changes things. I would argue that this makes a better system design - but let's discuss

mvrueden: I can see your point. If we want to do it that way, we should remove the inheritance and name the method `asVertexRef`. If we keep the inheritance, the vertex ref should be created using a copy constructor. However both is out of the scope of this issue/PR.

features/topology-map/plugins/org.opennms.features.topology.plugins.topo.application/src/main/java/org/opennms/features/topology/plugins/topo/application/GuiApplicationVertex.java
mvrueden: All the classes in `features/topology` are supposed to be "the old way" and all classes in `features/graph` are supposed to be "the new way". How about just prefix everything from "the old way" with `Legacy`. Then we don't have the naming collisions anymore.

mvrueden:
Cannot merge, as tests are failing.

PR #2397, opened 2019-03-15
NMS-10435: Refactor poller configuration to separate minion specific services.

cgorantla:
Minion nodes are special nodes added by default to establish Minion device Status.
These nodes by default have `Minion-Heartbeat` `Minion-RPC` `JMX-Minion` services.
Moving them to separate package and having different `down-time` model would yield quicker results when these nodes are down.

* JIRA: http://issues.opennms.org/browse/NMS-10435

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/poller-configuration.xml
+++ b/opennms-base-assembly/src/main/filtered/etc/poller-configuration.xml
@@ -242,14 +242,6 @@
          <parameter key="timeout" value="3000"/>
          <parameter key="rrd-repository" value="${install.share.dir}/rrd/response"/>
       </service>
-      <service name="JMX-Minion" interval="300000" user-defined="false" status="on">
-         <parameter key="port" value="18980"/>
-         <parameter key="retry" value="2"/>
-         <parameter key="timeout" value="3000"/>
-         <parameter key="rrd-repository" value="${install.share.dir}/rrd/response"/>
-         <parameter key="ds-name" value="jmx-minion"/>
-         <parameter key="use-foreign-id-as-system-id" value="true"/>
-      </service>
       <service name="JMX-Kafka" interval="300000" user-defined="false" status="on">
          <parameter key="port" value="9999"/>
          <parameter key="retry" value="2"/>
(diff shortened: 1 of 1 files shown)


opennms-base-assembly/src/main/filtered/etc/poller-configuration.xml
```diff
@@ -276,6 +268,24 @@
...
+      <downtime begin="300000" end="43200000" interval="300000"/><!-- 5m, 5m, 12h -->
+      <downtime begin="43200000" end="432000000" interval="600000"/><!-- 10m, 12h, 5d -->
+      <downtime begin="432000000" interval="3600000"/><!-- 1h, 5d -->
+   </package>
+
+   <!--
+      This package is defined for Minion nodes. Don't alter/remove this. If minions are not getting used, it is safe to remove below pkg.
+   -->
+   <package name="Minion">
+      <filter>foreignSource != 'Minions' AND IPADDR != '0.0.0.0'</filter>
```
j-white: Shouldn't the filter be: `foreignSource == 'Minions' AND IPADDR != '0.0.0.0'`?

opennms-base-assembly/src/main/filtered/etc/poller-configuration.xml
```diff
@@ -284,10 +294,16 @@
...
+      <service name="JMX-Minion" interval="30000" user-defined="false" status="on">
+         <parameter key="port" value="18980"/>
+         <parameter key="retry" value="2"/>
+         <parameter key="timeout" value="3000"/>
+         <parameter key="rrd-repository" value="${install.share.dir}/rrd/response"/>
+         <parameter key="ds-name" value="jmx-minion"/>
+         <parameter key="use-foreign-id-as-system-id" value="true"/>
+      </service>
+      <!-- Query every 30secs always -->
+      <downtime begin="0" interval="30000"/><!-- 30s -->
```
j-white: Do we need to set a downtime model if we just want to stick with the same interval?

opennms-base-assembly/src/main/filtered/etc/poller-configuration.xml
cgorantla: my bad, Thanks for catching

opennms-base-assembly/src/main/filtered/etc/poller-configuration.xml
cgorantla: `downtime` is not an optional field in `poller-configuration`

cgorantla:
> Does the Jsr160Monitor used by the JMX-Minion service accept the "ttl" as a parameter?
It doesn't.  All monitors run from [here](https://github.com/OpenNMS/opennms/blob/develop/opennms-services/src/main/java/org/opennms/netmgt/poller/pollables/PollableServiceConfig.java#L125)  which handles timeout.

MinionRpcMonitor is a special case since it only runs on OpenNMS and the poller itself invokes RpcClient, so it needs to know when to  `timeout` .

opennms-base-assembly/src/main/filtered/etc/poller-configuration.xml
```diff
@@ -276,6 +268,24 @@
...
+      <downtime begin="43200000" end="432000000" interval="600000"/><!-- 10m, 12h, 5d -->
+      <downtime begin="432000000" interval="3600000"/><!-- 1h, 5d -->
+   </package>
+
+   <!--
+      This package is defined for Minion nodes. Don't alter/remove this. If minions are not getting used, it is safe to remove below pkg.
+   -->
+   <package name="Minion">
+      <filter>foreignSource == 'Minions' AND IPADDR != '0.0.0.0'</filter>
+      <rrd step="300">
```
j-white: Step should be adjusted to match the interval.

cgorantla: Done!

j-white: Shouldn't it be 30 instead of 300?

PR #2399, opened 2019-03-20
HZN-1478: Upgrade Apache CXF to 3.2.8

mvrueden:
Here we migrate to Apache CXF to 3.2.8.
Most changes were made to the osgi-jax-rs-connector project.
Technically we now require Java 9 to run. 
However I made it work temporarily with Java 8.
Before we release 24.0.0 we should make Java 9 the requirement

We should also consider upgrading the maven-bundle-plugin to the latest version. That version also determines the required java runtime environment for each bundle.

* jira: https://issues.opennms.org/browse/HZN-1478

### DIFF
--- a/container/features/src/main/resources/features.xml
+++ b/container/features/src/main/resources/features.xml
@@ -695,7 +695,7 @@
         <bundle>mvn:org.opennms.features.events/org.opennms.features.events.traps/${project.version}</bundle>
     </feature>
     <feature name="tsrm-troubleticketer" version="${project.version}" description="OpenNMS :: Features :: Ticketing :: Tivoli Service Request Manager (TSRM)">
-        <feature version="3.1.11">cxf-jaxws</feature>
+        <feature version="${cxfVersion}">cxf-jaxws</feature>
         <feature>javax.servlet</feature>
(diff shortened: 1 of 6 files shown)


j-white:
Can you elaborate on why we need Java 9 to run now? What changed?

mvrueden:
I updated the osgi-jax-rs-connector to use the same jax-rs version as the Apache CXF project.
However with that, the maven-bundle-plugin used in that project no longer works (Some exceptions is thrown). So I updated the plugin to a newer version. With that the maven-bundle-plugin now checks the required java runtime environment for each bundle and adds it to the bundle header.
With the new updated jaxrs version, the required Java Runtime Environment is now Java 9. My guess is, that `javax.ws.rs/javax.ws.rs-api/2.1.1` is targeted for java 9, as that is the only thing which changed.
Also see [here](https://github.com/bndtools/bnd/issues/413) for more details.

I manually overrode the header to [require 8 instead of 9](https://github.com/OpenNMS/osgi-jax-rs-connector/commit/eee95a636808aa27a02145f1957dae2b7d62ca5d) for now.

PR #2401, opened 2019-03-20
HZN-1499: Added "ON DELETE CASCADE" to metadata liquibase changelog

christianpape:
* JIRA: https://issues.opennms.org/browse/HZN-1499

### DIFF
--- a/core/schema/src/main/liquibase/24.0.0/changelog.xml
+++ b/core/schema/src/main/liquibase/24.0.0/changelog.xml
@@ -401,4 +401,20 @@
 
   </changeSet>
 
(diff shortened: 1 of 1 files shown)


j-white:
Looks good. Given that we're altering the existing migrations instead of adding a new one, can we provide instructions here on how to upgrade? We have some users running 24.0.0-SNAPSHOT, and `install -dis` will break for them when they upgrade with this patch.

christianpape:
We can delete and recreate the constraint in another changeset. What do you think.

j-white:
Either way works, we just need a solution that allows users to upgrade.

Manually running some SQL for the few that are running snapshots is OK.

christianpape:
Created an additional changeset. So, upgrades will work...

PR #2408, opened 2019-03-22
NMS-10624: Upgrade kafka components to 2.2.0

Naicisum:
This PR will upgrade kafka components to 2.2.0 from 1.0.1

The major changes were to KafkaOffsetProvider which used deprecated Scala functions, these were changed to the newer API. Impact should be minimal as this is only used by the test KafkaOffsetIT

Main code changes were minor to KafkaAlarmDataSync which moved the import path of Consumed

* JIRA: http://issues.opennms.org/browse/NMS-10624

### DIFF
--- a/core/ipc/sink/kafka/server/src/main/java/org/opennms/core/ipc/sink/kafka/server/offset/KafkaOffsetProvider.java
+++ b/core/ipc/sink/kafka/server/src/main/java/org/opennms/core/ipc/sink/kafka/server/offset/KafkaOffsetProvider.java
@@ -52,12 +52,12 @@ package org.opennms.core.ipc.sink.kafka.server.offset;
 
 import java.nio.ByteBuffer;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.HashMap;
-import java.util.List;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -68,6 +68,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
+import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.protocol.types.Field;
 import org.apache.kafka.common.protocol.types.Schema;
 import org.apache.kafka.common.protocol.types.Struct;
@@ -87,14 +88,6 @@ import com.codahale.metrics.JmxReporter;
 import com.codahale.metrics.MetricRegistry;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 
-import kafka.api.PartitionOffsetRequestInfo;
-import kafka.common.TopicAndPartition;
-import kafka.javaapi.OffsetResponse;
-import kafka.javaapi.PartitionMetadata;
-import kafka.javaapi.TopicMetadata;
-import kafka.javaapi.TopicMetadataRequest;
-import kafka.javaapi.consumer.SimpleConsumer;
-
 public class KafkaOffsetProvider {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(KafkaOffsetProvider.class);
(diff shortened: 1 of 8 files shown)


j-white:
Thanks @Naicisum. @cgorantla can you give this a whirl?

We also may need to update some of the dependency versions in https://github.com/OpenNMS/opennms/blob/develop/container/features/src/main/resources/features.xml#L415 and https://github.com/OpenNMS/opennms/blob/develop/container/features/src/main/resources/features.xml#L871 so that they line up with what the updated packages are using.

cgorantla:
I pushed this to `jira/NMS-10624` branch and most of kafka related tests fail. We need to fix the karaf dependency issues/ test related failures and then check if KafkaOffset feature is actually working. 
 https://bamboo.opennms.org/browse/OPENNMS-ONMS2820-1

Naicisum:
@cgorantla Can you push this again for Bamboo, looks like this fixes a bunch of the Kafka failures.

agalue:
My apologies for interrupt here, but just wanted to let you guys know, for informational purposes, that 2.1.1 is not the latest version anymore. Kafka 2.2.0 was released a few days ago.

Naicisum:
Probably be much easier to go from 2.1.1 to 2.2.0. Looks like my last dependency fix passed all the integration tests, just waiting on the smoke tests to finish.

Naicisum:
Updated PR to Kafka 2.2.0 same smoke-test failures

mvrueden:
The `develop` branch is on an older version and most of the test failures are related to our CXF Upgrade, which we reverted and ultimately fixed. However this branch does not contain the revert or the fix. I rebased `jira/NMS-10624` and locally OpenNMS started properly. 

However, investigating the distribution, this is fishy:

```
> find $OPENNMS_HOME -iname "*kafka*.jar" | grep -v opennms
./system/org/apache/servicemix/bundles/org.apache.servicemix.bundles.kafka-clients/2.1.1_1/org.apache.servicemix.bundles.kafka-clients-2.1.1_1.jar
./system/org/apache/kafka/kafka-streams/2.1.1/kafka-streams-2.1.1.jar
./lib/org.apache.servicemix.bundles.kafka-clients-2.1.1_1.jar
./lib/kafka-clients-2.1.1.jar
./lib/kafka_2.12-2.1.1.jar
```
There shouldn't be two variants (org.apache vs servicemix) of the same kafka library.

Besides this, you should use the `org.apache.servicemix.bundles`, as they are proper OSGi bundles.
The original `org.apache.karaf` bundles will ultimately fail when used in sentinel or minion.

mvrueden:
I pushed a rebased version against latest `develop` to `jira/NMS-10624`

Naicisum:
I emailed Jean-Baptiste with Apache who is one of the maintainers of the ServiceMix bundles about releasing Kafka bundle modules based on Scala 2.12. He has graciously (in my humble opinion) said "Sure, I will add this tomorrow." Once these reach release and are pushed out to maven repos, I will attempt the same using them.

As a side note, trying to duplicate the last push to bamboo locally, I am still getting failures on tests that passed, even after purging my source directory of all untracked files and removing all local maven repo caches. Must be missing something...

Naicisum:
Apache ServiceMix Bundle for Kafka 2.2.0 on Scala 2.12
- https://issues.apache.org/jira/browse/SM-4034

cgorantla:
I gave this a try.  `kafka` still doesn't load on karaf properly ( On Minion) because of missing bundle information in default apache jars.  I think we should fall back to latest available `apache.servicemix` bundles.

Naicisum:
I have a secondary branch "b" that I used to build a snapshot build of the servicemix bundle on 2.2.0 that is working with kafka. Just need to wait for apache to release official bundling at the end of this month.

Naicisum:
Give this one a whirl again, it has been rebased against develop.

cgorantla:
Bamboo is happy now.  Looks good to me, Just need to validate Kafka offset thing once whether it works with these changes.

core/ipc/sink/kafka/server/src/main/java/org/opennms/core/ipc/sink/kafka/server/offset/KafkaOffsetProvider.java
```diff
@@ -322,8 +297,9 @@ public long getLastOffset(SimpleConsumer consumer, String topic, int partition,
     }
 
     public void closeConnection() throws InterruptedException {
-        for (SimpleConsumer consumer : consumerMap.values()) {
-            LOGGER.info("Closing connection for: " + consumer.host());
+        for (KafkaConsumer consumer : consumerMap.values()) {
+            //TODO Find a replacement for SimpleConsumer.host() for logging
```
mvrueden: Please remove or fix TODO.

Naicisum: Done

mvrueden:
@cgorantla your review is outstanding. Are you okay with this PR?

cgorantla:
Validated RPC/Sink and Offset seems to be working fine.

PR #2409, opened 2019-03-22
NMS-10581: Handle graceful shutdown for OpenNMS

cgorantla:
Currently OpenNMS doesn't shutdown with `kill -s TERM` or `kill -s INT` 
Adding shutdown hook in Manager will allow OpenNMS to shutdown gracefully without the need for force kill.
Disable shutdown hook in karaf for OpenNMS.

* JIRA: http://issues.opennms.org/browse/NMS-10581

### DIFF
--- a/container/karaf/src/main/filtered-resources/etc/system.properties
+++ b/container/karaf/src/main/filtered-resources/etc/system.properties
@@ -161,6 +161,10 @@ karaf.secured.services = (&(osgi.command.scope=*)(osgi.command.function=*))
 # OPENNMS: Enable SNMP4J Log4j logging
 snmp4j.LogFactory=org.snmp4j.log.Log4jLogFactory
 
+# OPENNMS: Disable SIGTERM in karaf as OpenNMS should handle SIGTERM.
+karaf.handle.sigterm = false
+
+
 # OPENNMS: If you're using Bouncy Castle, disable the Bouncy Castle security provider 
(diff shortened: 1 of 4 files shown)


container/karaf/src/main/filtered-resources/etc/system.properties
```diff
@@ -161,6 +161,10 @@ karaf.secured.services = (&(osgi.command.scope=*)(osgi.command.function=*))
 # OPENNMS: Enable SNMP4J Log4j logging
 snmp4j.LogFactory=org.snmp4j.log.Log4jLogFactory
 
+# OPENNMS: Disable SIGTERM in karaf as OpenNMS should handle SIGTERM.
+karaf.handle.sigterm = false
```
Naicisum: How will this affect karaf in Sentinel and Minion?

cgorantla: afaik,  this properties file is specific to OpenNMS. It won't effect Sentinel/Minion as they are forked from [here](https://github.com/OpenNMS/opennms/tree/develop/features/container/sentinel/src/main/filtered-resources/etc)

agalue: That's true. It is important to guarantee that this version of `system.properties` is *only* for OpenNMS, and that file won't be shared with Sentinel or Minion, as the Docker images for these 2 relies on having Karaf properly handling SIGTERM.

cgorantla:
Thanks @indigo423  for all the testing.

PR #2410, opened 2019-03-25
HZN-1478: Upgrade CXF to 3.2.x or greater

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1478

Original [PR](https://github.com/OpenNMS/opennms/pull/2399) was reverted in the `release-24.0.0` branch.

It looks like this affected the stability of the system tests - we'll need to to investigate this further before merging again.

### DIFF
--- a/container/features/src/main/resources/features.xml
+++ b/container/features/src/main/resources/features.xml
@@ -699,7 +699,7 @@
         <bundle>mvn:org.opennms.features.events/org.opennms.features.events.traps/${project.version}</bundle>
     </feature>
     <feature name="tsrm-troubleticketer" version="${project.version}" description="OpenNMS :: Features :: Ticketing :: Tivoli Service Request Manager (TSRM)">
-        <feature version="3.1.11">cxf-jaxws</feature>
+        <feature version="${cxfVersion}">cxf-jaxws</feature>
         <feature>javax.servlet</feature>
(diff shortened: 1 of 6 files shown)


mvrueden:
Added `triage` label and changed base branch to `release-24.0.0`.

mvrueden:
The problem causing the long build times were related to the maven-deploy-plugin. 
In Bamboo we override the default locations of the repositories to deploy to instead in the `pom.xml` directly. However in the `osgi-jax-rs-connector` project we don't define a `maven-deploy-plugin` section, and therefore also no version. This results in maven using the latest version of the`maven-deploy-plugin`. However from version `3.0.0` and upwards some properties the bamboo deploy step uses are no longer supported, which results in some artifacts beeing pushed to the wrong repository. In this case snapshot vs release repository. 

This results in weird behaviour, when downloading the artifacts and populating the system directory, causing the tests to fail or manually downloading them from remote, which probably caused the build to take much longer. I manually deleted the 1.0.2* artifacts from our maven repository (`/mnt/...`), and released a new version `1.0.2.ONMS` to have a release and not a snapshot version to depend on.

Looking at the last two build, the smoke tests took roughly 300 minutes. 
This is a bit higher compared to `release-24`.
I ran the branch a third time and would expect a smoke test run time of roughly 300 minutes.

From my side, this looks good now.

PR #2411, opened 2019-03-25
Proposal: Reorganisation of Navbar

mvrueden:
This is an approach to make the OpenNMS header more useful by reorganising the menus and also show additional information, such es notifications.

This proposal contains the following changes:
 * Unacknowledged Notifications are now shown as little pills in the header
 * The user menu was re-structured
   * Quick actions, like Quick add Node or Configure OpenNMS are now top level actions (top most right icons)
   * Support, Help and About are moved to its own menu "Help"
   * Change Password will now directly go to the password change site, not the landing page

**Notifications**
The notifications are split in two visualisations:
 - unacknowledged notifications for the current user
 - unacknowledged notifications for anyone but the current user
The notification indicator for the user uses the color of the "most critical" severity of the 10 newest notifications. The indicator for everyone else is always "info".
[The feedback so far was very positive](https://opennms.discourse.group/t/proof-of-concept-for-notification-indication/320/5)

Here is a gif to see it in action: 

**Proposed Navbar**
Screenshot 2019-03-25 at 11 09 17

**Notification Dropdown**
Screenshot 2019-03-25 at 11 09 46

**User Dropdown**
Screenshot 2019-03-25 at 11 09 39

**Help Dropdown**
Screenshot 2019-03-25 at 11 09 33

It is targeted for 25 on purpose.
I had to change the way Vaadin is pulling in the header and I am expecting something to break here.
Besides this, I want to wait for feedback and make some more code changes, which will probably not make it to 24 in time.

Please let me know what you think.

### DIFF
--- a/container/features/src/main/resources/features.xml
+++ b/container/features/src/main/resources/features.xml
@@ -1112,6 +1112,7 @@
         <bundle>mvn:org.opennms.features.geocoder/org.opennms.features.geocoder.nominatim/${project.version}</bundle>
         <bundle>mvn:org.opennms.features.geolocation/org.opennms.features.geolocation.api/${project.version}</bundle>
         <bundle>mvn:org.opennms.features.vaadin-components/core/${project.version}</bundle>
+        <bundle>mvn:org.opennms.features.vaadin-components/header/${project.version}</bundle>
         <bundle>mvn:org.opennms.features.topology/org.opennms.features.topology.api/${project.version}</bundle>
         <bundle>mvn:org.opennms.features.topology.plugins/org.opennms.features.topology.plugins.browsers/${project.version}</bundle>
         <bundle>mvn:org.opennms.features.geolocation/org.opennms.features.geolocation.api/${project.version}</bundle>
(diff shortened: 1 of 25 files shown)


RangerRick:
Very cool, but I recommend making the severity stuff in the notice list be icons too, for color-blind people.

christianpape:
It seems that the vertical placement of the right outermost icons is better in FF and Chrome but also not perfect. In Safari it seems the gap is bigger.

mvrueden:
The right icons should be aligned with the overall menu.
The notification indicator are slightly off however. 
That is caused by the bootstrap 4 badge class and nothing I was able to fix so far.
In Firefox and Chrome it looks okay.
I may play with it some more, but does not look like a show-stopper to me.

About the severity icons:
We need to rethink our presentation and visualisation of severities and status in general.
Our current approach is not ideal. This includes the usage of severity icons.
There is a more general issue to address the "severity information design" ([NMS-10611](https://issues.opennms.org/browse/NMS-10611)).

indigo423:
Here a gif which shows feature a bit more detail.

navbar-demo

pbrane:
Nice gif! ;)  Just as a note, I really like the notices improvements.  One suggestion for improvement is to make a singe bell style icon decorated by:
- color of highest severity notice for all notices
- badge with the count of the logged in user's quantity of outstanding notices

Just a thought.

PR #2417, opened 2019-04-01
HZN-1492: Add Tags and RootCause to Feedback

smith-opennms:
This PR adds support for tags and root cause in the situation feedback.
Additionally, existing tags can be querired from the datasource to assist front end work.
Health Check implementation is added that simply pings the ElasticSearch rest api.

* JIRA: https://issues.opennms.org/browse/HZN-1492

### DIFF
--- a/features/flows/elastic/src/main/resources/OSGI-INF/blueprint/blueprint.xml
+++ b/features/flows/elastic/src/main/resources/OSGI-INF/blueprint/blueprint.xml
@@ -148,8 +148,9 @@
 
     <!-- Health Check -->
     <service interface="org.opennms.core.health.api.HealthCheck">
-        <bean class="org.opennms.netmgt.flows.elastic.health.ElasticHealthCheck" >
+        <bean class="org.opennms.plugins.elasticsearch.rest.ElasticHealthCheck" >
             <argument ref="jestClient"/>
+            <argument value="Flows"/>
         </bean>
     </service>
 
--- a/features/situation-feedback/api/src/main/java/org/opennms/features/situationfeedback/api/AlarmFeedback.java
+++ b/features/situation-feedback/api/src/main/java/org/opennms/features/situationfeedback/api/AlarmFeedback.java
@@ -28,8 +28,12 @@
 
 package org.opennms.features.situationfeedback.api;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 import java.util.Objects;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.google.common.base.Enums;
 
@@ -42,6 +46,7 @@ public final class AlarmFeedback {
     public enum FeedbackType {
         FALSE_POSITIVE, // Alarm does not belong in this Situation
         FALSE_NEGATIVE, // Alarm was missing from this Situation
+        CREATE_SITUATION, // Alarm should be correlated in a new Situation
         CORRECT, // Alarm is correctly correlated
         UNKNOWN;
 
@@ -61,7 +66,11 @@ public final class AlarmFeedback {
 
     private final FeedbackType feedbackType;
 
+    private final boolean rootCause;
+
     private final String reason;
+
+    private final List<String> tags = new ArrayList<>();
 
     private final String user;
 
@@ -73,6 +82,8 @@ public final class AlarmFeedback {
         this.alarmKey = builder.alarmKey;
         this.feedbackType = builder.feedbackType;
         this.reason = builder.reason;
+        this.rootCause = builder.rootCause != null && builder.rootCause;
+        this.tags.addAll(builder.tags);
         this.user = builder.user;
         this.timestamp = builder.timestamp;
     }
@@ -83,6 +94,9 @@ public final class AlarmFeedback {
         private String alarmKey;
         private AlarmFeedback.FeedbackType feedbackType;
         private String reason;
+        @JsonProperty(value="rootCause")
+        private Boolean rootCause;
+        private List<String> tags = new ArrayList<>();
         private String user;
         private Long timestamp;
 
@@ -111,8 +125,18 @@ public final class AlarmFeedback {
             return this;
         }
 
+        public Builder withRootCause(Boolean rootCause) {
+            this.rootCause = rootCause;
+            return this;
+        }
+
         public Builder withReason(String reason) {
             this.reason = reason;
+            return this;
+        }
+
+        public Builder withTags(List<String> tags) {
+            this.tags.addAll(tags);
             return this;
         }
 
@@ -156,8 +180,16 @@ public final class AlarmFeedback {
         return feedbackType;
     }
 
+    public boolean getRootCause() {
+        return rootCause;
+    }
+
     public String getReason() {
         return reason;
+    }
+
+    public List<String> getTags() {
+        return Collections.unmodifiableList(tags);
     }
 
     public String getUser() {
@@ -179,12 +211,14 @@ public final class AlarmFeedback {
                 Objects.equals(alarmKey, that.alarmKey) &&
                 feedbackType == that.feedbackType &&
                 Objects.equals(reason, that.reason) &&
+                Objects.equals(rootCause, that.rootCause) &&
+                Objects.equals(tags, that.tags) &&
                 Objects.equals(user, that.user);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(situationKey, situationFingerprint, alarmKey, feedbackType, reason, user, timestamp);
+        return Objects.hash(situationKey, situationFingerprint, alarmKey, feedbackType, reason, rootCause, tags, user, timestamp);
     }
 
     @Override
@@ -195,6 +229,8 @@ public final class AlarmFeedback {
                 ", alarmKey='" + alarmKey + '\'' +
                 ", feedbackType=" + feedbackType +
                 ", reason='" + reason + '\'' +
+                ", rootCause='" + rootCause + '\'' +
+                ", tags='" + tags + '\'' +
                 ", user='" + user + '\'' +
                 ", timestamp=" + timestamp +
                 '}';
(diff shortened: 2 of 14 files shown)


features/situation-feedback/elastic/src/main/java/org/opennms/features/situationfeedback/elastic/ElasticHealthCheck.java
```diff
@@ -0,0 +1,71 @@
...
+import org.opennms.core.health.api.Response;
+import org.opennms.core.health.api.Status;
+
+import com.google.common.base.Strings;
+
+import io.searchbox.client.JestClient;
+import io.searchbox.client.JestResult;
+import io.searchbox.core.Ping;
+
+public class ElasticHealthCheck implements HealthCheck {
```
mvrueden: There is already an elastic health check implementation. Can't we just re-use that instead of duplicating the code?

smith-opennms: I wouldn't want another feature to depend on flows-elastic feature - so it would have to be  refactored it out of flows and into the jest feature.
Health check will need 'description' to be parameterized so that if multiple features have different ES config you can easily determine which one is failing. Let me know how strongly you feel about this.

mvrueden: If this is duplicated code, the need for re-usage is given.
To me it looks worth tackling.

mvrueden: https://github.com/opennms/opennms/blob/823865bf8ec097d693a48479af73d5136472fd84/features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/health/ElasticHealthCheck.java#L54

mvrueden: And should be refactored out, maybe as you already suggested, into the jest feature.

features/situation-feedback/api/src/main/java/org/opennms/features/situationfeedback/api/AlarmFeedback.java
```diff
@@ -61,8 +65,12 @@ public static FeedbackType valueOfOrUnknown(String type) {
 
     private final FeedbackType feedbackType;
 
+    private final Boolean rootCause;
+
     private final String reason;
 
+    private final List<String> tags = new ArrayList<>();;
```
mattixtech: Extra semicolon.

Also,

I suspect this should be a `Set` as I don't think a tag occurring twice would be meaningful. Not sure if ordering matters here, if so used a `LinkedHashSet`.

```suggestion
    private final List<String> tags = new ArrayList<>();
```

features/situation-feedback/api/src/main/java/org/opennms/features/situationfeedback/api/AlarmFeedback.java
```diff
@@ -156,10 +179,18 @@ public FeedbackType getFeedbackType() {
...
+    public Boolean getRootCause() {
+        return rootCause;
+    }
+
     public String getReason() {
         return reason;
     }
 
+    public List<String> getTags() {
+        return tags;
```
mattixtech: Change to `unmodifiableSet` if we switch to a set.
```suggestion
        return Collections.unmodifiableList(tags);
```

features/situation-feedback/api/src/main/java/org/opennms/features/situationfeedback/api/AlarmFeedback.java
```diff
@@ -61,8 +65,12 @@ public static FeedbackType valueOfOrUnknown(String type) {
 
     private final FeedbackType feedbackType;
 
+    private final Boolean rootCause;
```
mattixtech: Only the builder should allow the root cause field to be nullable.

```suggestion
    private final boolean rootCause;
```

features/situation-feedback/api/src/main/java/org/opennms/features/situationfeedback/api/AlarmFeedback.java
```diff
@@ -156,10 +179,18 @@ public FeedbackType getFeedbackType() {
         return feedbackType;
     }
 
+    public Boolean getRootCause() {
```
mattixtech:
```suggestion
    public boolean getRootCause() {
```

features/situation-feedback/api/src/main/java/org/opennms/features/situationfeedback/api/FeedbackRepository.java
```diff
@@ -57,4 +57,9 @@
      */
     List<AlarmFeedback> getAllFeedback() throws FeedbackException;
 
+    /**
+     * @return a list of unique SituationFeedback Tags filtered to include only those beginning with @prefix
+     */
+    List<String> getTags(String prefix) throws FeedbackException;
```
mattixtech: Consider changing this to a `Set` as well.

features/situation-feedback/elastic/src/main/java/org/opennms/features/situationfeedback/elastic/FeedbackDocument.java
```diff
@@ -63,6 +66,12 @@
     @SerializedName("reason")
     private String reason;
 
+    @SerializedName("root_cause")
+    private Boolean isRootCause;
+
+    @SerializedName("tags")
+    private List<String> tags = new ArrayList<>();
```
mattixtech: Will want `Set` here as well.

features/situation-feedback/elastic/src/main/resources/feedback-template.json
```diff
@@ -63,6 +63,12 @@
                     "type": "text",
                     "norms": false
                 },
+                "root_cause": {
+                    "type": "boolean"
+                },
+                "tags": {
+                    "type": "keyword"
```
mattixtech: Just checking, does this actually get mapped correctly to the collection? Or will it only capture one element?

smith-opennms: it's correctly mapped

features/situation-feedback/api/src/main/java/org/opennms/features/situationfeedback/api/AlarmFeedback.java
smith-opennms: it needs to be Boolean in case the json that is deserialized does not have the element.

features/situation-feedback/api/src/main/java/org/opennms/features/situationfeedback/api/AlarmFeedback.java
smith-opennms: In the json that this serialized to and from, it will be a List

features/situation-feedback/elastic/src/main/java/org/opennms/features/situationfeedback/elastic/ElasticFeedbackRepository.java
```diff
@@ -125,6 +127,42 @@ public void persist(List<AlarmFeedback> feedback) throws FeedbackException {
...
 
+    @Override
+    public List<String> getTags(String prefix) throws FeedbackException {
+        String query = "{\n" + 
+                "  \"size\": 0,\n" + 
+                "  \"aggs\": {\n" + 
+                "    \"terms\": {\n" + 
+                "      \"terms\": {\n" + 
+                "        \"field\": \"tags\",\n" + 
+                "        \"include\": \"" + prefix + ".*\"\n" +
```
j-white: Can we ensure that the prefix is properly escaped? Looks like the query will break if it contains a `"` now.

features/situation-feedback/api/src/main/java/org/opennms/features/situationfeedback/api/AlarmFeedback.java
mattixtech: Thats true of the builder, but in the constructor for this object it is always explicitly set to true or false.

smith-opennms:
@mattixtech yes - why do you ask?

mattixtech:
> @mattixtech yes - why do you ask?

Just checking as part of the new review process 👍

features/flows/elastic/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -148,8 +148,9 @@
 
     <!-- Health Check -->
     <service interface="org.opennms.core.health.api.HealthCheck">
-        <bean class="org.opennms.netmgt.flows.elastic.health.ElasticHealthCheck" >
+        <bean class="org.opennms.plugins.elasticsearch.rest.ElasticHealthCheck" >
             <argument ref="jestClient"/>
+            <argument value="Flows"/>
```
mvrueden: Nice, that you changed this as well, wanted to nit on it already :)

features/situation-feedback/api/src/main/java/org/opennms/features/situationfeedback/api/AlarmFeedback.java
mattixtech: The serialization should only care about the setter in the builder, in this case `withTags`. So the argument that takes can remain a list, but the internal data structure we use can be a set. This will let us de-duplicate immediately upon receiving tags from the client and we won't end up storing dupes.

smith-opennms: Feedback also goes in and out the other end (ElasticSearch) as a JSON array. I'd rather keep the internal structure as a List to keep it in line with the external representations

features/situation-feedback/elastic/src/main/java/org/opennms/features/situationfeedback/elastic/ElasticFeedbackRepository.java
smith-opennms: now using ```gson.toJson()``` to escape 'prefix'

features/situation-feedback/api/src/main/java/org/opennms/features/situationfeedback/api/AlarmFeedback.java
mattixtech: Let's still update this to the unmodifiable list to maintain immutability in this class.

PR #2419, opened 2019-04-02
HZN-1500: Removed references to Java modules in OpenNMS start script

christianpape:
* JIRA: https://issues.opennms.org/browse/HZN-1500


RangerRick:
Pretty straightforward.  We'll need to remember to make a note that we support only Java 8 or 11 in the release notes.

christianpape:
It also seems to work with JDK9 and JDK10.

indigo423:
> Pretty straightforward. We'll need to remember to make a note that we support only Java 8 or 11 in the release notes.

Yepp good catch, we should just add the line with this PR and we won't forget :)

PR #2420, opened 2019-04-02
OCE-55: Add topology(link) support to the OpenNMS Direct datasource

mattixtech:
- Updated topology service to support binding edge consumers from the integration API as topology consumers
- Added mapping code to convert a topology message into an integration API edge type

Had to move some files such as TopologyUpdater in order to fix a bundle dependency issue. Also fixed a bug in the topology code where topology deletes were not getting sent to consumers when using the TopologyGenerator command `enlinkd:delete-topology`.

Tested manually using the topology generator shell command.

This depends on the integration API changes in https://github.com/OpenNMS/opennms-integration-api/pull/10

JIRA: http://issues.opennms.org/browse/OCE-55

### DIFF
--- a/features/api-layer/pom.xml
+++ b/features/api-layer/pom.xml
@@ -140,6 +140,13 @@
             <groupId>org.opennms.features.enlinkd</groupId>
             <artifactId>org.opennms.features.enlinkd.service.api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.opennms.features.topologies</groupId>
+            <artifactId>org.opennms.features.topologies.service.api</artifactId>
+            <version>${project.version}</version>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <repositories>
--- a/features/api-layer/pom.xml
+++ b/features/api-layer/pom.xml
@@ -132,6 +132,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.opennms.features.topologies</groupId>
+            <artifactId>org.opennms.features.topologies.service.api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
             <groupId>org.opennms.features.enlinkd</groupId>
             <artifactId>org.opennms.features.enlinkd.service.api</artifactId>
         </dependency>
--- a/features/api-layer/pom.xml
+++ b/features/api-layer/pom.xml
@@ -114,6 +114,11 @@
             <artifactId>mapstruct</artifactId>
             <version>${mapstructVersion}</version>
         </dependency>
+        <dependency>
+            <groupId>org.opennms.features.topologies</groupId>
+            <artifactId>org.opennms.features.topologies.service.api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
 
         <!-- Test -->
         <dependency>
@@ -130,11 +135,6 @@
             <groupId>org.mockito</groupId>
             <artifactId>mockito-all</artifactId>
             <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.opennms.features.topologies</groupId>
-            <artifactId>org.opennms.features.topologies.service.api</artifactId>
-            <version>${project.version}</version>
         </dependency>
         <dependency>
             <groupId>org.opennms.features.enlinkd</groupId>
--- a/features/api-layer/src/main/java/org/opennms/features/apilayer/topology/TopologyEdgeConsumerManager.java
+++ b/features/api-layer/src/main/java/org/opennms/features/apilayer/topology/TopologyEdgeConsumerManager.java
@@ -29,8 +29,8 @@
 package org.opennms.features.apilayer.topology;
 
 import java.util.Collections;
-import java.util.HashSet;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.opennms.features.apilayer.utils.InterfaceMapper;
 import org.opennms.features.apilayer.utils.ModelMappers;
@@ -62,21 +62,16 @@ public class TopologyEdgeConsumerManager extends InterfaceMapper<TopologyEdgeCon
 
             @Override
             public Set<OnmsTopologyProtocol> getProtocols() {
-                if(ext.getProtocols() == null) {
+                if (ext.getProtocols() == null) {
+                    LOG.debug("Protocols was null, returning empty set");
                     return Collections.emptySet();
                 }
 
-                Set<OnmsTopologyProtocol> protocols = new HashSet<>();
-
-                for (String protocol : ext.getProtocols().split(",")) {
-                    try {
-                        protocols.add(OnmsTopologyProtocol.create(protocol));
-                    } catch (Exception e) {
-                        LOG.warn("Cannot add protocol {} for consumer {}", protocol, ext, e);
-                    }
-                }
-
-                return protocols;
+                LOG.trace("Returning mapped protocols from {}", ext.getProtocols());
+                return ext.getProtocols()
+                        .stream()
+                        .map(ModelMappers::toOnmsTopologyProtocol)
+                        .collect(Collectors.toSet());
             }
 
             @Override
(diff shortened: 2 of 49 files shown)


mattixtech:
Addressed feedback from the integration API review. Will re-test once all feedback is implemented.

features/api-layer/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -24,6 +24,7 @@
   <reference id="eventConfDao" interface="org.opennms.netmgt.config.api.EventConfDao" availability="mandatory"/>
   <reference id="eventForwarder" interface="org.opennms.netmgt.events.api.EventForwarder" availability="mandatory"/>
   <reference id="eventSubscriptionService" interface="org.opennms.netmgt.events.api.EventSubscriptionService"/>
+  <reference id="onmsTopologyDao" interface="org.opennms.netmgt.topologies.service.api.OnmsTopologyDao" availability="mandatory"/>
```
j-white: Is this DAO available if the Enlinkd service is disabled in `service-configuration.xml`? If not, we'll need make it optional, otherwise the whole `api-layer` feature won't start when the service is disabled.

j-white: (We also have to fix that problem with some of the existing service references.)

features/api-layer/src/main/java/org/opennms/features/apilayer/utils/ModelMappers.java
```diff
@@ -132,4 +138,45 @@ public static AlarmFeedback toFeedback(org.opennms.features.situationfeedback.ap
...
+
+    public static TopologyEdge toEdge(String protocol, OnmsTopologyEdge edge) {
+        ImmutableTopologyEdge.Builder topologyEdge = ImmutableTopologyEdge.newBuilder()
+                .setId(edge.getId())
+                .setProtocol(protocol)
+                .setTooltipText(edge.getToolTipText())
+                .setSource(ImmutableTopologyPort.newBuilder()
+                        .setId(edge.getSource().getId())
+                        .setTooltipText(edge.getSource().getToolTipText())
+                        .setIfIndex(edge.getSource().getIfindex())
```
j-white: This throws an NPE if the source ifIndex is null - the ImmutableTopologyEdge doesn't require an ifIndex though.

(Similar problem with the target)

features/api-layer/pom.xml
```diff
@@ -131,6 +131,11 @@
             <artifactId>mockito-all</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
j-white: Nit: This is "grouped" with the test dependencies at the bottom of the pom, could be above that section.

features/api-layer/src/main/java/org/opennms/features/apilayer/topology/TopologyEdgeConsumerManager.java
```diff
@@ -0,0 +1,108 @@
...
+import org.opennms.netmgt.topologies.service.api.OnmsTopologyConsumer;
+import org.opennms.netmgt.topologies.service.api.OnmsTopologyEdge;
+import org.opennms.netmgt.topologies.service.api.OnmsTopologyMessage;
+import org.opennms.netmgt.topologies.service.api.OnmsTopologyProtocol;
+import org.opennms.netmgt.topologies.service.api.TopologyVisitor;
+import org.osgi.framework.BundleContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class TopologyEdgeConsumerManager extends InterfaceMapper<TopologyEdgeConsumer, OnmsTopologyConsumer> {
```
mvrueden: Is this a consistent naming scheme in the integration API? `Manager` sounds misleading to me in this context. `Mapper` seems more appropriate to me, just looking at his one class, without knowing anything else, please enlighten me :)

mattixtech: Currently all the interface mapper's use that naming convention which is why I stuck with it.

features/api-layer/src/main/java/org/opennms/features/apilayer/utils/ModelMappers.java
mattixtech: Oops, originally I had the ifIndex as a mandatory parameter for ports which is why I had the non-null check. I should fix this in the API project to allow that field to be set with a null value.

mattixtech: Updated here:
https://github.com/OpenNMS/opennms-integration-api/pull/11

features/api-layer/src/main/resources/OSGI-INF/blueprint/blueprint.xml
mattixtech: Will have to check on that.

mattixtech: With the EnhancedLinkd service disabled the feature still starts up and the EdgeDao service is made available.

mattixtech:
Pushed a bunch of changes including reworking the kafka producer to support the different types of edges/segments.

mattixtech:
I suspect this will fail tests due to the unmerged API changes. Will re-trigger tests once those get merged.

features/enlinkd/adapters/updaters/cdp/src/main/java/org/opennms/netmgt/enlinkd/CdpOnmsTopologyUpdater.java
```diff
@@ -30,6 +30,7 @@
 
 import java.util.Map;
 
+import org.opennms.netmgt.enlinkd.common.TopologyUpdater;
```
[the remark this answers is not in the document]
mattixtech: Yes its needed since I had to move TopologyUpdater to a new package due to that issue where multiple bundles were including the same package.

features/api-layer/src/main/java/org/opennms/features/apilayer/utils/ModelMappers.java
```diff
@@ -135,4 +147,72 @@ public static AlarmFeedback toFeedback(org.opennms.features.situationfeedback.ap
...
+
+    private static TopologyPort getPort(OnmsTopologyPort port) {
+        return ImmutableTopologyPort.newBuilder()
+                .setId(port.getId())
+                .setTooltipText(port.getToolTipText())
+                .setIfIndex(port.getIfindex())
+                .setIfName(port.getIfname())
+                .setIfAddress(port.getAddr())
+                .setNodeCriteria(ImmutableNodeCriteria.newBuilder()
+                        .setId(port.getVertex().getNodeid())
```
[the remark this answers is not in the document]
mattixtech: Done

features/api-layer/src/main/java/org/opennms/features/apilayer/utils/ModelMappers.java
```diff
@@ -135,4 +147,72 @@ public static AlarmFeedback toFeedback(org.opennms.features.situationfeedback.ap
...
+                        .build())
+                .build();
+    }
+
+    private static Node getNode(OnmsTopologyPort port) {
+        return ImmutableNode.newBuilder()
+                .setId(port.getVertex().getNodeid())
```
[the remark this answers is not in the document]
mattixtech: Done

features/api-layer/src/main/java/org/opennms/features/apilayer/utils/NodeCriteriaLoadingCacheImpl.java
```diff
@@ -0,0 +1,84 @@
...
+package org.opennms.features.apilayer.utils;
+
+import java.util.Objects;
+import java.util.Optional;
+
+import org.opennms.integration.api.v1.model.NodeCriteria;
+import org.opennms.integration.api.v1.model.immutables.ImmutableNodeCriteria;
+import org.opennms.netmgt.dao.api.NodeDao;
+import org.opennms.netmgt.model.OnmsNode;
+import org.springframework.transaction.support.TransactionOperations;
```
j-white: Let's use `org.opennms.netmgt.dao.api.SessionUtils` instead of TransactionOperations here - otherwise we run into problems when reloading the apilayer bundle on the fly in Karaf.

features/kafka/producer/src/main/resources/OSGI-INF/blueprint/blueprint-kafka-producer.xml
```diff
@@ -17,7 +17,6 @@
       <cm:property name="eventTopic" value="events"/>
       <cm:property name="alarmTopic" value="alarms"/>
       <cm:property name="nodeTopic" value="nodes"/>
-      <cm:property name="topologyProtocols" value="bridge,cdp,isis,lldp,ospf"/>
```
j-white: We also need to update the docs. See references to `topologyProtocols` in `opennms-doc/guide-admin/src/asciidoc/text/kafka-producer/kafka-producer.adoc`.

PR #2421, opened 2019-04-02
HZN-1510: Refactor GraphProvider to populate Graphs instead of itself

mvrueden:
The motivation behind this PR is to align the old Topology model to be more like the new API.
This will allow us to better integrate it later on.

A first step for this is to refactor the current GraphProvider (old Topology API) to actually populate a Graph instead of being the provider and graph altogether.
This is what this PR is all about.

The main changes are:
 - remove grouping as the support is long gone
 - remove `VertexProvider`/`EdgeProvider` as this was only use the `ncs` plugin, which is gone
 - Move all graph-related methods to a `Graph`. I named it `BackendGraph` for now, as the UI also has an interface `Graph`. I am open for better names (-:

With the above change there is no need for a `VertexHopGraphProvider` anymore, so that is gone as well. The replacement is a `CollapsibleGraph` which implements all the things the provider did before.

Keep in mind that code-wise a lot of this will change while integrating it more with the new Graph API.

JIRA: https://issues.opennms.org/browse/HZN-1510


features/topology-map/org.opennms.features.topology.api/src/main/java/org/opennms/features/topology/api/support/hops/VertexHopCriteria.java
```diff
@@ -0,0 +1,86 @@
...
+
+package org.opennms.features.topology.api.support.hops;
+
+import java.util.Set;
+
+import org.opennms.features.topology.api.topo.Criteria;
+import org.opennms.features.topology.api.topo.VertexRef;
+
+public abstract class VertexHopCriteria extends Criteria {
+    private String m_label = "";
```
patrick-schweizer: Initialization should not be needed since m_label is set in each constructor

patrick-schweizer:
looks good to me...

PR #2428, opened 2019-04-04
HZN-1511: Added meta-data to models of requisitioning UI.

fooker:
This does not add any UI elements, but makes the model driving the UI aware of the meta-data. It also includes the split-up between the "requisition" context and all other contexts.

By adding the meta-data to the model, this preserves existing meta-data while editing other things in the UI.

* JIRA: http://issues.opennms.org/browse/HZN-1511

### DIFF
--- a/core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/scripts/model/RequisitionInterface.js
+++ b/core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/scripts/model/RequisitionInterface.js
@@ -3,6 +3,9 @@
 * @copyright 2014 The OpenNMS Group, Inc.
 */
 
+const RequisitionService = require('./RequisitionService');
+const RequisitionMetaData = require('./RequisitionMetaData');
+
 /**
 * @ngdoc object
 * @name RequisitionInterface
@@ -65,25 +68,21 @@ const RequisitionInterface = function RequisitionInterface(intf) {
   self.services = [];
 
   angular.forEach(intf['monitored-service'], function(svc) {
-    self.services.push({ name: svc['service-name'] });
+    self.services.push(new RequisitionService(svc));
   });
 
   /**
-  * @description Adds a new monitored service to the interface
-  *
-  * @name RequisitionInterface:addNewService
-  * @ngdoc method
-  * @methodOf RequisitionInterface
-  * @returns {object} the new service Object
-  */
-  self.addNewService = function() {
-    self.services.push({ name: '' });
-    return self.services.length - 1;
-  };
+   * @description The meta-data entries
+   * @ngdoc property
+   * @name RequisitionInterface#metaData
+   * @propertyOf RequisitionInterface
+   * @returns {object} The meta-data entries
+   */
+  self.metaData = new RequisitionMetaData(intf['meta-data']);
 
   self.className = 'RequisitionInterface';
 
   return self;
-}
+};
 
 module.exports = RequisitionInterface;
--- a/core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/scripts/model/RequisitionNode.js
+++ b/core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/scripts/model/RequisitionNode.js
@@ -4,6 +4,7 @@
 */
 
 const RequisitionInterface = require('./RequisitionInterface');
+const RequisitionMetaData = require('./RequisitionMetaData');
 
 // Internal function for initialization purposes
 const isEmpty = function(str) {
@@ -156,8 +157,17 @@ const RequisitionNode = function RequisitionNode(foreignSource, node, isDeployed
    */
   self.assets = [];
 
+  /**
+   * @description The meta-data entries
+   * @ngdoc property
+   * @name RequisitionNode#metaData
+   * @propertyOf RequisitionNode
+   * @returns {object} The meta-data entries
+   */
+  self.metaData = new RequisitionMetaData(node['meta-data']);
+
   angular.forEach(node['interface'], function(intf) {
-    self.interfaces.push(new RequisitionInterface(intf));
+      self.interfaces.push(new RequisitionInterface(intf));
   });
 
   angular.forEach(node['asset'], function(asset) {
@@ -218,6 +228,23 @@ const RequisitionNode = function RequisitionNode(foreignSource, node, isDeployed
       value: ''
     });
     return self.assets.length -1;
+  };
+
+  /**
+   * @description Adds a new meta-data entry to the node
+   *
+   * @name RequisitionNode:addNewMetaData
+   * @ngdoc method
+   * @methodOf RequisitionNode
+   * @returns {object} the new meta-data Object
+   */
+  self.addNewMetaData = function() {
+    let entry = {
+        key: '',
+        value: ''
+    };
+    self.requisitionMetaData.push(entry);
+    return entry;
   };
 
   /**
@@ -294,6 +321,7 @@ const RequisitionNode = function RequisitionNode(foreignSource, node, isDeployed
       'parent-foreign-id': self.parentForeignId,
       'parent-node-label': self.parentNodeLabel,
       'asset': [],
+      'meta-data': self.metaData.getOnmsMetaData(),
       'category': []
     };
 
@@ -303,12 +331,17 @@ const RequisitionNode = function RequisitionNode(foreignSource, node, isDeployed
         'descr': intf.description,
         'snmp-primary': intf.snmpPrimary,
         'status': (intf.status || intf.status === 'managed') ? '1' : '3',
+        'meta-data': intf.metaData.getOnmsMetaData(),
         'monitored-service': []
       };
+
       angular.forEach(intf.services, function(service) {
-        interfaceObject['monitored-service'].push({
-          'service-name': service.name
-        });
+        var serviceObject = {
+          'service-name': service.name,
+          'meta-data': service.metaData.getOnmsMetaData()
+        };
+
+        interfaceObject['monitored-service'].push(serviceObject);
       });
 
       nodeObject['interface'].push(interfaceObject);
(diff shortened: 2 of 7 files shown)


core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/scripts/model/RequisitionNode.js
```diff
@@ -156,8 +157,43 @@ const RequisitionNode = function RequisitionNode(foreignSource, node, isDeployed
...
+   * @returns {object} The requisition metaData entries
+   */
+  self.requisitionMetaData = [];
+
+  /**
+   * @description The array of other metaData entries
+   * @ngdoc property
+   * @name RequisitionNode#otherMetaData
+   * @propertyOf RequisitionNode
+   * @returns {object} The other metaData entries
```
christianpape: use same spelling for meta-data

core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/scripts/model/RequisitionNode.js
```diff
@@ -156,8 +157,43 @@ const RequisitionNode = function RequisitionNode(foreignSource, node, isDeployed
    */
   self.assets = [];
 
+  /**
+   * @description The array of requisition metaData entries
+   * @ngdoc property
+   * @name RequisitionNode#requisitionMetaData
+   * @propertyOf RequisitionNode
+   * @returns {object} The requisition metaData entries
```
christianpape: use same spelling for meta-data

core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/scripts/model/RequisitionNode.js
```diff
@@ -156,8 +157,43 @@ const RequisitionNode = function RequisitionNode(foreignSource, node, isDeployed
    */
   self.assets = [];
 
+  /**
+   * @description The array of requisition metaData entries
```
christianpape: use same spelling for meta-data

core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/scripts/model/RequisitionNode.js
```diff
@@ -220,6 +256,23 @@ const RequisitionNode = function RequisitionNode(foreignSource, node, isDeployed
     return self.assets.length -1;
   };
 
+  /**
+   * @description Adds a new meta-data entry to the node
+   *
+   * @name RequisitionNode:addNewMetaData
+   * @ngdoc method
+   * @methodOf RequisitionNode
+   * @returns {object} the new meta-data Object
```
christianpape: use same spelling for meta-data

core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/scripts/model/RequisitionNode.js
```diff
@@ -220,6 +256,23 @@ const RequisitionNode = function RequisitionNode(foreignSource, node, isDeployed
     return self.assets.length -1;
   };
 
+  /**
+   * @description Adds a new meta-data entry to the node
```
christianpape: use same spelling for meta-data

core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/scripts/model/RequisitionService.js
```diff
@@ -0,0 +1,69 @@
...
+   * @returns {string} The name of the service
+   */
+  self.name = svc['service-name'];
+
+  /**
+   * @description The array of requisition metaData entries
+   * @ngdoc property
+   * @name RequisitionNode#requisitionMetaData
+   * @propertyOf RequisitionNode
+   * @returns {object} The requisition metaData entries
```
christianpape: same here

core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/scripts/model/RequisitionService.js
```diff
@@ -0,0 +1,69 @@
...
+   */
+  self.requisitionMetaData = [];
+
+  /**
+   * @description The array of other metaData entries
+   * @ngdoc property
+   * @name RequisitionNode#otherMetaData
+   * @propertyOf RequisitionNode
+   * @returns {object} The other metaData entries
```
christianpape: same here

core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/scripts/model/RequisitionService.js
```diff
@@ -0,0 +1,69 @@
...
+* @copyright 2014 The OpenNMS Group, Inc.
+*/
+
+const _ = require('lodash');
+
+/**
+* @ngdoc object
+* @name RequisitionInterface
+* @module onms-requisitions
+* @param {Object} intf an OpenNMS interface JSON object
```
mvrueden: Is it possible to use something more suitable in terms of naming?
Maybe `apiObject` or `opennmsApiObject`. 
`intf` sounds like a mix between integer and float and hurts in my head, while reading, as it just doesn't parse to something sane.

agalue: I invite you to execute the following from the root of our repo:
```
git grep " intf"
```
That way you'll see the amount of places on which either an `IP Interface` or an `SNMP Interface` is referred using the name `intf` :)

fooker: :joy: I would totally opt for using `intf`, too. But the funny thing is: nobody noticed, that the whole comment block is wrong and this is all about services, not interfaces. So, this would be `svc` now...

mvrueden: That wasn't my point.
I interpreted the comment as `a generic api object from the OpenNMS Rest interface`, thus `apiObject` vs `interface`.

core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/scripts/model/RequisitionInterface.js
```diff
@@ -65,25 +68,47 @@ const RequisitionInterface = function RequisitionInterface(intf) {
...
+  self.otherMetaData = {};
+
+  angular.forEach(intf['meta-data'], function(entry) {
+    if (entry.context === 'requisition') {
+      self.requisitionMetaData.push({
+        'key': entry.key,
+        'value': entry.value,
+      });
+    } else {
+      if (!_.has(self.otherMetaData, entry.context)) {
```
RangerRick: You know you could avoid an entire lodash import by just making this:

```
if (!self.otherMetaData.hasOwnProperty(entry.context)) {
```

The only thing useful about using `._has()` instead is if `self.otherMetaData` is undefined, but you're unconditionally assigning to it anyways.

fooker: Yea, I know this. I was just mimicking the way it was done in the other models :sweat_smile:. Do you think it's woth changing?

RangerRick: I would say it makes sense to move away from lodash if we don't need an extra dependency.  Especially if you're having to _add_ it to existing files.

RangerRick: (To be clear, not saying you have to refactor all the things you're not touching in this PR or anything, just don't add new unnecessary dependencies.)

fooker: Changed.

core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/scripts/model/RequisitionNode.js
```diff
@@ -4,6 +4,7 @@
 */
 
 const RequisitionInterface = require('./RequisitionInterface');
+const _ = require('lodash');
```
mvrueden: @RangerRick added a comment recently to ensure that we really need this external dependency. If not, we should not introduce it.

core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/scripts/model/RequisitionNode.js
```diff
@@ -156,8 +157,43 @@ const RequisitionNode = function RequisitionNode(foreignSource, node, isDeployed
...
+      });
+    } else {
+      if (!_.has(self.otherMetaData, entry.context)) {
+        self.otherMetaData[entry.context] = []
+      }
+      self.otherMetaData[entry.context].push({
+        'key': entry.key,
+        'value': entry.value,
+      });
+    }
```
mvrueden: This whole block is duplicated a few times, can we somhow reduce that?

fooker: Not that I'm aware off. All things I've tried to do so are much more complex than these few lines of code.

core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/scripts/model/RequisitionNode.js
fooker: Fixed

core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/scripts/model/RequisitionNode.js
mvrueden: `angular.forEach(...)` could be rewritten, by something like 
```
 if (isArray(x)) {
   x.forEach(function(entry) {... });
};
```
or just `x.forEach(...)`. 

Besides this, putting it into its own file, maybe `Utils` shouldn't be that hard or do I miss something?

fooker: I've moved the duplicated code to a separate model class `RequisitionMetaData.js` whis also de-duplicates the code generating the DTO Model from the meta-data.

PR #2431, opened 2019-04-04
OIA-11: More Controller API Improvements

j-white:
JIRA: https://issues.opennms.org/browse/OIA-11

This patch includes a number of updates to implement the latest interfaces in the integration API that help improve the facilities available to plugins for integrating with controllers.

We added:
* Support for meta-data
* Support for user defined links
* Push metric support (OIA-11)


features/api-layer/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -259,4 +270,13 @@
     </bean>
   </service>
 
+  <!-- Topology -->
```
cgorantla: Since we are tied to `userDefinedLinkDao` being available for `UserDefinedLinkDaoImpl` to be started and in turn this bundle, If I  install `opennms-integration-api-features`  without enabling Enlinkd then this feature will fail which would be a problem for all other integration api that depend on this feature.

I think we have this problem with all other modules like  `Pollerd` , `Collectd` as well if they are not enabled.

j-white: The `userDefinedLinkDao` and `userDefinedLinkTopologyService` services are actually referenced as `optional` to avoid this problem.

PR #2433, opened 2019-04-05
Fix for jira/NMS-10642: DNSResolutionMonitor incorrectly sets port number

roskens:
jira/NMS-10642: DNSResolutionMonitor incorrectly sets port number 

* JIRA: http://issues.opennms.org/browse/NMS-10642

### DIFF
--- a/features/poller/monitors/core/src/main/java/org/opennms/netmgt/poller/monitors/DNSResolutionMonitor.java
+++ b/features/poller/monitors/core/src/main/java/org/opennms/netmgt/poller/monitors/DNSResolutionMonitor.java
@@ -143,10 +143,10 @@ public class DNSResolutionMonitor extends ParameterSubstitutingMonitor {
                     // hostname with port number
                     final Integer pos = nameserver.lastIndexOf(":");
                     final String hostname = nameserver.substring(0, pos);
-                    final String port = nameserver.substring(pos + 1);
+                    final Integer port = Integer.valueOf(nameserver.substring(pos + 1));
                     LOG.debug("nameserver: hostname={}, port={}", hostname, port);
                     resolver = new SimpleResolver(hostname);
-                    resolver.setPort(SystemProperties.getInteger(port));
+                    resolver.setPort(port);
 
                 } else {
                     // hostname or ip address


features/poller/monitors/core/src/main/java/org/opennms/netmgt/poller/monitors/DNSResolutionMonitor.java
```diff
@@ -143,10 +143,10 @@ public PollStatus poll(MonitoredService svc, Map<String, Object> parameters) {
                     // hostname with port number
                     final Integer pos = nameserver.lastIndexOf(":");
                     final String hostname = nameserver.substring(0, pos);
-                    final String port = nameserver.substring(pos + 1);
+                    final Integer port = Integer.valueOf(nameserver.substring(pos + 1));
```
RangerRick: `SystemProperties.getInteger()` should be used instead of `Integer.valueOf()` so we get better logging.

roskens: That's the problem with the prior changes that this reverts. port isn't the name of a system property, so which of the static SystemProperties.getInteger() methods would be used here?

`public static Integer getInteger(String propertyName)`
`public static Integer getInteger(String propertyName, int defaultValue)`
`public static Integer getInteger(String propertyName, Integer defaultValue)`

RangerRick: Oh! I see now, yeah.  That makes sense then.

RangerRick:
Also, this is outside of the scope of your bug report, but I just noticed https://github.com/OpenNMS/opennms/blob/682cde99d9d466bcf895e266647f8a50a0262672/features/poller/monitors/core/src/main/java/org/opennms/netmgt/poller/monitors/DNSResolutionMonitor.java#L133 which apparently thinks IPv6 addresses are only numeric and don't have `a-f` in there as well??

RangerRick:
Mind rebasing this to `foundation-2018`?  It looks good otherwise, I can merge it then.

roskens:
> Mind rebasing this to `foundation-2018`? It looks good otherwise, I can merge it then.

Foundation-2018, and master branches are all based on a version prior to the SystemProperties class work, so they don't need to be fixed. 
They all share the same hex-digit bug you noted though.

RangerRick:
makes sense, I'll go ahead and merge then

PR #2438, opened 2019-04-08
NMS-10646: Fix application contexts related to timeseries strategy : evaluate, tcp

cgorantla:
As part of refactoring, spring application context files related to timeseries were renamed from `applicationContext-#strategy#`  to `applicationContext-timeseries-#strategy#`.  
It was missed for `evaluate` strategy.  This  should fix the application loading for evaluate strategy.

* JIRA: http://issues.opennms.org/browse/NMS-10646


cgorantla:
Looks like we also missed refactoring it for tcp strategy as well. Fixed that.

PR #2440, opened 2019-04-09
OIA-12:  Make osgi service references optional 

cgorantla:
Fix bug where disabling of  `Pollerd` and `Collectd` will make features/api-layer bundle to fail.

* JIRA: http://issues.opennms.org/browse/OIA-12


cgorantla:
Verified this by disabling Pollerd/Collectd and checking if  integration api related features are started properly.

PR #2442, opened 2019-04-10
NMS-10640: Introduced reduction key filter for IFTTT trigger packages

christianpape:
Currently the IFTTT-integration only filters based on categories which will only allow alarms with associated nodes. A new reductionKeyFilter allows to filter also for reduction keys, so it is possible now to trigger IFTTT events for BS alarms.
* JIRA: https://issues.opennms.org/browse/NMS-10640

### DIFF
--- a/features/ifttt/src/main/java/org/opennms/features/ifttt/IfTttDaemon.java
+++ b/features/ifttt/src/main/java/org/opennms/features/ifttt/IfTttDaemon.java
@@ -166,18 +166,22 @@ public class IfTttDaemon {
             private List<OnmsAlarm> filterAlarms(List<OnmsAlarm> alarms, TriggerPackage triggerPackage) {
                 if (triggerPackage.getOnlyUnacknowledged()) {
                     return alarms.stream()
-                            .filter(alarm -> alarm.getNodeId() != null)
                             .filter(alarm -> !alarm.isAcknowledged())
+                            .filter(alarm -> Strings.isNullOrEmpty(triggerPackage.getCategoryFilter()) || alarm.getNodeId() != null)
                             .filter(alarm -> Strings.isNullOrEmpty(triggerPackage.getCategoryFilter()) ||
                                     alarm.getNode().getCategories().stream()
                                             .anyMatch(category -> category.getName().matches(triggerPackage.getCategoryFilter())))
+                            .filter(alarm -> Strings.isNullOrEmpty(triggerPackage.getUeiFilter()) ||
+                                    alarm.getUei().matches(triggerPackage.getUeiFilter()))
                             .collect(Collectors.toList());
                 } else {
                     return alarms.stream()
-                            .filter(alarm -> alarm.getNodeId() != null)
+                            .filter(alarm -> Strings.isNullOrEmpty(triggerPackage.getCategoryFilter()) || alarm.getNodeId() != null)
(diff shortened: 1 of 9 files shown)


features/ifttt/src/main/java/org/opennms/features/ifttt/IfTttDaemon.java
```diff
@@ -166,18 +166,22 @@ private void startPoller(final long pollInterval) {
             private List<OnmsAlarm> filterAlarms(List<OnmsAlarm> alarms, TriggerPackage triggerPackage) {
                 if (triggerPackage.getOnlyUnacknowledged()) {
                     return alarms.stream()
-                            .filter(alarm -> alarm.getNodeId() != null)
                             .filter(alarm -> !alarm.isAcknowledged())
+                            .filter(alarm -> Strings.isNullOrEmpty(triggerPackage.getCategoryFilter()) || alarm.getNodeId() != null)
```
fooker: You can simplify this something like this:
```
let s = alarms.stream();
if (triggerPackage.getOnlyUnacknowledged()) {
  s = s.filter(alarm -> !alarm.isAcknowledged());
}
if (Strings.isNullOrEmpty(triggerPackage.getCategoryFilter())) {
  s = s.filter(alarm.getUei().matches(triggerPackage.getUeiFilter()));
}
return s.[...].collect([...])
```

features/ifttt/src/main/java/org/opennms/features/ifttt/config/TriggerPackage.java
```diff
@@ -51,13 +55,22 @@
...
+        return categoryFilter != null ? categoryFilter : "";
     }
 
     public void setCategoryFilter(final String categoryFilter) {
         this.categoryFilter = categoryFilter;
     }
 
+    @XmlAttribute
+    public String getUeiFilter() {
+        return ueiFilter != null ? ueiFilter : ".*";
```
fooker: Shouldn't this be an empty string, too?

features/ifttt/src/test/java/org/opennms/features/ifttt/IfTttDaemonTest.java
```diff
@@ -187,29 +191,47 @@ public void setup() {
 
         addAlarm(10, null, OnmsSeverity.CRITICAL);  // -
 
-        addAlarm(11, 2, OnmsSeverity.CRITICAL, true);  // -
```
fooker: Add `addAlarm(11, 2, "", CRITICAL);`, too?

features/ifttt/src/main/java/org/opennms/features/ifttt/config/TriggerPackage.java
```diff
@@ -64,7 +66,7 @@ public void setCategoryFilter(final String categoryFilter) {
 
     @XmlAttribute
     public String getUeiFilter() {
-        return ueiFilter != null ? ueiFilter : ".*";
+        return Strings.isNullOrEmpty(ueiFilter) ? ".*" : ueiFilter;
```
fooker: I was talking about using the empty string instead of the match-all regex... Not sure if this works out.

christianpape:
Changed filter to match the alarm's reduction key instead of its UEI. This allows to filter for example for a single node outage or BS alarm.

mfuhrmann:
Small enhancement with great impact (if you are using alarm lamps). Thanks!

PR #2444, opened 2019-04-10
NMS-10653: add isAcknowledged search property

RangerRick:
This PR is a small change to add `isAcknowledged` as a search property in the alarm v2 ReST API.

* JIRA: http://issues.opennms.org/browse/NMS-10653

I tested this by referencing it in the Helm dynamic filtering branch: 
Is Acknowledged

It works when filtering alarms in the alarm table.


opennms-model/src/main/java/org/opennms/netmgt/model/OnmsAlarm.java
```diff
@@ -727,7 +727,11 @@ public void setAlarmAckUser(String alarmackuser) {
     @Transient
     @XmlTransient
     public boolean isAcknowledged() {
-        return getAlarmAckUser() != null;
+        return getAlarmAckTime() != null;
+    }
+
+    /* technically not a modifiable property, just here to make it show as a bean property */
```
j-white: Why do we need this to show up as a bean property? Seems like other search properties such as `isInSituation` work without it.

RangerRick: It  may not, really... originally I thought this was the change I needed but then it didn't actually help, so then I found the other bit.  I'll remove this and see if it still works.

PR #2449, opened 2019-04-12
NMS-10656: remove the unmaintained centric ticketer plugin

RangerRick:
This PR removes the centric ticketer plugin.  It's sat in `opennms-tools` forever because of weird build dependencies.  We can resurrect it if someone ends up actually needing it and wants us to update it to whatever it is you're supposed to use nowadays to talk to Centric since I assume billion year old jars is not it.

Tagging @agalue and @jeffgdotorg just to make sure support is cool with it going away.  I can't imagine we have a customer still actively using it though.

* JIRA: http://issues.opennms.org/browse/NMS-10656


RangerRick:
@jeffgdotorg said this is unused by any support folks (including the people it was written for) so I'm gonna merge this

PR #2450, opened 2019-04-12
NMS-10639: change dependencies to OpenJDK 11

RangerRick:
* change all java dependencies to OpenJDK 11
* move the java dependency to the minion meta-package
* remove the java dependency from sentinel since it doesn't
  have a meta-package

* JIRA: http://issues.opennms.org/browse/NMS-10639

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -167,7 +167,7 @@ Description: Enterprise-grade Open-source Network Management Platform (Common Fi
 
 Package: libopennms-java
 Architecture: all
-Depends: oracle-java8-installer | openjdk-8-jre | java8-runtime | java8-runtime-headless, jicmp (>= 2.0.0), jicmp6 (>= 2.0.0)
+Depends: openjdk-11-jdk | openjdk-11-jdk-headless, jicmp (>= 2.0.0), jicmp6 (>= 2.0.0)
 Conflicts: opennms-plugin-protocol-xml (<<${binary:Version})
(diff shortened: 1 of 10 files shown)


j-white:
Changes look good.

There are some references to Oracle JDK 8 in the installation guide, should we update those here too?

RangerRick:
Yeah, probably.  :)

Naicisum:
I noticed the documents don't seem to reference the package "java-11-openjdk-devel". However the opennms spec file does. 

For windows, since the Oracle 11 JDK license has changed significantly, how about using/referencing these packages. They are created from OpenJDK by a member from the redhat team for the windows platform.
 - https://github.com/ojdkbuild/ojdkbuild

My experience on the rhel side is that by installing "java-11-openjdk-devel" you get "java-11-openjdk" and "java-11-openjdk-headless" packages. Was it intentional not to use the JDK for minion/sentinel?

RangerRick:
As I understand it, the only reason we need the JDK specifically on the OpenNMS side is because of some runtime compilation stuff the web UI uses, and that's not an issue in minion or Sentinel.

As for mentioning Windows, I did the bare minimum of mentioning the right versions, but it's not like Windows is terribly supported in OpenNMS, even if it technically works.  Until we actually put some engineering into making OpenNMS on Windows more supportable, I'd still put it in the "you're on your own" category.  That said, I can add a link.

PR #2451, opened 2019-04-15
HZN-1529: Wrap Sink Message to ProtoBuf

cgorantla:
In order to support  `Tracing Info (for Distributed Tracing)` or handling `large buffer sizes` in Sink Message, we need to have more fields in `Sink Message`. Since this requirement is specific to `Kafka`, we can wrap `Sink Message to Protobuf` and add more fields in proto. This is first step in defining `sink-messsage.proto` fields and handle protobuf as the message in Kafka Client/Server layer.

Also fixes logging and handling kafka timeout errors as only case for sink messages to buffer. In all other cases we drop messages as they are non-recoverable errors.

* JIRA: http://issues.opennms.org/browse/HZN-1529

* JIRA: https://issues.opennms.org/browse/HZN-1509


core/ipc/sink/kafka/client/src/main/java/org/opennms/core/ipc/sink/kafka/client/KafkaRemoteMessageDispatcherFactory.java
```diff
@@ -89,12 +96,27 @@
                     Thread.currentThread().interrupt();
                     break;
                 } catch (ExecutionException e) {
-                    LOG.warn("Timeout occured while sending message to topic {}, it will be attempted again.", topic);
+                    // Timeout typically happens when Kafka is Offline or it didn't initialize yet.
+                    // For this case keep sending the message until it delivers, will cause sink messages to buffer.
+                    if (e.getCause() != null && e.getCause() instanceof TimeoutException) {
+                        LOG.warn("Timeout occured while sending message to topic {}, it will be attempted again.", topic, e);
```
j-white: Let's remove the stack trace from the logs here, since we know it's a TimeoutException, and if it happens there's bound to be a lot of these.
```suggestion
                        LOG.warn("Timeout occurred while sending message to topic {}, it will be attempted again.", topic);
```

core/ipc/sink/kafka/server/src/main/java/org/opennms/core/ipc/sink/kafka/server/KafkaMessageConsumerManager.java
```diff
@@ -101,9 +103,12 @@ public void run() {
                     ConsumerRecords<String, byte[]> records = consumer.poll(100);
                     for (ConsumerRecord<String, byte[]> record : records) {
                         try {
-                            dispatch(module, module.unmarshal(record.value()));
+                            byte[] messageInBytes = getSinkMessageFromProto(record.value());
+                            dispatch(module, module.unmarshal(messageInBytes));
                         } catch (RuntimeException e) {
                             LOG.warn("Unexpected exception while dispatching message", e);
+                        } catch (InvalidProtocolBufferException e) {
+                            LOG.warn("Error parsing protobuf", e);
```
j-white:
```suggestion
                            LOG.error("Error parsing procotol buffer in message. The message will be dropped. Ensure that all components are running the same version of the software.");
```

PR #2455, opened 2019-04-15
OIA-17: Move implementations for API types from Opennms to integration API

mattixtech:
Changes to OpenNMS to use new impls in integration API 0.2.0

Jira: https://issues.opennms.org/browse/OIA-17


j-white:
This should be re-targeted for the `release-24.1.0` branch. Note that this will also pull the OIA snapshots back in, which will break Sentinel right now.

mattixtech:
@j-white Given the Karaf upgrade was done in develop, should this be re-targeted to develop since this re-introduces the snapshot dependency for OIA.

j-white:
Yeah, let's re-target this for `develop`.

features/api-layer/src/main/java/org/opennms/features/apilayer/utils/ModelMappers.java
```diff
@@ -58,11 +61,28 @@
...
+                .setDescription(alarm.getDescription())
+                .setLastEventTime(alarm.getLastEventTime())
+                .setFirstEventTime(alarm.getFirstEventTime())
+                .setLastEvent(toEvent(alarm.getLastEvent()))
+                .build();
     }
 
     public static InMemoryEvent toEvent(Event event) {
-        return new InMemoryEventBean(event);
+        return event == null ? null : Mappers.getMapper(InMemoryEventMapper.class).map(event);
```
j-white: Mappers can be stored in `static final` variables instead of rebuilding these each time.

mattixtech:
> Can we add unit tests for the mappers?

Done.

j-white:
👍

PR #2457, opened 2019-04-16
NMS-9893: Alarm Clear Trigger query language performance improvement

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-9893

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/alarmd/drools-rules.d/alarmd.drl
+++ b/opennms-base-assembly/src/main/filtered/etc/alarmd/drools-rules.d/alarmd.drl
@@ -50,10 +50,13 @@ end
     <!-- Find all alarms that can be cleared -->
     <trigger name="selectAlarmsToClear" operator="&gt;=" row-count="1" >
       <statement>
-        SELECT *, now() AS _ts FROM alarms as problems
-        WHERE alarmType=1
-          AND severity &gt; 2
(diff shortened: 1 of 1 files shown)


opennms-base-assembly/src/main/filtered/etc/alarmd/drools-rules.d/alarmd.drl
```diff
@@ -50,10 +50,13 @@ end
     <!-- Find all alarms that can be cleared -->
     <trigger name="selectAlarmsToClear" operator="&gt;=" row-count="1" >
       <statement>
-        SELECT *, now() AS _ts FROM alarms as problems
-        WHERE alarmType=1
-          AND severity &gt; 2
-          AND reductionKey IN (SELECT clearKey FROM alarms WHERE alarmType=2 AND lastEventTime &gt; problems.lastEventTime)
+        SELECT A1.*, now() AS _ts FROM alarms A1
```
j-white: This is in a comment block - the previous vacuumd statements were only copied here as a reference.

christianpape: I was wondering too. Then neither the issue nor the changes make any sense to me, right? Should I close the PR & the issue?

j-white: Yeah, it's no longer a problem now that we've moved to Drools.

Earlier releases could benefit from the change, but we don't want to modify the configuration files there. I'd say close it.

PR #2461, opened 2019-04-18
NMS-10658: Exterminate alarm-change-notifier plugin

fooker:
* JIRA: http://issues.opennms.org/browse/NMS-10658


fooker:
I'm not sure if `AlarmElasticsearch5IT.java` still has a purpose. @j-white can you have a look?

j-white:
That test can be deleted altogether. We also need to remove `opennms.alarmChangeNotifier.events.xml` and the relevant code in the `opennms-es-rest` module.

PR #2465, opened 2019-04-22
HZN-1521: Edges should allow to contain vertices of different namespaces

patrick-schweizer:
* JIRA: https://issues.opennms.org/browse/HZN-1521


features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericEdge.java
```diff
@@ -39,24 +39,37 @@
...
+    }
+
+    public GenericEdge(GenericVertexRef source, GenericVertexRef target, Map<String, Object> properties) {
         super(properties);
         this.source = Objects.requireNonNull(source);
         this.target = Objects.requireNonNull(target);
-        this.setNamespace(source.getNamespace());
-        this.setId(source.getId() + "->" + target.getId());
+        // TODO patrick: discuss with mvr: can an edge connect the same vertexref? If not we want to add a check?
+        if(!source.getNamespace().equals(getNamespace()) && !target.getNamespace().equals(getNamespace())) {
```
mvrueden: This is in general a bit tricky.
This check implies, that the namespace is contained in the properties.
Is this really always the case?
The namespace should at least be enforced before trying to equal on it

Besides this, an edge can connect the same vertex, so pointing to itself. 
This may not be used often, but AFAIK it can already happen within linkd.

features/graph/api/src/test/java/org/opennms/netmgt/graph/api/generic/GenericEdgeTest.java
```diff
@@ -0,0 +1,69 @@
...
+        GenericVertexRef refWithForeignNamespace2 = new GenericVertexRef("I don't know you either, namespace", "v4");
+        new GenericEdge(NAMESPACE, refWithOurNamespace1, refWithOurNamespace2); // should throw no Excpetion
+        new GenericEdge(NAMESPACE, refWithOurNamespace1, refWithForeignNamespace1); // should throw no Excpetion
+        new GenericEdge(NAMESPACE, refWithForeignNamespace1, refWithOurNamespace1); // should throw no Excpetion
+        assertThrowsException(IllegalArgumentException.class, () -> new GenericEdge(NAMESPACE, refWithForeignNamespace1, refWithForeignNamespace2));
+        assertThrowsException(NullPointerException.class, () -> new GenericEdge(refWithForeignNamespace1, refWithForeignNamespace2, new HashMap<>()));
+    }
+
+    private static void assertThrowsException(Class<? extends Throwable> expectedException, Supplier<?> function) {
+        try {
```
mvrueden: Can you take a look at the Junit @Rules and check if we could re-write this as a Junit Rule, as this is something we need quite often.

features/graph/api/src/test/java/org/opennms/netmgt/graph/api/generic/GenericEdgeTest.java
```diff
@@ -0,0 +1,69 @@
...
+package org.opennms.netmgt.graph.api.generic;
+
+import static org.junit.Assert.*;
+
+import java.util.HashMap;
+import java.util.function.Supplier;
+
+import org.junit.Test;
+
+public class GenericEdgeTest {
```
mvrueden: We should also verify, that when adding an edge to a graph either the source or target matches the graph's namespace.

features/graph/dao/api/src/main/java/org/opennms/netmgt/topology/EdgeEntity.java
```diff
@@ -28,39 +28,39 @@
...
-    public VertexEntity getSource() {
-        return source;
+    public void setSource(VertexRef source) {
+        this.setProperty(PROPERTY_SOURCE, String.class, source.toString());
     }
 
-    public void setSource(VertexEntity source) {
-        this.source = source;
+    public VertexRef getTarget() {
+        return Optional.ofNullable(getPropertyValue(PROPERTY_TARGET)).map(VertexRef::new).orElse(null);
```
mvrueden: Using `if (...)` here makes it much more easy to read.

features/graph/dao/api/src/main/java/org/opennms/netmgt/topology/EdgeEntity.java
```diff
@@ -28,39 +28,39 @@
...
-    @JoinColumn(name="target_vertex_id", referencedColumnName = "id", nullable = false, updatable = true)
-    private VertexEntity target;
+    public void setSource(String namespace, String id) {
+        setSource(new VertexRef(namespace, id));
+    }
 
-    public VertexEntity getSource() {
-        return source;
+    public void setSource(VertexRef source) {
+        this.setProperty(PROPERTY_SOURCE, String.class, source.toString());
```
mvrueden: Wouldn't make `asStringRepresentation()` more sense instead of hijacking `toString()` for this purpose?
Maybe even allow `VertexRef` as supported property values instead.

patrick-schweizer: agree, lets make it a supported property, then we have it typesafe

features/graph/dao/api/src/main/java/org/opennms/netmgt/topology/EdgeEntity.java
patrick-schweizer: fixed

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericEdge.java
patrick-schweizer: I made sure that a GenericGraph, GenericVertex, GenericVertexRef and Edge always contain a non null namespace. This check is done in GenericElement: https://github.com/OpenNMS/opennms/blob/58756b55bc9e97ee619ec6479a958e41b48d9564/features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericElement.java#L48

features/graph/api/src/test/java/org/opennms/netmgt/graph/api/generic/GenericEdgeTest.java
patrick-schweizer: I am not sure if we can use @rules as a substitute since they are annotations to test methods or classes.
An alternative would be (in preferred order) to:
#  switch to Junit 5 which has this functionality already: https://junit.org/junit5/docs/current/api/org/junit/jupiter/api/Assertions.html#assertThrows(java.lang.Class,org.junit.jupiter.api.function.Executable)
# provide the functionality as a shared class in all of the tests in OpenNMS

features/graph/api/src/test/java/org/opennms/netmgt/graph/api/generic/GenericEdgeTest.java
patrick-schweizer: +1
done here: https://github.com/OpenNMS/opennms/blob/9e4067ebc10a759582aa2337fe56a91dc5ee43dc/features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericGraph.java#L200
and tested here:
https://github.com/OpenNMS/opennms/blob/9e4067ebc10a759582aa2337fe56a91dc5ee43dc/features/graph/api/src/test/java/org/opennms/netmgt/graph/api/generic/GenericGraphTest.java#L47

features/graph/dao/api/src/main/java/org/opennms/netmgt/topology/EdgeEntity.java
patrick-schweizer: done, see also comment above class

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericElement.java
```diff
@@ -63,23 +62,15 @@ public void setProperty(String key, Object value) {
...
-    public void setProperties(Map<String, Object> properties) {
-        this.properties.clear();
-        this.properties.putAll(properties);
-    }
-
     public Map<String, Object> getProperties() {
         return properties;
     }
 
+    // TODO: patrick: discuss with mvr if we don't want to have the namespace as a dedicated attribute in graph, vertex, edge
```
patrick-schweizer: @mvrueden , let's discuss

features/graph/dao/api/src/main/java/org/opennms/netmgt/topology/EdgeEntity.java
```diff
@@ -28,39 +28,46 @@
...
 
 @Entity
 @DiscriminatorValue("edge")
 public class EdgeEntity extends AbstractGraphEntity {
 
-    // TODO MVR we don't need this. This could be a property instead (due to vertex-ref), so source or target may not exist in the table
-    @ManyToOne(optional = false, cascade = CascadeType.ALL)
-    @JoinColumn(name="source_vertex_id", referencedColumnName = "id", nullable = false, updatable = true)
-    private VertexEntity source;
+    // TODO: Patrick: discuss with Markus if we want to have source and target as properties?
```
patrick-schweizer: @mvrueden lets discuss

features/graph/dao/api/src/main/java/org/opennms/netmgt/topology/VertexEntity.java
```diff
@@ -31,8 +31,16 @@
...
 import javax.persistence.Entity;
 
+import org.opennms.netmgt.graph.dao.api.EntityProperties;
+
 @Entity
 @DiscriminatorValue("vertex")
 public class VertexEntity extends AbstractGraphEntity {
 
+    // TODO: Patrick: discuss with mvr: we reuse "id" for 2 different things: databaseid and an identifier (part of VertexRef)
+    // we should come up with a better naming to reduce confusion?
```
patrick-schweizer: @mvrueden , lets discuss

features/graph/dao/api/src/main/java/org/opennms/netmgt/topology/VertexRef.java
```diff
@@ -0,0 +1,87 @@
...
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.netmgt.topology;
+
+
+import com.google.common.base.Objects;
+
+// TODO: Patrick discuss with mvr if the mapping from and to a property shouldn't happen through the converter service?
```
patrick-schweizer: @mvrueden lets discuss

features/graph/api/src/test/java/org/opennms/netmgt/graph/api/generic/GenericEdgeTest.java
mvrueden: I am not sure if moving to Junit 5 is reasonable at the moment. There are more pressing issues at hand. If we can simply bump the version to 5.x.y and everything still works, fine, but I doubt that would be the case. 

We could provide a similar function to `assertThrows(...)` and put it somewhere in `core/test`.

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericElement.java
mvrueden: The idea behind `Generic*` is that it has no dedicated attributes instead everything is persisted in a key/value-map.  What would be the benefit of introducing a `namespace` attribute? And if a `namespace` attribute is added, why not add an `id` attribute as well?

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericGraph.java
```diff
@@ -192,12 +197,20 @@ public void addEdge(GenericEdge edge) {
...
-            addVertex((GenericVertex) edge.getSource());
-        }
-        if (edge.getTarget().getNamespace().equalsIgnoreCase(getNamespace()) && getVertex(edge.getTarget().getId()) == null) {
-            addVertex((GenericVertex) edge.getTarget());
+        if(!this.getNamespace().equals(edge.getNamespace())){
+            throw new IllegalArgumentException(
+                    String.format("The namespace of the edge (%s) doesn't match the namespace of this graph (%s). Edge: %s ",
+                    edge.getNamespace(), this.getNamespace(), edge.toString()));
         }
+        // TODO patrick: discuss with mvr: the following casting won't work anymore since we are operating on actual Vertexrefs.
```
mvrueden: As an edge consists of two vertices, the idea was to add the vertices if they are not part of the graph yet, as they need to be added anyways. However, as an edge now only consists of `VertexRef`s, either source or target must be of type `Vertex` to add it to the Graph. To resolve a `VertexRef` it must already be known by the Graph. For now I'd say we have to ensure that the graph already knows all vertex refs of the same namespace, instead of source AND target.

features/graph/dao/api/src/main/java/org/opennms/netmgt/topology/EdgeEntity.java
```diff
@@ -28,39 +28,46 @@
...
+    // Advantage: we can use the current implementation for storage.
+    // Disadvantage: we need to filter out the source/target properties when converting to GenericEdge since these are
+    // there object attributes not properties (and should be since an edge is defined by them)
```
mvrueden: I am not sure about this.
Both variants have pros and cons.
The current graph structure is not very (postgres) database friendly. 
The only hard relation we enforce so far is: 
 - graph container -> graph
 - graph -> graph element
 - graph element -> property
 - edge -> source and target

Having source and target reference as a property is an option.
If we follow that route, we should make it part of the Converter and support it as a full type and not manually converting them.

If we use the more (postgres) database friendly approach, we should add a `VertexRefEntity` object and use that instead. 

I don't have a strong opinion, but leaning slightly towards property.
But let's have a call about this.

features/graph/dao/api/src/main/java/org/opennms/netmgt/topology/EdgeEntity.java
```diff
@@ -28,39 +28,46 @@
...
-        return target;
+    public void setTarget(String namespace, String id) {
+        setTarget(new VertexRef(namespace, id));
     }
 
-    public void setTarget(VertexEntity target) {
-        this.target = target;
+    public void setTarget(VertexRef target) {
+        Objects.requireNonNull(target, "target can not be null");
+        this.setProperty(PROPERTY_TARGET, String.class, target.toString());
```
mvrueden: FYI: This should be `target.asStringRepresentation()`

features/graph/dao/api/src/main/java/org/opennms/netmgt/topology/VertexEntity.java
mvrueden: I agree. However, this is only used in tests and should not be used by the persistence in any way.
It could be rewritten to something like `vertex.getProperty(EntityProperties.ID)` in general.

The naming is a bit confusing, but so far I don't have a better option. I am all for alternatives :)

features/graph/dao/api/src/main/java/org/opennms/netmgt/topology/VertexRef.java
```diff
@@ -0,0 +1,88 @@
...
+ *******************************************************************************/
+
+package org.opennms.netmgt.topology;
+
+
+import com.google.common.base.Objects;
+
+// TODO: Patrick discuss with mvr if the mapping from and to a property shouldn't happen through the converter service
+// leaving it like it is for now since we also might move it to a dedicated column
+public class VertexRef {
```
mvrueden: If we need to persist this, we should make this happen in the `ConverterService` and get rid of `asStringRepresentation`

features/graph/dao/api/src/main/java/org/opennms/netmgt/topology/VertexRef.java
```diff
@@ -0,0 +1,88 @@
...
+        this.id = string[1];
+    }
+
+    public VertexRef(String namespace, String id) {
+        this.namespace = namespace;
+        this.id = id;
+    }
+
+    /** Copy constructor */
+    public VertexRef(VertexRef genericVertexRef){
```
mvrueden: This is confusing with `GenericVertexRef`. I'd prefer a less confusing name, maybe just `vertexRef`?

features/graph/api/src/test/java/org/opennms/netmgt/graph/api/generic/GenericEdgeTest.java
patrick-schweizer: I looked a bit into moving to Junit5. It seems it is not as easy as to just change the version number but almost. Junit 5 is split into 3 jars:
core, api and vintage. Vintage allows to rung Junit4 tests to be run without changes in the junit 5 context => we would have to change all poms that reference junit to add the 3 jars.
See also: https://www.baeldung.com/junit-5-migration

I think this is out of scope for this pull request. If you are ok I would just leave the code as is and create a ticket to move to junit 5.
within that ticket would:
* move develop to junit 5
* remove all assertThrows() methods that have been added so far.

We could time box it to say 1 day and if it turns out to be more difficult as expected stop it. @mvrueden , @j-white what do you think?

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericElement.java
patrick-schweizer: it would make it a bit easier to set up the constructor chain, but agree we don't gain much by that. - let's leave it as a property for now.

features/graph/dao/api/src/main/java/org/opennms/netmgt/topology/VertexEntity.java
patrick-schweizer: we could add a comment to make this clear.
Just to make it clear for me: 
you say the id in the properties is a domain specific identifier and not a database id, correct?
What about the id in a graph? Isn't the namespace already domain identifier and thus a graph shouldn't have an id?

features/graph/api/src/test/java/org/opennms/netmgt/graph/api/generic/GenericEdgeTest.java
patrick-schweizer: as discussed: we will add a new ticket to evaluate the pros of Junit 5 and then decide what to do.
until then we will have a test util with the assertThrows method

patrick-schweizer: I moved it to org.opennms.core.test.OnmsAssert

features/graph/dao/api/src/main/java/org/opennms/netmgt/topology/VertexEntity.java
patrick-schweizer: ad discussed: we leave namespace as an attribute, the id as a property. Database ids are renamed to dbId.

features/graph/dao/api/src/main/java/org/opennms/netmgt/topology/EdgeEntity.java
patrick-schweizer: as discussed we will have source and target as embedded entities

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericGraph.java
patrick-schweizer: done

features/graph/dao/api/src/main/java/org/opennms/netmgt/topology/VertexRef.java
patrick-schweizer: I renamed it to VertexRefEntity since it became an (embedded) database entity

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericGraph.java
```diff
@@ -201,15 +201,16 @@ public void addEdge(GenericEdge edge) {
...
-        // and throw an exception if not.
-//        if (edge.getSource().getNamespace().equalsIgnoreCase(getNamespace()) && getVertex(edge.getSource().getId()) == null) {
-//            addVertex((GenericVertex) edge.getSource());
-//        }
-//        if (edge.getTarget().getNamespace().equalsIgnoreCase(getNamespace()) && getVertex(edge.getTarget().getId()) == null) {
-//            addVertex((GenericVertex) edge.getTarget());
-//        }
+        if (edge.getSource().getNamespace().equalsIgnoreCase(getNamespace()) && getVertex(edge.getSource().getId()) == null) {
+            throw new IllegalArgumentException(
+                    String.format("Adding an VertexRef to an unknown Vertex with id=%s in our namespace (%s). Please add the Vertex first to the graph",
```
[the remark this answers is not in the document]
patrick-schweizer: done.

mvrueden:
Cannot merge, as tests are failing.

mvrueden:
Let's squash merge this, if green.

mvrueden:
\o/ Bamboo went green

PR #2466, opened 2019-04-22
NMS-10663: better handling of missing JVMs and normalize error messages

RangerRick:
This PR cleans up the error messages in `runjava`, including better messages when a JVM is missing entirely.

* JIRA: http://issues.opennms.org/browse/NMS-10663


RangerRick:
Got the OK from @jeffgdotorg in chat, merging.

PR #2467, opened 2019-04-23
HZN-1535: Initial CircleCI Pipeline

j-white:
JIRA: https://issues.opennms.org/browse/HZN-1535

Here we add a new build pipeline for CircleCI that we can run in parallel to our existing Bamboo CI system:
image

As part of the build, we also generate Docker containers and use these containers to run the smoke/system tests based on the Testcontainers framework.

All of these components still need work, but the pipeline is in a working state now and we would like to start running CircleCI against all builds based off develop to help get an idea of cost going forward.

Bamboo will continue to be our primary CI while we evaluate using CircleCI.


.circleci/README.md
```diff
@@ -0,0 +1,25 @@
+# Smoke Tests
+
+## Running locally
+
+Download .octi artifacts
```
fooker: Should be ".oci artifacts", right?

j-white:
Merging as discussed in our scrum meeting today.

PR #2468, opened 2019-04-23
HZN-1496: Add RPC Metrics and graphs

cgorantla:
Expose JMX Metrics for RPC ( RequestSent, RequestFailed, RequestSize, ResponseSize, Duration )
and build graphs by default.

* JIRA: http://issues.opennms.org/browse/HZN-1496


core/ipc/rpc/api/src/main/java/org/opennms/core/rpc/api/RpcClientFactory.java
```diff
@@ -35,7 +35,14 @@
  */
 public interface RpcClientFactory {
 
-    static final String LOG_PREFIX = "ipc";
+    // RPC Metrics related constants.
+    public static final String JMX_DOMAIN_RPC = "org.opennms.core.ipc.rpc";
```
j-white: We can omit the `public` and `static` qualifiers here since we're in an interface.

core/ipc/rpc/api/src/main/java/org/opennms/core/rpc/api/RpcClientFactory.java
```diff
@@ -35,7 +35,14 @@
...
+    public static final String LOG_PREFIX = "ipc";
```
j-white: Should the LOG_PREFIX be lumped up with the other "RPC Metric related constants"?

core/ipc/rpc/camel/src/main/java/org/opennms/core/rpc/camel/CamelRpcClientFactory.java
```diff
@@ -184,9 +205,14 @@ public void setLocation(String location) {
...
         tracerRegistry.init(SystemInfoUtils.getInstanceId());
         tracer = tracerRegistry.getTracer();
+        // Initialize metrics reporter.
+        metricsRepoter = JmxReporter.forRegistry(metrics).
+                inDomain(JMX_DOMAIN_RPC).build();
+        metricsRepoter.start();
     }
 
     public void stop() {
         //pass
```
j-white: We can remove the `// pass` line now that we actually have something in this method. (I normally use that in methods I define that have an empty body.)

core/ipc/rpc/camel/src/main/java/org/opennms/core/rpc/camel/CamelRpcClientFactory.java
```diff
@@ -81,6 +86,9 @@
 
     private Tracer tracer;
 
+    private MetricRegistry metrics = new MetricRegistry();
+    private JmxReporter metricsRepoter = null;
```
j-white: `metricsReporter` vs `metricsRepoter ` (same spelling in other classes too)

core/ipc/rpc/camel/src/main/java/org/opennms/core/rpc/camel/CamelRpcClientFactory.java
```diff
@@ -184,9 +205,14 @@ public void setLocation(String location) {
...
         tracer = tracerRegistry.getTracer();
+        // Initialize metrics reporter.
+        metricsRepoter = JmxReporter.forRegistry(metrics).
+                inDomain(JMX_DOMAIN_RPC).build();
+        metricsRepoter.start();
     }
 
     public void stop() {
         //pass
+        metricsRepoter.close();
```
j-white: Can we make sure metricsRepoter is not null before closing? Calling stop before calling start would throw an NPE.

PR #2469, opened 2019-04-26
NMS-10626: Event parms using named capture groups in varbind matchers

j-white:
JIRA: https://issues.opennms.org/browse/NMS-10626

Values extract from varbinds using named capture groups like:
```
<vbvalue>~Node /(?<poolName>.*?)/(?<poolMember>\S+) address (?<poolAddr>\S+) monitor status down. .*\(slot(?<slotNum>[0-9]+)\)</vbvalue>
```

will now be automatically added as event parameters to the events at expansion time.


opennms-config-api/src/main/java/org/opennms/netmgt/config/api/EventMatch.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * For more information contact:
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.netmgt.config.api;
+
+public class EventMatch {
```
RangerRick: This class isn't used anywhere, as far as I can tell...

PR #2470, opened 2019-04-29
HZN-1536: Add metrics/graphs for IPC Sink

cgorantla:
* JIRA: http://issues.opennms.org/browse/HZN-1536


opennms-base-assembly/src/main/filtered/etc/resource-types.d/jmx-resource.xml
```diff
@@ -27,4 +27,23 @@
             <parameter key="replace-all" value="s/.*\.//"/>
         </storageStrategy>
     </resourceType>
+
+  <resourceType name="sinkProducerMetrics" label="Sink Producer Metrics"
```
j-white: Nit: Indentation isn't consistent with the rest of the file.

PR #2472, opened 2019-04-30
HZN-1540: Added option to search for nodes with or without flow data

christianpape:
* JIRA: https://issues.opennms.org/browse/HZN-1540

### DIFF
--- a/opennms-web-api/src/main/java/org/opennms/web/svclayer/support/DefaultNodeListService.java
+++ b/opennms-web-api/src/main/java/org/opennms/web/svclayer/support/DefaultNodeListService.java
@@ -163,9 +163,6 @@ public class DefaultNodeListService implements NodeListService, InitializingBean
     }
 
     private void addCriteriaForFlows(final OnmsCriteria criteria, final Boolean hasFlows) {
-        if (hasFlows == null) {
-            return;
-        }
         criteria.add(Restrictions.eq("node.hasFlows", hasFlows));
     }
 
--- a/opennms-web-api/src/main/java/org/opennms/web/svclayer/support/DefaultNodeListService.java
+++ b/opennms-web-api/src/main/java/org/opennms/web/svclayer/support/DefaultNodeListService.java
@@ -151,6 +151,8 @@ public class DefaultNodeListService implements NodeListService, InitializingBean
             addCriteriaForForeignSource(criteria, command.getForeignSource());
         } else if(command.hasMonitoringLocation()) {
             addCriteriaForMonitoringLocation(criteria, command.getMonitoringLocation());
+        } else if(command.hasFlows()) {
+                addCriteriaForFlows(criteria, command.getFlows());
         } else {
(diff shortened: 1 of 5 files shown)


smoke-test/src/test/java/org/opennms/smoketest/NodeListPageIT.java
```diff
@@ -88,6 +92,20 @@ private void createNode(final String foreignId, final String location) throws Ex
         sendPost("/rest/nodes", node, 201);
     }
 
+    @Test
+    public void testNodesWithFlows() {
+        this.m_driver.get(this.getBaseUrl() + "opennms/element/nodeList.htm?flows=true");
```
mvrueden: The usage of `this` and `m_` is weird, I'd omit the `this` here.

smoke-test/src/test/java/org/opennms/smoketest/NodeListPageIT.java
```diff
@@ -88,6 +92,20 @@ private void createNode(final String foreignId, final String location) throws Ex
...
+        assertThat(Iterables.transform(m_driver.findElements(By.xpath("//div[@class='NLnode']//a")), WebElement::getText),
+                containsInAnyOrder("TestMachine loc1node2", "TestMachine loc2node2"));
+    }
+
+    @Test
+    public void testNodesWithoutFlows() {
+        this.m_driver.get(this.getBaseUrl() + "opennms/element/nodeList.htm?flows=false");
```
mvrueden: The usage of `this` and `m_` is weird, I'd omit the `this` here.

opennms-webapp/src/main/webapp/element/index.jsp
```diff
@@ -200,6 +200,21 @@
                   </div>
               </div>
           </form>
+
+          <%-- Search by flow data --%>
+          <form role="form" class="form-group" action="element/nodeList.htm" method="get">
+              <input type="hidden" name="listInterfaces" value="false"/>
+              <label for="byflows_flows">Flows:</label>
```
mvrueden: Can we not use the `:`, as all the other labels don't have a colon anymore.

opennms-web-api/src/main/java/org/opennms/web/svclayer/support/DefaultNodeListService.java
```diff
@@ -160,6 +162,13 @@ private void addCriteriaForCommand(OnmsCriteria criteria, NodeListCommand comman
         }
     }
 
+    private void addCriteriaForFlows(final OnmsCriteria criteria, final Boolean hasFlows) {
+        if (hasFlows == null) {
```
mvrueden: This is a duplicated check, as this is only invoked after `command.hasFlows()` is true, so not required to invoke it again. If you absolutely want this check, I'd opt for 

```java
if (hasFlows != null) {
 ...
}
```
as it is much easier to read.

PR #2478, opened 2019-05-01
NMS-10676: fix shell exit codes for stop/start

RangerRick:
This PR fixes exit-code handling in a few of the functions and adds test coverage for `.rpmnew`/`.rpmsave`/`.dpkg-dist` checking.

* JIRA: http://issues.opennms.org/browse/NMS-10676

### DIFF
--- a/container/shared/src/main/resources/etc/container.init
+++ b/container/shared/src/main/resources/etc/container.init
@@ -90,6 +90,35 @@ run_as() {
 	fi
 }
 
+check_upgrade_files() {
+	for EXTENSION in rpmnew rpmsave dpkg-dist; do
+		if [ "$(find "$CONTAINER_HOME" -name \*."$EXTENSION" 2>/dev/null | wc -l)" -gt 0 ]; then
+			cat <<END >&2
+
+WARNING!  You have files that end in .$EXTENSION in your
+$CONTAINER_HOME directory.
+
+The format of the original files may have changed since
+you modified them before installing a new version.
+Please double-check that your configuration files are
+up-to-date and delete any leftover .$EXTENSION files or
+$DESC will not start.
+
+END
+			case "$COMMAND" in
+				status)
+					# when calling with `status` return 3 for "stopped"
+					return 3	# From LSB: 3 - program is stopped
+					;;
+				*)
+					# any other init command should return 6 on error
+					return 6	# From LSB: 6 - program is not configured
+					;;
+			esac
+		fi
+	done
+}
+
 get_root_pid() {
 	ROOT_INSTANCE_PID=0
 	if [ -f "$CONTAINER_HOME/instances/instance.properties" ];
@@ -132,6 +161,10 @@ case "$COMMAND" in
 		if is_running; then
 			echo "$DESC is running."
 			exit 0
+		fi
+
+		if ! check_upgrade_files; then
+			exit $?
 		fi
 
 		case "$PING_REQUIRED" in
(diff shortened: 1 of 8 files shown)


container/shared/src/main/resources/etc/container.init
```diff
@@ -215,6 +252,10 @@ case "$COMMAND" in
 		;;
 
 	status)
+		if ! check_upgrade_files; then
```
j-white: Is it necessary to run this check when calling `status`?

RangerRick: I suppose not... this is what the old behavior was, IIRC, but there's no reason it has to check there...

RangerRick: Hm, actually I don't think that's what the old (opennms) behavior was, so never mind.  ;)

RangerRick:
OK, so the debian part should be fixed now with that change since your report, but I can't reproduce your centos 7 behavior.  If I do your `cp` commands, it fails for me with both direct calling of `$OPENNMS_HOME/bin/opennms` or with `systemctl`.  Are you sure you were running the right snapshots?

RangerRick:
oho!  figured it out on debian... we're doing the equivalent of `find $OPENNMS_HOME/etc -type f -name \*.dpkg-dist` and by default... etc is a symlink!  Need the `-L` or `-H` flag to follow links.

indigo423:
@RangerRick Just finished the install from Debian packages from branches/jira-NMS-10676 and created a dpkg-dist file:

```
root@ubuntu:/usr/share/opennms/bin# find -L /usr/share/opennms/etc -iname "*.dpkg-dist"/usr/share/opennms/etc/service-configuration.xml.dpkg-dist
```

Start opennms with

```
root@ubuntu:/usr/share/opennms/bin# ./opennms start
Starting OpenNMS: (not waiting for startup) ok
root@ubuntu:/usr/share/opennms/bin# opennms -v status
OpenNMS.Eventd         : start_pending
OpenNMS.Alarmd         : start_pending
OpenNMS.Bsmd           : start_pending
OpenNMS.Ticketer       : start_pending
OpenNMS.Trapd          : start_pending
OpenNMS.Queued         : start_pending
OpenNMS.Actiond        : start_pending
OpenNMS.Notifd         : start_pending
OpenNMS.Scriptd        : start_pending
OpenNMS.Rtcd           : start_pending
OpenNMS.Pollerd        : start_pending
OpenNMS.PollerBackEnd  : start_pending
OpenNMS.EnhancedLinkd  : start_pending
OpenNMS.Collectd       : start_pending
OpenNMS.Discovery      : start_pending
OpenNMS.Vacuumd        : start_pending
OpenNMS.EventTranslator: start_pending
OpenNMS.PassiveStatusd : start_pending
OpenNMS.Statsd         : start_pending
OpenNMS.Provisiond     : start_pending
OpenNMS.Reportd        : start_pending
OpenNMS.Ackd           : start_pending
OpenNMS.JettyServer    : start_pending
OpenNMS.Telemetryd     : start_pending
opennms is partially running
```

It's starting and it shouldn't right?

RangerRick:
and if you run `$OPENNMS_HOME/bin/opennms` directly, does it do the right thing?

indigo423:
> and if you run `$OPENNMS_HOME/bin/opennms` directly, does it do the right thing?

opennms bin looks good:

```
cd /usr/share/opennms/bin
./opennms start
```

Output:
```
Starting OpenNMS:
WARNING!  You have files that end in .dpkg-dist in your
OPENNMS_HOME (/usr/share/opennms) directory.

The format of the original files may have changed since
you modified them before installing a new version.
Please double-check that your configuration files are
up-to-date and delete any leftover .dpkg-dist files or
OpenNMS will not start.

failed
```

```
echo $?
6
```

PR #2479, opened 2019-05-01
NMS-10677: custom.properties and karaf *.cfg files should be %config

RangerRick:
This PR changes it so that the following config files will overwrite their existing counterparts by default, creating an `.rpmsave` file:

* `custom.properties`
* `jmx.*.cfg`
* `org.apache.*.cfg`
* `org.ops4j.*.cfg`
* `profile.cfg`

Anything that matches `org.opennms.*.cfg` will _not_ be replaced by default, instead creating an `.rpmnew` file.

*feedback wanted:* should I exclude `org.ops4j.pax.logging.cfg` since it's a reasonably common thing to override?  Or leave it in the "create `.rpmsave` by default" category?

* JIRA: http://issues.opennms.org/browse/NMS-10677

### DIFF
--- a/tools/packages/minion/minion.spec
+++ b/tools/packages/minion/minion.spec
@@ -167,11 +167,27 @@ mv "%{buildroot}%{minioninstprefix}/etc/minion.conf" "%{buildroot}%{_sysconfdir}
 # container package files
 find %{buildroot}%{minioninstprefix} ! -type d | \
     grep -v %{minioninstprefix}/bin | \
+    grep -v %{minioninstprefix}/etc | \
     grep -v %{minionrepoprefix} | \
-    grep -v %{minioninstprefix}/etc/featuresBoot.d | \
-    grep -v %{minioninstprefix}/etc/org.opennms.minion.controller.cfg | \
     sed -e "s|^%{buildroot}|%attr(644,minion,minion) |" | \
     sort > %{_tmppath}/files.container
+
+# org.apache.karaf.features.cfg and org.ops4j.pax.logging.cfg should
+# be special-cased to not be replaced by default (and create .rpmnew files)
+find %{buildroot}%{minioninstprefix}/etc ! -type d | \
+    grep -E 'etc/(org.apache.karaf.features.cfg|org.ops4j.pax.logging.cfg)$' | \
+    sed -e "s|^%{buildroot}|%attr(644,minion,minion) %config(noreplace) |" | \
+    sort >> %{_tmppath}/files.container
+
+# all other etc files should replace by default (and create .rpmsave files)
+find %{buildroot}%{minioninstprefix}/etc ! -type d | \
+    grep -v etc/org.opennms. | \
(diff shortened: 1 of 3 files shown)


j-white:
Should we provide the same behavior the Minion/Sentinel packages?

I tend to modify the following files on installs frequently:
* org.apache.karaf.features.cfg (add/remove features on startup - not everything works in featuresBoot.d)
* org.ops4j.pax.logging.cfg (customize logging levels for debugging specific components)

RangerRick:
Yeah, probably... I can work on doing the same in minion/sentinel.  Currently almost nothing is set as `%config`, so they mostly get blown away on upgrade.

RangerRick:
should the features config be left `%config(noreplace)` in opennms then as well?

tools/packages/minion/minion.spec
```diff
@@ -167,11 +167,27 @@ mv "%{buildroot}%{minioninstprefix}/etc/minion.conf" "%{buildroot}%{_sysconfdir}
 # container package files
 find %{buildroot}%{minioninstprefix} ! -type d | \
     grep -v %{minioninstprefix}/bin | \
+    grep -v %{minioninstprefix}/etc | \
     grep -v %{minionrepoprefix} | \
-    grep -v %{minioninstprefix}/etc/featuresBoot.d | \
-    grep -v %{minioninstprefix}/etc/org.opennms.minion.controller.cfg | \
```
j-white: We should keep the user's `org.opennms.minion.controller.cfg` too.

RangerRick: We do, it's specified down in the %files section

https://github.com/OpenNMS/opennms/blob/b3bf1dcda7565e63e1ac775b705941e36cfe2c76/tools/packages/minion/minion.spec#L251

RangerRick:
FYI, this all works (did some testing in a VM) but I do want to see if there's anything I can do on the Debian side

RangerRick:
AFAICT we'd need to do the logic ourselves in pre/post scripts on the Debian side.  :(

Long-term it might be useful to do some "compare versions of these files with empty lines and comments removed" stuff as a managed, separate script, but I'm thinking that's outside of the scope of this.

mvrueden:
The problem already starts when starting the karaf upgrade process. Some of the config files just change comments/white spaces. If you were to update those as well, they might create a conflicting file when updating the packages on debian/centos. So we may need to put this into the "How to upgrade karaf guide": https://opennms.discourse.group/t/how-to-upgrade-apache-karaf/259

RangerRick:
Yeah. My main goal is for it to create the right combination of `.rpmnew` and `.rpmsave` files that if you delete them without looking at them, you still have a running system.  ;)

On Debian you at least have the option of viewing diffs while upgrading so it's not as scary there.

Anyways, I've done testing on this stuff and it's all working as described above. Removing `triage` and letting you folks take a look so I can merge.

PR #2480, opened 2019-05-02
HZN-1551: View traffic for specific applications over time (flows) (backend)

mattixtech:
This PR adds a new ReST endpoint to enumerate the applications with flows. It also updates the `flows/applications/series` endpoint to take either a top N value or a list of specific applications to take.

In addition to the integration test updates, both endpoints have been tested manually on a local install.

* JIRA: http://issues.opennms.org/browse/HZN-1551

### DIFF
--- a/features/flows/api/src/main/java/org/opennms/netmgt/flows/api/FlowRepository.java
+++ b/features/flows/api/src/main/java/org/opennms/netmgt/flows/api/FlowRepository.java
@@ -30,7 +30,6 @@ package org.opennms.netmgt.flows.api;
 
 import java.util.Collection;
 import java.util.List;
-import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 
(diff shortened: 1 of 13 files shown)


mattixtech:
Looks like the endpoint for the flow summary also needs to be updated to support the specific set of applications. So that needs to be changed before we should merge this.

mattixtech:
Latest changes include the updates to the summary endpoint for applications.

mattixtech:
Since this PR was created I've since worked on further API enhancements for hosts and conversations that dealt with a lot of the same code which has been refactored a bit to make re-use easier. I think it will be best to look at them all in a new PR.

mattixtech:
Going to close this PR in favour of the updated PR that solves the two other JIra issues as well. Here https://github.com/OpenNMS/opennms/pull/2503

PR #2483, opened 2019-05-03
NMS-9811: Thresholds should work without restart when putting nodes into categories

smith-opennms:
This PR makes Threshd re-evalute filter package maps when a NODE_CATEGORY_MEMBERSHIP_CHANGED_EVENT is received and also when a NODE_GAINED_SERVICE_EVENT_UEI event is received.. Due to the possibly heavy handedness of this, MBean timing metrics for the FilterDao call were added and  exposed.

* JIRA: https://issues.opennms.org/browse/NMS-9811

### DIFF
--- a/opennms-config/src/main/java/org/opennms/netmgt/filter/JdbcFilterDao.java
+++ b/opennms-config/src/main/java/org/opennms/netmgt/filter/JdbcFilterDao.java
@@ -64,6 +64,10 @@ import org.springframework.cache.annotation.Cacheable;
 import org.springframework.transaction.annotation.Transactional;
 import org.springframework.util.Assert;
 
+import com.codahale.metrics.JmxReporter;
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.Timer;
+
 /**
  * <p>JdbcFilterDao class.</p>
  *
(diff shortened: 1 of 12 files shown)


opennms-config/src/main/java/org/opennms/netmgt/filter/JdbcFilterDao.java
```diff
@@ -83,6 +87,16 @@
 	private DataSource m_dataSource;
     private DatabaseSchemaConfig m_databaseSchemaConfigFactory;
 
+    private final static MetricRegistry metricRegistry = new MetricRegistry();
+    private final JmxReporter jmxReporter;
+    private final Timer getIpListTimer;
+
+    public JdbcFilterDao() {
+        getIpListTimer = metricRegistry.timer("getIPAddressList");
+        jmxReporter = JmxReporter.forRegistry(metricRegistry).inDomain("org.opennms.netmgt.config").build();
```
j-white: Let's register the reporter when the bean is initialized (see org.springframework.beans.factory.InitializingBean) and remove it when the bean is destroyed (see org.springframework.beans.factory.DisposableBean) instead if just registering the constructor.

opennms-config/src/main/java/org/opennms/netmgt/filter/JdbcFilterDao.java
```diff
@@ -83,6 +87,16 @@
 	private DataSource m_dataSource;
     private DatabaseSchemaConfig m_databaseSchemaConfigFactory;
 
+    private final static MetricRegistry metricRegistry = new MetricRegistry();
+    private final JmxReporter jmxReporter;
+    private final Timer getIpListTimer;
+
+    public JdbcFilterDao() {
+        getIpListTimer = metricRegistry.timer("getIPAddressList");
```
j-white: Can we make the MBean domain/name more descriptive. Just by looking at the MBean, it's not clear what org.opennms.netmgt.config#getIPAddressList times.

smith-opennms: so ... org.opennms.netmgt.config.FilterDao#getIPAddressListForFilter ?
image

j-white: That helps. I'd keep the FilterDao in the domain portion lower case though :octocat:

PR #2484, opened 2019-05-03
NMS-10674: Added missing 'ON DELETE CASCADE' statements to ipNetToMedia table

JonathanDIT:
https://issues.opennms.org/browse/NMS-10674: Added 'ON DELETE CASCADE' statements to newly introduced nodeid foreign key constraints from NMS-9273 and HZN-1499 on the ipNetToMedia table. 
This change enables provisiond to successfully delete nodes using its simple 'DELETE FROM NODE WHERE nodeid = %ID%' syntax.

(I opened this PR against my bug HZN-1562 but that was closed as a duplicate, hence this is now re-purposed against the open bug for the same problem)

### DIFF
--- a/core/schema/src/main/liquibase/24.0.0/changelog.xml
+++ b/core/schema/src/main/liquibase/24.0.0/changelog.xml
@@ -284,11 +284,18 @@
     </addColumn>
     
     <addForeignKeyConstraint constraintName="fk_ipnettomedia_nodeid" 
-    	baseTableName="ipnettomedia" 
-    	baseColumnNames="nodeid" 
-    	referencedTableName="node" 
-    	referencedColumnNames="nodeid" />
+        baseTableName="ipnettomedia"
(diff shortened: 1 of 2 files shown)


JonathanDIT:
I have tested the principle of this change by manually altering the two foreign key constraints in postgres - automatic deletion of nodes is then successful.

j-white:
@JonathanDIT thanks for the patch.

For the Liquibase change, we need to do this is a new changelog rather than changing an existing one, otherwise users won't be able to upgrade cleanly.

mvrueden:
The issue you created is targeted for `24.1.0`, but the PR is targeted to `develop`, can you please retarget to `release-24.1.0`.

JonathanDIT:
I have opened a replacement Pull Request ￼ https://github.com/OpenNMS/opennms/pull/2486 and will close this one.

PR #2485, opened 2019-05-03
NMS-10684: Close the JMXConnector if an exception occurs when connecting

j-white:
JIRA: https://issues.opennms.org/browse/NMS-10684

### DIFF
--- a/core/jmx/impl/pom.xml
+++ b/core/jmx/impl/pom.xml
@@ -43,5 +43,11 @@
       <artifactId>junit</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>${mockitoVersion}</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 </project>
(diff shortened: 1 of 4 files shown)


mvrueden:
This PR is targeted for `foundation-2017`, however the issue defines a fixed version of `meridian.2016.*`. Please re-target or adjust fix version accordingly.

Besides this, LGTM.

smith-opennms:
I will re-target this to ```foundation-2016```

smith-opennms:
Cancel that. Porting back to ```Foundation-2016``` is a mess.
Recommend we update the JIRA fix version and back port if there's an absolute need.

mvrueden:
Okay, I updated the jira issue. Can you please enlighten us, why it is a mess? Just for transparency reasons.

smith-opennms:
Jesse's changes were based on JmxConfig changes you made back in foundation-2017.
I tried bringing the jmxConfig changes back to foundation-2016 too, but they were based on a variety of other changes made in foundation-2017. The number of dependent changes kept growing and IMO exceeded the merits of bringing this fix back to that branch.

mvrueden:
Yeah makes sense. Thanks for sharing.

PR #2486, opened 2019-05-03
NMS-10674: Added missing 'ON DELETE CASCADE' statement to ipNetToMedia table

JonathanDIT:
This replaces PR https://github.com/OpenNMS/opennms/pull/2484 in order to cleanly base off 24.1.0 and also use the correct bug ID from the start...

- Updated 'create.sql' to set delete cascade on fk_ipnettomedia_nodeid constraint
- Created new 24.1.0 changelog.xml for Liquibase and added a changeset to recreate the  fk_ipnettomedia_nodeid constraint with onDelete=CASCADE

Edit by mvrueden: 
 - JIRA: https://issues.openns.org/browse/NMS-10674

### DIFF
--- a/core/schema/src/main/liquibase/24.1.0/changelog.xml
+++ b/core/schema/src/main/liquibase/24.1.0/changelog.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd" >
+
+  <changeSet author="jondevit" id="24.1.0-ipnettomedia-on-delete-cascade">
+    <dropForeignKeyConstraint baseTableName="ipNetToMedia" constraintName="fk_ipnettomedia_nodeid"/>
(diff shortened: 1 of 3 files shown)


JonathanDIT:
I've updated with a missing liquibase include statement as pointed out by @j-white

mvrueden:
Pulled this into Bamboo.

pbrane:
Out of curiosity, wouldn't JPA annotations on the classes cover this requirement without having to put the constraint into the DB?  Unless of course there is an external force deleting the parent records which is a different problem entirely.

mvrueden:
It depends.

If you annotate the entities properly, hibernate (or JPA) will update the entities accordingly and also flush out the data properly to the database.
However, if you were to manually delete elements on the database, you don't have hibernate. 
Therefore it is always good to define the underlying schema properly.

christianpape:
In create.sql two foreign key constraints will be modified. The liquibase changset modifies only one constraint. This seems not right for me...

JonathanDIT:
> In create.sql two foreign key constraints will be modified. The liquibase changset modifies only one constraint. This seems not right for me...

Thanks - well spotted.. I've just updated the PR. Technically we only need the first constraint to support 'delete cascade' so I could have removed the change from create.sql but this other constraint seems logical to also do the same because if a node is deleted there's no point having an MAC address referencing it any longer. Hence I've updated liquibase to handle both

j-white:
Updated the `pr/jira/NMS-10674` branch with the latest commit. Looks good to merge once the build passes.

mvrueden:
MigratorIT is now failing. See https://bamboo.opennms.org/browse/OPENNMS-ONMS2933-UNIT-2

JonathanDIT:
> MigratorIT is now failing. See https://bamboo.opennms.org/browse/OPENNMS-ONMS2933-UNIT-2

I currently don't have access to Bamboo - Can anyone point me in the direction of where I can request access please?

PR #2488, opened 2019-05-06
HZN-1556: Use intermediate base images

indigo423:
Refactoring the OCI image build using a multi stage pattern. The first stage has all repetitive and expensive run commands and enhances a CentOS OpenJDK 11 base image. This stage can be a) cached when the instructions are not changed and b) can be build separately if needed using `docker build --target horizon-base .` The same pattern is applied for Minion and Sentinel.
The Docker Layer Caching feature is enabled which allows to reuse caches over different builds.

JIRA: https://issues.opennms.org/browse/HZN-1556

## Reviewer hints

All external dependencies are moved to the first build stage named `{horizon,minion,sentinel}-base`. In this PR I've tried to reduce as much as possible to install just RPMS which are a result and an artifact of the source repository. The "external" OpenNMS RPMS like, rrdtool, jicmp, jicmp6 and jrrd2 are installed from our stable yum repository. I had to rearrange the LABELS and ARGS to the end so they don't invalidate caches unnecessarily with more expensive RUN commands.


mattixtech:
Do we still need to update the circleci job to actually cache the base images? Or is that already being done.

indigo423:
> Do we still need to update the circleci job to actually cache the base images? Or is that already being done.

so basically this change introduces a cacheable artifact. So without any changes, it is stored at docker hub, gets downloaded as container image layer instead of running all the CentOS yum install commands. I'm not sure who CircleCI caches downloaded container images.

In general we could introduce a cache similar to the maven or source code cache in circleci to download the container image and load it before we build our container image.

opennms-container/horizon/Dockerfile
```diff
@@ -1,19 +1,12 @@
-ARG BASE_IMAGE="opennms/openjdk"
-ARG BASE_IMAGE_VERSION="latest"
+ARG BASE_IMAGE="opennms/base-horizon"
+ARG BASE_IMAGE_VERSION="jdk11-1.0.0"
```
RangerRick: The only thing I worry about this is that if you fix a bug in the packaging of the intermediate container, you'll have to change every branch going forward that references this image.  Should there be a `jdk11-1` tag that can be fixed without changing the source?  Semantic versioning style...

indigo423: we provide different tags and less specific floating tags and very specific tags. For example:

* jdk11-1.0.0: which references the last successful build
* jdk11-1.0.0-cb.1035: the exact build produced this image

Floating tags are dangerous cause they make things hard to reproduce. Build a with jdk11-1.0.0 can pass and the next build with a new jdk11-1.0.0 can fail and you don't have a clue why. On the other side if you make a change without touching all references than floating tags help to populate your changes automatically.

We have to make a decision here how we want to proceed here. I've picked the version here with floating tag.

opennms-container/sentinel/config.sh
```diff
@@ -3,8 +3,8 @@
 # shellcheck disable=SC2034
 
 # Base Image Dependency
-BASE_IMAGE="opennms/openjdk"
-BASE_IMAGE_VERSION="1.8.0.191.b12-b3"
+BASE_IMAGE="opennms/base-sentinel"
+BASE_IMAGE_VERSION="jdk11-1.0.1-cb.1239"
```
indigo423: @RangerRick this is where we set the specific version number. The Dockerfile uses a sane default in case you don't set the BASE_IMAGE_VERSION variable at image build time.

opennms-container/horizon/Dockerfile
RangerRick: Gotcha.  I mentioned semver because it would be a way to attempt to mitigate breaking something on a floating tag, but if it can be overridden anyways, that solves the biggest issue.

indigo423: I'm also very open to version tag ideas. I've just picked 1.0.0 from default semver.org which we can fully control. I'm not sure if we really need 1.0.0 or 1.0 is enough + the additional build number we can't control and comes from the CI/CD to make it absolutely specific.

j-white:
To solve the issue reported in HZN-1556 it would be sufficient to use a separate layer for the package and RPM installs and enable https://circleci.com/docs/2.0/docker-layer-caching/. We don't necessary need different base images to work from, but this can work too if it doesn't add too much overhead.

indigo423:
> To solve the issue reported in HZN-1556 it would be sufficient to use a separate layer for the package and RPM installs and enable https://circleci.com/docs/2.0/docker-layer-caching/. We don't necessary need different base images to work from, but this can work too if it doesn't add too much overhead.

I'll dig into that and then we can make a decision what path to go. Thanks for pointing this out.

indigo423:
@RangerRick and @j-white Passed the Smoke tests and rebased against latest develop.

indigo423:
> To solve the issue reported in HZN-1556 it would be sufficient to use a separate layer for the package and RPM installs and enable https://circleci.com/docs/2.0/docker-layer-caching/. We don't necessary need different base images to work from, but this can work too if it doesn't add too much overhead.

Addressed this topic with using the DLC feature and using a staged build. I've changed initial PR comment accordingly.

PR #2489, opened 2019-05-06
HZN-1539: Indicators for nodes with flow data

christianpape:
* JIRA: https://issues.opennms.org/browse/HZN-1539

I'm checking for the column hasFlows of the node and snmpinterface table. Please check the screenshots. This is how the indicators for available flow data look right now. 

### Node list (outdated screenshot, see comment below for the current version)

nodelist

### Node page

Bildschirmfoto 2019-05-06 um 16 32 21

### SNMP Interfaces

snmpinterfaces

### Resource Graphs
chooseresources


indigo423:
IMHO, there are too many inconsistent visualization combinations. a) In the node list, it forces me to read the text cause icons are always there with and without flow information. So the flow icon becomes meaningless whereas in the "interface box" it is used as an indicator.

I would suggest to make the behavior the same in the node list, show the icon when there is flow data otherwise don't. I would remove the text as well and show it as a tooltip.

christianpape:
I totally agree, but this isn't easy to be also consistent with the existing stuff.
### Node list: 
You suggest to only display the flow icon when there are flows. This will leave the most installations with the upper flow button without any function and without any visual feedback since there a no flows by default. Furthermore the flow icon will be the only information without any descriptive text beside it.  
Moreover, the node list uses normal icons and not the badge style.
### SNMP interfaces: 
This is also not easy, since I don't want to waste much space with an additional column that will probably be empty in most installations. So, I choose only to display the icon at this point.
### Resource Graphs:
At this page we have enough space. I tried several variations and the badge style seems for me the best option. A simple flow icon was easy to overlook. At this point the badge is only displayed if there exist flow data.

So, if you have any consistent, concrete and precise styling suggestions that will even comply with the existing stuff, let me know.

indigo423:
> Node list:

> You suggest to only display the flow icon when there are flows. This will leave the most installations with the upper flow button without any function and without any visual feedback since there a no flows by default.

The upper buttons to toggle database/requisitions/foreign-ids/locations is a very weak UI hack. The problem you trying to solve with the text is if there is no flow data, the user has no indication if there simply is no flow data nor it doesn't work or whatever. Instead of showing on all nodes "no flow data" I would make the node toggles behave like real "toggles" and visualize if any of them are toggled or untoggled.

christianpape:
So, this is what I've changed in the last commit:

### Node list
The flow toggle button will only be displayed when there is at least one node with flow data. Flow icon and text will only appear for nodes with flow data.

Bildschirmfoto 2019-05-07 um 10 17 55

### Node page
The upper flow badge will only appear for nodes with flow data. So, no "no flow data available" badge anymore.
### SNMP interfaces
Unchanged. Simple icon badge without text.
### Resource graphs
Unchangen. Badge with "flow data available" text

mvrueden:
**Node list**
I would prefer to only show the icon instead of what it actually means. The meaning can be determined by using a tooltip or a title. I like the simplified visualization of the "snmp interface has flow data". Plus if you "toggle all options" `flow data available` and all the other data will probably not fit on the page

**Node page**
Looks fine to me.

**SNMP interfaces**
Looks fine to me

**Resource graphs**
Here I would prefer the same solution as on the SNMP Interfaces table. One icon, one meaning. No explanation required. The user will learn pretty quickly what it means. If the data is easily overlooked, we can make it a column structure where the icon is always at the same position


About the icon:
The icon is already used in the context of flows. It means "bidirectional" rule. So I am not sure if this is the best icon to use.

christianpape:
Changed the indicator's text to 'flow data'.

christianpape:
Ok, I'll merge as soon as bamboo is happy.

PR #2494, opened 2019-05-08
Changes to compile under Java 9

mvrueden:
Here are some small changes to actually compile with JDK 9. 

This will only get it to compile, but not the tests run with JDK 9.

### DIFF
--- a/features/minion/repository/pom.xml
+++ b/features/minion/repository/pom.xml
@@ -318,14 +318,6 @@
             <artifactId>org.opennms.protocols.xml</artifactId>
             <version>${project.version}</version>
         </dependency>
-        <!-- TODO MVR this is including 4.1.7 with java 8, but 4.2.1-SNAPSHOT with java 9. WTF? -->
-        <!-- TODO MVR temporarily disable -->
-        <!--<dependency>-->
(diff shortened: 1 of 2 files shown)


RangerRick:
I can make sure all the agents have jdk9 installed...

Curious, is there a reason not to go right to jdk11?  9 is kind of a deadend...

mvrueden:
There is no reason. I started looking into this, where only JDK 9 was a thing.
I took the same approach: First make it work with JDK 9, then for 11.

roskens:
pom.xml has a jdk7+ profile. This needs to be capped at JDK 10 because tools.jar removed in JDK 11.
```
     <profile>
       <id>jdk7+</id>
       <activation>
-        <jdk>[1.7,)</jdk>
+        <jdk>[1.7,10)</jdk>
       </activation>
       <build>
         <plugins>
```

The next big issue that creeps up is the jaxb libraries...

mvrueden:
These changes are not really JDK 9 specific, it just happens that OpenNMS compiles under JDK 8 without those. Therefore I am going ahead and will merge them, as it will enable others to compile under JDK 9

PR #2496, opened 2019-05-08
NMS-10690: Upgrade Maven & Plugins to latest releases

roskens:
NMS-10690: Upgrade Maven & Plugins to latest releases

* JIRA: http://issues.opennms.org/browse/NMS-10690


mvrueden:
Is there a particular reason, why updating maven and related plugins? 
Are the plugin version updates required?
I am asking, because I know that at some point upgrading the `maven-deploy-plugin` will break Bamboo.

roskens:
Articles I've read on the internet recommend updating maven & plugins to their latest versions to fix issues when running jdk9+.

mvrueden:
> Articles I've read on the internet recommend updating maven & plugins to their latest versions to fix issues when running jdk9+.

In order to make these changes work, we must change our Bamboo jobs. If we merge this as is, this will break `mvn deploy`.

j-white:
Changing the `maven-bundle-plugin` is always risky too. 

I suggest we break this out instead of trying to do a bulk upgrade, too much risk involved and we're not actually solving anything specific.

roskens:
> > Articles I've read on the internet recommend updating maven & plugins to their latest versions to fix issues when running jdk9+.
> 
> In order to make these changes work, we must change our Bamboo jobs. If we merge this as is, this will break `mvn deploy`.

How would one know if a `mvn deploy` will break?  Can we get that documented somewhere?

I can split the plugin updates up into additional commits or into additional PRs if that would be preferred.

PR #2497, opened 2019-05-09
NMS-10667: Fix health:check getting stuck

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-10667

Tested this by providing a dummy health check blocking for 5 Minutes and not implementing Thread interrupts properly.


j-white:
That should fix it, could use a test 😝 though.

PR #2498, opened 2019-05-09
NMS-10688: Clarify Java installation

indigo423:
Recommend to install OpenJDK 11 JDK before you start to install OpenNMS Horizon. Removed the JAVA_HOME section and added a more detail section [here](https://opennms.discourse.group/t/dealing-with-java-environments-on-centos-and-ubuntu/455) which discusses how to deal with Java environments which are not tight to an OpenNMS Horizon release.

* JIRA: http://issues.opennms.org/browse/NMS-10688


RangerRick:
Can't say this commit _clarifies_ much but it does at least get rid of some outdated stuff.  ;)  Is it worth moving some of that info out of discourse into the official docs along with the rest of this?

PR #2502, opened 2019-05-10
NMS-10694: upgrade to latest c3p0

RangerRick:
This PR bumps c3p0 to the latest version.  This will apply to Foundation 2016 and up.

* JIRA: http://issues.opennms.org/browse/NMS-10694


indigo423:
@RangerRick seems like this PR was not made against branch foundation.

RangerRick:
yeah, I need to redo it because foundation is probably un-fixable.  2016 and up have already reworked the c3p0 deps though so I'm gonna try it there first.

RangerRick:
Looks like everything passed, and 2016 still uses C3P0 by default so that means things should all be working correctly database-wise, so this should be good to go.

PR #2503, opened 2019-05-10
Flow API enhancements for applications, hosts and conversations

mattixtech:
This PR enhances the flows API.

- can now enumerate all hosts, applications and conversations that have flows and filter those either based on a regex or with fuzzy matching.
- the summaries/series endpoints for hosts/applications/conversations can now take either a top N parameter or a set of specific hosts/applications/conversations

Flow documents will now have a new field called `hosts` that will contain the source and destination IP addresses for the flow and in the future should be updated to also contain the DNS hostnames for those IPs if available.

This PR only contains backend changes for the OpenNMS ReST api.

Tested all the changes manually using softflowd.

* JIRA: http://issues.opennms.org/browse/HZN-1551
* JIRA: http://issues.opennms.org/browse/HZN-1552
* JIRA: http://issues.opennms.org/browse/HZN-1554


features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/ConversationKeyUtils.java
```diff
@@ -49,15 +48,24 @@
  */
 public class ConversationKeyUtils {
     private static final Gson gson = new GsonBuilder().create();
+    private static final ConversationKey otherKey = new ConversationKey("Other", -1, "Other", "Other", "Other");
 
     public static ConversationKey fromJsonString(String json) {
+        if(json.equals("Other")) {
```
j-white: This seems sneaky

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/ElasticFlowRepository.java
```diff
@@ -291,35 +288,133 @@ public void enrichAndPersistFlows(final List<FlowDocument> flowDocuments, FlowSo
...
+                                                            String applicationPattern, long limit,
+                                                            List<Filter> filters) {
+        // Handle the unquoted null value
+        if(applicationPattern.equals(".*")) {
+            applicationPattern = String.format("(\\\"%s\\\"|null)", applicationPattern);
+        } else if (!applicationPattern.equals("null")) {
+            applicationPattern =String.format("\\\"%s\\\"", applicationPattern);
+        }
+        
+        String regex = String.format("\\[\\\"%s\\\",%s,\\\"%s\\\",\\\"%s\\\",%s\\]",
```
j-white: Building a regex to search the JSON string seems like it could problematic and lead to a lot of edge cases. Can we search the individual fields instead?

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/ConversationKeyUtils.java
mattixtech: Wasn't sneaky enough apparently!

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/ElasticFlowRepository.java
mattixtech: We could do this but I believe it will require adding additional fields to the document to handle searching the source/dest hosts accurately. Will defer it for now.

j-white:
Looks good. Can we also add the new `hosts` fields in the Elasticsearch template: https://github.com/OpenNMS/opennms/blob/7236ebd7c62b090b359a4c32eb8766896812762b/features/flows/elastic/src/main/resources/netflow-template.json#L762

mattixtech:
I pushed the change to the template and verified it is still working. I currently have the `hosts` field in the root of the document but it looks like you were pointing at putting it a part of the `netflow` object (as in netflow.hosts). Does this matter? I can change it.

j-white:
As long as it matches the name of property it's fine - looks good as is.

Looks like there's a conflict to fix now though.

mattixtech:
Merged with latest, should be conflict free now.

PR #2505, opened 2019-05-15
NMS-10678: Fix reload when drools engine encounters exception

cgorantla:
Fix the case where there is possibility of sending facts(events) while
the state being marshalled.

* JIRA: http://issues.opennms.org/browse/NMS-10678

### DIFF
--- a/opennms-correlation/drools-correlation-engine/src/main/java/org/opennms/netmgt/correlation/drools/DroolsCorrelationEngine.java
+++ b/opennms-correlation/drools-correlation-engine/src/main/java/org/opennms/netmgt/correlation/drools/DroolsCorrelationEngine.java
@@ -63,7 +63,6 @@ import org.opennms.netmgt.correlation.drools.config.EngineConfiguration;
 import org.opennms.netmgt.correlation.drools.config.RuleSet;
 import org.opennms.netmgt.events.api.EventConstants;
 import org.opennms.netmgt.model.events.EventBuilder;
-import org.opennms.netmgt.xml.event.AlarmData;
 import org.opennms.netmgt.xml.event.Event;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -82,6 +81,8 @@ import com.google.common.io.ByteStreams;
  * @version $Id: $
  */
 public class DroolsCorrelationEngine extends AbstractCorrelationEngine {
+
+    public static final boolean PRESERVE_STATE_WHEN_EXCEPTION = Boolean.getBoolean("org.opennms.netmgt.correlation.drools.preserveState");
     private static final Logger LOG = LoggerFactory.getLogger(DroolsCorrelationEngine.class);
 
     private KieBase m_kieBase;
@@ -122,6 +123,10 @@ public class DroolsCorrelationEngine extends AbstractCorrelationEngine {
     /** {@inheritDoc} */
     @Override
     public synchronized void correlate(final Event e) {
+        if (m_kieSession == null) {
+            LOG.info("No valid session");
+            return;
+        }
         LOG.debug("Begin correlation for Event {} uei: {}", e.getDbid(), e.getUei());
         m_kieSession.insert(e);
         try {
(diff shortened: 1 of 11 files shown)


cgorantla:
By Default, when Drools Engine encounters exception, it will reload the engine but not load any facts or recover any state.  This can be changed by setting system property `org.opennms.netmgt.correlation.drools.preserveState`.

Main fix here is to prevent facts being fired when `marshallStateToDisk` or `shutdownSession` is going on

UberPinguin:
This looks like it should work.

opennms-correlation/drools-correlation-engine/src/main/java/org/opennms/netmgt/correlation/drools/DroolsCorrelationEngine.java
```diff
@@ -122,6 +123,10 @@ public ApplicationContext getConfigContext() {
     /** {@inheritDoc} */
     @Override
     public synchronized void correlate(final Event e) {
+        if (m_kieSession == null) {
+            LOG.info("No valid session");
```
j-white:
```suggestion
            LOG.info("No valid session exists. Event with id: {} and UEI: {} will not be added as a fact.", e.getDbid(), e.getUei());
```

opennms-correlation/drools-correlation-engine/src/main/java/org/opennms/netmgt/correlation/drools/DroolsCorrelationEngine.java
```diff
@@ -136,6 +141,10 @@ public synchronized void correlate(final Event e) {
     /** {@inheritDoc} */
     @Override
     protected synchronized void timerExpired(final Integer timerId) {
+        if (m_kieSession == null) {
+            LOG.info("No valid session");
```
j-white:
```suggestion
            LOG.info("No valid session exists. Timer with id: {} will not be added as a fact.", timerId);
```

opennms-correlation/drools-correlation-engine/src/main/java/org/opennms/netmgt/correlation/drools/DroolsCorrelationEngine.java
```diff
@@ -381,7 +405,11 @@ public void reloadConfig() {
             EngineConfiguration cfg = JaxbUtils.unmarshal(EngineConfiguration.class, m_configPath);
             Optional<RuleSet> opt = cfg.getRuleSetCollection().stream().filter(rs -> rs.getName().equals(getName())).findFirst();
             if (opt.isPresent()) {
-                marshallStateToDisk(true);
+                if (PRESERVE_STATE_WHEN_EXCEPTION) {
```
j-white: This is misleading since the method may also be called during a normal reload.

cgorantla:
Updated code to  save facts in memory instead of marshalling state to disk for all soft reload cases ( reloadDaemonConfig).
Also added the ability to specify  `persistState`  as param in reloadDaemonConfig event ( by default it reloads previous state)

For the case of reloading drools engine after it encounters exception, engine will reload previous state only if  user specifies  `-Dorg.opennms.netmgt.correlation.drools.reloadState=true`

cgorantla:
I will add more unit tests to verify new `savefacts` functionality

UberPinguin:
I just want to double-check - will this preserve the current behavior where the `persist-state="true"` attribute on an engine is honored? Or will it _require_ adding `-Dorg.opennms.netmgt.correlation.drools.reloadState=true` and pass a `persistState` parm on the reload event ?

cgorantla:
> I just want to double-check - will this preserve the current behavior where the `persist-state="true"` attribute on an engine is honored? Or will it _require_ adding `-Dorg.opennms.netmgt.correlation.drools.reloadState=true` and passing a `persistState` parm on the reload event ?

When user sends reload event, state will be persisted and reloaded by default and user has option to set   event param `persistState=false`  to not to reload state.

When engine encounters exception, it will reload engine again by default but will not reload state and user has option to set `-Dorg.opennms.netmgt.correlation.drools.reloadState=true`  to reload state after exception.

May be good to change `reloadState` to 'reloadStateAfterException` so that it is easy to understand.

cgorantla:
Modified the property to `-Dorg.opennms.netmgt.correlation.drools.reloadStateAfterException=true`

RangerRick:
this passed in bamboo, FYI

PR #2509, opened 2019-05-16
HZN-1520: Remove unused dependency net.simon04.jelementtree:jelementtree

roskens:
HZN-1520 removed a dependency on net.simon04.jelementtree:jelementtree but it wasn't removed from opennms-base-assembly/pom.xml.


mvrueden:
Pulled this into Bamboo.

mvrueden:
LGTM, Thank you very much for your contribution.

PR #2510, opened 2019-05-17
NMS-10701: make sure all Jetty code is using 9.4.18

RangerRick:
A Jetty version bump.  Also, we were not changing Karaf to match our embedded Jetty.  I've changed it to use `${jettyVersion}` instead.

* JIRA: http://issues.opennms.org/browse/NMS-10701

### DIFF
--- a/features/springframework-security/src/main/java/org/opennms/web/springframework/security/JettyUserIdentityFilter.java
+++ b/features/springframework-security/src/main/java/org/opennms/web/springframework/security/JettyUserIdentityFilter.java
@@ -1,8 +1,8 @@
 /*******************************************************************************
  * This file is part of OpenNMS(R).
  *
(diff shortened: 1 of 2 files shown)


j-white:
foundation-2018 is also on Jetty 9.4.14 so it would be worth bumping it there as well

we don't actively use the jetty feature in Karaf AFAIK, so those changes wouldn't be necessary for f18

mvrueden:
We are not using the jetty Karaf feature and actually we cannot at the moment.
The jetty feature is just there, because it is a standard Karaf feature.
For the sake of simplicity, I'd say just let's keep it to whatever the Karaf guys set it to.
This will Make Karaf Updates much easier from my point.

RangerRick:
This actually passed, bamboo status updates aren't working right (what you see here is me testing scripts to replace the broken bamboo plugin).  I'm gonna merge.

PR #2513, opened 2019-05-18
NMS-10690: Update Maven & Plugins : maven enforcer plugin

roskens:
* JIRA: http://issues.opennms.org/browse/NMS-19690

This PR adds the maven-enforcer-plugin to ensure the version of maven used to compile the source is the same as the version bundled under the maven/ directory.

It also enforces running a specific version of java instead of an maven-ant-plugin target task.

### DIFF
--- a/checkstyle/pom.xml
+++ b/checkstyle/pom.xml
@@ -15,6 +15,29 @@
           <target>1.8</target>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-enforcer-plugin</artifactId>
+        <version>3.0.0-M3</version>
+        <executions>
+          <execution>
+            <id>enforce-maven</id>
+            <goals>
+              <goal>enforce</goal>
+            </goals>
+            <configuration>
+              <rules>
+                <requireMavenVersion>
+                  <version>[3.5,)</version>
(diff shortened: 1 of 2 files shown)


mvrueden:
What is the purpose of enforcing a certain maven version?

RangerRick:
Yeah, I could see maybe making it a minimum version of 3.5 just to be sure we're not hitting ancient bugs that don't exist anymore, but I don't understand locking it specifically to 3.5.2.

roskens:
> Yeah, I could see maybe making it a minimum version of 3.5 just to be sure we're not hitting ancient bugs that don't exist anymore, but I don't understand locking it specifically to 3.5.2.

Simply this: At the time of this PR, the bundled version of maven included with the source code was 3.5.2.

mvrueden:
I agree with @RangerRick here. If we really want to enforce a maven version, let's only define a minimum version. But in general I don't see the point of doing this.

fooker:
I'm not sure if the java version is to strict, too?

RangerRick:
For now, I would say no.  Obviously the eventual work on making the build pass with JDK 11 will require it to be changed, but we literally can't build with > JDK8 right now.

I missed that @roskens already updated the maven limit to be more forgiving, I'm going to go ahead and merge this.

PR #2515, opened 2019-05-21
NMS-10654: When editing a surveillance category from Admin flow, lists of nodes are not sorted by node label (reopened)

jeffgdotorg:
* JIRA: http://issues.opennms.org/browse/NMS-10654

### DIFF
--- a/opennms-webapp/src/main/webapp/WEB-INF/jsp/admin/editCategory.jsp
+++ b/opennms-webapp/src/main/webapp/WEB-INF/jsp/admin/editCategory.jsp
@@ -2,8 +2,8 @@
 /*******************************************************************************
  * This file is part of OpenNMS(R).
  *
- * Copyright (C) 2006-2014 The OpenNMS Group, Inc.
- * OpenNMS(R) is Copyright (C) 1999-2014 The OpenNMS Group, Inc.
+ * Copyright (C) 2006-2019 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2019 The OpenNMS Group, Inc.
  *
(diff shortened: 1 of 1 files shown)


jeffgdotorg:
@RangerRick if you don't mind having a quick look at this, I would appreciate it. I used the ES6-specific construct `for (const foo of bar)` and would like to be sure that's unlikely to cause issues.

RangerRick:
> @RangerRick if you don't mind having a quick look at this, I would appreciate it. I used the ES6-specific construct `for (const foo of bar)` and would like to be sure that's unlikely to cause issues.

IE doesn't support it at all, so depends on whether you care about that.  ;)  (same with `const` for that matter...)

In our other javascript stuff it goes through a babel compile stage so it gets turned into ES5, but if you're doing it straight in a JSP I'd still do it the old way.

opennms-webapp/src/main/webapp/WEB-INF/jsp/admin/editCategory.jsp
```diff
@@ -73,10 +73,14 @@ function populateOptGroupFromList(elementName, list) {
 		optgroupElem.remove(0);
 	}
 	
-	for (var nodeId in list) {
+	for (const nodeSpec of list) {
```
RangerRick: The (somewhat) efficient way to do this in es5 is:

```
for (var i=0, len=list.length, nodeSpec; i < len; i++) {
    nodeSpec = list[i];
```

jeffgdotorg: Thanks. Glad I asked :) I'll rework it to do this instead.

PR #2524, opened 2019-05-28
HZN-1581: Fixed type mismatch for VLAN value

christianpape:
* JIRA: https://issues.opennms.org/browse/HZN-1581


mvrueden:
This was reported as "working" by a customer.

mvrueden:
We discussed yesterday about adding tests here.
If too complicated tests may be omitted, as customer reported this as working.

j-white:
We've seen a number of these type mismatches happen now. What's the challenge in adding tests for this? Something like: Mock the data, use the classes to write the BSON document, read the BSON document and make sure that all of the fields can be parsed

christianpape:
Ok, you are right. I will start to work on a test to cover these errors...

christianpape:
Added tests. Please review...

PR #2525, opened 2019-05-28
HZN-1582: Increase logging clarity for flow processing

Naicisum:
* JIRA (Issue Tracker): http://issues.opennms.org/browse/HZN-1582


mvrueden:
LGTM. Should we target this for 24.x instead?

Naicisum:
That would be my preference

mvrueden:
I re-targeted it to `release-24.1.1`.

PR #2526, opened 2019-05-28
NMS-10716: Introduce rancidApiVersion property

roskens:
dependencies/rancid/pom.xml has a dependency on org.opennms/rancid-api/1.0.4 while container/features/src/main/resources/features-experimental.xml has a dependency on org.opennms/rancid-api/1.0.3.

This PR introduces a new property, rancidApiVersion, and updates the two xml files to use that property.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-10716


mvrueden:
Pulled this into Bamboo

PR #2530, opened 2019-05-30
NMS-12075: Provide Grafana Endpoint (UI + ReST)

mvrueden:
* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12075

I decided to target it against `features/drift2` as this feature still contains some TODOs which we will work on at a later state.

Ideally we would have a generic endpoints endpoint, but given the tough time frame, I decided to only focus on a Grafana Endpoint.


opennms-webapp/src/main/webapp/admin/index.jsp
```diff
@@ -133,6 +133,7 @@
         <ul class="list-unstyled mb-0">
             <li><a href="admin/sysconfig.jsp">System Configuration</a></li>
             <li><a href="admin/userGroupView/index.jsp">Configure Users, Groups and On-Call Roles</a></li>
+            <li><a href="admin/endpoint/index.jsp">Configure Grafana Endpoints (Reports only)</a></li>
```
j-white: Can we bring this down on the page? Maybe in "Additional Tools" on in a new section.

Doesn't seem important or accessed frequently enough to be at the top.

core/web-assets/src/main/assets/js/apps/onms-endpoints/grafana/grafana-modal.html
```diff
@@ -0,0 +1,95 @@
...
+  <div class="modal-header">
+    <h5 class="modal-title">{{endpoint.id === undefined ? "Add" : "Edit"}} Grafana Endpoint</h5>
+  </div>
+  <div class="modal-body">
+    <form name="endpointForm" novalidate>
+      <div class="form-group">
+        <div ng-show="error.entity" class="form-text text-danger">{{error.entity}}</div>
+      </div>
+      <div class="form-group form-row">
+        <label class="col-form-label" for="endpoint.uid">Grafana ID</label>
```
j-white: This field is called "Unique ID" in the table, but it's called "Grafana ID" in the modal. Is that intentional?

core/web-assets/src/main/assets/js/apps/onms-endpoints/grafana/grafana-modal.html
```diff
@@ -0,0 +1,95 @@
...
+        <input class="form-control" type="text" id="endpoint.uid" name="uid" placeholder="Unique ID" ng-model="endpoint.uid"
```
j-white: When editing this item in the modal, the changes are also reflected immediately in the table that's in the background. These shouldn't change until the values are valid and saved.

core/web-assets/src/main/assets/js/apps/onms-endpoints/grafana/grafana-modal.html
```diff
@@ -0,0 +1,95 @@
...
+               ng-class="{ 'is-invalid' : endpointForm.uid.$invalid || error.uid }" required/>
+        <div ng-show="endpointForm.uid.$invalid" class="invalid-feedback">The Unique ID is required.</div>
```
j-white: Can we enforce some validation on the unique id field? Only alpha-numeric perhaps?

opennms-webapp/src/main/webapp/admin/index.jsp
mvrueden: Done.

core/web-assets/src/main/assets/js/apps/onms-endpoints/grafana/grafana-modal.html
mvrueden: Oh, good catch. Originally the page was supposed to hold all endpoints and differentiate via a type column. This is now gone, so the column should be called `Grafana ID` instead.

core/web-assets/src/main/assets/js/apps/onms-endpoints/grafana/grafana-modal.html
mvrueden: Done

core/web-assets/src/main/assets/js/apps/onms-endpoints/grafana/grafana-modal.html
mvrueden: Sure. I added validation. It can contain all alpha-numeric values as well as - and _, but may not start with those.

mvrueden:
> This seems to work fine. I understand that it is necessary to have these TODOs in the code but can you remove the "TODO MVRs" from the UI tooltips?

Done.

mvrueden:
Addressed all feedback and fixed the tests. If no more feedback is provided I am going to merge this as soon as it goes green.

PR #2535, opened 2019-06-05
NMS-12087: Reverse hostname lookups for flows

fooker:
* http://issues.opennms.org/browse/NMS-12087


features/themes/onms-default-theme/src/main/java/VAADIN/themes/opennms/styles.css
```diff
@@ -0,0 +1,8436 @@
+@import url(../reindeer/legacy-styles.css);
```
j-white: Added by mistake?

features/flows/api/src/main/java/org/opennms/netmgt/flows/api/FlowRepository.java
```diff
@@ -74,4 +76,51 @@
     CompletableFuture<Table<Directional<String>, Long, Double>> getHostSeries(Set<String> hosts, long step, boolean includeOther, List<Filter> filters);
 
     CompletableFuture<Table<Directional<String>, Long, Double>> getTopNHostSeries(int N, long step, boolean includeOther, List<Filter> filters);
+
+    class Conversation {
```
j-white: Can we move this to a separate file?

features/flows/elastic/src/test/java/org/opennms/netmgt/flows/elastic/ConversationKeyUtilsTest.java
```diff
@@ -72,4 +72,43 @@ public void canCreateAndParseConversationKey() {
...
+        String outKey = ConversationKeyUtils.getConvoKeyAsJsonString(flowOut);
+
+        // We should have generated some key, and both should match
+        assertThat(inKey, notNullValue());
+        assertThat(inKey, equalTo(outKey));
+
+        ConversationKey expectedKey = new ConversationKey(
+                flowIn.getLocation(),
+                flowIn.getProtocol(),
+                flowIn.getSrcAddr() + " (" + flowIn.getSrcAddrHostname() + ")",
```
j-white: Do we still need this test? I thought we were no longer including the hostname in the convo key.

opennms-doc/guide-admin/src/asciidoc/text/telemetryd/introduction.adoc
```diff
@@ -50,3 +50,23 @@ If a _Listener_ its _Parsers_ is running on _Minion_, the received messages will
...
+Some _Parsers_ support reverse hostname resolution to annotate IP addresses with the actual hostname.
+If enabled, the reverse name lookup will be done on the _Minion_ while parsing the protocol data.
+The resolved hostname, if found, is attached to the address information and both will be send the the _Adapter_ for further processing.
+
+To enable reverse hostname resolution, the following command can be used:
+[source]
+----
+$ ssh -p 8201 admin@localhost
+...
+admin@minion()> system:property -p org.opennms.features.telemetry.enable true
```
j-white:
```suggestion
admin@minion()> system:property -p org.opennms.features.telemetry.dns.enable true
```

features/telemetry/common/src/test/java/org/opennms/netmgt/telemetry/common/utils/DnsUtilsTest.java
```diff
@@ -0,0 +1,323 @@
...
+
+public class DnsUtilsTest {
+
+    private Properties systemProperties;
+
+    @Before
+    public void before() {
+        this.systemProperties = new Properties();
+
+        DnsUtils.bundleContext = new BundleContext() {
```
j-white: Can we use a mock here instead of implementing the complete interface?

features/flows/elastic/src/test/java/org/opennms/netmgt/flows/elastic/ConversationKeyUtilsTest.java
fooker: removed

features/telemetry/common/src/test/java/org/opennms/netmgt/telemetry/common/utils/DnsUtilsTest.java
fooker: @Spy looks neat

features/telemetry/common/src/main/java/org/opennms/netmgt/telemetry/common/utils/DnsUtils.java
```diff
@@ -0,0 +1,149 @@
...
+
+        lookup.setResolver(resolver);
+
+        final Record records[] = lookup.run();
+
+        if (lookup.getResult() == Lookup.SUCCESSFUL) {
+
+            final PTRRecord ptrRecord = (PTRRecord) Arrays.stream(records)
+                    .filter(PTRRecord.class::isInstance)
+                    .findFirst()
```
Naicisum: Can we check `records` length and log a WARN or INFO if more than one record was returned?

pbrane:
This feature should use the "node label" if the location+IP address from the flow already exists in the DB.

Naicisum:
> This feature should use the "node label" if the location+IP address from the flow already exists in the DB.

This can be an issue, the nodes in opennms do not have a temporal time element. As such, if a flow from the past is IP A, and you then later on add a different node using IP A, the information will be incorrect when looking at the old flow. It needs to rely on dns resolution at the time the flow is processed for maximum reliability.

fooker:
> Looks great!
> 
> Any thought about extending the DnsUtils class to include a rudimentary short lived TTL (5 seconds per record) cache?

I've made a new issue for that: https://issues.opennms.org/browse/NMS-12117

features/flows/rest/impl/src/main/java/org/opennms/netmgt/flows/rest/internal/FlowRestServiceImpl.java
```diff
@@ -389,6 +395,52 @@ private static long getEffectiveEnd(long end) {
...
+            public String schwurbel(final String ip, final String hostname) {
+                if (!Strings.isNullOrEmpty(hostname)) {
+                    return hostname;
+                } else {
+                    return ip;
+                }
+            }
+        };
+
+        public abstract String schwurbel(final String ip, final String hostname);
```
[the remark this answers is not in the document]
fooker: `schwurbel` means doing something non-deterministic / non-scientific which seems to be a perfect fit.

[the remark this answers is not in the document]
fooker: changed to `buildDisplayName`

features/flows/api/src/main/java/org/opennms/netmgt/flows/api/FlowRepository.java
```diff
@@ -74,4 +76,51 @@
...
+            this.location = location;
+            this.protocol = protocol;
+            this.lowerIp = lowerIp;
+            this.upperIp = upperIp;
+            this.lowerHostname = lowerHostname;
+            this.upperHostname = upperHostname;
+            this.application = application;
+        }
+
+        public Conversation withHostnames(final Optional<String> lowerHostname,
```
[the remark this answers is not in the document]
fooker: Hope I got what you're talking about...

fooker:
> > This feature should use the "node label" if the location+IP address from the flow already exists in the DB.
> 
> This can be an issue, the nodes in opennms do not have a temporal time element. As such, if a flow from the past is IP A, and you then later on add a different node using IP A, the information will be incorrect when looking at the old flow. It needs to rely on dns resolution at the time the flow is processed for maximum reliability.

We can store this in elasticsearch during collection time. But I would opt for doing this in a separate issue...

features/flows/api/src/main/java/org/opennms/netmgt/flows/api/Conversation.java
```diff
@@ -0,0 +1,152 @@
...
+public class Conversation {
+    private final String location;
+    private final Integer protocol;
+    private final String lowerIp;
+    private final String upperIp;
+    private final Optional<String> lowerHostname;
+    private final Optional<String> upperHostname;
+    private final String application;
+
+    private Conversation(final String location,
```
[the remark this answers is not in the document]
fooker: That's neat. Never thought about doing it this way.

fooker:
@j-white I've added support for the host-queries.
@mattixtech Can you have a (hopefully) last look again?

features/flows/api/src/main/java/org/opennms/netmgt/flows/api/Host.java
```diff
@@ -0,0 +1,82 @@
...
+package org.opennms.netmgt.flows.api;
+
+import java.util.Objects;
+import java.util.Optional;
+
+public class Host {
+    private final String ip;
+    private final Optional<String> hostname;
+
+    public Host(final String ip, final Optional<String> hostname) {
```
mattixtech: Minor nit: this constructor should take a `String` and not an `Optional` as the parameter for hostname.

pbrane:
> > This feature should use the "node label" if the location+IP address from the flow already exists in the DB.
> 
> This can be an issue, the nodes in opennms do not have a temporal time element. As such, if a flow from the past is IP A, and you then later on add a different node using IP A, the information will be incorrect when looking at the old flow. It needs to rely on dns resolution at the time the flow is processed for maximum reliability.

The problem is that the node in the DB has a very high chance to have a different "label".  The associated label at the time of the flow will be persisted to ES the same as if it was done via DNS.  The node label can change just as the canonical name may change in DNS but we should prefer the node label if it exists as if we could add this to resolve.conf:

hosts:      opennms files dns

fooker:
While fixing the last unti tests, I found a conceptional problem with the host-related queries: If we store the resolved hostnames beside the IPs in the `hosts` field, a couple of thing break apart. First, the top-N queries will return bogus results, as the same host is inserted twice - so results will show the topmost speaker by its hostname and by its IP address in two distinct records. Second, the queries to match IPs with associated resolved hostnames will break, as we will query for the hostname of something that already is a hostname, which will fail by concept and as the hostname can not be parsed as an IP address as required by the fields we are filtering against.

My current idea is to not add the hostnames to the `hosts` field, but do a additional lookups while we query for the content of this field. This will solve the problems above, but does not allow to filter by hostnames any more.

@j-white and @mattixtech any thoughts about that?

Naicisum:
> While fixing the last unti tests, I found a conceptional problem with the host-related queries: If we store the resolved hostnames beside the IPs in the `hosts` field, a couple of thing break apart. First, the top-N queries will return bogus results, as the same host is inserted twice - so results will show the topmost speaker by its hostname and by its IP address in two distinct records. Second, the queries to match IPs with associated resolved hostnames will break, as we will query for the hostname of something that already is a hostname, which will fail by concept and as the hostname can not be parsed as an IP address as required by the fields we are filtering against.
> 
> My current idea is to not add the hostnames to the `hosts` field, but do a additional lookups while we query for the content of this field. This will solve the problems above, but does not allow to filter by hostnames any more.
> 
> @j-white and @mattixtech any thoughts about that?

I though the goal was to create new fields to persist the hostname information? This would prevent the conflict you have outlined. And when displaying the info use a `if hostname exist then use hostname else use ip'.

We need to always keep the IPs as apart of the elastic record. The hostname resolution should be optional fields in the elastic records that are null if empty or set to string if resolved.

fooker:
> I though the goal was to create new fields to persist the hostname information? This would prevent the conflict you have outlined. And when displaying the info use a `if hostname exist then use hostname else use ip'.
>
> We need to always keep the IPs as apart of the elastic record. The hostname resolution should be optional fields in the elastic records that are null if empty or set to string if resolved.

That's exactly what we're doing here. The problem is only related to the `getHosts` and `getHost*` queries. The `hosts` field contains a list of all addresses in record and is used to query for top talkers. We've thought that it will be a clever idea to add the hostnames to this list, too. But it turned out to cause a bunch of problems as the content of the field is used to query other fields (the list of hosts as seen in grafana is build by this field, but the filter works against another one) which results in type errors if these are not real IPs.
The second problem is, that we may end up adding the same host twice to the `hosts` field (once by IP and once by hostname) which will result in errors in the top talkers query. You will see the same host twice (once with the hostname already in the list and once with the IP - which in turn will be replaced by the IPs hostname).

The current state does NOT add the hostnames to the `hosts` field. Therefore the hostnames will not show up in the hosts lists in grafana. But the hostnames will be resolved for the `getHost*` queries.

I think to replace the IPs in the `getHosts` query requires to return both, the IP and the address to grafana so we can display the address but use the IP for further filtering. If we decide to do so, I would opt for doing it in another issue, as it involves changes to the grafana plugin.

fooker:
> Any thought about extending the DnsUtils class to include a rudimentary short lived TTL (5 seconds per record) cache?

Turns out, there was a cache already inside of this lousy DNS library. I've made it configurable now.
The default is to have 50000 entries (which is also the default of the DNS library) and to respect the TTL. In addition, you can set the the number of entries and the maximum TTL.

Naicisum:
@j-white Do we want to merge this with `develop` or merge it into `features/drift2`?

mattixtech:
@fooker Does this break the functionality of `getHostSeries` `getHostSummaries`? It looks like they will only function using an IP address as the criteria now. Since those both use the hosts field currently.

fooker:
> @j-white Do we want to merge this with `develop` or merge it into `features/drift2`?

He said, it should go into develop.

fooker:
> @fooker Does this break the functionality of `getHostSeries` `getHostSummaries`? It looks like they will only function using an IP address as the criteria now. Since those both use the hosts field currently.

Yea, I have not touched the filters at all. So it's still not possible to filter by hostname.
I would not like to squeeze this into the same filter field as it can have some strange corner cases: i.e. hostnames change during the query timespan. Basically it just does not feel right to filter against two fields (with different types) with the same input...

j-white:
Let's defer the host handling to another issue so that we can get this in - it should still work as expected for the application/conversation queries.

Given this is fairly complete, we can merge it directly to `develop` and merge develop into `features/drift2` immediately after.

Looks like some of the smoke tests may be failing, I triggered a run in Bamboo under the new plan, it was still on the legacy plan prior to this.

PR #2536, opened 2019-06-06
NMS-10721: Fixed session handling for Netflow v9

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-10721


j-white:
Can we re-target this `release-24.1.1`?

christianpape:
It seems that these changes will not resolve the issue, but nevertheless these changes are necessary for the mentioned NF9 data to be parsed at all. We also found the cause of the problem. We're currently working on a fix...

j-white:
This seems to solve the problem with the multiple we've been observing. Probably worth merging this as-is?

fooker:
> This seems to solve the problem with the multiple we've been observing. Probably worth merging this as-is?

Makes sense

PR #2537, opened 2019-06-07
Windows is failing tests due to spurious cleanup errors

smith-opennms:
This PR changes the behavior to simply print a stack trace on Windows if there is an error cleaning up after the test.

* https://issues.opennms.org/browse/NMS-12102

### DIFF
--- a/core/test-api/collection/src/main/java/org/opennms/core/collection/test/JUnitCollectorExecutionListener.java
+++ b/core/test-api/collection/src/main/java/org/opennms/core/collection/test/JUnitCollectorExecutionListener.java
@@ -154,7 +154,16 @@ public class JUnitCollectorExecutionListener extends AbstractTestExecutionListen
             }
         }
 
-        FileUtils.deleteDirectory(m_snmpRrdDirectory);
+        try {
+            FileUtils.deleteDirectory(m_snmpRrdDirectory);
+        } catch (Exception deletionException) {
+            // Windows is failing tests due to spurious cleanup errors
+            deletionException.printStackTrace();
+            String os = System.getProperty("os.name");
+            if (os == null || !os.contains("Windows")) {
+                throw deletionException;
+            }
+        }
 
         m_fileAnticipator.tearDown();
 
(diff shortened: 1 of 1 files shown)


core/test-api/collection/src/main/java/org/opennms/core/collection/test/JUnitCollectorExecutionListener.java
```diff
@@ -154,7 +154,16 @@ public void afterTestMethod(TestContext testContext) throws Exception {
             }
         }
 
-        FileUtils.deleteDirectory(m_snmpRrdDirectory);
+        try {
+            FileUtils.deleteDirectory(m_snmpRrdDirectory);
+        } catch (Exception deletionException) {
```
j-white: Can we be more specific in the exception we catch here? Would catching an IOException have the same affect?

core/test-api/collection/src/main/java/org/opennms/core/collection/test/JUnitCollectorExecutionListener.java
```diff
@@ -154,7 +154,16 @@ public void afterTestMethod(TestContext testContext) throws Exception {
...
+            // Windows is failing tests due to spurious cleanup errors
+            deletionException.printStackTrace();
```
j-white: Can we log a warning instead of just dumping the stack trace? i.e. "Failed to delete $path during cleanup. Contents may still exist in directory for further tests."

core/test-api/collection/src/main/java/org/opennms/core/collection/test/JUnitCollectorExecutionListener.java
smith-opennms: Yes, IOException is sufficient here.
I had missed the line below which requires RuntimeException.
Fixed.

core/test-api/collection/src/main/java/org/opennms/core/collection/test/JUnitCollectorExecutionListener.java
smith-opennms: Added the logger and now carp the message plus the stacktrace.
I dropped the 2nd sentence of your suggestion, as each test within a test class gets its own unique directory. e.g. **/tmp/FileAnticipator_temp_1560178460050_3785da4a20eb680a**

PR #2538, opened 2019-06-08
Clean up serviceUnresponsive wording

mmahacek:
The wording describing serviceUnresponsive was worded oddly, and this should make it a little more clear.

Yes, I know I didn’t make a JIRA issue for this change.


j-white:
LGTM

(No JIRA is OK, we don't need one to track minor documentation updates like these)

PR #2540, opened 2019-06-10
NMS-12083: Meta-data editor for the requisitions UI

j-white:
JIRA: https://issues.opennms.org/browse/NMS-12083

Added "Meta-data" tab in requisition UI to allow users to configure node/interface/service level meta-data from one screen.

Screenshots:
image

image

Shell output after importing:
```
admin@opennms> meta:test -n NODES:abc123 -i 127.0.0.1 -s SNMP "${ttl}" | grep ttl
  ttl='30000'
  ttl='60000'
  ttl='180000'
```


core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/scripts/model/RequisitionMetaData.js
```diff
@@ -2,86 +2,182 @@
...
+        'value': entry.value
       });
     });
 
     return metaDataObject;
   };
 
+  self.getOnmsMetaDataForNode = function() {
+    return self.getOnmsMetaData(function(entry) {
+      return entry.scope === 'node';
```
fooker: We should use `Scope.NODE` instead of the literal.

core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/scripts/model/RequisitionMetaData.js
```diff
@@ -2,86 +2,182 @@
...
+    });
+  };
+
+  self.getOnmsMetaDataForInterface = function(intf) {
+    return self.getOnmsMetaData(function(entry) {
+      return entry.scope === 'interface'
```
fooker: We should use `Scope.INTERFACE` instead of the literal.

core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/scripts/model/RequisitionMetaData.js
```diff
@@ -2,86 +2,182 @@
...
+          && entry.scoped_interface.ipAddress === intf.ipAddress;
+    });
+  };
+
+  self.getOnmsMetaDataForService = function(intf, svc) {
+    return self.getOnmsMetaData(function(entry) {
+      return entry.scope === 'service'
```
fooker: We should use `Scope.SERVICE` instead of the literal.

fooker:
The behavior when deleting stuff is strange: When I add meta-data for an interface and delete the interface afterwards, I can save the requisition with the existing meta-data but it is gone after a hard reload... Maybe we should delete the associated meta-data if we delete a interface or a service?

cgorantla:
Looks good to me.  Have checked that configured metadata works in the backend.

j-white:
Deleting behavior has been fixed. Also switched from using `N/A` to `*` with a "text-muted" class.

PR #2541, opened 2019-06-10
NMS-12095: Support search and lazy loading of graphs.

cgorantla:
* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12095


j-white:
This doesn't work for me. Using Chrome with a fresh build:

image

cgorantla:
Looks like minified js expects the hierarchy of angularjs controllers in js match the html file.  

Also added rendering all graphs when print functionality is triggered,.
@RangerRick Can you help to verify print functionality on safari as that seems to be deviate from other browsers here.

j-white:
Should I give it another try now? I can't see what changed since last time since it was force pushed.

cgorantla:
> Should I give it another try now? I can't see what changed since last time since it was force pushed.

yeah. Please try again. I moved  `graphSearchBoxCtrl` on top of `checkFlowsCtrl` and added `beforeprint` functionality

j-white:
A few comments:
1. After searching, I need to scroll to make the top-most graph render (see screen)
1. After searching, the first graph that is filtered is not pulled to the top (see screen)
1. It would be nice to show some message if none of the graphs match the given search string, as opposed to just a blank page.

image

core/web-assets/src/main/assets/js/apps/onms-graph/index.js
```diff
@@ -144,7 +144,6 @@ const getDimensionsForElement = (el, def) => {
 };
 
 const render = () => {
-  console.log('GraphContainer: rendering'); // eslint-disable-line no-console
```
j-white: Can we restore the console logs for when the graphs do get rendered?

cgorantla: Now we can add log based on  graph name. Is that okay since it could be lot. The other way is to count the total and log once everything gets rendered but that won't be solve for the filtering case.  I would prefer logging for each graph based on graph name.

cgorantla: Update logger with graph name

cgorantla:
I think I fixed all the issues mentioned here. https://github.com/OpenNMS/opennms/pull/2541#issuecomment-502235368

@j-white  Please try this one more time. I will try to upload a gif with results

core/web-assets/src/main/assets/js/apps/add-to-ksc/index.js
```diff
@@ -157,9 +198,39 @@ angular.module('onms-ksc', [
...
+  $scope.enableGraph = true;
+  // Handle search query update and enable graphs with matching search query.
+  $scope.$on('handleSearchQuery', function () {
+    let searchQuery = graphSearchFactory.searchQuery;
+    // Filter on graphName or graphTitle.
+    let matchingElements = $filter('filter')([graphName, graphTitle], searchQuery);
+    if (matchingElements && matchingElements.length) {
+      $scope.enableGraph = true;
+      if(searchQuery) {
+        // Send event for the specific div to check if they are in viewport.
```
cgorantla: Note that below code is small hack to get divs to render when they are in viewport.  I tried using  angular `$viewContentLoaded` but that doesn't work.  So I didn't find more deterministic way to handle this.

j-white:
Working well for me now. Tested on IE, Chrome and FF.

core/web-assets/src/main/assets/js/apps/add-to-ksc/index.js
cgorantla: Looks like `$timeout` is the default option here for what I want to achieve ( run after the DOM has been manipulated by Angular)  as I m triggering this event from controller.

https://stackoverflow.com/a/17303759/5216060

PR #2543, opened 2019-06-14
NMS-12115: Add an optional attribute to contain the total number of returned results of the enumeration

dino2gnt:
This commit adds a special attribute for `WsManCollector` that, if specified as part of a group, will be populated by the total number of results returned by an enumeration.  This is needed because WQL has no support for a sql-like `count(*)` function and in some cases, it is beneficial to be able to threshold on the number of results returned by an enumeration as opposed to thresholding on the data itself, e.g. monitoring Windows eventlog for new events matching specific criteria.

* JIRA : http://issues.opennms.org/browse/NMS-12115


j-white:
Closing in favor of https://github.com/OpenNMS/opennms/pull/2553.

Thanks @dino2gnt !

PR #2545, opened 2019-06-17
NMS-11766: Sentinel container does not contain integration API in maven repository (Karaf Upgrade)

mattixtech:
This PR upgrades Karaf from 4.2.3 -> 4.2.6. As of 4.2.5 the behaviour of the karaf maven plugin is to omit the timestamps by default. This fixes the issue we were having on sentinel where OIA artifacts would not get resolved when using snapshots (since they were suffixed by timestamps).

I followed the instructions in container/karaf/README.markdown. Let me know if there is additional work that is required for the upgrade.

Tested locally using runInPlace.sh that the minion and sentinel containers come up. Also tested that the Opennms container comes up.

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-11766

### DIFF
--- a/opennms-full-assembly/pom.xml
+++ b/opennms-full-assembly/pom.xml
@@ -338,7 +338,7 @@
                 <feature>eventadmin</feature>
                 <feature>feature</feature>
                 <feature>framework-security</feature>
-                <feature>httplite</feature>
+                <feature>felix-httplite</feature>
                 <feature>http</feature>
                 <feature>http-whiteboard</feature>
                 <feature>instance</feature>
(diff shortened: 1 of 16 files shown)


opennms-full-assembly/pom.xml
```diff
@@ -348,6 +348,8 @@
                 <!-- TODO: Should we import both versions of Jetty? -->
                 <feature>jetty/8.1.14.v20131031</feature>
                 <feature>jetty/9.3.9.v20160517</feature>
+                <!-- TODO: Needed? -->
+                <feature>jetty/9.4.18.v20190429</feature>
```
mattixtech: I added the new version of Jetty referenced by Karaf 4.2.6 but I'm not sure if that is needed, should I be replacing one of the versions above?

mvrueden: AFAIK, we should not need the jetty features, maybe it's dependencies.

mvrueden:
This looks good to me.

opennms-full-assembly/pom.xml
mattixtech: Ok, it seems to work without those added. So I've removed them in latest commit.

PR #2546, opened 2019-06-17
NMS-12116: Synchronize access to the list of handlers to avoid CMEs.

j-white:
JIRA: https://issues.opennms.org/browse/NMS-12116

No explicit test coverage was added for this, but this code path is covered by the system tests (which is how it was detected in the first place).


j-white:
Fast tracked this one too, since it's affecting build stability.

mvrueden:
Yeah, LGTM.

PR #2549, opened 2019-06-18
NMS-12081: System tests for thresholding

j-white:
JIRA: https://issues.opennms.org/browse/NMS-12081

Here we add a new system test used to verify thresholding of response time (pollerd) and generic resources (collectd w/ JDBC collector).

We also update the system tests APIs to be able to control the file permissions for overlaid files, so that we can set the executable bit on our script.

### DIFF
--- a/smoke-test/src/main/java/org/opennms/smoketest/utils/RestClient.java
+++ b/smoke-test/src/main/java/org/opennms/smoketest/utils/RestClient.java
@@ -59,8 +59,13 @@ import org.json.XML;
 import org.opennms.core.xml.JaxbUtils;
 import org.opennms.netmgt.measurements.model.QueryRequest;
 import org.opennms.netmgt.measurements.model.QueryResponse;
+import org.opennms.netmgt.model.OnmsAlarm;
+import org.opennms.netmgt.model.OnmsAlarmCollection;
+import org.opennms.netmgt.model.OnmsCategory;
 import org.opennms.netmgt.model.OnmsEvent;
+import org.opennms.netmgt.model.OnmsEventCollection;
 import org.opennms.netmgt.model.OnmsIpInterface;
+import org.opennms.netmgt.model.OnmsMetaData;
 import org.opennms.netmgt.model.OnmsMonitoredService;
 import org.opennms.netmgt.model.OnmsNode;
 import org.opennms.netmgt.model.minion.OnmsMinion;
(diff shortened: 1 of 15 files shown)


smoke-test/src/main/java/org/opennms/smoketest/utils/RestClient.java
```diff
@@ -104,6 +109,16 @@ public RestClient(URL url, String username, String password) {
...
     }
 
+    /**
+     * Create a REST client that connects to http://localhost:8980/ with the default credentials.
+     * Used for testing.
+     *
+     * @return a new REST client for OpenNMS on the localhost
+     */
+    public static RestClient forLocalhost() {
+        return new RestClient(InetSocketAddress.createUnresolved("127.0.0.1", 8980));
```
mvrueden: 🤸‍♀

smoke-test/src/test/java/org/opennms/smoketest/ThresholdingIT.java
```diff
@@ -0,0 +1,263 @@
...
+    private static final String TEST_NODE_FS = "test";
+    private static final String TEST_NODE_FID = "thresholding-test-node";
+    private static final String TEST_NODE_CRITERIA = TEST_NODE_FS + ":" + TEST_NODE_FID;
+    private static final String TEST_NODE_CATEGORY = "ThresholdingTest";
+    private static final String TEST_SVC_NAME = "SvcToThreshold";
+
+    @ClassRule
+    public static final OpenNMSStack stack = OpenNMSStack.withModel(StackModel.newBuilder()
+            .withOpenNMS(OpenNMSProfile.newBuilder()
+                    .withFile(getResource("thresholding/poller-configuration.xml"), "etc/poller-configuration.xml")
```
mvrueden: Is it worth trying to do the `getResouce(...)` call in `withFile(String, String)` instead as it is always performed anyways?

mvrueden:
LGTM, one minor suggestion. I like the OverlayFile and OverlayUtils 💯

smoke-test/src/test/java/org/opennms/smoketest/ThresholdingIT.java
j-white: Good point, this will be pretty common, so worth adding there.

PR #2552, opened 2019-06-19
NMS-12121: Opennms 24.1.0 on Ubuntu 19.04 won’t install ....

caminac:
Ubuntu server 19.04 installs PostgresSQL 11 which is not in dependency list

added version 11 dependencies in proper file and added a row in changelog

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12121

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+
+opennms (24.1.1-1) stable; urgency=medium
+
+  * New stable release.  This release is identical to 24.1.0-1 but adds support
+    for postgresql 11 as a dependency.
+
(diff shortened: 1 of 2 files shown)


opennms-bamboo:
Welcome @caminac.

I could not find a signed OpenNMS Contributor Agreement (OCA) assigned to the username @caminac.

caminac:
> You don't need to change changelog, I'll do that when release happens, but yeah this is fine. Good catch. Guess I need to go build `iplike` packages too. ;)

yes of course
thanks

PR #2558, opened 2019-06-26
NMS-12129: Add a bash tool to help analysing config diffs

indigo423:
Added a config diff script which helps to analyze config diffs when doing upgrades or backups from 
running OpenNMS instances.

* JIRA: https://issues.opennms.org/browse/NMS-12129

[asciicast](https://asciinema.org/a/253768)


indigo423:
@RangerRick @j-white I've added a `-d` option which uses the relative paths as quick defaults which work on CentOS and Debian based distros. What do you think?

j-white:
Looks reasonable. I don't see any risk in including this into the `foundation` branch either.

PR #2559, opened 2019-06-27
NMS-12134: Update information gathered by datachoices module

jeffgdotorg:
* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12134


features/datachoices/src/main/java/org/opennms/features/datachoices/internal/UsageStatisticsReporter.java
```diff
@@ -180,9 +204,83 @@ public UsageStatisticsReportDTO generateReport() {
...
+                    if (propName.endsWith(")") && propName.length() > 1) {
+                        propName = propName.substring(0, propName.length() - 2);
+                    }
+                    featureSet.add(propName.trim());
+                }
+            }
+        } catch (IOException e) {
+            LOG.error("Encountered IOException while loading Karaf features config. Features list will be empty.");
+        }
+        for (File featureDotDFile : featuresBootDotDDir.listFiles()) {
```
j-white: Would be nice to leverage the existing logic in the `karaf-extender` instead of duplicating that here.

jeffgdotorg: Done in latest revisions. Thanks for the suggestion!

jeffgdotorg: Well, I thought I'd gotten it done and tested successfully, but now I'm getting an NPE in the for() loop. I know it's because `extender` is null. I thought that annotating the `m_extender` field with `@Reference` would get the job done. Afraid I've hit the limit of my Karaf-fu.

features/datachoices/src/main/java/org/opennms/features/datachoices/internal/UsageStatisticsReporter.java
```diff
@@ -80,6 +83,10 @@
 
 @Service
 public class UsageStatisticsReporter implements StateChangeHandler {
+    
+    @Reference
```
jeffgdotorg: @j-white I thought that this annotation would take care of wiring in the KarafExtender, but it's ending up null as evidenced by the NPE thrown every time `generateReport()` runs. Does the KarafExtender class perhaps need an `@Service` annotation to get it into the registry?

features/datachoices/src/main/java/org/opennms/features/datachoices/internal/UsageStatisticsReporter.java
jeffgdotorg: Admin guide updated.

features/datachoices/src/main/java/org/opennms/features/datachoices/internal/UsageStatisticsReporter.java
j-white: The `@Service` and `@Reference` annotations should only be used for the shell commands AFAIK.

Also, the `KarafExtender` bean isn't being exported in the OSGi registry currently, so there is no way of referencing it. You would need to update blueprint to expose it, and then reference it, or re-declare it altogether in the datachoices features.

jeffgdotorg:
@j-white I'm not sure what to do with this PR. I don't want to lose momentum on it, but it feels hopelessly messy at this point. Thoughts?

j-white:
I think the only issue pending related to enumerating the features from the featuresboot. Let's either get this to work cleanly, or tear it out for now in favor of moving forward with the other enhancements.

PR #2560, opened 2019-06-27
NMS-12133: Allow to install Horizon from a tarball

indigo423:
Add capability to install OpenNMS Horizon OCI based on a tarball artifact. The logic to install from different sources is implemented as the following:

1. Try to install from a tar.gz file in horizon/tarball
2. Try to install from local RPM files in horizon/rpms
3. Try to install from public remote mirrors

For the reason a tarball assembly doesn't have etc-pristine, we create it from a plain install so we can use the same principle to initialize configurations. Additionally the path to deal with persisted files is normalized to the /opt/opennms directory instead of using the linked ones.

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12133

### DIFF
--- a/opennms-container/horizon/Dockerfile
+++ b/opennms-container/horizon/Dockerfile
@@ -33,24 +33,28 @@ ARG ONMS_PACKAGES="opennms-core opennms-webapp-jetty opennms-webapp-remoting ope
 ARG ADD_YUM_PACKAGES
 
 COPY ./rpms /tmp/rpms
+COPY ./tarball /tmp/tarball
 
 SHELL ["/bin/bash", "-c"]
 
 # Install repositories, system and OpenNMS packages and do some cleanup
-RUN if [[ "$(ls -1 /tmp/rpms/*.rpm 2>/dev/null | wc -l)" != 0 ]]; then yum -y localinstall /tmp/rpms/*.rpm; else yum install -y ${ONMS_PACKAGES}; fi && \
+# 1. Try to install from tarball
+# 2. Try to install from local RPMS
+# 3. Try to install from public yum repository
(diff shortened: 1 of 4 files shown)


opennms-container/horizon/Dockerfile
```diff
@@ -33,24 +33,28 @@ ARG ONMS_PACKAGES="opennms-core opennms-webapp-jetty opennms-webapp-remoting ope
...
+COPY ./tarball /tmp/tarball
 
 SHELL ["/bin/bash", "-c"]
 
 # Install repositories, system and OpenNMS packages and do some cleanup
-RUN if [[ "$(ls -1 /tmp/rpms/*.rpm 2>/dev/null | wc -l)" != 0 ]]; then yum -y localinstall /tmp/rpms/*.rpm; else yum install -y ${ONMS_PACKAGES}; fi && \
+# 1. Try to install from tarball
+# 2. Try to install from local RPMS
+# 3. Try to install from public yum repository
+RUN if [[ "$(ls -1 /tmp/tarball/*.tar.gz 2>/dev/null | wc -l)" != 0 ]]; then mkdir -p /opt/opennms && tar xzf /tmp/tarball/opennms-*.tar.gz -C /opt/opennms && cp -r /opt/opennms/etc /opt/opennms/share/etc-pristine; elif [[ "$(ls -1 /tmp/rpms/*.rpm 2>/dev/null | wc -l)" != 0 ]]; then yum -y localinstall /tmp/rpms/*.rpm; else yum install -y ${ONMS_PACKAGES}; fi && \
```
Naicisum: How about inserting some wrapping for readability:
```
RUN if [[ "$(ls -1 /tmp/tarball/*.tar.gz 2>/dev/null | wc -l)" != 0 ]]; then \
        mkdir -p /opt/opennms && tar xzf /tmp/tarball/opennms-*.tar.gz -C /opt/opennms && \
        cp -r /opt/opennms/etc /opt/opennms/share/etc-pristine; \
        elif [[ "$(ls -1 /tmp/rpms/*.rpm 2>/dev/null | wc -l)" != 0 ]]; then \
        yum -y localinstall /tmp/rpms/*.rpm; else yum install -y ${ONMS_PACKAGES}; fi && \
```

indigo423: totally agree. Added your change.

PR #2562, opened 2019-06-29
NMS-11985: make all graph related classes immutable

patrick-schweizer:
### All Contributors

* [X] Have you read and followed our [Contribution Guidelines](https://github.com/OpenNMS/opennms/blob/develop/CONTRIBUTING.md)?
* [X] Have you [made an issue in the OpenNMS issue tracker](https://issues.opennms.org)?<br>If so, you should:
* [X] Have you made a comment in that issue which points back to this PR?
* [X] Have you updated the JIRA link at the bottom of this comment to link to your issue?
* [X] If this is new code, are there unit and/or integration tests?

### Pull Request Process

### External References

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-11985


features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/AllowedValuesInPropertiesMap.java
```diff
@@ -0,0 +1,116 @@
...
+
+import org.opennms.netmgt.graph.api.info.NodeInfo;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+
+/** 
+ * This class holds all allowed types that can be used inside the properties of a GenericElement. They need to be immutable and persistable.
+ * See also org.opennms.netmgt.graph.persistence.converter.ConverterService
+ * TODO: Patrick: discuss with mvr if and how we want to synchronize this list with the converters in converter service.
```
patrick-schweizer: I introduced this class in order to ensure immutability in values of the properties map. Since java doesn't offer a built in way to ensure immutability I see two ways to solve this:
1.  explicitly allow the allow types (as implemented here). This is a fast straight forward approach but needs to be extended each time if we need to add different type. I think this should be ok since we anyway need to extend the converters in the persistence layer for each new type anyway. It would be nice to synchronize those 2 lists somehow.
2. make a generic solution to detect immutability (allowed type could then be cached for speed). Here is an example how this could be achieved: https://github.com/MutabilityDetector/MutabilityDetector4FindBugs

@mvrueden, lets have a discussion which way we want to go.

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/AllowedValuesInPropertiesMap.java
```diff
@@ -0,0 +1,116 @@
...
+		ALLOWED_VALUES.add((o)->o instanceof Boolean);
+		ALLOWED_VALUES.add((o)->o instanceof Float);
+		ALLOWED_VALUES.add((o)->o instanceof Integer);
+		ALLOWED_VALUES.add((o)->o instanceof Double);
+		ALLOWED_VALUES.add((o)->o instanceof String);
+		ALLOWED_VALUES.add((o)->o instanceof Short);
+		ALLOWED_VALUES.add((o)->o instanceof Byte);
+		ALLOWED_VALUES.add((o)->o instanceof Enum);
+		ALLOWED_VALUES.add(new AllowedValueCollection());
+		// TODO: Patrick make sure NodeInfo can also be persisted, discuss with mvr the strategy
```
patrick-schweizer: @mvrueden NodeInfo have no converter so far. Serialize as JSon?

features/graph/provider/bsm/src/main/java/org/opennms/netmgt/graph/provider/bsm/BusinessServiceGraph.java
```diff
@@ -0,0 +1,66 @@
...
+
+import org.opennms.netmgt.graph.api.Graph;
+import org.opennms.netmgt.graph.api.generic.GenericEdge;
+import org.opennms.netmgt.graph.api.generic.GenericGraph;
+import org.opennms.netmgt.graph.api.generic.GenericVertex;
+import org.opennms.netmgt.graph.simple.AbstractDomainGraph;
+
+
+// TODO: Patrick: I am not sure we need such a complicated object structure since we only build the objects and then give them
+// as AbstractDomain* to the outside. Evaluate if the builders wouldn't be enough?
```
patrick-schweizer: @mvrueden what is your take?

features/graph/simple/src/main/java/org/opennms/netmgt/graph/simple/SimpleGraphContainer.java
```diff
@@ -44,6 +44,7 @@
 // TODO MVR probably implement ContainerGraphInfo instead of a property
 // TODO MVR make more type safe
 // TODO MVR why is there a DefaultGraphContainerInfo, but no SimpleGraphContainerInfo... this is weird
+// TODO Patrick make it immutable if possible
```
patrick-schweizer: @mvrueden lets discuss if it makes sense to have a container immutable. I assume similar to graph?

patrick-schweizer:
I applied the following principles:
- all immutable type have a builder
- I removed copy constructors since the objects will be immutable and thus can be used directly.
- the non abstract types are final (otherwise we can't guarantee immutability)
- the domain views can also be constructed by passing in their generic counterpart
- value checks are done in the constructor of the immutable class and will throw a Nullointer or IllegalArgument exception if a property is missing or not correct.
- The graphs are not immutable (Vertices and Edges can be added and removed) but it's properties are. @mvrueden I assume this is how it is intended?

Open:
[ ] decided on immutable & serialization check in property values
[ ] decide on immutability of graphs and graph containers
[ ] object structure in bsm

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericEdge.java
```diff
@@ -125,4 +94,27 @@ public boolean equals(Object o) {
     public int hashCode() {
         return Objects.hash(super.hashCode(), source, target);
     }
+    
+    public static GenericEdgeBuilder builder() {
+        return new GenericEdgeBuilder();
+    }
+    
+    public final static class GenericEdgeBuilder extends GenericElementBuilder<GenericEdgeBuilder> {
```
mattixtech: Some things I learned from the immutability work in the OIA project:
- Builders should have a private default constructor to restrict access to building them via the static builder method of the parent class.
- The build() method of the builder should pass a reference of the builder itself to the parent class constructor rather than individual arguments, this way its a bit cleaner as argument list can get long depending...
- Checks should be done on the builder's setters as well as the builders build() method to catch null values where not allowed or invalid values and throw an exception to ensure an object is never in an illegal state
- For objects that have 2 or less fields, provide a convenience method in addition to builder method to construct an immutable copy without having to type out the builder syntax
- If applicable provide a `from()` method that takes the interface type and copies it to an immutable via the builder
- Collections must be copied to an immutable copy otherwise your collections can be mutated by anyone with a reference to the original collection
- If you have collections of mutable types, you should convert them to immutable types if deep immutability is desired, this can get complicated depending on the types (see the classes here for how I handled this in OIA: https://github.com/OpenNMS/opennms-integration-api/tree/master/common/src/main/java/org/opennms/integration/api/v1/util)

Examples:
https://github.com/OpenNMS/opennms-integration-api/blob/master/common/src/main/java/org/opennms/integration/api/v1/model/immutables/ImmutableAlarm.java
https://github.com/OpenNMS/opennms-integration-api/blob/master/common/src/main/java/org/opennms/integration/api/v1/model/immutables/ImmutableEventParameter.java

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/AllowedValuesInPropertiesMap.java
mattixtech: Rather than a static initialization block, it would be a bit cleaner to do:

`private final static List<AllowedValue> ALLOWED_VALUES = Arrays.asList(...);`

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/AllowedValuesInPropertiesMap.java
```diff
@@ -0,0 +1,117 @@
...
+		boolean isAllowed(Object o);
+	}
+	
+	private final static class AllowedValueCollection implements AllowedValue {
+		
+		private final static List<Class<?>> KNOWN_IMMUTABLE_COLLECTIONS = new ArrayList<>();
+		static {
+			KNOWN_IMMUTABLE_COLLECTIONS.add(ImmutableList.class);
+			KNOWN_IMMUTABLE_COLLECTIONS.add(ImmutableSet.class);
+			// Please note, the following Collections are not immutable and thus not allowed:
```
mattixtech: If you copy and wrap the collection you won't have to worry about this.

Defensively copy all collections and then wrap them in Collections.unmodifiable... and then you are guaranteed the collection is immutable (not necessarily that the items are immutable though).

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/AllowedValuesInPropertiesMap.java
mattixtech: Given that the Map's value type is `Object` achieving deep immutability is likely not worth the effort.

Could we just simply state via javadoc that all elements of this collection must be immutable and persistable and then drop the explicit checking?

Let me know if this functionality is must have.

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericGraph.java
```diff
@@ -45,12 +45,17 @@
...
 import com.google.common.collect.Lists;
 
 import edu.uci.ics.jung.graph.DirectedSparseGraph;
 
-// TODO MVR enforce namespace
-public class GenericGraph extends GenericElement implements Graph<GenericVertex, GenericEdge> {
+import static com.google.common.base.Preconditions.checkArgument;
+
+/**
+ * The graph itself is not immutable (Vertices and Edges can be added and removed) but it's properties are.
```
mattixtech: Is there value in having an immutable copy of the graph? If so we could add a `immutableCopy()` method to this class that creates a new instance of something that extends `DirectedSparseGraph` and adds the vertices and edges to it then returns that. The extended class can override the `add*()` methods with to throw unsupported exceptions.

Here is an example of copying the graph https://github.com/OpenNMS/alec/pull/7/files#diff-96fac5cebdf16f3307c72076ff349d96R82

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericGraphContainer.java
```diff
@@ -39,6 +39,7 @@
 import org.opennms.netmgt.graph.api.info.GraphInfo;
 
 // TODO MVR we must rework the generic and simple objects a bit more... that is kinda weird how it is implemented. The graph could work as an example but is also not fully defined yet
+// TODO: Patrick discuss with mvr is we should make the GraphContaner partially immutable similar to GenericGraph?
```
mattixtech: This might be another candidate for adding a `immutableCopy()` method.

features/graph/provider/bsm/src/main/java/org/opennms/netmgt/graph/provider/bsm/AbstractBusinessServiceVertex.java
```diff
@@ -45,56 +48,82 @@
...
-        this.delegate.setProperty(PROPERTY_TYPE, type);
-    }
-
     public Set<String> getReductionKeys() {
         return this.delegate.getProperty(PROPERTY_REDUCTION_KEYS);
     }
 
-    public void setReductionKeys(Set<String> reductionKeys) {
-        this.delegate.setProperty(PROPERTY_REDUCTION_KEYS, reductionKeys);// TODO MVR collections cannot be persisted
+    public final static AbstractBusinessServiceVertex from(GenericVertex genericVertex) {
```
mattixtech: From what I can see, you could probably get rid of the concrete class GenericVertex and replace the few instances where you instantiate GenericVertex directly with a factory pattern for building the correct concrete type.

This `from` method should ideally just be taking an interface type as its parameter.

Once that is done you won't have an intermediate generic type and you shouldn't have to switch on its type like below. If the intent here is to still make a copy you could either add a copy method to each type or you could also do it via visitor pattern.

If there is any other cases similar to this in the PR we could solve them the same way. Otherwise the types just get too confusing in my opinion. Let me know if you want to discuss how to do this.

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/AllowedValuesInPropertiesMap.java
patrick-schweizer: In order to persist a value we use this converter service: https://github.com/OpenNMS/opennms/blob/features/graph-service/features/graph/service/src/main/java/org/opennms/netmgt/graph/persistence/converter/ConverterService.java
which limits the types of allowed values to a handful. The question is if it doesn't makes sense to fail already at construction time instead at persisting time?

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/AllowedValuesInPropertiesMap.java
patrick-schweizer: agree. Fixed.

features/graph/provider/bsm/src/main/java/org/opennms/netmgt/graph/provider/bsm/AbstractBusinessServiceVertex.java
patrick-schweizer: The idea here is that GenericVertex and it's siblings GenericEdge, GenericGraph are the classes that hold the data and the ones we normally operate on. For convenience we can generate a domain specific view on those generic objects (such as ApplicationVertex) in order to perform domain specific operations. In this case (BSMGraph) however I am not even sure if we need all these domain classes.

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericEdge.java
patrick-schweizer: Hi @mattixtech , thanks for the feedback. These are some good points!
on 1.: makes sense, will do.
on 2.: in our case it passes a GenericXX object to the parent class constructor. I guess this has the same effect (no long list of parameters)
on 3. I am not sure I understand the benefit? Why not put the checks in the constructor of the parent class when all attributes are gathered? This gives in my opinion two benefits: no code duplication, some checks can be done in the super constructor of the parent class. A builder is anyway most of the time in an incomplete state (since we are adding until build is called).
on 4.: makes good sense to me but not sure where applicable here since we deal with a list of properties
on 5: agree
on 6: agree. This is done here https://github.com/OpenNMS/opennms/blob/d047a06016eebc677e84f53faa7bb1ac22364a68/features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericElement.java#L44
on 7. agree.

mattixtech: Hey Patrick,

For #3 I just meant checks need to be done on the builders setters and also at some point when it is built. That could be done in the constructor or build method, I don't think that particularly matters. So if it is more convenient in your case to do it in the constructor that is fine.

The main thing to take away from this point is that it is useful to do the check in two places. If you have a certain field that doesn't allow nulls, the setter should check input for non-null and the build()/constructor should also check for non-null in case the setter was never called. Just doing the latter will catch illegal state but doing the former will catch illegal state sooner in a more helpful manner. Hope that makes sense.

mattixtech: Also, for # 2 what I mean is that the build() method in this class for instance should look like the following:
```
public GenericEdge build() {
            return new GenericEdge(this);
        }
```

Then the constructor can be updated to take a builder as the only parameter and fields can be retrieved from the builder. I would try to do that any time you are passing more than 2 parameters to a constructor from the builder.

mvrueden:
The goal of this issue is to prevent that the `GraphContainer` or the `Graph` is changed after it leaves the `GraphProvider` or `GraphContainerProvider`, e.g.:
 - `graphService.getGraph(“dummy”).addVertex(…)` or  - `graphService.getGraph(“dummy”).getVertex(…).setProperty(“namespace”, “newNamespace”)`
 - The same is true for the meta data of the graph/graphcontainer -> `GraphInfo` and `GraphContainerInfo`.

As long as the `Graph/GraphContainer` is within its provider scope it can and should be modifiable.

As the overall API eventually returns a `GenericGraph/GenericGraphContainer` for ALL calls (see `GraphService`), we should enforce the immutability there.

My suggestion here would be to have the providers also return a `GenericGraph/GenericGraphContainer` instead of the `Graph` and then ensure that an immutable version of that `Graph` is returned.
This will solve changing the structure (adding/removing vertices, e.g.) but will not solve issues with objects within the properties map: `graph.getVertex(“xxx”).getProperty(“some-object”).setDescription(“new description”)`. 
We could address this, but at the moment we don’t have this scenario, maybe only for `NodeInfo` and that can be solved pretty easily. Whatever we do, we should not limit the usage of objects, as technically the only requirement for the properties is, that it they are serialisable as JSON. Only if it is persisted to the database it has special requirements to the types for each property. **Meaning all graphs should be persistable (to the database), but not all will be persisted to the database.**

Ultimately if we need a deep copy, we can simply serialise it to JSON and then deserialise it to the `GenericGraph/GenericGraphContainer` and apply some pre-processing (e.g. populate the `vertexId-to-vertex`-Map, which we don't populate at all, but have to later). However let’s not worry too much about deep cloning at the moment.

The above will solve the immutability for all Generic* types.

Now only the Domain-Elements (e.g. `BusinessServiceGraph`) are left, which is also simple to fix.
Technically each Domain-Element is already a builder itself, because it is just a convenient way of accessing the generic properties map plus some helper methods, and ultimately a call to `asGeneric*` is invoked, which is similar to the `build` method of a builder.

Also each DomainGraph is only used within its `GraphProvider` and if the `GraphProvider` would return a `GenericGraph` instead of `Graph` (as mentioned above) we can implement it in the `asGenericGraph` method, which is an equivalent of the `build()` method of a builder. See https://github.com/opennms/opennms/blob/1a2ac657b08ebc39f630e270df1e3158385f2ee0/features/graph/simple/src/main/java/org/opennms/netmgt/graph/simple/AbstractDomainGraph.java#L184

That all being said, technically we don’t need any Builder at all. However I like having one, just for convenience, but we should limit those for the `Generic*` objects, including the info types.

features/graph/provider/bsm/src/main/java/org/opennms/netmgt/graph/provider/bsm/AbstractBusinessServiceVertex.java
mattixtech: In that case couldn't `GenericVertex` be an interface that the domain specific vertices implement and get rid of the intermediate concrete class. We might also be able to get rid of `AbstractBusinessServiceVertex` as a concrete class and consolidate that into the interface.

When we instantiate a GenericVertex, at that time do we have the necessary information to create it as the appropriate domain vertex? I'm assuming yes since it seems that object and its properties are meant to be immutable. If we could do that we wouldn't have to derive the type from the properties it would already be typed correctly and we could visit it to get its type.

I'm just trying to see if we could avoid converting from one concrete type to another and work with interface types rather than a concrete type. As someone not that familiar with this code it looks like the object model is unnecessarily complicated, but I could be wrong.

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/AllowedValuesInPropertiesMap.java
mvrueden: Mainly each Graph must be serialisable (to JSON) and may be persisted to the database.
We should only enforce the property type if actually persisted to the database.
Besides that, as already mentioned let's not worry too much about deep cloning, as we don't really have that problem.

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/AllowedValuesInPropertiesMap.java
mvrueden: `NodeInfo` is a calculated/enriched value and represents a node. It should never be persisted, as the data is already in the database. If a node reference must be persisted, then use the according `NodeRef` object instead, which is a simple String -> `<foreignSource>:<foreignId>`.

features/graph/provider/bsm/src/main/java/org/opennms/netmgt/graph/provider/bsm/BusinessServiceGraph.java
mvrueden: This is a dummy implementation of the actual `BusinessServiceGraphProvider`. In order to convert this graph to the `BusinessServiceGraphProvider` which we have to implement at some point we may require a differentiation between types. I'd say do whatever you want here for now, however keep in mind, that we have to properly implement this provider in general.

features/graph/simple/src/main/java/org/opennms/netmgt/graph/simple/SimpleGraphContainer.java
mvrueden: Yes, we need this as well.

features/graph/provider/bsm/src/main/java/org/opennms/netmgt/graph/provider/bsm/AbstractBusinessServiceVertex.java
mvrueden: As mentioned earlier, this is a dummy implementation of the actual `BusinessServiceGraphProvider`. It was just copied over and its purpose is to just show data. However the proper implementation of this provider is still outstanding.

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericGraph.java
mvrueden: Each `Generic*` element should have an immutable counterpart.

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericGraphContainer.java
mvrueden: Each `Generic*` element should have an immutable counterpart.

patrick-schweizer:
As discussed with @mvrueden on the phone we think we will do the following:
1. rename interface Graph to ImmutableGraph and move all setters to the builder. GenericGraph and the Domain graphs will implement this interface.
2. remove AllowedValuesInPropertiesMap.java completely and add javadoc that specigfies to add only immutable objects (as @mattixtech suggested)
3. create a shortcut for building DomainGraphElements like `ApplicationGraph.from(GenericGraph)`

patrick-schweizer:
done: 1-3
open
- [ ] make GraphContainer immutable
- [ ] disallow changing namespace after adding vertex / edges

patrick-schweizer:
@mvrueden 
I changed the following:

1. rename interface Graph to ImmutableGraph and move all setters to the builder. GenericGraph and the Domain graphs will implement this interface.
2. remove AllowedValuesInPropertiesMap.java completely and add javadoc that specigfies to add only immutable objects (as @mattixtech suggested)
3. create a shortcut for building DomainGraphElements like ApplicationGraph.from(GenericGraph)
4. make GraphContainer immutable
5. disallow changing namespace after adding vertex / edges in GraphBuilder (otherwise we face inconsistencies)
6. make all tests pass
=> I think we are ready to merge or do you have anything else in mind that we should change?

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericGraphContainer.java
```diff
@@ -45,12 +45,12 @@
 // TODO MVR we must rework the generic and simple objects a bit more... that is kinda weird how it is implemented. The graph could work as an example but is also not fully defined yet
 public class GenericGraphContainer implements ImmutableGraphContainer<GenericGraph> {
 
-    private List<GenericGraph> graphs = new ArrayList<>();
-    private Map<String, Object> properties = new HashMap<>();
+    private List<GenericGraph> graphs;
+    private Map<String, Object> properties;
```
mvrueden: Shouldn't these be final?

smoke-test/README.md
```diff
@@ -6,9 +6,11 @@ We currently leverage Docker and the TestContainers framework for running the su
 
 ## Running the system tests manually
 
-The tests require Docker images to run.
 
-### A) Pull existing images from DockerHub
+### Getting the Docker Images
+The tests require Docker images to run. There are two alternatives to get them, a) or b):
+
+#### a) Pull existing images from DockerHub
```
mvrueden: Can we make these changes in develop and revert here? Otherwise we will face merge conflicts all the time, because this will probably evolve much faster in develop than in this branch.

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericGraphContainer.java
```diff
@@ -138,4 +128,68 @@ public boolean equals(Object o) {
...
+        
+        public GenericGraphContainerBuilder properties(Map<String, Object> properties){
+            Objects.requireNonNull(properties, "properties cannot be null");
+            for (Map.Entry<String, Object> entry : this.properties.entrySet()) {
+                property(entry.getKey(), entry.getValue());
+            }
+            return this;
+        }
+        
+        public GenericGraphContainerBuilder containerInfo(GraphContainerInfo containerInfo) {
```
mvrueden: Can we make this `applyContainerInfo` or something else? 
We are not setting the containerInfo but use the data from it.
This is different from the other methods which actually set the data the method suggest they would.

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericGraphContainer.java
patrick-schweizer: good catch. fixed.

smoke-test/README.md
patrick-schweizer: makes sense, I move them.

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericGraphContainer.java
patrick-schweizer: ok. Thats a bit of a grey area since we also copy some things (e.g. collections) in other methods. But I changed it.

patrick-schweizer:
@mvrueden finally green :-)

PR #2565, opened 2019-07-02
Proposal: Provide central search functionality

mvrueden:
This PR provides the basic infrastructure for a global search functionality into OpenNMS.

# The API
 - `SearchContext`: The context of the result, e.g. `Node`, `Action`, `Report`, etc.
 - `SearchQuery`: The query to execute. Contains limit options, etc.
 - `SearchResult`: The search result of the `SearchProvider`/`SpotlightService` for a given `SearchContext`. A `SearchResult` may contain multiple `SearchResultItem`s, but each item must be of the same context. No Context Sharing!
 - `SearchResultItem`: The actual item of a search result, e.g. representing a single node or user action. Each Item may contain additional options for the representing entity, e.g. node id, node label. It is also possible to add multiple matches to indicate to the user what criteria matched (e.g. the category name, or the node label).
 - `SearchProvider` provides a `SearchResult` with zero to n items. The result may never be null, but empty. The provider is also context focused. The `SearchProvider` must encounter for the `maxResults` property, in order to not stress the database too much. Each provider should sort the result properly, in order to produce a deterministic result. 

# The Basic Concept
The `DefaultSpotlightService` will either ask all or only a single context for a `SearchResult` when `query(SearchQuery)` is invoked. 
If a `SearchResult` from that `SearchContext` has already been provided, the results are merged.

Finally the `SearchResult`s itself are ordered by their `SearchContext`. 
The `SearchResultItem`s within the `SearchResult` are ordered by their weight and label property. Due to the structure of the API the limit is also applied again, to ensure only `SearchQuery.getMaxResults()` are returned per context.

# What can be searched for at the moment?

For the following items/contexts can be searched
 - `Node`, meaning a search for an OpenNMS Node
 - `KSC Report`, meaning a search fore any KSC report
 - `Action`, meaning a user action, such as `Logout`, `Configure Requisition`, etc. is implemented.

## `Node` Context
 
 - Asset fields, except topology related fields
 - Node Label, ForeignId and Node Id
 - Node associated with a category
 - Node associated with an ip address
 - Node associated with an existing service, e.g. `ICMP`
 - Node Location
 - Any valid Filter API statement can be used to filter for those nodes.

## `KSC Report` Context

 - Report Title

## `Action` Context
 - Any existing Requisition to `Edit Requisition $1`
 - Any existing Service to `Show all Nodes with service $1`
 - Any existing Location to `Show all Nodes in location $1`
 - Any label provided in the `${OPENNMS.HOME}/etc/spotlight-actions.xml` according to the `roles` statement. If no roles are defined it is visible to ALL users, e.g. `Logout`. And yes, you can edit this, it will be considered immediately. No restart/refresh/reload required.

## How does Search work?

Navigate to the input field, or hit `Double Shift` to focus the field.
Type in your search query (must be >= 3) and it will provide a result (see above for what you can search). If there are more results, you can load them by clicking the `more...` item.
Keyboard Navigation also works: 
 - Arrow UP/Down will navigate up and down. 
 - Enter will move to the item, or reload
 - double Shift will focus the input field again
 - Esc Cancel Search

# Known issues

 - If there is an error performing the request, it will not be shown
 - sometimes the user inputs too fast and may break the UI state (not critical, but it will show a loading indicator, but isn't doing anything)
 - Is not working in Full Sized Vaadin Applications (Topology Map and Node Maps).  PR https://github.com/OpenNMS/opennms/pull/2411 will most likely fix this

# Open Issues
- Rework the UI implementation (mostly make it more angularjs friendly) and remove duplicated code
- Rework Backend to reduce duplicated code
- Remove TODOs
 - Naming: Is Spotlight save to use? Also Class naming should be consistent (sometimes it is Search, sometimes it is Spotlight)

# Future Improvements
 - The node search could be improved by only perform a single database query instead of multiple. That would may also be allow for applying an offset and limit property, instead of only the limit property. Main problem here is that multiple providers contribute to the same context, so the offset is per provider and not global.
- Implement more providers to search for even more
 - Make it more performant, I'd suspect this to be not always fast
 - Await user feedback and address it, if feasonable
 - allow for more detailled search, e.g. for certain category or all nodes with alarms or similar

Please give this a try and let me know what you think.
Also let me know which actions we should put in the `spotlight-actions.xml` by default.

DEV JAM !!!!

JIRA: https://issues.opennms.org/browse/NMS-12347


indigo423:
There are three things which are probably more a matter taste issues.

I would like it better the "Search" icon would be part of the menu and expands to an input field when you hit "double shift" or click on the icon. The way it is in the middle of the navbar looks weird to me.
search

For the background color, I would pick a different color than the light blue and pick the same grey high light color as in the requisition UI:

highlight-color

I always hit enter to trigger the search and the automatic triggered search after the 3rd characters feels a little bit weird to me. The feature is amazing and I it makes using opennms so much faster. Awesome work.

christianpape:
Awesome feature. Really nice. There are a few things I would also like to search for, e.g. menu entries or nodes of a given category. For example enter "topology" to open the Topology UI or "surveillance" to open the Surveillance view. Or to see all nodes (by using the result page of the legacy search form) of a given category by entering "routers".

indigo423:
The only thing I would change is the representation in the UI. Instead of having it somehow floating in the middle I would add before the navbar icons see screenshot attached.
click-on-search

pbrane:
> On Oct 8, 2019, at 09:34, Ronny Trommer <notifications@github.com> wrote:
> 
> The only thing I would change is the representation in the UI. Instead of having it somehow floating in the middle I would add before the navbar icons see screenshot attached.

I like that!  How about adding it *after* the navbar items?

:D

indigo423:
The supports multiple search providers, the user doesn't have an idea what he can really search for. As an idea, if we hit enter in an empty search, we show which search providers are implemented, explained here without the users to force them to read documentation :)

> Node, meaning a search for an OpenNMS Node
> KSC Report, meaning a search fore any KSC report
> Action, meaning a user action, such as Logout, Configure Requisition, etc. is implemented.

agalue:
> On Oct 8, 2019, at 9:34 AM, Ronny Trommer <notifications@github.com> wrote:
> 
> The only thing I would change is the representation in the UI. Instead of having it somehow floating in the middle I would add before the navbar icons see screenshot attached.

I like how it looks on the screenshot. But I don't like having a link right after the Search input field called Search. It looks redundant, maybe it worth removing that? Or placed it differently?

pbrane:
> On Oct 8, 2019, at 09:39, Ronny Trommer <notifications@github.com> wrote:
> 
> The supports multiple search providers,

Does it work like the search providers in the topology app?  Can you qualify the search like

[ <Node> | <KSC Report> | <Logout> : <search string> ]

mvrueden:
> Awesome feature. Really nice. There are a few things I would also like to search for, e.g. menu entries or nodes of a given category. For example enter "topology" to open the Topology UI or "surveillance" to open the Surveillance view. Or to see all nodes (by using the result page of the legacy search form) of a given category by entering "routers".

Good points. 
I added the possibility to search all menu entries (leaf elements) as well as the admin entries (if admin role is assigned). Besides this it is no possible to show all nodes of a category. 
Also added a "home" entry, where you can just get to the landing/start page.

Also reworked the ordering of actions. `Configure OpenNMS` will always be the 1st element in the actions list, whereas Admin actions come next and normal actions afterwards.

> The only thing I would change is the representation in the UI. Instead of having it somehow floating in the middle I would add before the navbar icons see screenshot attached.
> click-on-search

I am also not 100% satisfied with the current position. I'd also like to find a better way of telling the user for what she can use, because that is not clear. Also the Search item from the menubar is kind of redundant. If we implement both, the input may change as well.


> On Oct 8, 2019, at 9:34 AM, Ronny Trommer ***@***.***> wrote: The only thing I would change is the representation in the UI. Instead of having it somehow floating in the middle I would add before the navbar icons see screenshot attached.
> I like how it looks on the screenshot. But I don't like having a link right after the Search input field called Search. It looks redundant, maybe it worth removing that? Or placed it differently?

See comment above.



> On Oct 8, 2019, at 09:39, Ronny Trommer ***@***.***> wrote: The supports multiple search providers,
> Does it work like the search providers in the topology app? Can you qualify the search like [ <Node> | <KSC Report> | <Logout> : <search string> ]

The API encounters for context aware search, but it is not implemented yet.
The following contexts are implemented at the moment: `Node` and `Action`. 
Also maybe a search provider wants to be addressed explicitly. However without further user and UI feedback I left it out. Ideally the user just does not have to worry about it. If you need more complex queries, just paste a valid filter api query in there and you'll get a very unique and specific result, however only works for nodes.

mfuhrmann:
I've tested the current state and I really love it! The `double shift` function is really awesome!

I've tested nodeLabel, foreignId, category, asset and interface searches with my working infrastructure requisitions. I always got good results.
Also, I did some searches with special characters like `:` in categorynames (because I know filters in categories.xml are not working with it) without problems.

I've noticed that the action "show all nodes in category XY" is missing, when I search for a category. It already works with services.

pioto:
FWIW, on the name "Spotlight", it appears Apple still holds a trademark on the word, and just renewed it in June 2019:

* http://tmsearch.uspto.gov/bin/showfield?f=doc&state=4803:4up5n9.4.1
* https://trademarks.justia.com/785/32/spotlight-78532480.html

IANAL, etc...

mvrueden:
@mfuhrmann you probably had the wrong version. Searching for categories should work and present you with the option "Show all nodes for category Routers" for example

@pioto yeah, I removed a few references to that name already and will rename the still missing parts next

mvrueden:
There were various concerns about the Look & Feel of this feature.

I would like to address those in the next iteration.
However if you say this feature cannot merged without addressing it now, please let me know.
Otherwise I will postpone it and merge the feature when it passes circleci.

j-white:
Given that we have a lot of time before we release H26 (H25 was just released a couple weeks ago) I don't see any problem with pulling this in as-is.

mvrueden:
Not ready for review. I'd like to address the TODOs before doing that. Adding `triage` label for now. When it is ready for a Formal Review, I will request it via the UI.

mvrueden:
Ready for Review

indigo423:
agree with @j-white, I miss double shift already and my comment about UX is not meant to be a blocker.

.circleci/config.yml
```diff
@@ -301,6 +301,7 @@ workflows:
                 - /^release-.*/
                 - /^foundation.*/
                 - /^features.*/
+                - /^mvr.*/
```
j-white: Do you want to keep these here?

features/search/api/pom.xml
```diff
@@ -0,0 +1,48 @@
...
+    <groupId>org.opennms.features</groupId>
+    <artifactId>org.opennms.features.search</artifactId>
+    <version>26.0.0-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>org.opennms.features.search</groupId>
+  <artifactId>org.opennms.features.search.api</artifactId>
+  <packaging>bundle</packaging>
+  <name>OpenNMS :: Features :: Search :: API</name>
+  <properties>
```
j-white: These seem redundant

features/search/providers/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+      <groupId>org.osgi</groupId>
+      <artifactId>osgi.core</artifactId>
+      <version>${osgiVersion}</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+      <version>${slf4jVersion}</version>
+      <scope>provided</scope>
```
j-white: What's the purpose of marking many of the dependencies as `provided`? This breaks transitive dependencies - the API module already depends on slf4j and guava, so there shouldn't be a need to re-declare these here.

features/search/providers/src/main/java/org/opennms/netmgt/search/providers/action/CategorySearchProvider.java
```diff
@@ -0,0 +1,80 @@
...
+                .orderBy("name")
+                .distinct();
+        final int totalCount = categoryDao.countMatching(builder.toCriteria());
+        final Criteria criteria = builder.limit(query.getMaxResults()).toCriteria();
+        final List<OnmsCategory> matchingResult = categoryDao.findMatching(criteria);
+        final List<SearchResultItem> searchResultItems = matchingResult.stream().map(category -> {
+            final SearchResultItem searchResultItem = new SearchResultItem();
+            searchResultItem.setIdentifier(category.getId().toString());
+            searchResultItem.setLabel("Show nodes in category '" + category.getName() + "'");
+            searchResultItem.setUrl("element/nodeList.htm?category1=" + category.getName());
```
j-white: Should `category.getName()` be URL encoded?

features/search/providers/src/main/java/org/opennms/netmgt/search/providers/action/LocationSearchProvider.java
```diff
@@ -0,0 +1,80 @@
...
+                .orderBy("locationName")
+                .distinct();
+        final int totalCount = monitoringLocationDao.countMatching(builder.toCriteria());
+        final Criteria criteria = builder.limit(query.getMaxResults()).toCriteria();
+        final List<OnmsMonitoringLocation> matchingResult = monitoringLocationDao.findMatching(criteria);
+        final List<SearchResultItem> searchResultItems = matchingResult.stream().map(monitoringLocation -> {
+            final SearchResultItem searchResultItem = new SearchResultItem();
+            searchResultItem.setIdentifier(monitoringLocation.getLocationName());
+            searchResultItem.setLabel("Show nodes in location '" + monitoringLocation.getLocationName() + "'");
+            searchResultItem.setUrl("element/nodeList.htm?monitoringLocation=" + monitoringLocation.getLocationName());
```
j-white: Similarly here, should `monitoringLocation.getLocationName()` be URL encoded?

features/search/providers/src/main/java/org/opennms/netmgt/search/providers/action/RequisitionSearchProvider.java
```diff
@@ -0,0 +1,83 @@
...
+        final List<Requisition> requisitions = Lists.newArrayList(requisitionAccessService.getRequisitions())
+                .stream()
+                .filter(r -> QueryUtils.matches(r.getForeignSource(), query.getInput()))
+                .sorted(Comparator.comparing(Requisition::getForeignSource))
+                .limit(query.getMaxResults())
+                .collect(Collectors.toList());
+        final List<SearchResultItem> resultItems = requisitions.stream().map(r -> {
+                final SearchResultItem searchResultItem = new SearchResultItem();
+                searchResultItem.setIdentifier(r.getForeignSource());
+                searchResultItem.setUrl(String.format("admin/ng-requisitions/index.jsp#/requisitions/%s",r.getForeignSource()));
```
j-white: URL encoding

opennms-full-assembly/pom.xml
```diff
@@ -580,6 +580,7 @@
                 <feature>opennms-endpoints-grafana</feature>
                 <feature>opennms-reporting</feature>
                 <feature>opennms-send-event-command</feature>
+                <feature>opennms-search</feature>
```
j-white: Dependency on the impl. projects should also be added to the pom, otherwise the 🐵 will not be happy.

mfuhrmann:
I'm really looking forward for this feature! Once tested I really miss it in my production system :-)

.circleci/config.yml
mvrueden: Yes for now I'd like to keep these. I usually remove any statements which are not needed anymore after merging.

features/search/providers/pom.xml
mvrueden: When I started working on these we were still using the `maven-features-plugin` instead of manually defining the features. Therefore all the dependencies are marked as `provided`. So no purpose, just forgot to update the poms

features/search/api/pom.xml
mvrueden: Yep. Fixed.

opennms-full-assembly/pom.xml
mvrueden: I am not sure what you mean by this.

mvrueden: I now know what this means. Will add dependencies to pom.

mvrueden: Dependency monkey is happy and jars occur in system directory. All good here.

PR #2566, opened 2019-07-02
NMS-12053: Dynamically change a node's location

cgorantla:
Add new event `uei.opennms.org/nodes/nodeLocationChanged`
Handle this event in Pollerd/Collectiond to reschedule nodes

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-12053


cgorantla:
I still need to add a `smoke-test` for this. Opening PR as this can be reviewed if I miss something.

opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/DefaultProvisionService.java
```diff
@@ -249,6 +251,9 @@ public void updateNode(final OnmsNode node, String rescanExisting) {
         m_nodeDao.update(dbNode);
         m_nodeDao.flush();
 
+        if(!prevLocation.equals(currentLocation)) {
+            accumulator.sendNow(EventUtils.createNodeLocationChangedEvent("Provisiond", dbNode.getId(), dbNode.getLabel(), prevLocation, currentLocation));
```
j-white: Is there a constant we can use instead the `"Provisiond" literal? (Similar bellow too)

opennms-base-assembly/src/main/filtered/etc/events/opennms.pollerd.events.xml
```diff
@@ -254,6 +254,16 @@
         </logmsg>
       <severity>Warning</severity>
    </event>
+   <event>
+      <uei>uei.opennms.org/nodes/nodeLocationChanged</uei>
+      <event-label>OpenNMS-defined node event: nodeLocationChanged</event-label>
+      <descr>&lt;p>A currently provisioned node (%parm[nodelabel]%) changed it's
```
j-white: `it's` vs `its`

opennms-base-assembly/src/main/filtered/etc/events/opennms.pollerd.events.xml
```diff
@@ -254,6 +254,16 @@
...
+        location from (%parm[nodePrevLocation]%) to (%parm[nodeCurrentLocation]%).&lt;/p></descr>
+      <logmsg dest="logndisplay">
+        A provisioned node (%parm[nodelabel]%) changed it's location to (%parm[nodeCurrentLocation]%).
```
j-white: `it's` vs `its`

opennms-base-assembly/src/main/filtered/etc/events/opennms.pollerd.events.xml
```diff
@@ -254,6 +254,16 @@
...
+      </logmsg>
+      <severity>Warning</severity>
```
j-white: Maybe we should use `Normal` instead?

j-white:
I think we may need to make some changes to the requisitions UI as well. Here's what I tried:
1. Provisioned a node in the "Default" location (no Minions were connected to the system at this time)
2. Connected a Minion
3. Changed location on the node in the requisition UI to the new location and hit "Synchronize"

The node's location was not changed, and no `nodeLocationChanged` event was sent.

cgorantla:
> I think we may need to make some changes to the requisitions UI as well. Here's what I tried:
> 
> 1. Provisioned a node in the "Default" location (no Minions were connected to the system at this time)
> 2. Connected a Minion
> 3. Changed location on the node in the requisition UI to the new location and hit "Synchronize"
> 
> The node's location was not changed, and no `nodeLocationChanged` event was sent.

Thanks Jesse for validating although that's not consistent with what I am seeing with same steps.  Digging more. 
Screenshot from 2019-07-09 12-14-47

j-white:
Tried it again, and it now works when I switch from one location to another. If I reproduce the problem again I'll open a separate issue, but this should be good to go.

PR #2568, opened 2019-07-06
NMS-11993: Clock skew detection

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-11993


Naicisum:
Looks great! 

Any chance this even can be sent on a configurable interval? Like one per day?

pbrane:
I believe that that would imply state and I don’t believe there isn’t such a thing. If we need this sort of behavior, we should add it as a rule within the rules engine.

Naicisum:
> I believe that that would imply state and I don’t believe there is such a thing. If we need this sort of behavior, we should add it as a rule within the rules engine.

In that event, every record from a skewed device will "always" trigger an event per packet in UDP or session in TCP (IpFix Only) until the condition is corrected. This could create a significant drain on resources as events flood into the system on a 24x7 basis per device. By providing a flood gate mechanism similar in fashion to the flood control logging mechanism implemented in Apache log4x modules, we would alleviate this pressure.

pbrane:
For now, we’ll just have to mark the events with alarm data to auto clean (prune) events and keep reducing the events into a single alarm.

pbrane:
Also, I’m sure the cause of this skew detection effects more than just flow analysis so I’m pretty sure putting a silencer on it is like being okay with a time bomb ticking away. This skew should be fixed as the obviously important flows are worthless and the worthless flows are using even more system resources.

christianpape:
Currently, the events per exporter are rate-limited to one per hour. Of course, we can make this interval configurable by using an additional parameter in the parser-definition.

pbrane:
Wow.  So, is this rate limited across the Minions or Sentinels?

features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/ParserBase.java
```diff
@@ -113,6 +166,32 @@ public static ByteBuffer serialize(final Protocol protocol, final Iterable<Value
...
+        if (getMaxClockSkew() > 0) {
+            long delta = Math.abs(packetTimestamp - System.currentTimeMillis());
+            if (delta > getMaxClockSkew() * 1000L) {
+                final Optional<Instant> instant = eventCache.getUnchecked(remoteAddress);
+
+                if (!instant.isPresent() || Duration.between(instant.get(), Instant.now()).getSeconds() > getClockSkewEventRate()) {
+                    eventCache.put(remoteAddress, Optional.of(Instant.now()));
+
+                    eventForwarder.sendNow(new EventBuilder()
+                            .setUei("uei.opennms.org/internal/telemetry/clockSkewDetected")
```
j-white: Can we move this to a constant?

features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/ParserBase.java
```diff
@@ -113,6 +166,32 @@ public static ByteBuffer serialize(final Protocol protocol, final Iterable<Value
...
+                            .setTime(new Date())
+                            .setSource(getName())
+                            .setInterface(remoteAddress)
+                            .setDistPoller(identity.getId())
+                            .addParam("monitoringSystemId", identity.getId())
+                            .addParam("monitoringSystemLocation", identity.getLocation())
+                            .setParam("delta", (int) delta)
```
j-white: Can we include the unit as part of the parameter name? i.e. `deltaMs`

opennms-base-assembly/src/main/filtered/etc/events/opennms.internal.events.xml
```diff
@@ -230,4 +230,19 @@
...
+      <uei>uei.opennms.org/internal/telemetry/clockSkewDetected</uei>
+      <event-label>Clock Skew detected</event-label>
+      <descr>Clock skew detected for flow exporter</descr>
+      <logmsg dest="logndisplay">Clock skew for exporter with interface '%interface%' in location '%parm[monitoringSystemLocation]%' detected by '%parm[monitoringSystemId]%'.</logmsg>
+      <severity>Warning</severity>
+   </event>
+   <event>
+      <uei>uei.opennms.org/translator/telemetry/clockSkewDetected</uei>
+      <event-label>Clock Skew detected</event-label>
+      <descr>Clock skew detected for flow exporter</descr>
```
j-white: Can we include both the delta and the configured limits in the description?

features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/ParserBase.java
```diff
@@ -67,18 +80,57 @@
...
+                      final AsyncDispatcher<TelemetryMessage> dispatcher,
+                      final EventForwarder eventForwarder,
+                      final Identity identity) {
         this.protocol = Objects.requireNonNull(protocol);
         this.name = Objects.requireNonNull(name);
         this.dispatcher = Objects.requireNonNull(dispatcher);
+        this.eventForwarder = Objects.requireNonNull(eventForwarder);
+        this.identity = Objects.requireNonNull(identity);
+
+        setClockSkewEventRate(3600);
```
j-white: Can we make this configurable, similar to the `maxClockSkew`?

christianpape: This is configurable by parameter. This is just the default value (1 hour).

features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/ParserBase.java
christianpape: done

features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/ParserBase.java
christianpape: done

opennms-base-assembly/src/main/filtered/etc/events/opennms.internal.events.xml
christianpape: done

features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/ParserBase.java
j-white: Right.

PR #2569, opened 2019-07-08
NMS-12148: Add better datetime selection

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-12148

This does not consider time zone handling.
There is another issue for that.


mvrueden:
Closing. Mixed up the issue number. See PR https://github.com/OpenNMS/opennms/pull/2570

PR #2571, opened 2019-07-09
NMS-12141: Add ability to send events from sentinel

cgorantla:
Add ability to send events from sentinel.
Enable `sentinel-events-forwarder` feature when either `sentinel-jms` or `sentinel-kafka` is enabled.
Since this is mostly used for internal purposes, didn't add any docs for this feature.

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12141


mvrueden:
At the moment the `Sentinel` uses the `Identity` but we don't enforce it at all. 
If we follow this route, we must add documentation around the `identity` concept for sentinel and probably also adjust the docker images.

mvrueden:
My comment was related to the location of the identity.

cgorantla:
Currently all the dependent code needs is an `identity` bean to exist.  It doesn't fail if id and location are not set.  Fortunately sink code is not dependent on the `id` and `location`. 
I think for `25.0.0` we should force id and location to be set on `sentinel`. That will be useful for `OpenTracing` stuff and other rpc stuff in future.

cgorantla:
In latest commits, tried to fix smoke-test.
And I also added code to handle a corner case in large buffer size case ( kafka partition re-balance).
For this case, we now check if all the chunks are sent to same partition, if not  we will re-send the all the chunks again.

cgorantla:
full set of smoke tests are also passing in bamboo, merging now.

PR #2574, opened 2019-07-11
NMS-12158: Remove Database Reports UI (Legacy - using Spring Web Flows)

mvrueden:
Here we remove the Legacy Database Reports UI page, which was build with Spring Webflow, which is also removed with this PR.

For now I kept the old tests, which are ignored for now, as new replacements should be provided.

PR: https://issues.opennms.org/browse/NMS-12158


opennms-web-api/src/main/java/org/opennms/web/svclayer/SchedulerMessageSeverity.java
```diff
@@ -26,27 +26,12 @@
...
-     * @param triggerDetails a {@link org.opennms.web.report.database.model.TriggerDetails} object.
-     * @param context a {@link org.springframework.binding.validation.ValidationContext} object.
-     */
-    public void ValidateTriggerDetails(TriggerDetails triggerDetails, ValidationContext context) {
-        //TODO validate the cron expression here
-    }
+package org.opennms.web.svclayer;
 
+// TODO MVR delete me
+public enum SchedulerMessageSeverity {
```
mvrueden: The backend will be reworked at a later and with that these classes will be removed as well. As the focus of this PR is to remove spring-webflow the clasess `SchedulerMessageSeverity`, `SchedulerContext` and `SchedulerMessage` were created as a temporary replacement for the old `(Spring)Context`.

PR #2577, opened 2019-07-12
NMS-12161: Find Java 8 and 11

indigo423:
Changed the Debian init script to find Java 8 and Java 11 installations. On systems which have just Java 11 installed, it is otherwise not possible to start OpenNMS.

JIRA: https://issues.opennms.org/browse/NMS-12161

### DIFF
--- a/debian/opennms-server.opennms.init
+++ b/debian/opennms-server.opennms.init
@@ -38,7 +38,7 @@ fi
 
 # Check for JAVA_HOME
 if [ -z "$JAVA_HOME" ]; then
-	for dir in `find /usr/lib/jvm/* -maxdepth 0 -type d | grep java-8 | sort -u -r`; do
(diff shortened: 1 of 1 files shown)


indigo423:
@RangerRick @j-white I've branched from release-24.1.2 as a target for this PR. Why does the branch release-24.1.2 does not have a `.circleci` configuration?

j-white:
`.circleci` is only in `develop` currently

indigo423:
> I think what you do here is not as it is intended:
> `$ echo java-1 | grep "java-[8|11]"`
> `java-1`
> `$ echo java-18 | grep "java-[8|11]"`
> `java-18`
> `echo java-8181818 | grep "java-[8|11]"`
> `java-8181818`
> You want something like `grep -E "java-(8|11)"`, right?

yes, you are totally right. Thank you!

PR #2579, opened 2019-07-15
NMS-10721: Route flow messages depending on exporter

fooker:
This allows each sink module to derive a routing key for every message. Messages with the same routing key are then routed to the same consumer.

### DIFF
--- a/core/ipc/sink/camel/client/src/main/java/org/opennms/core/ipc/sink/camel/client/CamelRemoteMessageDispatcherFactory.java
+++ b/core/ipc/sink/camel/client/src/main/java/org/opennms/core/ipc/sink/camel/client/CamelRemoteMessageDispatcherFactory.java
@@ -30,6 +30,7 @@ package org.opennms.core.ipc.sink.camel.client;
 
 import static org.opennms.core.ipc.sink.api.Message.SINK_METRIC_PRODUCER_DOMAIN;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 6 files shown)


core/ipc/sink/camel/client/src/main/java/org/opennms/core/ipc/sink/camel/client/CamelRemoteMessageDispatcherFactory.java
```diff
@@ -78,29 +79,31 @@
                 CamelSinkConstants.JMS_QUEUE_PREFIX, module.getId());
         Map<String, Object> headers = new HashMap<>();
         headers.put(CamelSinkConstants.JMS_QUEUE_NAME_HEADER, queueNameFactory.getName());
-        return headers;
+        return Collections.unmodifiableMap(headers);
```
cgorantla: Since in every dispatch we are adding tracing headers, we can't have `UnmodifiableMap` here.   I had to change this when I was doing tracing related changes.

cgorantla: Looks like you are creating new HashMap below

j-white:
I'm guessing this is only a partial solution to NMS-10721? 😄

core/ipc/sink/camel/client/src/main/java/org/opennms/core/ipc/sink/camel/client/CamelRemoteMessageDispatcherFactory.java
fooker: Yeah, there have been some conditional adds to the list which result in leaving the provious value in the list. Therefore I wanted to ensure that the list is immutable

fooker:
@jwhite it is. But somebody told me that this is needed for other stuff, too. So I want to get the PR out as soon as possible

cgorantla:
We could change the issue to https://issues.opennms.org/browse/NMS-12142 which just captures what is getting fixed in this PR.

PR #2582, opened 2019-07-16
Flip flop

j-white:
These changes help address a number of smoke tests flapping on both Bamboo & CircleCI.

I can create issues for any of the changes if we think it's worthy to include them in the release notes.

### DIFF
--- a/container/features/src/main/resources/features.xml
+++ b/container/features/src/main/resources/features.xml
@@ -1416,7 +1416,7 @@
     </feature>
     <feature name="opennms-jest" version="${project.version}" description="OpenNMS :: Features :: Jest :: Feature definition">
         <details>Feature definition for Jest (ElasticSearch Java ReST Client)</details>
-        <bundle>wrap:mvn:com.google.code.gson/gson/${jestGsonVersion}</bundle>
+        <bundle>mvn:com.google.code.gson/gson/${jestGsonVersion}</bundle>
         <bundle>wrap:mvn:org.apache.httpcomponents/httpcore-nio/4.4.6</bundle>
         <bundle>mvn:org.opennms.features.jest/org.opennms.features.jest.client/${project.version}</bundle>
         <bundle>mvn:io.searchbox/jest-complete-osgi/${jestVersion}</bundle>
--- a/container/shared/src/main/filtered-resources/etc/org.apache.karaf.features.cfg
+++ b/container/shared/src/main/filtered-resources/etc/org.apache.karaf.features.cfg
@@ -1,0 +1,93 @@
+################################################################################
+#
+#    Licensed to the Apache Software Foundation (ASF) under one or more
+#    contributor license agreements.  See the NOTICE file distributed with
+#    this work for additional information regarding copyright ownership.
+#    The ASF licenses this file to You under the Apache License, Version 2.0
+#    (the "License"); you may not use this file except in compliance with
+#    the License.  You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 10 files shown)


container/shared/src/main/filtered-resources/etc/org.apache.karaf.features.cfg
```diff
@@ -0,0 +1,93 @@
...
+    mvn:org.apache.karaf.features/spring/${karafVersion}/xml/features, \
+    mvn:org.opennms.karaf/opennms/${project.version}/xml/karaf-extensions, \
+    mvn:org.apache.karaf.features/standard/${karafVersion}/xml/features, \
+    mvn:org.apache.karaf.features/spring-legacy/${karafVersion}/xml/features, \
+    mvn:io.hawt/hawtio-karaf/2.0.0/xml/features
+
+#
+# Comma separated list of features to install at startup
+#
+featuresBoot = \
```
j-white: Define this file so that we can control the order of the features listed in `featuresBoot`. Don't rely on the file generated by the `karaf-maven-plugin` in https://github.com/OpenNMS/opennms/blob/develop/container/shared/pom.xml#L70 anymore.

container/features/src/main/resources/features.xml
```diff
@@ -1416,7 +1416,7 @@
     </feature>
     <feature name="opennms-jest" version="${project.version}" description="OpenNMS :: Features :: Jest :: Feature definition">
         <details>Feature definition for Jest (ElasticSearch Java ReST Client)</details>
-        <bundle>wrap:mvn:com.google.code.gson/gson/${jestGsonVersion}</bundle>
+        <bundle>mvn:com.google.code.gson/gson/${jestGsonVersion}</bundle>
```
j-white: Was encountering class-loading issues when using gson, so I bumped it to a newer version that includes the OSGi meta-data to make sure that wasn't the problem.

opennms-webapp/src/main/webapp/report/index.jsp
```diff
@@ -149,6 +149,8 @@
 </div>
   <hr />
 
-<jsp:include page="/js/angular.min.jsp" flush="false" />
-<jsp:include page="/js/search.min.jsp" flush="false" />
+<jsp:include page="/assets/load-assets.jsp" flush="false">
```
j-white: Use a valid `.jsp`. The other `.jsp`s included here don't actually exist. This caused to fail the smoke tests that used to page to fail *sometimes*.

smoke-test/src/main/java/org/opennms/smoketest/containers/OpenNMSContainer.java
```diff
@@ -366,7 +366,7 @@ public WaitForOpenNMS(OpenNMSContainer container) {
         protected void waitUntilReady() {
             LOG.info("Waiting for startup to begin.");
             final Path managerLog = Paths.get("/opt", "opennms", "logs", "manager.log");
-            await().atMost(2, MINUTES).ignoreExceptions()
+            await().atMost(3, MINUTES).ignoreExceptions()
```
j-white: Increased this for better reliability locally.

smoke-test/src/test/java/org/opennms/smoketest/AssetsPageForNodeIT.java
```diff
@@ -98,16 +96,13 @@ public void tearDown() throws Exception {
...
         findElementByLink("Asset Info").click();
         Alert alert = driver.switchTo().alert();
         alert.accept();
 
         final String descriptionXpath = "(//input[@ng-model='asset[field.model]'])[1]";
-        waitForElement(By.xpath(descriptionXpath));
-        WebElement description = findElementByXpath(descriptionXpath);
-        Assert.assertNotNull(description);
-        Assert.assertEquals("Right here, Right now", description.getAttribute("value"));
+        waitForValue(By.xpath(descriptionXpath), "Right here, Right now");
```
j-white: Use the helper function instead, which was made much more resilient.

mvrueden:
As various branches fail and this will most likely fix those failing builds, I am going ahead and merge this.

PR #2583, opened 2019-07-16
NMS-12164: Fix minion label handling

mvrueden:
Here we fix minion label handling.

This probably never worked.
So it might be worth backporting to `foundation-2018` or even `foundation-2017`.

JIRA: https://issues.opennms.org/browse/NMS-12164


indigo423:
Tested the feature with the following:

* Created a Minion label
* Changed an existing label

Just a side note and is no blocker for merging.

While using I've missed a tooltip on the "x" icon, cause it was not obvious to me it cancels when you edit a label. My first impression was "deleting" the label as an action.

Screenshot 2019-07-17 at 17 40 50

indigo423:
@mvrueden you should add a reviewer for code :)

RangerRick:
Yeah, maybe change it to an undo kind of thing instead: 

(assuming fontawesome or ionicons has something we already include)

RangerRick:
Otherwise, LGTM.  Also agree this should probably go to F2018 at least.

mvrueden:
Will leave it as is for now as the `X` icon is consistent with other elements on the UI, even if it may not be the ideal choice. See adding/editing/removing properties.

PR #2584, opened 2019-07-16
Unify Database Report Actions

mvrueden:
Here we merge all of the database report actions into one single page.

I did not create an issue as we have to iterate over this a bunch more times and it is tedious to have an issue everytime stating the same. We can however assign these commits to issue https://issues.opennms.org/browse/NMS-11723 or create a consolidating new one if required.

I think we have to iterate over this some more, but this is how it looks now:

This is the list of templates
Screenshot 2019-07-16 at 13 44 02

This shows a selected template
Screenshot 2019-07-16 at 13 44 12

This shows a selected template, running in scheduled mode
Screenshot 2019-07-16 at 13 45 11

### DIFF
--- a/core/web-assets/src/main/assets/js/apps/onms-reports/details.html
+++ b/core/web-assets/src/main/assets/js/apps/onms-reports/details.html
@@ -3,7 +3,7 @@
     <span>{{errorMessage}}</span>
 </div>
 
-<form name="reportForm" class="form mx-4" role="form" novalidate>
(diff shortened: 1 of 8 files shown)


indigo423:
reporting-ui-proposal

j-white:
Gave this a whirl and it flows much better than the previous UI. +1

indigo423:
Yeah absolutely, tested it when I made the animated gif and :+1: from me as well. Big improvement than the original UI :)

PR #2586, opened 2019-07-17
NMS-12163: Add SonarQube Code Coverage to CircleCI Workflow

Naicisum:
NMS-12163: Add SonarQube Code Coverage to CircleCI Workflow
NMS-12163: Clarity changes, typo name fix, formatting changes
NMS-12163: Cleanup scripts, add new CircleCI cron workflow
NMS-12163: Rename variable prefix circleci to ci
NMS-12163: Reconfigure retry count on failed tests

This enhancement adds new functionality to the CircleCI Workflow, it uses SonarQube Cloud to provide code coverage and analysis.

- Add CircleCI `integration-tests-with-coverage` job for enabling code coverage modules
- Add ability to easily specify how many times to re-run failed tests
- Add new CircleCI job `code-coverage` for running SonarQube analysis
- Add new workflow `nightly-coverage` to run on CircleCI cron job at 00:00 UTC for `develop` branch

### All Contributors

### Pull Request Process

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12163


.circleci/config.yml
```diff
@@ -184,6 +206,15 @@ commands:
             cp ./failsafe_classname* ~/generated-tests/
             cp ./projects_with_path ~/generated-tests/
             cp /tmp/this_node* ~/generated-tests/
+      - run:
+          name: Compress Target Directories (Code Coverage)
+          when: always
+          command: |
+            .circleci/scripts/cc-save.sh
```
mvrueden: Can we name the file a bit more meaningful, e.g. `codecoverage-save.sh`, or `coverage-save.sh`?

.circleci/config.yml
```diff
@@ -416,7 +450,28 @@ jobs:
...
+    resource_class: medium
+    steps:
+      - attach_workspace:
+          at: ~/
+      - restore-sonar-cache
+      - run:
+          name: Restore Target Directories (Code Coverage)
+          when: always
+          command: |
+            .circleci/scripts/cc-load.sh
```
mvrueden: Can we name the file a bit more meaningful, e.g. `codecoverage-load.sh`, or `coverage-load.sh`?

Also maybe, `restore` is a more suitable suffix here.

.circleci/config.yml
```diff
@@ -103,8 +116,8 @@ commands:
           when: always
           command: |
             mkdir -p ~/test-results/junit
-            find . -type f -regex ".*/target/surefire-reports/.*xml" -exec cp {} ~/test-results/junit/ \;
-            find . -type f -regex ".*/target/failsafe-reports/.*xml" -exec cp {} ~/test-results/junit/ \;
+            find . -type f -regex ".*/target/surefire-reports-[0-9]+/.*xml" -exec cp {} ~/test-results/junit/ \;
+            find . -type f -regex ".*/target/failsafe-reports-[0-9]+/.*xml" -exec cp {} ~/test-results/junit/ \;
```
roskens: Would using
 $ find ... | xargs -r ....
be more performant here?

This is walking the full build tree twice to copy files over.  Could this be condensed into one statement?

roskens:
Regarding circleci.rerunFailingTestsCount:

This corresponds directly to the surefire and failsafe plugin property rerunFailingTestsCount. The JVM properties to set on the command line are surefire.rerunFailingTestsCount and failsafe.rerunFailingTestsCount.

Regarding the code-coverage property, I'd echo Markus's comments earlier about naming the scripts after their function instead of shortening the name (ie: code-coverage-save.sh instead of cc-save.sh)

.circleci/config.yml
RangerRick: Well, you can't use xargs because the output of the find needs to be _before_ the target   path, and it appears support for extended regex and other things are spotty between OSes in  `find`, but you could do something like:

```
find . -type f -regex '.*/target/.*-reports.*\.xml' | grep -E '(surefire|failsafe)-reports' > /tmp/$$.filelist
rsync -v --files-from=/tmp/$$.filelist . ~/test-results/junit/
rm /tmp/$$.filelist
```

.circleci/config.yml
```diff
@@ -160,21 +173,30 @@ commands:
             - project
             - .m2
   run-integration-tests:
+    parameters:
+      run-code-coverage:
+        default: false
+        type: boolean
+      fail-rerun-count:
+        default: 2
```
j-white: We shouldn't change the behavior of re-running failed tests here - we're working on fixing the flappers, re-running them automatically hides real problems.

pom.xml
```diff
@@ -1217,24 +1155,24 @@
...
+    <webstart.keystore>${project.build.directory}/build-generated-keystore.ks</webstart.keystore>
+    <webstart.keypass>opennms</webstart.keypass>
+    <webstart.storepass>opennms</webstart.storepass>
+    <webstart.keyalias>opennms-webstart-key</webstart.keyalias>
+    <webstart.keystore.delete>true</webstart.keystore.delete>
+    <webstart.keygen>true</webstart.keygen>
+    <webstart.dnameCn>${user.name}</webstart.dnameCn>
+    <webstart.dnameL/>
+    <webstart.dnameSt/>
+    <webstart.dnameC/>
```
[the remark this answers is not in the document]
Naicisum: This is my OCD, this actually corrects the alignment to match the rest of the pom

.circleci/config.yml
Naicisum: Considering this step take under 20 seconds on average I think keeping it simple like this is acceptable.

.circleci/config.yml
Naicisum: Done

.circleci/config.yml
Naicisum: Done

.circleci/scripts/itest.sh
```diff
@@ -74,8 +74,11 @@ mvn verify -P'!checkstyle' \
            -Dbuild.skip.tarball=true \
            -DfailIfNoTests=false \
            -DskipITs=false \
+           -Dcircleci.instance=$CIRCLE_NODE_INDEX \
+           -Dcircleci.rerunFailingTestsCount=${CCI_RERUN_FAILTEST:-1} \
+           -Dcode-coverage=${CCI_CODE_COVERAGE:-false} \
```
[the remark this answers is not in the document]
Naicisum: Done

.circleci/config.yml
Naicisum: Done, changed defaults to 0

Naicisum:
> Regarding circleci.rerunFailingTestsCount:
> 
> This corresponds directly to the surefire and failsafe plugin property rerunFailingTestsCount. The JVM properties to set on the command line are surefire.rerunFailingTestsCount and failsafe.rerunFailingTestsCount.

I found that this was being overridden by the pom configuration so I created these settings and adjusted the pom so any value set would be obeyed.

> Regarding the code-coverage property, I'd echo Markus's comments earlier about naming the scripts after their function instead of shortening the name (ie: code-coverage-save.sh instead of cc-save.sh)

This was done

Naicisum:
Well, changing the retry failing tests to 0 definitely increased the number of failures :)

j-white:
> Thoughts?

We want to job to fail fast if something is wrong.

Looks like enabling jacoco changes the test timing significantly in a way that causes a lot of tests to start failing.

Maybe we should just run the code coverage job nightly on the `develop` branch and ignore test failures for that run.

Naicisum:
> We want to job to fail fast if something is wrong.
Done, there are now two workflows. One based on a cron job that runs at 00:00 UTC and only performs `build`, `integration-tests-with-coverage`, and `code-coverage` and the original job.

> Looks like enabling jacoco changes the test timing significantly in a way that causes a lot of tests to start failing.
I have also changed the workflow to have 0 retries on both trees for full visibility. We might want to have the maven `-fn` run always and then in a next step, scan the results files for failures and if found, then exit with 1 to cause failure. Reasoning is that this would allow all tests to run for full visibility.

> Maybe we should just run the code coverage job nightly on the `develop` branch and ignore test failures for that run.
Done

Naicisum:
@j-white Do we want to create some weekly jobs against the other long trees?
```
master
/^release-.*/
/^foundation.*/
```

pom.xml
```diff
@@ -927,10 +881,8 @@
           </systemPropertyVariables>
           <forkCount>1</forkCount>
           <reuseForks>false</reuseForks>
-          <rerunFailingTestsCount>1</rerunFailingTestsCount>
-          <includes>
-            <include>**/*Test.java</include>
-          </includes>
+          <rerunFailingTestsCount>${circleci.rerunFailingTestsCount}</rerunFailingTestsCount>
```
j-white: Can we change the variable names to just be `ci` instead of `circleci` inside the poms - since they aren't specific to CircleCI.

Naicisum: Done

Naicisum:
@j-white @RangerRick This should be ready to merge.

Be aware that with retries set to 0 (was 1), this will cause a lot of previous test that passed on retry to now fail.

j-white:
Ah, I didn't realize it was actually set to 1 before. It is set to 0 for the smoke/system tests though.

Maybe we should keep this at 1 and tackle that separately then. Changes here look good though 👍

Naicisum:
Created a happy medium, kept at 1 for main workflow and 0 for nightly code coverage. Since nightly workflow never fails out, it should help expose tests that need adjustments.

j-white:
Perfect. To recap:
1. We have a new job that will trigger integration tests with code coverage enabled and upload these to sonarcloud.io nightly.
1. This is only enabled on the `develop` branch currently
1. Test retries remain unchanged

Naicisum:
Yep!

PR #2588, opened 2019-07-17
NMS-12172: Fix default kafka consumer properties for Kafka RPC

cgorantla:
Use default `auto.offset.reset` settings that is `latest`.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12172


cgorantla:
> What's the motivation for removing the auto commit settings too?
Default value for `enable.auto.commit` seems to be true  when there is a `group.id` is specified which we do.
`auto.commit.interval.ms` default value is `5000` instead of `1000` but I think `consumer.poll()` does auto commit if  it fetched some data. So default settings seems to be sufficient.

j-white:
`enable.auto.commit` defaults to `true`, so makes sense to remove it.

Doesn't seem like we have a good reason for changing the value of the `auto.commit.interval.ms` though.

cgorantla:
> `enable.auto.commit` defaults to `true`, so makes sense to remove it.
> 
> Doesn't seem like we have a good reason for changing the value of the `auto.commit.interval.ms` though.

Agree, On thinking more on this, If there are multiple minions and one minion is down but if it already processed some rpc requests the other minion will not process the same rpc request if we commit more often.

Naicisum:
By following the `latest` strategy, in the event that the offset is not recorded, this will cause the consumer thread to jump to the `latest` or newest message.

We should also look at situations where these topic have a long retention time, the last offset was saved and the records still exist, there is a large lag, and the consumer connects. By my understanding the reset would not come into play and the lag would attempt to be procesed

One solution would be to outline/document `rpc-request/rpc-response` recommended/required retention times. And/Or programmatically set these options on the topics from within Core/Minion.

Side Note, interesting read on Offset Management and using external systems for offsets.
https://blog.cloudera.com/blog/2017/06/offset-management-for-apache-kafka-with-apache-spark-streaming/

j-white:
From the core's perspective, if there's a lag on the response topic on reconnect, it will zoom through the backlog since there won't be any outstanding requests for those: https://github.com/OpenNMS/opennms/blob/0fc4e5e0691b9c088f3af33f3d88c247de0e0011/core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcClientFactory.java#L481

From the Minion's perspective, if there's a lag on the request topic on reconnect, again it will just drop requests with already expired TTLs:
https://github.com/OpenNMS/opennms/blob/0fc4e5e0691b9c088f3af33f3d88c247de0e0011/core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcServerManager.java#L236

Given that these will be dropped quickly before moving on to processing live requests having some lag / stale requests/responses in the topics shouldn't add much overhead.

cgorantla:
This seems to be passing in bamboo now.,

cgorantla:
@j-white can we merge this?   Bamboo seems to be passing.

PR #2589, opened 2019-07-17
NMS-12143: Share thresholding state across Sentinels (Phase 1 - Keystore)

mattixtech:
Phase 1 - Adding a key value store backed by Cassandra

This PR introduces a key value store API with an implementation backed by Cassandra. The intent is for this implementation to be used for storing thresholding states on distributed Sentinels in an upcoming PR that will depend on this one. Currently the implementation derives its Cassandra session via newts and thus relies on newts to be installed in order to work.

The API has been made generic to make other KV store implementations easy to add in the future and/or to remove the dependency on newts from the Cassandra implementation (by providing a new Session factory and Schema factory).

There is one part that is a bit of a hack and that is how I am obtaining the SchemaManager from newts and using that to write the kv store table schema. I'm doing this for now to make standing this feature up easier in smoke tests given that newts will be installed. We can change this later if we don't like it or it doesn't end up making anything easier (it won't hurt anything right now).

The changes here define the new Karaf features but do not install anything new by default since there is no consumer of this API yet. That will come in part 2 when thresholding is updated to use this service to store state.

I've tested this out manually on sentinel with the Cassandra backend, you can see that here:
https://recordit.co/dX4FiS2ZHB

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12143


mattixtech:
The failing IT passes locally, might need to rethink how to test that aspect.

features/distributed/kv-store/cassandra/src/test/java/org/opennms/features/distributed/kvstore/cassandra/CassandraKVStoreIT.java
```diff
@@ -0,0 +1,278 @@
...
+
+        for (int i = 0; i < 1000; i++) {
+            String iStr = Integer.toString(i);
+            putFutures.add(kvStore.putAsync(iStr, new CassandraKVStoreIT.State(i, iStr)));
+        }
+
+        // Since we did 1000 puts async they shouldn't be all done immediately but we should be able to wait for them
+        // to finish
+        CompletableFuture[] allPutFutures = putFutures.toArray(new CompletableFuture[0]);
+        assertThat(CompletableFuture.allOf(allPutFutures).isDone(), equalTo(false));
```
j-white: I'd remove this - what *if* they do finish quickly?

features/distributed/kv-store/cassandra/src/main/java/org/opennms/features/distributed/kvstore/cassandra/CassandraKVStore.java
```diff
@@ -0,0 +1,207 @@
...
+        selectStmt = session.prepare(String.format("SELECT %s FROM %s where %s = ?", VALUE_COLUMN, TABLE_NAME,
+                KEY_COLUMN));
+        timestampStmt = session.prepare(String.format("SELECT %s FROM %s where %s = ?", TIMESTAMP_COLUMN, TABLE_NAME,
+                KEY_COLUMN));
+    }
+
+    @Override
+    protected void putSerializedValueWithTimestamp(String key, ByteBuffer serializedValue, long timestamp) {
+        // Cassandra will throw a runtime exception here if the execution fails
+        session.execute(insertStmt.bind(key, serializedValue, new Date(timestamp)));
```
j-white: We should also consider including a TTL and exposing this up through the kvstore API.

i.e. Persist this key, but only keep it for 24 hours unless I write to it again.

features/distributed/kv-store/api/src/main/java/org/opennms/features/distributed/kvstore/api/KeyValueStore.java
```diff
@@ -0,0 +1,98 @@
...
+ * original exception.
+ *
+ * @param <T> the type that will be persisted by this store
+ */
+public interface KeyValueStore<T> {
+    /**
+     * @param context a context used to differentiate between keys with the same name (forms a compound key)
+     * @return the timestamp the value was persisted with
+     */
+    long put(String key, T value, String context);
```
cgorantla: Can we have `key` as generic too.  `Rocksdb` takes (byte[], byte[]) as (key, value).   Having sequential numbers as keys makes it useful as queue.   I may want to use this interface for Offheap feature as well.

mattixtech: Sure, I'll make that update as part of this PR.

mattixtech: Actually going to leave this as is and put that on hold for now in the interest of getting this in with minimal changes. Not sure yet if this interface is the best fit for the roxdb use case.

mattixtech:
@j-white thanks for pointing out the issue with the generic type, fixed in latest commit.

container/features/src/main/resources/features-sentinel.xml
```diff
@@ -205,4 +205,27 @@
...
     </feature>
+
+    <feature name="sentinel-distributed-cassandra-api"
+             description="OpenNMS :: Features :: Distributed :: Cassandra :: API"
+             version="${project.version}">
+        <bundle>mvn:org.opennms.features.distributed/org.opennms.features.distributed.cassandra.api/${project.version}</bundle>
+    </feature>
+    
+    <feature name="sentinel-newts-cassandra"
+             description="OpenNMS :: Features :: Distributed :: Newts Cassandra" version="${project.version}">
```
j-white:
```suggestion
             description="OpenNMS :: Features :: Distributed :: Cassandra :: Newts Impl." version="${project.version}">
```

container/features/src/main/resources/features-sentinel.xml
```diff
@@ -205,4 +205,27 @@
...
+        <bundle>mvn:org.opennms.features.distributed/org.opennms.features.distributed.cassandra.newts-impl/${project.version}</bundle>
+    </feature>
+
+    <feature name="sentinel-kvstore-cassandra"
+             description="OpenNMS :: Features :: Distributed :: Key Value Store :: Cassandra"
+             version="${project.version}">
+        <feature>sentinel-newts-cassandra</feature>
+        <feature>opennms-kvstore-api</feature>        
+        <bundle>mvn:org.opennms.features.distributed/org.opennms.features.distributed.kv-store.cassandra/${project.version}</bundle>
+        <bundle>mvn:org.opennms.features.distributed/org.opennms.features.distributed.kv-store.shell/${project.version}</bundle>
```
j-white: Should we move the `shell` bundle into the `opennms-kvstore-api` feature instead? Seems like it would be useful to have for all impls.

features/distributed/distributed-cassandra/cassandra-api/src/main/java/org/opennms/features/distributed/cassandra/api/CassandraSession.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.concurrent.Future;
+
+import com.datastax.driver.core.PreparedStatement;
+import com.datastax.driver.core.RegularStatement;
+import com.datastax.driver.core.ResultSet;
+import com.datastax.driver.core.ResultSetFuture;
+import com.datastax.driver.core.Statement;
+
+/**
+ * A thin facade on top of Cassandra's {@link com.datastax.driver.core.Session session}.
```
j-white: Is there a reason we don't just use the `Session` interface directly? We're referencing specific types from the `com.datastax.driver.core` package already here so we don't shield much.

features/distributed/kv-store/cassandra/src/main/resources/cql/kv.cql
```diff
@@ -0,0 +1,11 @@
+
+CREATE KEYSPACE $KEYSPACE$
+    WITH replication = {'class': 'SimpleStrategy', 'replication_factor': $REPLICATION_FACTOR$};
+
+CREATE TABLE $KEYSPACE$.kvstore (
+  key varchar,
+  context varchar,
+  value blob,
+  lastUpdated timestamp,
+  PRIMARY KEY ((context, key))
```
j-white: Can we change this to `((context), key)` so that it's possible to enumerate all of the keys within a specific context?

mattixtech: I was considering that. It does allow use to enumerate by context as you mentioned. However the downside is that all keys for a given context will have the same primary Cassandra node instead of being evenly distributed throughout the cluster. For now I will leave it as is since we don't currently have the use case of enumerating a context.

container/features/src/main/resources/features-sentinel.xml
mattixtech: Yes, good point.

features/distributed/distributed-cassandra/cassandra-api/src/main/java/org/opennms/features/distributed/cassandra/api/CassandraSession.java
mattixtech: Will update the class documentation to explain why.

mattixtech:
Smoke test failed on bamboo but passed locally, going to re-run it on bamboo.

PR #2591, opened 2019-07-18
NMS-12175: WsManWQLDetectorFactory missing setBeanProperties

dino2gnt:
* JIRA: https://issues.opennms.org/browse/NMS-12175

### DIFF
--- a/features/wsman/src/main/java/org/opennms/netmgt/provision/detector/wsman/WsManWQLDetector.java
+++ b/features/wsman/src/main/java/org/opennms/netmgt/provision/detector/wsman/WsManWQLDetector.java
@@ -56,7 +56,7 @@ import com.google.common.collect.Lists;
 public class WsManWQLDetector extends SyncAbstractDetector {
     public static final Logger LOG = LoggerFactory.getLogger(WsManWQLDetector.class);
 
-    private static String PROTOCOL_NAME = "WsManWQLService";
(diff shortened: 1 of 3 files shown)


j-white:
@cgorantla can you take a look at this one? Looks like a regression from OIA-2.

cgorantla:
Looks like it is. I missed it for this specific class. I checked again if we are missing it in any other places,   looks like newly created ActiveMQDetector also missed this.

I think if we depend on create detector from GenericServiceDetectorFactory using  `createDetector(Map<String, String> properties)`  that will do setBeanProperties.  

@dino2gnt  Do you mind if I push another commit for ActiveMQDetector here ?

dino2gnt:
> @dino2gnt Do you mind if I push another commit for ActiveMQDetector here ?

Please feel free to do so.

cgorantla:
@j-white Should we target this to  `release-24.1.3` ?

cgorantla:
> @j-white Should we target this to `release-24.1.3` ?

I think we can only target `WsManWQLDetectorFactory` part  towards `release-24.1.3`.

j-white:
> > @j-white Should we target this to `release-24.1.3` ?
> I think we can only target `WsManWQLDetectorFactory` part towards `release-24.1.3`.

Yeah. Given 25 is not too far away I though it would be sufficient to merge this in `develop`, but we can put that in `release-24.1.3` if you see fit.

PR #2596, opened 2019-07-22
NMS-12143: Share thresholding state across Sentinels (Phase 2 state persistence)

mattixtech:
Phase 2 - Implementing state storage in thresholding

This PR expands on #2589 to use the new key value store service to persist thresholding states. Threshold states are simple serializable POJOs that encapsulate all the state of a threshold.

When a threshold is evaluated, it first checks to see if it has the latest copy of its own state and restores that if necessary (using getIfStale() via the key value store). This is to satisfy the case where threshold evaluation for a given threshold occurs on one sentinel and then subsequently on another (potentially due to kafka partition rebalancing). @j-white noted that we might be able to do away with this fetch in the future if we can sequence the input such that it always arrives at the same sentinel.

Then after evaluation, if the value has changed it is persisted to the key value store. The key is persisted with a default TTL of 24hours configurable via system property.

Since this change introduces a new dependency on the key value store to thresholding, I've added a NO-OP implementation that is installed on boot of the OpenNMS karaf container so that existing behavior is maintained in the non-distributed case.

The main focus point of this review is `AbstractThresholdEvaluatorState`, the rest is mostly wiring and minor refactoring.

This PR is currently targeted against #2589 (for a clean diff) and should be changed to target either develop or the drift feature branch after #2589 is merged.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12143


features/collection/thresholding/impl/src/main/java/org/opennms/netmgt/threshd/AbstractThresholdEvaluatorState.java
```diff
@@ -30,28 +30,169 @@
...
+
+        this.thresholdingSession = thresholdingSession;
+        kvStore = new SerializingKeyValueStore<>(thresholdingSession.getKVStore(),
+                AbstractThresholdEvaluatorState::serialize,
+                AbstractThresholdEvaluatorState::deserialize);
+        key = String.format("%d-%s-%s-%s-%s-%s", thresholdingSession.getKey().getNodeId(),
+                thresholdingSession.getKey().getLocation(), thresholdingSession.getKey().getResource(),
+                threshold.getDatasourceExpression(), threshold.getDsType(), threshold.getType());
+
+        String ttlProp = System.getProperty("org.opennms.netmgt.threshd.state_ttl");
```
j-white: Use `org.opennms.core.sysprops.SystemProperties` instead.

features/collection/thresholding/impl/src/main/java/org/opennms/netmgt/threshd/AbstractThresholdEvaluatorState.java
```diff
@@ -151,5 +292,36 @@ protected String formatValue(double value) {
         return valueFormatter.format(value);
     }
 
+    @Override
+    public ThresholdingSession getThresholdingSession() {
+        return thresholdingSession;
+    }
+
+    private static byte[] serialize(Serializable value) {
```
j-white: Just a note. We already have [fst](https://github.com/RuedigerMoeller/fast-serialization) in the class-path which may give us better performance here.

mattixtech:
Smoke test that failed on bamboo passes locally. I'll try re-running it.

mattixtech:
Had to rework things a bit since I was depending on an OSGI reference that would never be satisfied due to startup order. Changed the behaviour to now look for the reference at runtime after the bean is instantiated when a session is requested.

mattixtech:
Retargeted and rebased for develop.

container/features/src/main/resources/features-sentinel.xml
```diff
@@ -224,7 +237,8 @@
              description="OpenNMS :: Features :: Distributed :: Key Value Store :: Cassandra"
              version="${project.version}">
         <feature>sentinel-newts-cassandra</feature>
-        <feature>opennms-kvstore-api</feature>        
+        <feature>opennms-kvstore-shell</feature>
+        <feature>sentinel-kvstore-api</feature><feature>sentinel-kvstore-api</feature>
```
j-white: double double

container/features/src/main/resources/features.xml
```diff
@@ -1440,14 +1440,13 @@
...
-    <feature name="opennms-kvstore-api" description="OpenNMS :: Features :: Distributed :: Key Value Store :: API" version="${project.version}">
-        <bundle>mvn:org.opennms.features.distributed/org.opennms.features.distributed.kv-store.api/${project.version}</bundle>
+    <feature name="opennms-kvstore-shell" description="OpenNMS :: Features :: Distributed :: Key Value Store :: Shell" version="${project.version}">
         <bundle>mvn:org.opennms.features.distributed/org.opennms.features.distributed.kv-store.shell/${project.version}</bundle>
     </feature>
     
     <feature name="opennms-kvstore-noop"
              description="OpenNMS :: Features :: Distributed :: Key Value Store :: No-Op" version="${project.version}">
-        <feature>opennms-kvstore-api</feature>
+        <feature>opennms-kvstore-shell</feature>
```
j-white: We normally depend directly on the api feature and not the shell.

I guess the shell bundle should always be loaded by the `opennms-kvstore-api` feature.

features/collection/thresholding/impl/src/main/java/org/opennms/netmgt/threshd/AbstractThresholdEvaluatorState.java
```diff
@@ -30,28 +30,167 @@
...
+    @SuppressWarnings("unchecked")
+    public AbstractThresholdEvaluatorState(BaseThresholdDefConfigWrapper threshold,
+                                           ThresholdingSession thresholdingSession) {
+        Objects.requireNonNull(threshold);
+        Objects.requireNonNull(thresholdingSession);
+        Objects.requireNonNull(thresholdingSession.getKVStore());
+
+        this.thresholdingSession = thresholdingSession;
+        kvStore = new SerializingKeyValueStore<>(thresholdingSession.getKVStore(),
+                fst::asByteArray,
```
j-white: Nice. Good to see FST back in action.

container/features/src/main/resources/features.xml
mattixtech: I can't depend on the API here since it is being provided by the root class loader on OpenNMS. Only the sentinel version of the kvstore features are bringing along an api bundle since it won't be provided otherwise there.

j-white: Ack. That make sense.

PR #2597, opened 2019-07-23
NMS-12171: Fixed client-side NPE in node-map

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-12171

### DIFF
--- a/smoke-test/src/test/java/org/opennms/smoketest/NodeMapIT.java
+++ b/smoke-test/src/test/java/org/opennms/smoketest/NodeMapIT.java
@@ -1,0 +1,71 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2019 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2019 The OpenNMS Group, Inc.
+ *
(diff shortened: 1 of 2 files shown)


smoke-test/src/test/java/org/opennms/smoketest/NodeMapIT.java
```diff
@@ -0,0 +1,71 @@
...
+ *******************************************************************************/
+
+package org.opennms.smoketest;
+
+import static com.jayway.awaitility.Awaitility.with;
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+import org.junit.Test;
+
+public class NodeMapIT extends OpenNMSSeleniumTestCase {
```
j-white: Is there an existing class we can add this test under instead? (If there isn't it's OK.)

Right now adding a new test class to the smoke tests adds about 4 minutes to every run.

christianpape: There is no node-map related smoke-test yet, but of course this can be added to any other unrelated smoke test since the test method itself isn‘t very complex. What do you think?

j-white: Yeah, I can't see any other good place to put it for now - let's keep it as-is then.

We'll rework these as a separate effort.

PR #2598, opened 2019-07-23
NMS-12182: Collect and graph packet reception rates for telemetryd listeners

j-white:
JIRA: https://issues.opennms.org/browse/NMS-12182

Here we re-work the way that telemetryd expose metrics and make it easier to expose listener/adapter statistics.

We leverage this functionality to expose the rate at which packets are received by the listeners, and update the JMX collection configs to collect and graph these.


cgorantla:
Looks good to me.

PR #2600, opened 2019-07-25
NMS-12154: Provide a Schedule Editor to more easily create scheduled reports

mvrueden:
Here we add a new angular directive to more easily generate scheduled reports without having to type in a cron expression.

Jira: https://issues.opennms.org/browse/NMS-12154

# Known issues

 - [x] This PR also includes parsing a cron expression to populate the `ScheduledOptions` UI object, to only persist the cron expression and detect the UI state from that. However some corner cases are not considered properly and will result in a wrong UI state. For example only 0, 5, 10, 15, 30, 60, 120 and 180 are valid intervals, but we accept all, which means an `0/2` minute expression will result in a daily UI state with an invalid interval :-/
 - [x] Still has some TODOs

## This is how it looks in general
Screenshot 2019-07-25 at 15 09 37

# Daily at 00:00
Screenshot 2019-07-25 at 15 55 44

# Daily every 3 hours in the time range from 1 to 8
Screenshot 2019-07-25 at 15 55 54

# Every Wednesday, Thursday and Saturday at 3:25
Screenshot 2019-07-25 at 15 56 24

# Every 1st Tuesday of the month
Screenshot 2019-07-25 at 15 56 43

## Define a cron expression manually
Screenshot 2019-07-25 at 15 19 52

For now I left the UI in "debug mode" so you see the cron expression which is generated when you make your changes. It will also "print" the populated UI model.
Before merging, I will disable this.
Screenshot 2019-07-25 at 16 22 04


core/web-assets/src/main/assets/js/lib/onms-schedule-editor/templates/schedule-editor.tpl.html
```diff
@@ -0,0 +1,169 @@
...
+            </div>
+
+            <!-- Short selection -->
+            <div class="form-inline my-1 ml-1">
+                <input class="" type="radio" ng-model="options.dayOfMonthToggle" value="dayOfWeek">
+                <span class="mx-1">the</span>
+                <select class="form-control custom-select" ng-model="options.weekOfMonth">
+                    <option value="1">first</option>
+                    <option value="2">second</option>
+                    <option value="3">third</option>
```
RangerRick: This year August has 5 weeks, should there be a `fourth`? I realize it's quite an edge case, but...

mvrueden: Technically you can define a `fourth` or `fifth`.
However if you define, let's say the `fifth Thursday` and there is no `fifth Thursday` in that month, the report is not delivered.
From my point of view the `Schedule Editor` should make creating a schedule very easy.
If you can configure an edge case it may not be very trivial what actually happens behind the scene.
Therefore I decided to drop the `fourth` and `fifth`, just use `Last` in that scenario.
If required I can add a fourth, but would leave the fifth. 
Besides that you can always use `Cron Expression` and define everything in more detail.

> Note that if you specify “#5” and there is not 5 of the given day-of-week in the month, then no firing will occur that month.
Source: http://www.quartz-scheduler.org/documentation/quartz-2.3.0/tutorials/crontrigger.html

mvrueden:
@RangerRick the `Edit Button` is a place holder, as a schedule cannot be edited at this time. The functionality is just missing (-:

mvrueden:
Addressed reviewer feedback. The edit functionality will be implemented at a later state. See https://issues.opennms.org/browse/NMS-12150 for more details.

indigo423:
I've noticed the following problem when you click the following parameters:

Screenshot 2019-07-29 at 14 42 23

Screenshot 2019-07-29 at 14 42 33

It is not possible to schedule the report with the following error message:

> The report could not be scheduled.
> 
> **Reason:** An error occurred parsing the cron expression. It was not possible to schedule the report
> 

The seems to be the cron expression `0 0 1-14 * * TUE, THU`. As @mvrueden figured out we have to convert it to `0 0 1-14 ? * 3,5`.

Additionally: The wording in the error message is a little bit strange. "The report could not be scheduled" and "It was not possible to schedule the report" is duplicated. The last sentence doesn't have a dot :)

mvrueden:
@indigo423 thanks for reporting this. I looked into the report scheduling error. It seems this is a quartz limitation. I will look into this.

About the error message: Those are generated by the backend at the moment and are not optimal, but they are better than nothing. There is an issue to rework that. See https://issues.opennms.org/browse/NMS-12151 for more details.

j-white:
Worked well for me functionally - this is a much easier than manually editing cron expressions.

When scheduling via 'Daily' it would be nice to have the From & To fields span the entire day by default.

mvrueden:
@indigo423 I addressed the issue you mentioned.
@j-white I made the expression span the full day by default
@RangerRick I added the `fourth` as an option.

PR #2602, opened 2019-07-27
NMS-12166: Hide Vaadin Apps headers inside iFrames

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-12166

The code for hiding the header is still in place and works for normal pages. Our Vaadin UIs include the html/script code in a CustomLayout but Vaadin strips script-tags so the JavaScript-stuff is never executed. I added an AttachListener to execute the JavaScript-Code. The Topology UI uses the @PreserveOnRefresh annotation which means that a page reload will not fire detach/attach listeners. So, I added a custom refresh-method that invokes the JS-code manually on page reload.

### DIFF
--- a/features/topology-map/org.opennms.features.topology.api/src/main/java/org/opennms/features/topology/api/HeaderUtil.java
+++ b/features/topology-map/org.opennms.features.topology.api/src/main/java/org/opennms/features/topology/api/HeaderUtil.java
@@ -1,0 +1,48 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2019 The OpenNMS Group, Inc.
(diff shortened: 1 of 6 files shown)


mvrueden:
I created an Ops Board Configuration, with the `Surveillance View` and the `Topology Map`.
Hitting the Preview button hid the Header for the `Surveillance View` but not for the `Topology Map`.
Afterwards going to the Ops Board View, the View does not initialize at all, instead it is in a infinite loading loop. Restarting OpenNMS also didn't work. I tried with latest Chrome.

mvrueden:
It works great 👍 
Actually helps, when building the right branch 
# 🤦‍♂

christianpape:
Still waiting for PR’s approval :-)

PR #2603, opened 2019-07-29
NMS-10657: Added %primaryinterface% to events

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-10657

### DIFF
--- a/features/events/daemon/src/main/java/org/opennms/netmgt/eventd/StandardExpandableParameterResolvers.java
+++ b/features/events/daemon/src/main/java/org/opennms/netmgt/eventd/StandardExpandableParameterResolvers.java
@@ -238,6 +238,28 @@ public enum StandardExpandableParameterResolvers implements ExpandableParameterR
         }
     },
 
+    PRIMARY_INTERFACE {
+
+        @Override
+        public boolean matches(final String parm) {
+            return AbstractEventUtil.TAG_PRIMARY_INTERFACE_ADDRESS.equals(parm);
+        }
+
+        @Override
+        public String getValue(final String parm, final String parsedParm, final Event event, final EventUtil eventUtil) {
+            if (event.getNodeid() != null) {
+                try {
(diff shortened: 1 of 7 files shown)


features/events/daemon/src/main/java/org/opennms/netmgt/eventd/StandardExpandableParameterResolvers.java
```diff
@@ -238,6 +238,28 @@ public String getValue(String parm, String parsedParm, Event event, EventUtil ev
...
+        }
+
+        @Override
+        public String getValue(final String parm, final String parsedParm, final Event event, final EventUtil eventUtil) {
+            if (event.getNodeid() != null) {
+                try {
+                    return eventUtil.getPrimaryInterface(event.getNodeid());
+                } catch (SQLException ex) {
+                    // do nothing
+                    LoggerFactory.getLogger(getClass()).info("primary interface ipaddr unavailable for node with id:", event.getNodeid(), ex);
```
j-white:
```suggestion
                    LOG.info("primary interface ipaddr unavailable for node with id: {}", event.getNodeid(), ex);
```

Missing `{}` placeholder.

Would also be a good opportunity to refactor the existing `LoggerFactory.getLogger(getClass())` statements to use a static logger instead.

j-white:
> (Looks like there's no good place to document to new expansion option.)

Actually, it looks like there is. Replacement tokens are listed here:
http://docs.opennms.org/opennms/releases/24.1.2/guide-admin/guide-admin.html#_the_eventconf_xml_file_and_its_tributaries

indigo423:
@mfuhrmann As you requested, a docker image for your testing is [here](https://hub.docker.com/r/opennms/horizon/tags)

mfuhrmann:
"Faster than the police allows"  (German idiom)
I'm happy you've done that @christianpape

PR #2605, opened 2019-07-29
NMS-11720: Run as non-root

indigo423:
Refactored the container image to run as non-root by default and runs with an arbitrary user id to be compatible to run in Open Shift. The following things are changed on a high level:

* cap_net_raw is set for the Java binary
* support [Arbitrary User IDs](https://docs.openshift.com/enterprise/3.0/creating_images/guidelines.html)
* Unify the directory structure so we don't have different situations when we install from RPM vs. tarball installation

JIRA: https://issues.opennms.org/browse/NMS-11720

As requested, I've uploaded an image to DockerHub with opennms/horizon:nms-11720.

### DIFF
--- a/opennms-container/horizon/Dockerfile
+++ b/opennms-container/horizon/Dockerfile
@@ -40,15 +40,10 @@ ARG ONMS_PACKAGES="opennms-core opennms-webapp-jetty opennms-webapp-remoting ope
 # Allow to install optional packages via YUM
 ARG ADD_YUM_PACKAGES
 
-ARG OPENNMS_UID
-
 COPY ./rpms /tmp/rpms
 COPY ./tarball /tmp/tarball
 
 SHELL ["/bin/bash", "-c"]
-
-RUN groupadd -g 1000 elasticsearch && \
-adduser -u 1000 -g 1000 -d /usr/share/elasticsearch elasticsearch
 
 # Install repositories, system and OpenNMS packages and do some cleanup
 # 1. Try to install from tarball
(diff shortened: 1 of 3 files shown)


indigo423:
@agalue @Naicisum it would be nice to verify if this image can be tested in an Open Shift infrastructure to make sure it works as expected.

agalue:
Here is a screenshot proving it is running:

opennms-in-openshift

```bash
➜  openshift git:(master) ✗ oc project
Using project "opennms" on server "https://192.168.99.140:8443".
➜  openshift git:(master) ✗ oc get pod
NAME                 READY     STATUS    RESTARTS   AGE
onms-0               1/1       Running   0          10m
postgresql-1-q2pvw   1/1       Running   1          23m
```

indigo423:
> I'll defer to @agalue on this one and his tests looks good for me. My OpenShift environment is much more restrictive and would not be as generalized.
> 
> I would however add `umask 002` to the beginning of your `bootstrap.sh` script. It seems to help alleviate some issues inside OpenShift with how some programs create/manipulate files.

@Naicisum I've set umask 002 in our entry point as first instruction. I'll did a test locally and it had no other side effects. Let's see if we pass the smoke tests with it.

indigo423:
* ✅ [Minimal](https://circleci.com/workflow-run/043527e6-1271-4b0a-9648-7c3c3afdc280) smoke tests pass
* ✅ [Full](https://circleci.com/workflow-run/3c8485fe-aabf-416c-a68c-e43aad004709) smoke tests pass

PR #2606, opened 2019-07-29
NMS-12082: Add support for adding prefixes to the index names in Elasticsearch

j-white:
JIRA: https://issues.opennms.org/browse/NMS-12082

Here we update the features that use Elastiscearch to support a new `indexPrefix` property that can be used to prepend arbitrary stings to the currently fixed index names. Using prefixes it now becomes possible for multiple OpenNMS instances to share the same ES cluster and write to different indices.

### DIFF
--- a/features/jest/client/src/main/java/org/opennms/plugins/elasticsearch/rest/index/IndexStrategy.java
+++ b/features/jest/client/src/main/java/org/opennms/plugins/elasticsearch/rest/index/IndexStrategy.java
@@ -62,12 +62,12 @@ public enum IndexStrategy {
                 .withZone(UTC);
     }
 
-    public String getIndex(IndexSettings indexSettings, String indedName, TemporalAccessor temporal) {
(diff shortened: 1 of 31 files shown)


features/jest/client/src/main/java/org/opennms/plugins/elasticsearch/rest/index/IndexStrategy.java
```diff
@@ -58,8 +62,15 @@
                 .withZone(UTC);
     }
 
-    public String getIndex(String indexPrefix, TemporalAccessor temporal) {
-        return String.format("%s-%s", indexPrefix, dateFormat.format(temporal));
+    public String getIndex(IndexSettings indexSettings, String indedName, TemporalAccessor temporal) {
```
RangerRick: `s/indedName/indexName/g`  ;)

j-white:
Merged this right away since we have some other work in the pipeline that's going to cause conflicts with this patch set. If there's additional feedback, feel free to share and we'll look at addressing it while working on the other enhancements (ES 7.x support).

PR #2609, opened 2019-07-30
NMS-12099: Elasticsearch 7.x support

j-white:
JIRA: https://issues.opennms.org/browse/NMS-12099

Here we upgrade the Elasticsearch integration to be compatible with with v7.x.x and drop support for earlier versions.


features/jest/client/src/main/java/org/opennms/plugins/elasticsearch/rest/SearchResultUtils.java
```diff
@@ -0,0 +1,59 @@
...
+
+package org.opennms.plugins.elasticsearch.rest;
+
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+
+import io.searchbox.core.SearchResult;
+
+public class SearchResultUtils {
+    private static final String[] PATH_TO_TOTAL = "hits/total/value".split("/");
```
mvrueden:
```suggestion
    private static final String[] PATH_TO_TOTAL = new String[]{"hits", "total", "value"};
```

PR #2611, opened 2019-07-31
NMS-12176: recursively resolve links in find-java.sh

RangerRick:
### All Contributors

### Pull Request Process

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12176

### DIFF
--- a/opennms-base-assembly/src/main/filtered/bin/find-java.sh
+++ b/opennms-base-assembly/src/main/filtered/bin/find-java.sh
@@ -47,16 +47,6 @@ get_java_version_string() {
 	fi
 }
 
-get_real_path() {
-	file_to_find="$1"
-	readlink="$(command -v readlink 2>/dev/null)"
-	if [ -n "$readlink" ] && [ -L "$file_to_find" ]; then
-		get_real_path "$("$readlink" "$file_to_find" || :)"
-		return
-	fi
-	echo "$file_to_find"
-}
-
 usage() {
 	cat <<END
 usage: $0 [-h] [-v] [minimum_jdk_version] [maximum_jdk_version] [path-to-java]
@@ -160,7 +150,7 @@ main() {
 			[ -n "$DEBUG" ] && (>&2 printf 'Scanning: %s\n' "${dir}")
 			find -L "$dir" -type f -name java 2>/dev/null | grep -E '/bin/java$' | sort -u > /tmp/$$.javabins
 			while read -r javabin; do
-				javabin="$(get_real_path "${javabin}")"
+				javabin="$(__onms_get_real_path "${javabin}")"
 				javahome="$(printf '%s' "${javabin}" | sed -e 's,/bin/java$,,')"
 				version_string="$(get_java_version_string "$javahome")"
 				[ -n "$DEBUG" ] && (>&2 printf '* %s = %s\n' "${javahome}" "${version_string}")
--- a/opennms-base-assembly/src/main/filtered/bin/find-java.sh
+++ b/opennms-base-assembly/src/main/filtered/bin/find-java.sh
@@ -50,12 +50,9 @@ get_java_version_string() {
 get_real_path() {
 	file_to_find="$1"
 	readlink="$(command -v readlink 2>/dev/null)"
-	if [ -n "$readlink" ]; then
-		__new_filename="$("$readlink" "$file_to_find" || :)"
-		if [ -n "$__new_filename" ] && [ -e "$__new_filename" ]; then
-			get_real_path "$__new_filename"
-			return
-		fi
+	if [ -n "$readlink" ] && [ -L "$file_to_find" ]; then
+		get_real_path "$("$readlink" "$file_to_find" || :)"
+		return
 	fi
 	echo "$file_to_find"
 }
--- a/opennms-base-assembly/src/main/filtered/bin/find-java.sh
+++ b/opennms-base-assembly/src/main/filtered/bin/find-java.sh
@@ -50,11 +50,14 @@ get_java_version_string() {
 get_real_path() {
 	file_to_find="$1"
 	readlink="$(command -v readlink 2>/dev/null)"
-	if [ -n "$readlink" ] && [ -n "$file_to_find" ] && [ -e "$file_to_find" ]; then
-		"$readlink" "$file_to_find" || echo "$file_to_find"
-	else
-		echo "$file_to_find"
+	if [ -n "$readlink" ]; then
+		__new_filename="$("$readlink" "$file_to_find" || :)"
+		if [ -n "$__new_filename" ] && [ -e "$__new_filename" ]; then
+			get_real_path "$__new_filename"
+			return
+		fi
 	fi
+	echo "$file_to_find"
 }
 
 usage() {
--- a/opennms-base-assembly/src/main/resources/bin/_lib.sh
+++ b/opennms-base-assembly/src/main/resources/bin/_lib.sh
@@ -47,3 +47,45 @@ __onms_read_conf() {
     rm "${__onms_read_conf_tmp_file}"
   fi
 }
+
+__onms_is_absolute() {
+  case "$1" in
+    [/]*)
+      return 0
+      ;;
+    *)
+      ;;
+  esac
+  return 1
+}
+
+__onms_get_absolute_path() {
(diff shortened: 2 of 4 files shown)


fooker:
Any reason why not using `readlink -f` which resolves recursively and gives you the full path?

opennms-base-assembly/src/main/filtered/bin/find-java.sh
```diff
@@ -50,11 +50,14 @@ get_java_version_string() {
 get_real_path() {
 	file_to_find="$1"
 	readlink="$(command -v readlink 2>/dev/null)"
-	if [ -n "$readlink" ] && [ -n "$file_to_find" ] && [ -e "$file_to_find" ]; then
-		"$readlink" "$file_to_find" || echo "$file_to_find"
-	else
-		echo "$file_to_find"
+	if [ -n "$readlink" ]; then
+		__new_filename="$("$readlink" "$file_to_find" || :)"
+		if [ -n "$__new_filename" ] && [ -e "$__new_filename" ]; then
```
fooker: You can use `-h` or `-L` in tests to determine if the file is a link and use this to end your recursion.

RangerRick: on Mac OS X `readlink` only supports the `-n` option

fooker: I'm not talking about `readlink` but about `test`. Something like
```
if [ -h "$file_to_find" ]; then
    get_real_path "$("$readlink" "$file_to_find")
else
  echo "$file_to_find"
fi
```

RangerRick: oh!  I gotcha, yeah, that makes sense

RangerRick:
Found another bug, `readlink` doesn't print out the full paths properly!  I'mma write some unit tests for this.

opennms-base-assembly/src/main/filtered/bin/find-java.sh
```diff
@@ -50,11 +50,11 @@ get_java_version_string() {
 get_real_path() {
 	file_to_find="$1"
 	readlink="$(command -v readlink 2>/dev/null)"
-	if [ -n "$readlink" ] && [ -n "$file_to_find" ] && [ -e "$file_to_find" ]; then
-		"$readlink" "$file_to_find" || echo "$file_to_find"
-	else
-		echo "$file_to_find"
+	if [ -n "$readlink" ] && [ -L "$file_to_find" ]; then
+		get_real_path "$("$readlink" "$file_to_find" || :)"
```
mattixtech: Is the no-op necessary here? We aren't doing anything with the return code.
```suggestion
		get_real_path "$("$readlink" "$file_to_find")"
```

opennms-base-assembly/src/main/filtered/bin/find-java.sh
```diff
@@ -50,11 +50,11 @@ get_java_version_string() {
 get_real_path() {
 	file_to_find="$1"
 	readlink="$(command -v readlink 2>/dev/null)"
```
mattixtech: Could move this out of this function since we don't need to recursively set this

opennms-base-assembly/src/main/filtered/bin/find-java.sh
RangerRick: true, previously I was detecting if it returns nothing, but now since it's guarded by `-L`, failure should actually shortcut out as an error

RangerRick:
OK, I think this should all be good now.  I wrote proper tests and such.

RangerRick:
It also got a lot more complicated... :D

opennms-base-assembly/src/main/resources/bin/_lib.sh
```diff
@@ -47,3 +47,45 @@ __onms_read_conf() {
...
+    *)
+      ;;
+  esac
+  return 1
+}
+
+__onms_get_absolute_path() {
+  __abspath_dir="$(dirname "$1")"
+  __basename="$(basename "$1")"
+  pushd "$__abspath_dir" >/dev/null 2>&1 || exit 1
```
mattixtech: Couldn't we just do the following:
`echo "${__abspath_dir}/${__basename}"`

opennms-base-assembly/src/main/resources/bin/_lib.sh
```diff
@@ -47,3 +47,45 @@ __onms_read_conf() {
...
+__onms_bin_readlink="$(command -v readlink 2>/dev/null)"
+__onms_bin_realpath="$(command -v realpath 2>/dev/null)"
+
+__onms_get_real_path() {
+  if [ -n "$__onms_bin_realpath" ]; then
+    "$__onms_bin_realpath" "$1"
+    return
+  fi
+
+	file_to_find="$(__onms_get_absolute_path "$1")"
```
mattixtech: Nit: the recursive logic here should be in a separate function, the above realpath evaluation only needs to happen once.

opennms-base-assembly/src/main/resources/bin/_lib.sh
```diff
@@ -47,3 +47,45 @@ __onms_read_conf() {
...
+	if [ -n "$__onms_bin_readlink" ] && [ -L "$file_to_find" ]; then
```
mattixtech: Nit: the `[ -n "$__onms_bin_readlink" ]` could also be moved to the non-recursive part.

opennms-base-assembly/src/main/resources/bin/_lib.sh
```diff
@@ -47,3 +47,45 @@ __onms_read_conf() {
...
+  __basename="$(basename "$1")"
+  pushd "$__abspath_dir" >/dev/null 2>&1 || exit 1
+    echo "${PWD}/${__basename}"
+  popd >/dev/null 2>&1 || exit 1
+}
+
+__onms_bin_readlink="$(command -v readlink 2>/dev/null)"
+__onms_bin_realpath="$(command -v realpath 2>/dev/null)"
+
+__onms_get_real_path() {
```
mattixtech: Nit: The indentation for some of the below is a bit wonky.

opennms-base-assembly/src/main/resources/bin/_lib.sh
RangerRick: If you call `__onms_get_absolute_path foo` that will return `.` as dirname, and `foo` as basename.  So just echoing would print `./foo` rather than getting the full path of `.`.  This is a trick to ensure relative paths get turned into absolute.

opennms-base-assembly/src/main/resources/bin/_lib.sh
RangerRick: I'm not sure how you would do this, you mean outside `__onms_get_real_path()` ?  Or do you just mean check it right after the `[ -n "$__onms_bin_realpath" ]` bit?  Either way it's not really any different AFAICT...

RangerRick:
Oh, I think I get what you're getting at now.  You mean make a separate function for resolving links, and then `__onms_get_real_path()` should only be a wrapper for either `realpath` or `recurse-links`?

opennms-base-assembly/src/main/resources/bin/_lib.sh
```diff
@@ -70,22 +71,41 @@ __onms_get_absolute_path() {
...
+
+	if [ -n "$__onms_bin_realpath" ]; then
+		"$__onms_bin_realpath" "$1"
 		return
 	fi
+
+	file_to_find="$(__onms_get_absolute_path "$1")"
+	if [ -n "$__onms_bin_readlink" ]; then
+		__lib_resolve_symbolic_links "$file_to_find"
+    return
```
mattixtech: nit: indentation

mattixtech:
> Oh, I think I get what you're getting at now. You mean make a separate function for resolving links, and then `__onms_get_real_path()` should only be a wrapper for either `realpath` or `recurse-links`?

Yep, what you updated it to is what I had in mind.

opennms-base-assembly/src/main/resources/bin/_lib.sh
RangerRick: dangit! vscode is not indenting nicely, I should just do it in `vim`  :D

PR #2614, opened 2019-07-31
NMS-1240: Store and read thresholding configuration from PostgreSQL (PART 1)

mattixtech:
This PR adds a JSON document key value store service. The store is available on OpenNMS and Sentinel.

This store will be used in the future to persist configuration in JSON format. Currently there are no users, however in NMS-1240 part 2 I will be adding a user that persists thresholding configuration via this store.

The behaviour as of this PR will be that on OpenNMS a Postgres impl of JSONStore is installed and the `jsonstore` shell command is installed. On sentinel neither are installed by default.

A lot of the churn in this PR is simply splitting the previous key value store up to make it more generic with new implementations for Blobs and JSON documents. The main area to focus on for this review should be the `AbstractPostgresKeyValueStore`.

I chose JDBC for the Postgres impl over hibernate for two reasons. 1 - There is not really any object model for the persisted entity that would leverage hibernate. 2 - Straight JDBC will allow us to take advantage of the `jsonb` data type we are using in Postgres to query specific fields in the JSON document for instance.

This has been tested with the integration test as well as new smoke tests for coverage on Sentinel and OpenNMS. Also tested on local OpenNMS instance.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12140


features/distributed/distributed-postgres/api/src/main/java/org/opennms/features/distributed/postgres/api/PostgresConnectionFactory.java
```diff
@@ -0,0 +1,36 @@
...
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.features.distributed.postgres.api;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+
+public interface PostgresConnectionFactory {
```
j-white: Can we just use the `javax.sql.DataSource` interface directly? `javax.sql.DataSource` is already available in the OSGi registry on Sentinel and we can easily retrieve this from Spring in OpenNMS too.

features/distributed/kv-store/blob/blob-shell/src/main/java/org/opennms/features/distributed/kvstore/blob/shell/BlobStoreGet.java
```diff
@@ -36,13 +36,13 @@
 import org.apache.karaf.shell.api.action.Command;
 import org.apache.karaf.shell.api.action.lifecycle.Reference;
 import org.apache.karaf.shell.api.action.lifecycle.Service;
-import org.opennms.features.distributed.kvstore.api.KeyValueStore;
+import org.opennms.features.distributed.kvstore.api.BlobStore;
 
-@Command(scope = "kvstore", name = "get", description = "Get a record from the key value store")
+@Command(scope = "blobstore", name = "get", description = "Get a record from the blob store decoded as a string")
```
j-white: Can we prefix the shell commands with `opennms-`? (See [NMS-10413](https://issues.opennms.org/browse/NMS-10413))

features/distributed/kv-store/postgres-shared/src/main/java/org/opennms/features/distributed/kvstore/pgshared/AbstractPostgresKeyValueStore.java
```diff
@@ -0,0 +1,244 @@
...
+        try {
+            Connection connection = postgresConnectionFactory.getConnection();
+
+            // Prepare the statements now that we have the table name
+            selectStatement = connection.prepareStatement(String.format("SELECT %s, %s FROM %s WHERE %s = ? AND %s = ?",
+                    VALUE_COLUMN, EXPIRES_AT_COLUMN, getTableName(), KEY_COLUMN, CONTEXT_COLUMN));
+            upsertStatement = connection.prepareStatement(String.format(
+                    "INSERT INTO %s (%s, %s, %s, %s, %s) VALUES (?, ?, ?, ?, " + getValueStatementPlaceholder() + ") " +
+                            "ON CONFLICT ON CONSTRAINT " +
+                            "pk_kvstore_jsonb" +
```
j-white: Shouldn't the constraint depend on the impl.?

core/schema/src/main/liquibase/25.0.0/changelog.xml
```diff
@@ -0,0 +1,32 @@
...
+            <column name="context" type="text">
+                <constraints nullable="false"/>
+            </column>
+            <column name="last_updated" type="timestamp">
+                <constraints nullable="false"/>
+            </column>
+            <column name="expires_at" type="timestamp">
+                <constraints nullable="true"/>
+            </column>
+            <column name="value" type="jsonb">
```
j-white: What version of PostgreSQL was the `jsonb` type added in? Just wondering if this will change our PostgreSQL version requirements in any way.

features/distributed/kv-store/postgres-shared/src/main/java/org/opennms/features/distributed/kvstore/pgshared/AbstractPostgresKeyValueStore.java
mattixtech: Yep, good catch.

core/schema/src/main/liquibase/25.0.0/changelog.xml
mattixtech: As we discussed in scrum, sounds like we will be requiring 10+ going forward after 25.0. Looks like jsonb was introduced in 9.4.

features/distributed/distributed-postgres/api/src/main/java/org/opennms/features/distributed/postgres/api/PostgresConnectionFactory.java
mattixtech: Yep, that simplifies it quite a bit, will do.

features/distributed/kv-store/postgres-shared/src/main/java/org/opennms/features/distributed/kvstore/pgshared/AbstractPostgresKeyValueStore.java
```diff
@@ -0,0 +1,250 @@
...
+
+    private final PreparedStatement selectStatement;
+    private final PreparedStatement upsertStatement;
+    private final PreparedStatement lastUpdatedStatement;
+
+    public AbstractPostgresKeyValueStore(DataSource dataSource) {
+        Objects.requireNonNull(dataSource);
+
+        try {
+            Connection connection = dataSource.getConnection();
```
j-white: This connection should eventually be closed. There's also a chance of it going "stale", so we'll need to do some better connection management here.

features/distributed/kv-store/postgres-shared/src/main/java/org/opennms/features/distributed/kvstore/pgshared/AbstractPostgresKeyValueStore.java
```diff
@@ -0,0 +1,250 @@
...
+    @Override
+    public long put(String key, T value, String context, Integer ttlInSeconds) {
+        Objects.requireNonNull(key);
+        Objects.requireNonNull(value);
+
+        long now = System.currentTimeMillis();
+
+        try {
+            // The below sets the prepared values for both the INSERT and UPDATE cases hence some values being repeated
+            upsertStatement.setString(1, key);
```
j-white: Are these prepared statements and the connection thread safe?

mattixtech: No, good catch. I sent them back to thread safety training, think they should be good now.

container/features/src/main/resources/features-sentinel.xml
```diff
@@ -235,12 +235,27 @@
...
              version="${project.version}">
         <feature>sentinel-newts-cassandra</feature>
-        <feature>opennms-kvstore-shell</feature>
+        <feature>opennms-blobstore-shell</feature>
         <feature>sentinel-kvstore-api</feature>
-        <bundle>mvn:org.opennms.features.distributed/org.opennms.features.distributed.kv-store.cassandra/${project.version}</bundle>
+        <bundle>mvn:org.opennms.features.distributed/org.opennms.features.distributed.kv-store.blob.cassandra/${project.version}</bundle>
+    </feature>
+
+    <feature name="sentinel-kvstore-postgres-shared"
```
j-white: This feature won't resolve correctly if installed alone.

Since it only contains a single bundle, I would just move the bundle directly into the `sentinel-jsonstore-postgres` feature.

features/distributed/kv-store/postgres-shared/src/main/java/org/opennms/features/distributed/kvstore/pgshared/AbstractPostgresKeyValueStore.java
```diff
@@ -0,0 +1,256 @@
...
+    private static final String LAST_UPDATED_COLUMN = "last_updated";
+    private static final String EXPIRES_AT_COLUMN = "expires_at";
+
+    private final DataSource dataSource;
+
+    public AbstractPostgresKeyValueStore(DataSource dataSource) {
+        this.dataSource = Objects.requireNonNull(dataSource);
+    }
+
+    private static boolean expired(ResultSet resultSet) throws SQLException {
```
mvrueden: NIT: Should probably be called `isExpired(..)`.

PR #2619, opened 2019-08-04
Small enhancements for the smoke test api

mvrueden:
I ran into two issues while using the new smoke/system test api:

 1. When running a test locally in Debug Mode, the Container running the WebDriver and Firefox with that may run out of memory as it does not close the firefox application properly. This may not be a problem all the time, but will blow up if you debug locally (which you can prevent by just closing the firefox application manually after each test). However the problem may cause issues with the container when running a parameterised test. With this change after each test the firefox application is closed properly, freeing the occupied memory with it.

2. When running a test in debug mode locally, the downloads folder is not `./target/downloads` but `./smoke-test/target/downloads`. I added a system property to override it, if necessary to help debug tests requiring that property, e.g. the `DatabaseReportIT`.

### DIFF
--- a/smoke-test/src/main/java/org/opennms/smoketest/selenium/AbstractOpenNMSSeleniumHelper.java
+++ b/smoke-test/src/main/java/org/opennms/smoketest/selenium/AbstractOpenNMSSeleniumHelper.java
@@ -206,6 +206,11 @@ public abstract class AbstractOpenNMSSeleniumHelper {
         @Override
         protected void finished(final Description description) {
             cleanUp();
+
+            // Ensure we close the driver, otherwise the browser is still open
+            // which might cause the container to run out of memory when running either
+            // in debug mode or a parameterized test
+            getDriver().quit();
         }
 
         protected void cleanUp() {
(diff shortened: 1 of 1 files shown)


smoke-test/src/main/java/org/opennms/smoketest/selenium/AbstractOpenNMSSeleniumHelper.java
```diff
@@ -206,6 +206,11 @@ protected void failed(final Throwable e, final Description description) {
         @Override
         protected void finished(final Description description) {
             cleanUp();
+
+            // Ensure we close the driver, otherwise the browser is still open
+            // which might cause the container to run out of memory when running either
+            // in debug mode or a parameterized test
+            getDriver().quit();
```
j-white: Looks like this may be causing some failures now:
```
org.openqa.selenium.NoSuchSessionException: 
Session ID is null. Using WebDriver after calling quit()?
```

mvrueden: Yep, I saw this, will take a look. Adding the `triage` label though, as it may take a while.

mvrueden:
Not going to work on this any time soon. Closing.

PR #2620, opened 2019-08-04
Bring back DatabaseReportIT

mvrueden:
Here we bring back the DatabaseReportIT.
The Grafana Reports are not tested at the moment, as we either need to mock the Grafana Endpoint Requests or also start a Grafana Container with pre-configured Dashboards

### DIFF
--- a/smoke-test/src/test/java/org/opennms/smoketest/DatabaseReportIT.java
+++ b/smoke-test/src/test/java/org/opennms/smoketest/DatabaseReportIT.java
@@ -55,13 +55,9 @@ import org.slf4j.LoggerFactory;
  * a downloads directory which the test verifies.
  */
 @RunWith(Parameterized.class)
-public class DatabaseReportIT extends OpenNMSSeleniumDebugIT {
+public class DatabaseReportIT extends OpenNMSSeleniumIT {
 
     private static final Logger LOG = LoggerFactory.getLogger(DatabaseReportIT.class);
-
-    public DatabaseReportIT() {
-        super("http://localhost:32775/wd/hub");
-    }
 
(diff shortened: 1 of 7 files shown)


smoke-test/src/test/java/org/opennms/smoketest/DatabaseReportIT.java
```diff
@@ -65,19 +63,20 @@
...
-                {"ResponseTimeByNode", "PDF", "Response time by node", 1},
-                {"SerialInterfaceUtilizationSummary", "PDF", "Serial Interface Utilization Summary", 2},
-                {"TotalBytesTransferredByInterface", "PDF", "Total Bytes Transferred by Interface ", 2},
-                {"AverageAndPeakTrafficRatesForNodesByInterface", "PDF", "Average and Peak Traffic rates for Nodes by Interface", 2},
-                {"InterfaceAvailabilityReport", "PDF", "Interface Availability Report", 2},
-                {"SnmpInterfaceAvailabilityReport", "PDF", "Snmp Interface Availability Report", 2},
-                {"MaintenanceContractsExpired", "PDF", "Maintenance contracts expired", 2},
-                {"MaintenanceContractsStrategy", "PDF", "Maintenance contracts strategy", 2},
-                {"EventAnalysisReport", "PDF", "Event Analysis report", 2},
+                // TODO MVR what about the grafana reports?
```
j-white: Safe to remove this? Given we've opted not to test these.

mvrueden: Yeah, if we don't want to test them I can remove this.

PR #2623, opened 2019-08-05
NMS-12191: Upgrade Drools from 7.7.0.Final to 7.24.0.Final

j-white:
JIRA: https://issues.opennms.org/browse/NMS-12191

Here we upgrade Drools to the latest stable release.

Protobuf 2 in the root classpath has been upgrade to Protobuf 3 - existing code that we generated for Protobuf 2 continues to work though.

### DIFF
--- a/opennms-correlation/drools-correlation-engine/src/test/opennms-home/etc/drools-engine.d/nodeParentRules/NodeParentRules.drl
+++ b/opennms-correlation/drools-correlation-engine/src/test/opennms-home/etc/drools-engine.d/nodeParentRules/NodeParentRules.drl
@@ -3,6 +3,7 @@ package org.opennms.netmgt.correlation.drools
 import java.util.Date;
 
 import org.opennms.netmgt.correlation.drools.DroolsCorrelationEngine;
+import org.opennms.netmgt.correlation.drools.Cause;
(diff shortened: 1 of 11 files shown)


opennms-correlation/drools-correlation-engine/src/test/opennms-home/etc/drools-engine.d/nodeParentRules/NodeParentRules.drl
```diff
@@ -65,18 +66,10 @@ rule "resolveIssue"
 		retract( $up );
 end
 
-function Cause possibleCause( DroolsCorrelationEngine engine, Long nodeId, Event symptom, Long POLL_INTERVAL ) {
+function Cause createPossibleCause( DroolsCorrelationEngine engine, Long nodeId, Event symptom, Long POLL_INTERVAL ) {
```
j-white: Had to rename this function since the updated Drools compiler would confuse the function named `possibleCause` with the class named `PossibleCause`.

RangerRick: (sigh)

PR #2625, opened 2019-08-05
NMS-12165: Make Meta-Data-DSL work with arbitrary values

fooker:
* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12165

### DIFF
--- a/core/ipc/rpc/utils/src/main/java/org/opennms/core/rpc/utils/mate/Interpolator.java
+++ b/core/ipc/rpc/utils/src/main/java/org/opennms/core/rpc/utils/mate/Interpolator.java
@@ -28,12 +28,20 @@
 
 package org.opennms.core.rpc.utils.mate;
 
-import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
+import java.util.Stack;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import com.google.common.collect.Maps;
+import org.opennms.netmgt.poller.ComplexPollerParameter;
+import org.opennms.netmgt.poller.PollerParameter;
+import org.opennms.netmgt.poller.SimplePollerParameter;
+import org.w3c.dom.Attr;
+import org.w3c.dom.Element;
+import org.w3c.dom.NamedNodeMap;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
 
 public class Interpolator {
     private static final String OUTER_REGEXP = "\\$\\{(.+?:.+?)\\}";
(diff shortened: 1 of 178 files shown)


core/ipc/rpc/utils/src/main/java/org/opennms/core/rpc/utils/mate/Interpolator.java
```diff
@@ -52,10 +54,18 @@ private Interpolator() {}
...
-        if (value instanceof String) {
+        if (value == null) {
+            return null;
+        } else if (value instanceof String) {
             return interpolate((String) value, scope);
         } else {
-            return value;
+            // Serialize, interpolate and deserialize the value
+            final Class<?> clazz = value.getClass();
+            final String xml = JaxbUtils.marshal(value);
```
j-white: Should we try catching any exceptions that may happen during this process and return the original value if something fails, or leave that up to the caller?

(I suspect the marshaling/unmarshaling may break with some types of objects or in certain circumstances.)

core/ipc/rpc/utils/src/test/java/org/opennms/core/rpc/utils/mate/RpcMetaDataUtilsTest.java
```diff
@@ -36,6 +37,11 @@
 import org.junit.Before;
 import org.junit.Test;
 
+import static org.junit.Assert.assertThat;
+import static org.hamcrest.Matchers.*;
```
j-white: Nit: *

core/ipc/rpc/utils/src/test/java/org/opennms/core/rpc/utils/mate/RpcMetaDataUtilsTest.java
```diff
@@ -78,4 +84,33 @@ public void testMetaDataInterpolation() {
...
         Assert.assertEquals("aaa${abc}bbb", interpolatedAttributes.get("attribute10"));
     }
+
+    @Test
+    public void testNonStringMetaDataInterpolation() {
+        final TestObject testObject = new TestObject();
+        testObject.setField("chaos.${ctx1:key1}.example.com");
+
+        final Map<String, Object> attributes = Collections.singletonMap("test-object", testObject);
+        final Map<String, Object> interpolatedAttributes = Interpolator.interpolateObjects(attributes, new MapScope(this.metaData));
```
j-white: Can we add a test that specifically covers the PSM objects?

core/ipc/rpc/utils/src/main/java/org/opennms/core/rpc/utils/mate/Interpolator.java
fooker: God idea. I'm not sure which exceptions to catch? But `Throwable` should do the trick, right?

core/ipc/rpc/utils/src/test/java/org/opennms/core/rpc/utils/mate/RpcMetaDataUtilsTest.java
fooker: jup

fooker:
I've just realized, that this solution does not allow to put expressions into fields which expect to be non-string types (e.g. the port on the PSM config). Therefore I would like to investigate into other solutions...

fooker:
I've just realized, that this solution does not allow to put expressions into fields which expect to be non-string types (e.g. the port on the PSM config). Therefore I would like to investigate into other solutions...

fooker:
Follow up: https://github.com/OpenNMS/opennms/pull/2708

PR #2626, opened 2019-08-05
Fix for sentinel-flows feature not installing

mattixtech:
The sentinel-flows feature depends on `opennms-thresholding-api` which was recently updated and now depends on `opennms-kvstore-api`. However the feature definition was not updated to include that dependency. This PR updates that feature definition to include the new dependency. The `sentinel-flows` feature now installs correctly.

Test manually in sentinel container.

- No Jira


j-white:
Looks good.

We need better ways to test our Karaf features to help prevent issues like these.

PR #2635, opened 2019-08-07
NMS-11721: Proposal to replace custom bash templating with confd

indigo423:
Introduce confd and replace our custom bash templating with confd. The logic to deal with templates is delegated to confd so it can be a) easier maintained and b) gives users the ability to leverage from confd features. The environment variables have been unchanged, so we provide compatibility with old container setups.

### Reviewer Hint
To make it more obvious which property files are generated by environment variables, they get prefixed with _confd.<property-filename>. Additionally, @agalue suggested a [few default settings](https://github.com/OpenNMS/opennms-drift-kubernetes/blob/master/docker-images/opennms/Dockerfile#L7) which are incorporated as default values in the templates when they are not set via ENVs. The dedicated startup command for Cassandra with `-c` is removed and the time series strategy variable Newts is used instead.

* JIRA: https://issues.opennms.org/browse/NMS-11721


opennms-container/horizon/container-fs/confd/templates/newts.properties.tpl
```diff
@@ -0,0 +1,8 @@
+#
+# DON'T EDIT THIS FILE :: GENERATED WITH CONFD
+#
+org.opennms.newts.config.hostname={{getv "/opennms/cassandra/hostnames" "localhost"}}
+org.opennms.newts.config.keyspace={{getv "/opennms/cassandra/keyspace" "newts"}}
```
agalue: I believe having Newts as the default is not good, especially because that is not OpenNMS's default. Why not using what we have (including the confd syntax) for all "conditional" stuff from the branch on which we've originally introduced confd?

agalue: Disregard, the actual property to enable newts is on a different confd template. I should read everything first, and then comment ;)

PR #2636, opened 2019-08-07
NMS-12170: Render exceptions in images instead of aborting.

j-white:
JIRA: https://issues.opennms.org/browse/NMS-12170

Render exceptions that may occur while rendering the panels as images instead of aborting the report altogether.

Example PDF is available here:
[error_as_image.pdf](https://github.com/OpenNMS/opennms/files/3477315/error_as_image.pdf)

### DIFF
--- a/integrations/opennms-jasper-extensions/src/main/java/org/opennms/netmgt/jasper/grafana/ExceptionToPngRenderer.java
+++ b/integrations/opennms-jasper-extensions/src/main/java/org/opennms/netmgt/jasper/grafana/ExceptionToPngRenderer.java
@@ -38,6 +38,8 @@ import java.io.IOException;
 
 import javax.imageio.ImageIO;
 
+import org.opennms.core.sysprops.SystemProperties;
+
 import com.google.common.base.Throwables;
 
 /**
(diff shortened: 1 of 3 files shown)


integrations/opennms-jasper-extensions/src/main/java/org/opennms/netmgt/jasper/grafana/ExceptionToPngRenderer.java
```diff
@@ -0,0 +1,132 @@
...
+     * @return a byte array of a .png representing the given exception
+     * @throws IOException if an error occurs rendering the .png
+     */
+    public static byte[] renderExceptionToPng(Exception e) {
+        // Build a string representation of the exception
+        final StringBuilder sb = new StringBuilder();
+        sb.append("Exception occurred: ");
+        final String stack = Throwables.getStackTraceAsString(e) ;
+        // Limit the length of the stack
+        sb.append(getFirstNLines(stack, 5));
```
mvrueden: Is this something we want to make configurable via a system property?

j-white: Good call. Done.

PR #2638, opened 2019-08-08
Database Reports: Cleanup, Refactoring and improved error handling

mvrueden:
Here we cleanup and refactor most of the backend to produce better error messages in the front end. There are still some things left to do, but mostly it should work.

Things left to do related to validation/error handling:
 - Sometimes a modal is still shown in case of an error (in case of 500 error message)
 - Parameter type validation is not implemented, meaning you can enter a text for int parameters


mvrueden:
I am going ahead and will merge this. If you have any comments, feel free to add them and I will work them in a follow up PR/commit.

PR #2640, opened 2019-08-08
Database  Reporting: Bring back ReportsPageIT

mvrueden:
Here we bring back the temporarily disabled `ReportsPageIT`.


mvrueden:
Having the `ReportsPageIT` as its individual test is too expensive in terms of what we gain, as we have to spawn a new container for that test, which makes everything slower. I decided to put the logic into the `MenuHeaderIT` as it is basically doing the same thing.

PR #2641, opened 2019-08-08
Database Reporting: Add pagination

mvrueden:
Here we add pagination for Report Schedules and already persisted Reports.


mvrueden:
I am going ahead and will merge this. As there are still some bits left in the reporting UI, just comment here and I will address them in a separate commit/PR.

PR #2642, opened 2019-08-08
NMS-12212: Show configured rule groups

patrick-schweizer:
* Jira: https://issues.opennms.org/browse/NMS-12212

# Test Instructions:
* Manually add new rule group into database, e.g. `INSERT INTO classification_groups ("id","name", "description", "readonly", "enabled", "priority") VALUES ('5','group5', 'another user defined group', false, true, 40);`
* group should show up in ui
* add rule
* edit rule and select different group: rule should be moved to different group. Positions in both group should have been rearranged

### DIFF
--- a/features/flows/rest/api/src/main/java/org/opennms/netmgt/flows/rest/classification/GroupDTOBuilder.java
+++ b/features/flows/rest/api/src/main/java/org/opennms/netmgt/flows/rest/classification/GroupDTOBuilder.java
@@ -1,0 +1,72 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2018-2018 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2018 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License,
+ * or (at your option) any later version.
+ *
+ * OpenNMS(R) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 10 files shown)


mvrueden:
I tried this and it works. 

Some things I noticed, which probably needs some offline discussions: 

Do we want to allow to create read only groups?
If we do allow, this means we should also allow for editing those (meaning make them editable).
But for the System-defined rules we probably don't want to make them editable.
So probably best to not allow to create read-only groups at all. -> let's talk about this

The group name must be unique. Can you add a test to the rest endpoint verifying that when posting a group with an already existing name, you get a `400 Bad Request: { context: 'name', message: 'A group with name "bla" already exists }`.

The definition of a rule must be unique within a group.
If I were to move a rule to a group where a rule with the same definition already exists, the UI only reports `A rule with this definition already exists`. Maybe just a "move" action is suited better for this use case instead of adding it to the edit dialogue. -> Let's talk about this

When adding a new rule to a group, the drop down allows to change the group while creating it.
Not sure if we want to allow this, as the user already selected the group in order to add a rule. -> Let's talk about this

features/flows/rest/api/src/main/java/org/opennms/netmgt/flows/rest/classification/GroupDTOBuilder.java
```diff
@@ -0,0 +1,72 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2018-2018 The OpenNMS Group, Inc.
```
mvrueden: Should be `2019-2019`.

features/flows/rest/api/src/main/java/org/opennms/netmgt/flows/rest/classification/GroupDTOBuilder.java
```diff
@@ -0,0 +1,72 @@
...
+ * OpenNMS(R) is Copyright (C) 1999-2018 The OpenNMS Group, Inc.
```
mvrueden: Should be `1999-2019`.

features/flows/rest/api/src/main/java/org/opennms/netmgt/flows/rest/classification/GroupDTOBuilder.java
patrick-schweizer: fixed

j-white:
This was closed because the `features/drift2` branch was deleted. Please re-open and retarget to `release-25.0.0`.

patrick-schweizer:
replaced by: https://github.com/OpenNMS/opennms/pull/2661

PR #2643, opened 2019-08-08
NMS-12157: Update the Netty resolver to use a cache based on Caffeine

j-white:
JIRA: https://issues.opennms.org/browse/NMS-12157

Here we update the NettyDnsResolver to use a new custom cache based on Caffeine. This allows us to limit the cache size and expose resolution statistics.

While adding tests for this I found that the previous implementation didn't actually cache any of the reverse lookups that were being performed - only the standard A and AAAA lookups were cached 🤦‍♂

### DIFF
--- a/features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineCache.java
+++ b/features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineCache.java
@@ -1,0 +1,138 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2019 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2019 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License,
+ * or (at your option) any later version.
+ *
+ * OpenNMS(R) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with OpenNMS(R).  If not, see:
+ *      http://www.gnu.org/licenses/
+ *
+ * For more information contact:
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.netmgt.dnsresolver.netty;
+
+import java.util.Collection;
+import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.concurrent.TimeUnit;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.Expiry;
+import com.github.benmanes.caffeine.cache.stats.CacheStats;
+
+/**
+ * Underlying cache used to back the {@link CaffeineDnsCache}.
+ *
+ * Uses the Caffeine caching library as opposed to Netty's io.netty.resolver.dns.Cache implementation so that we can:
+ *  1) Limit the cache's size (and intelligently evict entries)
+ *  2) Expose cache stats
+ *
+ * @author jwhite
+ */
+public abstract class CaffeineCache<E> {
+
+    private final Cache<String, Entries<E>> cache;
+
+    /**
+     * Returns {@code true} if this entry should replace all other entries that are already cached for the hostname.
+     */
+    protected abstract boolean shouldReplaceAll(E entry);
+
+    public CaffeineCache(long maxSize) {
+        final Caffeine<String, Entries> caffeine = Caffeine.newBuilder()
+                .expireAfter(new Expiry<String, Entries>() {
+                    @Override
+                    public long expireAfterCreate(String key, Entries value, long currentTime) {
+                        // Use the TTL of the first entry added for the key as the expiry time
+                        return TimeUnit.SECONDS.toNanos(value.ttl());
+                    }
+
+                    @Override
+                    public long expireAfterUpdate(String key, Entries value, long currentTime, long currentDuration) {
+                        // Don't expire after update
+                        return currentDuration;
+                    }
+
+                    @Override
+                    public long expireAfterRead(String key, Entries value, long currentTime, long currentDuration) {
+                        // Don't expire after read
+                        return currentDuration;
+                    }
+                })
+                .recordStats();
+        if (maxSize > 0) {
+            // Apply a size limit, if set
+            caffeine.maximumSize(maxSize);
+        }
+        cache = caffeine.build();
+    }
+
+    public void clear() {
+        cache.invalidateAll();
+    }
+
+    public void clear(String key) {
+        cache.invalidate(key);
+    }
+
+    public Collection<E> get(String key) {
+        final Entries<E> entries = cache.getIfPresent(key);
+        if (entries == null) {
+            return null;
+        }
+        return entries.get();
+    }
+
+    public void cache(String key, E entry, int ttl) {
+        cache.get(key, k -> new Entries<>(ttl)).add(entry, shouldReplaceAll(entry));
+    }
+
+    public long size() {
+        return cache.estimatedSize();
+    }
+
+    public CacheStats stats() {
+        return cache.stats();
+    }
+
+    private static class Entries<E> {
+        private final int ttl;
+        private final CopyOnWriteArraySet<E> container = new CopyOnWriteArraySet<>();
+
+        public Entries(int ttl) {
+            this.ttl = ttl;
+        }
+
+        public int ttl() {
+            return ttl;
+        }
+
+        public void add(E entry, boolean shouldReplaceAll) {
+            if (shouldReplaceAll) {
+                container.clear();
+            }
+            container.add(entry);
+        }
+
+        public Collection<E> get() {
+            return container;
+        }
+    }
+}
--- a/features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineCache.java
+++ b/features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineCache.java
@@ -31,6 +31,7 @@ package org.opennms.netmgt.dnsresolver.netty;
 import java.util.Collection;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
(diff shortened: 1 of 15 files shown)


j-white:
Here's what the output of the stats look like:
```
admin@opennms> health:metrics-display -m dns                                                                                                                                                                                         [11/1919]
Metric set:                                                                                                                                                                                                                                   
Netty DNS Resolver (Netty DNS resolution metrics)
08/08/19 1:47:07 PM ============================================================
                                                           
-- Gauges ----------------------------------------------------------------------
cacheEvictionCount                    
             value = 0                
cacheMaxSize                          
             value = 10000            
cacheSize
             value = 2
                                                           
-- Meters ----------------------------------------------------------------------
cacheHits
             count = 111
         mean rate = 0.15 events/second
     1-minute rate = 0.40 events/second
     5-minute rate = 0.15 events/second
    15-minute rate = 0.08 events/second
cacheMisses
             count = 10
         mean rate = 0.01 events/second
     1-minute rate = 0.01 events/second
     5-minute rate = 0.01 events/second
    15-minute rate = 0.01 events/second
lookupsFailed
             count = 0
         mean rate = 0.00 events/second
     1-minute rate = 0.00 events/second
     5-minute rate = 0.00 events/second
    15-minute rate = 0.00 events/second
lookupsRejectedByCircuitBreaker
             count = 0
         mean rate = 0.00 events/second
     1-minute rate = 0.00 events/second
     5-minute rate = 0.00 events/second
    15-minute rate = 0.00 events/second
lookupsSuccessful
             count = 121
         mean rate = 0.16 events/second
     1-minute rate = 0.42 events/second
     5-minute rate = 0.16 events/second
    15-minute rate = 0.09 events/second

-- Timers ----------------------------------------------------------------------
lookups
             count = 121
         mean rate = 0.16 calls/second
     1-minute rate = 0.42 calls/second
     5-minute rate = 0.16 calls/second
    15-minute rate = 0.09 calls/second
               min = 0.02 milliseconds
               max = 58.92 milliseconds
              mean = 0.51 milliseconds
            stddev = 1.19 milliseconds
            median = 0.11 milliseconds
              75% <= 0.56 milliseconds
              95% <= 2.88 milliseconds
              98% <= 6.63 milliseconds
              99% <= 6.63 milliseconds
            99.9% <= 6.99 milliseconds
```

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineCache.java
```diff
@@ -0,0 +1,138 @@
...
+
+    private final Cache<String, Entries<E>> cache;
+
+    /**
+     * Returns {@code true} if this entry should replace all other entries that are already cached for the hostname.
+     */
+    protected abstract boolean shouldReplaceAll(E entry);
+
+    public CaffeineCache(long maxSize) {
+        final Caffeine<String, Entries> caffeine = Caffeine.newBuilder()
```
fooker: Shouldn't this be `Caffeine<String, Entries<E>>` ?

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineCache.java
```diff
@@ -0,0 +1,138 @@
...
+                .expireAfter(new Expiry<String, Entries>() {
+                    @Override
+                    public long expireAfterCreate(String key, Entries value, long currentTime) {
+                        // Use the TTL of the first entry added for the key as the expiry time
```
fooker: Using the minimum TTL seems to be more correct here.
This also requires to update the TTL on each `cache(...)` call.

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineCache.java
```diff
@@ -0,0 +1,138 @@
...
+    public Collection<E> get(String key) {
+        final Entries<E> entries = cache.getIfPresent(key);
+        if (entries == null) {
+            return null;
+        }
+        return entries.get();
+    }
+
+    public void cache(String key, E entry, int ttl) {
+        cache.get(key, k -> new Entries<>(ttl)).add(entry, shouldReplaceAll(entry));
```
fooker: If `shouldReplaceAll(...)` returns `true`, this can be simplified to a `cache.put(...)`.

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineCache.java
```diff
@@ -0,0 +1,138 @@
...
+
+        public void add(E entry, boolean shouldReplaceAll) {
+            if (shouldReplaceAll) {
+                container.clear();
+            }
+            container.add(entry);
+        }
+
+        public Collection<E> get() {
+            return container;
```
fooker: We should make this immutable to allow manipulation only via the `CaffeineCache` class.

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineDnsCache.java
```diff
@@ -0,0 +1,370 @@
...
+    /**
+     * Create a cache.
+     * @param minTtl the minimum TTL
+     * @param maxTtl the maximum TTL
+     * @param negativeTtl the TTL for failed queries
+     */
+    public CaffeineDnsCache(int minTtl, int maxTtl, int negativeTtl, long maxSize) {
+        this.minTtl = Math.min(MAX_SUPPORTED_TTL_SECS, checkPositiveOrZero(minTtl, "minTtl"));
+        this.maxTtl = Math.min(MAX_SUPPORTED_TTL_SECS, checkPositiveOrZero(maxTtl, "maxTtl"));
+        if (minTtl > maxTtl) {
```
fooker: This can be simplified by using guavas `Precondition.checkArgument`.

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineDnsCache.java
```diff
@@ -0,0 +1,370 @@
...
+    private static final class DefaultExtendedDnsCacheEntry implements ExtendedDnsCacheEntry {
+        private final String hostname;
+        private final String hostnameFromPtrRecord;
+        private final InetAddress address;
+        private final Throwable cause;
+
+        DefaultExtendedDnsCacheEntry(String hostname, InetAddress address) {
+            this.hostname = hostname;
+            this.address = address;
+            cause = null;
```
fooker: Nit: `this` consistency.

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineDnsCache.java
```diff
@@ -0,0 +1,370 @@
...
+        public int hashCode() {
+            return Objects.hash(hostname, hostnameFromPtrRecord, address, cause);
+        }
+    }
+
+    private static boolean emptyAdditionals(DnsRecord[] additionals) {
+        return additionals == null || additionals.length == 0;
+    }
+
+    private static String appendDot(String hostname) {
```
fooker: Call this `ensureTrailingDot`?

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineDnsCache.java
```diff
@@ -0,0 +1,370 @@
...
+        }
+    }
+
+    public void unregisterMetrics(MetricRegistry metrics) {
+        for (CacheMetric cm : CacheMetric.values()) {
+            metrics.remove(cm.getName());
+        }
+    }
+
+    private static final class DefaultExtendedDnsCacheEntry implements ExtendedDnsCacheEntry {
```
fooker: Maybe this should be split in three implementations. On for `hostnameFromPtrRecord`, `address` and `cause` each.

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/ExtendedDnsCache.java
```diff
@@ -0,0 +1,39 @@
...
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.netmgt.dnsresolver.netty;
+
+import io.netty.channel.EventLoop;
+import io.netty.handler.codec.dns.DnsPtrRecord;
+import io.netty.resolver.dns.DnsCache;
+
+public interface ExtendedDnsCache extends DnsCache {
```
fooker: Call this `ReverseDnsCache` (and rename the entry accordingly)?

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineDnsCache.java
fooker: I'm not sure if it makes sense to mix forward and reverse lookup. Maybe this should be split into two distinct classes / interfaces and use the same generic cache implementation in the background.

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/NettyResolverContext.java
```diff
@@ -145,25 +147,53 @@ public void destroy() {
...
+            // We've got a hit, so we have some cached result for this entry
+            // Try and find a hostname, or return an empty optional if none was found
+            final Optional<String> cachedHostname = entries.stream()
+                    .filter(e -> e instanceof ExtendedDnsCacheEntry)
+                    .map(e -> ((ExtendedDnsCacheEntry)e).hostnameFromPtrRecord())
+                    .filter(Objects::nonNull)
+                    .findFirst();
+            if (cachedHostname.isPresent()) {
+                // We found a cached hostname
+                return CompletableFuture.completedFuture(Optional.of(removeTrailingDot(cachedHostname.get())));
```
fooker: We remove the trailing dot here, just to add it back later on.

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineCache.java
j-white: Done.

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineCache.java
j-white: Good point. Updated.

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineCache.java
j-white: Method was revised while changing the TTL logic. Not sure this has much value in changing anymore.

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineCache.java
j-white: I reduced the visibility of this method. We already convert the collection to a new list before returning the results to the caller in the `CaffeineDnsCache` class, so cache users can't directly manipulate the contents of the container.

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineDnsCache.java
j-white: Done.

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineDnsCache.java
j-white: Done. I tried not using `this.` unless necessary, but it does "look" better when using `this.` consistently 😄

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineDnsCache.java
j-white: Done.

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineDnsCache.java
j-white: I tried splitting this up into different classes, but found that its add a lot of complexity. We then need to introduce something a long the line of a visitor pattern to figure out the type of entries in the container. Can do it, but not sure it's worth it.

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/ExtendedDnsCache.java
j-white: `ReverseDnsCache` seems to imply that it only caches results from reverse lookups.

I added a description on the interface.

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/NettyResolverContext.java
j-white: Where do we add it back on?

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineDnsCache.java
j-white: I think I found a cleaner way to do this while keeping it simple. I'll come back to this.

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineCache.java
fooker: I'm fine with this, too.

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineCache.java
fooker: ok.

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/ExtendedDnsCache.java
fooker: :+1:

features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/NettyResolverContext.java
fooker: Ok, now I get it. We ensure it's there first, then we remove it. Seems fine.

PR #2644, opened 2019-08-09
Remove SNMP proxy hint

indigo423:
The SNMP collector works not just with an SNMP Proxy.

### DIFF
--- a/opennms-doc/guide-admin/src/asciidoc/text/performance-data-collection/collectors/snmp/SnmpCollector.adoc
+++ b/opennms-doc/guide-admin/src/asciidoc/text/performance-data-collection/collectors/snmp/SnmpCollector.adoc
@@ -13,7 +13,7 @@ Access to the _SNMP Agent_ is configured through the _SNMP configuration_ in the
 |===
 | Class Name          | `org.opennms.netmgt.collectd.SnmpCollector`
 | Package             | core
-| Supported on Minion | `Yes` (via the SNMP proxy)
+| Supported on Minion | `Yes`
 |===
 
 ===== Collector Parameters


opennms-doc/guide-admin/src/asciidoc/text/performance-data-collection/collectors/snmp/SnmpCollector.adoc
```diff
@@ -13,7 +13,7 @@ Access to the _SNMP Agent_ is configured through the _SNMP configuration_ in the
 |===
 | Class Name          | `org.opennms.netmgt.collectd.SnmpCollector`
 | Package             | core
-| Supported on Minion | `Yes` (via the SNMP proxy)
+| Supported on Minion | `Yes`
```
j-white: It does use the SNMP proxy, which is different then how other collectors operate.

indigo423: But it is not something like Net-SNMP proxy functionality, right?

j-white: Nope, it's some internal proxy for the RPCs.

What's important to note here is that this collector behaves differently then others - in particular how the TTL is handled.

indigo423: The reason for this change is, user tried to set up SNMP data collection for a Minion and configured the SNMP proxy host using the Minions IP address in "Configure OpenNMS -> Configures SNMP Community Names by IP Address", which is in my point of view a little always, open to any suggestions.

indigo423: Do you think renaming from `(via the SNMP proxy)` to `(via internal RPC proxy)` is better?

j-white: I guess less is more in this case - users don't really need to care *how* it works, just know that it is supported on Minion.

PR #2650, opened 2019-08-14
NMS-12167: Add SNMP Profiles support

cgorantla:
Support SNMP profiles where each profile consists of SNMP configuration with a label and filter expression to match an IpAddress/hostname.
These pre-configured profiles will be matched in Provisiond and saved in definitions in `snmp-config.xml` when there is a match.
Also provides support for  shell command `opennms-snmp:fit` to check and verify a profile match
Also provides support for removing an IpAddress from saved definitions so that it can match a new profile.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12167


pbrane:
I've ran a trial of this and it appears to work as expected.  I tested the provisioning use case as well as the shell command.  Great work and I think the community will really appreciate this feature.  Only think I will say is that the et/examples prototype isn't XML formatted properly but I'm not sure if this is a holdover or introduced in this update.

core/snmp/commands/src/main/java/org/opennms/netmgt/snmp/commands/FitProfileCommand.java
```diff
@@ -0,0 +1,117 @@
...
+    public Object execute() throws Exception {
+
+        InetAddress inetAddress = InetAddress.getByName(ipAddress);
+        CompletableFuture<SnmpAgentConfig> future = fitProfile(label, inetAddress, location, oid);
+        while (true) {
+            try {
+                SnmpAgentConfig agentConfig = future.get(1, TimeUnit.SECONDS);
+                if (agentConfig != null) {
+                    System.out.printf("Fitted IpAddress '%s' with profile '%s', agent config: \n", ipAddress, agentConfig.getProfileLabel());
+                    System.out.println(agentConfig.toString());
```
j-white: Can we display the config the same way as is done in the `ShowConfigCommand`?

Maybe we can refactor this out as `public static void prettyPrint(SnmpAgentConfig agent)`

core/snmp/api/src/main/java/org/opennms/netmgt/snmp/SnmpProfileMapper.java
```diff
@@ -0,0 +1,63 @@
...
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.netmgt.snmp;
+
+import java.net.InetAddress;
+import java.util.Optional;
+
+public interface SnmpProfileMapper {
```
j-white: Can we make these async and wrap the results in a `CompletableFuture`? Callers shouldn't need to tie up a thread given the SNMP requests are already async.

core/snmp/api/src/main/java/org/opennms/netmgt/snmp/SnmpProfileMapper.java
```diff
@@ -0,0 +1,63 @@
...
+     *  Get @{@link SnmpAgentConfig} from Snmp profiles.
+     * @param inetAddress IpAddress for which agent config need to be retrieved.
+     * @param location    location of IpAddress
+     * @return  snmpAgentConfig, a @{@link SnmpAgentConfig} if ipaddress matches the profile.
+     */
+    Optional<SnmpAgentConfig> getAgentConfigFromProfiles(InetAddress inetAddress, String location);
+
+    /**
+     * Fit a profile with specified IpAddress.
+     * @param profileLabel  label of profile with which Snmp profile will be retrieved
```
j-white: The implementation actual has special behavior when `profileLabel` is null or empty, but this is not clearly reflected here.

core/snmp/profile-mapper/src/main/java/org/opennms/core/snmp/profile/mapper/impl/SnmpProfileMapperImpl.java
```diff
@@ -0,0 +1,176 @@
...
+    @Autowired
+    private LocationAwareSnmpClient locationAwareSnmpClient;
+
+    public SnmpProfileMapperImpl(FilterDao filterDao, SnmpAgentConfigFactory agentConfigFactory, LocationAwareSnmpClient locationAwareSnmpClient) {
+        this.agentConfigFactory = Objects.requireNonNull(agentConfigFactory);
+        this.filterDao = Objects.requireNonNull(filterDao);
+        this.locationAwareSnmpClient = Objects.requireNonNull(locationAwareSnmpClient);
+    }
+
+    public SnmpProfileMapperImpl() {
```
j-white: Nit: Do we need to define this?

core/snmp/profile-mapper/src/main/java/org/opennms/core/snmp/profile/mapper/impl/SnmpProfileMapperImpl.java
```diff
@@ -0,0 +1,176 @@
...
+
+    private boolean isFilterExpressionValid(InetAddress inetAddress, String filterExpression) {
+        // Consider profiles without filter expression or an empty one as matching.
+        if (Strings.isNullOrEmpty(filterExpression)) {
+            return true;
+        } else {
+            try {
+                return filterDao.isValid(inetAddress.getHostAddress(), filterExpression);
+            } catch (FilterParseException e) {
+                LOG.warn("Filter expression '{}' is invalid. ", e);
```
j-white: Missing filter expression?

core/snmp/profile-mapper/src/main/java/org/opennms/core/snmp/profile/mapper/impl/SnmpProfileMapperImpl.java
```diff
@@ -0,0 +1,176 @@
...
+
+        List<SnmpProfile> snmpProfiles = agentConfigFactory.getProfiles();
+        // Get matching profiles for this IpAddress.
+        List<SnmpProfile> matchedProfiles = snmpProfiles.stream()
+                .filter(snmpProfile -> isFilterExpressionValid(inetAddress, snmpProfile.getFilterExpression()))
+                .collect(Collectors.toList());
+
+        for (SnmpProfile snmpProfile : matchedProfiles) {
+
+            Optional<SnmpAgentConfig> config = fitProfile(snmpProfile, inetAddress, location, oid);
```
j-white: Profiles should be fitted in parallel instead of serially (take advantage of the fact the SNMP lookups are already async.). When all have returned, we can iterate through the results using the same order as which the profiles are defined in the configuration file and take the first that matched.

core/snmp/profile-mapper/src/main/java/org/opennms/core/snmp/profile/mapper/impl/SnmpProfileMapperImpl.java
```diff
@@ -0,0 +1,176 @@
...
+    public SnmpProfileMapperImpl(FilterDao filterDao, SnmpAgentConfigFactory agentConfigFactory, LocationAwareSnmpClient locationAwareSnmpClient) {
+        this.agentConfigFactory = Objects.requireNonNull(agentConfigFactory);
+        this.filterDao = Objects.requireNonNull(filterDao);
+        this.locationAwareSnmpClient = Objects.requireNonNull(locationAwareSnmpClient);
+    }
+
+    public SnmpProfileMapperImpl() {
+    }
+
+    private static final String SYS_OBJECTID_INSTANCE = ".1.3.6.1.2.1.1.2.0";
```
j-white: Store this locally with type `SnmpObjId` so that the client doesn't need to perform the conversion every call.

core/snmp/profile-mapper/src/main/java/org/opennms/core/snmp/profile/mapper/impl/SnmpProfileMapperImpl.java
cgorantla: yeah, we need this since Spring needs default constructor.

cgorantla:
Thanks @j-white for the review. I addressed the review comments and made scanning with profiles parallel.

core/snmp/profile-mapper/src/main/java/org/opennms/core/snmp/profile/mapper/impl/SnmpProfileMapperImpl.java
```diff
@@ -61,83 +63,104 @@
...
-                return Optional.of(agentConfig);
-            }
-        } catch (Exception e) {
-            LOG.warn("Exception while trying to get sysObjectID for the profile {}", snmpProfile.getLabel(), e);
+        final SnmpAgentConfig agentConfig = agentConfigFactory.getAgentConfigFromProfile(snmpProfile, inetAddress);
+        if (!Strings.isNullOrEmpty(oid)) {
+            snmpObjId = SnmpObjId.get(oid);
+        }
+        if (Strings.isNullOrEmpty(location)) {
+            location = "Default";
```
j-white: Let's use a constant here instead.

core/snmp/profile-mapper/src/main/java/org/opennms/core/snmp/profile/mapper/impl/SnmpProfileMapperImpl.java
```diff
@@ -61,83 +63,104 @@
...
-                    .execute()
-                    .get();
-            if (snmpValue != null && !snmpValue.isError()) {
-                return Optional.of(agentConfig);
-            }
-        } catch (Exception e) {
-            LOG.warn("Exception while trying to get sysObjectID for the profile {}", snmpProfile.getLabel(), e);
+        final SnmpAgentConfig agentConfig = agentConfigFactory.getAgentConfigFromProfile(snmpProfile, inetAddress);
+        if (!Strings.isNullOrEmpty(oid)) {
+            snmpObjId = SnmpObjId.get(oid);
```
j-white: We should use a local variable here.

core/snmp/profile-mapper/src/main/java/org/opennms/core/snmp/profile/mapper/impl/SnmpProfileMapperImpl.java
```diff
@@ -61,83 +63,104 @@
     @Autowired
     private LocationAwareSnmpClient locationAwareSnmpClient;
 
+    private static final String SYS_OBJECTID_INSTANCE = ".1.3.6.1.2.1.1.2.0";
+
+    private SnmpObjId snmpObjId = SnmpObjId.get(SYS_OBJECTID_INSTANCE);
```
j-white: Should be final.

core/snmp/profile-mapper/src/main/java/org/opennms/core/snmp/profile/mapper/impl/SnmpProfileMapperImpl.java
```diff
@@ -61,83 +63,104 @@
...
+                .execute();
+        snmpResult.whenComplete(((snmpValue, throwable) -> {
+            if(throwable == null) {
+                if (snmpValue != null && !snmpValue.isError()) {
+                    future.complete(Optional.of(agentConfig));
+                } else {
+                    future.complete(Optional.empty());
+                }
+            } else {
+                LOG.warn("Exception while doing SNMP get on OID '{}' with profile '{}'", snmpObjId, snmpProfile.getLabel(), throwable);
```
j-white: Can we change this to INFO, and not log the stack?

PR #2652, opened 2019-08-14
NMS-1240: Store and read thresholding configuration from PostgreSQL (PART 2)

mattixtech:
This PR refactors the way thresholding configuration is handled by removing the static factory config access and replacing it with injectable DAOs.

The DAO implementation will depend on the deployment model. On OpenNMS the DAO will read/write from/to the disk and also write to the database via the JsonStore. On Sentinel the DAO will read only from the DB and not allow any write operations.

With this model any OpenNMS and Sentinel containers that are sharing a DB instance will be able to keep their thresholding configuration in sync (within a polling interval).

If Sentinel starts up before the config is ready in the database it will continue to poll for it for a period of 5 minutes before giving up and throwing an exception.

I've tested this manually at runtime and verified that when OpenNMS starts up, it writes the config to the database and that when the config is modified the database is updated. I've also tested that on Sentinel the DAOs can be installed (if a jsonstore impl is provided) and start up and read from the DB.

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12140


mattixtech:
Drafting this pending a clean test run. Somehow I manged to move a bunch of files I didn't mean to move and that was causing test failures.

mattixtech:
Looks like the test failures are unrelated to changes in this PR.

mattixtech:
Rebased to bring in the system test fixes.

opennms-webapp/src/main/webapp/WEB-INF/web.xml
```diff
@@ -54,7 +54,7 @@
 
   <context-param>
     <param-name>parentContextKey</param-name>
-    <param-value>daoContext</param-value>
+    <param-value>restContext</param-value>
```
j-white: Given that these are also used by some of the JSP and not exclusively for REST. Can we rename the context to `webContext`?

opennms-webapp-rest/src/test/java/org/opennms/web/rest/v1/ScheduledOutagesRestServiceIT.java
```diff
@@ -80,7 +80,9 @@
         "classpath:/META-INF/opennms/mockEventIpcManager.xml",
         "file:src/main/webapp/WEB-INF/applicationContext-svclayer.xml",
         "file:src/main/webapp/WEB-INF/applicationContext-cxf-common.xml",
-        "classpath:/applicationContext-rest-test.xml"
+        "classpath:/applicationContext-rest-test.xml",
+        "classpath:/META-INF/opennms/applicationContext-testThresholdingDaos.xml",
```
j-white: Doesn't the rest-test context defined above already include these?

opennms-webapp-rest/src/main/webapp/WEB-INF/web.xml
```diff
@@ -26,7 +26,7 @@
 
   <context-param>
     <param-name>parentContextKey</param-name>
-    <param-value>daoContext</param-value>
+    <param-value>restContext</param-value>
```
j-white: `webContext`? (Since it's also shared by the webapp)

opennms-services/src/test/java/org/opennms/netmgt/statsd/StatsdIT.java
```diff
@@ -48,10 +50,12 @@
         "classpath:/META-INF/opennms/applicationContext-pinger.xml",
         "classpath:/META-INF/opennms/applicationContext-statisticsDaemon.xml",
         "classpath:/META-INF/opennms/applicationContext-commonConfigs.xml",
-        "classpath:/META-INF/opennms/applicationContext-minimal-conf.xml"
+        "classpath:/META-INF/opennms/applicationContext-minimal-conf.xml",
+        "classpath:/META-INF/opennms/applicationContext-testPollerConfigDaos.xml"
 })
 @JUnitConfigurationEnvironment
-//@JUnitTemporaryDatabase
+// TODO: Should this db be used?
```
j-white: TODO

opennms-config-dao/thresholding/impl/src/main/java/org/opennms/netmgt/config/dao/thresholding/impl/AbstractThresholdingDao.java
```diff
@@ -0,0 +1,104 @@
...
+import org.opennms.netmgt.config.dao.thresholding.api.ReadableThresholdingDao;
+import org.opennms.netmgt.config.threshd.Basethresholddef;
+import org.opennms.netmgt.config.threshd.Group;
+import org.opennms.netmgt.config.threshd.ThresholdingConfig;
+
+public abstract class AbstractThresholdingDao implements ReadableThresholdingDao {
+    private final Map<String, Group> groupMap = new HashMap<>();
+    final JsonStore jsonStore;
+    public static final String JSON_STORE_KEY = "thresholds";
+    public static final String JSON_STORE_CONTEXT = "config";
```
j-white: Can we move this constant somewhere that is accessible by all the config impls?

features/distributed/kv-store/api/src/main/java/org/opennms/features/distributed/kvstore/api/AbstractAsyncKeyValueStore.java
```diff
@@ -55,18 +54,13 @@ protected AbstractAsyncKeyValueStore(Executor executor) {
...
-            return t;
-        };
         // A reasonable default executor based on available processors that degrades to synchronous processing when full
         executor = new ThreadPoolExecutor(1,
                 Runtime.getRuntime().availableProcessors() * 10,
                 60,
                 TimeUnit.SECONDS,
                 new SynchronousQueue<>(),
-                threadFactory,
+                r -> new Thread(r, "kvstore-async-thread"),
```
j-white: Would be nice to give the threads unique names. We use Guava's  `ThreadFactoryBuilder` to help with this for other thread pools.

PR #2653, opened 2019-08-15
NMS-12188: Added exception to LOG.error() statement

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-12188


mvrueden:
Is that printing each item which failed and its reason? If not, we should enhance it even further.
Please take a look at the [PersistenceException](https://github.com/OpenNMS/opennms/blob/2f7cf992d5767df80a71539c55273273505216f9/features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/PersistenceException.java#L37) which has each failed item including the cause why it failed.

christianpape:
done

features/telemetry/protocols/flows/pom.xml
```diff
@@ -42,6 +42,17 @@
       <artifactId>org.opennms.features.telemetry.protocols.common</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
```
j-white: This module should only depend on the Flow API (`org.opennms.features.flows.api`) and not have direct knowledge of the implementation.

j-white:
Can we re-target this to `release-25.0.0`.

christianpape:
done

j-white:
LGTM, if @mvrueden approves, then we're good.

PR #2655, opened 2019-08-15
NMS-12224: added replyTo field

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-12224


j-white:
👍 Nice update to the northbounder too.

PR #2663, opened 2019-08-21
NMS-12241: Fix flapping DatabaseReportPageIT

mvrueden:
Here we fix the flapping `DatabaseReportPageIT`.

JIRA: https://issues.opennms.org/browse/NMS-12241


smoke-test/src/test/java/org/opennms/smoketest/DatabaseReportPageIT.java
```diff
@@ -339,7 +343,7 @@ public ReportDetailsForm applyDeliveryOptions(DeliveryOptions options) {
         }
 
         public ReportDetailsForm applyCronExpression(String cronExpression) {
-            LOG.debug("Apply corn expression {}", cronExpression);
+            LOG.debug("Applying cron expression '{}'", cronExpression);
```
j-white: 🌽

PR #2670, opened 2019-08-26
NMS-12243: Leverage telemetry sequencing for better thresholding performance

mattixtech:
This PR updates thresholding to take into account the sequence number of collections sets derived from JTI messages. Using the sequence number in a distributed thresholding environment, we can determine if we were the instance that thresholded the previous sequence and in this case we can avoid retrieving the thresholding state from the KV store since we know we are up to date.

With this in place, the number of reads from the KV store that will be required by thresholding in a distributed deployment should be greatly reduced, resulting in better performance. In a steady state scenario we would expect that the same instance will receive the metrics for a given agent/resource and this should represent that majority of incoming messages. In an outage/Kafka partition rebalancing scenario the sequence number will indicate to us that we need to fetch state to resume and this should not happen often.

This was tested by adding coverage to existing IT.

This only works for metrics derived from JTI messages as I did not see a sequence number equivalent for NXOS/Netflow.

### All Contributors

### Pull Request Process

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12243

### DIFF
--- a/features/collection/thresholding/impl/src/main/java/org/opennms/netmgt/threshd/AbstractThresholdEvaluatorState.java
+++ b/features/collection/thresholding/impl/src/main/java/org/opennms/netmgt/threshd/AbstractThresholdEvaluatorState.java
@@ -84,7 +84,11 @@ public abstract class AbstractThresholdEvaluatorState<T extends Serializable> im
     protected final ThresholdingSession thresholdingSession;
     
     private static final String THRESHOLDING_KV_CONTEXT = "thresholding";
+    
     private final int stateTTL;
+    
+    private Long sequenceNumber;
+
     private boolean firstEvaluation = true;
 
     /**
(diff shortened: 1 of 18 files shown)


j-white:
Looks good - this should help performance a lot.

For sFlow, the sequence number is written by the parser here: https://github.com/OpenNMS/opennms/blob/3cb76440e46c33fc77e3b010723b1c59f0c91d8d/features/telemetry/protocols/sflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/sflow/parser/proto/flows/SampleDatagramV5.java#L102

This can then be accessed by the `SFlowTelemetryAdapter` similarly to the other fields in the bson document.

mattixtech:
Rebased and updated sflow to also set the sequence number.

features/collection/thresholding/impl/src/main/java/org/opennms/netmgt/threshd/AbstractThresholdEvaluatorState.java
```diff
@@ -155,7 +155,7 @@ private void fetchState() {
         // If both of those conditions are false, then we must be on a standalone instance of OpenNMS and have the state
         // already in memory so there is no need to fetch it
         if (!isDistributed() && !firstEvaluation) {
-            fetchState();
+            return;
```
mattixtech: Note: this change is just due to screwing up the merge. Just fixed it in this commit since it will all get squashed anyway.

PR #2671, opened 2019-08-27
LTS-234: Allow special characters in scheduled outage name

christianpape:
* JIRA: https://issues.opennms.org/browse/LTS-234


indigo423:
Tested the branch with:
* Create a scheduled outage named `My Outage #42`
* Edit the scheduled outage
* Deleted the scheduled outage

without any issues, looks good to me 👍

PR #2673, opened 2019-08-28
NMS-12261: use backwards-compatible date serialization

RangerRick:
### All Contributors

### Pull Request Process

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12261

### DIFF
--- a/features/events/api/src/main/java/org/opennms/netmgt/events/api/EventConstants.java
+++ b/features/events/api/src/main/java/org/opennms/netmgt/events/api/EventConstants.java
@@ -32,10 +32,6 @@ import java.math.BigInteger;
 import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
-import java.time.ZonedDateTime;
-import java.time.format.DateTimeFormatter;
-import java.time.format.DateTimeParseException;
-import java.util.Arrays;
 import java.util.Date;
 import java.util.Locale;
 import java.util.TimeZone;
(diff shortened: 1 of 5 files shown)


j-white:
No problems in setting these flags when still using JDK 8?

Ideally we'd solve this in the code instead of having to set these flags everywhere, but it looks like that's captured in NMS-12263.

We also need to update: https://github.com/OpenNMS/opennms/blob/develop/opennms-container/horizon/assets/entrypoint.sh#L168

RangerRick:
> No problems in setting these flags when still using JDK 8?

Hmm, good point.  Looks like JDK8 expects "JRE" and "CLDR" rather than "COMPAT" and "CLDR".  In JDK 11 they say:

`JRE: Represents a synonym to "COMPAT". This name is deprecated and will be removed in the future release of JDK.`

:/

Guess I should version-box them in the other scripts too.

> Ideally we'd solve this in the code instead of having to set these flags everywhere, but it looks like that's captured in NMS-12263.

Yeah, this is a short-term fix.  Long-term we need to make sure we can be forgiving of our inputs before CLDR becomes the default.

> We also need to update: https://github.com/OpenNMS/opennms/blob/develop/opennms-container/horizon/assets/entrypoint.sh#L168

Ooh, good catch.  Missed this one.

RangerRick:
I just committed something that handles this, but it got real complicated.  I think I can do it in a simpler way, so marking as triage for now.  Don't bother reviewing until I've released it.

j-white:
This is starting to looks like its more work than actual fixing the date formats.

RangerRick:
Well, I was trying to reuse the code for getting the java version, which can get complicated across JDK/OpenJDK/betas/etc.  Breaking some of this out is actually something I'd been meaning to do for a while, this was just the trigger.  :)

RangerRick:
Alright, I made [NMS-12266](http://issues.opennms.org/browse/NMS-12266) to track the actual base changes for the CLI module.  That simplifies _this_ patch immensely.  I went ahead and rebased it against `NMS-12266`, cleaned it up, and force-pushed it, because things had changed so much.

I'll wait for `NMS-12266` to finish and confirm it doesn't break anything, and then un-triage this after it at least passes circle.  (If it does. <grin>)

RangerRick:
OK, this is the simplistic version that just changes OpenNMS to `COMPAT` but _not_ anything else.  All the tests pass, and it should be the most minimal version of this, but ultimately I think post-25 updating all our dates to be ISO-8601 is better and the previous patch should be revisited for [NMS-12263](https://issues.opennms.org/browse/NMS-12263).

j-white:
We still need to update the start script for the container: https://github.com/OpenNMS/opennms/blob/bd4bd561276b389008ccaf5251acd13dd4bac50e/opennms-container/horizon/container-fs/entrypoint.sh#L134

Without this, I believe that the Docker container (& system tests) won't run with the new flag.

RangerRick:
D'oh, good catch.  Cleaned that out when I pulled the other bits.  Guess *now* we'll see what happens...  :)

RangerRick:
Ah good, looks like it still passes in circle with that change: https://github.com/OpenNMS/opennms/pull/2673

RangerRick:
Closing this PR for now, I'm working on it under NMS-12263 anyway.

PR #2679, opened 2019-09-03
NMS-12256: Resource Graphs : Fix long URL issues with Graph Selected

cgorantla:
In order to resolve long URL issues associated with  graph selected, we save resources in DB ( as json type) with an Id and use this `generatedId` to retrieve all resources.

The ID is generated by UUID from resources (bytes).
It should be same if same resources are selected. 

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12256

### DIFF
--- a/core/web-assets/src/main/assets/js/apps/onms-resources/index.js
+++ b/core/web-assets/src/main/assets/js/apps/onms-resources/index.js
@@ -172,7 +172,7 @@ angular.module('onms-resources', [
       if ($scope.resources.hasOwnProperty(key)) {
         _.each($scope.filteredResources[key], function(r) {
           if (r.selected) {
-            selected.push('resourceId=' + r.id);
+            selected.push(r.id);
           }
         });
       }
@@ -189,7 +189,7 @@ angular.module('onms-resources', [
     }
   };
 
-  $scope.doGraph = function(selected) {
+  $scope.doGraph = function (selected) {
     // Save resources with an ID and form url with generatedId.
     if (selected.length > 0) {
       $http.post('rest/resources/generateId', selected)
@@ -198,14 +198,21 @@ angular.module('onms-resources', [
           if ($scope.generatedId) {
             $window.location.href = getBaseHref() + $scope.url + '?generatedId=' + $scope.generatedId + ($scope.reports ? '&reports=' + $scope.reports : '');
           } else {
-            $window.location.href = getBaseHref() + $scope.url + '?' + selected.join('&') + ($scope.reports ? '&reports=' + $scope.reports : '');
+            $scope.setResourceIds(selected);
           }
         }).error(function (error, status) {
-          $window.location.href = getBaseHref() + $scope.url + '?' + selected.join('&') + ($scope.reports ? '&reports=' + $scope.reports : '');
+          $scope.setResourceIds(selected);
         });
     } else {
       growl.error('Please select at least one resource.');
     }
+  };
+
+  $scope.setResourceIds = function (selected) {
+    for (var i = 0; i < selected.length; i++) {
+      selected[i] = 'resourceId=' + selected[i];
+    }
+    $window.location.href = getBaseHref() + $scope.url + '?' + selected.join('&') + ($scope.reports ? '&reports=' + $scope.reports : '');
   };
 
   $scope.$watch('searchQuery', function() {
--- a/core/web-assets/src/main/assets/js/apps/onms-resources/index.js
+++ b/core/web-assets/src/main/assets/js/apps/onms-resources/index.js
@@ -83,6 +83,7 @@ angular.module('onms-resources', [
   $scope.url = 'graph/results.htm';
   $scope.reports = 'all';
   $scope.loaded = false;
+  $scope.generatedId = '';
 
   $scope.init = function(nodeCriteria, reports, endUrl) {
     if (!nodeCriteria) {
@@ -105,7 +106,7 @@ angular.module('onms-resources', [
       $scope.loaded = true;
       $scope.hasResources = data.children.resource.length > 0;
       var reduced = _.map(data.children.resource, function(obj) {
-	var resource = {
+	    var resource = {
           id: obj.id,
           label: obj.label,
           typeLabel: obj.typeLabel,
@@ -190,7 +191,17 @@ angular.module('onms-resources', [
 
   $scope.doGraph = function(selected) {
     if (selected.length > 0) {
-      $window.location.href = getBaseHref() + $scope.url + '?' + selected.join('&') + ($scope.reports ? '&reports=' + $scope.reports : '');
+      $http.post('rest/resources/generateId', selected)
+      .success(function(response) {
+        $scope.generatedId = response;
+        if($scope.generatedId) {
+            $window.location.href = getBaseHref() + $scope.url + '?generatedId=' + $scope.generatedId + ($scope.reports ? '&reports=' + $scope.reports : '');
+        } else {
+            $window.location.href = getBaseHref() + $scope.url + '?' + selected.join('&') + ($scope.reports ? '&reports=' + $scope.reports : '');
+        }
+      }).error(function(error, status) {
+        $window.location.href = getBaseHref() + $scope.url + '?' + selected.join('&') + ($scope.reports ? '&reports=' + $scope.reports : '');
+      });
     } else {
       growl.error('Please select at least one resource.');
     }
--- a/core/web-assets/src/main/assets/js/apps/onms-resources/index.js
+++ b/core/web-assets/src/main/assets/js/apps/onms-resources/index.js
@@ -105,8 +105,8 @@ angular.module('onms-resources', [
       $scope.nodeLabel = data.label;
       $scope.loaded = true;
       $scope.hasResources = data.children.resource.length > 0;
-      var reduced = _.map(data.children.resource, function(obj) {
-	    var resource = {
+      var reduced = _.map(data.children.resource, function (obj) {
+        var resource = {
           id: obj.id,
           label: obj.label,
           typeLabel: obj.typeLabel,
@@ -190,18 +190,19 @@ angular.module('onms-resources', [
   };
 
   $scope.doGraph = function(selected) {
+    // Save resources with an ID and form url with generatedId.
     if (selected.length > 0) {
       $http.post('rest/resources/generateId', selected)
-      .success(function(response) {
-        $scope.generatedId = response;
-        if($scope.generatedId) {
+        .success(function (response) {
+          $scope.generatedId = response;
+          if ($scope.generatedId) {
             $window.location.href = getBaseHref() + $scope.url + '?generatedId=' + $scope.generatedId + ($scope.reports ? '&reports=' + $scope.reports : '');
-        } else {
+          } else {
             $window.location.href = getBaseHref() + $scope.url + '?' + selected.join('&') + ($scope.reports ? '&reports=' + $scope.reports : '');
-        }
-      }).error(function(error, status) {
-        $window.location.href = getBaseHref() + $scope.url + '?' + selected.join('&') + ($scope.reports ? '&reports=' + $scope.reports : '');
-      });
+          }
+        }).error(function (error, status) {
+          $window.location.href = getBaseHref() + $scope.url + '?' + selected.join('&') + ($scope.reports ? '&reports=' + $scope.reports : '');
+        });
     } else {
       growl.error('Please select at least one resource.');
     }
--- a/features/distributed/kv-store/json/postgres/src/main/resources/META-INF/opennms/applicationContext-postgresJsonStore.xml
+++ b/features/distributed/kv-store/json/postgres/src/main/resources/META-INF/opennms/applicationContext-postgresJsonStore.xml
@@ -7,9 +7,9 @@
 
     <bean name="dataSource" class="org.opennms.core.db.DataSourceFactory" factory-method="getInstance"/>
 
-    <bean name="postgresJsonStore" class="org.opennms.features.distributed.kvstore.json.postgres.PostgresJsonStore">
+    <bean name="jsonStore" class="org.opennms.features.distributed.kvstore.json.postgres.PostgresJsonStore">
         <constructor-arg ref="dataSource"/>
     </bean>
-    <onmsgi:service interface="org.opennms.features.distributed.kvstore.api.JsonStore" ref="postgresJsonStore"/>
+    <onmsgi:service interface="org.opennms.features.distributed.kvstore.api.JsonStore" ref="jsonStore"/>
 
 </beans>
--- a/features/distributed/kv-store/json/postgres/src/main/resources/META-INF/opennms/applicationContext-postgresJsonStore.xml
+++ b/features/distributed/kv-store/json/postgres/src/main/resources/META-INF/opennms/applicationContext-postgresJsonStore.xml
@@ -7,9 +7,9 @@
 
     <bean name="dataSource" class="org.opennms.core.db.DataSourceFactory" factory-method="getInstance"/>
 
-    <bean name="jsonStore" class="org.opennms.features.distributed.kvstore.json.postgres.PostgresJsonStore">
+    <bean name="postgresJsonStore" class="org.opennms.features.distributed.kvstore.json.postgres.PostgresJsonStore">
         <constructor-arg ref="dataSource"/>
     </bean>
-    <onmsgi:service interface="org.opennms.features.distributed.kvstore.api.JsonStore" ref="jsonStore"/>
+    <onmsgi:service interface="org.opennms.features.distributed.kvstore.api.JsonStore" ref="postgresJsonStore"/>
 
 </beans>
--- a/opennms-web-api/src/main/java/org/opennms/web/svclayer/support/DefaultGraphResultsService.java
+++ b/opennms-web-api/src/main/java/org/opennms/web/svclayer/support/DefaultGraphResultsService.java
@@ -166,17 +166,15 @@ public class DefaultGraphResultsService implements GraphResultsService, Initiali
             }
         }
         // GraphSelected case where all resources are fetched from generatedId
-        if (!Strings.isNullOrEmpty(generatedId) && m_jsonStore != null) {
+        if (!Strings.isNullOrEmpty(generatedId)) {
             Optional<String> result = m_jsonStore.get(generatedId, RESOURCE_IDS_CONTEXT);
             if (result.isPresent()) {
                 try {
                     String[] resourceArray = m_gson.fromJson(result.get(), String[].class);
                     for (String resourceId : resourceArray) {
                         try {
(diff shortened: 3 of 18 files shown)


cgorantla:
> Is there an explanation as to why you have to reference the `postgresJsonStore` directly? That seems like it defeats the purpose of exporting services as interfaces instead. I know you were running into issues with making it happy but I wonder if it's possible to grab it out of the `onmsgi:` registry instead or something?

The issue I faced was having two copies of `kvstore-api` That is resolved. Here I just changed name of the bean from `jsonStore` to `postgresJsonStore` to indicate what it is explicitly.

RangerRick:
> The issue I faced was having two copies of `kvstore-api` That is resolved. Here I just changed name of the bean from `jsonStore` to `postgresJsonStore` to indicate what it is explicitly.

I guess what I mean is, if we know that the "normal" jsonStore object is published to `<onmsgi:service>`, could you just make a bean ref in the web context using `<onmsgi:reference>` to pull that same bean back out?  I believe that's one way to cross those boundaries without ending up with multiple copies.

core/web-assets/src/main/assets/js/apps/onms-resources/index.js
```diff
@@ -105,7 +106,7 @@ angular.module('onms-resources', [
       $scope.loaded = true;
       $scope.hasResources = data.children.resource.length > 0;
       var reduced = _.map(data.children.resource, function(obj) {
-	var resource = {
+	    var resource = {
```
mattixtech: Looks like the indenting of the object's fields needs to be updated since this is being updated.

core/web-assets/src/main/assets/js/apps/onms-resources/index.js
```diff
@@ -190,7 +191,17 @@ angular.module('onms-resources', [
 
   $scope.doGraph = function(selected) {
     if (selected.length > 0) {
-      $window.location.href = getBaseHref() + $scope.url + '?' + selected.join('&') + ($scope.reports ? '&reports=' + $scope.reports : '');
+      $http.post('rest/resources/generateId', selected)
+      .success(function(response) {
+        $scope.generatedId = response;
+        if($scope.generatedId) {
```
mattixtech:
```suggestion
        if ($scope.generatedId) {
```

opennms-web-api/src/main/java/org/opennms/web/svclayer/support/DefaultGraphResultsService.java
```diff
@@ -349,5 +382,12 @@ public void setRrdDao(RrdDao rrdDao) {
     public void setEventProxy(EventProxy eventProxy) {
         m_eventProxy = eventProxy;
     }
-    
+
+    /**
```
mattixtech: Delete the javadoc here as it is implied

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/ResourceRestService.java
```diff
@@ -123,4 +138,21 @@ public ResourceDTO getResourceForNode(@PathParam("nodeCriteria") final String no
...
         return ResourceDTO.fromResource(resource, depth);
     }
+
+
+    @POST
+    @Path("generateId")
+    @Consumes({MediaType.APPLICATION_JSON})
+    @Produces({MediaType.APPLICATION_JSON})
+    public Response saveResourcesWithId(String[] resources) {
+        if (m_jsonStore == null) {
```
mattixtech: Is this condition necessary? If there was no JsonStore to wire in then this bean would not be created by Spring.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/ResourceRestService.java
```diff
@@ -123,4 +138,21 @@ public ResourceDTO getResourceForNode(@PathParam("nodeCriteria") final String no
...
+            throw getException(Status.NOT_FOUND, "No JsonStore registered.");
+        }
+        String resourcesInJson = new Gson().toJson(resources);
```
mattixtech: Create a private final member in this bean that contains the `new Gson()` instance and reuse it here rather than create a new object each `POST`

```suggestion
        String resourcesInJson = m_gson.toJson(resources);
```

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/ResourceRestService.java
```diff
@@ -123,4 +138,21 @@ public ResourceDTO getResourceForNode(@PathParam("nodeCriteria") final String no
...
+        String key = UUID.nameUUIDFromBytes(resourcesInJson.getBytes()).toString();
+        m_jsonStore.put(key, resourcesInJson, RESOURCE_IDS_CONTEXT);
+        String jsonKey = new Gson().toJson(key);
```
mattixtech: Same here
```suggestion
        String jsonKey = m_gson.toJson(key);
```

features/distributed/kv-store/json/postgres/src/main/resources/META-INF/opennms/applicationContext-postgresJsonStore.xml
```diff
@@ -7,9 +7,9 @@
 
     <bean name="dataSource" class="org.opennms.core.db.DataSourceFactory" factory-method="getInstance"/>
 
-    <bean name="jsonStore" class="org.opennms.features.distributed.kvstore.json.postgres.PostgresJsonStore">
+    <bean name="postgresJsonStore" class="org.opennms.features.distributed.kvstore.json.postgres.PostgresJsonStore">
```
mattixtech: Currently the intent is that there would only ever be one instance fo the jsonStore. Changing the name here creates an implicit dependency on this specific implementation (the postgres one). I think we should keep the naming as is since it simplifies swapping out implementations in the future.

Since there is only one instance, we do not need to be able to differentiate based on the name.

mattixtech:
> > The issue I faced was having two copies of `kvstore-api` That is resolved. Here I just changed name of the bean from `jsonStore` to `postgresJsonStore` to indicate what it is explicitly.
> 
> I guess what I mean is, if we know that the "normal" jsonStore object is published to `<onmsgi:service>`, could you just make a bean ref in the web context using `<onmsgi:reference>` to pull that same bean back out? I believe that's one way to cross those boundaries without ending up with multiple copies.

From what I can tell, we are still referencing by the interface `JsonStore`, so that part is fine. It just looks like the bean names changed to use an implementation specific name. Unless I am missing something, please point me to it if I am.

features/distributed/kv-store/json/postgres/src/main/resources/META-INF/opennms/applicationContext-postgresJsonStore.xml
cgorantla: I specifically want to use `postgresJsonStore` for the current usecase.  A `NoSQL` jsonStore would be slow for this usecase.  I think  having `postgresJsonStore` as bean name would be more visible for anybody trying to understand the code or changing the code in future instead of assuming that it may be Postgres.  That is my intention of changing this.

RangerRick:
@mattixtech Yeah, Chandra and I just talked about it, I was misunderstanding the changes to the spring context and hadn't noticed the `@AutoWired` was still using the generic interface properly.  I rescind my comments. ;)

opennms-web-api/src/main/java/org/opennms/web/svclayer/support/DefaultGraphResultsService.java
cgorantla: Just made it look similar to the other methods in the Class.

cgorantla:
Addressed review comments, Please review again.

opennms-web-api/src/main/java/org/opennms/web/svclayer/support/DefaultGraphResultsService.java
```diff
@@ -383,9 +385,11 @@ public void setEventProxy(EventProxy eventProxy) {
         m_eventProxy = eventProxy;
     }
 
+
     /**
+     * <p>setJsonStore</p>
```
mattixtech: Still think this javadoc should be removed since it doesn't really add any value.

opennms-web-api/src/main/java/org/opennms/web/svclayer/support/DefaultGraphResultsService.java
```diff
@@ -83,6 +88,10 @@
 
     private EventProxy m_eventProxy;
 
+    private JsonStore m_jsonStore;
+
+    private Gson m_gson = new Gson();
```
mattixtech:
```suggestion
    private final Gson m_gson = new Gson();
```

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/ResourceRestService.java
```diff
@@ -62,12 +73,20 @@
...
     private NodeDao m_nodeDao;
 
     @Autowired
     private ResourceDao m_resourceDao;
 
+    @Autowired
+    @Qualifier("postgresJsonStore")
+    private JsonStore m_jsonStore;
+
+    private Gson m_gson = new Gson();
```
mattixtech:
```suggestion
    private final Gson m_gson = new Gson();
```

features/distributed/kv-store/json/postgres/src/main/resources/META-INF/opennms/applicationContext-postgresJsonStore.xml
mattixtech: The client code shouldn't be deciding which implementation it is going to use. There is only going to be one implementation per deployment right now so a qualifier is going to fail if the one you are looking for is not installed. The bean was named that way for precisely the reason that people should not be assuming that it is implemented with Postgres, that implementation detail should be irrelevant for use cases of this API.

Client code should be agnostic of implementation based on the current paradigm. We may look into providing multiple implementations in the future at which point qualifier based selection would be reasonable, but that is not the case currently.

Can you explain why you believe the Cassandra implementation would be too slow? The API only provides primary key based get/set. NoSQL DB such as Cassandra handle that case very well.

cgorantla: From my limited understanding some NoSQL DBs are optimized for writing lots of data.  In case of our Newts solution, Rest APIs take longer fetching the graphs. Anyways I will go with your naming pattern.

opennms-web-api/src/main/java/org/opennms/web/svclayer/support/DefaultGraphResultsService.java
cgorantla: I prefer consistency with the rest of the code in the file.

opennms-web-api/src/main/java/org/opennms/web/svclayer/support/DefaultGraphResultsService.java
```diff
@@ -154,6 +165,29 @@ public GraphResults findResults(ResourceId[] resourceIds, String[] reports, Stri
...
         }
+        // GraphSelected case where all resources are fetched from generatedId
+        if (!Strings.isNullOrEmpty(generatedId) && m_jsonStore != null) {
+            Optional<String> result = m_jsonStore.get(generatedId, RESOURCE_IDS_CONTEXT);
+            if (result.isPresent()) {
+                try {
+                    String[] resourceArray = m_gson.fromJson(result.get(), String[].class);
+                    for (String resourceId : resourceArray) {
+                        try {
+                            //Resources are saved as keyValue resourceId=node[12312].
```
mattixtech: Can you provide an example as to what the JSON content looks like? What does the content of the POST to the ReST endpoint look like?

Rather than splitting strings based on `=` it would probably better to send/store JSON objects and reference the fields we want by their key.

So this:
```
{
    [
        {"id": "a", "value": "node[123123]"},
        {"id": "b", "value": "node[456567]"}
    ]
}
```

instead of this (assuming this is what it currently looks like):
```
{
    [
        "a=node[123123]",
        "b=node[456567]"
    ]
}
```

cgorantla: The values are just strings. `resourceId=node[3232]` is how servlet request parms were sent. I m using the same inputs here. I think it would be better just parse the values at front end before `POST` call and just send strings.

mattixtech: Ok, I think the ReST endpoint should be updated to take a JSON array of IDs rather than key value pairs. It doesn't make sense to persist the whole key value pair when we only care about the ID portion.

The ReST endpoint's contract should be:
Input:
A JSON encoded list of resource IDs. i.e. `{["id1", "id2", "id3"]}`
Result:
A generated ID that maps all the given IDs.

Once that is done, we won't need to split the results anymore when we retrieve them from the JSON store. The client side will have to be smarter about how it sends the request however.

Also the list of resource IDs should be contained in the body of the `POST` message not as part of query parameter string (Assuming this is already the case, just making sure).

mattixtech: Looks like you thought the same, seems the commit you just pushed does this 👍

opennms-web-api/src/main/java/org/opennms/web/svclayer/support/DefaultGraphResultsService.java
```diff
@@ -154,6 +165,25 @@ public GraphResults findResults(ResourceId[] resourceIds, String[] reports, Stri
                 }
             }
         }
+        // GraphSelected case where all resources are fetched from generatedId
+        if (!Strings.isNullOrEmpty(generatedId) && m_jsonStore != null) {
```
mattixtech: If the jsonStore is null, this will just get silently skipped. Probably better to let line 170 throw an NPE so we know why it failed.
```suggestion
        if (!Strings.isNullOrEmpty(generatedId)) {
```

opennms-web-api/src/main/java/org/opennms/web/svclayer/support/DefaultGraphResultsService.java
```diff
@@ -154,6 +165,25 @@ public GraphResults findResults(ResourceId[] resourceIds, String[] reports, Stri
...
+            Optional<String> result = m_jsonStore.get(generatedId, RESOURCE_IDS_CONTEXT);
+            if (result.isPresent()) {
+                try {
+                    String[] resourceArray = m_gson.fromJson(result.get(), String[].class);
+                    for (String resourceId : resourceArray) {
+                        try {
+                            OnmsResource resource = m_resourceDao.getResourceById(ResourceId.fromString(resourceId));
+                            graphResults.addGraphResultSet(createGraphResultSet(null, resource, reports, graphResults));
+                        } catch (IllegalArgumentException e) {
```
mattixtech: Is this catch block attempting to cover the case where a resource is no longer valid?

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/ResourceRestService.java
```diff
@@ -123,4 +140,21 @@ public ResourceDTO getResourceForNode(@PathParam("nodeCriteria") final String no
...
     }
+
+
+    @POST
+    @Path("generateId")
+    @Consumes({MediaType.APPLICATION_JSON})
+    @Produces({MediaType.APPLICATION_JSON})
+    public Response saveResourcesWithId(String[] resources) {
+
+        String resourcesInJson = m_gson.toJson(resources);
```
mattixtech: Should we verify that all the given resource IDs are valid at insertion time?

cgorantla: I want the `POST` call to be responsive enough. That's the reason I don't even check `key` to exist.

opennms-web-api/src/main/java/org/opennms/web/svclayer/support/DefaultGraphResultsService.java
cgorantla: It is sanity check. We don't want to throw NPE in the application

opennms-web-api/src/main/java/org/opennms/web/svclayer/support/DefaultGraphResultsService.java
cgorantla: Fixed

cgorantla:
@mattixtech It would be helpful if you can review at once. I know that it is not always possible if there is new code but it saves me lot of time going back to the same branch if code exist from the start

mattixtech:
> @mattixtech It would be helpful if you can review at once. I know that it is not always possible if there is new code but it saves me lot of time going back to the same branch if code exist from the start

I try, I only noticed the other things after re-reviewing your latest commits.

opennms-web-api/src/main/java/org/opennms/web/svclayer/support/DefaultGraphResultsService.java
mattixtech: Wouldn't we rather throw an NPE than silently fail? How will the user know what went wrong here?

There is lots of cases where we explicitly throw an NPE when sanity checking, thats why we use `Objects.requireNonNull()`.

You could change the setter so that it does the `nonNull` check and then remove this null check from the conditional.

mattixtech: Actually, the `afterPropertiesSet()` already does the equivalent. So you could simply remove the null check from the conditional.

cgorantla: If you look at the changes completely, it's already there. https://github.com/OpenNMS/opennms/pull/2679/files#diff-9c9a5247b8b6fcade3daaae74fe929c9R301

cgorantla: > Actually, the `afterPropertiesSet()` already does the equivalent. So you could simply remove the null check from the conditional.
Having null check is okay. I don't want to go through another iteration.

mattixtech: I think I missed it from the force push.

cgorantla: > I think I missed it from the force push.

It existed from start of this PR

mattixtech: > > I think I missed it from the force push.
> 
> It existed from start of this PR

I was referring to your removal of the null check in the conditional. It was there in previous commit and removed after force push. So I didn't see the it change in a new commit which is why I missed it.

PR #2685, opened 2019-09-07
NMS-9784: JMX config clean up

mfuhrmann:
The branch https://github.com/OpenNMS/opennms/tree/jira/NMS-9784/opennms-base-assembly/src/main/filtered/etc was initially created to solve NMS-9784. But it was to inversive because of metric name changes etc.
It can be deleted, after merging this one.

This PR now contains only a separation of the `jmx-datacollection-config.xml` content into single files.
No impact for all OpenNMS' out there.
`jmx-datacollection-config.xml` is now empty.

Also, I've used identical names for data collection files and graph config files.

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-9784


opennms-webapp-rest/src/test/java/org/opennms/web/rest/v1/config/JmxDataCollectionConfigResourceIT.java
```diff
@@ -84,17 +84,49 @@ public void testJmxConfig() throws Exception {
 
         assertThat(config.getJmxCollectionCount(), greaterThanOrEqualTo(6));
 
+        assertEquals("activemq", config.getJmxCollection("activemq").getName());
```
j-white: We can remove these. The purpose of this test is to verify that we can retrieve the JMX configuration via REST. We don't need to assert all of the groups - it's a pain to maintain otherwise.

opennms-base-assembly/src/main/filtered/etc/jmx-datacollection-config.d/jvm.xml
```diff
@@ -0,0 +1,155 @@
+<?xml version="1.0"?>
+<jmx-datacollection-config rrdRepository="${install.share.dir}/rrd/snmp/">
+    <jmx-collection name="jvm">
```
indigo423: @mfuhrmann can you please double check here? The OpenNMS-JVM is looking for a `jsr160` collection and it seems like this is renamed here to `jvm` see [here](https://github.com/OpenNMS/opennms/blob/a0fd2dcbfeed754fd8b8a1188d8123fc74e3c569/opennms-base-assembly/src/main/filtered/etc/collectd-configuration.xml#L112)

opennms-base-assembly/src/main/filtered/etc/collectd-configuration.xml
```diff
@@ -109,10 +109,30 @@
          <parameter key="retry" value="2"/>
          <parameter key="timeout" value="3000"/>
          <parameter key="rrd-base-name" value="java"/>
-         <parameter key="collection" value="jsr160"/>
+         <parameter key="collection" value="jmx-opennms"/>
```
mvrueden: We have to ensure that this does not break existing behaviour, e.g. does with this change the `JMXDetector` still work for OpenNMS?

indigo423: > We have to ensure that this does not break existing behavior, e.g. does with this change the `JMXDetector` still work for OpenNMS?

yes, this is the reference to the data collection package in collectd and does not interfere with the JMX Detector.

mvrueden:
I added the `triage` label as we need to verify that this change does not break existing behaviour.

opennms-provision/opennms-provision-persistence/src/main/resources/org/opennms/netmgt/provision/persist/default-foreign-source.xml
```diff
@@ -26,6 +26,31 @@
...
+			<parameter key="port" value="18980"/>
+			<parameter key="password" value="admin"/>
+			<parameter key="urlPath" value="/jmxrmi"/>
+			<parameter key="retries" value="2"/>
+			<parameter key="type" value="default"/>
+			<parameter key="protocol" value="rmi"/>
+			<parameter key="timeout" value="3000"/>
+			<parameter key="object" value="org.opennms.newts:name=ring-buffer.size"/>
+		</detector>
+		<detector name="JMX-Java-Virtual-Machine" class="org.opennms.netmgt.provision.detector.jmx.Jsr160Detector">
```
mvrueden: Shouldn't this be `JMX-Java-8`, otherwise the detector finds a service called `JMX-Java-Virtual-Machine` which is `NOT MONITORED`

opennms-base-assembly/src/main/filtered/etc/collectd-configuration.xml
mvrueden: That was probably a bad example. I found other references to "jsr160" in the code, e.g. in https://github.com/OpenNMS/opennms/blob/9eaac5f862c6c8891bb63326b84a7f6e946d451d/opennms-doc/guide-admin/src/asciidoc/text/performance-data-collection/collectors/jmx.adoc#L150 or some data-collection examples. We should make them work as well, if we change the value here, that's all I am saying :)

mvrueden:
I fixed the failing test. Let's see how far we can get. Some references, e.g. in the JMX.adoc and some examples are still referencing `jsr160`. Please make sure that they either work as given or must be changed as well @mfuhrmann

mfuhrmann:
I've changed the example data collection name to `jmx-wildfly`. So it should not confuse someone.
The other `jsr160` hits are still correct, since the collector name itself is still the same.

mfuhrmann:
@indigo423 @mvrueden So, are we done here?

mfuhrmann:
@indigo423 Can we proceed here? I mean closing this one and pushing https://github.com/OpenNMS/opennms/pull/2899

indigo423:
Closed in favor of #2899

PR #2686, opened 2019-09-09
NMS-10071: Poller config reload for reload events.

cgorantla:
Update internal data for reload events to fetch any updated config from included files.

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-10071


opennms-config/src/test/resources/poller-configuration.xml
```diff
@@ -0,0 +1,512 @@
...
+    <!--<include-range begin="1.1.1.1" end="254.254.254.254" />-->
+    <include-url>file:${INCLUDE_URL_FILE}</include-url>
+    <rrd step="300">
+      <rra>RRA:AVERAGE:0.5:1:2016</rra>
+      <rra>RRA:AVERAGE:0.5:12:1488</rra>
+      <rra>RRA:AVERAGE:0.5:288:366</rra>
+      <rra>RRA:MAX:0.5:288:366</rra>
+      <rra>RRA:MIN:0.5:288:366</rra>
+    </rrd>
+    <service name="ICMP" interval="300000" user-defined="false" status="on">
```
j-white: Can we trim this file down to only keep what's necessary? (Seems like it's only used to testing the `include-url`, so we don't need all of the services / monitors defined.)

opennms-config/src/main/java/org/opennms/netmgt/config/PollerConfigManager.java
```diff
@@ -133,7 +133,9 @@ protected void setUpInternalData() {
      * @throws java.io.IOException if any.
      */
     @Override
-    public abstract void update() throws IOException;
+    public void update() throws IOException {
+        setUpInternalData();
```
j-white: We may need to review the locking here. Looks like there may be cases when callers can get an invalid state (`createUrlIpMap` doesn't lock,)

cgorantla: Sure. Since this is only updating internal state and not changing anything in config. added read lock for `update`.

cgorantla:
Bamboo passed.

PR #2688, opened 2019-09-10
NMS-12265: Fixed ServiceInstancePool housekeeping timer

christianpape:
The ServiceInstancePool housekeeping timer wasn't declared as daemon thread.
* JIRA: https://issues.opennms.org/browse/NMS-12265


agalue:
As the problem happens regardless if the communication was successful, I was able to reproduce the problem trying to access a non-existing server. Then, I manually applied the simple fix, and now the command works!

Interestingly, while I was doing that I found that the "example" usage of that tool is incorrect. The tool says you should pass a URL in the form of `vmware://user:pass@vcenter-server/my-requisition`, but that doesn't work. It should be `vmware://vcenter-server/my-requisition?username=user;password=pass`, or use `vmware-config.xml`.

christianpape:
I created an issue for that:
* JIRA: https://issues.opennms.org/browse/NMS-12275

PR #2689, opened 2019-09-10
NMS-12233: Generate list of unit and integration tests based on files modified

j-white:
JIRA: https://issues.opennms.org/browse/NMS-12233

Here we add a new script used by the integration test phase in CircleCI that will determine the list of ITs to run based on:
1) The files that were modified
2) The modules that depend on those with files modified

We also include an script in the top-level directory that can be used to execute the same logic locally.

This is expected to drastically reduce the number of tests we run on commit, and in some cases eliminate test runs altogether (i.e. when modifying docs).


.circleci/scripts/find-tests/maven_test.py
```diff
@@ -0,0 +1,61 @@
...
+
+    @unittest.skip("requires filesystem access")
+    def test_get_modules_for_classnames(self):
+        project = MavenProject(self.sample_graph)
+        modules = project.get_modules_for_classnames("org.opennms.netmgt.poller.PollerIT")
+        self.assertEqual(len(modules), 0)
+
+    def test_get_modules_using(self):
+        project = MavenProject(self.sample_graph)
+        schema_module = project.get_module("org.opennms.core:org.opennms.core.schema:25.0.0-SNAPSHOT")
```
indigo423: Is `25.0.0-SNAPSHOT` hardcoded and needs to be adjusted over time?

.circleci/scripts/find-tests/maven_test.py
```diff
@@ -0,0 +1,61 @@
...
+        self.assertEqual(len(project.modules), 4)
+        # First project should be the root module
+        self.assertEqual(project.modules[0].is_root(), True)
+        # Subsequent module should not be
+        self.assertEqual(project.modules[1].is_root(), False)
+
+    def test_get_related_modules(self):
+        project = MavenProject(self.sample_graph)
+        related = project.get_modules_related_to([os.path.abspath(
+            "/home/jesse/git/opennms/opennms-services/src/main/java/org/opennms/netmgt/poller/Poller.java")])
```
indigo423: We use your home dir :)

.circleci/scripts/find-tests/maven_test.py
j-white: This is just an example in the test case.

.circleci/scripts/find-tests/maven_test.py
j-white: This is just an example in the test case.

.nightly
```diff
@@ -1 +1,2 @@
 repo: develop
+branch: release-25.0.0
```
RangerRick: Instead of `branch:` which is kind of ambiguous (what branch? this branch?) should this maybe be `parent_branch:` or `root_branch:` or something?

j-white: I'll update it to be `parent_branch`

runtests.sh
```diff
@@ -0,0 +1,43 @@
...
+      --output=target/test_projects \
+      .
+
+# Run the tests
+mvn verify -P'!checkstyle' \
+           -DupdatePolicy=never \
+           -Dbuild.skip.tarball=true \
+           -DfailIfNoTests=false \
+           -DskipITs=false \
+           -Dci.rerunFailingTestsCount=0 \
```
Naicisum:
```suggestion
           -Dci.rerunFailingTestsCount=${CCI_RERUN_FAILTEST:-0} \
```

j-white: This one was purposely set to `0`. This script is meant to be invoked manually to test local changes - we want to expose flappers here.

PR #2691, opened 2019-09-11
NMS-12275: Fixed usage description of vmwarereqtool utility

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-12275


iojeda71:
This fix also needs to be applied to vmwareconfigbuilder as it complains for the same missing library

PR #2700, opened 2019-09-13
NMS-12120: Add health wrapper for health:check

indigo423:
### All Contributors

Using the Karaf client directly with the health check in Docker and Kubernetes doesn't work with the way the client is implemented. The output can not be correctly fetched and evaluated. A workaround would be using an SSH client and passing the credentials to the Karaf shell. This would be a) requires to install sshpass and ssh client packages in the container and b) exposing admin credentials in the health check command in Docker compose or Kubernetes.

This fix resolves the problem by creating a /tmp/file from the client command and evaluate it separately without the disadvantages described in a) and b).

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12120

### DIFF
--- a/opennms-container/minion/Dockerfile
+++ b/opennms-container/minion/Dockerfile
@@ -33,8 +33,7 @@ RUN groupadd -g 10001 minion && \
 # Create SSH Key-Pair to use with the Karaf Shell
 RUN mkdir /opt/minion/.ssh && \
     chmod 700 /opt/minion/.ssh && \
-    ssh-keygen -t rsa -f /opt/minion/.ssh/id_rsa -q -N "" && \
-    chmod 600 /opt/minion/.ssh/id_rsa
+    ssh-keygen -t rsa -f /opt/minion/.ssh/id_rsa -q -N ""
 
 ##
 # Install and setup OpenNMS Minion RPMS
@@ -64,7 +63,8 @@ RUN if [[ "$(ls -1 /tmp/rpms/*.rpm 2>/dev/null | wc -l)" != 0 ]]; then dnf -y in
     echo "_g_\\:admingroup = group,admin,manager,viewer,systembundles,ssh" >> /opt/minion/etc/keys.properties && \
     chown 10001:10001 -R /opt/minion && \
     chgrp -R 0 /opt/minion && \
-    chmod -R g=u /opt/minion
+    chmod -R g=u /opt/minion && \
+    chmod 600 /opt/minion/.ssh/id_rsa
 
 COPY ./assets/* /
 
--- a/opennms-container/minion/Dockerfile
+++ b/opennms-container/minion/Dockerfile
@@ -6,7 +6,7 @@ ARG BASE_IMAGE_VERSION="11"
 
 FROM ${BASE_IMAGE}:${BASE_IMAGE_VERSION} as minion-base
 
-ARG REQUIRED_RPMS="wget gettext jicmp jicmp6"
+ARG REQUIRED_RPMS="wget gettext jicmp jicmp6 openssh-clients"
 
 ARG REPO_KEY_URL="https://yum.opennms.org/OPENNMS-GPG-KEY"
 ARG REPO_RPM="https://yum.opennms.org/repofiles/opennms-repo-stable-rhel8.noarch.rpm"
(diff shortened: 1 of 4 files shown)


agalue:
Excellent!

I believe the way to test it with Kubernetes should be:
```yaml
livenessProbe:
    exec:
        command:
            - bash
            - -c
            - /health.sh
        initialDelaySeconds: 60
        periodSeconds: 60
        timeoutSeconds: 20
```
Is this correct?

indigo423:
> Excellent!
> 
> I believe the way to test it with Kubernetes should be:
> 
> ```yaml
> livenessProbe:
>     exec:
>         command:
>             - bash
>             - -c
>             - /health.sh
>         initialDelaySeconds: 60
>         periodSeconds: 60
>         timeoutSeconds: 20
> ```
> 
> Is this correct?

Yes

agalue:
I found an alternative solution based on https://karaf.apache.org/manual/latest/security

The idea is to modify the Dockerfile to install `openssh-clients` to be able to have access to the `ssh` command, then create a key-pair using the `ssh-keygen` command, then update `$MINION_HOME/etc/keys.properties` with the generated public key for the `minion` user and associate the `viewer` role as we just need to execute the `health:check`, and finally enjoy a secured password-less solution in the form of:
```bash
ssh -p 8201 localhost health:check
```
That applies to `sentinel` as well.
If that sounds good, I can either re-use this PR or build a new one. Let me know your thoughts.
This might not be ready for 25.1.0, but I hope we can re-build and re-push the image after the solution is merged.

agalue:
FYI, I tested the proposed solution and it works flawlessly.

indigo423:
@agalue I have applied your patch using SSH with pub/key authentication to run the health check via SSH against Karaf.

agalue:
@indigo423 Thank you! Although, yesterday I found an extra change we should add. Due to the last command (i.e. `chmod -R g=u /opt/minion`), the permissions on `/opt/minion/.ssh/id_rsa.pub` will be changed and therefore, SSH will reject any authentication attempt due to bad permissions.

So, we should make the following the last line:
```
chmod 600 /opt/minion/.ssh/id_rsa
```
```

indigo423:
> Like with `minion`, this is incorrect, as `sshpass` is no longer necessary, and it is unavailable, it should be:
> 
> ```
> if ssh -o StrictHostKeyChecking=no -p 8301 localhost health:check | grep --quiet "Everything is awesome"; then
> ```
> 
> Also note the port ;)

👍 Copied from Minion :) Thank you for watching!

agalue:
No worries but do not forget to remove `sshpass` and fix the permissions on `/opt/minion/.ssh/id_rsa` as the last entry. I have my K8s lab up and running with those changes, which is why I know they are mandatory.

indigo423:
> No worries but do not forget to remove `sshpass` and fix the permissions on `/opt/minion/.ssh/id_rsa` as the last entry. I have my K8s lab up and running with those changes, which is why I know they are mandatory.

do you mean by fix permissions this here:
https://github.com/OpenNMS/opennms/pull/2700/files#diff-d3e81309706a55112b49fe7bfee8ee6bR35
* https://github.com/OpenNMS/opennms/pull/2700/files#diff-531442c857e9e911efcfd79aaa37c815R35

What do you mean with remove `sshpass`?

agalue:
The `health.sh` is using `sshpass`, but that package is not installed, and thanks to using certificates it is no longer needed, so the `health.sh` must be fixed as I outlined to avoid `command not found` errors.

The "current" last instruction will change the permissions on the public key form 600 to 640 which will be rejected by SSH.

As mentioned, my K8s lab is currently running (I can offer the KUBE_CONFIG if you want to take a look), and this is the change I've applied to my custom images (using Minion as an example, but also applies to Sentinel):

https://github.com/OpenNMS/opennms-drift-kubernetes/blob/master/docker-images/minion/Dockerfile#L38

Without that entry, the health check will fail because the SSH authentication will be rejected (bad permissions on the public key).

Makes sense?

indigo423:
@agalue Thanks for feedback, updated the branch related to your comments.

agalue:
Everything looks perfect, thank you! 👍

PR #2702, opened 2019-09-16
NMS-12283: Be able to execute ICMP requests through Java when running as non-root

indigo423:
### All Contributors

Added the network capabilities to the Java process similar as in the Horizon image to allow sending ICMP packets with Java. Additionally, the OpenJDK version is bumped to the latest available version 11.0.4.11 in the CentOS 7 repositories. We have also removed `gettext` and added `jq` which was a request to make things easier from @agalue. The 3 changes are separate commits so they can be reverted individually if necessary.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12283

### DIFF
--- a/opennms-container/horizon/Dockerfile
+++ b/opennms-container/horizon/Dockerfile
@@ -6,7 +6,7 @@ ARG BASE_IMAGE_VERSION="11"
 
 FROM ${BASE_IMAGE}:${BASE_IMAGE_VERSION} as horizon-base
 
-ARG REQUIRED_RPMS="rrdtool jrrd2 jicmp jicmp6 R-core rsync perl-XML-Twig perl-libwww-perl gettext"
+ARG REQUIRED_RPMS="rrdtool jrrd2 jicmp jicmp6 R-core rsync perl-XML-Twig perl-libwww-perl jq"
 
 ARG REPO_KEY_URL="https://yum.opennms.org/OPENNMS-GPG-KEY"
 ARG REPO_RPM="https://yum.opennms.org/repofiles/opennms-repo-stable-rhel7.noarch.rpm"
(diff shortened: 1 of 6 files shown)


agalue:
LGTM, thanks!

agalue:
Although, the title should say "be able to perform ICMP through Java when running as non-root", because the non-root nature is already present.

agalue:
Unfortunately, net.ipv4.ping_group_range requires using sysctl which only runs when you have root privileges. This is forbidden in strict environments like OpenShift.

Also, that flag works when the host (the machine that runs Docker/Kubelet) has a fairly recent kernel. In other words, that option won't work on RHEL/CentOS 7 or older. So, for OpenShift you would need CoreOS, which is the only option without going to OpenShift 4 with RHEL 8 to be able to use that privileged operation with OpenShift 3. Besides that, this operation has to be done by the sysadmins at OS level, as OpenShift itself won't allow you to do it through YAML or the UI for obvious reasons.

A similar situiation will happen with plain Docker, Swarm or Kubernets on restricted environments.

We have customers that are not allowed to run applications as root. While we help them deploying OpenNMS in their restricted environments, we found that the best and most elegant solution that doesn't impose kernel restrictions (in this customers case, only RHEL/CentOS 7 is allowed) to perform ICMP requests from within the JVM as a non-privileged user was using the netcap tool over the java binary.

In other words, the netcap solution for the java binary is well tested and verified with multiple customers regardless of how OpenNMS or Minion are running (containerized or bare-metal).

My $.02

Naicisum:
Apologies, the echo command works on OpenShift when use in the Dockerfile and building the image. I have validated that this indeed does work on RHEL7 inside Openshift.

indigo423:
Thank you @agalue and @Naicisum for taking the time to look into these changes! 🚀

PR #2709, opened 2019-09-17
NMS-12185: Ignore SFlow messages without IP data

fooker:
SFlow is capable of sending samples about other network layers beside
IP. To make a usefull flow out of the sample, the IP data is required.
Flows not containing IP related information are silently ignored now.

JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12185


fooker:
This should be cherry-picked into `foundation-2019`, too.

PR #2714, opened 2019-09-18
NMS-8294: Run Scriptd event scripts in a Hibernate transaction

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-8294

### DIFF
--- a/opennms-services/src/main/java/org/opennms/netmgt/scriptd/Executor.java
+++ b/opennms-services/src/main/java/org/opennms/netmgt/scriptd/Executor.java
@@ -49,6 +49,7 @@ import org.opennms.netmgt.config.scriptd.StartScript;
 import org.opennms.netmgt.config.scriptd.StopScript;
 import org.opennms.netmgt.config.scriptd.Uei;
 import org.opennms.netmgt.dao.api.NodeDao;
+import org.opennms.netmgt.dao.api.SessionUtils;
 import org.opennms.netmgt.events.api.EventConstants;
 import org.opennms.netmgt.model.OnmsNode;
 import org.opennms.netmgt.xml.event.Event;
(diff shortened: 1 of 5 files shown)


cgorantla:
Updated code to run scripts in transaction by default and an option to disable if desired.

opennms-services/src/main/java/org/opennms/netmgt/scriptd/Executor.java
```diff
@@ -202,87 +209,102 @@ public void run() {
                 }
             }
 
-            Script[] attachedScripts = m_event.getScript();
+            if (m_config.getTransactional()) {
+                m_sessionUtils.withReadOnlyTransaction(() -> {
```
j-white: @agalue any opinion on whether we make this read-only or not? Read-only will have less of an impact on the system but this mean that you'll continue to have to manage your own transaction if you want to create or update any database entities.

j-white:
Looks good. Just need to make sure we want to use a read-only transaction here.

opennms-services/src/main/java/org/opennms/netmgt/scriptd/Executor.java
agalue: I know for a fact that there are customers that need to modify nodes from within Scriptd. They are currently doing that within a transaction using Groovy, meaning that the `read-only` nature should not affect them if applied.

For this reason, I think it makes sense to enforce `read-only`, as most of the times that object should just provide more context while processing the event.

j-white: Thanks. We can keep it to read-only then as this will make sure that the LazyInitExceptions no longer occur. Changing objects will require the script developer to manage their own lookups and transactions though (which is the current case).

PR #2716, opened 2019-09-19
NMS-12298: Expose health via rest

mvrueden:
Here we expose the health of the system via a rest endpoint.

There are 2 options to get the health of the system:
 - `GET /rest/health/overview`: Returns `Everything is awesome` or `Oh no, something went wrong`. Requires no authentication.
 - `GET /rest/health`: Returns a details json object, similar to the result of the `health:check` command. Requires an authenticated user.

Even if the system health is not okay, the http response code is always `200 OK`. 

It would be nice to have the health of the system also exposed for minion and sentinel, but as of now we don't expose any ReST services there, so that is a bit more work which wasn't done in this PR.

JIRA: https://issues.opennms.org/browse/NMS-12298


agalue:
This is a great addition, thank you for working on this,

However, the way Kubernetes expects an HTTP health check to work (or readiness/liveness probes as they are called) is that an HTTP 200 indicates "everything is awesome", and something different (400 for instance), something is not healthy.

Here is Kubernetes documentation about it:

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#define-a-liveness-http-request

To be more precise, here is the code:

https://github.com/kubernetes/kubernetes/blob/master/pkg/probe/http/http.go#L123

For this reason, I believe it worth changing the code to return an error code when everything is not awesome.

indigo423:
Just to clarify, the code you described and the link to the documentation evaluates the get request from 200 to 399.

> Any code greater than or equal to 200 and less than 400 indicates success. Any other code indicates failure.

That means even the health check itself is failing the response must be 200 cause a) the request was successfully processed and the result needs to be evaluated in the custom header described in the documentation. 400 for example needs to be reserved for the case you send a bad request to the health check rest service.

mvrueden:
We also expose a header called 'Health', which is either `Everything is awesome` or `Oh no, something went wrong`. So looking at the documentation you can still do that, even if a 200 OK was returned.

The reason why I don't like to change the status code is, that 400 means `Bad Request`, but the request wasn't bad, the health is. Also any other http code does not really work.

If there is absoluetly no other way of determining of the prope is healthy or unhealthy besides a different status code, e.g. 400, I can change it, but looking at the go code you showed, it just shows, that the request for the probe was successful, but not the evaluation of the result.

agalue:
HTTP 400 was just an example to illustrate something different than HTTP 200. It doesn't have to be that precise code.

If the actual code is not required (which contradicts K8s documentation and the extensive tests I've been doing with this technology), I'd love to see a working liveness probe that validates the output instead of the HTTP returning code with Kubernetes.

agalue:
Just to clarify a bit about compliance and elegance in terms of what's expected to use with Kubernetes.

Right now, I would have to write the liveness probe like this:

```yaml
livenessProbe:
    exec:
        command:
            - sh
            - -c
            - 'curl http://localhost:8980/opennms/rest//health/overview | grep --quiet "Everything is awesome"; then exit 0; else exit 1; fi'
        initialDelaySeconds: 60
        periodSeconds: 60
        timeoutSeconds: 20
```

That works, but it is overcomplicated for no reason.

What I was trying to explain is write the HTTP health check properly (using HTTP return codes besides what you wish to return with it), so I can use the predefined well-known and widely used http probe like the following:

```yaml
livenessProbe:
          httpGet:
            path: /opennms/rest/health
            port: http
        initialDelaySeconds: 60
        periodSeconds: 60
        timeoutSeconds: 20
```

Makes sense?

Of course, that should be an unauthenticated end-point for obvious reasons.

j-white:
I agree that being able to use `httpGet` directly would be ideal, and for this to work we need to return a status code >= 400 on failure. How about we return a 500 Internal Server Error on failure?

As a compromise, we could also look at changing the behavior based on the presence of an HTTP header (which `httpGet` can set) or the look for the `kube-probe` user agent.

indigo423:
On startup I got this error message here:

```
 	at org.apache.karaf.features.internal.service.FeaturesServiceImpl.doProvision(FeaturesServiceImpl.java:1062)
	at org.apache.karaf.features.internal.service.FeaturesServiceImpl.lambda$doProvisionInThread$13(FeaturesServiceImpl.java:998)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
	Suppressed: java.io.IOException: Error downloading mvn:org.opennms.core.health/org.opennms.core.health.rest/25.0.0-SNAPSHOT
		at org.apache.karaf.features.internal.download.impl.AbstractRetryableDownloadTask.run(AbstractRetryableDownloadTask.java:77)
		at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
		at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
		at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
		... 3 more
	Caused by: java.io.IOException: Error resolving artifact org.opennms.core.health:org.opennms.core.health.rest:jar:25.0.0-SNAPSHOT: [Could not find artifact org.opennms.core.health:org.opennms.core.health.rest:jar:25.0.0-SNAPSHOT]
		at org.ops4j.pax.url.mvn.internal.AetherBasedResolver.configureIOException(AetherBasedResolver.java:803)
		at org.ops4j.pax.url.mvn.internal.AetherBasedResolver.resolve(AetherBasedResolver.java:774)
		at org.ops4j.pax.url.mvn.internal.AetherBasedResolver.resolve(AetherBasedResolver.java:657)
		at org.ops4j.pax.url.mvn.internal.AetherBasedResolver.resolve(AetherBasedResolver.java:598)
		at org.ops4j.pax.url.mvn.internal.AetherBasedResolver.resolve(AetherBasedResolver.java:565)
		at org.apache.karaf.features.internal.download.impl.MavenDownloadTask.download(MavenDownloadTask.java:52)
		at org.apache.karaf.features.internal.download.impl.AbstractRetryableDownloadTask.run(AbstractRetryableDownloadTask.java:60)
		... 6 more
		Suppressed: shaded.org.eclipse.aether.transfer.ArtifactNotFoundException: Could not find artifact org.opennms.core.health:org.opennms.core.health.rest:jar:25.0.0-SNAPSHOT
			at shaded.org.eclipse.aether.internal.impl.DefaultArtifactResolver.resolve(DefaultArtifactResolver.java:403)
			at shaded.org.eclipse.aether.internal.impl.DefaultArtifactResolver.resolveArtifacts(DefaultArtifactResolver.java:215)
			at shaded.org.eclipse.aether.internal.impl.DefaultArtifactResolver.resolveArtifact(DefaultArtifactResolver.java:192)
			at shaded.org.eclipse.aether.internal.impl.DefaultRepositorySystem.resolveArtifact(DefaultRepositorySystem.java:247)
			at org.ops4j.pax.url.mvn.internal.AetherBasedResolver.resolve(AetherBasedResolver.java:767)
			... 11 more
	Caused by: shaded.org.eclipse.aether.resolution.ArtifactResolutionException: Error resolving artifact org.opennms.core.health:org.opennms.core.health.rest:jar:25.0.0-SNAPSHOT
		at shaded.org.eclipse.aether.internal.impl.DefaultArtifactResolver.resolve(DefaultArtifactResolver.java:413)
		at shaded.org.eclipse.aether.internal.impl.DefaultArtifactResolver.resolveArtifacts(DefaultArtifactResolver.java:215)
		at shaded.org.eclipse.aether.internal.impl.DefaultArtifactResolver.resolveArtifact(DefaultArtifactResolver.java:192)
		at shaded.org.eclipse.aether.internal.impl.DefaultRepositorySystem.resolveArtifact(DefaultRepositorySystem.java:247)
		at org.ops4j.pax.url.mvn.internal.AetherBasedResolver.resolve(AetherBasedResolver.java:767)
		... 11 more 
```

indigo423:
I've used the container image artefact from the [last successful built](https://circleci.com/gh/OpenNMS/opennms/11250#artifacts/container).

mvrueden:
I agree, we should use the `httpGet` as well. Originally I misunderstood the documentation. I thought with the `httpHeaders` one can use those settings to expect a certain header value to consider the "probe" successful. However it is the header you want to send when making the request 🤦‍♂ 

I changed the health check endpoint to `/rest/health/probe/` instead of `/rest/health/overview` as there is no overview, just an OK, NOK response. Besides that in case of an error `500` will be returned. Not ideal but probably the most "sane" alternative.

mvrueden:
Also the issue @indigo423 saw, should be fixed.

opennms-webapp/src/main/webapp/WEB-INF/applicationContext-spring-security.xml
```diff
@@ -12,6 +12,9 @@
     <!-- OPTIONS pre-flight requests should always be accepted -->
     <intercept-url pattern="/rest/**" method="OPTIONS" access="ROLE_ANONYMOUS,ROLE_REST,ROLE_ADMIN,ROLE_USER,ROLE_MOBILE"/>
 
+    <!-- Calls to health/overview are always accepted -->
+    <intercept-url pattern="/rest/health/overview" method="GET" access="ROLE_ANONYMOUS,ROLE_REST,ROLE_ADMIN,ROLE_USER,ROLE_MOBILE"/>
```
mvrueden: should be `/rest/health/probe` 🤦‍♂

Naicisum:
Can we create a parameter to make the check for OpenNMS rest api optional for minion/sentinel? Either that or take it out completely as there is no dependency on needing core up.

mvrueden:
That is out of the scope of the current issue. Create https://issues.opennms.org/browse/NMS-12305 to reflect the request.

PR #2717, opened 2019-09-19
NMS-12272: Fixed Metadata ReST API

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-12272

### DIFF
--- a/opennms-model/src/main/java/org/opennms/netmgt/model/OnmsMetaData.java
+++ b/opennms-model/src/main/java/org/opennms/netmgt/model/OnmsMetaData.java
@@ -33,9 +33,14 @@ import java.util.Objects;
 
 import javax.persistence.Column;
 import javax.persistence.Embeddable;
+import javax.xml.bind.annotation.XmlElement;
 import javax.xml.bind.annotation.XmlRootElement;
 
+import org.codehaus.jackson.annotate.JsonProperty;
+import org.codehaus.jackson.map.annotate.JsonRootName;
+
 @XmlRootElement(name="meta-data")
+@JsonRootName("metaData")
 @Embeddable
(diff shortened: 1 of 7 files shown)


agalue:
Thank you for working on this.

I'm testing your changes, and I found the following:

* The `/api/v2/nodes/1/metadata` works and provides information about metadata in JSON as default (which is consistent with the rest of the end-points).

* The `/api/v2/nodes/1/metadata/ipinterfaces/127.0.0.1/metadata` works, but it provides the information in XML by default (including the nested paths for context and name), which is inconsistent with the behavior of other endpoints. The good news is that the content make sense, even after forcing the `Accept: application/json` header.

* The `/api/v2/nodes/1/metadata/ipinterfaces/127.0.0.1/services/OpenNMS-JVM/metadata` suffers the same problem as the IP interfaces.

To provide more context, here is the test requisition I've used:

```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<model-import xmlns="http://xmlns.opennms.org/xsd/config/model-import" date-stamp="2019-09-19T12:11:51.255-04:00" foreign-source="Local" last-import="2019-09-19T12:12:01.623-04:00">
  <node foreign-id="opennms-server" node-label="opennms-server">
    <interface ip-addr="127.0.0.1" status="1" snmp-primary="P">
      <monitored-service service-name="OpenNMS-JVM">
        <meta-data context="requisition" key="kind" value="JMX"/>
      </monitored-service>
      <meta-data context="requisition" key="type" value="loopback"/>
    </interface>
    <meta-data context="requisition" key="owner" value="agalue"/>
  </node>
</model-import>
```

christianpape:
You are right. I fixed that for ipInterfaces and services.

PR #2724, opened 2019-09-22
NMS-9827: Event parameters no longer preserve ordering

patrick-schweizer:
Open issues:
* [x] check if migration from release 20 preserves order of parameters: test was ok

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-9827

### DIFF
--- a/features/events/api/src/main/java/org/opennms/netmgt/events/api/EventParameterUtils.java
+++ b/features/events/api/src/main/java/org/opennms/netmgt/events/api/EventParameterUtils.java
@@ -29,8 +29,12 @@
 package org.opennms.netmgt.events.api;
 
 import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
@@ -156,5 +160,21 @@ public abstract class EventParameterUtils {
 
     public static Map<String, Parm> normalize(final List<Parm> parmList) {
         return parmList.stream().collect(Collectors.toMap(Parm::getParmName, Function.identity(), (p1, p2) -> p2));
+    }
+
+    public static List<Parm> normalizePreserveOrder(final List<Parm> parmList) {
+        Objects.requireNonNull(parmList);
+        Set<String> existingNames = new HashSet<>();
+        LinkedList<Parm> filteredList = new LinkedList<>();
+        // go through the list backwards since we want to keep the last element of duplicates names
+        // (existing behaviour)
+        for(int i = parmList.size() -1 ; i >= 0; i--) {
+            Parm parm = parmList.get(i);
+            if(!existingNames.contains(parm.getParmName())){
+                filteredList.addFirst(parm);
+                existingNames.add(parm.getParmName());
+            }
+        }
+        return filteredList;
     }
 }
--- a/opennms-base-assembly/src/main/filtered/bin/send-event.pl
+++ b/opennms-base-assembly/src/main/filtered/bin/send-event.pl
@@ -56,7 +56,7 @@ if ($version)  { print "$0 version $VERSION\n"; exit; }
 if ($help)     { print get_help(); exit; }
 
 # parm array is numerically referenced in OpenNMS' templates
-@PARMS = reverse map { parse_parm($_) } @PARMS;
+@PARMS = map { parse_parm($_) } @PARMS;
 
 my $hostname = hostname;
 
--- a/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/EventRestService.java
+++ b/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/EventRestService.java
@@ -99,6 +99,8 @@ public class EventRestService extends OnmsRestService {
         if (e == null) {
             throw getException(Status.NOT_FOUND, "Event object {} was not found.", Integer.toString(eventId));
         }
+        // Make sure we expose the params sorted:
+        e.setEventParameters(e.getEventParametersInOrder());
         return e;
     }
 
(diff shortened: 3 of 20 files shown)


features/events/api/src/main/java/org/opennms/netmgt/events/api/EventParameterUtils.java
```diff
@@ -157,4 +161,20 @@ public static String format(Parm parm) {
     public static Map<String, Parm> normalize(final List<Parm> parmList) {
         return parmList.stream().collect(Collectors.toMap(Parm::getParmName, Function.identity(), (p1, p2) -> p2));
     }
+
+    public static List<Parm> normalizePreserveOrder(final List<Parm> parmList) {
+        Objects.requireNonNull(parmList);
+        Set<String> existingNames = new HashSet<>();
+        LinkedList<Parm> filteredList = new LinkedList<>();
+        // go through the list backwards since we want to keep the last element of duplicates names
+        // (existing behaviour)
```
patrick-schweizer: @j-white do we want to preserve the existing behavior (keep the last element of duplicates names). If not we can simplify this method and pick the first duplicate parameter instead of the last.

opennms-base-assembly/src/main/filtered/bin/send-event.pl
```diff
@@ -56,7 +56,7 @@
 if ($help)     { print get_help(); exit; }
 
 # parm array is numerically referenced in OpenNMS' templates
-@PARMS = reverse map { parse_parm($_) } @PARMS;
+@PARMS = map { parse_parm($_) } @PARMS;
```
patrick-schweizer: @j-white I am not sure why this code reverts the order of the parameters that we send. It will mean that we will save them in reverse order. I think it should be
```@PARMS = map { parse_parm($_) } @PARMS;``` instead of ```@PARMS = reverse map { parse_parm($_) } @PARMS;```

File: opennms-webapp-rest/src/test/resources/v1/alarms.json [hunk 3]
```diff
@@ -21,7 +21,7 @@
                 "nodeId": 1, 
                 "nodeLabel": "node1", 
                 "parameters": [
-                    {"name": "testParm", "value": "HelloWorld", "type":"string" }
+                    {"name": "testParm", "value": "HelloWorld", "type":"string", "position": 0}
```
patrick-schweizer: @j-white the api exposes the database model (OnmsEvent) via rest and thus we have the new field "position" in the response. Are we ok with that or do we need to do something about it?

features/events/api/src/main/java/org/opennms/netmgt/events/api/EventParameterUtils.java
j-white: Let's preserve the existing behavior.

opennms-base-assembly/src/main/filtered/bin/send-event.pl
j-white: Not sure either,  looks like it's been that way for > 12 years though.

Keeping the order specified by the user makes sense to me.

opennms-webapp-rest/src/test/resources/v1/alarms.json
j-white: I would hide it. Marking the attribute with `@XmlTransient` instead of `@XmlAttribute` should do the trick.

features/events/api/src/main/java/org/opennms/netmgt/events/api/EventParameterUtils.java
patrick-schweizer: ok.

opennms-base-assembly/src/main/filtered/bin/send-event.pl
patrick-schweizer: ok, then let's leave it with my change.

opennms-webapp-rest/src/test/resources/v1/alarms.json
patrick-schweizer: :+1: makes sense

patrick-schweizer:
@j-white do you mind checking if I found all places where we expose an event and if you are happy with the solution? If so I would squash all commits since I tried now lot's of different stuff in this PR.

patrick-schweizer:
Please squash commits (instead of just merge)

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/EventRestService.java
```diff
@@ -99,6 +99,8 @@ public OnmsEvent getEvent(@PathParam("eventId") final Integer eventId) {
         if (e == null) {
             throw getException(Status.NOT_FOUND, "Event object {} was not found.", Integer.toString(eventId));
         }
+        // Make sure we expose the params sorted:
+        e.setEventParameters(e.getEventParametersInOrder());
```
mvrueden: Isn't the `EventDao` the better place for this, as it ensures that ALL calls to load events from the database will have the correct order applied?

patrick-schweizer: Ideally we would have one central place where we could ensure the ordering of the event parameters. The ideal place would be OnmsEvent itself which I tried for a while but I couldn't get it to work for various reasons (see https://issues.opennms.org/browse/NMS-9827?focusedCommentId=56051&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-56051).
Other than that I believe there is no other central place where we can ensure the ordering. We can either use
- the DAO level (as you suggested ) which I think would be OnmsEventDaoHibernate, AlarmDaoHibernate, NotificationDaoHibernate, OutageDaoHibernate or
- on the API level
I went for the API level since this seemed less invasive and we order on demand (on those API calls where it is relevant). I fear the API level will be more complicated. But I have no strong opinion on this. I can do either way, whatever is preferred. @mvrueden what do you think?

mvrueden: I just don't understand why we need a new `getEventParametersInOrder()` method. 
We could just implement the same logic in `getEventParameters()` or am I missing something?

patrick-schweizer: closing this PR in favor of: https://github.com/OpenNMS/opennms/pull/2777

PR #2725, opened 2019-09-23
NMS-10073: Move Telephone PIN to Notification  Information

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-10073


cgorantla:
I verified that storing something in Telephone PIN and clicking Finish would save it and I can retrieve it later.

PR #2732, opened 2019-09-24
fix up runtest script for my environment  :)

RangerRick:
This PR contains fixes for running on my mac (which doesn't have `/usr/bin/bash`) and for building offline without having run tests before.


j-white:
Looks reasonable 👍

PR #2736, opened 2019-09-26
NMS-12104: Ensure order in flow queries

fooker:
This adds a test which ensures the order of flow series and summaries is
correctly sorted.

In addition, this makes some clean-ups to the used datastructures which
makes it easier for testing.

JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12104


fooker:
@mattixtech The original issue was not clear about whether the problem was in series or summaries but both seems to be ordered correctly with the current code.

I've verified that the results of the queries are in the wrong order and get sorted by the topN-list afterwards.

mattixtech:
@fooker I stepped through the code and you are right, it does appear that the ordering is working correctly. Not sure how I could have got that incorrect result in the first place... Looks good.

PR #2739, opened 2019-09-28
NMS-10582: fix problem with not writable map

patrick-schweizer:
Jira: https://issues.opennms.org/browse/NMS-10582

### DIFF
--- a/features/enlinkd/adapters/collectors/bridge/src/main/java/org/opennms/netmgt/enlinkd/NodeDiscoveryBridge.java
+++ b/features/enlinkd/adapters/collectors/bridge/src/main/java/org/opennms/netmgt/enlinkd/NodeDiscoveryBridge.java
@@ -102,7 +102,7 @@ public final class NodeDiscoveryBridge extends NodeCollector {
 
         SnmpAgentConfig peer = getSnmpAgentConfig();
         String community = peer.getReadCommunity();
-        Map<Integer, String> vlanmap = (m_disableBridgeVlanDiscovery) ? Collections.emptyMap() : getVtpVlanMap(peer);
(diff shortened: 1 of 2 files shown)


features/enlinkd/adapters/collectors/bridge/src/main/java/org/opennms/netmgt/enlinkd/NodeDiscoveryBridge.java
```diff
@@ -102,7 +102,7 @@ public void collect() {
 
         SnmpAgentConfig peer = getSnmpAgentConfig();
         String community = peer.getReadCommunity();
-        Map<Integer, String> vlanmap = (m_disableBridgeVlanDiscovery) ? Collections.emptyMap() : getVtpVlanMap(peer);
```
patrick-schweizer: @j-white the fix is easy - I missed that we are writing a null entry to the map :-(
I am not sure if / how we should test just in case we run into more problems?

j-white: Let's add a test this time. We can re-use an existing test that go down this code-path, and update it to set `disableBridgeVlanDiscovery=true`.

patrick-schweizer: @j-white, I added this test: https://github.com/OpenNMS/opennms/pull/2739/files#diff-6983d05f17a2f4b85883291113fc1f9a is that what you imagined?

PR #2743, opened 2019-09-30
NMS-10697: Recover from Connection pool shut down

mvrueden:
When elastic is not reachable and somehow a bundle refresh occurred on a bundle using the JestClient a "Connection pool shut down" exception is raised. This is caused by invoking "shutdownClient" on the client bean on stopping the bundle. This PR addresses this issue, by gracefully dropping the affected messages. We should look for a better solution in the long run, to not drop the messages.

This PR also fixes a package clash, which caused the JestClient objects to be created more than required.

While being at it, I also used `SessionUtils` instead of `TransactionOperations` in the `opennms-es-rest` module.

JIRA: https://issues.opennms.org/browse/NMS-10697


features/telemetry/protocols/flows/src/main/java/org/opennms/netmgt/telemetry/protocols/flows/AbstractFlowAdapter.java
```diff
@@ -116,6 +116,10 @@ public void handleMessageLog(TelemetryMessageLog messageLog) {
             }
         } catch (FlowException ex) {
             LOG.error("Error while persisting flows: {}", ex.getMessage(), ex);
+            if (ex.getMessage().equals("Connection pool shut down")) {
```
j-white: Is there a better way to catch this, or match the exception? Seems like this could break easily.

mvrueden: Added a dedicated exception to handle this more properly.

features/opennms-es-rest/src/main/java/org/opennms/plugins/elasticsearch/rest/EventForwarderQueueImpl.java
```diff
@@ -117,9 +122,13 @@ public EventIndexProducer(int batchSize, int batchInterval) {
...
-			// Ensure we are initialized correctly.
-			if (!elasticSearchInitializer.isInitialized()) {
-				elasticSearchInitializer.initialize(); // blocks until initialized properly
+			try {
+				// Ensure we are initialized correctly.
+				if (!elasticSearchInitializer.isInitialized()) {
+					elasticSearchInitializer.initialize(); // blocks until initialized properly
+				}
+			} catch (ConnectionPoolShutdownException ex) { // Connection Pool is gone, nothing we can do
+				ExceptionUtils.handle(getClass(), ex, events);
```
mvrueden: Should return afterwards. There is no need in forwarding the events afterwards, as that will fail as well.

mvrueden:
Retargeted to `foundation-2019`. Circle CI passed. Merging...

PR #2746, opened 2019-10-01
NMS-10578: Update severity for BSM alarms

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-10578


mvrueden:
Should be re-targeted for `foundation-2019` instead, though.

PR #2754, opened 2019-10-09
NMS-12288: Create events for invalidated BSM graphs

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-12288


features/bsm/daemon/src/main/java/org/opennms/netmgt/bsm/daemon/Bsmd.java
```diff
@@ -196,7 +203,70 @@ public void serviceInterfaceOrNodeDeleted(Event e) {
                 || EventConstants.SERVICE_DELETED_EVENT_UEI.equals(e.getUei()) && reductionKeys.contains(String.format("uei.opennms.org/nodes/nodeLostService::%d:%s:%s", e.getNodeid(), e.getInterface(), e.getService()))
                 || EventConstants.APPLICATION_DELETED_EVENT_UEI.equals(e.getUei()) && m_stateMachine.getGraph().getVertexByApplicationId(Integer.parseInt(e.getParm("applicationId").getValue().getContent())) != null) {
 
+            final Set<BusinessService> businessServices = m_stateMachine.getGraph().getVertices().stream().filter(v -> v.getBusinessService() != null).map(v -> v.getBusinessService()).collect(Collectors.toSet());
```
j-white: The BusinessServiceGraph already has the knowledge of `application id -> vertex` and `reduction key -> vertex`. I'd rather see us leverage that indexing instead of having to iterate and search the whole graph again here.

features/bsm/daemon/src/main/java/org/opennms/netmgt/bsm/daemon/Bsmd.java
```diff
@@ -196,7 +203,70 @@ public void serviceInterfaceOrNodeDeleted(Event e) {
...
+
+                if (EventConstants.SERVICE_DELETED_EVENT_UEI.equals(e.getUei())) {
+                    affectedBusinessServices.addAll(reductionKeyToBusinessService.get(String.format("uei.opennms.org/nodes/nodeLostService::%d:%s:%s", e.getNodeid(), e.getInterface(), e.getService())));
+                }
+            }
+
             reloadConfigurationAt = System.currentTimeMillis() + RELOAD_DELAY;
+
+            final EventBuilder eventBuilder = new EventBuilder(EventConstants.BUSINESS_SERVICE_GRAPH_INVALIDATED, "bsmd");
+            eventBuilder.addParam("businessServiceNames", affectedBusinessServices.stream().map(b -> String.valueOf(b.getName())).sorted().collect(Collectors.toList()));
```
j-white: Should we create one alarm per business service instead?

features/bsm/daemon/src/main/java/org/opennms/netmgt/bsm/daemon/Bsmd.java
christianpape: Yes, but I do not need the application or reduction key leaf node at this point. I need the business service that has the application edge or the reduction key edge attached.

features/bsm/daemon/src/main/java/org/opennms/netmgt/bsm/daemon/Bsmd.java
christianpape: What do you think makes more sense? I don't care about that. I can also generate an event for each affected business service... what do you think?

features/bsm/daemon/src/main/java/org/opennms/netmgt/bsm/daemon/Bsmd.java
j-white: We can traverse the graph from there. My main concern here is performance, these searches will take a lot of cycles on large graphs.

features/bsm/daemon/src/main/java/org/opennms/netmgt/bsm/daemon/Bsmd.java
j-white: What led me here was looking at the reduction key for the alarm associated with the event, seems like it would make more sense to de-duplicate these events into alarms based on individual business services, as opposed to sets of business services.

features/bsm/daemon/src/main/java/org/opennms/netmgt/bsm/daemon/Bsmd.java
christianpape: Since the graph is built from top-bottom, I only have one way to go down based on the business services to see which BS definitions are really affected. I can't get the business services for one leaf that are direct predecessors. Or am I missing something?

j-white: We'll need to extend some interfaces to grab it from this context, but the complete graph is available. The state propagation goes upward, so it's possible to find the next affected BS from the leaf vertices.

christianpape: Ok. I see. I will change this...

features/bsm/daemon/src/main/java/org/opennms/netmgt/bsm/daemon/Bsmd.java
christianpape: Ok

mfuhrmann: I think per BSM would be better, since you can use the alarms as some kind of todo list and process/fix each step by step. Also, you could have more teams/coworkers that are responsible for different BSM. The alarm acknowledging logic fits better then.

mfuhrmann: @christianpape Will test this PR asap.

features/bsm/daemon/src/main/java/org/opennms/netmgt/bsm/daemon/Bsmd.java
christianpape: done

mfuhrmann:
Looks like this when I delete a node:
grafik
I'm wondering why I got 3 events: 
grafik

I would like to see the name of the BSM and the deleted node in the messages.
Also I'm suggesting to change the wording:
Some thing like:

logmsg: `top1 business service is affected by the deletion of associated entities`
description: `Since Node XY was deleted and thus affects the functionality of BSM XY.`

And a link to the corresponding BSM config in the description would be nice to have.

Also I'm wondering what ` uei.opennms.org/bsm/graphInvalidated ` mean?  I never would associate this UEI to the fact that a node was deleted and broke a BSM. But maybe I'm too picky :-)

features/bsm/daemon/src/main/java/org/opennms/netmgt/bsm/daemon/Bsmd.java
christianpape: done

christianpape:
If you delete a node, you receive events for all services, interfaces and finally for the node itself. The code matches the corresponding reductionKeys and send out the events. Finally the events get reduced into one alarm.

christianpape:
I think leaf is not the right term since BSM has only one type of leafs, reduction keys. These are represented by strings and not deletable entities in the database. A BS, application, IP service can be deleted which represents in the BSM graph inner objects, right? So, my goal was to reflect this by the uei "uei.opennms.org/bsm/graphInvalidated", that the graph BSM is working on is no longer valid concerning the real entities in the DB.

mvrueden:
Yeah you are right. Leaf is a bit misleading as those are the parents of the leafs. Also the term orphan does not really match, as nothing is orphaned. It is more like a "missing child" which is different.

opennms-base-assembly/src/main/filtered/etc/events/opennms.bsm.events.xml
```diff
@@ -33,4 +33,14 @@
       <severity>Indeterminate</severity>
       <alarm-data reduction-key="uei.opennms.org/bsm/serviceProblem:%parm[businessServiceId]%" alarm-type="2" auto-clean="false"/>
    </event>
+    <event>
+        <uei>uei.opennms.org/bsm/graphInvalidated</uei>
+        <event-label>Business Service Monitoring: Graph invalidated</event-label>
+        <descr>&lt;p>Business service '%parm[businessServiceName]%' with ID '%parm[businessServiceId]%' is affected by the deletion of %parm[cause]%. A reload of the BMS daemon is scheduled.&lt;/p></descr>
+        <logmsg dest="logndisplay">Business service '%parm[businessServiceName]%' with ID '%parm[businessServiceId]%' is affected by the deletion of %parm[cause]%. A reload of the BMS daemon is scheduled.</logmsg>
```
j-white:
```suggestion
        <logmsg dest="logndisplay">Business service '%parm[businessServiceName]%' with ID '%parm[businessServiceId]%' is affected by the deletion of %parm[cause]%.</logmsg>
```

Keep the log message short and save the last part of the description

PR #2760, opened 2019-10-11
NMS-12302: Reset state on change

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-12302


mvrueden:
Looks good to me and works. Should probably add a smoke test for this, though (-:

j-white:
> Looks good to me and works. Should probably add a smoke test for this, though (-:

Not worth adding a smoke test IMO - we have most of these for BSM already disabled because of flapping issues.

mvrueden:
Ah yes, forgot about that (-:

@christianpape can you do a rebase instead of a merge in the future. The changes are so small that rebasing shouldn't be that hard and it gets rid of the merge commits.

PR #2765, opened 2019-10-15
NMS-9684: Introduce version compatibility list for Horizon 25.1.0

indigo423:
### All Contributors

Add a version compatibility list in the Admin guide which gives administrators the information with which components this specific version of OpenNMS Horizon is compatible with.

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-9684


opennms-doc/guide-admin/src/asciidoc/text/compatibility.adoc
```diff
@@ -0,0 +1,18 @@
...
+|===
+| Component               | Version Compatibility
+| OpenNMS Integration API | 0.2.1
+| Java Development Kit    | OpenJDK 8, OpenJDK 11
+| PostgreSQL              | 10.x and 11.x
+| Cassandra               | 3.11+
+| Newts                   | 1.5.2
+| Helm                    | 3+
+| Kafka                   | 2.3.0
+| Elasticsearch           | 6.2.4 until 7.3.0
```
RangerRick: Pretty sure we don't support Elasticsearch 6.x in Horizon 25.

indigo423:
> Also this should be targeted to `foundation-2019` rather than `develop`.

Done! Thanks good point :)

opennms-doc/guide-admin/src/asciidoc/text/compatibility.adoc
indigo423: What I know here so far, Elasticsearch version dependency is driven by our elasticsearch-drift-plugin which is installed on the Elasticsearch node. The first version was 1.0.0 which pointed to Elasticsearch 6.2.4 which is the reason I've picked it :) Now is the best chance to clarify this :)

j-white: H25 -> Only ES 7.x: http://docs.opennms.org/opennms/releases/25.0.0/releasenotes/releasenotes.html#_elasticsearch_7_x_support

opennms-doc/guide-admin/src/asciidoc/text/compatibility.adoc
```diff
@@ -0,0 +1,18 @@
...
+[options="header, autowidth"]
+|===
+| Component               | Version Compatibility
+| OpenNMS Integration API | 0.2.1
+| Java Development Kit    | OpenJDK 8, OpenJDK 11
+| PostgreSQL              | 10.x and 11.x
+| Cassandra               | 3.11+
+| Newts                   | 1.5.2
+| Helm                    | 3+
+| Kafka                   | 2.3.0
```
j-white: That's the version of the client library we use, but it's backwards compatible.

RangerRick:
I haven't really done much with it, but as I understand it, besides the plugin there is also code and schemas on the OpenNMS side that require Eliasticsearch 7 because it's difficult to support both simultaneously.  @j-white or @mattixtech could probably answer better.

RangerRick:
Oops, for some reason I didn't get an email about Jesse's comment, but yeah, ES7 only.  :)

opennms-doc/guide-admin/src/asciidoc/text/compatibility.adoc
indigo423: @j-white Yes, I've picked it up from the pom.xml. Should we write 1.x - 2.x here?

indigo423:
Do we follow the [semantic version](https://semver.org/) for our own components? If so, we could write here for our own components `<major version number>.x`, i.e. Helm, Newts, OIA? What about ALEC? Can we break compatibility with new ALEC versions which aren't compatible with Horizon Versions?

opennms-doc/guide-admin/src/asciidoc/text/compatibility.adoc
```diff
@@ -0,0 +1,18 @@
+[[compatibility]]
+== Compatibility
+
+{opennms-product-name} {opennms-version} requires the following component versions:
+
+[options="header, autowidth"]
+|===
+| Component               | Version Compatibility
+| OpenNMS Helm            | 3+
+| OpenNMS Integration API | 0.2.1
```
j-white:
```suggestion
| OpenNMS Integration API | 0.2.x
```

opennms-doc/guide-admin/src/asciidoc/text/compatibility.adoc
```diff
@@ -0,0 +1,18 @@
...
+| Newts                   | 1.5.2
```
j-white: Given that Newts is embedded, do we need to mention it here?

opennms-doc/guide-admin/src/asciidoc/text/compatibility.adoc
```diff
@@ -0,0 +1,18 @@
...
+| Cassandra               | 3.11.+
+| Elasticsearch           | 7.x - 7.3.0
```
j-white:
```suggestion
| Elasticsearch           | 7.x
```

opennms-doc/guide-admin/src/asciidoc/text/compatibility.adoc
j-white: Works for me.

indigo423:
Thank you for help.

indigo423:
@j-white @RangerRick anything we need to add here?

opennms-doc/guide-admin/src/asciidoc/text/compatibility.adoc
```diff
@@ -0,0 +1,17 @@
...
+[options="header, autowidth"]
+|===
+| Component               | Version Compatibility
+| OpenNMS Helm            | 3+
+| OpenNMS Integration API | 0.2.x
+| Cassandra               | 3.11.+
+| Elasticsearch           | 7.x
+| Java Development Kit    | OpenJDK 8, OpenJDK 11
+| Kafka                   | 2.3.0
+| PostgreSQL              | 10.x - 11.x
```
RangerRick: we technically allow 10-12, only recommend 11 because 12 had come out moments before the horizon release

indigo423:
@j-white @RangerRick I've fixed all comments related compatibility. I noticed two things:

1) I've moved it from the admin guide to the install guide, because this is the place where you need the information, what you need to install things. Any objections here?
2) What do you think about adding supported operating systems to the compatibility list?

Thank you in advance.

opennms-doc/guide-install/src/asciidoc/text/compatibility.adoc
```diff
@@ -0,0 +1,17 @@
...
+
+[options="header, autowidth"]
+|===
+| Component               | Version Compatibility
+| OpenNMS Helm            | 3+
+| OpenNMS Integration API | 0.2.x
+| Cassandra               | 3.11.+
+| Elasticsearch           | 7.x
+| Java Development Kit    | OpenJDK 8, OpenJDK 11
+| Kafka                   | 2.3.0
```
j-white:
```suggestion
| Kafka                   | 1.x - 2.x
```

j-white:
Looks good - just the Kafka version needs updating.

RangerRick:
>     1. I've moved it from the admin guide to the install guide, because this is the place where you need the information, what you need to install things. Any objections here?

Makes sense.

>     2. What do you think about adding supported operating systems to the compatibility list?

Couldn't hurt I guess, though since it's JVM-based there isn't that much that _isn't_ technically compatible, especially if you use all JNA-based stuff.  Not sure it's worth trying to nail down specifics.

indigo423:
> Looks good - just the Kafka version needs updating.

Fixed, sorry I haven't understood the conversation in the comments correctly :)

indigo423:
> > ```
> > 1. I've moved it from the admin guide to the install guide, because this is the place where you need the information, what you need to install things. Any objections here?
> > ```
> 
> Makes sense.
> 
> > ```
> > 2. What do you think about adding supported operating systems to the compatibility list?
> > ```
> 
> Couldn't hurt I guess, though since it's JVM-based there isn't that much that _isn't_ technically compatible, especially if you use all JNA-based stuff. Not sure it's worth trying to nail down specifics.

Ok we can skip this in the first place and think about it more, adding stuff later isn't a big deal.

PR #2768, opened 2019-10-16
NMS-12343: Handle duplicate messages from Kafka

cgorantla:
When `acks=1` there is a chance of duplicate messages in kafka.
For RPC, each message has an ID associated so it will only be processed once.
In case of large buffer, this duplicate chunk could produce invalid response, handle this by checking chunk number.

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12343


core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcClientFactory.java
```diff
@@ -459,16 +461,25 @@ public void run() {
...
                             // For larger messages which get split into multiple chunks, cache them until all of them arrive.
                             if (rpcMessage.getTotalChunks() > 1) {
-                                String rpcId = rpcMessage.getRpcId();
+                                // Avoid duplicate chunks. discard if chunk is repeated.
+                                chunkNumberMap.putIfAbsent(rpcId, 0);
+                                Integer chunkNum = chunkNumberMap.get(rpcId);
+                                if(chunkNum == rpcMessage.getCurrentChunkNumber()) {
+                                    chunkNumberMap.put(rpcId, ++chunkNum);
+                                } else {
+                                    continue;
```
j-white: Worth logging something here?

core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcClientFactory.java
```diff
@@ -486,9 +497,13 @@ public void run() {
...
                             responseHandlerExecutor.execute(() ->
                                     responseCb.sendResponse(rpcMessageContent));
+                            // Clear maps so that duplicate response will not be handled.
+                            rpcResponseMap.remove(rpcId);
+                            messageCache.remove(rpcId);
+                            chunkNumberMap.remove(rpcId);
                         } else {
-                            LOG.warn("Received a response for request with ID:{}, but no outstanding request was found with this id, The request may have timed out.",
-                                    record.key());
+                            LOG.debug("Received a response for request with ID:{}, but no outstanding request was found with this id, " +
```
j-white:
```suggestion
                            LOG.debug("Received a response for request with ID:{}, but no outstanding request was found with this id. " +
```

core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcClientFactory.java
```diff
@@ -486,9 +497,13 @@ public void run() {
...
+                                            "The request may have timed out or the response may be duplicate", record.key());
```
j-white:
```suggestion
                                            "The request may have timed out or the response may be a duplicate.", record.key());
```

core/ipc/sink/kafka/server/src/main/java/org/opennms/core/ipc/sink/kafka/server/KafkaMessageConsumerManager.java
```diff
@@ -181,8 +195,12 @@ public void run() {
                                 if (sinkMessage.getTotalChunks() > sinkMessage.getCurrentChunkNumber() + 1) {
                                     continue;
                                 }
-                                messageInBytes = largeMessageCache.getIfPresent(messageId).toByteArray();
+                                byteString = largeMessageCache.getIfPresent(messageId);
+                                if(byteString != null) {
```
j-white: If it is `null`, then the contents of `messageInBytes` aren't valid. Looks like there should be some additional logging / handling for that case here.

cgorantla:
> Nice catch, looks good overall.
> 
> Would be nice to be able to unit test this - can we find a way to mock the Kafka consumer?

I think we could mock KafkaServerManager and send duplicate messages to kafka. That may be easiest way to test the whole path.

cgorantla:
I have changed the logic of handling duplicates little bit so that we only process message if only all of the chunks are received.  
`RpcKafkaLargeBufferIT` will verify handling of this duplicate. The same logic is applied in Sink as well.

core/ipc/rpc/kafka/src/test/java/org/opennms/core/ipc/rpc/kafka/MockKafkaServerManager.java
```diff
@@ -0,0 +1,199 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.math.IntMath;
+import com.google.protobuf.ByteString;
+
+/**
+ * Mocking Kafka Server Manager to push duplicate messages to kafka.
+ */
+public class MockKafkaServerManager {
```
j-white: The `Mock` prefix is deceiving here since this class still talks to the actual Kafka broker.

Can we rework this so we don't have to copy so much of the existing `KafkaServerManager` code?

cgorantla: yeah, That was quickest I can think of. I will try to rewrite this test.

cgorantla:
Took this opportunity to make rpc code little bit organized and easy to test.

PR #2769, opened 2019-10-17
NMS-9352: Added root cause data to uei.opennms.org/bsm/serviceProblem events/alarms

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-9352

### DIFF
--- a/smoke-test/src/test/java/org/opennms/smoketest/bsm/BsmEventsIT.java
+++ b/smoke-test/src/test/java/org/opennms/smoketest/bsm/BsmEventsIT.java
@@ -33,7 +33,9 @@ import static java.util.concurrent.TimeUnit.MINUTES;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.hamcrest.Matchers.notNullValue;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 
+import org.elasticsearch.common.Strings;
 import org.junit.ClassRule;
 import org.junit.Test;
 import org.opennms.core.criteria.CriteriaBuilder;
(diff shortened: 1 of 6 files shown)


smoke-test/src/test/java/org/opennms/smoketest/bsm/BsmEventsIT.java
```diff
@@ -33,7 +33,9 @@
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.hamcrest.Matchers.notNullValue;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 
+import org.elasticsearch.common.Strings;
```
j-white: elasticsearch import?

smoke-test/src/test/java/org/opennms/smoketest/bsm/BsmEventsIT.java
```diff
@@ -76,6 +78,7 @@ public void testSeverityChange() throws Exception {
                         .toCriteria()), notNullValue());
 
         assertEquals(1L, (long) onmsAlarm1.getCounter());
+        assertFalse(Strings.isNullOrEmpty(onmsAlarm1.getEventParameter("rootCause")));
```
j-white:
```suggestion
        assertThat(onmsAlarm1.getEventParameter("rootCause"), not(isEmptyOrNullString());
```

smoke-test/src/test/java/org/opennms/smoketest/bsm/BsmEventsIT.java
```diff
@@ -86,6 +89,7 @@ public void testSeverityChange() throws Exception {
                         .toCriteria()), notNullValue());
 
         assertEquals(2L, (long) onmsAlarm2.getCounter());
+        assertFalse(Strings.isNullOrEmpty(onmsAlarm2.getEventParameter("rootCause")));
```
j-white:
```suggestion
        assertThat(onmsAlarm2.getEventParameter("rootCause"), not(isEmptyOrNullString());
```

j-white:
The updated BsmEventsIT appears to have passed on CircleCI, why did we remove those assertions?

christianpape:
That was a mistake in my thinking. In this test rootCause cannot be set at all, because the serviceProblem alarm is generated in the test and not by the Bsmd. Therefore the assertions made no sense at all.

christianpape:
I added a check to an existing method in BsmdIT.

christianpape:
Still waiting for approval... :-)

PR #2772, opened 2019-10-18
NMS-10346: default event description is incorrect

patrick-schweizer:
* Jira: https://issues.opennms.org/browse/NMS-10346


j-white:
Let's re-target this to `foundation-2019` - this patch is probably not worth the potential of introducing .rpmnew files on existing M2017 or M2018 installs.

patrick-schweizer:
> Let's re-target this to `foundation-2019` - this patch is probably not worth the potential of introducing .rpmnew files on existing M2017 or M2018 installs.

retarget to foundation-2019 -> done

PR #2774, opened 2019-10-19
NMS-12352: Fix R installation for CentOS 8

indigo423:
### All Contributors

Fix installation instructions to install R on CentOS 8.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12352

### DIFF
--- a/opennms-doc/guide-install/src/asciidoc/text/r-statistical-computing.adoc
+++ b/opennms-doc/guide-install/src/asciidoc/text/r-statistical-computing.adoc
@@ -8,7 +8,7 @@ _{opennms-product-name}_ interfaces with _R_ via _stdin_ and _stdout_, and for t
 as _{opennms-product-name}_.
 Note that installing _R_ is optional, and not required by any of the core components.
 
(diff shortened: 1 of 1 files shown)


agalue:
I tried all the above commands on a Docker instance based on CentOS 8 and the last one didn't work for me:
```bash
[root@6978839a5a18 /]# dnf install -y R
Failed to set locale, defaulting to C
Last metadata expiration check: 0:00:16 ago on Tue Oct 22 14:25:31 2019.
No match for argument: R
Error: Unable to find a match
```
I don't think a full VM would be required for it. Am I missing something?

agalue:
It might be a limitation of the Docker image `centos:8`, as with `Vagrant`, using `generic/centos8`, the commands work.

That said, if you want to save 100+ MB, why not using `R-core` instead of `R`?

indigo423:
@agalue thanks for testing, used the R-core package instead. We found a bug and have to install epel-release which has the R packages. The PowerTools repository packages to satisfy the R package dependencies.

PR #2777, opened 2019-10-22
NMS-9827: Event parameters no longer preserve ordering attributes: solution 2: JPA annotations via attributes

patrick-schweizer:
Alternative to https://github.com/OpenNMS/opennms/pull/2724

We try here to use the OnmsEvent attributes instead of the setter for the JPA/Hibernate annotation. Rational: Getter / Setter won't be used by hibernate anymore with its problems (see https://issues.opennms.org/browse/NMS-9827?focusedCommentId=56051&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-56051)

* JIra: https://issues.opennms.org/browse/NMS-9827

### DIFF
--- a/core/schema/src/main/liquibase/25.0.0/changelog.xml
+++ b/core/schema/src/main/liquibase/25.0.0/changelog.xml
@@ -123,5 +123,17 @@
             WHERE classification_groups.id = t2.id
         </sql>
     </changeSet>
+    <changeSet author="patrick-schweizer" id="25.0.0-add-position-column-to-event_parameters">
+        <addColumn tableName="event_parameters">
+        <column name="position"
+                type="integer"
+                defaultValueNumeric="0" />
 
+        </addColumn>
+        <rollback>
+            <dropColumn
+                    columnName="position"
+                    tableName="event_parameters"/>
+        </rollback>
+    </changeSet>
 </databaseChangeLog>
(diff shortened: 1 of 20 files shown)


patrick-schweizer:
@j-white, @mvrueden, this version seems to work.

I tested:
* sending events via: POSThttp://localhost:8980/opennms/rest/events (xml & json)
* retrieving events via:
** GET http://localhost:8980/opennms/rest/events?eventUei=uei.opennms.org/alarms/trigger (xml & json)
** GET http://localhost:8980/opennms/api/v2/events?eventUei=uei.opennms.org/alarms/trigger

Are we happy with it? I would prefer this solution over https://github.com/OpenNMS/opennms/pull/2724

RangerRick:
Just an FYI, the reason annotations for all the old model objects were on the methods was to preserve behavior backwards-compatibility with castor serialization, which means that integers would _always_ resolve to `0` (because they were originally `int`) rather than being able to return a `null` value.

This is _probably_ OK nowadays, but be aware there might be some slight changes in how things work in the publicly-facing XML side.

patrick-schweizer:
> Just an FYI, the reason annotations for all the old model objects were on the methods was to preserve behavior backwards-compatibility with castor serialization, which means that integers would _always_ resolve to `0` (because they were originally `int`) rather than being able to return a `null` value.
> 
> This is _probably_ OK nowadays, but be aware there might be some slight changes in how things work in the publicly-facing XML side.

Thanks @RangerRick , thats good input! I left the XML annotations on the methods and only moved the JPA annotations to the attributes. Not sure if that brings unintended side effects?

RangerRick:
> Thanks @RangerRick , thats good input! I left the XML annotations on the methods and only moved the JPA annotations to the attributes. Not sure if that brings unintended side effects?

¯\\\_(ツ)\_/¯ 

It just means that serializing to the database won't match serializing to the XML, but I'd imagine that won't hurt anything.

PR #2781, opened 2019-10-24
NMS-12322: Switch Drools in alarmd from STREAM to CLOUD processing

j-white:
JIRA: https://issues.opennms.org/browse/NMS-12322

We appear to be hitting a memory leak and some unexpected behavior in Drools when using STREAM processing with our alarms. I suspect this is due to the fact that alarms are not necessarily ordered sequentially in time - we use the given event time and this is driven by external sources.

Here we switch from using STREAM to CLOUD processing in order to help remedy the problem.

CLOUD processing is not as efficient as STREAM processing for our use case, but stability is more important for the time being until we can determine the root cause of the problem.


RangerRick:
That is less complicated than I expected it to be.

PR #2782, opened 2019-10-24
CircleCI stability updates

j-white:
Here we update the build so that nodejs & yarn no longer run when building the RPM packages for OpenNMS/Minion/Sentinel - the artifacts are already available from the compile step.

We also add a script that will help us monitor JVM running the systems tests in order to help diagnose the intermittent "timeout due to no output" errors.

### DIFF
--- a/.circleci/scripts/jvmprocmon-start.sh
+++ b/.circleci/scripts/jvmprocmon-start.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+TARGET="/tmp/jvmprocmon"
+
+echo "Starting JVMProcMon writing to: ${TARGET}"
+mkdir -p "$TARGET"
+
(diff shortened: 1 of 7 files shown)


.circleci/scripts/jvmprocmon-start.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
j-white: Validated with shellcheck:
```
$ shellcheck ./.circleci/scripts/jvmprocmon-start.sh  && echo "OK"
OK
```

mvrueden: Hehe, I just wanted to add a comment like "Did you validate this with shellcheck?" 👍

PR #2787, opened 2019-10-25
NMS-12307: Add ability to update definitions when SNMP profile changes and fix tests

cgorantla:
This reverts revert of #2738  and fix tests.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12307


cgorantla:
Integration tests on Bamboo passed which were failing earlier. I would wait for CI to pass again and merge this since it is already reviewed before on #2738

PR #2789, opened 2019-10-27
NMS-12356: Detect and help resolve Karaf bootstrap issues: first version

patrick-schweizer:
* JIra: https://issues.opennms.org/browse/NMS-12356


container/karaf/src/main/filtered-resources/etc/org.apache.karaf.features.cfg
```diff
@@ -144,8 +144,8 @@ featuresBoot = ( \
   geolocation, \
   ifttt-integration, \
   opennms-endpoints-grafana, \
-  opennms-reporting, \
-  opennms-bundle-refresher
+  opennms-bundle-refresher, \
```
patrick-schweizer: @j-white, the bundle-refresher lookup doesn't seem to work. That's why I chose opennms-reporting instead.
`SERVICE_LOOKUP.lookup(org.opennms.features.vaadin.components.bundlerefresher.Activator.class, null);`
or it's implemented interface
`SERVICE_LOOKUP.lookup(org.osgi.framework.BundleActivator, null);`
returns null. I assume this is because the interface is not opennms specific.

features/telemetry/daemon/pom.xml
```diff
@@ -46,7 +46,21 @@
       <artifactId>org.opennms.features.telemetry.config.jaxb</artifactId>
       <version>${project.version}</version>
     </dependency>
-
+    <dependency>
```
patrick-schweizer: needs to be removed again once we find a better home for KarafStartupMonitor

features/telemetry/daemon/src/main/java/org/opennms/netmgt/telemetry/daemon/KarafStartupMonitor.java
```diff
@@ -0,0 +1,135 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.client.RestTemplate;
+
+/**
+ * Used in startup script to detect if Karaf started properly.
+ * TODO: patrick move this class to better place
```
patrick-schweizer: @j-white what do you think where this class should live?

features/telemetry/daemon/src/main/java/org/opennms/netmgt/telemetry/daemon/KarafStartupMonitor.java
```diff
@@ -0,0 +1,135 @@
...
+        try {
+            final ReportStoreService service = SERVICE_LOOKUP.lookup(serviceClass, null);
+            return service != null;
+        } catch (Exception e) {
+            LOG.warn("Could not lookup {}, will fail.", serviceClass.getName(), e);
+            return false;
+        }
+    }
+
+    private boolean isKarafOkAlternative() {
```
patrick-schweizer: this seems to work as well as an alternative but isKarafOk() should do => we can delete this method

container/karaf/src/main/filtered-resources/etc/org.apache.karaf.features.cfg
mvrueden: Just for clarification:

The `SERVICE_LOOKUP` is (as far as I know) a non OSGi-specific method to get service references from the OSGi registry when not running inside Karaf itself. Basically if you need a service from the OSGI-registry but run in the jetty classpath context. 

This also means, that only interfaces can be requested, which have been exposed as a service before.

The opennms-bundle-refresher bundle defines an Activator (So it is on the classpath of that bundle) but it never exposes it to the underlying OSGi service registry, thus the call `SERVICE_LOOKUP.lookup(org.osgi.framework.BundleActivator, null);` always correctly returns null.
The call `SERVICE_LOOKUP.lookup(org.opennms.features.vaadin.components.bundlerefresher.Activator.class, null);` should also not work as the Activcator is a class not an interface.

features/telemetry/daemon/src/main/java/org/opennms/netmgt/telemetry/daemon/KarafStartupMonitor.java
```diff
@@ -0,0 +1,135 @@
...
+        try {
+            // TODO: if we use this method we wil hae to move the health service to the end of the list in org.apache.features.cfg
+            String healthEndpoint = "http://admin:admin@localhost:8980/opennms/rest/health"; // TODO: patrick Is this URL always correct?
+            String base64Creds = Base64.getEncoder().encodeToString("admin:admin".getBytes());
+            HttpHeaders headers = new HttpHeaders();
+            headers.add("Authorization", "Basic " + base64Creds);
+            HttpEntity request = new HttpEntity(headers);
+            ResponseEntity<String> response = new RestTemplate().exchange(healthEndpoint, HttpMethod.GET, request, String.class);
```
mvrueden: This is not going to work well:
1. the URL is never the same. Should use `/opennms/rest/health` or configure username, password and url
2. Use /rest/health/probe instead, as it does not require authentication and simply returns 200 or > 500

Generally I would prefer to use the HealthCheckService here instead.
We could also add a custom HealthCheck, which verifies that ALL features defined in the features.boot.cfg have been installed correctly. Also if you cannot find the HealthCheckService, the container is not started properly

opennms-base-assembly/src/main/filtered/bin/fix-karaf-setup.sh
```diff
@@ -0,0 +1,44 @@
...
+# Restore files
+PRISTINE_DIR="$OPENNMS_HOME/share/etc-pristine"
+ETC_DIR="$OPENNMS_HOME/etc"
+echo "Copying pristine config files to $ETC_DIR"
+cp -p "$PRISTINE_DIR/jmx."*".cfg" $ETC_DIR
+cp -p "$PRISTINE_DIR/org.apache."*".cfg" "$ETC_DIR/"
+cp -p "$PRISTINE_DIR/org.ops4j.pax."*".cfg" "$ETC_DIR/"
+cp -p "$PRISTINE_DIR/profile.cfg" "$ETC_DIR/"
+
+echo "Done. Please try again to run OpenNMS."
```
mvrueden: This does not pass shellcheck:

Screenshot 2019-10-28 at 09 24 16

See https://www.shellcheck.net for more information.

patrick-schweizer: fixed.

features/telemetry/daemon/src/main/java/org/opennms/netmgt/telemetry/daemon/KarafStartupMonitor.java
patrick-schweizer: I think we have two ways to check for Karaf health:

1. expose an http endpoint and query for it, done in isKarafOkAlternative() or
2. call a Karaf service directly, done in isKarafOk()
I think option 2 is the preferred one, once we get it to work properly we can remove isKarafOkAlternative().

features/telemetry/daemon/src/main/java/org/opennms/netmgt/telemetry/daemon/KarafStartupMonitor.java
patrick-schweizer: moved to its own feature (karaf-health)

features/telemetry/daemon/src/main/java/org/opennms/netmgt/telemetry/daemon/KarafStartupMonitor.java
patrick-schweizer: we removed the method

features/karaf-health/src/main/java/org/opennms/features/karaf/health/bundle/Activator.java
```diff
@@ -26,25 +26,32 @@
...
 
-
 public class Activator implements BundleActivator {
 
+    private ServiceRegistration karafHealthServiceSvcReg;
+
     @Override
     public void start(BundleContext context) throws Exception {
-        // Determine if any vaadin theme fragments are unresoled
+        // Determine if any vaadin theme fragments are unresolved
```
j-white: @mvrueden is it work keeping the Vaadin bundle refreshing parts around? Or do you think it's safe to remove them now?

features/karaf-health/src/main/java/org/opennms/features/karaf/health/daemon/KarafStartupMonitor.java
```diff
@@ -0,0 +1,89 @@
...
+                    "3.) run the script bin/fix-karaf-setup.sh");
+            throw new IllegalStateException(message);
+        }
+        LOG.info("{} is started.", NAME);
+    }
+
+    private boolean isKarafOk() {
+        Class serviceClass = KarafHealthService.class;
+        try {
+            final KarafHealthService service = SERVICE_LOOKUP.lookup(serviceClass, null);
```
j-white: We should add a log message to indicate that we're going to block while waiting for a service to be available.

features/karaf-health/src/main/java/org/opennms/features/karaf/health/daemon/KarafStartupMonitor.java
```diff
@@ -0,0 +1,89 @@
...
+            return service != null;
+        } catch (Exception e) {
+            LOG.warn("Could not lookup {}, will fail.", serviceClass.getName(), e);
```
j-white:
```suggestion
            LOG.error("Could not lookup {}, will fail.", serviceClass.getName(), e);
```

opennms-base-assembly/src/main/filtered/bin/fix-karaf-setup.sh
```diff
@@ -0,0 +1,45 @@
...
+fi
+
+# Prune data directory, exept for history.txt
+echo "Pruning directory $DATA_DIR"
+find "$DATA_DIR" -mindepth 1 -maxdepth 1 -not -name 'history.txt' -exec rm -r {} \;
+
+# Restore files
+PRISTINE_DIR="$OPENNMS_HOME/share/etc-pristine"
+ETC_DIR="$OPENNMS_HOME/etc"
+echo "Copying pristine config files to $ETC_DIR"
```
j-white: Should be skipped if $PRISTINE_DIR does not exist.

opennms-base-assembly/src/main/filtered/bin/fix-karaf-setup.sh
```diff
@@ -0,0 +1,45 @@
...
+   echo "Ok, goodbye!"
+   exit 0
+fi
+
+# Prune data directory, exept for history.txt
+echo "Pruning directory $DATA_DIR"
+find "$DATA_DIR" -mindepth 1 -maxdepth 1 -not -name 'history.txt' -exec rm -r {} \;
+
+# Restore files
+PRISTINE_DIR="$OPENNMS_HOME/share/etc-pristine"
```
j-white: Need to make sure this works with the Debian packages as well.

opennms-base-assembly/src/main/filtered/bin/fix-karaf-setup.sh
```diff
@@ -0,0 +1,45 @@
...
+	true # all is well
+else
+	echo "Error: you must run this script as $RUNAS, not '$myuser'" >&2
+	exit 4	# According to LSB: 4 - user had insufficient privileges
+fi
+
+echo "This script will try to fix karaf configuration problems by:"
+echo "- pruning $DATA_DIR directory. This is where the Karaf cache sits."
+echo "- restore all Karaf related configuration files to a pristine state"
+echo "It might be a good idea to take a backup of $OPENNMS_HOME/etc"
```
j-white:
```suggestion
echo "We recommend making a backup of $OPENNMS_HOME/etc before proceeding."
```

opennms-base-assembly/src/main/filtered/bin/fix-karaf-setup.sh
```diff
@@ -0,0 +1,45 @@
...
+# Restore files
+PRISTINE_DIR="$OPENNMS_HOME/share/etc-pristine"
+ETC_DIR="$OPENNMS_HOME/etc"
+echo "Copying pristine config files to $ETC_DIR"
+cp -p "$PRISTINE_DIR/jmx."*".cfg" $ETC_DIR
+cp -p "$PRISTINE_DIR/org.apache."*".cfg" "$ETC_DIR/"
+cp -p "$PRISTINE_DIR/org.ops4j.pax."*".cfg" "$ETC_DIR/"
+cp -p "$PRISTINE_DIR/profile.cfg" "$ETC_DIR/"
+
+echo "Done. Please try again to run OpenNMS."
```
j-white:
```suggestion
echo "Done. Please try restarting OpenNMS."
```

features/karaf-health/src/main/java/org/opennms/features/karaf/health/bundle/Activator.java
mvrueden: I am not 100% sure. We talked about removing it, as Karaf added a fix for this at some point. However we have never actually verified it. I suggest we create an issue to remove the bundle refresher code and leave it for now. After this is merged we can then focus on removing the bits (or verify if they can remove and then remove it)

opennms-base-assembly/src/main/filtered/bin/fix-karaf-setup.sh
patrick-schweizer: @j-white how do I test that?

j-white: I confirmed this works. For future reference, you can pull the .deb package from the Bamboo build i.e. https://bamboo.opennms.org/browse/OPENNMS-OPENNMS381-DEB-5/artifact, and then install that on a Debian-based system. In this case, I tested on Ubuntu 18.04 using `sudo docker run -ti ubuntu:18.04 /bin/bash`.

patrick-schweizer: ah ok, so you meant by that comment that the script runs on a Debian-based system? I guess we are good then since I develop on Ubuntu :-)

PR #2795, opened 2019-10-28
NMS-12367: Update Drools dependency to 7.18.0

UberPinguin:
The 'From with modify fires unexpected rule' bug appears to be fixed in Drools 7.8.0.
Comparing 'mvn dependency:list' does not indicate any transitive dependency impact.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12367

### DIFF
--- a/container/features/src/main/resources/features.xml
+++ b/container/features/src/main/resources/features.xml
@@ -1133,7 +1133,7 @@
     <feature name="opennms-kafka-producer" description="OpenNMS :: Kafka :: Producer" version="${project.version}">
       <feature version="${guavaVersion}">guava</feature>
       <feature version="${kafkaStreamsVersion}">kafka-streams</feature>
-      <bundle>mvn:com.google.protobuf/protobuf-java/${protobuf3Version}</bundle>
+      <bundle>mvn:com.google.protobuf/protobuf-java/${protobufVersion}</bundle>
       <bundle>mvn:org.opennms.features.kafka/org.opennms.features.kafka.producer/${project.version}</bundle>
     </feature>
 
--- a/dependencies/drools/pom.xml
+++ b/dependencies/drools/pom.xml
@@ -15,7 +15,7 @@
     can depend on to use the drools library.
   </description>
   <properties>
-    <droolsVersion>7.8.0.Final</droolsVersion>
+    <droolsVersion>7.18.0.Final</droolsVersion>
   </properties>
   <dependencies>
     <dependency>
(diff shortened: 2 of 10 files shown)


j-white:
Thanks @UberPinguin!

Pushed to `pr/jira/NMS-12367` for a Bamboo run.

j-white:
Existing ITs are failing in Bamboo: https://bamboo.opennms.org/browse/OPENNMS-OPENNMS388-1

Looks like we may need more changes here.

UberPinguin:
They're all failing with the same stack trace. I'll try to dig into it today, but I don't see why a Drools version bump would stop `org.opennms.netmgt.correlation.drools.PossibleCause.possibleCause` from loading.

UberPinguin:
This seems really dumb, but I think Drools is trying to resolve the function `possibleCause` in `NodeParentRules.drl` as a method in the `PossibleCause` class.
I renamed the function to `possiblCause`, and all of the tests pass.
How do you want to proceed?

j-white:
👍 Good catch

UberPinguin:
I borrowed and modified the test case from the upstream bug, DROOLS-3606.
When testing with that case, I didn't find a working version of Drools until 7.18.0.Final.

j-white:
Updated `pr/jira/NMS-12367` with the new patch. Confirmed mvel is the only transitive dependency update in `lib/`. Waiting on Bamboo.

j-white:
Tests are failing due to a requirement on protobuf-java 3.x, whereas we pin it to 2.x in our root pom. We can upgrade this dependency in this branch if (and only if) the serialization format does not change as a result of the update. This serialization is used for the Sink API over Kafka between Minion & OpenNMS as well as in the TCP exporter for RRD metrics.

UberPinguin:
It looks like the only difference is in whether repeated fields are packed by default.
It was optional in proto2, and is default in proto3.

From https://developers.google.com/protocol-buffers/docs/encoding#embedded:
"If a proto2 message definition has repeated elements (without the [packed=true] option), the encoded message has zero or more key-value pairs with the same field number. These repeated values do not have to appear consecutively; they may be interleaved with other fields. The order of the elements with respect to each other is preserved when parsing, though the ordering with respect to other fields is lost. In proto3, repeated fields use packed encoding, which you can read about below.

For any non-repeated fields in proto3, or optional fields in proto2, the encoded message may or may not have a key-value pair with that field number.
[....]
In proto3, repeated fields of scalar numeric types are packed by default.

UberPinguin:
May have to abandon this, the proto3 upgrade changes a lot of method signatures, and breaks several OpenNMS classes.

j-white:
Closing in favor of https://github.com/OpenNMS/opennms/pull/2808. Thanks for the help @UberPinguin.

PR #2796, opened 2019-10-29
NMS-12318: Process newSuspect events in a single threaded fashion

cgorantla:
In discovery, new suspect could be generated for the same node which has multiple interfaces. This could cause multiple nodes with same interfaces. 
This will restrict  scans that are coming from new suspect events to be single threaded.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12318

### DIFF
--- a/opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/Provisioner.java
+++ b/opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/Provisioner.java
@@ -40,8 +40,11 @@ import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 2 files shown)


opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/Provisioner.java
```diff
@@ -104,7 +108,12 @@
     private SnmpAgentConfigFactory m_agentConfigFactory;
     
     private volatile TimeTrackingMonitor m_stats;
-    
+
+    private final ThreadFactory newSuspectThreadFactory = new ThreadFactoryBuilder()
+            .setNameFormat("newSuspectExecutor")
+            .build();
+    private ExecutorService m_newSuspectExectuor = Executors.newSingleThreadExecutor(newSuspectThreadFactory);
```
j-white: Should be shutdown in `destroy()` method?

PR #2798, opened 2019-10-31
NMS-10396: Config tester doesn't detect missing xml datacollection file

patrick-schweizer:
* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-10396

### DIFF
--- a/opennms-config-tester/src/test/java/org/opennms/netmgt/config/tester/ConfigTesterTest.java
+++ b/opennms-config-tester/src/test/java/org/opennms/netmgt/config/tester/ConfigTesterTest.java
@@ -47,9 +47,12 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Properties;
 import java.util.Set;
+import java.util.UUID;
 
 import org.junit.After;
+import org.junit.AfterClass;
 import org.junit.Before;
+import org.junit.BeforeClass;
 import org.junit.FixMethodOrder;
 import org.junit.Ignore;
 import org.junit.Test;
@@ -59,6 +62,7 @@ import org.opennms.core.test.ConfigurationTestUtils;
 import org.opennms.netmgt.config.tester.checks.ConfigCheckValidationException;
 import org.opennms.netmgt.filter.FilterDaoFactory;
 import org.opennms.test.DaoTestConfigBean;
+import org.springframework.util.FileSystemUtils;
 import org.springframework.util.StringUtils;
 
 import com.google.common.io.Files;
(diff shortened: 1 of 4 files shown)


opennms-config-tester/src/test/java/org/opennms/netmgt/config/tester/ConfigTesterTest.java
```diff
@@ -67,32 +71,46 @@
...
+        // once that is done the home dir cannot be changed via a property since the application context has been loaded
+        // already in a static way.
+        setUpOpennmsHomeInTmpDir();
+    }
+
+    private static void setUpOpennmsHomeInTmpDir() throws IOException {
+        String opennmsHomeOriginal = System.getProperty(OPENNMS_HOME_NAME);
+        opennmsHomeDir = Files.createTempDir();
+        System.setProperty(OPENNMS_HOME_NAME, opennmsHomeDir.getAbsolutePath());
+        FileSystemUtils.copyRecursively(new File(opennmsHomeOriginal, "etc"), new File(opennmsHomeDir, "etc"));
```
j-white: This seems expensive. Can you explain why we need to copy the tree? Is it feasible to limit to copying to `.xml` files?

patrick-schweizer: @j-white ,
The reason why I copied the files is a combination of the following: 

1. the tested config files don't live in the same directory.
    Most are in `opennms-base-assembly/src/main/filtered/etc` but the newly added test (datacollection) tests the files in `protocols/xml/src/main/etc`.
2. the ConfigTester uses BeanUtils: https://github.com/OpenNMS/opennms/blob/78b73f19c28b79cad50910c74e6ac6a3b99b2a9e/opennms-config-tester/src/main/java/org/opennms/netmgt/config/tester/ConfigTester.java#L173 which creates and holds the ApplicationContext in a static way (it won't recreate the aplication context for every test method). Therefor setting the system property `opennms.home` has no effect after the first call => we can't point to a different directory.

Possible solutions (in order of effectiveness but also effort):
1. Leave as is (copying 36,4 MB of data)
2. Copying only relevant files (e.g. the subfolder events is 32 MB alone)
3. Change the ConfigTester in a way so that we can reload the application context (and therefore point to another directory in each test)

I think 3. is probably the cleanest solution. I will look into it...

patrick-schweizer: Hmm, I did a test and creating the tmp directory and copying the files takes ~ 25 ms on my machine. It is also only done only once for this test, not for each test method. I'm not sure it's worth optimizing that?  @j-white what do you think?

j-white: Not worth it then. Let's just keep it as-is.

patrick-schweizer: :+1:

PR #2799, opened 2019-11-04
NMS-12376: Allow fetching graphs with SZL and Vertices in Focus (or apply default focus)

mvrueden:
Here we allow fetching a "view" of a given graph.
The view is calculated by a given set of vertices and focus and the semantic zoom level (szl).
The returned graph contains all vertices and edges which are in focus and have a connection with hops lesser or equal the number of the szl parameter.
In case the vertices in focus do not exist in the graph, they are ignored.

JIRA: https://issues.opennms.org/browse/NMS-12376

### DIFF
--- a/features/graph/api/src/test/java/org/opennms/netmgt/graph/api/generic/GenericGraphTest.java
+++ b/features/graph/api/src/test/java/org/opennms/netmgt/graph/api/generic/GenericGraphTest.java
@@ -28,11 +28,21 @@
 
 package org.opennms.netmgt.graph.api.generic;
 
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertThat;
 import static org.opennms.core.test.OnmsAssert.assertThrowsException;
 
+import java.util.List;
+
+import org.hamcrest.Matchers;
 import org.junit.Test;
+import org.opennms.netmgt.graph.api.ImmutableGraph;
+import org.opennms.netmgt.graph.api.VertexRef;
+import org.opennms.netmgt.graph.api.focus.Focus;
+import org.opennms.netmgt.graph.api.focus.FocusStrategy;
 import org.opennms.netmgt.graph.api.generic.GenericGraph.GenericGraphBuilder;
+
+import com.google.common.collect.Lists;
 
 public class GenericGraphTest {
 
@@ -66,7 +76,7 @@ public class GenericGraphTest {
     
     @Test
     public void shouldNotAllowNamespaceChangeAfterAddingElements(){
-        GenericGraphBuilder graphBuilder =  GenericGraph.builder();
+        GenericGraphBuilder graphBuilder = GenericGraph.builder();
         
         // 1.) set namespace on "empty graph" => shouldn't be a problem
         graphBuilder.namespace("some namespace");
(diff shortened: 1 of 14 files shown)


features/graph/api/src/test/java/org/opennms/netmgt/graph/api/generic/GenericGraphTest.java
```diff
@@ -86,4 +96,53 @@ public void shouldNotAllowNamespaceChangeAfterAddingElements(){
         assertThrowsException(IllegalStateException.class, () -> graphBuilder.properties(new MapBuilder<String, Object>()
                 .withProperty(GenericProperties.NAMESPACE, otherNamespace).build()));
     }
+
+    @Test
+    public void shouldConsiderSemanticZoomLevel() {
+        final String NAMESPACE = "dummy";
```
patrick-schweizer: shouldn't the variable called "namespace" (small letters) since it is not a constant?

features/graph/rest/src/main/java/org/opennms/netmgt/graph/rest/api/Query.java
```diff
@@ -0,0 +1,56 @@
...
+
+public class Query {
+
+    public static final int DEFAULT_SEMANTIC_ZOOM_LEVEL = 1;
+
+    private Integer szl;
+
+    private List<String> verticesInFocus;
+
+    public Integer getSzl() {
```
patrick-schweizer: I wonder if we shouldn't call the method getSemanticZoomLevel()? I think this makes the code more readable than the abbreviation szl. (there are multiple occurrences in the code of "szl")

features/graph/rest/src/main/java/org/opennms/netmgt/graph/rest/impl/renderer/JsonGraphRenderer.java
```diff
@@ -75,40 +76,51 @@ public String render(ImmutableGraphContainer<?> graphContainer) {
...
+                edgeProperties.put("source", genericEdge.getSource().getId());
+                edgeProperties.put("target", genericEdge.getTarget().getId());
+                final JSONObject jsonEdge = new JSONObject(edgeProperties);
+                jsonEdgesArray.put(jsonEdge);
             });
-        return jsonContainer.toString();
+
+            // Convert Vertices
+            graph.getVertices().stream().forEach(vertex -> {
+                final JSONObject jsonVertex = new JSONObject(vertex.asGenericVertex().getProperties());
```
patrick-schweizer: I wonder if we need here a similar converter mechanism as for serializing into the database: https://github.com/OpenNMS/opennms/blob/jira/NMS-12376-smoke/features/graph/service/src/main/java/org/opennms/netmgt/graph/persistence/converter/ConverterService.java

mvrueden: Good point.
Probably yes and no.
For now we don't need it, as all properties are converted using its `.toString()`-method.
As we are only storing primitive datatypes in the map, this is not required.
However, later on we may store more complex objects, which may need to be manually serialized.

patrick-schweizer: But what about Collections? We implemented storing Collections in HZN-1523.

mvrueden: This code is serializing the properties to JSON, not a persistable String. Therefore the logic is different. At the moment we only save primitive datatypes/enums (or collections of primitive datatypes/enums) which JSONObject already serializes correctly. For other objects (e.g. IpInfo) it may serialize correctly as well. Therefore at the moment there is no need to implement a more generic solution. At a later state, this may vary.

features/graph/rest/src/main/java/org/opennms/netmgt/graph/rest/api/Query.java
RangerRick: Or at least "zoom" or "zoomLevel" instead, that would make more sense.

mvrueden: I decided to name it `semanticZoomLevel`. Even if the statement `SZL` or `szl` is a common term in context of Topology/Graph Service.

features/graph/api/src/test/java/org/opennms/netmgt/graph/api/generic/GenericGraphTest.java
mvrueden: Technically, the correct naming would be `Namespace`. Anyways, I have no strong opinion about this, and just named it `namespace` as that seems to be the common use in OpenNMS

mvrueden:
I addressed the review comments and fixed some test failures. Also rebased against latest `features/graph-service` as that branch was failing.

PR #2802, opened 2019-11-06
NMS-12382: Provide enrichment of graphs (using node enrichment as an example)

mvrueden:
Here we provide the feature to enrich graphs with additional information.
As an example the graph is enriched with node information if the graph requested it (property on the graph).
Later on this functionality can be used to enrich the graph with status information etc.

Originally the idea was to put the logic inside the `GraphContainerProvider` or `GraphProvider` itself. However while working on this, separating it made more sense, as only when viewing the graph (which should always be using the focus and szl approach), the data needs to be enriched.

Requires PR https://github.com/OpenNMS/opennms/pull/2799 to be merged.

Jira: https://issues.opennms.org/browse/NMS-12382


features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericGraph.java
```diff
@@ -119,10 +121,16 @@ public Focus getDefaultFocus() {
...
-//    }
+    @Override
+    public GenericVertex getVertex(NodeRef nodeRef) {
+        Objects.requireNonNull(nodeRef);
+        for (NodeRef eachVariant : nodeRef.getVariants()) {
+            if (nodeRefToVertexMap.containsKey(eachVariant)) {
+                return nodeRefToVertexMap.get(eachVariant);
+            }
+        }
+        return null;
```
patrick-schweizer: A general thought here: since we are designing a new module: 
How is our stance towards Optional as return parameter vs. return null? I think that is the preferred way these days to handle possible null return values?

features/graph/service/src/main/java/org/opennms/netmgt/graph/enrichment/NodeEnrichmentProcessor.java
```diff
@@ -0,0 +1,78 @@
...
+    private final Logger log = LoggerFactory.getLogger(getClass());
+    private final NodeService nodeService;
+
+    public NodeEnrichmentProcessor(NodeService nodeService) {
+        this.nodeService = Objects.requireNonNull(nodeService);
+    }
+
+    @Override
+    public boolean canEnrich(GenericGraph graph) {
+        final Boolean value = graph.getProperty("enrichment.resolveNodes", false);
```
patrick-schweizer: not sure if "enrichment.resolveNodes" shouldn't be a constant somewhere?

features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericGraph.java
mvrueden: In most cases Optionals are just not necessary. In the rare cases where an Optional would be neat, in my opinion, don't make up for the need to use them everywhere. Especially on an API level. The reason for this is, that you have to document the API level anyways and "must not be null" or "may be null" is as sufficient as the optional. And if it really helps, you can always create an Optional and use it.

TL;DR: may be nice, but not necessary. If you need it for exception handling or so, just create it ad-hoc, instead of using it everywhere.

features/graph/service/src/main/java/org/opennms/netmgt/graph/enrichment/NodeEnrichmentProcessor.java
mvrueden: Will add a TODO after merged and address it later, as there will be more, but I am not sure where to put them yet.

PR #2809, opened 2019-11-12
NMS-12393: Add HTTP header injection in HTTP Monitors

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-12393

HttpPostMonitor now uses the same parameters for the defintion of headers like the HttpMonitor. The PageSequenceMonitor now allows an additional element &lt;header name="foo" value="bar"/> inside of &lt;page>&lt;/page> elements.


mfuhrmann:
@christianpape So... the HttpMonitor was already able to do such a heartbeat post?

christianpape:
> @christianpape So... the HttpMonitor was already able to do such a heartbeat post?

The HttpMonitor was able to set headers.

mfuhrmann:
> > @christianpape So... the HttpMonitor was already able to do such a heartbeat post?
> 
> The HttpMonitor was already able to set headers but GET-only.

Since Opsgenie is also able to receive GET requests with an auth header for the heartbeat, I could have used this one to create a heartbeat monitor, correct?

Are the docs of HttpMonitor sufficient for that case?

christianpape:
> > > @christianpape So... the HttpMonitor was already able to do such a heartbeat post?
> > 
> > 
> > The HttpMonitor was already able to set headers but GET-only.
> 
> Since Opsgenie is also able to receive GET requests with an auth header for the heartbeat, I could have used this one to create a heartbeat monitor, correct?
> 
> Are the docs of HttpMonitor sufficient for that case?

Yes, you can use header[0-9]+ parameters to specify header entries.

christianpape:
Closed and reopened with new branch name  jira/NMS-12393-smoke.

PR #2810, opened 2019-11-12
Graph-Service: Rework/fix graph persistence related issues

mvrueden:
Here we address some flaws with the current persistence:
 - entities not having their own package
 - entities not using their own sequence
 - graph info not persisting/loading the correct vertex type
 - edges do not allow referencing vertices outside their scope


features/graph/api/src/main/java/org/opennms/netmgt/graph/api/generic/GenericGraph.java
```diff
@@ -480,5 +509,10 @@ public String getLabel() {
             return GenericVertex.class;
         }
 
+        @Override
+        public Class<Vertex> getDomainVertexType() {
```
patrick-schweizer: I think this violates the design principle we defined behind the class structure:
GenericX is the generic data storage. If needed a domain specific view (DomainX) can be projected on that data. There can be potentially more than one DomainX view for a GenericX element. The GenericX element shouldn't know anything of its views (since it is generic).
Not sure if we want to alter our design principle?

mvrueden: I had to solve the following issue:
The `GraphRepository` can persist ANY `ImmutableGraphContainer`. 
To do so the persistence is converting the container to its generic counterpart (thus `GenericGraphContainer`) and persists that to database. Afterwards only `GenericGraphContainer`s can be loaded.

Besides this, it should be possible to load the `GenericGraphContainer`s `ContainerInfo` to get the meta data. However, as we persisted it as a `GenericGraphContainer` that information is not persisted and therefore lost.

It is not breaking our design principle. It is similar to the namespace. It just says, that the vertex information in this graph were provided according to a certain vertex type. That is similar to the namespace. Yes in theory we could have multiple/different views (Domain objects) of the same generic object, but in practice we never will or most likely never will.

I am not super happy with the current solution, but not sure on how to otherwise solve the "read meta data" issue without persisting it.

mvrueden: TL;DR: Not worth adding this restriction now. Removed the `domainVertexType` code.

I thought about this some more:

The main reasons for the VertexType is to allow implementing SearchProvider and EnrichmentProcessors which can work on the vertex types of a given graph. Usually the namespace is sufficient, however when providing a more generic implementation (e.g. search for nodes or enrich node information) not every graph provider wants to implement this functionality. Enrichment/Searching can be implemented for all vertices of a certain type vs a certain namespace.

At the moment not many SearchProvider/EnrichmentProcessors make use of this design decision. Also the only code section which actually uses the "problematic situation" is an example provider showing how to use the `GraphRepository`. Given that there is no actual need to "know" what type of vertex was persisted (other than shown in the example) it is not worth adding this now. Let's wait until we know more about what information we actually need.

PR #2811, opened 2019-11-13
NMS-12365: Fixed WS-MAN poller and test

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-12365

### DIFF
--- a/features/wsman/src/main/java/org/opennms/netmgt/poller/monitors/WsManMonitor.java
+++ b/features/wsman/src/main/java/org/opennms/netmgt/poller/monitors/WsManMonitor.java
@@ -28,12 +28,16 @@
 
 package org.opennms.netmgt.poller.monitors;
 
+import java.net.MalformedURLException;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.NoSuchElementException;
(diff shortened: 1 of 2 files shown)


agalue:
I tested the changes and the solution works with and without Minions involved, thanks!

```
admin@opennms> poller:poll org.opennms.netmgt.poller.monitors.WsManMonitor 172.20.12.154 resource-uri=http://schemas.microsoft.com/wbem/wsman/1/wmi/root/cimv2/Win32_Service selector.Name=SNMP rule="#State matches 'Running' and #Status matches 'OK'"
.
Service is Up on 172.20.12.154 using org.opennms.netmgt.poller.monitors.WsManMonitor:
(No properties were returned by the monitor.

admin@opennms> poller:poll -l Apex org.opennms.netmgt.poller.monitors.WsManMonitor 172.20.12.154 resource-uri=http://schemas.microsoft.com/wbem/wsman/1/wmi/root/cimv2/Win32_Service selector.Name=SNMP rule="#State matches 'Running' and #Status matches 'OK'"
.
Service is Up on 172.20.12.154 using org.opennms.netmgt.poller.monitors.WsManMonitor:
(No properties were returned by the monitor.
```

PR #2812, opened 2019-11-13
NMS-12393: Add HTTP header injection in HTTP Monitors

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-12393

HttpPostMonitor now uses the same parameters for the defintion of headers like the HttpMonitor. The PageSequenceMonitor now allows an additional element &lt;header name="foo" value="bar"/> inside of &lt;page>&lt;/page> elements.


features/poller/monitors/core/src/main/java/org/opennms/netmgt/poller/monitors/PageSequenceMonitor.java
```diff
@@ -355,6 +356,10 @@ public void process(HttpRequest request, HttpContext ctx) throws HttpException,
                     method.setQueryParameters(expandParms(svc));
                 }
 
+                for(final org.opennms.netmgt.config.pagesequence.Header header : m_page.getHeaders()) {
+                    method.setHeader(new BasicHeader(header.getName(), header.getValue()));
```
mvrueden: Should probably also add some logging statements which header is used.
Also what happens if header name and value is empty?

features/poller/monitors/core/src/main/java/org/opennms/netmgt/poller/monitors/HttpPostMonitor.java
```diff
@@ -201,6 +205,13 @@ public PollStatus poll(MonitoredService svc, Map<String, Object> parameters) {
                 LOG.debug("HttpPostMonitor: Constructed URL is {}", ub);
 
                 HttpPost post = new HttpPost(ub.build());
+
+                for (final String key : parameters.keySet()) {
+                    if (HEADER_PATTERN.matcher(key).matches()) {
+                        post.setHeader(getHeader(ParameterMap.getKeyedString(parameters, key, null)));
```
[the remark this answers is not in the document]
christianpape: done

features/poller/monitors/core/src/main/java/org/opennms/netmgt/poller/monitors/PageSequenceMonitor.java
christianpape: done

PR #2797, opened 2019-10-30
NMS-12368: Add an easy to follow step-by-step guide to use Kafka for Minion

indigo423:
### All Contributors

* [x] Have you read and followed our [Contribution Guidelines](https://github.com/OpenNMS/opennms/blob/develop/CONTRIBUTING.md)?
* [x] Have you [made an issue in the OpenNMS issue tracker](https://issues.opennms.org)?<br>If so, you should:
  1. update the title of this PR to be of the format: `${JIRA-ISSUE-NUMBER}: subject of pull request`
  2. update the JIRA link at the bottom of this comment to refer to the real issue number
  3. prefix your commit messages with the issue number, if possible
* [x] Have you made a comment in that issue which points back to this PR?
* [x] Have you updated the JIRA link at the bottom of this comment to link to your issue?
* [x] If this is a new or updated feature, is there documentation for the new behavior?
* [x] If this is new code, are there unit and/or integration tests?
* [x] If this PR targets a `foundation-*` branch, does it avoid changing files in `$OPENNMS_HOME/etc/`?

Refactored documentation to setup OpenNMS and Minion to use Apache Kafka instead of the built-in ActiveMQ messaging system. The guide is now consolidated into simple to follow step-by-step guide.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12368

PR #2813, opened 2019-11-13
NMS-12375: Migrate docker configuration official install guide

indigo423:
### All Contributors

Migrate our readme documentation to our official guide on how to run OpenNMS with Docker and create a full reference with mount point conventions and environment variables. Describe a minimal docker-compose file that allows you to run an OpenNMS service stack with PostgreSQL, Horizon and connect a Minion via ActiveMQ.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12375

Screenshot 2019-11-13 at 14 02 12


opennms-doc/guide-install/src/asciidoc/text/run-with-docker/introduction.adoc
```diff
@@ -0,0 +1,190 @@
...
+    volumes:<7>
+      - data-postgres:/var/lib/postgresql/data
+    healthcheck:<8>
+      test: [ "CMD-SHELL", "pg_isready -U postgres" ]
+      interval: 10s
+      timeout: 30s
+      retries: 3
+
+  horizon:
+    image: opennms/horizon:25.0.0<9>
```
indigo423: I was not able to use maven properties version properties in source files, If don't want to maintain this, we need to add a placeholder and the user needs to replace it with the version he wants to install.

opennms-doc/guide-install/src/asciidoc/text/run-with-docker/introduction.adoc
```diff
@@ -0,0 +1,190 @@
...
+      - "8101:8101/tcp"
+      - "61616:61616/tcp"
+    healthcheck:<14>
+      test: [ "CMD", "curl", "-f", "-I", "http://localhost:8980/opennms/login.jsp" ]
+      interval: 1m
+      timeout: 5s
+      retries: 3
+  
+  minion:
+    image: opennms/minion:25.0.0
```
indigo423: I was not able to use maven properties version properties in source files, If don't want to maintain this, we need to add a placeholder and the user needs to replace it with the version he wants to install.

agalue:
Nice, thanks! I'll carefully review the content soon.

What about the `README.md` on each directory inside the `opennms-container` directory? As we will have usage configuration as part of the official docs, maybe we can include `development` and `building` details on those `README.md` (like be able to build your images on your own), and remove the `usage` bits from them. What do you think?

indigo423:
> What about the `README.md` on each directory inside the `opennms-container` directory? As we will have usage configuration as part of the official docs, maybe we can include `development` and `building` details on those `README.md` (like be able to build your images on your own), and remove the `usage` bits from them. What do you think?
 
Very good hint. I'll dig into this.

agalue:
For Kafka, I believe it worth explaining how the prefix and env-var conversion works.

If the env-var starts with the prefix `KAFKA_RPC`, that will end up adding something to `$MINION_HOME/etc/org.opennms.core.ipc.rpc.kafka.cfg`. Similarly, if the env-var starts with the prefix `KAFKA_SINK`, that will end up adding something to `$MINION_HOME/etc/org.opennms.core.ipc.sink.kafka.cfg`.

When the `KAFKA_RPC_BOOSTRAP_SERVERS` is added, Kafka for RPC will be enabled by creating a file called `featuresBoot.d/kafka-rpc.boot` with the expected content. Something similar happens for Sink.

Now, after the prefix, what the user must know is that a Kafka variable is expected. For example, `KAFKA_RPC_BOOSTRAP_SERVERS` will add the Kafka property `bootstrap.servers` (note how `_` was changed too `.`) to the appropriate file.

You can add more complex settings as well as I outlined on my K8s Playground [here](https://github.com/OpenNMS/opennms-drift-kubernetes/blob/master/README.md) (look for the sample Minion execution). Speaking of which, you can also see how to add advanced Flow settings there. If you think it worth adding those as well, let me know and I can explain how they work.

indigo423:
> For Kafka, I believe it worth explaining how the prefix and env-var conversion works.
> 
> If the env-var starts with the prefix `KAFKA_RPC`, that will end up adding something to `$MINION_HOME/etc/org.opennms.core.ipc.rpc.kafka.cfg`. Similarly, if the env-var starts with the prefix `KAFKA_SINK`, that will end up adding something to `$MINION_HOME/etc/org.opennms.core.ipc.sink.kafka.cfg`.
> 
> When the `KAFKA_RPC_BOOSTRAP_SERVERS` is added, Kafka for RPC will be enabled by creating a file called `featuresBoot.d/kafka-rpc.boot` with the expected content. Something similar happens for Sink.
> 
> Now, after the prefix, what the user must know is that a Kafka variable is expected. For example, `KAFKA_RPC_BOOSTRAP_SERVERS` will add the Kafka property `bootstrap.servers` (note how `_` was replaced to `.`) to the appropriate file. Another example would be, `KAFKA_RPC_AUTO_OFFSET_RESET` will be translated to `auto.offset.reset` for RPC.
> 
> You can add more complex settings as well as I outlined on my K8s Playground [here](https://github.com/OpenNMS/opennms-drift-kubernetes/blob/master/README.md) (look for the sample Minion execution). Speaking of which, you can also see how to add advanced Flow settings there. If you think it worth adding those as well, let me know and I can explain how they work.

yes, this would be nice explaining it a little bit more in detail. I just picked the ones from the docs. Just add it in the PR and we iterate in the branch until we have it as we want it.

To be perfectly honest, thinking more in detail about the properties I'm unsure, here is thought on this:

* PRO as we have it right now: As I understand it right now, we have all the flexibility to create any property for UDP listener and Kafka without making a change
* CONTRA: The user needs to know before which properties are available and he there is currently no place to look this up

Probably this is a problem we have to solve in a different spot and we move on :) What doy ou think?

agalue:
That is a very good point.

The reason why I believe mentioning Kafka is important is because in my experience with support, ActiveMQ is the main source of Minion communication issues (at least for about 90% of the high number of tickets I've been working since the technology exists). The majority of the customers are using Kafka, and those that are still using AMQ is because they haven't upgraded to a version on which they can fully use Kafka.

On top of that, like Cassandra, Kafka is designed for zero downtime and high availability in mind, which is not the case of ActiveMQ. Having what they call a "Network of Brokers" requires a deep knowledge about the technology, compared with what you need to know to have Kafka up and running in production.

In terms of your comments, Kafka is not a problem, as their documentation is superb. Here are all the consumer/producer options you can use:

For producers:
https://kafka.apache.org/documentation/#producerconfigs

For consumers:
https://kafka.apache.org/documentation/#consumerconfigs

For the Sink pattern, Minion is always the producer, whereas OpenNMS and Sentinels are always consumers.

For the RPC pattern, OpenNMS is a producer and Minion is a consumer when OpenNMS sends a request to be executed by the Minion; but OpenNMS is a consumer and Minion is a producer when Minion sends the response back to OpenNMS. Topics are different on each use case.

So, with all that information, I cannot see a reason for not mentioning Kafka.

For the UDP listeners, I leave that to you, as since H23 those settings have been changed on every major version and each major version also introduces new settings. All that is supposed to be explained on our official docs, so if those docs are good enough, I can't find a reason for not mentioning this when configuring Minon.

I mean, if the entrypoint has been designed to offer these features, why hidding them?

indigo423:
> That is a very good point.
> 
> The reason why I believe mentioning Kafka is important is because in my experience with support, ActiveMQ is the main source of Minion communication issues (at least for about 90% of the high number of tickets I've been working since the technology exists). The majority of the customers are using Kafka, and those that are still using AMQ is because they haven't upgraded to a version on which they can fully use Kafka.

I totally agree with you. This is the reason for this #2797. 
> 
> On top of that, like Cassandra, Kafka is designed for zero downtime and high availability in mind, which is not the case of ActiveMQ. Having what they call a "Network of Brokers" requires a deep knowledge about the technology, compared with what you need to know to have Kafka up and running in production.
> 
> In terms of your comments, Kafka is not a problem, as their documentation is superb. Here are all the consumer/producer options you can use:
> 
> For producers:
> https://kafka.apache.org/documentation/#producerconfigs
> 
> For consumers:
> https://kafka.apache.org/documentation/#consumerconfigs

These links are in the description how to use in Kafka see link above.

> So, with all that information, I cannot see a reason for not mentioning Kafka.

I totally agree with you. This is the reason for this #2797. 

 > For the UDP listeners, I leave that to you, as since H23 those settings have been changed on every major version and each major version also introduces new settings. All that is supposed to be explained on our official docs, so if those docs are good enough, I can't find a reason for not mentioning this when configuring Minon.

> I mean, if the entrypoint has been designed to offer these features, why hidding them?

This should be anyway be described in our docs, no matter if we ran on bare-metal or in the container.

So the main question to me here is the following. Assume the topics for configuring Kafka and the UDP listener are documented to run Minion / Sentinel on bare-metal.

In the special case running in the container, there is an additional barrier which we introduced in our entrypoint script which translates environment variables from `UDP_...something` to `KAFKA_...something`. This is where I specifically refer to the

* PRO as we have it right now: As I understand it right now, we have all the flexibility to create any property for UDP listener and Kafka without making a change
* CONTRA: The user needs to know before which properties are available and there is currently no place to look this up

agalue:
> I'm trying to figure where do we draw the line between describing all the things configuring on bare-metal and doing the same thing running it in a container without describing the whole world twice :) You know what I mean?

Yes, but even if the final content on the configuration files is the same, the way to configure them is different when containers are involved, so the way I can see how to avoid describing them twice is by describing how the env-var to properties translation works, and then add a link to the proper reference that has the detailed description (for Kafka #2797, and for the Listeners, what we have on the official docs).

Makes sense?

indigo423:
> > I'm trying to figure where do we draw the line between describing all the things configuring on bare-metal and doing the same thing running it in a container without describing the whole world twice :) You know what I mean?
> 
> Yes, but even if the final content on the configuration files is the same, the way to configure them is different when containers are involved, so the way I can see how to avoid describing them twice is by describing how the env-var to properties translation works, and then add a link to the proper reference that has the detailed description (for Kafka #2797, and for the Listeners, what we have on the official docs).
> 
> Makes sense?

yes make sense :)

indigo423:
Anything else here we need to address?

agalue:
> Anything else here we need to address?

As I've asked before, what about the `README.md` on each directory inside the `opennms-container` directory?

indigo423:
> > Anything else here we need to address?
> 
> As I've asked before, what about the `README.md` on each directory inside the `opennms-container` directory?

Removed the readme and added one with the build instructions.

PR #2814, opened 2019-11-13
NMS-12392: Fix send-event.pl with time zone support

indigo423:
### All Contributors

This PR applies a patch from attached to the issue to fix our send-event.pl script which got busted when we introduced time zone changes.

With all the kudos to the contributor: @jb4415🏅

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12392

### DIFF
--- a/opennms-base-assembly/src/main/filtered/bin/send-event.pl
+++ b/opennms-base-assembly/src/main/filtered/bin/send-event.pl
@@ -56,7 +56,7 @@ if ($version)  { print "$0 version $VERSION\n"; exit; }
 if ($help)     { print get_help(); exit; }
 
 # parm array is numerically referenced in OpenNMS' templates
-@PARMS = map { parse_parm($_) } @PARMS;
+@PARMS = reverse map { parse_parm($_) } @PARMS;
 
 my $hostname = hostname;
 
@@ -163,11 +163,11 @@ if (defined $SERVICE) {
 
 my ($sec,$min,$hour,$mday,$mon,$year,$wday,$yday,$isdst) = gmtime(time);
 $year += 1900;
-my $month = sprintf("%02d", $mon + 1);
-$mday     = sprintf("%02d", $mday);
-$min      = sprintf("%02d", $min);
-$sec      = sprintf("%02d", $sec);
-
+my $month = $mon;
+$min = sprintf("%02d", $min);
+$sec = sprintf("%02d", $sec);
+my @week = ('Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday');
+my @month = ('January', 'February', 'March', 'April', 'May', 'June', 'July', 'August', 'September', 'October', 'November', 'December');
 my $uuidattribute;
 if (defined $UUID) {
 	$uuidattribute = "uuid=\"$UUID\"";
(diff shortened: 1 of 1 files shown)


opennms-base-assembly/src/main/filtered/bin/send-event.pl
```diff
@@ -187,7 +188,7 @@ END
 $event .= "   <nodeid>$NODEID</nodeid>\n"          if (defined $NODEID);
 
 $event .= <<END;
-   <time>${year}-${month}-${mday}T${hour}:${min}:${sec}+00:00</time>
+   <time>$week[$wday], $mday $month[$month] $year $hour:$min:$sec o'clock $ZONE</time>
```
j-white: We should continue using ISO-8601, and not revert to the old format.

RangerRick:
Oh, hah, @j-white beat me to it.  :)

indigo423:
> Oh, hah, @j-white beat me to it. :)

yeah my perl skills are extremely limited :D

jb4415:
Timezone wasn't the issue that I reported. The value being inserted between the time tags was the correct GMT time for events.

The issue was the hour component of that time was not being zero padded so that if I submitted an event between midnight and 9:59 am GMT, the hour value only had a single digit.

My fix was to add a sprintf statement for the hour field to ensure the hour was always 2 digits. Similar to the other values that already have the sprintf command.

opennms-base-assembly/src/main/filtered/bin/send-event.pl
```diff
@@ -163,11 +163,12 @@
...
 $year += 1900;
-my $month = sprintf("%02d", $mon + 1);
-$mday     = sprintf("%02d", $mday);
-$min      = sprintf("%02d", $min);
-$sec      = sprintf("%02d", $sec);
-
+my $month = $mon;
+$hour = sprintf("%02d", $hour);
+$min = sprintf("%02d", $min);
+$sec = sprintf("%02d", $sec);
```
jb4415: This is the only change that is required for this.  line 164 gets gmtime so there should be no need to be concerned about timezones as internally, the script is working in GMT.

RangerRick:
I went ahead and reverted the  `c06dcdaa1ad48b488dbfe1ffe55607d1d5a61111` commit since the only thing necessary was the hours fix.  The zone stuff is wrong, because the code uses `gmtime()` so it's always UTC, thus the offset in the ISO bit is always 0.

If we _did_ want to implement actual time zone support, it would require different code than the bit from the old `send-event.pl` that was not really working (which is why I removed it when I made the original ISO changes).

RangerRick:
What is the `incomplete` tag for?

indigo423:
> What is the `incomplete` tag for?

Don't merge the PR, there is stuff missing and the PR is also not complete to be reviewed. You can remove the tag if this is not the case anymore.

RangerRick:
I guess I'm just confused about what's missing.  The timezone stuff was a red herring, timezone support (that came back when you copied over some other `send-event.pl`) was _removed_ when I switched the script to support ISO-8601, because it was incomplete/incorrect in the first place.  The original issue is _only_ about fixing the format of the event sent.

RangerRick:
(The diff in his original comment on the ticket was the diff between `25.0.0` and `25.1.0`, @jb4415 was only including it in the ticket to show the changes, AFAICT.)

jb4415:
> (The diff in his original comment on the ticket was the diff between `25.0.0` and `25.1.0`, @jb4415 was only including it in the ticket to show the changes, AFAICT.)

Yes you're quite right. The diff I included was between the 25.0.0 and 25.1.0 versions. I was just trying to show that I could see that send-event.pl had been updated to reflect the new event XML format. I posted that before i figured out the fix. My fix was just to add the 0 padding of the hour field. Sorry for the confusion.

PR #2816, opened 2019-11-14
Refactoring/Cleanup: Move graphml related classes to their own module and make it more osgi

mvrueden:
This PR converts the Graphml related modules and classes to pure OSGi modules and consolidates some of the classes while doing so.

Originally the `GraphmlRestService` was introduced before we had the capabilities to expose new ReST services via the OSGi service registry, but it also was never migrated to the "new way of exposing rest services". 

There is no issue for this, as it is purely refactoring.


mvrueden:
I am going ahead and will merge this. Feel free to make any comments/suggestions and I will address it in a seperate PR/branch.

PR #2819, opened 2019-11-19
NMS-12389: Add cache for all graph containers

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-12389


features/graph/service/src/main/java/org/opennms/netmgt/graph/service/DefaultGraphContainerCache.java
```diff
@@ -0,0 +1,131 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+import org.opennms.netmgt.graph.api.ImmutableGraphContainer;
+import org.opennms.netmgt.graph.api.service.GraphContainerCache;
+
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+
+public class DefaultGraphContainerCache implements GraphContainerCache {
```
patrick-schweizer: I wonder why you implemented the locking and caching "by hand" instead of using something like a ConcurrentHashMap? Maybe you want to add a comment about the reasons behind?

An alternative would maybe be to use Guava Caching? https://github.com/google/guava/wiki/CachesExplained
That could be an easy and elegant solution which offers lots of possibilities out of the box?

mvrueden: Ideally we would use the `CacheBuilder` from guava. However periodically invalidating only SOME of the cache values is not supported by guava. That is why I implemented it myself.

j-white: Take a look at the Caffeine cache - based on Guava with more features. We use it for caching DNS responses with varying TTLs: https://github.com/OpenNMS/opennms/blob/develop/features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/CaffeineCache.java#L61

mvrueden: Cool, I can take a look :)

mvrueden: Caffeine cache is providing the features we need to make this work. Nice 👍

PR #2822, opened 2019-11-22
Perform code coverage weekly instead of nightly

j-white:
Code coverage is currently performed nightly and made available here:
  https://sonarcloud.io/dashboard?id=OpenNMS_opennms

I don't think we use it enough to justify running nightly, I propose we switch this to weekly instead.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -275,10 +275,11 @@ commands:
           destination: generated-tests
 
 workflows:
-  nightly-coverage:
+  weekly-coverage:
     triggers:
       - schedule:
-          cron: "0 0 * * *"
+          # Saturday at 12:00 AM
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -275,10 +275,11 @@ commands:
           destination: generated-tests
 
 workflows:
-  nightly-coverage:
+  weekly-coverage:
     triggers:
       - schedule:
-          cron: "0 0 * * *"
+          # Saturday at 12:00 AM
+          cron: "0 0 * * 5"
```
mvrueden: I couldn't find anything about which underlying cron implementation this is using, but 
ssn't 5 supposed to be Friday and 6 Saturday?

j-white: Yes.

PR #2824, opened 2019-11-25
NMS-12396: Handle SNMP profiles by default in NodeScan

cgorantla:
Since SNMPDetector doesn't support SNMP profiles yet, NodeScan skips the scanning node if the default config doesn't detect SNMP service. 
This PR fixes that by always trying to detect SNMP service using config from profiles.

Also fixes a bug related to RPC exceptions.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12396
* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-12399


j-white:
Looks good 👍

PR #2826, opened 2019-11-26
NMS-12411: Allow foreign source definitions in new suspect scan

cgorantla:
Add foreign source for NodeScan and IpInterfaceScan so that relevant
foreign source definitions may be selected.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12411


j-white:
Is it feasible to add test coverage here?

cgorantla:
> Is it feasible to add test coverage here?

Should be possible. Will add one.

cgorantla:
Above changes in tests should test loading of custom foreign source definitions  when foreign source is supplied with new suspect event call.

PR #2827, opened 2019-11-27
Rework BusinessServiceGraphProvider and expose all relevant properties (except status)

mvrueden:
Here we rework the bsm provider a bit and remove the complexity due to vertex inheritance, which is not really required. We also expose the MapFunction and ReduceFunction as well as provide a method to allow defining how property values are serialised (to json).


features/graph/provider/bsm/src/main/java/org/opennms/netmgt/graph/provider/bsm/converter/ReduceFunctionJsonConverter.java
```diff
@@ -0,0 +1,77 @@
...
+        jsonOutput.put("type", input.getClass().getSimpleName().toLowerCase());
+        input.accept(new ReduceFunctionVisitor<Void>() {
+            @Override
+            public Void visit(HighestSeverity highestSeverity) {
+                return null;
+            }
+
+            @Override
+            public Void visit(HighestSeverityAbove highestSeverityAbove) {
+                jsonOutput.put("threshold", highestSeverityAbove.getThreshold());
```
patrick-schweizer: is the property name correct?

mvrueden: Yes

mvrueden:
Yeah, I will add some more test coverage.

PR #2828, opened 2019-11-27
NMS-12408: Implement GraphNotificationService

mvrueden:
Here we expose a GraphNotificationService which allows to notify various graph change listeners.
This provides the basis on pushing changes actively to kafka, however this is not implemented here.

JIRA: https://issues.opennms.org/browse/NMS-12408


features/graph/api/src/main/java/org/opennms/netmgt/graph/api/updates/ChangeSet.java
```diff
@@ -39,8 +39,6 @@
 import org.opennms.netmgt.graph.api.focus.Focus;
 import org.opennms.netmgt.graph.api.info.DefaultGraphInfo;
 import org.opennms.netmgt.graph.api.info.GraphInfo;
-import org.opennms.netmgt.graph.updates.listener.GraphChangeListener;
-import org.opennms.netmgt.graph.updates.listener.GraphChangeSetListener;
 
 public class ChangeSet<G extends ImmutableGraph<V, E>, V extends Vertex, E extends Edge> {
```
patrick-schweizer: I wonder if a change set shouldn't be immutable since it represents a diff between two immutable graphs? Maybe the "builder" methods should go into a builder?

mvrueden: Yes it should be immutable. I already worked on that in some other branch, so will skip that here. The builder change could be addressed there as well.

mvrueden:
Will go ahead and merge this. Mentioned issues will be addressed separately

PR #2829, opened 2019-11-27
Final cleanup of Graph-Service branch

mvrueden:
Here we clean up the last bits, mostly rewording/reorganisation of code and also removing/addressing of TODOs.

### Still Work in Progress ###


features/graph/api/src/main/java/org/opennms/netmgt/graph/api/updates/ChangeSet.java
```diff
@@ -40,67 +41,30 @@
 import org.opennms.netmgt.graph.api.info.DefaultGraphInfo;
 import org.opennms.netmgt.graph.api.info.GraphInfo;
 
-public class ChangeSet<G extends ImmutableGraph<V, E>, V extends Vertex, E extends Edge> {
+public final class ChangeSet<G extends ImmutableGraph<V, E>, V extends Vertex, E extends Edge> {
```
patrick-schweizer: +1 for an immutable change set

PR #2830, opened 2019-11-27
NMS-12413: Fix SNMP Remove from definitions

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12413


cgorantla:
Found issues with previous commit when multiple profiles were involved.
Also handled case where definitions with different locations but same IP Address.

PR #2839, opened 2019-12-04
Use SessionUtils instead of TransactionOperations

mvrueden:
Here we migrate yet another module to use `SessionUtils` instead of `TransactionOperations` to get rid of weird `Dependency not resolvable` issues when refreshing a bundle and some bundles are using `TransactionOperations`


j-white:
LGTM

PR #2840, opened 2019-12-04
NMS-12429: Reuse rules when defining port ranges

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-12429

### DIFF
--- a/features/flows/classification/engine/impl/src/main/java/org/opennms/netmgt/flows/classification/internal/DefaultClassificationEngine.java
+++ b/features/flows/classification/engine/impl/src/main/java/org/opennms/netmgt/flows/classification/internal/DefaultClassificationEngine.java
@@ -32,10 +32,12 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.concurrent.ExecutionException;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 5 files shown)


features/flows/classification/engine/impl/src/test/java/org/opennms/netmgt/flows/classification/internal/DefaultClassificationEngineTest.java
```diff
@@ -191,6 +192,18 @@ public void verifyAllPortsToEnsureEngineIsProperlyInitialized() {
...
+    // See NMS-12429
+    @Test
+    public void verifyDoesNotRunOutOfMemory() {
+        final List<Rule> rules = Lists.newArrayList();
+        for (int i=0; i<100; i++) {
+            final Rule rule = new RuleBuilder().withName("rule1").withPosition(i+1).withProtocol("UDP").withDstAddress("192.168.0." + i).build();
+            rules.add(rule);
+        }
+        final DefaultClassificationEngine engine = new DefaultClassificationEngine(() -> rules, FilterService.NOOP);
+        System.gc();
```
j-white: What's the purpose of this explicit gc call?

features/flows/classification/engine/impl/src/test/java/org/opennms/netmgt/flows/classification/internal/value/PortValueTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+import org.hamcrest.Matchers;
+import org.junit.Test;
+
+public class PortValueTest {
+
+    @Test
+    public void verifySingleValue() {
+        final PortValue portValue = new PortValue("5");
+        assertThat(portValue.getPorts(), Matchers.hasItems(5));
```
j-white: These read better with the `Matchers.` prefix.

mvrueden: Don't understand what that means.

features/flows/classification/engine/impl/src/test/java/org/opennms/netmgt/flows/classification/internal/DefaultClassificationEngineTest.java
mvrueden: The next test is using a timeout and I saw it fail because too much memory was allocated and cleaning it took too long. The `System.gc()` call should prevent that.

features/flows/classification/engine/impl/src/test/java/org/opennms/netmgt/flows/classification/internal/value/PortValueTest.java
j-white: `import static org.hamcrest.Matchers.hasItems`

features/flows/classification/engine/impl/src/test/java/org/opennms/netmgt/flows/classification/internal/DefaultClassificationEngineTest.java
j-white: Can we add a comment explaining that?

mvrueden: Yep. I forgot to push that.

mvrueden:
I gave this another spin and it looks like the memory footprint is much smaller, but it is also very slow in comparison to the old solution. Need to investigate a bit further and provide another fix.

mvrueden:
I reworked this a bit and now use a Guava Cache instead of a simple hash map as it is faster, even if it is pre-populated.
Got the memory footprint down from ~1200 MB to ~420 MB for roughly 250 rules.

features/flows/classification/engine/impl/src/test/java/org/opennms/netmgt/flows/classification/internal/DefaultClassificationEngineTest.java
mvrueden: Turns out, the whole thing was damn slow instead of just using memory. Reworked it so no `System.gc()` call required.

features/flows/classification/engine/impl/src/test/java/org/opennms/netmgt/flows/classification/internal/value/PortValueTest.java
mvrueden: Done

features/flows/classification/engine/impl/src/main/java/org/opennms/netmgt/flows/classification/internal/DefaultClassificationEngine.java
```diff
@@ -50,20 +53,65 @@
...
+                            // Only src or dst ports are defined (or none)
+                            // if none, the value of either src or dst port may be empty, as the filtering already occurred
+                            // through the index of the rule in the classifierPortList.
+                            if (rule.hasDstPortDefinition()) {
+                                portRule.setDstPort(rule.getDstPort());
+                            }
+                            if (rule.hasSrcPortDefinition()) {
+                                portRule.setSrcPort(rule.getSrcPort());
+                            }
+                        }
```
Naicisum: Am I missing something? It looks like if the rule has a dst and src, it only set a SrcPort and ignores DstPort. Shouldn't just the code in the else be present as it would work on all cases?

mvrueden: Yes, you are absolutely right. Before the change it was necessary to distinguish between both cases, now it can be merged.

PR #2842, opened 2019-12-04
NMS-12427: Allow telemetry flows to balance across Kafka partitions

Naicisum:
### All Contributors

### Pull Request Process

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12427

### DIFF
--- a/features/telemetry/api/src/main/java/org/opennms/netmgt/telemetry/api/receiver/TelemetryMessage.java
+++ b/features/telemetry/api/src/main/java/org/opennms/netmgt/telemetry/api/receiver/TelemetryMessage.java
@@ -28,24 +28,37 @@
 
 package org.opennms.netmgt.telemetry.api.receiver;
 
-import org.opennms.core.ipc.sink.api.Message;
 import java.net.InetSocketAddress;
 import java.nio.ByteBuffer;
 import java.util.Date;
+
+import org.opennms.core.ipc.sink.api.Message;
 
 public class TelemetryMessage implements Message {
     private final InetSocketAddress source;
     private final ByteBuffer buffer;
     private final Date receivedAt;
+    private final boolean isFlowMessage;
+    private final boolean useRoutingKey;
 
     public TelemetryMessage(InetSocketAddress source, ByteBuffer buffer) {
-        this(source, buffer, new Date());
+        this(source, buffer, new Date(), false, true);
     }
 
     public TelemetryMessage(InetSocketAddress source, ByteBuffer buffer, Date receivedAt) {
+        this(source, buffer, receivedAt, false, true);
+    }
+
+    public TelemetryMessage(InetSocketAddress source, ByteBuffer buffer, Date receivedAt, boolean isFlowMessage) {
+        this(source, buffer, receivedAt, isFlowMessage, true);
+    }
+
+    public TelemetryMessage(InetSocketAddress source, ByteBuffer buffer, Date receivedAt, boolean isFlowMessage, boolean useRoutingKey) {
         this.source = source;
         this.buffer = buffer;
         this.receivedAt = receivedAt;
+        this.isFlowMessage = isFlowMessage;
+        this.useRoutingKey = useRoutingKey;
     }
 
     public InetSocketAddress getSource() {
@@ -59,4 +72,8 @@ public class TelemetryMessage implements Message {
     public Date getReceivedAt() {
         return receivedAt;
     }
+
+    public boolean getIsFlowMessage() { return isFlowMessage; }
+
+    public boolean getUseRoutingKey() { return useRoutingKey;}
 }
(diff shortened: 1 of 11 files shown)


Naicisum:
Skipping CircleCI tests until basic code review performed, will hit CI once changes and suggestions are implemented.

features/telemetry/api/src/main/java/org/opennms/netmgt/telemetry/api/receiver/TelemetryMessage.java
```diff
@@ -28,24 +28,37 @@
...
+
+    public TelemetryMessage(InetSocketAddress source, ByteBuffer buffer, Date receivedAt, boolean isFlowMessage) {
+        this(source, buffer, receivedAt, isFlowMessage, true);
+    }
+
+    public TelemetryMessage(InetSocketAddress source, ByteBuffer buffer, Date receivedAt, boolean isFlowMessage, boolean useRoutingKey) {
         this.source = source;
         this.buffer = buffer;
         this.receivedAt = receivedAt;
+        this.isFlowMessage = isFlowMessage;
```
fooker: the naming pattern for boolean properties is just `property` for the field and `isProperty` and `setProperty` for getters and setter. So the field should be called `flowMessage` and the getter should be called `isFlowMessage`.

features/telemetry/common/src/main/java/org/opennms/netmgt/telemetry/common/ipc/TelemetrySinkModule.java
```diff
@@ -171,7 +173,11 @@ public boolean isBlockWhenFull() {
 
     @Override
     public Optional<String> getRoutingKey(final TelemetryProtos.TelemetryMessageLog message) {
-        return Optional.of(String.format("%s@%s:%d", message.getLocation(), message.getSourceAddress(), message.getSourcePort()));
+        if (message.getUseRoutingKey()) {
+            return Optional.of(String.format("%s@%s:%d", message.getLocation(), message.getSourceAddress(), message.getSourcePort()));
+        } else {
+            return null;
```
fooker: This is an `Optional` value and thus should not return `null`. Use `Optional.of()` and `Optional.empty()` instead.

fooker:
Why not just disable it always? Is there any use-case where it which requires this option to be changeable? If we can remove it, it can just be done by removing the custom implementation for `getRoutingKey`.

Naicisum:
> What is the `isFlowMessage` flag is used for?

If a flow contains Streaming Telemetry data such as thresholding data, then order becomes important and playing messages to a single topic will help process messages in order and possibly reduce out of order events. SFlow is a primary example, at present this does not take telemetry data into account for sflow. Will need to find a detection method early on to flag this data. Any ideas/suggestions?

> Why not just disable it always? Is there any use-case where it which requires this option to be changeable? If we can remove it, it can just be done by removing the custom implementation for `getRoutingKey`.

See first answer.

Naicisum:
Actually after some thought, it does seem redundant. We can simply override the routing key value if streaming telemetry data is detected.

features/telemetry/common/src/main/resources/telemetry.proto
```diff
@@ -1,15 +1,18 @@
-option java_package = "org.opennms.netmgt.telemetry.ipc";
+syntax = "proto3";
```
j-white: A couple things here:
1) We need to keep this in proto2 if we're targeting the fix to foundation-2019 or 25.x - we can't break backwards compatibility
2) We shouldn't need to serialize the `isFlowMessage` or `useRoutingKey` fields. These are used to make the decision, but don't need to be in the payload once that's done - I think the overall strategy may need to be revised. The SinkModule for the respective types should be the one driving this logic, and not the individual messages.

features/telemetry/common/src/main/java/org/opennms/netmgt/telemetry/common/ipc/TelemetrySinkModule.java
Naicisum: Done

features/telemetry/api/src/main/java/org/opennms/netmgt/telemetry/api/receiver/TelemetryMessage.java
Naicisum: Done

features/telemetry/common/src/main/resources/telemetry.proto
Naicisum: No arguments on either point. I just looked for a way to preserve the flag/behavior from the parser all the way to the Kafka dispatcher.

If it can be done at the sink modules prior to dispatch, then that would be more efficient.

fooker:
@Naicisum You're right. Didn't think about thresholding. While thresholding supports switching between two sentinals, it's only meant for failover and very slow.

We need to be able to decide on a per message base. SFlow may or may not have data relevant for thresholding in it (and it's the same for upcoming BMP support, too).

features/telemetry/common/src/main/resources/telemetry.proto
Naicisum: Changed to proto2

Naicisum:
This PR to be closed in lieu of the alternative: https://github.com/OpenNMS/opennms/pull/2855

Naicisum:
PR Closed, see https://github.com/OpenNMS/opennms/pull/2855 for implementation.

PR #2844, opened 2019-12-05
NMS-12430: Fix TCP parsers

fooker:
This replaces the the buffers used for parsing in telemtryd.

* The first commit replace the used NIO `ByteBuffer`s with nettys _`ByteBuf` to avoid complete consumption of incoming data.
* The second commit fix the TCP parsers by make the check the available data before consuming it - allowing for more data to arrive if needed to complete a full packet.

JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12430

_I'm really sorry..._

PR #2848, opened 2019-12-08
NMS-12430: Fix TCP parsers

fooker:
The back-port of #2844 for `foundation-2019`.
This is the same change-set as in the other PR but with all BMP related changes removed.

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12430


smoke-test/src/test/java/org/opennms/smoketest/flow/TCPFlowsIT.java
```diff
@@ -0,0 +1,85 @@
...
+    @ClassRule
+    public static final OpenNMSStack stack = OpenNMSStack.withModel(StackModel.newBuilder()
+            .withTelemetryProcessing()
+            .build());
+
+    // Verifies that when OpenNMS and ElasticSearch is running and configured, that sending a flow packet
+    // will actually be persisted in elastic
+    @Test
+    public void verifyFlowStack() throws Exception {
+        final InetSocketAddress flowTelemetryAddress = stack.opennms().getNetworkProtocolAddress(NetworkProtocol.IPFIX_TCP);
```
j-white: 👍

PR #2849, opened 2019-12-09
NMS-12397: Refactoring of the Cassandra installation instructions

mfuhrmann:
### All Contributors

### Pull Request Process

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12397


mfuhrmann:
I removed our install steps since the apache.org guide already contains the complete procedure.
I only added the autostart and apt mark hold part as recommendation.

indigo423:
IMHO, I would remove the Cassandra install guide for *all* distributions not just for Debian and add a generic reference with something like ... install Cassandra from the official installation  instructions.

christianpape:
> IMHO, I would remove the Cassandra install guide for all distributions not just for Debian and add a generic reference with something like ... install Cassandra from the official installation instructions.

I agree.

indigo423:
I've reworked the Newts setup guide and removed the Cassandra installation instructions with references to Cassandra so we don't have to maintain it. I'll rewrite the JIRA issue as well so it reflects the refactoring of this change better.

indigo423:
@agalue I've worked in your comment and instead of using the newts init command, I've used your CQL statement to initialize the keyspace. I've shortened a bit the explanation in the first section why we recommend to use TWCS instead of STSC to avoid adding a wall of text :) What do you think?

Thanks a lot for adding this input here.

opennms-doc/guide-install/src/asciidoc/text/time-series-storage/newts.adoc
```diff
@@ -0,0 +1,148 @@
...
+EOF
+----
+
+<1> Associate time series data by the foreign ID instead the database generated node ID
+<2> Set time series strategy to use `newts`
+<3> Host or IP addresses of the _Cassandra_ cluster nodes, can be a comma separated list
+<4> Name of the keyspace which is initialized and used
+<5> Port to connect to _Cassandra_
+
+===== Step 2: Initialize the _Newts_ schema in _Cassandra_
```
j-white: If this is what we're recommending to use for the default schema, should we not just update `newts init` to use this instead?

agalue: That's another valid approach. I envision that particular enhancement as follow:

Request not only the replication factor but also the desired retention (TTL), from there, calculate the window settings accordingly and generate the schema. Now, I think that would be in the Newts code base, right? But, without having those changes in place, the documentation should be something useful to have.

agalue:
@indigo423 It is STCS (Size Tiered Compaction Strategy), not STSC ;)

opennms-doc/guide-install/src/asciidoc/text/time-series-storage/newts.adoc
indigo423: > If this is what we're recommending to use for the default schema, should we not just update `newts init` to use this instead?

that would be my next issue and investigate how to change it there :)

indigo423:
@agalue Fixed the TODO and @j-white I've created an enhancement issue with our findings in [NEWTS-109](https://issues.opennms.org/browse/NEWTS-109).

indigo423:
I've dug a little bit deeper and I think we should solve the problem @agalue mentioned in NEWTS-109 instead of here. To address the STCS vs. TWCS issue, I've added a disclaimer to explain this is not a production-ready tuned setup and referenced to @agalue article which describes a lot of things in [greater detail](https://opennms.discourse.group/t/sizing-cassandra-for-newts/771).

Any comments?

indigo423:
> LGTM!
> 
> Maybe it worth updating the discourse article to include how to set up the whole newts keyspace instead of focusing on how to tune just the samples table for TWCS, now that those details are not here anymore. What do you think?

I made the Discourse article a Wiki post so people can maintain it. I'll add the instructions in the discourse article.

PR #2852, opened 2019-12-10
NMS-12441: Add documentation for the new Graph Service API

mvrueden:
JIRA: https://issues.opennms.org/browse/NMS-12441


opennms-doc/guide-development/src/asciidoc/text/graph/graph.adoc
```diff
@@ -0,0 +1,201 @@
...
+* `GenericVertex` and
+* `GenericEdge`
+
+Each of the above have besides their designated functionality described in link:#ga-development-graph-service-api-terminology[Terminology] also properties assigned to them.
+Some properties are mandatory and determined by the element's type.
+Others are optionally and totally up to the entity providing a graph to the system.
+
+The following rules apply to the _Generic Graph Model_:
+
+- Each Graph container must be uniquely identified.
```
patrick-schweizer: somehow we have the explanation about identifiers twice. Here and in chapter Identifier and Namespace. Maybe consolidate?

opennms-doc/guide-development/src/asciidoc/text/graph/graph.adoc
```diff
@@ -0,0 +1,201 @@
...
+===== Identifier and Namespace
+
+In order to uniquely identify each element and each graph container within the system different strategies are applied.
+
+1. The container is uniquely identified by an `id` property.
+2. The graph container is uniquely identified by a `namespace` property.
+3. Each element within the graph is uniquely identified by a `namespace` and `id` property.
+The `id` property must not be unique in the system, but within the graph.
+
+If a graph defines an `id` property as well, this SHOULD but MUST not be identically to the `namespace` property.
```
patrick-schweizer: Reading this now I wonder why we do that? That seems to make things more complicated than they have to. Shouldn't we merge the id and the namespace and use only one?

opennms-doc/guide-development/src/asciidoc/text/graph/graph.adoc
```diff
@@ -0,0 +1,201 @@
...
+
+Refer to module `features/graph/api` in package `org.opennms.netmgt.graph.api.generic` for implementation details.
+
+
+==== Domain Graph Model
+
+A _Domain Graph Model_ is a model implementing a use case in its domain using the _Generic Graph Model_.
+This can be achieved by simply using the _Generic Graph Model_ and setting the according properties.
+However in more complex use cases it may be necessary to provide a more Java-friendly way.
+Therefore basic Domain Graph objects are provided to act as a view on the _Generic Graph Model_ to provide a more convenient implementation.
```
patrick-schweizer: To make it even more clear maybe add something like : "Domain objects don't hold information on their own but provide only access to information that can be deducted from the underlying Generic Model and it's properties"

opennms-doc/guide-development/src/asciidoc/text/graph/graph.adoc
mvrueden: The `id` property is only there to be compatible with GraphML.
This means, you can set an id property on the graph, but it is identified using the namespace. In order to make things easier, simply make them the same. 
This is more or less only relevant when importing graphs via GraphML.

opennms-doc/guide-development/src/asciidoc/text/graph/graph.adoc
mvrueden: Good point.

opennms-doc/guide-development/src/asciidoc/text/graph/graph.adoc
patrick-schweizer: hm ok, I think you are right. It is the same thing but since the children of a graph (Vertex and Edge) need to reference the graph but have their own attribute named "id" we had to come up with a different name. In the end we are talking about the graph identifier. Let's leave it as is.

mvrueden: As mentioned above the separate `Namespace and Identifier` chapter is not ideal. When rearranging (or consolidating) things I will consider rewriting this a bit to make it more clear.

opennms-doc/guide-development/src/asciidoc/text/graph/graph.adoc
mvrueden: DONE

PR #2853, opened 2019-12-11
NMS-12406: Support snmp profiles with SnmpDetector

cgorantla:
SnmpDetector would handle multiple agent configs specified in snmp profiles to detect SNMP service on the interface.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12406

### DIFF
--- a/opennms-provision/opennms-detector-simple/src/main/java/org/opennms/netmgt/provision/detector/snmp/GenericSnmpDetectorFactory.java
+++ b/opennms-provision/opennms-detector-simple/src/main/java/org/opennms/netmgt/provision/detector/snmp/GenericSnmpDetectorFactory.java
@@ -29,12 +29,17 @@
 package org.opennms.netmgt.provision.detector.snmp;
 
 import java.net.InetAddress;
+import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 import org.opennms.netmgt.config.api.SnmpAgentConfigFactory;
+import org.opennms.netmgt.config.snmp.SnmpProfile;
 import org.opennms.netmgt.provision.DetectRequest;
+import org.opennms.netmgt.provision.support.AgentBasedSyncAbstractDetector;
 import org.opennms.netmgt.provision.support.DetectRequestImpl;
 import org.opennms.netmgt.provision.support.GenericServiceDetectorFactory;
+import org.opennms.netmgt.snmp.SnmpAgentConfig;
 import org.springframework.beans.factory.annotation.Autowired;
 
 public class GenericSnmpDetectorFactory<T extends SnmpDetector> extends GenericServiceDetectorFactory<SnmpDetector> {
@@ -62,7 +67,22 @@ public class GenericSnmpDetectorFactory<T extends SnmpDetector> extends GenericS
         if (m_agentConfigFactory == null) {
             throw new IllegalStateException("Cannot determine agent configuration without a SnmpAgentConfigFactory.");
         }
-        return m_agentConfigFactory.getAgentConfig(address, location).toMap();
+        List<SnmpProfile> snmpProfiles = m_agentConfigFactory.getProfiles();
+        if (snmpProfiles.isEmpty()) {
+            return m_agentConfigFactory.getAgentConfig(address, location).toMap();
+        } else {
+            Map<String, String> agentConfigMap = new HashMap<>();
+            agentConfigMap.put(AgentBasedSyncAbstractDetector.HAS_MULTIPLE_AGENT_CONFIGS, Boolean.toString(true));
+            // Add default config as string with profile label as key.
+            String defaultConfig = m_agentConfigFactory.getAgentConfig(address, location).toProtocolConfigString();
+            agentConfigMap.put(SnmpAgentConfig.PROFILE_LABEL_FOR_DEFAULT_CONFIG, defaultConfig);
+            // Add snmp profile label as key and config converted to string as value.
+            snmpProfiles.forEach((snmpProfile -> {
+                SnmpAgentConfig snmpAgentConfig = m_agentConfigFactory.getAgentConfigFromProfile(snmpProfile, address);
+                agentConfigMap.put(snmpAgentConfig.getProfileLabel(), snmpAgentConfig.toProtocolConfigString());
+            }));
+            return agentConfigMap;
+        }
     }
 
     public void setAgentConfigFactory(SnmpAgentConfigFactory agentConfigFactory) {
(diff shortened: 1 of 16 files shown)


opennms-provision/opennms-detector-simple/src/main/java/org/opennms/netmgt/provision/detector/snmp/SnmpDetector.java
```diff
@@ -220,6 +221,27 @@ public boolean isServiceDetected(InetAddress address, SnmpAgentConfig agentConfi
...
+            runTimeAttributes.forEach((label, configAsString) -> {
+                agentConfigList.add(SnmpAgentConfig.parseProtocolConfigurationString(configAsString));
+            });
+        }
+        return agentConfigList;
+    }
+
+    @Override
+    public boolean isServiceDetected(InetAddress address, List<SnmpAgentConfig> agentConfigList) {
+        return agentConfigList.parallelStream().anyMatch(snmpAgentConfig -> {
```
cgorantla: Since `SnmpUtils.getOids` and `SnmpUtils.getOid` are static and isServiceDetected doesn't update any state, I think it is safe to use `parallelStream` here.  

Note that since we are updating the profiles list into map and getting back them from map, the order of snmp profiles is not preserved.

opennms-provision/opennms-detector-simple/src/main/java/org/opennms/netmgt/provision/detector/snmp/GenericSnmpDetectorFactory.java
```diff
@@ -62,7 +67,22 @@ public DetectRequest buildRequest(String location, InetAddress address, Integer
         if (m_agentConfigFactory == null) {
             throw new IllegalStateException("Cannot determine agent configuration without a SnmpAgentConfigFactory.");
         }
-        return m_agentConfigFactory.getAgentConfig(address, location).toMap();
+        List<SnmpProfile> snmpProfiles = m_agentConfigFactory.getProfiles();
+        if (snmpProfiles.isEmpty()) {
+            return m_agentConfigFactory.getAgentConfig(address, location).toMap();
+        } else {
+            Map<String, String> agentConfigMap = new HashMap<>();
+            agentConfigMap.put(AgentBasedSyncAbstractDetector.HAS_MULTIPLE_AGENT_CONFIGS, Boolean.toString(true));
```
cgorantla: hasMultipleConfigs can be moved to DetectRequest, don't want to change this in point release.

opennms-provision/opennms-detector-simple/src/main/java/org/opennms/netmgt/provision/detector/snmp/SnmpDetector.java
```diff
@@ -220,6 +221,27 @@ public boolean isServiceDetected(InetAddress address, SnmpAgentConfig agentConfi
...
     }
 
+    @Override
+    public List<SnmpAgentConfig> getListOfAgentConfigs(DetectRequest request) {
+        List<SnmpAgentConfig> agentConfigList = new ArrayList<>();
+        Map<String, String> runTimeAttributes = request.getRuntimeAttributes();
+        if (hasMultipleAgentConfigs(runTimeAttributes)) {
+            runTimeAttributes.remove(HAS_MULTIPLE_AGENT_CONFIGS);
+            //Rest of the attribute values are snmp configs.
+            runTimeAttributes.forEach((label, configAsString) -> {
```
j-white: Can we prefix the keys in the `runTimeAttributes` map with something like `SNMP_AGENT_CONFIG_` and only derive the agent configs from keys matching this prefix?

This would be a bit safer than assuming everything else is an agent config.

opennms-provision/opennms-detector-simple/src/main/java/org/opennms/netmgt/provision/detector/snmp/SnmpDetector.java
j-white: We shouldn't use `parallelStream()` with the default thread pool for blocking operations like these.

opennms-wmi/src/main/java/org/opennms/netmgt/provision/detector/wmi/WmiDetector.java
```diff
@@ -134,7 +135,17 @@ public boolean isServiceDetected(final InetAddress address, final WmiAgentConfig
         return (result != null && result.getResultCode() != WmiResult.RES_STATE_CRIT
                 && result.getResultCode() != WmiResult.RES_STATE_UNKNOWN);
     }
-    
+
+    @Override
```
j-white: Having to implement these methods and throw `UnsupportedOperationException`s is a little odd. Can we implement them in the base class and only override them when needed instead?

cgorantla:
Updated SnmpDetector to use custom threadpool to run detection in parallel and update the results.

jeffgdotorg:
I tried this out on Friday with a Circle build from 12 December:
```
This is an OpenNMS build from the jira/NMS-12406 branch.  For a complete log, see:
https://github.com/OpenNMS/opennms/commit/5c3aa988bec724efdc471c82addadd054c5e27fb
```
With one valid profile and a whole bunch of invalid profiles (started with 80, backed off to 30, then to 20) configured in snmp-config.xml, and a single `<definition>` in discovery-config.xml which uses `ReverseDNSLookupDetector` and `SnmpDetector`, and default-foreign-source.xml pared down to just SNMP and ICMP, I could not get the valid profile to hit.
Furthermore, adding `timeout` and `retries` parameters to the `SnmpDetector` in the discovery configuration did not seem to override the timeout and retries specified in the SNMP profiles.

cgorantla:
> I tried this out on Friday with a Circle build from 12 December:
> 
> ```
> This is an OpenNMS build from the jira/NMS-12406 branch.  For a complete log, see:
> https://github.com/OpenNMS/opennms/commit/5c3aa988bec724efdc471c82addadd054c5e27fb
> ```
> 
> With one valid profile and a whole bunch of invalid profiles (started with 80, backed off to 30, then to 20) configured in snmp-config.xml, and a single `<definition>` in discovery-config.xml which uses `ReverseDNSLookupDetector` and `SnmpDetector`, and default-foreign-source.xml pared down to just SNMP and ICMP, the changes appear to work as designed. Adding `timeout` and `retries` parameters to the `SnmpDetector` in the discovery configuration correctly overrides the timeout and retries specified in the SNMP profiles.
> I did note one unintended consequence. Once the `newSuspect` is created by discovery and Provisiond invokes the SNMP detector, the profiles get fitted again unnecessarily. This results in at least twice the noise and potentially failed SNMP access attempts (many times more on nodes having many IP interfaces). To avoid this situation, we should make the SNMP detector _not_ do profile fitting by default. I think the best way to implement this is by adding a `useProfiles` parameter, with a default value of `false`, to the SnmpDetector. This means that the user must explicitly request profile fitting when it is desired, including in the context of a discovery definition.

Addressed this.  Set `useSnmpProfiles` property to `true` in order to use snmp profiles in SnmpDetector.  By default it is false.

PR #2855, opened 2019-12-11
NMS-12427: Allow telemetry flows to balance across Kafka partitions (alternate)

j-white:
JIRA: https://issues.opennms.org/browse/NMS-12427

Here we use the queue configuration to drive the behavior of the routing key as opposed to using the listener configuration as done in: https://github.com/OpenNMS/opennms/pull/2842

This makes the solution more flexible, allow us to use different properties for different queues and avoids adding the routing-key flag to the actual messages sent.

### DIFF
--- a/features/telemetry/common/src/main/java/org/opennms/netmgt/telemetry/common/ipc/TelemetrySinkModule.java
+++ b/features/telemetry/common/src/main/java/org/opennms/netmgt/telemetry/common/ipc/TelemetrySinkModule.java
@@ -171,6 +171,10 @@ public class TelemetrySinkModule implements SinkModule<TelemetryMessage, Telemet
 
     @Override
     public Optional<String> getRoutingKey(final TelemetryProtos.TelemetryMessageLog message) {
+        // Allow the queue configuration to drive whether or not the routing key is used. Default to true.
+        if (!queueConfig.getUseRoutingKey().orElse(true)) {
+            return Optional.empty();
+        }
         return Optional.of(String.format("%s@%s:%d", message.getLocation(), message.getSourceAddress(), message.getSourcePort()));
     }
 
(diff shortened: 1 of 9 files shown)


j-white:
@Naicisum this is what I was thinking of doing instead. Any concerns?

We can add your `features/telemetry/common/build-proto.sh` script to the tree too for future use.

features/telemetry/common/src/main/java/org/opennms/netmgt/telemetry/common/ipc/TelemetrySinkModule.java
```diff
@@ -171,6 +171,10 @@ public boolean isBlockWhenFull() {
 
     @Override
     public Optional<String> getRoutingKey(final TelemetryProtos.TelemetryMessageLog message) {
+        // Allow the queue configuration to drive whether or not the routing key is used. Default to true.
+        if (!queueConfig.getUseRoutingKey().orElse(true)) {
+            return Optional.empty();
+        }
         return Optional.of(String.format("%s@%s:%d", message.getLocation(), message.getSourceAddress(), message.getSourcePort()));
```
fooker: I had to read it three times to get the negation. Maybe this is simpler to read?
```suggestion
        if (queueConfig.getUseRoutingKey().orElse(true)) {
            return Optional.of(String.format("%s@%s:%d", message.getLocation(), message.getSourceAddress(), message.getSourcePort()));
        } else {
            return Optional.empty();
        }
```

j-white: Ack.

Naicisum:
LGTM 👍

PR #2858, opened 2019-12-16
NMS-12448: Use one builder instead of re-creating it in each enrichment processor

mvrueden:
JIRA: https://issues.opennms.org/browse/NMS-12448


features/graph/api/src/main/java/org/opennms/netmgt/graph/api/enrichment/EnrichmentProcessor.java
```diff
@@ -34,5 +34,5 @@
 
     boolean canEnrich(GenericGraph graph);
 
-    GenericGraph enrich(GenericGraph graph);
+    void enrich(EnrichmentGraphBuilder graphBuilder);
```
j-white: Nice!

PR #2859, opened 2019-12-18
NMS-12455: Upgrade Drools and go back to STREAMing!

j-white:
JIRA: https://issues.opennms.org/browse/NMS-12455

Here we upgrade Drools now that https://issues.jboss.org/browse/DROOLS-4805 is fixed and revert most of the changes done in NMS-12322.

Memory leak appears to be resolved, and system performance is much better when using STREAM vs CLOUD.


j-white:
I've been running this for over a week now in a test environment - no signs of any leakage. Going to go ahead and merge.

PR #2860, opened 2019-12-20
NMS-12464: publish packages to CloudSmith

RangerRick:
This PR updates the CircleCI build to publish packages [to CloudSmith](https://cloudsmith.io/~opennms/repos/) for `foundation-*`, `release-*`, and `develop` branches to the `opennms/foundation-*`, `opennms/testing`, and `opennms/develop` repositories, respectively.

The idea is to enable builds for a live test/evaluation of CloudSmith to potentially replace our existing RPM/Debian infrastructure.

Currently they don't support noarch/"any" repositories for packages that are targeted to multiple releases of a distro, so I am currently using `el5` and `etch` for distros until this feature is implemented (due in January).

I still have some work to get distro-specific packages up there, but this is a start.

### All Contributors

### Pull Request Process

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12464

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -225,8 +225,6 @@ commands:
         default: true
         type: boolean
     steps:
-      - attach_workspace:
-          at: ~/
       - run:
           name: Integration Tests
           no_output_timeout: 1.0h
@@ -613,12 +611,16 @@ jobs:
     executor: integration-test-executor
     parallelism: 4
     steps:
+      - attach_workspace:
+          at: ~/
       - run-integration-tests:
           rerun-failtest-count: 1
   integration-test-with-coverage:
     executor: integration-test-executor
     parallelism: 12
     steps:
+      - attach_workspace:
+          at: ~/
       - run-integration-tests:
           run-code-coverage: true
           rerun-failtest-count: 0
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,6 +13,9 @@ executors:
     machine:
       image: ubuntu-1604:201903-01
 
+orbs:
+  cloudsmith: cloudsmith/cloudsmith@1.0.3
+
 commands:
   cached-checkout:
       description: "Checkout with caching"
@@ -271,6 +274,14 @@ commands:
           when: always
           path: ~/generated-tests
           destination: generated-tests
+  publish-cloudsmith:
+    steps:
+      - cloudsmith/ensure-api-key
+      - cloudsmith/install-cli
+      - run:
+          name: Publish Packages
+          command: |
+            .circleci/scripts/publish-cloudsmith.sh
 
 workflows:
   weekly-coverage:
@@ -395,6 +406,22 @@ workflows:
                 - /^foundation.*/
                 - /^features.*/
                 - /.*smoke.*/
+      - publish-cloudsmith:
+          # technically only requires the RPM/deb builds, but only publish
+          # if everything passes
+          requires:
+            - horizon-deb-build
+            - minion-deb-build
+            - sentinel-deb-build
+            - smoke-test-full
+            - integration-test
+          filters:
+            branches:
+              only:
+                - master
+                - develop
+                - /^release-.*/
+                - /^foundation.*/
 
 jobs:
   build:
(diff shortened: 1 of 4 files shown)


RangerRick:
Note that during development, I ran the task in a normal workflow, so the current circle run won't show what it would look like for real.  [Here](https://circleci.com/workflow-run/65aa872f-7517-43ea-94e9-54a5803b2390) is an example of a successful run.

.circleci/config.yml
```diff
@@ -273,8 +276,61 @@ commands:
...
+    steps:
+      - cloudsmith/ensure-api-key
+      - cloudsmith/install-cli
+      - run:
+          name: Publish Packages
+          command: |
+            .circleci/scripts/publish-cloudsmith.sh
 
 workflows:
+  nightly-publish:
```
j-white: These nightly jobs are petty heavy - these will add 3+ full smoke test and IT runs daily even if there are no changes in the given branches.

Can we find a way to reduce the compute time required?

RangerRick: I wanted to make sure we don't publish something unless it actually truly passes everything.  I guess the alternative would be to always publish the first passing package of the day, and keep track of it somewhere or query the repo?

indigo423: Just some naive quick idea, can we add a gate something like here: https://github.com/opennms-forge/opennms-base-container/blob/master/.circleci/build.sh to just build if we have changes in the given branch?

RangerRick: OK, latest version adds to the "normal" full test builds, but checks against the last commit date of a previously-published branch.  Stuck a temporary commit at the top of the branch to let it run through everything to make sure it does the right thing, then I'll hit it one more time to make sure it does nothing.  :)

RangerRick:
Alright, everything still passed: https://circleci.com/gh/OpenNMS/opennms/17805
You can see the packages [on cloudsmith here](https://cloudsmith.io/~opennms/repos/foundation-2019/packages/).

I just removed the temporary commit that made it publish from this branch.  Everything should be a go to merge if you're cool with the current state of it.

.circleci/config.yml
```diff
@@ -613,12 +638,16 @@ jobs:
     executor: integration-test-executor
     parallelism: 4
     steps:
+      - attach_workspace:
```
j-white: Is this necessary? Add 3+ minutes.

.circleci/config.yml
```diff
@@ -613,12 +638,16 @@ jobs:
...
+          at: ~/
       - run-integration-tests:
           rerun-failtest-count: 1
   integration-test-with-coverage:
     executor: integration-test-executor
     parallelism: 12
     steps:
+      - attach_workspace:
```
j-white: Similarly here.

.circleci/config.yml
```diff
@@ -660,4 +689,10 @@ jobs:
           at: ~/
       - run-smoke-tests:
           minimal: true
+  publish-cloudsmith:
+    executor: cloudsmith/default
+    steps:
+      - attach_workspace:
+          at: ~/
+      - publish-cloudsmith
```
j-white: 30+ minutes to publish the packages seems like a long time. Is is because of large packages and slow upload speeds? :D

.circleci/config.yml
RangerRick: Hrm, maybe I just don't understand circle enough.

It was in the _command_ before, I just moved it to the _job_ entries that call that command to match the way all the other jobs are set up -- but it should have been happening either way.  (Those jobs call the command, so the attach should happen once per job whether it's defined in the job or the command, right?)

I can take it out, it's not directly related to my changes, I just ran into that inconsistency while I was working on the config.

.circleci/config.yml
RangerRick: Not really sure where the bottleneck is, but yeah, it really does seem to take that long.  I could split it into RPM and Deb but I was trying to avoid another big attach that eats time.

.circleci/config.yml
j-white: My mistake, let's keep it.

.circleci/config.yml
j-white: Ack.

We could consider using the [small resource class](https://circleci.com/docs/2.0/configuration-reference/#resource_class) here if it doesn't affect the performance too much - that would make these runs less costly.

RangerRick: Good point.  It shouldn't need bigger, it's all network operations really.

PR #2865, opened 2020-01-10
NMS-12460: Merged poll and test command

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-12460


agalue:
In terms of functionality, both methods work as expected; either using the class name and the parameters or the package and the service.

I also tried overriding parameters when using package/service and it works very well.

My only observation is what I mentioned in the short description you would see on the Karaf shell, and of course, the [official docs](https://meridian.opennms.com/docs/2019/latest/guide-admin/#ga-pollerd-packages-test-service-manually), especially because there is no more `poller:test`, now, it is either just `poll` or `opennms-poller:poll`.

features/poller/shell/src/main/java/org/opennms/netmgt/poller/shell/Poll.java
```diff
@@ -44,60 +47,147 @@
...
     public LocationAwarePollerClient locationAwarePollerClient;
 
+    @Reference
+    public NodeDao nodeDao;
+
+    @Reference
+    public IpInterfaceDao ipInterfaceDao;
+
+    @Reference
+    public TransactionOperations transactionTemplate;
```
mvrueden: Can you use `SessionUtils` instead? 
Using `TransactionOperations` may result in weird bundle states when `bundle:watch *` is enabled or a bundle is trying to refresh.

PR #2867, opened 2020-01-14
NMS-12475: remove threshd entry in log4j2.xml

mfuhrmann:
* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-12475


agalue:
Why do you want to remove the entry for vacuumd? That daemon is still available and customers are using it, despite the fact that the alarm life-cycle is now managed by Alarmd using Drools.

agalue:
Keep in mind that even if it makes sense removing the entry for threshd because it doesn't exist anymore, it is not strictly necessary. Log4j2 will ignore it.

mfuhrmann:
> Why do you want to remove the entry for vacuumd? That daemon is still available...

@agalue Lack of knowledge :-) I thought is was already removed. Will add it again.

indigo423:
@agalue set target branch to 26.0.0, what do you think?

agalue:
> @agalue set target branch to 26.0.0, what do you think?

Yes, that sounds good.

PR #2868, opened 2020-01-14
NMS-10586: Sink API: Persistent Off-Heap Storage

mattixtech:
This PR updates the off-heap functionality of the Sink process on Minion. The H2 db implementation that was using off-heap memory is now removed. It is replaced by a file based fifo queue to avoid using memory and allowing us to store much larger amounts of messages off-heap.

The new implementation is installed and active by default. It can optionally be disabled by setting the max file size to 0.

Additionally, fixed a memory leak where we were storing futures on-heap for off-heap messages.

Tested on Minion using udpgen. I was seeing a throughput of 10-15MB/s to the disk, limited mostly by the time to serialize SNMP trap objects.

### All Contributors

### Pull Request Process

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-10586


core/ipc/sink/common/src/main/java/org/opennms/core/ipc/sink/common/AsyncDispatcherImpl.java
```diff
@@ -67,226 +65,162 @@
...
-     *
-     * If the implementation is changed, make sure that that executor is built accordingly.
-     */
-    
-    private static class OfferBlockingQueue<E> extends LinkedBlockingQueue<E> {
-        private static final long serialVersionUID = 1L;
+    private void dispatchFromQueue() {
+        while (true) {
+            try {
+                RATE_LIMITED_LOGGER.trace("Asking dispatch queue for the next entry...");
```
j-white: We can use the standard logger for these instead of the `RATE_LIMITED_LOGGER`. The `RATE_LIMITED_LOGGER` is intended to be used to avoid flooding the log with error messages.

core/ipc/sink/off-heap/src/main/java/org/opennms/core/ipc/sink/offheap/QueueFileOffHeapDispatchQueueFactory.java
```diff
@@ -0,0 +1,117 @@
...
+    public <T> DispatchQueue<T> getQueue(AsyncPolicy asyncPolicy, String moduleName, Function<T, byte[]> serializer,
+                                         Function<byte[], T> deserializer) {
+        if (asyncPolicy.getNumThreads() > inMemoryEntrySize) {
+            throw new IllegalArgumentException("The in memory queue size must be greater than or equal to the number" +
+                    " of consuming threads");
+        }
+
+        return (DispatchQueue<T>) queues.computeIfAbsent(moduleName, (k) -> {
+            try {
+                // TODO: Configurable
```
j-white: TODO

core/ipc/sink/off-heap/src/main/java/org/opennms/core/ipc/sink/offheap/QueueFileOffHeapDispatchQueueFactory.java
```diff
@@ -0,0 +1,117 @@
...
+        switch (suffix) {
+            case "kb":
+                bytes = (long) value * 1024;
+                break;
+            case "mb":
+                bytes = (long) value * 1024 * 1024;
+                break;
+            case "gb":
+                bytes = (long) value * 1024 * 1024 * 1024;
+                break;
```
j-white: What's the expect behavior for unknown suffixes?

core/ipc/sink/off-heap/src/main/java/org/opennms/core/ipc/sink/offheap/QueueFileOffHeapDispatchQueueFactory.java
```diff
@@ -0,0 +1,117 @@
...
+            }
+        });
+    }
+
+    private static long convertToBytes(String sizeWithSuffix) {
+        if (sizeWithSuffix == null || sizeWithSuffix.length() == 0) {
+            return 0;
+        }
+
+        String suffix = sizeWithSuffix.substring(sizeWithSuffix.length() - 2).toLowerCase();
```
j-white: Will cause error if length is 1

j-white: Might be a good idea to trim trailing white-space off of this too

core/ipc/sink/api/src/main/java/org/opennms/core/ipc/sink/api/AsyncDispatcher.java
```diff
@@ -40,13 +40,12 @@
 public interface AsyncDispatcher<S extends Message> extends AutoCloseable {
```
bouff: Update copyright year.

core/ipc/sink/off-heap/src/main/java/org/opennms/core/ipc/sink/offheap/QueueFileOffHeapDispatchQueueFactory.java
mattixtech: Updated to throw an error.

core/ipc/sink/api/src/main/java/org/opennms/core/ipc/sink/api/WriteFailedException.java
```diff
@@ -35,4 +35,9 @@
     public WriteFailedException(String message) {
         super(message);
     }
+
```
bouff: Copyright year for this one as well.

core/ipc/sink/off-heap/src/main/java/org/opennms/core/ipc/sink/offheap/QueueFileOffHeapDispatchQueueFactory.java
mattixtech: Catching that error and logging now.

core/ipc/sink/common/src/main/java/org/opennms/core/ipc/sink/common/AsyncDispatcherImpl.java
mattixtech: Ok should I apply this to the trace logging in `QueueFileOffHeapDispatchQueue` as well?

j-white: Yeah, I would do the same there too.

core/ipc/sink/off-heap/src/test/java/org/opennms/core/ipc/sink/offheap/QueueFileOffHeapDispatchQueueTest.java
```diff
@@ -0,0 +1,192 @@
...
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+import org.opennms.core.ipc.sink.api.DispatchQueue;
+import org.opennms.core.ipc.sink.api.WriteFailedException;
+
+import com.jayway.awaitility.core.ConditionTimeoutException;
+
+public class QueueFileOffHeapDispatchQueueTest {
```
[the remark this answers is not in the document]
mattixtech: Updated one of the test cases to pay attention to the return type.

opennms-doc/guide-admin/src/asciidoc/text/minion/offheap.adoc
```diff
@@ -3,35 +3,31 @@
 
 === Using Off-heap Storage for Sink Messages
 
-If a _Minion_ loses connectivity with the broker (i.e. _Kafka_ or _ActiveMQ_), then any received messages (i.e. syslog, flows, SNMP traps) are queued in the JVM heap until connectivity is restored.
-This queue is limited by a fixed (and configurable) number of messages.
+If a _Minion_ loses connectivity with the broker (i.e. _Kafka_ or _ActiveMQ_), then any received messages (i.e. syslog, flows, SNMP traps) are queued until connectivity is restored.
+This queue is limited by a fixed (and configurable) number of messages queued in the JVM heap and can optionally queue
+additional messages directly to the filesystem to avoid using heap memory.
```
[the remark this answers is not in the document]
mattixtech: Not sure I understand the update you want me to make. FYI you can submit your comment on github as a suggestion which will give a direct diff to the author of the change you want them to make.

mattixtech:
Rebased and force pushed to pick up the fix for the date parsing.

opennms-doc/guide-admin/src/asciidoc/text/minion/offheap.adoc
```diff
@@ -3,35 +3,31 @@
...
-Install off-heap feature on _Minion_:
+Configure storage limits:
 
 [source, sh]
 ----
-echo 'opennms-core-ipc-sink-offheap' >> "$MINION_HOME/etc/featuresBoot.d/features.boot"
+echo 'offHeapSize = 1GB
+entriesAllowedOnHeap = 100000
+offHeapFilePath =' > "$MINION_HOME/etc/org.opennms.core.ipc.sink.offheap.cfg"
 ----
```
cgorantla: We should probably add a note saying that this file size is per module.

core/ipc/sink/off-heap/src/main/java/org/opennms/core/ipc/sink/offheap/QueueFileOffHeapDispatchQueue.java
```diff
@@ -0,0 +1,537 @@
...
+                if (offHeapQueue.size() > 0 && inMemoryQueue.remainingCapacity() >= batchSize) {
+                    LOG.trace("Found an entry off-heap and there was room in-memory, moving it");
+
+                    try {
+                        byte[] entry = offHeapQueue.peek();
+                        if (entry != null) {
+                            offHeapQueue.remove();
+
+                            try {
+                                inMemoryQueue.addAll(unbatchSerializedBatch(new SerializedBatch(entry)));
```
cgorantla: I'm not sure if I understand this code completely.   If there is only one thread that does `dequeue` the order should be maintained. Would like to see a test case.

mattixtech: Good idea, I will add a test case to verify the ordering. Since batches are synchronized the order of contents should be guaranteed to be in the same order.

Let me know if you want me to elaborate on how anything works.

core/ipc/sink/off-heap/src/test/java/org/opennms/core/ipc/sink/offheap/QueueFileOffHeapDispatchQueueTest.java
```diff
@@ -0,0 +1,194 @@
...
+
+    @Rule
+    public TemporaryFolder folder = new TemporaryFolder();
+
+    @Test
+    public void canQueueAndDequeue() throws IOException, WriteFailedException, InterruptedException {
```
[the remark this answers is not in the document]
mattixtech: Will do.

core/ipc/sink/off-heap/src/main/java/org/opennms/core/ipc/sink/offheap/QueueFileOffHeapDispatchQueue.java
```diff
@@ -0,0 +1,537 @@
...
+            }
+
+            checkFileSize();
+        } else {
+            offHeapQueue = null;
+            usedBytesMethod = null;
+        }
+    }
+
+    @VisibleForTesting
```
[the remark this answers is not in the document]
mattixtech: The only other class that uses it is a test class. It would normally be private otherwise so the annotation is there to explain why the access modifier is not private.

core/ipc/sink/common/src/main/java/org/opennms/core/ipc/sink/common/AsyncDispatcherImpl.java
```diff
@@ -67,226 +65,162 @@
...
+                    CompletableFuture<DispatchStatus> messageFuture = futureMap.remove(messageEntry.getKey());
+
+                    if (messageFuture != null) {
+                        messageFuture.complete(DispatchStatus.DISPATCHED);
+                        LOG.trace("Completed future for message {}", messageEntry);
+                    } else {
+                        RATE_LIMITED_LOGGER.warn("No future found for message {}", messageEntry);
+                    }
+                } else {
+                    LOG.trace("Dequeued an entry with a null key");
```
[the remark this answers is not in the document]
mattixtech: This will happen any time we get an off-heap entry since we don't bother keeping keys for them as we don't associate a future with them.

core/ipc/sink/off-heap/src/main/java/org/opennms/core/ipc/sink/offheap/QueueFileOffHeapDispatchQueue.java
cgorantla: Adding a test case with some data on file, some data in memory would be sufficient. Then I can play with that and understand the logic better.

core/ipc/sink/common/src/main/java/org/opennms/core/ipc/sink/common/AsyncDispatcherImpl.java
```diff
@@ -67,226 +65,162 @@
...
-     */
-    
-    private static class OfferBlockingQueue<E> extends LinkedBlockingQueue<E> {
-        private static final long serialVersionUID = 1L;
+    private void dispatchFromQueue() {
+        while (true) {
+            try {
+                LOG.trace("Asking dispatch queue for the next entry...");
+                Map.Entry<String, S> messageEntry = dispatchQueue.dequeue();
+                LOG.trace("Received message entry from dispatch queue {}", messageEntry);
```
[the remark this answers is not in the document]
mattixtech: AFAIK the toString() will only be called if trace is enabled so adding the conditional around the message won't make a difference.

opennms-doc/guide-admin/src/asciidoc/text/minion/offheap.adoc
mattixtech: Will do.

core/ipc/sink/off-heap/src/main/java/org/opennms/core/ipc/sink/offheap/QueueFileOffHeapDispatchQueue.java
mattixtech: Looks like there was an issue with the ordering, will fix. Thanks for catching this.

mattixtech:
Thanks for the feedback, there was some issues with dequeue order. Also found some other potential threading issues in the process of testing.

core/ipc/sink/off-heap/src/main/java/org/opennms/core/ipc/sink/offheap/QueueFileOffHeapDispatchQueue.java
cgorantla: If I understand this correctly,  If `dequeue` is called from two different threads, it would pull two different batches of messages in parallel then order may not be maintained when they dispatch in two different threads.

mattixtech: All calls to `dequeue` end up taking from the inMemoryQueue data structure and ordering is ensured there by guarding manipulation of that queue with a mutex. It is not possible for two threads to be simultaneously retrieving two different batches at once.

This means that entries are always dequeued in order.

However it is still possible to dispatch entries in a different order since there is multiple consuming threads calling `dequeue` since all the threads ultimately race to finish dispatching. If we need that to be guaranteed to be in order we need make sure there is only one thread consuming from the queue which will likely drastically impact the performance of the AsyncDispatcher.

Note that that race condition existed in the previous implementation so this PR is not violating any previously established ordering guarantee.

cgorantla: > All calls to `dequeue` end up taking from the inMemoryQueue data structure and ordering is ensured there by guarding manipulation of that queue with a mutex. It is not possible for two threads to be simultaneously retrieving two different batches at once.
> 
> This means that entries are always dequeued in order.
> 
> However it is still possible to dispatch entries in a different order since there is multiple consuming threads calling `dequeue` since all the threads ultimately race to finish dispatching. If we need that to be guaranteed to be in order we need make sure there is only one thread consuming from the queue which will likely drastically impact the performance of the AsyncDispatcher.
> 
> Note that that race condition existed in the previous implementation so this PR is not violating any previously established ordering guarantee.

Got it. I think it does maintain order. Can we modify test case for verifying that multiple threads doing `dequeue` also maintains order. Probably synchronize both `dequeue` and putting elements back.

I think we don't need to worry about dispatching in order since even after pushing them to kafka, there is no guarantee in order.

mattixtech: Multiple threads calling dequeue won't necessarily be in order since there is a race condition on which thread will get to `take()` from the queue first. This should be acceptable since we don't have an ordering guarantee on the consumption side anyways (dispatching).

The queue itself will dispense items in the correct order but there is no way of knowing the order the threads will actually dequeue in so there really isn't any predictable behaviour to test without synchronizing the entire dequeue & process task.

cgorantla: yeah, the dequeue method is not synchronized. That's fine then.

mattixtech:
All feedback should be addressed now, let me know if there is anything else you see. Thanks!

mattixtech:
Rebased for the maven url fix.

core/ipc/sink/common/src/main/java/org/opennms/core/ipc/sink/common/AsyncDispatcherImpl.java
```diff
@@ -67,226 +65,162 @@
...
-            rejectedExecutionHandler = new ThreadPoolExecutor.AbortPolicy();
+        SinkModule<S, T> sinkModule = state.getModule();
+        Optional<DispatchQueueFactory> factory = DispatchQueueServiceLoader.getDispatchQueueFactory();
+
+        if (factory.isPresent()) {
+            LOG.debug("Using queue from factory");
+            dispatchQueue = factory.get().getQueue(asyncPolicy, sinkModule.getId(),
+                    sinkModule::marshalSingleMessage, sinkModule::unmarshalSingleMessage);
         } else {
-            queue = new LinkedBlockingQueue<Runnable>(asyncPolicy.getQueueSize());
```
cgorantla: Consider using `asyncPolicy.getQueueSize()` as the default inMemoryQueueSize for the specific module. We can extend AsyncPolicy to set `offHeapSize` then everything is associated with module

core/ipc/sink/off-heap/src/main/java/org/opennms/core/ipc/sink/offheap/QueueFileOffHeapDispatchQueue.java
```diff
@@ -0,0 +1,560 @@
...
+            QueueFile qf;
+            try {
+                qf = new QueueFile.Builder(file).build();
+            } catch (Exception e) {
+                LOG.warn("Exception while loading queue file", e);
+
+                if (file.delete()) {
+                    qf = new QueueFile.Builder(file).build();
+                } else {
+                    throw new IOException("Could delete corrupted queue file " + file.getAbsolutePath());
```
cgorantla: I think the exception should be  `Could not delete`

mattixtech: Yep it should be.

core/ipc/sink/common/src/main/java/org/opennms/core/ipc/sink/common/AsyncDispatcherImpl.java
mattixtech: It should be easy to re-use the queue size value but what about the batch size and max file size? Do they belong in the XML configuration for the module? They are implementation specific details for this queueing strategy. Do you think it makes sense to move those into the per-module XML as well?

Another thing to think about... the original meaning of the queue size setting is not really applicable in the case of off-heap storage. For instance if the queue gets filled up it doesn't mean we are going to start rejecting or blocking, we will just be going off-heap. This was the original reasoning why I didn't re-use the existing config.

cgorantla: Yeah, I was thinking that it makes sense to have module specific configuration for off-heap as well so that we can change off-heap size based on needs of module.  

But looking at configuration more closely, it may be confusing to add `batchSize` and `queueSize` in module configuration as they are also used for `aggregation`.  Let's keep them separate for now.

mattixtech: Ok thanks for the feedback. I think ultimately what we would want is an aggregate max file size instead of a per-module file size so the user could choose to use a max of XGB across all modules. To keep it simple I just went with the fixed file size  per module for now.

mattixtech:
@cgorantla let me know if I missed anything or if there is any other feedback

@bouff Can you update your review to approved if the changes are satisfactory

Thanks

cgorantla:
Looks good to me.

I opened an issue for handling `off-heap` across minion restarts https://issues.opennms.org/browse/NMS-12485

mattixtech:
Looks like smoke tests failed on bamboo could be related to the changes, will have to take a look.

mattixtech:
> Just to clarify, was `DefaultQueue` not sufficient for sentinel as it only dispatches internal events.

My intention was that the `DefaultQueue` would only be used for testing. On sentinel it was blowing up before it could get to creating a default queue because Karaf was trying to find an implementation of the dispatch queue factory service.

core/ipc/sink/common/src/main/java/org/opennms/core/ipc/sink/offheap/DispatchQueueServiceLoader.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    public static Optional<DispatchQueueFactory> getDispatchQueueFactory() {
+        if (dispatchQueueFactory != null) {
+            return Optional.of(dispatchQueueFactory);
+        }
+
+        if (context != null) {
+            try {
+                dispatchQueueFactory = context.getService(context.getServiceReference(DispatchQueueFactory.class));
```
cgorantla: I think you have to do null check on the reference before passing it to  `context.getService`.  In the previous case, it used to check this only when off-heap feature is enabled config is present.

mattixtech: Yeah you are right. I'll add that too. I think it is still preferably to install the feature on sentinel additionally so we use the same impl.

mattixtech:
Will retarget this to 25 before merging.

PR #2870, opened 2020-01-15
NMS-12443: Fix getTopology() is blocking when discovery is in progress

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-12443

### DIFF
--- a/features/enlinkd/adapters/common/src/main/java/org/opennms/netmgt/enlinkd/common/TopologyUpdater.java
+++ b/features/enlinkd/adapters/common/src/main/java/org/opennms/netmgt/enlinkd/common/TopologyUpdater.java
@@ -139,37 +139,49 @@ public abstract class TopologyUpdater extends Discovery implements OnmsTopologyU
     @Override
     public synchronized void runDiscovery() {
         LOG.debug("run: start {}", getName());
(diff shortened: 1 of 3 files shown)


features/enlinkd/adapters/common/src/main/java/org/opennms/netmgt/enlinkd/common/TopologyUpdater.java
```diff
@@ -217,8 +229,10 @@ private static IpInterfaceTopologyEntity getPrimary(IpInterfaceTopologyEntity n1
     public abstract OnmsTopology buildTopology();
 
     @Override
-    public synchronized OnmsTopology getTopology() {
-        return m_topology.clone();
+    public OnmsTopology getTopology() {
+        synchronized(m_topology) {
```
j-white: We may want to add a synchronized block in the setter too.

mvrueden: Yep, good point :)

mvrueden:
I originally mentioned (in Sprint meeting) that there is one other thing to look at. I did that and this is now good to go if it goes green.

PR #2873, opened 2020-01-20
NMS-12453: Allow exposing status information (New Graph API)

mvrueden:
Here we provide a foundation to expose status information.
At the moment only the status for applications are enriched.
Others can be added very easily.

The solution is very flexible and generic.
Besides severity and count, custom properties can be added as well.
This allows to be compatible even with the old Topology where we expose some custom styling details via the status object.

JIRA: https://issues.opennms.org/browse/NMS-12453


features/graph/api/pom.xml
```diff
@@ -32,6 +32,10 @@
             <artifactId>org.opennms.features.graph.jung</artifactId>
             <version>${jungVersion}</version>
         </dependency>
+        <dependency>
+            <groupId>org.opennms</groupId>
+            <artifactId>opennms-model</artifactId>
```
j-white: This is a pretty heavy dependency to add to the API module :)

Looks like there's no easy way around it if we want to use the `OnmsSeverity` model though.

smoke-test/src/test/java/org/opennms/smoketest/graph/GraphRestServiceIT.java
```diff
@@ -361,6 +378,136 @@ public void verifyNodeEnrichment() throws InterruptedException {
...
+                    .contentType(ContentType.JSON)
+                    .content("vertices", Matchers.hasSize(1))
+                    .content("vertices[0].status", Matchers.is("Normal"));
+        } finally {
+            karafShell.runCommand("opennms-bsm:delete-generated-hierarchies");
+        }
+    }
+
+    @Test
+    public void verifyStatusEnrichmentApplication() {
```
j-white: 👍

PR #2874, opened 2020-01-21
NMS-12479: Remove module name from Kafka RPC topic names.

cgorantla:
By moving module name to proto from the topic names, we can save the number of topics by order of number of RPC modules.

Remove the need for module name in topic names.
Include topic name in proto message.
upgrade kafka-rpc.proto to proto3.
Use kafka specific topic name generation.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12479


core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcServerManager.java
```diff
@@ -258,40 +256,24 @@ public void run() {
...
-                                } else {
-                                    // No exception occurred, use the given response
-                                    response = res;
-                                }
-                                // Finish minion Span
-                                minionSpan.finish();
-                                sendResponse(rpcId, response);
-                            });
+                            final RpcModule module = modulesById.get(rpcMessage.getModuleId());
+                            if(module == null) {
```
j-white: This is a problem - if a Minion doesn't have a module loaded it still consumes the relevant requests and drops them silently.

cgorantla: yeah.  If you are considering a scenario where certain modules are loaded on one minion and rest of the modules are loaded on another minion, that may not work with this approach.

cgorantla: I realized that we can't disable `RPC` as that would disable Echo module as well which is needed  for maintaining minion status.  We can still create a topic specific to Echo and subscribe to it by default.  subscribe to generic `rpc-request` topic only if at least one of the other `rpc` modules are loaded.  Let me know what do you think.

cgorantla: Or introduce two more Minion Status, `RPC_UP`, `SINK_UP`

PR #2883, opened 2020-02-06
NMS-12422: Allow multiple values for ip address input

mvrueden:
Here we fix an issue with the ip src and dest fields when defining Classifcation Rules.
Somehow it was possible to define an IP-Like expression using commas. However that was a bit misleading and didn't work under the hood. Instead it is now possible to define actual IP-Ranges, e.g. `10.0.0.0-10.255.255.255`, as well as multiple definitions at once:
`192.168.0.1,10.0.0.0-10.255.255.255`

Support for wildcard entries was dropped.

A next step is to allow CIDR expressions (which is not done here)

It may be possible that invalid rules exist and when applying this patch to a system the classification engine may not load. 

~At the moment this is targeted for `develop` but should go into `foundation-2019`~

JIRA: https://issues.opennms.org/browse/NMS-12422

### DIFF
--- a/features/flows/classification/engine/api/src/main/java/org/opennms/netmgt/flows/classification/error/Errors.java
+++ b/features/flows/classification/engine/api/src/main/java/org/opennms/netmgt/flows/classification/error/Errors.java
@@ -54,6 +54,8 @@ public interface Errors {
     ErrorTemplate RULE_EXPORTER_FILTER_INVALID = new ErrorTemplate("rule.filter.invalid", "The provided filter ''{0}'' is invalid: {1}");
 
     ErrorTemplate RULE_IP_ADDRESS_INVALID = new ErrorTemplate("rule.ipaddress.invalid", "The provided IP Address ''{0}'' is not valid.");
+    ErrorTemplate RULE_IP_ADDRESS_RANGE_INVALID = new ErrorTemplate("rule.ipaddress.range.invalid", "The provided IP Address range ''{0}'' is not valid.");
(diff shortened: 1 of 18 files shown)


mvrueden:
As this breaks things, the current behaviour is, that OpenNMS does not come up at all as the Classification Engine crashes. This also means it is not possible to fix any possible rule problems using the UI. I would suggest to simply ignore invalid rules from the classification engine and have OpenNMS come up (maybe with a bunch of error log statements). A user can then fix the problematic rules in the UI. We could also try to add a UI indicator for invalid rules. What do you think?

features/flows/classification/engine/api/src/main/java/org/opennms/netmgt/flows/classification/error/Errors.java
```diff
@@ -54,6 +54,8 @@
     ErrorTemplate RULE_EXPORTER_FILTER_INVALID = new ErrorTemplate("rule.filter.invalid", "The provided filter ''{0}'' is invalid: {1}");
 
     ErrorTemplate RULE_IP_ADDRESS_INVALID = new ErrorTemplate("rule.ipaddress.invalid", "The provided IP Address ''{0}'' is not valid.");
+    ErrorTemplate RULE_IP_ADDRESS_RANGE_INVALID = new ErrorTemplate("rule.ipaddress.range.invalid", "The provided IP Address range ''{0}'' is not valid.");
```
patrick-schweizer: I wonder if we should give here an example of how a valid range should look like?

mvrueden:
I am going ahead and will merge this.

PR #2887, opened 2020-02-13
NMS-12526: Enrich content of nodeAdded event

mfuhrmann:
I've rewritten the content of nodeAdded event because the old text was a little too simplistic in my opinion.
I've also added the operator instruction field because I really like it if there are instructions for an user. But I'm pretty sure that we don't have default operator instructions right now in the default OpenNMS events. Is there any reason why we shouldn't?

Also I'm wondering why this event is stored in the poller event file. Is that correct? Which component is responsible for `uei.opennms.org/nodes/` events?

### External References

* https://issues.opennms.org/browse/NMS-12526

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/events/opennms.pollerd.events.xml
+++ b/opennms-base-assembly/src/main/filtered/etc/events/opennms.pollerd.events.xml
@@ -214,11 +214,9 @@
    <event>
       <uei>uei.opennms.org/nodes/nodeAdded</uei>
       <event-label>OpenNMS-defined node event: nodeAdded</event-label>
-      <descr>&lt;p>A new node (%parm[nodelabel]%) was discovered by
-            OpenNMS.&lt;/p></descr>
-      <logmsg dest="logndisplay">
-            A new node (%parm[nodelabel]%) was discovered by OpenNMS.
-        </logmsg>
+      <descr>The node (%parm[nodelabel]%) was added and is now being monitored.</descr>
+      <logmsg dest="logndisplay">A new node (%parm[nodelabel]%) was added.</logmsg>
+      <operinstruct>This event is for information only. Please make sure that the newly added device &lt;a href=&quot;element/node.jsp?node=%nodeid%&quot;>%nodeid%&lt;/a> is monitored as desired.</operinstruct>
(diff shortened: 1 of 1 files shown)


opennms-base-assembly/src/main/filtered/etc/events/opennms.pollerd.events.xml
```diff
@@ -214,11 +214,9 @@
...
       <uei>uei.opennms.org/nodes/nodeAdded</uei>
       <event-label>OpenNMS-defined node event: nodeAdded</event-label>
-      <descr>&lt;p>A new node (%parm[nodelabel]%) was discovered by
-            OpenNMS.&lt;/p></descr>
-      <logmsg dest="logndisplay">
-            A new node (%parm[nodelabel]%) was discovered by OpenNMS.
-        </logmsg>
+      <descr>The node (%parm[nodelabel]%) was added and is now being monitored.</descr>
+      <logmsg dest="logndisplay">A new node (%parm[nodelabel]%) was added.</logmsg>
+      <operinstruct>This event is purely informative. Please make sure that the newly added device &lt;a href=&quot;element/node.jsp?node=%nodeid%&quot;>%nodeid%&lt;/a> is monitored as desired.</operinstruct>
```
christianpape: Evtl. wäre "This is for information only" schöner.

indigo423:
Changed target to release-26.0.0 @christianpape @RangerRick @agalue any objections?

RangerRick:
Now's the time!

RangerRick:
Although maybe instead of parentheses it should be quotes in the `descr`?  It sounds a little weird to say:

> The node (foo) was added

Whereas before it was:

> A new node (foo) was discovered

(where the "foo" was an interjection)

mfuhrmann:
Thanks @indigo423 @RangerRick. I've did the suggested changes.
Now it looks like this:
grafik

mfuhrmann:
Regarding to:

> Also I'm wondering why this event is stored in the poller event file. Is that correct? Which component is responsible for uei.opennms.org/nodes/ events?

There is also `dataCollectionFailed` in the poller event file. I will create an other issue to correct them all.

PR #2890, opened 2020-02-17
NMS-12513: Fix HQL Injection issue

mvrueden:
Here we address an HQL Injection issue when using the nodeList page.

* JIRA: https://issues.opennms.org/browse/NMS-12513


opennms-web-api/src/main/java/org/opennms/web/svclayer/support/DefaultNodeListService.java
```diff
@@ -189,8 +191,16 @@ private static void addCriteriaForSnmpParm(OnmsCriteria criteria,
         if(snmpParmMatchType.equals("contains")) {
             criteria.add(Restrictions.ilike("snmpInterface.".concat(snmpParm), snmpParmValue, MatchMode.ANYWHERE));
         } else if(snmpParmMatchType.equals("equals")) {
+            final List<String> acceptedParamNames = Lists.newArrayList(
```
j-white: Using a `Set` here instead of a list would be more efficient for lookups.

We should also move to a static constant rather than defining it in the function here.

PR #2892, opened 2020-02-18
NMS-12505: Provide ID and IpInterface in JSON results

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-12505


opennms-model/src/main/java/org/opennms/netmgt/model/OnmsMonitoredService.java
```diff
@@ -416,14 +421,20 @@ public void removeMetaData(final String context) {
...
     @XmlElement(name="ipInterfaceId")
     @ManyToOne(optional=false, fetch=FetchType.LAZY)
     @JoinColumn(name="ipInterfaceId")
     public OnmsIpInterface getIpInterface() {
         return m_ipInterface;
     }
 
+    @JsonProperty("ipInterfaceId")
+    @Transient
+    public Integer getFoo() {
```
mvrueden: Fu und Fara oder was?

Should probably be named differently :D

opennms-model/src/main/java/org/opennms/netmgt/model/OnmsMonitoredService.java
```diff
@@ -416,14 +421,20 @@ public void removeMetaData(final String context) {
...
+        return m_ipInterface.getId();
```
mvrueden: Can `m_ipInterface` ever be null?

opennms-model/src/main/java/org/opennms/netmgt/model/OnmsMonitoredServiceDetail.java
```diff
@@ -73,6 +75,7 @@ public OnmsMonitoredServiceDetail(OnmsMonitoredService service) {
         m_statusCode = service.getStatus();
         m_status = service.getStatusLong();
         m_id = service.getXmlId();
+        m_ipInterfaceId = service.getIpInterface().getId();
```
mvrueden: Do we need to store this as a field, can't we just return that in the `getIpinterfaceId()` method?

opennms-model/src/main/java/org/opennms/netmgt/model/OnmsMonitoredServiceDetail.java
```diff
@@ -140,6 +143,11 @@ public void setId(String id) {
         m_id = id;
     }
 
+    @XmlAttribute(name="ipInterfaceId")
+    public Integer getIpinterfaceId() {
```
mvrueden: Should be `getIpInterfaceId()` instead.

opennms-model/src/main/java/org/opennms/netmgt/model/OnmsMonitoredService.java
christianpape: Yes, of course.

opennms-model/src/main/java/org/opennms/netmgt/model/OnmsMonitoredService.java
christianpape: I don't think so.

opennms-model/src/main/java/org/opennms/netmgt/model/OnmsMonitoredServiceDetail.java
christianpape: done

opennms-model/src/main/java/org/opennms/netmgt/model/OnmsMonitoredServiceDetail.java
christianpape: done

PR #2893, opened 2020-02-18
NMS-12473: Add persistent local engine Id for Snmp Informs

cgorantla:
Modify the way of creating session to include fixed local engine Id.

Some Snmp engines may not have ability to discover authoritative
engine id. They will need to have a fixed engine id that is
persistent across engine reboots. This is necessary for Informs
as Snmp  uses receiver's engine id as authoritative engine Id.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12473


cgorantla:
Verified this with `snmpinform` utility using packet capture to check that authoritative engine id is consistent across reboots.

PR #2894, opened 2020-02-18
NMS-12531: Fix geolocation info panel not showing

mvrueden:
* JIRA: https://issues.opennms.org/browse/NMS-12531


mvrueden:
@christianpape reported an issue with this. Will take a look.

PR #2897, opened 2020-02-24
NMS-10413: Use 'opennms' scope for all Karaf shell commands

j-white:
JIRA: https://issues.opennms.org/browse/NMS-10413

Updates the scope of all Karaf shell commands to be 'opennms'.
Removes the 'opennms-minion:ping' command in favor of using the health check command instead.
Add mapping of old to new command names in the release notes.


cgorantla:
LGTM

PR #2903, opened 2020-02-26
NMS-12521: Use protobuf instead of bson for Netflow

cgorantla:
Move the serialization from bson to protobuf for all Netflow related protocols.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12521


features/telemetry/api/pom.xml
```diff
@@ -45,5 +50,10 @@
       <groupId>io.dropwizard.metrics</groupId>
       <artifactId>metrics-core</artifactId>
     </dependency>
+    <dependency>
```
j-white: I would like to try and maintain avoiding having knowledge of specific payloads in the Telemetry API.

I realize this may be required if we want to combine the protobuf definitions though and avoid intermediary byte arrays though. IMO it would only be worth doing this if we we're able to show some associated performance benefits.

features/telemetry/protocols/bmp/adapter/src/test/java/org/opennms/netmgt/telemetry/protocols/bmp/adpater/routes/BmpRoutesAdapterTest.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.concurrent.CompletableFuture;
+
+import org.junit.Test;
+import org.opennms.core.ipc.sink.api.AsyncDispatcher;
+import org.opennms.netmgt.telemetry.api.receiver.TelemetryMessage;
+import org.opennms.netmgt.telemetry.protocols.bmp.parser.BmpParser;
+
+import com.codahale.metrics.MetricRegistry;
+
+public class BmpRoutesAdapterTest {
```
j-white: BMP? :)

features/telemetry/api/pom.xml
cgorantla: I currently have `netflow.proto` generated files in `flows/api` project but we can move this to `telemetry/api` ( which is the model I initially had).  I moved this to `flows/api` as I was trying to avoid conversion from `FlowMessage` to `Flow`.  I don't think it offers any performance benefits as it is just converting from one DTO to another.

j-white: We can keep it in `flows/api` then. This dependency shouldn't be needed then no?

cgorantla: yes. I still have the conversion of `FlowMessage` to `Flow` in  `Netflow5Converter`.  Once everything is moved to `FlowMessage`  it will be removed.

cgorantla: Sorry, I think we can remove this dependency on `flows/api` it is only needed in adapter.

features/flows/api/src/main/java/org/opennms/netmgt/flows/api/Converter.java
```diff
@@ -30,6 +30,8 @@
 
 import java.util.List;
 
+// TODO: Remove the converter once proto maps to Flow interface directly.
 public interface Converter<P> {
     List<Flow> convert(P packet);
+    Flow convertTo(P packet);
```
fooker: Not sure if this will work out well. In sflow it's still possible to return multiple flows from one incoming message.
`Collections.singletonList` should make it easy to implement the original interface.

features/telemetry/protocols/netflow/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/adapter/netflow5/Netflow5Converter.java
```diff
@@ -28,17 +28,223 @@
...
+            }
+
+            @Override
+            public String getDstAddr() {
+                return flowMessage.getDstAddress();
+            }
+
+            @Override
+            public Optional<String> getDstAddrHostname() {
+                return Optional.of(flowMessage.getDstHostname());
```
fooker: Should be `Optional.ofNullable`.

features/telemetry/protocols/netflow/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/adapter/netflow5/Netflow5Converter.java
```diff
@@ -28,17 +28,223 @@
...
+            }
+
+            @Override
+            public String getNextHop() {
+                return flowMessage.getNextHopAddress();
+            }
+
+            @Override
+            public Optional<String> getNextHopHostname() {
+                return Optional.of(flowMessage.getNextHopHostname());
```
fooker: Should be `Optional.ofNullable`

features/telemetry/protocols/netflow/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/adapter/netflow5/Netflow5Converter.java
```diff
@@ -28,17 +28,223 @@
...
+            }
+
+            @Override
+            public Integer getProtocol() {
+                return flowMessage.getProtocol();
+            }
+
+            @Override
+            public SamplingAlgorithm getSamplingAlgorithm() {
+                switch (flowMessage.getSamplingAlgorithm().getNumber()) {
```
fooker: `SmaplingAlgorithm` is an enum by itself. We should use the enum variants instead of the number.

features/telemetry/protocols/netflow/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/adapter/netflow5/Netflow5Converter.java
```diff
@@ -28,17 +28,223 @@
...
+            }
+
+            @Override
+            public String getSrcAddr() {
+                return flowMessage.getSrcAddress();
+            }
+
+            @Override
+            public Optional<String> getSrcAddrHostname() {
+                return Optional.of(flowMessage.getSrcHostname());
```
fooker: Should be `Optional.ofNullable`

features/telemetry/protocols/netflow/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/adapter/netflow5/Netflow5Converter.java
```diff
@@ -28,17 +28,223 @@
...
+            }
+
+            @Override
+            public Integer getTos() {
+                return flowMessage.getTos();
+            }
+
+            @Override
+            public NetflowVersion getNetflowVersion() {
+                switch (flowMessage.getNetflowVersion().getNumber()) {
```
fooker: Use the variants here, too.

features/telemetry/protocols/netflow/adapter/src/test/java/org/opennms/netmgt/telemetry/protocols/netflow/adapter/netflow5/Netflow5ConverterTest.java
```diff
@@ -78,11 +76,11 @@ public void canParseNetflow5Flows() {
         assertThat(flow.getSamplingInterval(), is(0.0));
         assertThat(flow.getSamplingAlgorithm(), is(Flow.SamplingAlgorithm.Unassigned));
         MatcherAssert.assertThat(flow.getSrcAddr(), equalTo("10.0.2.2"));
-        MatcherAssert.assertThat(flow.getSrcAddrHostname(), equalTo(Optional.empty()));
+        MatcherAssert.assertThat(flow.getSrcAddrHostname(), equalTo(Optional.of("")));
         MatcherAssert.assertThat(flow.getSrcPort(), equalTo(54435));
         MatcherAssert.assertThat(flow.getSrcMaskLen(), equalTo(0));
         MatcherAssert.assertThat(flow.getDstAddr(), equalTo("10.0.2.15"));
-        MatcherAssert.assertThat(flow.getDstAddrHostname(), equalTo(Optional.empty()));
+        MatcherAssert.assertThat(flow.getDstAddrHostname(), equalTo(Optional.of("")));
```
fooker: I'm not sure if `""` is semantically equal to `null` in this case.

features/telemetry/protocols/netflow/adapter/src/test/java/org/opennms/netmgt/telemetry/protocols/netflow/adapter/netflow5/Netflow5ConverterTest.java
```diff
@@ -95,8 +93,8 @@ public void canParseNetflow5Flows() {
         MatcherAssert.assertThat(flow.getPackets(), equalTo(5L));
         MatcherAssert.assertThat(flow.getDirection(), equalTo(Flow.Direction.INGRESS));
         MatcherAssert.assertThat(flow.getNextHop(), equalTo("0.0.0.0"));
-        MatcherAssert.assertThat(flow.getNextHopHostname(), equalTo(Optional.empty()));
-        MatcherAssert.assertThat(flow.getVlan(), nullValue());
+        MatcherAssert.assertThat(flow.getNextHopHostname(), equalTo(Optional.of("")));
+        MatcherAssert.assertThat(flow.getVlan(), equalTo(0));
```
fooker: VLAN 0 is not equal to no VLAN

features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/netflow5/proto/Header.java
```diff
@@ -111,4 +114,28 @@ public String toString() {
                 .add(new UnsignedValue("@samplingInterval", this.samplingInterval))
                 .build();
     }
+
+    void updateBuilder(FlowMessageProto.Builder builder) {
```
fooker: I would prefer to have these messages in the parser and keep the proto implementations clean of that knowledge.

features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/netflow5/proto/Packet.java
```diff
@@ -70,6 +75,18 @@ public Packet(final Header header,
         this.records = records;
     }
 
+    private FlowMessageProto.Builder createFlowMessage(Header header, Record record) {
```
fooker: I would prefer to have these messages in the parser and keep the proto implementations clean of that knowledge.

features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/netflow5/proto/Record.java
```diff
@@ -54,9 +56,9 @@
 
     public final Packet packet; // Enclosing packet
 
-    public final Inet4Address srcAddr;
-    public final Inet4Address dstAddr;
-    public final Inet4Address nextHop;
+    public Inet4Address srcAddr;
```
fooker: Any reason to remove these finals?

features/telemetry/protocols/netflow/adapter/src/test/java/org/opennms/netmgt/telemetry/protocols/netflow/adapter/netflow5/Netflow5ConverterTest.java
cgorantla: yeah, proto has primitives and never returns null  even for Strings.  need to find a way to differentiate between actual values and not having a value.

features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/netflow5/proto/Header.java
cgorantla: Makes sense. will do that.

features/telemetry/api/pom.xml
fooker: IMO, the transport protocol stuff should go to `/features/telemetry/protocols/netflow/transport` ans it is related to netflow only.

cgorantla: > IMO, the transport protocol stuff should go to `/features/telemetry/protocols/netflow/transport` ans it is related to netflow only.

You mean protobuf generated files should be moved to transport ?

fooker: Yeah. Abd they should reside under `proto/netflow/`. This allows parser and adapter to use it.

cgorantla:
@fooker @christianpape All tests seems to be passing now. Also added tests to validate `Flow` values to be consistent with previous bson implementation. Please review.

features/telemetry/protocols/netflow/transport/src/main/proto/netflow.proto
```diff
@@ -0,0 +1,74 @@
+
+
+syntax = "proto3";
+
+option java_multiple_files = true;
+option java_package = "org.opennms.netmgt.telemetry.protocols.netflow.proto";
```
fooker: The package should be `netflow.transport`, to. This was a little confusing...

features/telemetry/protocols/netflow/transport/src/main/proto/netflow.proto
```diff
@@ -0,0 +1,74 @@
...
+    PropertyMatchFiltering = 5;
+    HashBasedFiltering = 6;
+    FlowStateDependentIntermediateFlowSelectionProcess = 7;
+}
+
+enum NetflowVersion {
+    V5 = 0;
+    V9 = 1;
+    IPFIX = 2;
+    SFLOW = 3;
```
fooker: This should be removed as it's not used.

features/telemetry/protocols/netflow/transport/src/main/proto/netflow.proto
```diff
@@ -0,0 +1,74 @@
...
+
+
+enum Direction {
+    INGRESS = 0;
+    EGRESS = 1;
+}
+
+enum SamplingAlgorithm {
+    Unassigned = 0;
+    SystematicCountBasedSampling = 1;
```
fooker: This should be formatted as `CAPITALS_WITH_UNDERSCORES`.

features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/ie/RecordProvider.java
```diff
@@ -30,6 +30,7 @@
 
 import java.util.stream.Stream;
 
+
```
fooker: Do not commit stuff like this, please.

features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/InvalidPacketException.java
```diff
@@ -40,6 +40,10 @@ public InvalidPacketException(final ByteBuf buffer, final String message, final
         super(appendPosition(message, buffer), cause);
     }
 
+    public InvalidPacketException(final String message, final Throwable cause) {
```
fooker: I can't find any usage of this - is it still needed?

features/telemetry/protocols/netflow/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/adapter/common/FlowMapper.java
```diff
@@ -0,0 +1,270 @@
...
+
+import org.opennms.netmgt.flows.api.Flow;
+import org.opennms.netmgt.telemetry.protocols.netflow.common.FlowMessage;
+
+import com.google.common.base.Strings;
+
+public class FlowMapper {
+
+    public static Flow buildFlowFromProto(FlowMessage flowMessage) {
+        return new Flow() {
```
fooker: This should be moved to a standalone class.
The old `*Flow` (like `IpfixFlow`) classes can be removed.

features/telemetry/protocols/netflow/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/adapter/common/FlowMapper.java
```diff
@@ -0,0 +1,270 @@
...
+                    case V5:
+                        return NetflowVersion.V5;
+                    case V9:
+                        return NetflowVersion.V9;
+                    case IPFIX:
+                        return NetflowVersion.IPFIX;
+                    case SFLOW:
+                        return NetflowVersion.SFLOW;
+                    default:
+                        return NetflowVersion.V5;
```
fooker: This should throw an `IllegalStateException`.

features/telemetry/protocols/netflow/transport/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/proto/FlowMessageProtoOrBuilder.java
```diff
@@ -0,0 +1,800 @@
...
+// Generated by the protocol buffer compiler.  DO NOT EDIT!
+// source: netflow.proto
+
+package org.opennms.netmgt.telemetry.protocols.netflow.proto;
+
+import org.opennms.netmgt.telemetry.protocols.netflow.common.FlowMessage;
+
+public interface FlowMessageProtoOrBuilder extends
+    // @@protoc_insertion_point(interface_extends:FlowMessageProto)
+    com.google.protobuf.MessageOrBuilder, FlowMessage {
```
fooker: Is this hand edited?

features/telemetry/protocols/netflow/transport/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/common/FlowMessage.java
```diff
@@ -0,0 +1,266 @@
...
+
+import org.opennms.netmgt.telemetry.protocols.netflow.proto.Direction;
+import org.opennms.netmgt.telemetry.protocols.netflow.proto.NetflowVersion;
+import org.opennms.netmgt.telemetry.protocols.netflow.proto.SamplingAlgorithm;
+
+import com.google.protobuf.DoubleValue;
+import com.google.protobuf.UInt32Value;
+import com.google.protobuf.UInt64Value;
+
+public interface FlowMessage {
```
fooker: I don't think we need this interface at all. Instead, we should use the `FlowMessageProto` directly in the converters.

features/telemetry/protocols/netflow/adapter/src/test/java/org/opennms/netmgt/telemetry/protocols/netflow/adapter/ipfix/IpFixProtobufValidationTest.java
```diff
@@ -0,0 +1,179 @@
...
+import com.google.protobuf.InvalidProtocolBufferException;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+
+
+/**
+ * This test validates netflow protobuf values against bson output.
+ */
+public class IpFixProtobufValidationTest {
```
fooker: I can see this is useful during the transition but I would prefer to get rid of the BSON stuff completely. My proposal is to generate the BSON output one last time and use it to generate some JSON which will be used in this tests instead.

features/telemetry/protocols/netflow/adapter/src/test/java/org/opennms/netmgt/telemetry/protocols/netflow/adapter/netflow5/Netflow5ConverterTest.java
```diff
@@ -96,7 +96,7 @@ public void canParseNetflow5Flows() {
         MatcherAssert.assertThat(flow.getDirection(), equalTo(Flow.Direction.INGRESS));
         MatcherAssert.assertThat(flow.getNextHop(), equalTo("0.0.0.0"));
         MatcherAssert.assertThat(flow.getNextHopHostname(), equalTo(Optional.empty()));
-        MatcherAssert.assertThat(flow.getVlan(), nullValue());
```
fooker: What's the difference?

features/telemetry/protocols/netflow/adapter/src/test/java/org/opennms/netmgt/telemetry/protocols/netflow/adapter/netflow5/Netflow5ConverterTest.java
```diff
@@ -121,9 +121,15 @@ public void canParseNetflow5Flows() {
...
                 packet.getRecords().forEach(rec -> {
-                    final ByteBuffer bf = ParserBase.serialize(Protocol.NETFLOW9, rec);
-                    final BsonDocument doc = new RawBsonDocument(bf.array());
-                    flows.addAll(nf5Converter.convert(doc));
+
+                    byte[] message = FlowMessageBuilderFactory.buildAndSerialize(Protocol.NETFLOW5, rec);
+                    try {
+                        FlowMessage flowMessage = FlowMessageProto.parseFrom(message);
+                        flows.addAll(nf5Converter.convert(flowMessage));
+                    } catch (InvalidProtocolBufferException e) {
```
fooker: Just make the method `throws Exception`.

features/telemetry/protocols/netflow/adapter/src/test/java/org/opennms/netmgt/telemetry/protocols/netflow/adapter/netflow9/Netflow9ConverterTest.java
```diff
@@ -113,9 +116,15 @@ public void canParseNetflow9Flows() {
...
                 packet.getRecords().forEach(rec -> {
-                    final ByteBuffer bf = ParserBase.serialize(Protocol.NETFLOW9, rec);
-                    final BsonDocument doc = new RawBsonDocument(bf.array());
-                    flows.addAll(nf9Converter.convert(doc));
+
+                    byte[] message = FlowMessageBuilderFactory.buildAndSerialize(Protocol.NETFLOW9, rec);
+                    try {
+                        FlowMessage flowMessage = FlowMessageProto.parseFrom(message);
+                        flows.addAll(nf9Converter.convert(flowMessage));
+                    } catch (InvalidProtocolBufferException e) {
```
fooker: Same here

features/telemetry/protocols/netflow/adapter/src/test/java/org/opennms/netmgt/telemetry/protocols/netflow/adapter/netflow5/Netflow5ConverterTest.java
cgorantla: It's in lambda expression

cgorantla:
@fooker Think I handled all the review comments. Please take a look. Let me know if I miss anything.

features/telemetry/protocols/netflow/adapter/src/test/java/org/opennms/netmgt/telemetry/protocols/netflow/adapter/ipfix/FlowTimeoutTest.java
```diff
@@ -1,129 +0,0 @@
...
-package org.opennms.netmgt.telemetry.protocols.netflow.adapter.ipfix;
-
-import static org.hamcrest.CoreMatchers.is;
-import static org.junit.Assert.assertThat;
-
-import org.bson.BsonDocument;
-import org.bson.BsonDocumentWriter;
-import org.junit.Test;
-
-public class FlowTimeoutTest {
```
fooker: We should adopt this test. I remember this was quite tricky to get right - so it would be nice to test the behavior on the parser, too.

features/telemetry/protocols/netflow/adapter/src/test/java/org/opennms/netmgt/telemetry/protocols/netflow/adapter/ipfix/IpfixTestData.java
```diff
@@ -0,0 +1,207 @@
...
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.netmgt.telemetry.protocols.netflow.adapter.ipfix;
+
+// Generated using protobuf json utils.
+public interface IpfixTestData {
+
+    String[] ipFixData = {"{\n" +
```
fooker: Why not load this as resource?

features/telemetry/protocols/netflow/adapter/src/test/java/org/opennms/netmgt/telemetry/protocols/netflow/adapter/netflow5/Netflow5ConverterTest.java
fooker: Right. But in the end it could be a good old for loop, right?

features/telemetry/protocols/netflow/adapter/src/test/java/org/opennms/netmgt/telemetry/protocols/netflow/adapter/netflow9/FlowTimeoutTest.java
```diff
@@ -1,123 +0,0 @@
...
-package org.opennms.netmgt.telemetry.protocols.netflow.adapter.netflow9;
-
-import static org.hamcrest.CoreMatchers.is;
-import static org.junit.Assert.assertThat;
-
-import org.bson.BsonDocument;
-import org.bson.BsonDocumentWriter;
-import org.junit.Test;
-
-public class FlowTimeoutTest {
```
fooker: Same here, we should adopt this test.

features/telemetry/protocols/netflow/adapter/src/test/java/org/opennms/netmgt/telemetry/protocols/netflow/adapter/netflow9/Netflow9TestData.java
```diff
@@ -0,0 +1,311 @@
...
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.netmgt.telemetry.protocols.netflow.adapter.netflow9;
+
+// Generated using protobuf json utils.
+public interface Netflow9TestData {
+
+    String[] Netflow9Message = {
```
fooker: Make this a resource, too.

features/telemetry/protocols/netflow/adapter/src/test/java/org/opennms/netmgt/telemetry/protocols/netflow/adapter/ipfix/FlowTimeoutTest.java
cgorantla: Updated tests to handle this directly on proto builder.

features/telemetry/protocols/netflow/adapter/src/test/java/org/opennms/netmgt/telemetry/protocols/netflow/adapter/netflow5/Netflow5ConverterTest.java
cgorantla: Since `packet.getRecords()` is a stream, I'm inclined to keep this way.

cgorantla:
> One last change request: can you rename the `AbstractNetflowAdapter` to `NetflowAdapter` and make it not abstract? This would simplify a `MultiProtocolListener` config a lot. It would allow us to use a single queue for all the prayers.

Done.  Not sure if we want to make those configuration changes now in `release-26`. For now retaining individual adapters.

fooker:
@cgorantla No, just leave the config as it is. Maybe we should open another issue for this (changing the example and docs)?

PR #2906, opened 2020-02-27
NMS-12413: Fix remove from definitions

cgorantla:
Using `getAgentConfig`  comes with some defaults which results in equality check failure.
Get config directly from snmp config by matching specifics and ranges.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12413

### DIFF
--- a/opennms-config/src/main/java/org/opennms/netmgt/config/SnmpPeerFactory.java
+++ b/opennms-config/src/main/java/org/opennms/netmgt/config/SnmpPeerFactory.java
@@ -43,18 +43,21 @@ import java.net.InetAddress;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.Lock;
 
 import org.apache.commons.io.IOUtils;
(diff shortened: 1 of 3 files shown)


cgorantla:
I validated this against various combinations. Since now we are longer get the definition derived from agent config but directly compare it against what is in definitions  this should resolve issues where definitions are not matching.

I have asked @jeffgdotorg  for help in validating this change.

opennms-config/src/main/java/org/opennms/netmgt/config/SnmpPeerFactory.java
```diff
@@ -470,6 +480,84 @@ public boolean removeFromDefinition(InetAddress inetAddress, String location, St
...
+        definition.setPort(matchingDefinition.getPort());
+        definition.setTimeout(matchingDefinition.getTimeout());
+        definition.setTTL(matchingDefinition.getTTL());
+        definition.setRetry(matchingDefinition.getRetry());
+        return definition;
+    }
+
+    private boolean matchDefinition(Definition definition, InetAddress inetAddress, String location) {
+        String locationFromDefinition = definition.getLocation();
+        if(DEFAULT_LOCATION.equals(location)) {
```
j-white: Let's use LocationUtils here.

PR #2910, opened 2020-03-05
NMS-12573: Refine Parameter handling of Telemetryd

fooker:
This allows Telemetryd {Adapter,Parser,Listener} Factories to modify the
parameter set during construction before applying the parameters to the
created beans. Each Factory implementation can implement custom
parameter handling and modify/remove parameters for the according
definition. The modified parameters are then applyed using BeanWrapper.

In addition this contains some cleanup. It simplyfies all factory
implementations be removing duplicated BeanWrapper calls, refines the
BmpIntegrationAdapter initialisation to leverage the modifiable
parameters and renames a couple of things to make projects structure mor
concise.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12573


features/telemetry/listeners/src/main/java/org/opennms/netmgt/telemetry/listeners/factory/TcpListenerFactory.java
```diff
@@ -68,9 +66,6 @@ public Listener createBean(ListenerDefinition listenerDefinition) {
         if (parser.size() != listenerDefinition.getParsers().size()) {
             throw new IllegalArgumentException("Each parser must be of type TcpParser but was not.");
         }
-        final TcpListener listener = new TcpListener(listenerDefinition.getName(), parser.iterator().next(), telemetryRegistry.getMetricRegistry());
-        final BeanWrapper wrapper = PropertyAccessorFactory.forBeanPropertyAccess(listener);
```
j-white: I believe this was done within the factory itself rather that outside because of reflection issues with the OSGi proxy classes.

Does this work differently now?

fooker: I don't know if this has been the case before. But I verified that it works on standalone OpenNMS and on Minion.

The instances returned from the factories are not proxied (whereas the factories itself are). Therefore the `BeanWrapper` should work just fine.

j-white: Sounds good. Thanks for the confirmation.

PR #2911, opened 2020-03-05
NMS-12574: Apply default to OpenBMP kafka producer

fooker:
This applies default values to the Kafka producer of the OpenBMP
Integration Adapter. The defaults are the same as the ones used in the
OpenBMP reference implementation.

This is a follow-up of #2910 and this PR should be re-targeted to `features/bmp` after merging #2910.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12574

### DIFF
--- a/features/telemetry/protocols/bmp/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/adapter/openbmp/BmpKafkaProducer.java
+++ b/features/telemetry/protocols/bmp/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/adapter/openbmp/BmpKafkaProducer.java
@@ -31,6 +31,7 @@ package org.opennms.netmgt.telemetry.protocols.bmp.adapter.openbmp;
 import java.util.Map;
 
 import org.apache.kafka.clients.producer.KafkaProducer;
+import org.apache.kafka.clients.producer.ProducerConfig;
 import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.common.serialization.StringSerializer;
 import org.opennms.core.utils.StringUtils;
@@ -59,7 +60,12 @@ public class BmpKafkaProducer implements BmpMessageHandler {
     }
 
     private static KafkaProducer<String, String> buildProducer(final Map<String, Object> kafkaConfig) {
-        // TODO fooker: Apply defaults (steal from https://github.com/SNAS/openbmp/blob/1a615a3c75a0143cc87ec70458471f0af67d3929/Server/src/kafka/MsgBusImpl_kafka.cpp#L162) (see https://issues.opennms.org/browse/NMS-12574)
+        // Apply some defaults
+        kafkaConfig.putIfAbsent(ProducerConfig.BATCH_SIZE_CONFIG, 100);
+        kafkaConfig.putIfAbsent(ProducerConfig.COMPRESSION_TYPE_CONFIG, "snappy");
+        kafkaConfig.putIfAbsent(ProducerConfig.MAX_REQUEST_SIZE_CONFIG, 1000000);
+        kafkaConfig.putIfAbsent(ProducerConfig.RETRIES_CONFIG, 2);
+        kafkaConfig.putIfAbsent(ProducerConfig.RETRY_BACKOFF_MS_CONFIG, 100);
 
         return new KafkaProducer<>(kafkaConfig, new StringSerializer(), new StringSerializer());
     }


features/telemetry/protocols/bmp/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/adapter/openbmp/BmpKafkaProducer.java
```diff
@@ -59,7 +59,18 @@ public BmpKafkaProducer(final String topicPrefix,
     }
 
     private static KafkaProducer<String, String> buildProducer(final Map<String, Object> kafkaConfig) {
-        // TODO fooker: Apply defaults (steal from https://github.com/SNAS/openbmp/blob/1a615a3c75a0143cc87ec70458471f0af67d3929/Server/src/kafka/MsgBusImpl_kafka.cpp#L162) (see https://issues.opennms.org/browse/NMS-12574)
+        // Apply some defaults
+        kafkaConfig.putIfAbsent("batch.num.messages", 100);
```
j-white: The Kafka client library has constants for these strings, would nice to use those.

features/telemetry/protocols/bmp/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/adapter/openbmp/BmpKafkaProducer.java
```diff
@@ -59,7 +59,18 @@ public BmpKafkaProducer(final String topicPrefix,
...
+        kafkaConfig.putIfAbsent("compression.codec", "snappy");
+        kafkaConfig.putIfAbsent("message.max.bytes", 1000000);
+        kafkaConfig.putIfAbsent("receive.message.max.bytes", 100000000);
```
cgorantla: Can't find this property with quick search on producer config.  If you can directly get them from kafka client as Jesse suggested, that would rule out any outdated configuration.

fooker:
I've reduced the settings to the required minimum.

@cgorantla Can you have a look if these make sens?

features/telemetry/protocols/bmp/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/adapter/openbmp/BmpKafkaProducer.java
```diff
@@ -59,7 +60,12 @@ public BmpKafkaProducer(final String topicPrefix,
     }
 
     private static KafkaProducer<String, String> buildProducer(final Map<String, Object> kafkaConfig) {
-        // TODO fooker: Apply defaults (steal from https://github.com/SNAS/openbmp/blob/1a615a3c75a0143cc87ec70458471f0af67d3929/Server/src/kafka/MsgBusImpl_kafka.cpp#L162) (see https://issues.opennms.org/browse/NMS-12574)
+        // Apply some defaults
+        kafkaConfig.putIfAbsent(ProducerConfig.BATCH_SIZE_CONFIG, 100);
+        kafkaConfig.putIfAbsent(ProducerConfig.COMPRESSION_TYPE_CONFIG, "snappy");
```
cgorantla: I think these two parameters are sufficient for increasing the default message size. By default it is 1MB, you may want to increase it to 10MB.
 `kafkaConfig.putIfAbsent(ProducerConfig.COMPRESSION_TYPE_CONFIG, "snappy");`
 `kafkaConfig.putIfAbsent(ProducerConfig.MAX_REQUEST_SIZE_CONFIG, 1000000);`

Also we should allow the config to be added from osgi configuration just like [here](https://github.com/OpenNMS/opennms/blob/3de338a7044ed5d45df5b70f26fda5798530e8e5/core/ipc/common/kafka/src/main/java/org/opennms/core/ipc/common/kafka/OsgiKafkaConfigProvider.java#L59)

The loaded config should override anything that we set here.  So we can totally skip these defaults and allow user to add them.

fooker: We don't use our default kafka config mechanism here, because we talk to the external OpenBMP kafka cluster. Therefore the config is only set in the according adapter. And this config is already overriding the defaults, right?

cgorantla: okay, looks like it is set in adapter config. That works.

PR #2912, opened 2020-03-05
NMS-12578: Confd templates for Minion configuration

mattixtech:
This PR adds the ability for some of Minion's configuration to be applied by confd via a yaml file mounted on container startup. Similar to how the OpenNMS container was updated to use confd in this PR https://github.com/opennms-forge/docker-horizon-core-web/pull/58.

This PR isn't meant to cover all of Minion's configuration, just a part 1 covering the low hanging fruit.

### All Contributors

### Pull Request Process

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12578


opennms-container/minion/confd/templates/system.properties.tmpl
```diff
@@ -0,0 +1,161 @@
+################################################################################
```
j-white: Can we update the default/existing system.properties files to include a statement like `${optionals} = confd.system.properties`  and append the properties there rather than templating this file?

That would save us from having to maintain this template for Karaf upgrades going forward.

mattixtech: Yes, that should work. Do we have to worry about duplicates in that case? Or would we treat the values that are currently in the file as non-configurable by confd.

j-white: Karaf documents this as "Included files will override the values specified in this file", so shouldn't have to worry about dupes.

agalue:
At a glance, it looks very well.

I can see there are settings to configure the off-heap feature, but nothing about that in the documentation. Also, the documentation mentions the single-port listener, but nothing about dedicated ports, although there are config settings for this purpose. Is this intended?

My time is minimal at the moment, but I'll try to find some time to give it a try.

opennms-container/minion/Dockerfile
```diff
@@ -68,6 +68,15 @@ RUN if [[ "$(ls -1 /tmp/rpms/*.rpm 2>/dev/null | wc -l)" != 0 ]]; then dnf -y in
 
 COPY ./assets/* /
 
+# Install confd
+ARG CONFD_VERSION="0.16.0"
+ARG CONFD_ARCH="amd64"
```
indigo423: 👍

opennms-container/minion/assets/entrypoint.sh
```diff
@@ -149,12 +152,42 @@ applyOverlayConfig() {
...
 
+runConfd() {
+  # Create any directories that confd might write to
+  while IFS= read -r dir; do
+    local dirToCreate="$MINION_HOME"/"$dir"
+    echo "Creating $dirToCreate so confd can write to it"
+    mkdir -p "$dirToCreate"
+  done < "$CONFD_CONFIG_DIR"/directories
+
+  "$CONFD_BIN" -log-level debug -onetime -backend=file -file "$CONFD_KEY_STORE" -confdir "$CONFD_CONFIG_DIR"
```
indigo423: Just to be sure, we nail it down to use only the minion-config.yaml file and don't allow other configuration backends right? Just as a quick idea, we could gain some flexibility using a `confd.toml` and call the binary with `-config-file` 

```
confdir = "/etc/confd"
backend = "file"
file = "/opt/minion/minion-config.yaml"
```
I don't worry much - no blocker for and really just a minor comment.

features/container/minion/src/main/filtered-resources/etc/system.properties
```diff
@@ -148,3 +148,5 @@ karaf.secured.services = (&(osgi.command.scope=*)(osgi.command.function=*))
 
 # OPENNMS: Enable SNMP4J Log4j logging
 snmp4j.LogFactory=org.snmp4j.log.Log4jLogFactory
+
+${optionals} = confd.system.properties
```
j-white: Given this is in the default `system.properties` file and not just in the container, maybe we should actually rename it.

Karaf has this stanza in their default configuration, maybe we just reuse it? https://github.com/apache/karaf/blob/master/assemblies/features/base/src/main/resources/resources/etc/system.properties#L25

Note that it's also at the top of the time, and not the bottom, not sure if that makes a difference.

opennms-container/minion/confd/templates/confd.system.properties.tmpl
```diff
@@ -0,0 +1,10 @@
+{{$snmp2In1Key := "/snmp/snmp4j/org.opennms.snmp.snmp4j.allowSNMPv2InV1" -}}
```
j-white: Should be be strongly modeling all of the system properties, or try and keep this generic from a template perspective?

mattixtech: Generic will let us be a lot more flexible, will update.

features/container/minion/src/main/filtered-resources/etc/system.properties
mattixtech: Will follow the pattern from the default config.

opennms-container/minion/assets/entrypoint.sh
mattixtech: I didn't mean to lock it down to a file. Agreed using a confd.toml will allow us to be more flexible since this file could be optionally mounted to changed the config. Will use it.

mattixtech:
> At a glance, it looks very well.
> 
> I can see there are settings to configure the off-heap feature, but nothing about that in the documentation. Also, the documentation mentions the single-port listener, but nothing about dedicated ports, although there are config settings for this purpose. Is this intended?
> 
> My time is minimal� at the moment, but I'll try to find some time to give it a try.

In this case, currently the dedicated ports config is the default and gets applied if single port is no specified. Not sure if this is correct.

@j-white Do we want to always apply either the single port config or dedicated port config, or do we also want the option of neither.

j-white:
> @j-white Do we want to always apply either the single port config or dedicated port config, or do we also want the option of neither.

For flows, I would always apply the single port configuration if enabled. We can add additional keys and flexibility for customizing telemtryd later.

Going back over this, I think it would be worthwhile to update the Minion container in the smoke test to use the templates when possible, rather than building the overlays itself. For example, the following logic is embedded in the smoke tests for the broker configuration: https://github.com/OpenNMS/opennms/blob/develop/smoke-test/src/main/java/org/opennms/smoketest/containers/MinionContainer.java#L187

And the following overlay is used for the single port configuration:
https://github.com/OpenNMS/opennms/blob/develop/smoke-test/src/main/resources/minion-overlay/org.opennms.features.telemetry.listeners-udp-50000-single-port-flows.cfg

Updating the smoke tests to use confd makes sure we have test coverage for the template and removes the need to maintain the logic twice.

mattixtech:
I've updated the PR to change the smoke tests to use the yaml file/confd for configuration of the minion container rather than overlaying individual files.

I've updated the telemetry flows configuration to now have a default single port configuration that can be enabled/disabled (and have parameters added). Additionally an arbitrary list of individual listeners along with their parsers/parameters can be specified for individual ports.

Full smoke test is passing.

PR #2914, opened 2020-03-11
NMS-12583: Write enriched flows to Kafka

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12583


cgorantla:
> I have some uneasy feelings about moving the `FlowDocument` to the `api` module. It contains a lot of elasticsearch specific bits.
> 
> I can see two ways to make this more clear:
> 
> 1. Make the whole thing elasticsearch specific - this would move the `Persister` API and the kafka implementation (which I would call a `Forwarder`) into the elasticsearch module and will be used only there.
> 2. Move the `Enricher` out of the `elastic` module and have an intermediate format which will sit between `Enricher`, `Repository` implementation and the `Persister`. With this, the `Repository`, can be just another `Persister`.
> 
> In addition, should there be a flag allowing to disable the forwarding completely?

I considered (2) but there are too many changes for small feature. IMO `FlowDocument` and other related classes are just model objects with json serialization annotations which we use in `KafkaFlowPersister` as well.  This seems to be least disruptive change as I understand.

I will add a flag to disable forwarding.

cgorantla:
Chatted with @j-white about the restructuring and we decided to keep `FlowDocument` related classes in  `elastic` module itself and create new model objects for the persister module.

features/flows/api/pom.xml
```diff
@@ -36,5 +36,11 @@
           <groupId>org.opennms.core.ipc.rpc</groupId>
           <artifactId>org.opennms.core.ipc.rpc.utils</artifactId>
       </dependency>
+    <dependency>
+      <groupId>com.google.code.gson</groupId>
+      <artifactId>gson</artifactId>
+      <version>${jestGsonVersion}</version>
+      <scope>provided</scope>
```
fooker: Still needed?

features/flows/api/src/main/java/org/opennms/netmgt/flows/api/EnrichedFlow.java
```diff
@@ -0,0 +1,506 @@
...
+    public enum Locality {
+        PUBLIC,
+        PRIVATE
+    }
+
+    private Flow flow;
+
+    private String application;
+
+    private String convoKey;
```
fooker: This is pure elasticsearch specific and can be calculated from other values.

features/flows/api/src/main/java/org/opennms/netmgt/flows/api/EnrichedFlow.java
```diff
@@ -0,0 +1,506 @@
...
+
+    private String host;
+
+    private List<String> hosts = new ArrayList<>();
```
fooker: Same here: pure elasticsearch specific

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/ElasticFlowRepository.java
```diff
@@ -164,6 +165,10 @@
 
     private final IndexSettings indexSettings;
```
fooker: If the property to enable it is called "forwarding", should the class be called "Forwarder", too?

cgorantla: I debated this in my mind but somehow I didn't like Forwarder for the class name.  For ElasticFlowRepository it is forwarding to kafka but when it comes to KafkaFlowPersister it is just persisting.

pbrane: > Class names should be nouns or noun phrases. This means that the name of the class should be something that would be the subject of a verb. In the case of object-oriented design, methods would be the verbs that take place on the thing that the class is a representation of. Some words should be avoided.

cgorantla: Modified to forwarder.

cgorantla:
> I think we are on the right track.
> Docs are still missing, right?

Updated docs although currently this is more of internal consumption.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/FlowDocument.java
```diff
@@ -724,4 +741,59 @@ public static FlowDocument from(final Flow flow) {
 
         return doc;
     }
+
+    public static EnrichedFlow buildEnrichedFlow(FlowDocument flowDocument) {
```
j-white: I'd move this logic to the EnrichedFlow class instead of defining it here.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/ElasticFlowRepository.java
```diff
@@ -238,8 +244,14 @@ public void persist(final Collection<Flow> flows, final FlowSource source) throw
         } catch (Exception e) {
             throw new FlowException("Failed to enrich one or more flows.", e);
         }
-
+        if(enableFlowForwarding) {
+            // Persist to kafka.
+            LOG.debug("Forwarding {} flow documents to Kafka", flowDocuments.size());
```
j-white: We don't *know* that it's Kafka, but rather just some flow forwarder.

features/flows/kafka-persistence/src/main/java/org/opennms/netmgt/flows/persistence/KafkaFlowForwarder.java
```diff
@@ -0,0 +1,139 @@
...
+                .getConfiguration(KAFKA_CLIENT_PID).getProperties();
+        if (properties != null) {
+            final Enumeration<String> keys = properties.keys();
+            while (keys.hasMoreElements()) {
+                final String key = keys.nextElement();
+                producerConfig.put(key, properties.get(key));
+            }
+        }
+
+        if (producerConfig.getProperty(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG) != null) {
```
j-white: Can we just have it fail if not set?

features/flows/kafka-persistence/src/main/java/org/opennms/netmgt/flows/persistence/KafkaFlowForwarder.java
```diff
@@ -0,0 +1,139 @@
...
+            // Overwrite the serializers
+            producerConfig.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, StringSerializer.class.getCanonicalName());
+            producerConfig.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, ByteArraySerializer.class.getCanonicalName());
+            // Class-loader hack for accessing the kafka classes when initializing producer.
+            producer = runWithGivenClassLoader(() -> new KafkaProducer<>(producerConfig), KafkaProducer.class.getClassLoader());
+            LOG.info(" Kafka Producer initialized with config {} ", producerConfig);
```
j-white:
```suggestion
            LOG.info("Kafka Producer initialized with config: {}", producerConfig);
```

Remove leading and trailing spaces from log message.

features/flows/kafka-persistence/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -0,0 +1,26 @@
...
+">
+  <!-- Enables update of bundle when kafka config is updated -->
+  <cm:property-placeholder id="kafkaForwarderProperties" persistent-id="org.opennms.features.flows.persistence.kafka" update-strategy="reload"/>
+
+  <reference id="configurationAdmin" interface="org.osgi.service.cm.ConfigurationAdmin"/>
+
+  <bean id="kafkaFlowForwarder" class="org.opennms.netmgt.flows.persistence.KafkaFlowForwarder"
+        init-method="init" destroy-method="destroy">
+    <argument ref="configurationAdmin"/>
+    <property name="topicName" value="flowDocuments"/>
```
j-white: Topic name must be configurable.

opennms-doc/guide-admin/src/asciidoc/text/flows/setup.adoc
```diff
@@ -187,3 +187,31 @@ The following cache properties are available to be set in `${OPENNMS_HOME/etc/or
 | `true`
 
 |===
+
+==== Configure Kafka forwarder
```
j-white: 👍

features/flows/api/src/main/java/org/opennms/netmgt/flows/api/EnrichedFlow.java
j-white: It would actually be nice to have this populated the same way that is done is Elasticsearch and avoid having to replicate that logic elsewhere.

j-white:
Thanks @cgorantla!

features/flows/kafka-persistence/src/main/java/org/opennms/netmgt/flows/persistence/KafkaFlowForwarder.java
cgorantla: Without  `bootstrap.servers` in the config,  `init` would throw exception and bundle wouldn't load. Since this bundle is not needed by default and should be available when needed, chose this approach.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/FlowDocument.java
cgorantla: Have added these back.

cgorantla:
@j-white Thanks for the review. Handled review comments on this. Please take one more look

features/flows/api/src/main/java/org/opennms/netmgt/flows/api/EnrichedFlow.java
j-white: This one we can skip.

PR #2915, opened 2020-03-11
NMS-12580: Improve NodeInfoCache handling

cgorantla:
Query the `InterfaceNodeCache` first and if there is a relavant `nodeId` query `NodeInfoCache` for the node document.
Also modify `nodeCache` settings to cache `10000` elements and set the cache to not to expire.
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12580

### DIFF
--- a/features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/DocumentEnricher.java
+++ b/features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/DocumentEnricher.java
@@ -40,6 +40,7 @@ import java.util.stream.Collectors;
 import org.opennms.core.cache.Cache;
 import org.opennms.core.cache.CacheBuilder;
 import org.opennms.core.cache.CacheConfig;
+import org.opennms.core.rpc.utils.mate.ContextKey;
 import org.opennms.core.utils.InetAddressUtils;
 import org.opennms.netmgt.dao.api.InterfaceToNodeCache;
 import org.opennms.netmgt.dao.api.NodeDao;
@@ -56,6 +57,7 @@ import org.slf4j.LoggerFactory;
 
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
+import com.google.common.base.Strings;
 import com.google.common.cache.CacheLoader;
 
 public class DocumentEnricher {
@@ -69,8 +71,11 @@ public class DocumentEnricher {
 
     private final ClassificationEngine classificationEngine;
 
-    // Caches NodeDocument data
+    // Caches NodeDocument data for a given node Id.
     private final Cache<Integer, Optional<NodeDocument>> nodeInfoCache;
+
+    // Caches NodeDocument data for a given node metadata.
+    private final Cache<NodeMetadataKey, Optional<NodeDocument>> nodeMetadataCache;
 
     private final Timer nodeLoadTimer;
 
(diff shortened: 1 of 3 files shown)


features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/DocumentEnricher.java
```diff
@@ -150,40 +148,24 @@ private static boolean isPrivateAddress(String ipAddress) {
...
 
-    private Optional<NodeDocument> getNodeInfo(final String location, final String ipAddress, final ContextKey contextKey, final String value) {
-        return getNodeInfo(location, InetAddressUtils.addr(ipAddress), contextKey, value);
-    }
 
-    private Optional<NodeDocument> getNodeInfo(final String location, final InetAddress ipAddress, final ContextKey contextKey, final String value) {
+    private Optional<NodeDocument> getNodeInfo(Integer nodeId) {
         OnmsNode onmsNode = null;
-
-        if (contextKey != null && !Strings.isNullOrEmpty(value)) {
```
j-white: We need to keep this logic. We can perform the check before the `interfaceToNodeCache` lookup, and store the results in a separate cache (keyed by `contextKey`).

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/DocumentEnricher.java
```diff
@@ -151,68 +163,40 @@ private static boolean isPrivateAddress(String ipAddress) {
     }
 
     private Optional<NodeDocument> getNodeInfoFromCache(final String location, final String ipAddress, final ContextKey contextKey, final String value) {
-        final NodeInfoKey key = new NodeInfoKey(location, ipAddress, contextKey, value);
+
+        Optional<NodeDocument> nodeDocument;
+        final NodeMetadataKey metadataKey = new NodeMetadataKey(contextKey, value);
```
j-white: Let's only perform this if `contextKey != null && !Strings.isNullOrEmpty(value)` and remove that check from the cache loader (i.e. getNodeInfoFromMetadataContext)

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/DocumentEnricher.java
```diff
@@ -221,19 +205,58 @@ private NodeInfoKey(final String location, final String ipAddress, final Context
...
         public int hashCode() {
-            return Objects.hash(location, ipAddress, contextKey, value);
+            return Objects.hash(contextKey, value);
         }
     }
 
+    private Optional<NodeDocument> getNodeInfoFromMetadataContext(ContextKey contextKey, String value) {
+        OnmsNode onmsNode = null;
+        if (contextKey != null && !Strings.isNullOrEmpty(value)) {
+            final List<OnmsNode> nodes = nodeDao.findNodeWithMetaData(contextKey.getContext(), contextKey.getKey(), value);
```
j-white: Would be nice to time this with the `nodeLoadTimer` too.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/DocumentEnricher.java
```diff
@@ -221,19 +205,58 @@ private NodeInfoKey(final String location, final String ipAddress, final Context
...
+
+            if (!nodes.isEmpty()) {
+                onmsNode = nodes.get(0);
```
j-white:
```suggestion
                return mapOnmsNodeToNodeDocument(nodes.get(0));
```

then can remove `onmsNode` local variable and other `if` statement bellow.

PR #2916, opened 2020-03-12
NMS-12569: Resolve Hostnames in BMP parser

fooker:
This resolves IPs to hostnames while parsing BMP messages.
IPs are resolved for Initiation messages and for each message containing
a peer header.

The DNS resolution is done asynchonous before dispatching the message to
the sink module. The behaviour is disabled by default and can be enabled
using parsing parameters.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12569


features/telemetry/protocols/bmp/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/parser/BmpParser.java
```diff
@@ -130,11 +133,16 @@
 
     private HashSet<InetAddress> connections = Sets.newHashSet();
 
+    private final DnsResolver dnsResolver;
+    private boolean dnsLookupsEnabled;
```
j-white: Can we enable this by default?

features/telemetry/protocols/bmp/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/parser/BmpParser.java
```diff
@@ -187,25 +195,39 @@ public Handler accept(final InetSocketAddress remoteAddress,
...
+                                       .thenCompose(BmpParser.this.resolveSysName(packet, remoteAddress.getAddress()));
+                }
+
+                // Dispatch the final message
+                final CompletableFuture<AsyncDispatcher.DispatchStatus> dispatched = enriched.thenCompose(msg -> {
+                    final ByteBuffer payload = ByteBuffer.wrap(msg.build().toByteArray());
+                    return BmpParser.this.dispatcher.send(new TelemetryMessage(remoteAddress, payload));
+                });
+
                 BmpParser.this.recordsDispatched.mark();
```
j-white: We haven't necessarily dispatched it yet. We should move this into the `thenCompose` block above.

features/telemetry/protocols/bmp/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/parser/BmpParser.java
```diff
@@ -187,25 +195,39 @@ public Handler accept(final InetSocketAddress remoteAddress,
...
+                    message.setBgpId(BmpParser.address(bgpId));
                 }
 
-                packet.accept(new Serializer(message));
 
-                final CompletableFuture<AsyncDispatcher.DispatchStatus> dispatched = dispatcher.send(new TelemetryMessage(remoteAddress, ByteBuffer.wrap(message.build().toByteArray())));
+                // Enrich the message with resolved hostnames
+                CompletableFuture<Transport.Message.Builder> enriched = CompletableFuture.completedFuture(message);
+                if (BmpParser.this.dnsLookupsEnabled) {
+                    enriched = enriched.thenCompose(BmpParser.this.resolvePeer(packet))
```
j-white: We need to find a way to put backpressure on the listener if we're not able to keep up. If there are already K enrichments in flight, then we should block until one of these finishes, otherwise we can OOM.

We handle this in `org.opennms.netmgt.telemetry.protocols.netflow.parser.ParserBase`, rather non-intutively. Maybe we can find a better way here? We already have resilience4j in the class-path - we could try and make use of the bulkhead: https://resilience4j.readme.io/docs/bulkhead

PR #2920, opened 2020-03-17
NMS-12571: Parse BGP capabilities

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-12571


features/telemetry/protocols/bmp/transport/src/main/resources/transport.proto
```diff
@@ -68,6 +68,7 @@ message PeerUpPacket {
         fixed32 as = 2;
         uint32 hold_time = 3;
         IpAddress id = 4;
+        string capabilities = 5;
```
j-white: Can we carry the capability structures over and only convert to strings in the BMP forwarder?

This will give us more flexibility to do things with them later.

christianpape: Ok. I'll change it.

christianpape: done

features/telemetry/protocols/bmp/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/parser/proto/bgp/packets/Capability.java
```diff
@@ -0,0 +1,61 @@
...
+    public int getCode() {
+        return code;
+    }
+
+    public int getLength() {
+        return length;
+    }
+
+    public ByteString getValue() {
+        return ByteString.copyFrom(value.array());
```
fooker: Not that it currently is required to implement right now, but do we have a clue about what this value could be?

features/telemetry/protocols/bmp/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/parser/proto/bgp/packets/OpenPacket.java
```diff
@@ -66,8 +92,10 @@ public OpenPacket(final Header header, final ByteBuf buffer, final PeerFlags fla
 
         final int parametersLength = uint8(buffer);
 
-        // Skip the parameters (see https://tools.ietf.org/html/rfc4271#section-4.2)
-        skip(buffer, parametersLength);
+        final List<Parameter> list = repeatRemaining(slice(buffer, parametersLength), Parameter::new);
+
+        parameters = list.stream().filter(p -> p.type != 2).collect(Collectors.toList());
```
fooker: I don't think we should filter the parameters, as capabilities are parameters, too. An we should keep the link to the rfc for further reference to there parameter types, if there are any.

pbrane:
> I lick it

LOL

features/telemetry/protocols/bmp/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/parser/proto/bgp/packets/Capability.java
christianpape: We currently implement the parsing for this value for code 1 and 69, like the OpenBMP implementation does.

features/telemetry/protocols/bmp/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/parser/proto/bgp/packets/OpenPacket.java
christianpape: Ok

PR #2921, opened 2020-03-17
NMS-12586: Use DroolsStreamUtils methods to serialize/deserialize facts

UberPinguin:
Copying facts between sessions leads to incompatible class signatures.
Take advantage of the Drools streaming utilities and protobuf to persist
facts between sessions, instead.

### DIFF
--- a/opennms-correlation/drools-correlation-engine/src/main/java/org/opennms/netmgt/correlation/drools/DroolsCorrelationEngine.java
+++ b/opennms-correlation/drools-correlation-engine/src/main/java/org/opennms/netmgt/correlation/drools/DroolsCorrelationEngine.java
@@ -28,6 +28,9 @@
 
 package org.opennms.netmgt.correlation.drools;
 
+import static org.drools.core.util.DroolsStreamUtils.streamIn;
+import static org.drools.core.util.DroolsStreamUtils.streamOut;
+
 import java.io.File;
 import java.io.FileInputStream;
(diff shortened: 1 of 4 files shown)


UberPinguin:
Do not merge yet - this fails with a ClassNotFound exception, but I have not been able to figure out why.

opennms-correlation/drools-correlation-engine/src/main/java/org/opennms/netmgt/correlation/drools/DroolsCorrelationEngine.java
```diff
@@ -270,7 +273,13 @@ public void initialize() throws Exception {
         }
 
         if (factObjects != null) {
-            factObjects.forEach(fact -> m_kieSession.insert(fact));
+            factObjects.forEach(fact -> {
+                try {
+                    m_kieSession.insert(streamIn(fact, m_kieSession.getClass().getClassLoader()));
```
UberPinguin: DroolsStreamUtils#streamIn throws a ClassNotFoundException for facts that are declared in and compiled from DRL.
I have not yet found a ClassLoader that can find those classes.

cgorantla:
@UberPinguin Thanks for all the groundwork. I think I got this working in #2935. Let me know if that works.

cgorantla:
@UberPinguin Took commit from here into #2935.  Closing this.

PR #2932, opened 2020-03-23
NMS-12602: Upgrade Kafka SinkMessage to Proto3

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12602


core/ipc/sink/common/src/main/proto/sink-message.proto
```diff
@@ -0,0 +1,16 @@
+syntax = "proto3";
+option java_multiple_files = true;
+option java_package = "org.opennms.core.ipc.sink.model";
+option java_outer_classname = "SinkMessageProtos";
+
+
+message SinkMessage {
+string message_id = 1;
+bytes  content = 2;
+string system_id = 3;
```
j-white: Are we using `system_id` and `location`?

cgorantla: No. I was using the same SinkMessage as gRPC but we don't need them as of now. Will remove them

PR #2933, opened 2020-03-23
NMS-12513 Open Redirect security issue

patrick-schweizer:
This Pull Request removes open redirect vulnerabilities by restricting the redirect to a whitelist.
To be tested: I hope I found all links for the whitelist but I might have missed some.

### External References

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-12612

### DIFF
--- a/opennms-webapp/src/main/java/org/opennms/web/controller/alarm/AcknowledgeAlarmByFilterController.java
+++ b/opennms-webapp/src/main/java/org/opennms/web/controller/alarm/AcknowledgeAlarmByFilterController.java
@@ -40,6 +40,7 @@ import org.opennms.netmgt.model.OnmsCriteria;
 import org.opennms.web.alarm.AcknowledgeType;
 import org.opennms.web.alarm.AlarmUtil;
 import org.opennms.web.alarm.filter.AlarmCriteria;
+import org.opennms.web.controller.RedirectRestricter;
 import org.opennms.web.filter.Filter;
 import org.opennms.web.servlet.MissingParameterException;
 import org.springframework.beans.factory.InitializingBean;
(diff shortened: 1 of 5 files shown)


j-white:
Given there's some risk in breaking some existing workflows with this, I would re-target this to the `release-26.0.0` branch. We can then back-port for `foundation-2019` at a later time.

opennms-webapp/src/main/java/org/opennms/web/controller/alarm/AcknowledgeAlarmByFilterController.java
```diff
@@ -61,7 +62,10 @@
     private AlarmRepository m_webAlarmRepository;
     
     private String m_redirectView;
-    
+
+    private RedirectRestricter redirectRestricter = RedirectRestricter.builder()
```
j-white: No allowed redirects? Is this intentional?

patrick-schweizer: yes, I didn't find any link with a redirect pointing to this controller (although I might have missed one). The alternative would be to remove the redirect parameter completely from this controller (would result in the same thing but viewer code). I wonder why we have the redirect parameter if we don't use it:
Either this is just a copy and paste leftover or
I didn't find the link...

patrick-schweizer: TODO: Patrick add as comment in code

patrick-schweizer: done.

PR #2937, opened 2020-03-25
NMS-12616: Change key for events forwarded to kafka

cgorantla:
To balance events across partitions, change key from `uei` to `id`
Fix event createTime in ProtobufMapper

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12616


j-white:
Looks good. Should we re-target this for `foundation-2019`?

j-white:
We should also update documentation, specifically this line: 
`The records are keyed by event UEI and contain a GPB encoded model of the event.`

cgorantla:
> We should also update documentation, specifically this line:
> `The records are keyed by event UEI and contain a GPB encoded model of the event.`

Done!

PR #2954, opened 2020-04-03
update CircleCI checksums to conserve space/bandwidth

RangerRick:
These changes make builds take anywhere from a few seconds to a few minutes less time, by reducing the amount of workspace junk that gets passed around from different caches.

In general, the major thing it does is add a level of caching based on the `<version>` tag in `pom.xml`, since dependency churn among single versions is less common.  This greatly reduces the size of the cache that's passed around between workspaces (in some cases up to 1GB!).

I also got rid of the "generic" keys for nodejs and maven dependencies since they increase the chances of passing old junk along, at the expense of the first build of a pom version being a bit slower (since it needs to download new dependencies).  This will have the side-effect of catching if upstream dependencies go away, which we're probably pretty immune to now and could make it hard to know that upstream repos are permanently gone or changed.

Finally, I got rid of checksumming-the-checksums in favor of just catting everything together in a repeatable way.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,11 +42,12 @@ commands:
         - run:
             name: Calculate cache key from pom files
             command: |
-              find . -type f -name "pom.xml" -exec sha256sum "{}" \; | sort -nr >> maven-dependency-cache.key
+              grep -E '<version>' pom.xml | head -n 1 | sed -e 's,^[^>]*>,,' -e 's,<.*$,,' > maven-dependency-version-cache.key
+              find . -type f -name "pom.xml" | grep -v /target/ | sort -u | xargs cat > maven-dependency-pom-cache.key
         - restore_cache:
             keys:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -55,26 +61,24 @@ commands:
...
             - ~/.m2
   restore-nodejs-cache:
       description: "NodeJS: Calculate cache key and restore cache"
       steps:
         - run:
             name: Calculate cache key
-            command: |
-              sha256sum opennms-webapp/bower.json > nodejs-dependency-cache.key
-              sha256sum opennms-webapp/package.json >> nodejs-dependency-cache.key
+            command: find opennms-webapp -name package\*.json -o -name bower.json | grep -v /target/ | sort -u | xargs cat > nodejs-dependency-json-cache.key
```
j-white, line 72: This will need to be adapted for foundation-2019+

PR #2956, opened 2020-04-04
NMS-12625: Cannot create monitored-service with JSON via ReST

patrick-schweizer:
This Pull Request allows the creation of a monitored-service via JSON in api v1 and v2

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-12625

### DIFF
--- a/opennms-model/src/main/java/org/opennms/netmgt/model/OnmsMonitoredService.java
+++ b/opennms-model/src/main/java/org/opennms/netmgt/model/OnmsMonitoredService.java
@@ -569,7 +569,6 @@ public class OnmsMonitoredService extends OnmsEntity implements Serializable, Co
     )
     @XmlElementWrapper(name="applications")
     @XmlElement(name="application")
-    // @JsonManagedReference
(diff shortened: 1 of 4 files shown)


opennms-model/src/main/java/org/opennms/netmgt/model/OnmsMonitoredService.java
```diff
@@ -570,7 +569,7 @@ public void setCurrentOutages(Set<OnmsOutage> currentOutages) {
     )
     @XmlElementWrapper(name="applications")
     @XmlElement(name="application")
-    @JsonManagedReference
+    // @JsonManagedReference
```
j-white, line 572: Remove if no longer being used.

patrick-schweizer: :+1:

patrick-schweizer:
> Can we add test coverage the scenarios this addresses?

yes :-) done.

PR #2958, opened 2020-04-05
NMS-12637: Fix NPE on graph service

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12637

### DIFF
--- a/features/graph/service/src/main/java/org/opennms/netmgt/graph/service/DefaultGraphService.java
+++ b/features/graph/service/src/main/java/org/opennms/netmgt/graph/service/DefaultGraphService.java
@@ -133,10 +133,12 @@ public class DefaultGraphService implements GraphService {
     }
 
     public void onUnbind(GraphContainerProvider graphContainerProvider, Map<String, String> props) {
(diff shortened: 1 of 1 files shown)


features/graph/service/src/main/java/org/opennms/netmgt/graph/service/DefaultGraphService.java
```diff
@@ -133,10 +133,12 @@ public void onBind(GraphContainerProvider graphContainerProvider, Map<String, St
     }
 
     public void onUnbind(GraphContainerProvider graphContainerProvider, Map<String, String> props) {
-        graphContainerProviders.remove(graphContainerProvider);
-        final ServiceRegistration<GraphContainerProviderRegistration> serviceRegistration = serviceRegistrationMap.remove(graphContainerProvider);
-        if (serviceRegistration != null) {
-            serviceRegistration.unregister();
+        if(graphContainerProvider != null) {
```
j-white, line 136: Could be simplified by returning immediately if null - would save having to indent the whole block.

PR #2970, opened 2020-04-10
 NMS-12605: Provisiond accepts multiple primary SNMP interfaces

patrick-schweizer:
There should be only one primary interface allowed per node. But provisiond accepts node definitions with two primary tagged snmp interfaces.
This Pull Request addresses that problem and disallows the import of nodes with multiple primary interfaces (an error is thrown during validation phase).

This fix works also for the import via the Rest API: multiple primaries will lead to a HTTP 400 with a descriptive error message.

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-12605

### DIFF
--- a/core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/scripts/controllers/Interface.js
+++ b/core/web-assets/src/main/assets/js/apps/onms-requisitions/lib/scripts/controllers/Interface.js
@@ -28,7 +28,7 @@ const RequisitionService = require('../model/RequisitionService');
   *
   * @description The controller for manage the modal dialog for add/edit IP interfaces of requisitioned nodes
   */
-  .controller('InterfaceController', ['$scope', '$uibModalInstance', 'RequisitionsService', 'foreignSource', 'foreignId', 'requisitionInterface', 'ipBlackList', 'isPrimaryExists', function($scope, $uibModalInstance, RequisitionsService, foreignSource, foreignId, requisitionInterface, ipBlackList, isPrimaryExists) {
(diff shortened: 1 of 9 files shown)


patrick-schweizer:
In case of 2 primary interfaces the API refuses to safe and gives back a 400 with an error message:

Selection_187

On the user interface itself however we don't display a good message:

Selection_186

I think we should add a check in the front end similar to the ip address:

Selection_189

@j-white what is your take?

j-white:
Not ideal - the error indicates there is a problem, but doesn't tell you what.

patrick-schweizer:
@j-white I am stuck with the validation. Let's take a look on Tuesday?

cgorantla:
Added changed to UI to permit only one primary interface.
Screenshot from 2020-05-13 09-16-17
Screenshot from 2020-05-13 09-16-41

patrick-schweizer:
@cgorantla thank you for the front end changes! I tested them and they look good to me. The functionality is there and I think it's clear to the user that 2 x a primary  is not possible.

I found only one little glitch (not sure if it's worth addressing): create a node with 2 interfaces (1 primary, 2 secondary) and save the node.
Then go back and edit the primary interface. Click into the primary field and delete the "P". It won't let you change it back to "P" even though there is no other primary interface.

cgorantla:
> @cgorantla thank you for the front end changes! I tested them and they look good to me. The functionality is there and I think it's clear to the user that 2 x a primary is not possible.
> 
> I found only one little glitch (not sure if it's worth addressing): create a node with 2 interfaces (1 primary, 2 secondary) and save the node.
> Then go back and edit the primary interface. Click into the primary field and delete the "P". It won't let you change it back to "P" even though there is no other primary interface.

Should be fixed with latest commit.

Screenshot from 2020-05-18 13-51-10

patrick-schweizer:
> Should be fixed with latest commit.
It is. Looks good :-)

PR #2977, opened 2020-04-15
NMS-12645: Add smoke test for admin password change

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12645


smoke-test/src/test/java/org/opennms/smoketest/AdminUserIT.java
```diff
@@ -0,0 +1,58 @@
...
+package org.opennms.smoketest;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.openqa.selenium.By;
+import org.openqa.selenium.support.ui.ExpectedConditions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class AdminUserIT extends OpenNMSSeleniumIT {
```
j-white, line 38: Let's add this test to an existing class rather than creating a new one - saves us from spawning another stack.

PR #2982, opened 2020-04-20
NMS-12577: Prometheus Collector

j-white:
JIRA: https://issues.opennms.org/browse/NMS-12577

Add support for scraping metrics from HTTP endpoints using the Prometheus exposition format.


opennms-doc/guide-admin/src/asciidoc/text/performance-data-collection/collectors/prometheus.adoc
```diff
@@ -0,0 +1,132 @@
...
+[options="autowidth"]
+|===
+| Class Name          | `org.opennms.netmgt.collectd.prometheus.PrometheusCollector`
+| Package             | core
+| Supported on Minion | `Yes`
+|===
+
+===== Collector Parameters
+
+.Collector specific parameters for the WsManCollector
```
indigo423, line 24: PrometheusCollector vs. WsManCollector

features/prometheus-collector/pom.xml
```diff
@@ -0,0 +1,145 @@
...
+        <executions>
+          <execution>
+            <phase>package</phase>
+            <goals>
+              <goal>single</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+<!--
```
indigo423, line 51: Do we need comment?

opennms-config-tester/src/test/java/org/opennms/netmgt/config/tester/ConfigTesterTest.java
```diff
@@ -375,6 +375,11 @@ public void testPollerConfiguration() {
         testConfigFile("poller-configuration.xml");
     }
 
+    @Test
+    public void testPrometheusConfigFiles() {
```
indigo423, line 379: Thank you very much! :D

opennms-doc/guide-admin/src/asciidoc/text/performance-data-collection/collectors/prometheus.adoc
j-white: Fixed

features/prometheus-collector/pom.xml
j-white: Worth keeping, adding a comment to explain.

features/prometheus-collector/src/main/java/org/opennms/netmgt/collectd/prometheus/PrometheusCollector.java
```diff
@@ -0,0 +1,361 @@
...
+                        metric.visit(new MetricVisitor() {
+                            @Override
+                            public void visitCounter(Counter counter) {
+                                builder.withNumericAttribute(resource, group.getName(), attributeName, counter.getValue(), attributeTypeMapper.apply(AttributeType.COUNTER));
+                            }
+
+                            @Override
+                            public void visitGauge(Gauge gauge) {
+                                builder.withNumericAttribute(resource, group.getName(), attributeName, gauge.getValue(), attributeTypeMapper.apply(AttributeType.GAUGE));
+                            }
```
indigo423, line 237: Counter and Gauge are the only metrics we have ways to handle on our side.

indigo423:
Tested the Prometheus collector with the node_exporter, with RRD graphs for node load, disks and file systems.

Screenshot 2020-04-23 at 11 30 33

The fun parts are working :) 

When I've tested the collector over a Minion I ran into the case where the Node Exporter was not reachable from my Minion and I got the following data collection event. The error reason, in this case, was a little bit strange to understand the problem is the Minion can't connect to the Node Exporter, not sure if we have ways to provide a user-friendlier error message in this case:

```
Prometheus-Node-Exporter data collection on interface 192.168.152.243 failed because of the following condition: 'Collection failed : java.util.concurrent.ExecutionException: org.springframework.jms.UncategorizedJmsException: Uncategorized exception occurred during JMS processing; nested exception is javax.jms.JMSException: Error while attempting to add new Connection to the pool; nested exception is javax.jms.JMSException: Could not create Transport. Reason: java.io.IOException: Broker named 'localhost' does not exist.'.
```

I've fixed the connection between Minion and the Node Exporter and I got the [errors](https://paste.opennms.eu/emikumekuv.md) in the Minions karaf.log and the data collection doesn't work. I've used the minion.oci image from [here](https://app.circleci.com/pipelines/github/OpenNMS/opennms/5655/workflows/6b1cdf0d-9a5b-447a-bc52-c5aba09c8415/jobs/26071/artifacts).

I've created the following service in the collectd-configuration.xml

```xml
<service name="Prometheus-Node-Exporter" interval="30000" user-defined="false" status="on">
  <parameter key="collection" value="node_exporter" />
  <parameter key="thresholding-enabled" value="true" />
  <parameter key="url" value="http://INTERFACE_ADDRESS:9100/metrics" />
</service>
<collector service="Prometheus-Node-Exporter" class-name="org.opennms.netmgt.collectd.prometheus.PrometheusCollector" />
```

j-white:
Thanks for giving this a whirl. The Minion support was in fact broken, but is fixed now.


> When I've tested the collector over a Minion I ran into the case where the Node Exporter was not reachable from my Minion and I got the following data collection event. The error reason, in this case, was a little bit strange to understand the problem is the Minion can't connect to the Node Exporter, not sure if we have ways to provide a user-friendlier error message in this case:
> 
> ```
> Prometheus-Node-Exporter data collection on interface 192.168.152.243 failed because of the following condition: 'Collection failed : java.util.concurrent.ExecutionException: org.springframework.jms.UncategorizedJmsException: Uncategorized exception occurred during JMS processing; nested exception is javax.jms.JMSException: Error while attempting to add new Connection to the pool; nested exception is javax.jms.JMSException: Could not create Transport. Reason: java.io.IOException: Broker named 'localhost' does not exist.'.
> ```

I believe this error is unrelated - you should see a different exception if the node_exporter is not available. I agree there's room for improvement here, but out of the scope of this PR.


> 
> I've created the following service in the collectd-configuration.xml
> 
> ```
> <service name="Prometheus-Node-Exporter" interval="30000" user-defined="false" status="on">
>   <parameter key="collection" value="node_exporter" />
>   <parameter key="thresholding-enabled" value="true" />
>   <parameter key="url" value="http://INTERFACE_ADDRESS:9100/metrics" />
> </service>
> <collector service="Prometheus-Node-Exporter" class-name="org.opennms.netmgt.collectd.prometheus.PrometheusCollector" />
> ```

Consider using the meta-data DSL instead of the placeholder i.e.:
```
 <parameter key="url" value="http://${interface:address}:9100/metrics" />
```

mfuhrmann:
@jesse :-) This is really awesome!

I've also tested this branch and the collector created wonderful RRD files.

We should also provide default threshold definitions based on those metrics.

What about the collecting cycle? Any doubts to use a 1 minute collecting cycle? 1 minute cycle has the benefit that 1 trigger = 1 minute which makes it easier to understand. Also you can do in event definitions something like `CPU usage above %parm[threshold]%% for %parm[trigger]% minutes.`.

mfuhrmann:
Oh. It is already merged. I should refresh more often the tabs in my browser :joy:

PR #2984, opened 2020-04-21
NMS-12653: Remove remote poller

fooker:
Remove remote poller GUI, packaging and backend.

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12653


fooker:
@RangerRick can you have a look at the packaging and the .spec files?

RangerRick:
The RPM spec changes look fine.  I've removed the debian remote poller stuff, looks like there's still a lot to remove.  `system-report` references, `-Dpoller.logfile`, probably basically anything that shows in `git grep -i remoting`...

fooker:
Thanks for the Debian stuff. This is not complete right now. I just wanted to get your feedback because packaging feels like black magic to me...

opennms-base-assembly/src/main/filtered/etc/events/opennms.correlation.events.xml
```diff
@@ -1,26 +1,4 @@
 <events xmlns="http://xmlns.opennms.org/xsd/eventconf">
-   <event>
```
christianpape, line 2: Don't know. Maybe we want to keep such kind of event?

opennms-config/src/test/resources/org/opennms/netmgt/config/eventd/eventconf-speedtest/eventconf.xml
```diff
@@ -1129,160 +1129,6 @@
     <alarm-data reduction-key="%uei%:%dpname%:%nodeid%:%interface%:%service%" alarm-type="3" auto-clean="false" />
 -->
   </event>
-  <!-- Remote Poller Events -->
```
christianpape, line 1132: Same here. Should we keep different events for remote outages?

opennms-base-assembly/src/main/filtered/etc/events/opennms.remote.poller.events.xml
```diff
@@ -1,59 +0,0 @@
-<events xmlns="http://xmlns.opennms.org/xsd/eventconf">
```
christianpape, line 1: Same here. Should we keep separate events for remote outages?

opennms-web-api/src/main/java/org/opennms/web/svclayer/model/LocationMonitorListModel.java
```diff
@@ -162,10 +157,7 @@ public int compare(Entry<String, String> one, Entry<String, String> two) {
             });
             for (Entry<String, String> detail : details) {
                 if (
-                    !detail.getKey().equals(PollerBackEnd.HOST_NAME_KEY) &&
```
christianpape, line 165: Wait! What?

fooker: It's just to get it working. Have to revisit this code anyways.

opennms-base-assembly/src/main/filtered/etc/events/opennms.correlation.events.xml
fooker: Then we will recreate it...

opennms-base-assembly/src/main/filtered/etc/events/opennms.remote.poller.events.xml
fooker: Then we will recreate it...

PR #2990, opened 2020-04-30
NMS-12654: Update remote-poller model to link service to locations instead of individual RPs

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-12654

This PR includes the model changes and the deletion of the OnmsLocationManager and all of its classes for the RP removal. This branch will compile but of course some tests will fail, since it's too much effort to keep the tests green at this stage.


opennms-model/src/main/java/org/opennms/netmgt/model/OnmsLocationMonitor.java
```diff
@@ -1,123 +0,0 @@
...
- * view of the controlling OpenNMS daemon.
- *
- * Note: this class has a natural ordering that is inconsistent
- * with equals.
- *
- * @author <a href="mailto:brozow@opennms.org">Mathew Brozowski</a>
- * @author <a href="mailto:david@opennms.org">David Hustace</a>
- */
-@Entity
-@DiscriminatorValue(OnmsMonitoringSystem.TYPE_REMOTE_POLLER)
```
fooker, line 52: The constant can be removed, too.

fooker, line 52: As the sub-class is removed, the liquibase update should ensure that all values with this type are removed from the `monitoringsystems` table, too.

christianpape: Yes, you're soooo right.

christianpape: Good catch.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/AvailCalculator.java
```diff
@@ -154,23 +154,23 @@ private double uptime(TimeChunk chunk, TimeIntervalSequence uptime) {
     }
     
     public static class ServiceAvailCalculator {
-        Map<OnmsLocationMonitor, UptimeCalculator> m_uptimeCalculators = new HashMap<OnmsLocationMonitor, UptimeCalculator>();
```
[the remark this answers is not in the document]
christianpape: Done.

opennms-dao/src/main/java/org/opennms/netmgt/dao/support/DistributedStatusResourceType.java
```diff
@@ -117,11 +117,11 @@ public boolean isResourceTypeOnParent(OnmsResource parent) {
 
         final OnmsNode node = ResourceTypeUtils.getNodeFromResource(parent);
         final List<OnmsResource> resources = Lists.newLinkedList();
-        final Collection<LocationMonitorIpInterface> statuses = m_locationMonitorDao.findStatusChangesForNodeForUniqueMonitorAndInterface(node.getId());
+        final Collection<LocationIpInterface> statuses = m_locationSpecificStatusDao.findStatusChangesForNodeForUniqueMonitorAndInterface(node.getId());
 
-        for (LocationMonitorIpInterface status : statuses) {
-            String definitionName = status.getLocationMonitor().getLocation();
-            String id = status.getLocationMonitor().getId();
+        for (LocationIpInterface status : statuses) {
```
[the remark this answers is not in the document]
christianpape: Done.

PR #2994, opened 2020-05-04
NMS-12695: graphite telemetry adapter

RangerRick:
This PR adds basic support for the Graphite `plaintext` protocol to Telemetryd.

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12695

### DIFF
--- a/features/telemetry/protocols/graphite/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/graphite/adapter/GraphiteAdapter.java
+++ b/features/telemetry/protocols/graphite/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/graphite/adapter/GraphiteAdapter.java
@@ -100,7 +100,7 @@ public class GraphiteAdapter extends AbstractScriptedCollectionAdapter {
                     final GraphiteMetric metric = new GraphiteMetric(entry[0], entry[1], Long.valueOf(entry[2], 10));
                     final CollectionSet collectionSet = builder.build(agent, metric, metric.getTimestamp());
                     collectionSets.add(new CollectionSetWithAgent(agent, collectionSet));
-                } catch (final ScriptException e) {
+                } catch (final NumberFormatException | ScriptException e) {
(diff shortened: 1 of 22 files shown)


cgorantla:
Looks good to me. But if this not core network telemetry we should find a way to integrate this through integration api and make it available as plugin that we can load into OpenNMS

j-white:
> But if this not core network telemetry we should find a way to integrate this through integration api and make it available as plugin that we can load into OpenNMS

This can be "core" - the idea is to have an easy way to get arbitrary metrics in that aren't tied to specific vendor implementations.

cgorantla:
> > But if this not core network telemetry we should find a way to integrate this through integration api and make it available as plugin that we can load into OpenNMS
> 
> This can be "core" - the idea is to have an easy way to get arbitrary metrics in that aren't tied to specific vendor implementations.

:+1:

features/telemetry/protocols/graphite/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/graphite/adapter/GraphiteAdapter.java
```diff
@@ -95,12 +96,12 @@ public GraphiteAdapter(final AdapterDefinition adapterConfig, final MetricRegist
             if (entry.length != 3) {
                 LOG.warn("Unparseable graphite plaintext message: {}", line);
             } else {
-                final GraphiteMetric metric = new GraphiteMetric(entry[0], entry[1], Long.valueOf(entry[2], 10));
                 try {
+                    final GraphiteMetric metric = new GraphiteMetric(entry[0], entry[1], Long.valueOf(entry[2], 10));
```
j-white, line 100: The `NumberFormatException`  still won't get caught here right?

RangerRick: d'oh, you're right

PR #2997, opened 2020-05-05
NMS-12688: Use custom thread pool to run the Startup of consumers

cgorantla:
Without using the custom thread pool, the thread that is going to run `CompletableFuture.runAsync` seems to be coming from thread pool that is using different classloader (may be custom classloader from osgi ? ) than default class loader which is causing class loader issues while running telemetry adapters.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12688

### DIFF
--- a/core/ipc/sink/common/src/main/java/org/opennms/core/ipc/sink/common/AbstractMessageConsumerManager.java
+++ b/core/ipc/sink/common/src/main/java/org/opennms/core/ipc/sink/common/AbstractMessageConsumerManager.java
@@ -31,6 +31,9 @@ package org.opennms.core.ipc.sink.common;
 import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
 
 import org.opennms.core.ipc.sink.api.Message;
 import org.opennms.core.ipc.sink.api.MessageConsumer;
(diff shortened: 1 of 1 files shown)


core/ipc/sink/common/src/main/java/org/opennms/core/ipc/sink/common/AbstractMessageConsumerManager.java
```diff
@@ -44,13 +47,20 @@
...
 
     private static final Logger LOG = LoggerFactory.getLogger(AbstractMessageConsumerManager.class);
 
     public static final String SINK_INITIAL_SLEEP_TIME = "org.opennms.core.ipc.sink.initialSleepTime";
 
+    private final ThreadFactory threadFactory = new ThreadFactoryBuilder()
+            .setNameFormat("consumer-starter-%d")
+            .build();
+
+    protected final ExecutorService startupExecutor = Executors.newCachedThreadPool(threadFactory);
```
j-white, line 62: Let's make sure this thread pool gets shutdown.

j-white: We can merge without this, but let's make sure to circle back.

agalue:
I used the test environment mentioned on the JIRA issue to test the latest OCI image generated by CircleCI with the changes, and everything worked as expected. The problem is gone, no exceptions, RRDs are correctly updated.

@cgorantla Thanks a lot for fixing this problem.

RangerRick:
the issue was created with meridian 2019 tagged too, should this be backported to foundation-2019 @cgorantla @agalue ?

cgorantla:
Not sure if it is reproducible on `foundation-2019` but doesn't hurt to put it there.

PR #3005, opened 2020-05-12
NMS-12391: Use bulkhead to restrict number of threads

cgorantla:
During node scan, one node can have 100/1000s of interfaces and those interfaces
can have 10s of services. When those nodes are on minion, OpenNMS
asynchronously sends all requests to minion. This may result in too many
concurrent requests.

Idea is to restrict number of concurrent requests at a given point.
Using bulkhead to gate requests that are incoming with tunable timeout
Default timeout is 100msecs and max concurrent calls is 1000.
So when 1000 rpc requests are processing currently, only 10 extra requests per sec
will be allowed. This will allow not having a hard limit on threads but also
gives chance to tune the system depending on the load.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12391

### DIFF
--- a/core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcServerManager.java
+++ b/core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcServerManager.java
@@ -30,8 +30,9 @@ package org.opennms.core.ipc.rpc.kafka;
 
 import static org.opennms.core.ipc.common.kafka.KafkaRpcConstants.MAX_CONCURRENT_CALLS_PROPERTY;
 import static org.opennms.core.ipc.common.kafka.KafkaRpcConstants.MAX_DURATION_BULK_HEAD;
-import static org.opennms.core.ipc.common.kafka.KafkaRpcConstants.MAX_WAIT_DURATION_BULK_HEAD;
 import static org.opennms.core.ipc.common.kafka.KafkaRpcConstants.SINGLE_TOPIC_FOR_ALL_MODULES;
+import static org.opennms.core.rpc.api.RpcClientFactory.RPC_COUNT;
+import static org.opennms.core.rpc.api.RpcClientFactory.RPC_FAILED;
 import static org.opennms.core.tracing.api.TracerConstants.TAG_LOCATION;
 import static org.opennms.core.tracing.api.TracerConstants.TAG_RPC_FAILED;
 import static org.opennms.core.tracing.api.TracerConstants.TAG_SYSTEM_ID;
@@ -52,6 +53,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
@@ -80,6 +82,10 @@ import org.opennms.distributed.core.api.MinionIdentity;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.Meter;
+import com.codahale.metrics.MetricRegistry;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
 import com.google.common.math.IntMath;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
@@ -108,8 +114,12 @@ public class KafkaRpcServerManager {
             .maxRate(5).every(Duration.standardSeconds(30))
             .build();
     private Map<String, RpcModule<RpcRequest, RpcResponse>> modulesById = new ConcurrentHashMap<>();
+    final static String ACTIVE_RPC_REQUESTS = "activeRpcRequests";
+    final static String AVAILABLE_CONCURRENT_CALLS = "availableConcurrentCalls";
+    private final static String MAX_ALLOWED_CONCURRENT_CALLS = "maxAllowedConcurrentCalls";
     private final Properties kafkaConfig = new Properties();
     private final KafkaConfigProvider kafkaConfigProvider;
+    private final MetricRegistry metrics;
     private KafkaProducer<String, byte[]> producer;
     private MinionIdentity minionIdentity;
     private Integer maxBufferSize = KafkaRpcConstants.MAX_BUFFER_SIZE_CONFIGURED;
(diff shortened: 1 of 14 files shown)


core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcServerManager.java
```diff
@@ -287,8 +301,12 @@ public void run() {
                             }
                             // Should have complete message by this point.
                             final ByteString requestMessage = rpcContent;
+                            // Incoming RPC requests are gated to restrict number of threads used by Kafka RPC.
+                            // Need to process the requests after timeout even though permission acquisition fails.
```
j-white, line 305: Can you elaborate on why we still need to process the request?

If we continue processing the request there may still be >maxConcurrentCalls active.

core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcServerManager.java
```diff
@@ -162,6 +171,11 @@ public void init() throws IOException {
         });
         tracerRegistry.init(minionIdentity.getLocation() + "@" + minionIdentity.getId());
 
+        BulkheadConfig bulkheadConfig = BulkheadConfig.custom()
+                .maxConcurrentCalls(maxConcurrentCalls)
+                .maxWaitDuration(java.time.Duration.ofMillis(maxWaitDuration))
+                .build();
+        bulkhead = Bulkhead.of("ipc-rpc-kafka", bulkheadConfig);
```
j-white, line 178: Let's expose the bulk head metrics like we've done here: https://github.com/OpenNMS/opennms/blob/develop/features/dnsresolver/netty/src/main/java/org/opennms/netmgt/dnsresolver/netty/NettyDnsResolver.java#L126

core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcServerManager.java
cgorantla: I'm thinking this as gradual increase of threads instead of a burst that could cause issues with too many threads.  As per current defaults,  only 10 extra threads will be allowed per sec after a threshold of `maxConcurrentCalls`.   Currently burst only happens with `Provisiond` and this allows it to not to have a burst of requests within few secs.

core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcServerManager.java
```diff
@@ -329,10 +347,12 @@ private void handleRequest(RpcMessageProto rpcRequestProto, ByteString rpcConten
                 }
                 // Finish minion Span
                 minionSpan.finish();
+                bulkhead.onComplete();
```
j-white, line 350: This should only be called if `tryAcquirePermission` returned `true`.

core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcServerManager.java
j-white: Makes sense - this will slow down and put back-pressure on the topic for bursts.

Let's also make sure to expose a Gauge for the actual number of threads active - since this may differ from the number of active calls being tracked by the bulkhead.

cgorantla:
@j-white Handled review comments. Added `requestSent` and `requestFailed` metrics as well on the Minion.   
@Bonrob2 Please review documentation. 
file:///home/chandra/Pictures/Screenshot%20from%202020-05-19%2015-07-21.png

opennms-doc/guide-admin/src/asciidoc/text/minion/kafka-rpc.adoc
```diff
@@ -0,0 +1,20 @@
+// Allow GitHub image rendering
+:imagesdir: ../../images
+
+===  Tuning Kafka RPC on Minion
+
+To avoid too many requests hogging the system at once, Kafka RPC on Minion limits the maximum number of concurrent RPC requests.
+Currently max number of concurrent requests is set to 1000 with maximum wait time of 100ms.
```
Bonrob2, line 7:
```suggestion
Currently the maximum number of concurrent requests is set to 1,000 with a maximum wait time of 100ms.
```

opennms-doc/guide-admin/src/asciidoc/text/minion/kafka-rpc.adoc
```diff
@@ -0,0 +1,20 @@
...
+With these settings when concurrent requests reach 1000, it only allow 10 extra requests per sec.
```
Bonrob2, line 8:
```suggestion
With these settings, when concurrent requests reach 1000, Kafka RPC will allow only 10 extra requests per second.
```

opennms-doc/guide-admin/src/asciidoc/text/minion/kafka-rpc.adoc
```diff
@@ -0,0 +1,20 @@
...
+
+These settings can be tuned as below
```
Bonrob2, line 10:
```suggestion
Tune these settings as below:
```

cgorantla:
@Bonrob2 Thanks for the review. Handled the comments.

core/ipc/rpc/kafka/src/main/java/org/opennms/core/ipc/rpc/kafka/KafkaRpcServerManager.java
```diff
@@ -315,6 +368,8 @@ private void handleRequest(RpcMessageProto rpcRequestProto, ByteString rpcConten
             setTagsOnMinion(rpcRequestProto, request, minionSpan);
             // Modules may run the execution in their own thread pool.
             CompletableFuture<RpcResponse> future = module.execute(request);
+            final Meter requestSentMeter = getMetrics().meter(MetricRegistry.name(request.getLocation(), module.getId(), RPC_COUNT));
```
j-white, line 371: We can omit location here since all the requests will be for the same location right?

j-white: Maybe we should rename to `requestsReceived`? From this perspective we're processing the received request.

j-white:
Looks good. I think it's important that we also collect & graph the metrics by default so that we have the visibility we need to tune if necessary.

That said, I'm good with merging this as is, but I think we should come back with the JMX datacollection config & graph definition(s) before this is part of a release.

cgorantla:
> Looks good. I think it's important that we also collect & graph the metrics by default so that we have the visibility we need to tune if necessary.
> 
> That said, I'm good with merging this as is, but I think we should come back with the JMX datacollection config & graph definition(s) before this is part of a release.

Sure. Created https://issues.opennms.org/browse/NMS-12737 for handling this as well as add any missing RPC/Sink related metrics.

PR #3016, opened 2020-05-19
NMS-12723: Events forwarded by Kafka Producer doesn't have any parameters set

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12723

### DIFF
--- a/core/daemon/src/main/java/org/opennms/netmgt/daemon/DaemonTools.java
+++ b/core/daemon/src/main/java/org/opennms/netmgt/daemon/DaemonTools.java
@@ -28,6 +28,8 @@
 
 package org.opennms.netmgt.daemon;
 
+import java.util.function.Consumer;
+
 import org.apache.commons.lang.StringUtils;
 import org.opennms.netmgt.events.api.EventConstants;
 import org.opennms.netmgt.events.api.EventIpcManagerFactory;
@@ -36,8 +38,6 @@ import org.opennms.netmgt.xml.event.Event;
 import org.opennms.netmgt.xml.event.Parm;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import java.util.function.Consumer;
 
 public class DaemonTools {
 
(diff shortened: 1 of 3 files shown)


features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/ProtobufMapper.java
```diff
@@ -220,16 +223,26 @@ public ProtobufMapper(EventConfDao eventConfDao, HwEntityDao hwEntityDao, Sessio
...
-        if (event.getInterface() != null) {
-            builder.setIpAddress(event.getInterface());
+
+        getString(event.getInterface()).ifPresent(builder::setIpAddress);
+
+        for (Parm parm : event.getParmCollection()) {
+            if (parm.getParmName() == null || parm.getValue() == null) {
+                continue;
+            }
+            String value = EventDatabaseConstants.escape(parm.getValue().getContent() == null ? "" :
```
j-white, line 240: We don't need to escape here do we?

```suggestion
            String value = parm.getValue().getContent() == null ? "" : parm.getValue().getContent();
```

core/daemon/src/main/java/org/opennms/netmgt/daemon/DaemonTools.java
```diff
@@ -60,7 +60,6 @@ public static void handleReloadEvent(Event e, String daemonName, Consumer<Event>
                 LOG.info("Reload successful.");
 
                 ebldr = new EventBuilder(EventConstants.RELOAD_DAEMON_CONFIG_SUCCESSFUL_UEI, daemonName);
-                ebldr.addParam(EventConstants.PARM_DAEMON_NAME, daemonName);
```
j-white, line 63: Related?

cgorantla: Not related but saw duplicate params for this event while validating this.

cgorantla:
Except for log/display all the fields are filled for event.
All fields are filled for Alarm.

core/daemon/src/main/java/org/opennms/netmgt/daemon/DaemonTools.java
j-white: Ah I see - it's still added bellow.

PR #3018, opened 2020-05-22
use lighttpd to get RPMs

RangerRick:
This reduced the horizon image from 2.7gb to 2.1gb.  I imagine a similar reduction in the minion and sentinel images will occur when this merges forward (and they are updated).


RangerRick:
FYI I have equivalent branches for foundations 2017 through 2019 too, with Minion and Sentinel updated to use the same process.  Appears to pretty consistently shrink Horizon by about half a gig, and Minion/Sentinel by about 200MB.

PR #3029, opened 2020-06-04
NMS-7547: fix UTF-8 comparison decoding hex strings

RangerRick:
We pulled this out of the backlog this week, it's an old patch that Seth did that supposedly broke some Enlinkd tests, but we've refactored plenty of the related code and everything passes here.  I've gone ahead and made a `foundation-2019` version of it [here](https://app.circleci.com/pipelines/github/OpenNMS/opennms/6315/workflows/50ce4c23-550c-4c4c-8b2c-41bf48eed934) just to make sure it passes there too, and and as long as it does, I'd say we should hopefully be safe to merge.

JIRA: https://issues.opennms.org/browse/NMS-7547

### DIFF
--- a/opennms-services/src/main/java/org/opennms/netmgt/enlinkd/snmp/LldpLocalGroupTracker.java
+++ b/opennms-services/src/main/java/org/opennms/netmgt/enlinkd/snmp/LldpLocalGroupTracker.java
@@ -31,6 +31,7 @@ package org.opennms.netmgt.enlinkd.snmp;
 import org.opennms.core.utils.LldpUtils;
 import org.opennms.core.utils.LldpUtils.LldpChassisIdSubType;
 import org.opennms.netmgt.model.LldpElement;
+import org.opennms.netmgt.snmp.AbstractSnmpValue;
 import org.opennms.netmgt.snmp.AggregateTracker;
 import org.opennms.netmgt.snmp.ErrorStatus;
 import org.opennms.netmgt.snmp.ErrorStatusException;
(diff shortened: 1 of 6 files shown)


RangerRick:
Hah!  Bamboo actually still failed: https://bamboo.opennms.org/browse/OPENNMS-OPENNMS697-IT-1/log

Gonna see if I can reproduce it anywhere else.  Marked as `incomplete` until I can see if it fails locally.

cgorantla:
Also https://github.com/OpenNMS/opennms/pull/1169  notes that workaround needs to be removed once this is fixed

RangerRick:
good catch

RangerRick:
looks like the bamboo failure showed up [in the 2019 version too](https://app.circleci.com/pipelines/github/OpenNMS/opennms/6315/workflows/50ce4c23-550c-4c4c-8b2c-41bf48eed934/jobs/30724); maybe it's a flapper

opennms-services/src/main/java/org/opennms/netmgt/enlinkd/snmp/LldpLocalGroupTracker.java
```diff
@@ -92,10 +93,10 @@
     public static String getDisplayable(final SnmpValue snmpValue) {
         String decodedsnmpValue = snmpValue.toHexString();
         try {
-            if (snmpValue.isDisplayable())
+            if (AbstractSnmpValue.allBytesUTF_8(snmpValue.getBytes()))
```
cgorantla, line 96: So basically we are limiting the displayable to UTF-8 validation. 
If Enlinkd expects this way, that's fine. 
Otherwise I'm good with skipping the test validation for values in order to get the original fix of preventing `ArrayIndexOutOfBoundsException`

RangerRick: I could still make it fall back to trying ASCII at least.  Though in this case that's still a single `ï` character, which also seems wrong -- I suspect these devices have a bug in their "null entry" responses that cause this.

RangerRick: It _used_ to return `cf` though, so treating it as UTF-8 text seems to keep the same behavior here.

RangerRick: I'm open to either way of doing it, if you have a preference.

cgorantla: Let's go with this one.

RangerRick:
FYI, it was _not_ a flapper, the walk used in this test does have weird data that is either bad UTF-8 or puzzling ASCII (`ï`).

PR #3035, opened 2020-06-11
NMS-12750: Fix delta switched for Netflow5

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12750


features/telemetry/protocols/netflow/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/adapter/common/NetflowMessage.java
```diff
@@ -110,6 +110,9 @@ public Integer getEngineType() {
 
     @Override
     public Long getDeltaSwitched() {
+        if (getNetflowVersion().equals(NetflowVersion.V5)) {
+            return flowMessageProto.hasFirstSwitched() ? flowMessageProto.getFirstSwitched().getValue() : null;
```
fooker, line 114: Maybe we can always use this as an fallback for `deltaSwitchen` without making it protocol specific

cgorantla: Yeah, If you think that would be correct protocol wise, I can change that.

features/telemetry/protocols/netflow/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/adapter/common/NetflowMessage.java
```diff
@@ -110,7 +110,7 @@ public Integer getEngineType() {
 
     @Override
     public Long getDeltaSwitched() {
-        return flowMessageProto.hasDeltaSwitched() ? flowMessageProto.getDeltaSwitched().getValue() : null;
+        return flowMessageProto.hasDeltaSwitched() ? new Long(flowMessageProto.getDeltaSwitched().getValue()) : getFirstSwitched();
```
fooker, line 113: I don't think the extra boxing with `new Long` is necessary here.

cgorantla: `ternary` operator tries to auto unbox null and encounters NPE.  WIth boxing the first expression it doesn't unbox the second expression.

PR #3040, opened 2020-06-17
NMS-12729: Add openconfig support for Telemetryd

cgorantla:
`OpenConfig` uses gRPC for streaming telemetry data.
This PR adds support for `StreamListener` on Telemetryd which can subscribe to telemetry data through gRPC clients and tunnels the received data to openconfig adapter.
`OpenConfig` configuration for each  Interface is fetched from metadata when initializing Telemetryd as well as whenever new Interface is discovered. 

There are two modes that Junos specifically operates. It's own openconfig model and `gNMI` mode.
Currently this PR supports `jti` mode by default but plans to make it configurable.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12729


core/ipc/rpc/utils/src/main/java/org/opennms/core/rpc/utils/MetadataUtils.java
```diff
@@ -0,0 +1,101 @@
...
+    @Autowired
+    private EntityScopeProvider entityScopeProvider;
+
+    public Map<InetAddress, Integer> getIpInterfacesWithContext(String context) {
+
+        Map<InetAddress, Integer> matchingIpInterfaceMap = new HashMap<>();
+        sessionUtils.withReadOnlyTransaction(() -> {
+            try {
+                Map<InetAddress, Integer> ipInterfaces = ipInterfaceDao.getInterfacesForNodes();
+                ipInterfaces.forEach(((inetAddress, nodeId) -> {
```
fooker, line 66: This should be done on the database level. The `EntityScopeProvider` is meant to be used for for Meta-DSL as it provides a whole set of additional values. In fact, I would like to see this as part of the `IpInterfaceDao`.

core/ipc/rpc/utils/src/main/java/org/opennms/core/rpc/utils/MetadataUtils.java
```diff
@@ -0,0 +1,101 @@
...
+
+            } catch (Exception e) {
+                LOG.error("Encountered exception while getting interfaces with context {}", context, e);
+            }
+            return null;
+        });
+        return matchingIpInterfaceMap;
+    }
+
+    public Map<String, String> getConfigFromMetadata(Integer nodeId, String ipAddress, String context) {
```
fooker, line 81: This should be replaced by a call to `Interpolator.interpolateStrings` where you have to pass in a map of Meta-DSL expressions and get a map of interpolated values. Preferable passed on to a `BeanWrapper` which created a domain specific config bean.

cgorantla:
> About the client interface: We have to decide how much management the client should do.
> First option is to have a very simple client which just propagates errors.
> Second option is a client which does reconnects and therefore runs in it's own thread or leverages a netty thread pool. It also needs to keep track of subscriptions, because it has to resubscribe on connection failure.
> 
> I see some problems with the fact that `telemetryd` has special handling for the `StreamListeners`. The listener itself should figure out which targets to connect to. Preferable by talking to the parser, as the listeners are protocol agnostic.
> I would imagine to have some kind of tracking interface for the parser `StreamParser` which provides a stream of connect and disconnect events providing (host, port) tuples for the listener to connect to. All parsers working with the `StreamListener` has to implement this interface (similar to how all parsers for the `TcpListener` must implement the `TcpParser` interface). The implementation would then hook up to the events.
> 
> I like that this is driven my metadata. But it should be more flexible. The user will have a hard time to configure metadata in the `openconfig` context as only the `requisition` context is editable using the UI.
> To make this more usable, the parser should allow to specify meta-dsl expressions which will be interpolated. I.e. the parser config has a parameter called `hostExpression` containing `${openconfig:port|requisition:openconfig_port}` and the node is used as target only if this evaluates to something not empty. Other options should be treated similar.
> 
> I'm uncertain of the event listening process is complete. What happens if user changes the requisition by adding or removing openconfig specific attributes. Does these connections get opened /closed without breaking unchanged connections?

Thanks for the review.  I vote for the client to be simple and have only one grpc connection and to propagate errors through events.  Then we can have detector/poller mechanism to know when to connect.

On metadata, yes. It would be best to have this in `requisition` context as that is editable in UI. I will add this.
For handling metadata changes to openconfig, a `reload daemon` event is needed to update the current connections.

I will address the other comments individually.

fooker:
> On metadata, yes. It would be best to have this in `requisition` context as that is editable in UI. I will add this.
> For handling metadata changes to openconfig, a `reload daemon` event is needed to update the current connections.

To avoid this problem completely we should use used-defined expression instead of fixed keys. By doing so, the user can decide if he wants to use the requisition editor or some complex provisioning mechanism.

features/telemetry/api/src/main/java/org/opennms/netmgt/telemetry/api/receiver/StreamListener.java
```diff
@@ -0,0 +1,36 @@
...
+ * For more information contact:
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.netmgt.telemetry.api.receiver;
+
+
+public interface StreamListener extends Listener {
```
fooker, line 32: The whole interface is not required.

features/telemetry/stream-listener/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -0,0 +1,29 @@
...
+  <reference id="telemetryRegistry" interface="org.opennms.netmgt.telemetry.api.registry.TelemetryRegistry" availability="mandatory" />
+  <reference id="openConfigClientFactory" interface="org.opennms.features.openconfig.api.OpenConfigClientFactory" availability="optional"/>
+  <reference id="eventSubscriptionService" interface="org.opennms.netmgt.events.api.EventSubscriptionService" availability="mandatory"/>
+
+
+  <!-- Stream Listener Factory -->
+  <bean id="telemetryStreamListenerFactory" class="org.opennms.netmgt.telemetry.listeners.factory.TelemetryStreamListenerFactory">
+    <argument ref="telemetryRegistry" />
+    <argument ref="eventSubscriptionService"/>
+    <property name="openConfigClientFactory" ref="openConfigClientFactory" />
```
fooker, line 20: Make this an argument, too?

features/telemetry/stream-listener/src/main/java/org/opennms/netmgt/telemetry/listeners/factory/TelemetryStreamListenerFactory.java
```diff
@@ -0,0 +1,82 @@
...
+        // Expects only one parser for stream listener.
+        StreamParser parser = parsers.get(0);
+        return new TelemetryStreamListener(listenerDefinition.getName(),
+                parser,
+                openConfigClientFactory,
+                eventSubscriptionService,
+                telemetryRegistry.getMetricRegistry());
+    }
+
+    public void setOpenConfigClientFactory(OpenConfigClientFactory openConfigClientFactory) {
```
fooker, line 79: Pass that as an constructor argument, too?

features/telemetry/stream-listener/src/main/java/org/opennms/netmgt/telemetry/listeners/StreamParser.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Invoked initially when stream listener is starting
+     */
+    List<Config> getMatchingConfig();
+
+
+    CompletableFuture<?> parse(final ByteBuf buffer,
+                               final InetSocketAddress remoteAddress) throws Exception;
+}
```
fooker, line 54: This is still to much bound to OpenConfig. Can we push the connection part even further to the parser? I imagine the following: The Listener is responsible for identifyng which connections should exit whereas the parser is responsible for opening and closing the connections as dictated by the listener. A proposal for a according interface:

```suggestion
public interface StreamParser extends Parser {
    public interface Connection extends Closeable {}

    public Connection connect(final int nodeId, final String ipAddress) throws Exception;
}
```

The listener then has to keep track of connection instances.

features/telemetry/stream-listener/src/main/java/org/opennms/netmgt/telemetry/listeners/TelemetryStreamListener.java
```diff
@@ -0,0 +1,227 @@
...
+
+    private static final int DEFAULT_PORT = 50051;
+    private static final int DEFAULT_FREQUENCY = 300000; //5minutes
+    private final String name;
+
+    private final StreamParser parser;
+
+    private final MetricRegistry metrics;
+
+    private final OpenConfigClientFactory openConfigClientFactory;
```
fooker, line 78: This should live in the `OpenConfigParserFactory`.

j-white:
Closing in favor of https://github.com/OpenNMS/opennms/pull/3113

PR #3053, opened 2020-06-30
NMS-12784: Remove key for the events sent to kafka 

cgorantla:
Chose to send events without key since number of events could be too many and this may be better way to evenly distribute these messages.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12784


agalue:
I just realized that the fix was merged to `release-26.x`, but this also affects Meridian 2019 (as mentioned in the JIRA issue), meaning the changes should be cherry-picked into `foundation-2019`.

cgorantla:
Pushed this to `foundation-2019` with d962df4794ce0bc3e06c5bb3d78d74660a7d1fb3

PR #3054, opened 2020-07-01
NMS-12753: Add encryption option for snmp config

cgorantla:
Add encryption for snmp credentials

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12753


RangerRick:
Shouldn't this PR be against `features/space`?

opennms-bamboo:
Welcome CI/CD System (cicd-system@opennms.com).

I could not find a signed OpenNMS Contributor Agreement (OCA) assigned to the username CI/CD System (cicd-system@opennms.com).

core/snmp/profile-mapper/src/test/java/org/opennms/core/snmp/profile/mapper/SnmpProfileMapperIT.java
```diff
@@ -121,6 +121,7 @@ public void testSnmpProfileMapperWithCustomOID() {
         int timeout = 6000;
         long ttl = 7000;
         URL url = getClass().getResource("/snmp-config2.xml");
+        url.getPath().toString();
```
RangerRick, line 124: Does this do anything?  Seems like a no-op.

opennms-config-jaxb/src/main/java/org/opennms/netmgt/config/snmp/Configuration.java
```diff
@@ -942,6 +957,13 @@ public boolean equals(final Object obj) {
         } else if (!m_writeCommunity.equals(other.m_writeCommunity)) {
             return false;
         }
+        if (m_ttl == null) {
+            if (other.m_ttl != null) {
+                return false;
+            }
+        } else if (!m_ttl.equals(other.m_ttl)) {
+            return false;
+        }
```
RangerRick, line 966: If `m_encrypted` is included in the hashcode above, shouldn't it be part of the `equals()` too?

opennms-config/src/main/java/org/opennms/netmgt/config/SnmpPeerFactory.java
```diff
@@ -426,7 +443,7 @@ public SnmpConfig getSnmpConfig() {
      * with many other attributes.  Uses new classes the wrap JAXB-generated code to
      * help with merging, comparing, and optimizing definitions.  Thanks for your
      * initial work on this Gerald.
-     *
+     * <p>
      * Puts a specific IP address with associated read-community string into
      * the currently loaded snmp-config.xml.
```
RangerRick, line 448: `</p>` ?  ;)

opennms-config-jaxb/src/main/java/org/opennms/netmgt/config/snmp/Configuration.java
cgorantla: yeah, it doesn't need to be part of hascode as well. Don't need this for comparison

opennms-config/src/main/java/org/opennms/netmgt/config/SnmpPeerFactory.java
cgorantla: IntelliJ auto-generated one, removed

cgorantla:
Moved persisting key (digested) to SCV instead of saving randomID in Postgres.
@RangerRick can you take a one more look ?

core/api/src/main/java/org/opennms/core/config/api/TextEncryptor.java
```diff
@@ -0,0 +1,37 @@
...
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.core.config.api;
+
+public interface TextEncryptor {
+
+    String encrypt(String alias, String key, String text);
```
j-white, line 33: Would help to explain the use of the `alias` and `key` attribute here.

core/encrypt-util/src/main/java/org/opennms/core/text/encryptor/TextEncryptorImpl.java
```diff
@@ -0,0 +1,108 @@
...
+    public String encrypt(String alias, String key, String text) {
+        try {
+            final AES256TextEncryptor textEncryptor = new AES256TextEncryptor();
+            String password = getPasswordFromCredentials(alias, key);
+            if (!Strings.isNullOrEmpty(password)) {
+                textEncryptor.setPassword(password);
+                return textEncryptor.encrypt(text);
+            }
+        } catch (Exception e) {
+            LOG.error("Exception while encrypting {} with key {}", text, key, e);
```
j-white, line 65: I'd opt to re-throw here instead. If someone is calling `encrypt` I wouldn't expect it to silently fail and return a un-encrypted payload back.

core/encrypt-util/src/main/java/org/opennms/core/text/encryptor/TextEncryptorImpl.java
```diff
@@ -0,0 +1,108 @@
...
+    public String decrypt(String alias, String key, String encrypted) {
+        final AES256TextEncryptor textEncryptor = new AES256TextEncryptor();
+        try {
+            String password = getPasswordFromCredentials(alias, key);
+            if (!Strings.isNullOrEmpty(password)) {
+                textEncryptor.setPassword(password);
+                return textEncryptor.decrypt(encrypted);
+            }
+        } catch (Exception e) {
+            LOG.error("Exception while decrypting {} with key {}", encrypted, key, e);
```
j-white, line 81: Similarly here.

features/scv/jceks-impl/src/main/java/org/opennms/features/scv/jceks/JCEKSSecureCredentialsVault.java
```diff
@@ -86,11 +87,15 @@ public JCEKSSecureCredentialsVault(String keystoreFile, String password, byte[]
     }
 
     public JCEKSSecureCredentialsVault(String keystoreFile, String password, byte[] salt, int iterationCount, int keyLength) {
-        m_password = Objects.requireNonNull(password).toCharArray();
         m_salt = Objects.requireNonNull(salt);
         m_iterationCount = iterationCount;
         m_keyLength = keyLength;
         m_keystoreFile = new File(keystoreFile);
+        String keyFromProperties = System.getProperty(SCV_KEY_PROPERTY);
```
j-white, line 94: What's the purpose of this override? This logic should probably be in the application context instead.

opennms-doc/guide-admin/src/asciidoc/text/snmp/snmp-encryption.adoc
```diff
@@ -0,0 +1,17 @@
...
+=== Encrypting SNMP Credentials
+
+SNMP credentials like readCommunity/WriteCommunity, authPassPhrase/privPassPhrase can be saved on the file system in encrypted format.
+
+==== Enabling encrypting on SNMP Config
+
+Specifying a key for encryption would enable encryption on SNMP config.
+[source, sh]
+----
+echo 'org.opennms.snmp.encryption.key=abc12345' >> "$OPENNMS_HOME/etc/opennms.properties.d/snmp-encryption.properties"
```
j-white, line 11: Requiring the key set via a system property and storing it directly on the file-system negates the benefit subsequently storing it in SCV.

cgorantla: Yeah, not storing it directly in SCV.  They key is digested to a password and then that is used to encrypt and decrypt password.
Both key/password are stored in SCV.

j-white: Got it. Is there any need to expose this key then? Seems like we could keep this fixed and then expose a simple boolean on/off instead.

cgorantla: yeah, we could keep it fixed since the digested password will always be different for each application and key never changes in the life time of application.

cgorantla:
> Some of these unit tests are pretty dependent upon the type of garbage collector the testing framework JVM is using. (ex: G1GC vs ParallelGC vs SerialGC, etc.) When I was testing the build/test process against JDK 11, I had to force the use of ParalellGC by adding -XX:+UseParallelGC to the argLine prooperty in the top level pom.xml.

Thanks @roskens. Good to know. Will check that

cgorantla:
@j-white Thanks for the review. Can you take one more look.

cgorantla:
> > Some of these unit tests are pretty dependent upon the type of garbage collector the testing framework JVM is using. (ex: G1GC vs ParallelGC vs SerialGC, etc.) When I was testing the build/test process against JDK 11, I had to force the use of ParalellGC by adding -XX:+UseParallelGC to the argLine prooperty in the top level pom.xml.
> 
> Thanks @roskens. Good to know. Will check that

Looks like that still fails for this test. https://circleci.com/gh/OpenNMS/opennms/34950#tests/containers/2

core/encrypt-util/src/test/java/org/opennms/core/text/encryptor/TextEncryptorTest.java
```diff
@@ -0,0 +1,89 @@
...
+import org.opennms.features.scv.jceks.JCEKSSecureCredentialsVault;
+import org.springframework.beans.factory.annotation.Autowired;
+
+
+public class TextEncryptorTest {
+
+    @Rule
+    public TemporaryFolder tempFolder = new TemporaryFolder();
+
+    @Autowired
```
j-white, line 52: No Spring context is used here.

core/api/src/main/java/org/opennms/core/config/api/TextEncryptor.java
```diff
@@ -0,0 +1,48 @@
...
+ * For more information contact:
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.core.config.api;
+
+/**
+ * For a given context/alias, encrypt/decrypt text with specified encryptionKey.
```
j-white, line 32: I find the use of the `encryptionKey` here a little confusing.

We hash the key with some salt using `StrongPasswordEncryptor` to generate a unique password - this password is stored in SCV and we subsequently use that to encrypt/decrypt the payload.

If we call `decrypt` after `encrypt` with a given key, the key is ignored (since the actual password is retrieved from SCV).
If we call `encrypt` with a different key, but the same alias, then original key is still used.

Given this, it seems like the key can be dropped completed and we can just generate a random password for a given alias and store/retrieve this in SCV.

opennms-doc/guide-admin/src/asciidoc/text/snmp/snmp-encryption.adoc
```diff
@@ -0,0 +1,19 @@
...
+==== Enabling encrypting on SNMP Config
+
+Enable encryption on SNMP config by setting following system property.
+[source, sh]
+----
+echo 'org.opennms.snmp.encryption.enabled=true' >> "$OPENNMS_HOME/etc/opennms.properties.d/snmp-encryption.properties"
+----
+
+SNMP config that is modified on User Interface will be encrypted directly on file system.
+Config that is updated on the file system needs to be saved with karaf command `opennms-snmp:save-config` to get encrypted.
```
j-white, line 15: Would be good to document how the key used to encrypt/decrypt the attributes is managed, the fact that the key is stored in SCV and what alias the key is stored under.

opennms-config-jaxb/src/main/java/org/opennms/netmgt/config/snmp/Configuration.java
```diff
@@ -763,6 +767,15 @@ public void setTTL(Long ttl) {
         m_ttl = ttl;
     }
 
+
+    public Boolean getEncrypted() {
+        return m_encrypted != null ? m_encrypted : false;
+    }
+
+    public void setEncrypted(Boolean encrypted) {
+        this.m_encrypted = encrypted;
```
j-white, line 776: `this.m_`

core/api/src/main/java/org/opennms/core/config/api/TextEncryptor.java
cgorantla: Yeah, initially I was saving the key,value in `Credentials` attributes. Since that's moved to Credentials and there is no user input for the key guess that can be dropped.

opennms-doc/guide-admin/src/asciidoc/index.adoc
```diff
@@ -399,6 +399,10 @@ include::text/alarm-correlation/situation-feedback.adoc[]
 == Meta-Data
 include::text/meta-data.adoc[]
 
+[[ga-snmp-encryption]]
+== Snmp Encryption
+include::text/snmp/snmp-encryption.adoc[]
```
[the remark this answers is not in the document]
Bonrob2: Spoke with Chandra, and I think we agree it should go in 10.3.1. Provisioning the SNMP Configuration.

opennms-doc/guide-admin/src/asciidoc/index.adoc
```diff
@@ -399,6 +399,10 @@ include::text/alarm-correlation/situation-feedback.adoc[]
 == Meta-Data
 include::text/meta-data.adoc[]
 
+[[ga-snmp-encryption]]
+== Snmp Encryption
```
Bonrob2, line 403:
```suggestion
== SNMP Encryption
```

Bonrob2: And move to the end of section 10.3.1: Provisioning the SNMP Configuration.

PR #3057, opened 2020-07-02
NMS-12768: Add ignore certificates property for ElasticSearch

cgorantla:
Add property to allow insecure SSL/HTTPS connections

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12768


cgorantla:
> It works as intended.
> 
> I compiled the branch, started OpenNMS and enabled `opennms-es-rest` and `opennms-alarm-history-elastic` for testing purposes.
> 
> Both features work against an ES instance running version 7.6.1 via HTTPS with a self-signed CA and certificate generated according to ES documentation.

Thanks for the confirmation @agalue

j-white:
This new field should be documented here: https://docs.opennms.org/opennms/releases/26.1.1/guide-admin/guide-admin.html#ga-elasticsearch-integration-configuration

PR #3074, opened 2020-07-21
NMS-12818: Log and throw custom exception when the response  is truncated.

cgorantla:
Although call stack doesn't provide good clue on where this is happening
this looks like more probable cause for the bug.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12818

### DIFF
--- a/core/snmp/proxy-rpc-impl/src/main/java/org/opennms/netmgt/snmp/proxy/common/SnmpProxyRpcModule.java
+++ b/core/snmp/proxy-rpc-impl/src/main/java/org/opennms/netmgt/snmp/proxy/common/SnmpProxyRpcModule.java
@@ -176,7 +176,8 @@ public class SnmpProxyRpcModule extends AbstractXmlRpcModule<SnmpRequestDTO, Snm
         final CompletableFuture<SnmpValue[]> future = SnmpUtils.getAsync(request.getAgent(), oids);
         return future.thenApply(values -> {
             final List<SnmpResult> results = new ArrayList<>(oids.length);
-            for (int i = 0; i < oids.length; i++) {
+            int len = Math.min(oids.length, values.length);
+            for (int i = 0; i < len; i++) {
                 final SnmpResult result = new SnmpResult(oids[i], null, values[i]);
                 results.add(result);
             }
--- a/core/snmp/proxy-rpc-impl/src/main/java/org/opennms/netmgt/snmp/proxy/common/SnmpProxyRpcModule.java
+++ b/core/snmp/proxy-rpc-impl/src/main/java/org/opennms/netmgt/snmp/proxy/common/SnmpProxyRpcModule.java
@@ -47,11 +47,14 @@ import org.opennms.netmgt.snmp.ColumnTracker;
 import org.opennms.netmgt.snmp.SingleInstanceTracker;
 import org.opennms.netmgt.snmp.SnmpInstId;
 import org.opennms.netmgt.snmp.SnmpObjId;
+import org.opennms.netmgt.snmp.SnmpResponseException;
 import org.opennms.netmgt.snmp.SnmpResult;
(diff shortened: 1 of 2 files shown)


core/snmp/proxy-rpc-impl/src/main/java/org/opennms/netmgt/snmp/proxy/common/SnmpProxyRpcModule.java
```diff
@@ -176,7 +176,8 @@ private static final void addResult(SnmpResult result, String correlationId, Map
         final CompletableFuture<SnmpValue[]> future = SnmpUtils.getAsync(request.getAgent(), oids);
         return future.thenApply(values -> {
             final List<SnmpResult> results = new ArrayList<>(oids.length);
-            for (int i = 0; i < oids.length; i++) {
+            int len = Math.min(oids.length, values.length);
```
j-white, line 179: if `oids.length != values.length` then the oid/value mapping bellow will not be accurate

cgorantla: True, was discussing with @agalue. He proposed to compare OIDs coming from the result and map accordingly.  That seems better approach

cgorantla: scratch that, since we don't carry request PDU in SNMPValue, it may be good to throw exception in `SNMP4Strategy` when request and response size are not matching.

cgorantla: Since returning wrong response is worse than not having response, I reverted that.
Now logging whenever response size is not matching with request size and throwing custom exception.

core/snmp/proxy-rpc-impl/src/main/java/org/opennms/netmgt/snmp/proxy/common/SnmpProxyRpcModule.java
```diff
@@ -176,6 +181,10 @@ private static final void addResult(SnmpResult result, String correlationId, Map
         final CompletableFuture<SnmpValue[]> future = SnmpUtils.getAsync(request.getAgent(), oids);
         return future.thenApply(values -> {
             final List<SnmpResult> results = new ArrayList<>(oids.length);
+            if (oids.length != values.length) {
```
j-white, line 184: In what cases might this happen?

core/snmp/proxy-rpc-impl/src/main/java/org/opennms/netmgt/snmp/proxy/common/SnmpProxyRpcModule.java
```diff
@@ -176,6 +181,10 @@ private static final void addResult(SnmpResult result, String correlationId, Map
...
+                LOG.error("Response values size {} is not matching request oids size {}", values.length, oids.length);
+                throw new SnmpResponseException("Invalid SNMP Response, Number of Response elements are not matching number of Requested oids");
```
j-white, line 186: I think better behavior would be to return the results we did retrieve rather than none at all.

core/snmp/proxy-rpc-impl/src/main/java/org/opennms/netmgt/snmp/proxy/common/SnmpProxyRpcModule.java
cgorantla: I was thinking when values returned are less than number of oids requested.

cgorantla:
Closing in favor of #3159

PR #3075, opened 2020-07-21
NMS-6470: allow hashing of password when creating or updating users

RangerRick:
This PR allows optionally hashing passwords upon saving when creating or updating users in the ReSTv1 interface.  I made it an optional URL parameter to make it easy to stay backwards-compatible with any existing scripts that might be interacting with the user rest service; you have to opt-in by adding `hashPassword=true` to trigger hashing before storing.

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-6470

### DIFF
--- a/opennms-doc/guide-development/src/asciidoc/text/rest/users.adoc
+++ b/opennms-doc/guide-development/src/asciidoc/text/rest/users.adoc
@@ -22,6 +22,7 @@ You may need to touch the `$OPENNMS_HOME/etc/users.xml` file on the filesystem f
 | Parameter | Description
 | `/users`  | Add a user. If supplying a password it is assumed to be hashed or encrypted already, at least as of 1.12.5.
               To indicate that the supplied password uses the salted encryption algorithm rather than the older _MD5_ based algorithm, you need to pass an element named `passwordSalt` with text `true` after the password element (or key/value pairs if using _JSON_).
+              Note: You may add the query parameter `hashPassword=true` to tell OpenNMS you are passing an unencrypted password; it will hash and salt the password when it is saved.
 |===
 
 ===== PUTs (Modifying Data)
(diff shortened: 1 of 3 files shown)


dschlenk:
👍

agalue:
The following is the only way I found it works without this great patch, and it works, but the password is not hashed in `users.xml`:

```bash
curl -u admin:admin -H 'Content-Type: application/xml' -v -d '
<user>
  <user-id>agalue</user-id>
  <full-name>Alejandro</full-name>
  <user-comments>Do not delete</user-comments>
  <password>5EBE2294ECD0E0F08EAB7690D2A6EE69</password>
  <passwordSalt>false</passwordSalt>
  <role>ROLE_ADMIN</role>
</user>
' http://localhost:8980/opennms/rest/users
```

So, with the patch, is it correct to assume the following will work, and I'll get a hashed password in `users.xml`? Or, would this be a proper usage example?

```bash
curl -u admin:admin -H 'Content-Type: application/xml' -v -d '
<user>
  <user-id>agalue</user-id>
  <full-name>Alejandro</full-name>
  <user-comments>Do not delete</user-comments>
  <password>secret</password>
  <role>ROLE_ADMIN</role>
</user>
' "http://localhost:8980/opennms/rest/users?hashPassword=true"
```

RangerRick:
Correct. If you include `hashPassword=true`, it assumes your user object has an _unencrypted/plaintext_ password.

On a `POST` or `PUT`, with `hashPassword=true`, it will:

1. ignore the `passwordSalted` value on the user object
2. take the password from the `password` field on the object, and hash it (with salt enabled, `SHA-256`)
3. store that hashed password _back_ in the object's `password` field
4. set `passwordSalted=true`
5. save it

I just had a thought... If you do a `PUT` and you set `hashPassword=true`, but you are not passing `&password=foo` on the URL, should it hash the existing object's password or not?  Right now it processes the contents of `getPassword()` on the user object, no matter what.  I can see an argument for either, but I could see people wanting to always add `hashPassword=true` when they make API calls no matter which metadata they happen to be updating.

agalue:
Perfect, thanks!

About your last comment, I agree with assuming people would always add `hashPassword=true`.

I believe it would be good hashing the existing password when the `password` attribute is not passed to the URL. That could help to solve the problem I have right now: be able to hash the password after the user was created, without the need for changing the password programmatically.

RangerRick:
Except in that case, the password _might_ be an already-hashed MD5 password. We may have no way of knowing. It wouldn't "upgrade" the password, it would just make gibberish.

agalue:
And now you know why I always use `POST` and never tried `PUT`, even when I want to "update" something from an existing user.

So, as I use `POST`, I'm good with any option you decide for `PUT`.

RangerRick:
Yeah, I'm gonna make it only transform the object if you are passing a password.  That seems least likely to be destructive.

PR #3099, opened 2020-08-06
NMS-12841: fixed case in broken column names

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-12841


patrick-schweizer:
@christianpape don't we also need to adopt OnmsApplication? I believe it has Camel Case references...

PR #3103, opened 2020-08-07
NMS-12846: When hostname is not resolvable, use source address from syslog message

cgorantla:
### External References

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-12846


cgorantla:
Closing this as we think it is better to stick to default option instead of creating more configurable options.

PR #3073, opened 2020-07-20
NMS-12792: Add outages for remote poller status changes

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-12792

This PR adds outages for remote poller status changes. For this an additional column ‚perspective‘ was added to the outage's table specifying the location from which the service was polled. An null-valued perspective indicates that the service was polled from OpenNMS itself. Please test this PR extensive and thoroughly, since this is a big change that affects many spots (availability calculation, DB views, reports, DB charts, trend charts, PL/SQL functions, etc.) in OpenNMS.

For testing, it is useful to manually create some entries in the outage table of the database whose perspective refers to other locations.

## Testplan mitigating merge risk

Compare 26.1.2 vs. NMS-12792 for the following topics:

### Web UI

* [x] Compare results from RTCD calculating availability
* [x] Compare results from outage records
* [x] Compare Trend charts
* [ ] Compare availability Jasper Reports

### REST API

* [x] Compare results from RTCD calculating availability
* [x] Compare results from Availability REST API
* [x] Compare results from outage records

PR #3107, opened 2020-08-12
NMS-12842: Remote Poller: add remote outages to UI

patrick-schweizer:
This PR adds "Perspective" to outage and outage detail view.

Open TODOs:
* [x] get perspective from data base
* [x] verify that sorting works

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-12842

### DIFF
--- a/opennms-webapp/src/main/java/org/opennms/web/filter/EqualsFilterNullAware.java
+++ b/opennms-webapp/src/main/java/org/opennms/web/filter/EqualsFilterNullAware.java
@@ -1,0 +1,61 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2007-2014 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2014 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License,
+ * or (at your option) any later version.
+ *
+ * OpenNMS(R) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with OpenNMS(R).  If not, see:
+ *      http://www.gnu.org/licenses/
+ *
+ * For more information contact:
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.web.filter;
+
+import org.hibernate.criterion.Criterion;
+import org.hibernate.criterion.Restrictions;
+
+/** Allows null as value for equals operations. */
+public abstract class EqualsFilterNullAware extends OneArgFilter<String> {
+
+    public EqualsFilterNullAware(final String filterType, final SQLType<String> type, final String fieldName, final String propertyName, final String value){
+        super(filterType, type, fieldName, propertyName, toValueAllowNull(value));
+    }
+
+    private static String toValueAllowNull(String value) {
+        return "null".equals(value) ? null : value;
+    }
+
+    @Override
+    public Criterion getCriterion() {
+        if(getValue() == null) {
+            return Restrictions.isNull(getPropertyName());
+        }
+        return Restrictions.eq(getPropertyName(), getValue());
+    }
+
+    @Override
+    public String getSQLTemplate() {
+        if(getValue() == null) {
+            return " " + getSQLFieldName() + " IS NULL ";
+        }
+        return " " + getSQLFieldName() + " =  %s ";
+    }
+
+}
--- a/opennms-webapp/src/main/java/org/opennms/web/filter/EqualsFilterNullAware.java
+++ b/opennms-webapp/src/main/java/org/opennms/web/filter/EqualsFilterNullAware.java
@@ -1,8 +1,8 @@
 /*******************************************************************************
  * This file is part of OpenNMS(R).
  *
- * Copyright (C) 2007-2014 The OpenNMS Group, Inc.
- * OpenNMS(R) is Copyright (C) 1999-2014 The OpenNMS Group, Inc.
+ * Copyright (C) 2007-2020 The OpenNMS Group, Inc.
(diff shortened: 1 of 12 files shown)


opennms-webapp/src/main/java/org/opennms/web/outage/DaoWebOutageRepository.java
```diff
@@ -195,6 +195,7 @@ private Outage mapOnmsOutageToOutage(OnmsOutage onmsOutage) {
             outage.serviceName = onmsOutage.getMonitoredService() != null ? onmsOutage.getMonitoredService().getServiceName() : "";
             outage.suppressedBy = onmsOutage.getSuppressedBy();
             outage.suppressTime = onmsOutage.getSuppressTime();
+            outage.perspectiveLocation = "TODO Patrick";
```
patrick-schweizer, line 198: @christianpape I'm not sure from where do we get the perspective location?

christianpape: The PR https://github.com/OpenNMS/opennms/pull/3073 isn't merged to features/rpwm yet. We are still waiting for the final approval of @indigo423 . You can of course base your branch off jira/NMS-12792.

patrick-schweizer: fixed

patrick-schweizer:
Selection_381

Selection_380

christianpape:
> Selection_381
> 
> Selection_380
> 
> @christianpape is that how you imagined it? You can sort by Perspective. Or do you want (maybe additionally) a filter on top?

Nice work. But yes, I think we need an additional filter for perspective here. I'm pretty sure, that an user wants to filter for outages for a specific perspective.

fooker:
> > Selection_381
> > Selection_380
> > @christianpape is that how you imagined it? You can sort by Perspective. Or do you want (maybe additionally) a filter on top?
> 
> Nice work. But yes, I think we need an additional filter for perspective here. I'm pretty sure, that an user wants to filter for outages for a specific perspective.

This is right. In addition, we need a filter for the "empty" (read `null`) perspective to see only non-remote polled outages.

patrick-schweizer:
> This is right. In addition, we need a filter for the "empty" (read `null`) perspective to see only non-remote polled outages.

Hm I wonder how it will look like if we have outages from 10 different perspectives. The current UI design doesn't support that I think. Either we need to change the way we display filters or we just filter by "local" and "remote". For now I'll implement the latter but happy to discuss how to filter by all Perspectives.

patrick-schweizer:
voila: With filtering:

Selection_382

opennms-webapp/src/main/java/org/opennms/web/filter/EqualsFilterNullAware.java
```diff
@@ -0,0 +1,61 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2007-2014 The OpenNMS Group, Inc.
```
christianpape, line 4: Use 
  Copyright (C) 2002-2020 The OpenNMS Group, Inc.
  OpenNMS(R) is Copyright (C) 1999-2020 The OpenNMS Group, Inc.
for new files.

opennms-webapp/src/main/java/org/opennms/web/filter/NotEqualsFilterNullAware.java
```diff
@@ -0,0 +1,61 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2007-2014 The OpenNMS Group, Inc.
```
christianpape, line 4: Same here

opennms-webapp/src/main/java/org/opennms/web/filter/NotEqualsFilterNullAware.java
```diff
@@ -0,0 +1,61 @@
...
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.web.filter;
+
+import org.hibernate.criterion.Criterion;
+import org.hibernate.criterion.Restrictions;
+
+/** Allows null as a valid value for not equals filtering. */
+public abstract class NotEqualsFilterNullAware<T> extends OneArgFilter<String> {
```
christianpape, line 35: <T> isn'd used anywhere, right?

indigo423:
> voila: With filtering: I called the "null" perspectives <local> since they are not remote.
> 
> Selection_382

It seems like a "Perspective" is represented as "location + component" detected the outage which can be:

* [1] In location "Default" and the outage is detected from the Core System (OpenNMS itself)
* [2] In location "Default" and the outage is detected from a Minion
* [3] In location "other-location-than-default" and the outage is detected from a Minion

The solution for [1] with `<local>` seems like a hacky workaround. Here some naive brainstorm thoughts on this:

1. Keeping it empty. Is also misleading cause the OpenNMS system itself is in location "Default" and is not correct
2. `<local>`: Doesn't explain a lot to a user. We have to explain what makes it different to "Default"
3. "Default" with a component identifier

I would vote for a solution around 3. Which can be something like this:

suggestion

We have just FontAwesome 4.7.0 so we could use the home icon to indicate a remote location with a Minion and the Server icon to indicate the OpenNMS server itself. We can add a tooltip to the icon to explain the icon. That way we can always show. Instead of the home icon, a Kiwi would be nice but is only available in latest Fontawesome version :D

opennms-webapp/src/main/java/org/opennms/web/filter/EqualsFilterNullAware.java
patrick-schweizer: :+1: fixed

opennms-webapp/src/main/java/org/opennms/web/filter/NotEqualsFilterNullAware.java
patrick-schweizer: :+1: fixed

opennms-webapp/src/main/java/org/opennms/web/filter/NotEqualsFilterNullAware.java
patrick-schweizer: it is used by the NegativePerspectiveLocationFilter. I had to introduce this class to handle the NULL perspectives properly.

christianpape: Sorry, I meant the <T> isn't used at all, right? This can be omitted, right?

patrick-schweizer: true. Will remove...

patrick-schweizer:
@indigo423 ,
I don't think we can use the location name in case perspective == null since this would change the filter semantic.
I solved it now in the following way (close to your suggestion):
* perspective != null => server icon + perspective name
* perspective == null => home icon

Selection_384
Selection_385

PR #3108, opened 2020-08-12
NMS-12693: remove reliance on Getopt::Mixed

RangerRick:
This PR removes Getopt::Mixed in favor of Getopt::Long which has been a part of Perl5's official distribution for some time now.  This fixes installation on newer ubuntu/debian distros which finally dropped the `libgetopt-mixed-perl` package.


RangerRick:
Note that I changed the help to list `--destination` but the previous style of just `--dest` still works, `Getopt::Long` automatically supports shortened versions of the options.

PR #3112, opened 2020-08-13
NMS-12817: allow forcing use of SHA-256 passwords

RangerRick:
This PR makes it so you can configure OpenNMS to disallow writing old-style unsalted passwords.

JIRA: https://issues.opennms.org/browse/NMS-12817


RangerRick:
Hrm, I can't reproduce any failure if there are existing salted passwords in there.  Only difference is my build had been rebased against newest `foundation-2019` tho' I can't imagine that would change things.

indigo423 (approved):
Verified with the RPM artifacts and works also with my 26.2.0 users.xml. 👍

RangerRick:
thanks!

PR #3120, opened 2020-08-24
NMS-12865: Use 'native' to clarify 'not remote'

fooker:
* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12865

### DIFF
--- a/opennms-webapp/src/main/java/org/opennms/web/outage/OutageUtil.java
+++ b/opennms-webapp/src/main/java/org/opennms/web/outage/OutageUtil.java
@@ -448,9 +448,9 @@ public abstract class OutageUtil extends Object {
     public static String getPerspectiveLabel(final String perspective) {
         String html;
         if (perspective == null) {
-            html = "<i class=\"fa fa-home\" aria-hidden=\"true\"></i> ";
+            html = "<i>none</i>";
         } else {
-            html = "<i class=\"fa fa-server\" aria-hidden=\"true\"></i> " + WebSecurityUtils.sanitizeString(perspective);
+            html = WebSecurityUtils.sanitizeString(perspective);
         }
         return html;
     }
--- a/opennms-webapp/src/main/java/org/opennms/web/outage/filter/NegativePerspectiveLocationFilter.java
+++ b/opennms-webapp/src/main/java/org/opennms/web/outage/filter/NegativePerspectiveLocationFilter.java
@@ -30,6 +30,7 @@ package org.opennms.web.outage.filter;
 
 import java.util.Objects;
 
+import org.opennms.core.utils.WebSecurityUtils;
 import org.opennms.web.filter.NotEqualsFilterNullAware;
 import org.opennms.web.filter.SQLType;
 import org.opennms.web.outage.OutageUtil;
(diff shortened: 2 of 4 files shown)


christianpape (approved):
Good. Term #2 of thousand of terms to come...

patrick-schweizer:
I am not sure if "native" makes the wording more clear. I would argue the opposite of _remote_ is _local_. If we have remote polling we should also have local polling, no? I am not a native English speaker nor a writer so I might be off here. Maybe @Bonrob2 could help us sorting out the wording?

Bonrob2:
> I am not sure if "native" makes the wording more clear. I would argue the opposite of _remote_ is _local_. If we have remote polling we should also have local polling, no? I am not a native English speaker nor a writer so I might be off here. Maybe @Bonrob2 could help us sorting out the wording?

@patrick-schweizer, I think "local" would be better than "native". "Local" is more clearly the opposite of "remote". I also think that within a few years English will be replacing "native" with a different term anyway, similar to what we are currently doing to change "master" and "slave" terminology.

christianpape:
> I am not sure if "native" makes the wording more clear. I would argue the opposite of _remote_ is _local_. If we have remote polling we should also have local polling, no? I am not a native English speaker nor a writer so I might be off here. Maybe @Bonrob2 could help us sorting out the wording?




> > I am not sure if "native" makes the wording more clear. I would argue the opposite of _remote_ is _local_. If we have remote polling we should also have local polling, no? I am not a native English speaker nor a writer so I might be off here. Maybe @Bonrob2 could help us sorting out the wording?
> 
> @patrick-schweizer, I think "local" would be better than "native". "Local" is more clearly the opposite of "remote". I also think that within a few years English will be replacing "native" with a different term anyway, similar to what we are currently doing to change "master" and "slave" terminology.

We are searching for a term, that represents a remotely noticed outage. Just to clarify, we have three cases:
- first, the outage of a service was noticed from the OpenNMS instance itself. In this case we have no perspective location. Maybe we can display _Default_ or _OpenNMS_ here.
- second, the outage of a service was noticed from a minion in the associated remote location of the node. In this case the minion itself polled a node within its same locality but of course from the OpenNMS' point of view this minion is also a remote entity. So, we need a term for this case. Maybe we can use something like _minion-local_ here.
- third, the outage of a service was noticed by a minion from a minion in another location (not the node's location). This case is the new _Remote Poller_ case, where a minion polls a node that is associated with another location, so here the _perspective_ can be displayed.

fooker:
I agree. Wording is super hard on this. Let me try to clarify some more:

The terms `local` and `remote` mean something different in this context. Usually we talk about `local` and `default location` when we talk about the OpenNMS-Horizon instance. In contrast `remote` means something is executed on a minion.

The default polling mechanism currently works by having nodes assigned to locations. It uses one of the `monitoring systems` assigned to the nodes location to monitor that node.

The term `Remote Polling` is inherited from a time before minion. Preferably the is would be called `Perspective Monitoring` now.
What it does is selecting locations (maybe) different from the location assigned to that node and use one `monitoring system` at these `perspective location` to monitor the node.

This all results in two datums, both referencing a location. The first location is the one assigned to the node - usually seen as the location in which the node is defined (or in which it exists). The second one is the perspective from which the node was inspected.

For the second datum, the term `perspective location` fits quite well, i think. But outages can also have the case where they are created by the default poller and the field remains empty. This means, that the node node is polled by a monitoring system in the location it is assigned to. My idea was to use something that represents the fact that the location is "the home" of that node (but not use the term `home` as it can be misinterpreted with the `default location`). Maybe something like `resident location` fits better than the word `native location`.

> We are searching for a term, that represents a remotely noticed outage. Just to clarify, we have three cases:

I think this is still over-complicated: It just boils down to two cases: `perspective location == null` which means "the nodes own location" or `perspsective != null` which means polled form a different perspective.

I'm happy to discus this further... Bikeshedding ahead!

patrick-schweizer:
About the naming: Some food for thought:

> It just boils down to two cases: perspective location == null which means "the nodes own location" or perspsective != null which means polled form a different perspective.

We are talking about perspective: That would be from where a node is viewed, right ?: How about:
_outside perspective_ vs. _inside perspective_ ?

fooker:
I've reworked this by making the filter description more verbose/smart. The now use the phrasing `polling perspective is X`, `polling perspective is not X`. `from non-perspective polling` and `from perspective polling`.
In addition, the empty perspective is indicated as an italic `none` in the lists.

christianpape (approved):
LGTM

Bonrob2 (approved):
LGTM

PR #3122, opened 2020-08-24
NMS-12866: Show only native outages in detail pages

fooker:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12866


christianpape (approved):
OK

PR #3127, opened 2020-08-25
NMS-12782: Add SHA2 authentication support for SnmpV3

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12782


cgorantla:
Verified locally with Traps and SNMP get. Opened an issue for improving documentation on SNMP config https://issues.opennms.org/browse/NMS-12873

PR #3128, opened 2020-08-25
NMS-12870: Use generatedId only for Standard Resource graphs

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12870


cgorantla:
My dev environment is somewhat broken to run `TestContainers` because of fedora32 and this seems non-trivial to resolve quickly. I will add the IT later.  Merging this for now.

PR #3129, opened 2020-08-26
NMS-12860: Added some documentation for the remote poller

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-12860

Maybe we can add later ome documentation/screenshots about the application topology that is currently still in development.


opennms-doc/guide-admin/src/asciidoc/text/application-perspective-monitoring/application-perspective-monitoring.adoc
```diff
@@ -2,36 +2,36 @@
...
 After configuring the application, the minions at the remote perspective locations start to monitor the services associated with this application.
-The next screenshot show an HTTP outage noticed from all perspective locations and the _{opennms-product-name}_ instance itself.
+The next screenshot show an HTTP outage noticed from all perspective locations and the _{opennms-product-name}_ poller daemon itself.
 
 .Remote outages
-image::remote-poller/remote-outages.png[]
+image::application-perspective-monitoring/remote-outages.png[]
 
 The column _Perspective_ shows the remote location from which a minion has detected this outage.
 If no _Perspective_ is shown, the outage was detected by the normal process: the _{opennms-product-name}_ instance detected it in the _Default_ location or a minion detected it in the corresponding node's location.
```
Bonrob2, line 37:
```suggestion
An empty *Perspective* column indicates that the normal process detected the outage: either the _{opennms-product-name}_ instance detected it in the default location or a Minion detected it in the corresponding node's location.
```
Please verify that saying the column is empty is correct. Would it be blank, or a null value?

christianpape: We don't know it yet. Maybe we will use an icon or term when no perspective is given. I'll change this accordingly when we have a final solution.

PR #3133, opened 2020-08-27
NMS-12878: Make backend more resilient

fooker:
* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12878


features/remotepollerng/daemon/src/main/java/org/opennms/netmgt/remotepollerng/PerspectiveServiceTracker.java
```diff
@@ -0,0 +1,259 @@
...
+
+    private final Timer timer = new Timer("PerspectiveServiceTracker-Timer");
+
+    @Autowired
+    public PerspectiveServiceTracker(final SessionUtils sessionUtils,
+                                     final ApplicationDao applicationDao) {
+        this.sessionUtils = Objects.requireNonNull(sessionUtils);
+        this.applicationDao = Objects.requireNonNull(applicationDao);
+
+        final long timerIntevalMs = Math.min(REFRESH_RATE_LIMIT_MS, TimeUnit.SECONDS.toMillis(5));
```
j-white, line 143: Do we intend to use `min` (vs `max` here?). The default value of `REFRESH_RATE_LIMIT_MS` is 30 seconds.

features/remotepollerng/daemon/src/test/java/org/opennms/netmgt/remotepollerng/RemotePollerdIT.java
```diff
@@ -215,6 +215,11 @@ public void setUp() throws Exception {
         locationAwarePollerClient.setRpcTargetHelper(new RpcTargetHelper());
         locationAwarePollerClient.afterPropertiesSet();
 
+        System.setProperty("org.opennms.netmgt.remotepollerng.trackerRefreshRateLimit", "5");
```
j-white, line 218: I'd move the `"org.opennms.netmgt.remotepollerng.trackerRefreshRateLimit"` part to a constant and refer to that instead.

features/remotepollerng/daemon/src/main/java/org/opennms/netmgt/remotepollerng/PerspectiveServiceTracker.java
```diff
@@ -0,0 +1,259 @@
...
+            EventConstants.APPLICATION_DELETED_EVENT_UEI,
+    })
+    public void handleEvent(final IEvent event) {
+        update(true);
+    }
+
+    private void update(final boolean dirty) {
+        synchronized (PerspectiveServiceTracker.this.sessions) {
+            try {
+                PerspectiveServiceTracker.this.sessions.forEach(session -> session.update(dirty));
```
j-white, line 187: Every session will create its own transaction and repeat the same DB query. This is not a big problem since there's likely only 1 session, but would be nicer to do that outside the loop.

features/remotepollerng/daemon/src/main/java/org/opennms/netmgt/remotepollerng/PerspectiveServiceTracker.java
christianpape, line 143: typo: timerInte<r>valMs

features/remotepollerng/daemon/src/main/java/org/opennms/netmgt/remotepollerng/PerspectiveServiceTracker.java
```diff
@@ -0,0 +1,259 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+
+import com.google.common.collect.Sets;
+
+@EventListener(name = "PerspectiveServiceTracker")
+public class PerspectiveServiceTracker implements DisposableBean {
+
+    private static final Logger LOG = LoggerFactory.getLogger(PerspectiveServiceTracker.class);
+
+    private static final long REFRESH_RATE_LIMIT_MS = SystemProperties.getLong("org.opennms.netmgt.remotepollerng.trackerRefreshRateLimit", TimeUnit.SECONDS.toMillis(30));
```
christianpape, line 62: Please also add an entry with description and default value in opennms-base-assembly/src/main/filtered/etc/opennms.properties.

christianpape (changes requested):
Beautiful and simple solution. Had only two tiny remarks.

features/remotepollerng/daemon/src/main/java/org/opennms/netmgt/remotepollerng/PerspectiveServiceTracker.java
fooker: I think so. This is the polling interval denoting how often it checks for outstanding refresh (those who are not executed immediately due to the rate limit). This should usually be smaller than the rate limit.

features/remotepollerng/daemon/src/test/java/org/opennms/netmgt/remotepollerng/RemotePollerdIT.java
fooker: Ay

features/remotepollerng/daemon/src/main/java/org/opennms/netmgt/remotepollerng/PerspectiveServiceTracker.java
fooker: But the session is only created if an update is required for that session. By moving it out of the session, it will be created every time. Most of the time, calls to the `update` method will result in an early exit.

features/remotepollerng/daemon/src/main/java/org/opennms/netmgt/remotepollerng/PerspectiveServiceTracker.java
christianpape: Ah yes, got it. Every 5ms the timer checks if an update is necessary and updates if the last one is older than REFRESH_RATE_LIMIT_MS, right? This way events are handled faster than waiting for REFRESH_RATE_LIMIT_MS.

christianpape (approved):
Slim and effective and does exactly what it should. Very nice.

features/remotepollerng/daemon/src/main/java/org/opennms/netmgt/remotepollerng/PerspectiveServiceTracker.java
fooker: Exactly (beside the check is every 5s not every 5ms)

christianpape: Right.

PR #3137, opened 2020-08-31
NMS-12881: Clear outages after unscheduling services

fooker:
* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12881


christianpape (approved):
Looks good.

PR #3139, opened 2020-09-02
NMS-12866: Only show non-perspective outages in recent outages

fooker:
* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12866

This filters out the non-perspective outages from the recent outages view and added some minor improvements/fixes to the outages lists.

### DIFF
--- a/opennms-webapp/src/main/java/org/opennms/web/controller/outage/InterfaceOutagesController.java
+++ b/opennms-webapp/src/main/java/org/opennms/web/controller/outage/InterfaceOutagesController.java
@@ -77,8 +77,9 @@ public class InterfaceOutagesController extends AbstractController implements In
             filters.add(new InterfaceFilter(ipAddr));
             filters.add(new NodeFilter(nodeId, getServletContext()));
             filters.add(new RecentOutagesFilter());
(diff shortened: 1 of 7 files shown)


christianpape (approved):
Nice.

indigo423:
Tested the feature and is much better now. The only thing I've noticed, we derrive from the normal and introduce "Grey" as UP. This is it consistent with the rest of the UI as we show UP things in Green :)

Screenshot 2020-09-03 at 14 26 09

For others, here some screenshots how it is shown on Interface and Node level

Screenshot 2020-09-03 at 14 30 11

Screenshot 2020-09-03 at 14 27 08

Screenshot 2020-09-03 at 14 30 07

indigo423 (changes requested):
enjoy-the-little-things

fooker:
> Tested the feature and is much better now. The only thing I've noticed, we derrive from the normal and introduce "Grey" as UP. This is it consistent with the rest of the UI as we show UP things in Green :)

Down=Red and Up=Grey mimics the behavior from the Recent Outages list on the same page.

indigo423 (commented):
Regarding the color thing, just a suggestion and not a blocker for me.

PR #3141, opened 2020-09-03
NMS-12886: Application status page

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-12886


indigo423:
The only thing I noticed during testing, the overall availability super weird at the top. Instead of letting it floating away right aligned. Suggestion, put before the application label?

Screenshot 2020-09-03 at 15 57 22

indigo423:
The availability calculation looks a little bit weird, when I take the service back up, the service has availabilty around 98% and the application shows 100%.

Screenshot 2020-09-03 at 16 14 18
Screenshot 2020-09-03 at 16 14 23

christianpape:
> The availability calculation looks a little bit weird, when I take the service back up, the service has an availability of 98,248% and in the application status availability, it shows 100%.
> 
> The percent sign is missing in the overall application availability :)
> 
> Screenshot 2020-09-03 at 16 14 18
> Screenshot 2020-09-03 at 16 14 23

Please give it another try...

indigo423:
Ok I got it the availability calculation logic is done in the JSPs and not covered by tests - Just the fun parts :D

christianpape:
> Ok, I got it the availability calculation logic in not covered by tests - Just the fun parts :D

No, that's not right. The calculation is covered by tests but the API is flexible concerning the timeframe of the availability calculation. So, you can compute the availability for the last hour, the last 24 hours or whatever you like. The original RP calculated the availability over the last 24h. So, I used the same timeframe in the JSP-page, but I forgot to multiply the value I subtract from the end timestamp by 1000. So, the calculation covered only the last 86,4 seconds and not 24h.

indigo423:
Percentage-wise fixed. The formatting is now different weird :) The suggestion I was thinking of was more like this:

Screenshot 2020-09-03 at 19 12 32

In a way, the percentage value is not floating off the screen on the right.

indigo423:
It's getting close, if you like the big availability numbers better, so be it :)

Screenshot 2020-09-04 at 09 59 55

opennms-dao/src/main/java/org/opennms/netmgt/dao/support/ApplicationStatusUtil.java
```diff
@@ -0,0 +1,201 @@
...
+import org.opennms.netmgt.model.OnmsOutage;
+import org.opennms.netmgt.model.monitoringLocations.OnmsMonitoringLocation;
+import org.opennms.netmgt.model.perspectivepolling.ApplicationServiceStatus;
+import org.opennms.netmgt.model.perspectivepolling.ApplicationStatus;
+import org.opennms.netmgt.model.perspectivepolling.Location;
+
+import com.google.common.base.MoreObjects;
+
+public class ApplicationStatusUtil {
+    static class DowntimeInterval {
```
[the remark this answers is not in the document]
christianpape: The Java Duration class models an amount of time. This does not include a start or end timestamp of the duration, just the duration itself. I query the start and end timestamps of the associated outages and have to merge them together. So, I need the start and end timestamps for this, because the outages of different services may overlap.

opennms-dao/src/main/java/org/opennms/netmgt/dao/support/ApplicationStatusUtil.java
```diff
@@ -0,0 +1,201 @@
...
+        public long start, end;
+
+        public DowntimeInterval(final OnmsOutage onmsOutage, long start, long end) {
```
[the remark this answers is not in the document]
christianpape: see above

opennms-dao/src/main/java/org/opennms/netmgt/dao/support/ApplicationStatusUtil.java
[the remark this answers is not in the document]
patrick-schweizer: you are right - makes sense!

opennms-webapp/src/main/java/org/opennms/web/element/NetworkElementFactory.java
```diff
@@ -1009,4 +1020,50 @@ public boolean nodeExistsInRequisition(final String foreignSource, final String
...
+                onmsApplication,
+                m_outageDao.getStatusChangesForApplicationIdBetween(new Date(start), new Date(end), onmsApplication.getId()),
+                start,
+                end
+        );
+    }
+
+    @Override
+    public Map<OnmsMonitoredService, Map<String, Double>> getApplicationServiceStatus(final OnmsApplication onmsApplication, final long start, final long end) {
+        final Map<String, OnmsMonitoredService> serviceMap = new TreeMap<>();
```
fooker, line 1041: The key of the map is never used. Can `onmsApplication.getMonitoredServices()` be used directly in the loop below?

opennms-webapp/src/main/java/org/opennms/web/element/NetworkElementFactory.java
```diff
@@ -1009,4 +1020,50 @@ public boolean nodeExistsInRequisition(final String foreignSource, final String
...
+        final Map<OnmsMonitoredService, Map<String, Double>> status = new TreeMap<>();
+
+        for(final Map.Entry<String, OnmsMonitoredService> serviceMapEntry : serviceMap.entrySet()) {
+            final Map<String, Double> serviceStatus = new TreeMap<>();
+
+            ApplicationServiceStatus applicationServiceStatus = ApplicationStatusUtil.buildApplicationServiceStatus(
+                    m_monSvcDao,
+                    onmsApplication,
+                    serviceMapEntry.getValue().getId(),
+                    m_outageDao.getStatusChangesForApplicationIdBetween(new Date(start), new Date(end), onmsApplication.getId()),
```
fooker, line 1055: This call can be moved out of the loop.

fooker (changes requested):

opennms-webapp/src/main/java/org/opennms/web/element/NetworkElementFactory.java
christianpape: done.

opennms-webapp/src/main/java/org/opennms/web/element/NetworkElementFactory.java
christianpape: done.

fooker (approved):

PR #3148, opened 2020-09-09
NMS-12896: Register JmxReporter only once

fooker:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12896

### DIFF
--- a/core/ipc/sink/camel/server/src/main/java/org/opennms/core/ipc/sink/camel/server/CamelMessageConsumerManager.java
+++ b/core/ipc/sink/camel/server/src/main/java/org/opennms/core/ipc/sink/camel/server/CamelMessageConsumerManager.java
@@ -28,6 +28,8 @@
 
 package org.opennms.core.ipc.sink.camel.server;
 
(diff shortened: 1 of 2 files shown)


core/ipc/sink/camel/server/src/main/java/org/opennms/core/ipc/sink/camel/server/CamelMessageConsumerManager.java
```diff
@@ -69,6 +72,8 @@
     @Autowired
     private Identity identity;
 
+    private JmxReporter jmxReporter = null;
```
j-white, line 75: Do we need to do something similar for Kafka IPC?

cgorantla (commented):
Looks good to me.

core/ipc/sink/camel/server/src/main/java/org/opennms/core/ipc/sink/camel/server/CamelMessageConsumerManager.java
cgorantla: KafkaMessageConsumer seems to have only one instance of JmxReporter

PR #3150, opened 2020-09-10
NMS-12794: Add support for metadata for threshold fields

cgorantla:
Support metadata in Threshold fields like `value` `rearm` `trigger` in Simple-DS and Expression threshold configs.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12794

### DIFF
--- a/core/lib/src/main/java/org/opennms/core/utils/StringUtils.java
+++ b/core/lib/src/main/java/org/opennms/core/utils/StringUtils.java
@@ -53,6 +53,8 @@ import javax.xml.transform.stream.StreamSource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Strings;
+
 public abstract class StringUtils {
 
     private static final Logger LOG = LoggerFactory.getLogger(StringUtils.class);
@@ -455,8 +457,22 @@ public abstract class StringUtils {
     }
 
     public static Integer parseInt(String value, Integer defaultValue) {
+        if(Strings.isNullOrEmpty(value)) {
+            return defaultValue;
+        }
         try {
             return Integer.parseInt(value);
+        } catch (NumberFormatException e) {
+            return defaultValue;
+        }
+    }
+
+    public static Double parseDouble(String value, Double defaultValue) {
+        if(Strings.isNullOrEmpty(value)) {
+            return defaultValue;
+        }
+        try {
+            return Double.parseDouble(value);
         } catch (NumberFormatException e) {
             return defaultValue;
         }
--- a/features/collection/thresholding/impl/src/main/java/org/opennms/netmgt/threshd/AbstractThresholdEvaluatorState.java
+++ b/features/collection/thresholding/impl/src/main/java/org/opennms/netmgt/threshd/AbstractThresholdEvaluatorState.java
@@ -285,53 +285,53 @@ public abstract class AbstractThresholdEvaluatorState<T extends AbstractThreshol
     }
 
     @Override
-    public ValueStatus evaluate(ExpressionThresholdValue valueSupplier, Long sequenceNumber)
+    public ValueStatus evaluate(ExpressionThresholdValueSupplier valueSupplier, Long sequenceNumber)
             throws ThresholdExpressionException {
-        ExpressionConfigWrapper.ExpressionValue expressionValue = getValueForExpressionThreshold(valueSupplier);
-        Status status = evaluate(expressionValue.value, expressionValue.getThresholdValues(), sequenceNumber);
+        ExpressionConfigWrapper.ExpressionThresholdValues expressionThresholdValues = getValueForExpressionThreshold(valueSupplier);
+        Status status = evaluate(expressionThresholdValues.value, expressionThresholdValues.getThresholdValues(), sequenceNumber);
 
-        return new ValueStatus(expressionValue.value, status, expressionValue.getThresholdValues());
+        return new ValueStatus(expressionThresholdValues.value, status, expressionThresholdValues.getThresholdValues());
     }
 
     @Override
-    public ValueStatus evaluate(ThresholdValuesConsumer thresholdValuesConsumer, Long sequenceNumber)
+    public ValueStatus evaluate(ThresholdValuesSupplier thresholdValuesSupplier, Long sequenceNumber)
             throws ThresholdExpressionException {
-        ThresholdValues thresholdValues = getThresholdValues(thresholdValuesConsumer);
+        ThresholdValues thresholdValues = getThresholdValues(thresholdValuesSupplier);
         Status status = evaluate(thresholdValues.getDsValue(), thresholdValues, sequenceNumber);
         return new ValueStatus(thresholdValues.getDsValue(), status, thresholdValues);
     }
 
-    private ThresholdValues getThresholdValues(ThresholdValuesConsumer thresholdValuesConsumer) {
+    private ThresholdValues getThresholdValues(ThresholdValuesSupplier thresholdValuesSupplier) {
 
         if (!state.isCached()) {
-            return thresholdValuesConsumer.get(thresholdValues -> {
-                state.setThresholdValues(thresholdValues);
-                state.setCached(true);
-            });
+            ThresholdValues thresholdValues = thresholdValuesSupplier.get();
+            state.setThresholdValues(thresholdValues);
+            state.setCached(true);
+            return thresholdValues;
         } else {
-            Double dsvalue = thresholdValuesConsumer.getDsValue();
+            Double dsvalue = thresholdValuesSupplier.getDsValue();
             ThresholdValues thresholdValues = state.getThresholdValues();
             thresholdValues.setDsValue(dsvalue);
             return thresholdValues;
         }
     }
 
-    private ExpressionConfigWrapper.ExpressionValue getValueForExpressionThreshold(ExpressionThresholdValue valueSupplier)
+    private ExpressionConfigWrapper.ExpressionThresholdValues getValueForExpressionThreshold(ExpressionThresholdValueSupplier valueSupplier)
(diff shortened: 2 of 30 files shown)


cgorantla:
Still need some UI changes and more testing. Opening Draft PR for feedback.

features/collection/thresholding/impl/src/main/java/org/opennms/netmgt/threshd/ThresholdConfigWrapper.java
```diff
@@ -83,4 +83,5 @@ public double evaluate(Map<String, Double> values) {
     public void accept(ThresholdDefVisitor thresholdDefVisitor) {
         thresholdDefVisitor.visit(this);
     }
+
```
mattixtech, line 86: Looks like we don't need to churn this file.

features/collection/thresholding/impl/src/main/java/org/opennms/netmgt/threshd/ThresholdEvaluatorHighLow.java
```diff
@@ -201,32 +211,53 @@ public Status evaluateAfterFetch(double dsValue) {
             return Status.NO_CHANGE;
         }
 
-        protected boolean isThresholdExceeded(double dsValue) {
+        protected boolean isThresholdExceeded(double dsValue, Double value) {
```
mattixtech, line 214: Since this method and isRearmExceeded will NPE if given nulls can we update the method signature to take a primitive. That way their preconditions will be enforced.

mattixtech (changes requested):
Its been a while since I've looked at this code so I didn't dive too deep and I also didn't run any of the code. If you need a more thorough review let me know.

I didn't notice anything that looked like much of a functional problem, just recommending some changes for readability and safety.

features/collection/thresholding/impl/src/main/java/org/opennms/netmgt/threshd/AbstractThresholdEvaluatorState.java
```diff
@@ -128,6 +128,8 @@ public Long load(String key) {
 
     static abstract class AbstractState implements Serializable {
         String interpolatedExpression = null;
+        boolean cached = false;
```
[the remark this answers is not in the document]
cgorantla: Yeah, this is needed since `thresholdValues` can exist before and after reload.  
Doesn't state itself get `reset` when reloading ? Looks like it's not. When merging thresholds, looks like it retains previous ThresholdStates and just loads the config again.  There are several tests with  `ThresholdingVisitor`  `reload` that mimics this  behavior.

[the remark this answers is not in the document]
mattixtech: Not sure I follow, looks like very writer of `thresholdValues` also sets the cached field. Also upon deserializing an object with `thresholdValues` set wouldn't `cached` also be set in that case?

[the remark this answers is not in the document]
cgorantla: `clearStateBeforePersist` unsets the cache which gets called in `ThresholdingVisitor` `reload`.  If you are saying that `thresholdValues` can be set as null or `Optional.empty` in `clearStateBeforePersist` that's an option but I chose this as this is common to both `interpolatedExpression` and `thresholdValues`.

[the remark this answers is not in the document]
mattixtech: Ok, I missed the part where it was getting cleared in `clearStateBeforePersist`, I was thinking it was always being set in a tandem with the value. No worries, can leave as is.

cgorantla:
Modified Editing Thresholds UI looks like this

Screenshot from 2020-09-15 15-28-32
Screenshot from 2020-09-15 15-29-07

cgorantla:
Edit Group UI looks like this when using metadata in threshold values. Haven't changed this.
Screenshot from 2020-09-15 15-33-27

cgorantla:
Verified thresholds using XMLCollector.  Modifying thresholds seems to work fine.

j-white:
What's the expected behavior someone enters a value that is neither a number, nor a valid meta-data expression in the UI?

cgorantla:
> What's the expected behavior someone enters a value that is neither a number, nor a valid meta-data expression in the UI?

Yeah, there is no UI validation currently. Earlier webapp used to throw exception if it's not Double, now it accepts but resulting value could be null and no thresholds  are triggered.

core/lib/src/main/java/org/opennms/core/utils/StringUtils.java
```diff
@@ -455,11 +457,25 @@ public static String getHumanReadableByteCount(long bytes, boolean si) {
     }
 
     public static Integer parseInt(String value, Integer defaultValue) {
+        if(Strings.isNullOrEmpty(value)) {
```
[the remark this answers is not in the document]
mattixtech: Oh yeah, you are right, `parseDouble` only handles the empty case and not the null. Weird inconsistency in the stdlib!

features/collection/thresholding/impl/src/main/java/org/opennms/netmgt/threshd/ThresholdValuesConsumer.java
```diff
@@ -0,0 +1,38 @@
...
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.netmgt.threshd;
+
+import java.util.function.Consumer;
+
+public interface ThresholdValuesConsumer {
+
+    ThresholdEvaluatorState.ThresholdValues get(Consumer<ThresholdEvaluatorState.ThresholdValues> thresholdsConsumer);
```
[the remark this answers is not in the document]
mattixtech: 👍

features/collection/thresholding/impl/src/main/java/org/opennms/netmgt/threshd/ThresholdEvaluatorHighLow.java
mattixtech: Doesn't look like this changed, anyways not a big deal...

mattixtech (approved):

features/collection/thresholding/impl/src/main/java/org/opennms/netmgt/threshd/ThresholdEvaluatorHighLow.java
cgorantla: Oh, yeah. method with only one argument is only used in tests so added `@VisibleForTesting` annotation and other will always pre-check for null so this seems safe.

cgorantla:
@Bonrob2  Updated docs to indicate that metadata is supported in  Value, Rearm, Trigger fields. Let me know if we need to expand this

features/collection/thresholding/impl/src/main/java/org/opennms/netmgt/threshd/ThresholdEvaluatorHighLow.java
mattixtech: Not that it matters at this point, but the reason I originally commented on this is because the method would assert its preconditions by using a primitive `double` in the signature rather than implicitly only being valid with a non-null `Double`.

If we had changed the other code to return `OptionalDouble` rather than a nullable `Double` we would have either `empty` or a primitive `double` to pass in and we would satisfy the preconditions.

No need to action this, just explaining my train of thought.

j-white:
> > What's the expected behavior someone enters a value that is neither a number, nor a valid meta-data expression in the UI?
> 
> Yeah, there is no UI validation currently. Earlier webapp used to throw exception if it's not Double, now it accepts but resulting value could be null and no thresholds are triggered.

Thresholds are hard to debug, so we should do what we can do make sure the input is valid. Can we add input validation to the forms and to the XSD? We can test the input with a regex.

Bonrob2:
> Modified Editing Thresholds UI looks like this
> 
> Screenshot from 2020-09-15 15-28-32
> Screenshot from 2020-09-15 15-29-07

Please change "Rearm" to "Re-arm" for consistency.

opennms-doc/guide-admin/src/asciidoc/text/thresholds/thresholding.adoc
```diff
@@ -173,11 +173,10 @@ During evaluation of an expression, the following scopes are available:
 * Interface metadata
 * Service metadata
 
+Metadata is also supported in Value, Rearm, Trigger fields for Single-DS and expression-based thresholds.
```
Bonrob2, line 176:
```suggestion
Metadata is also supported in Value, Re-arm, and Trigger fields for Single-DS and expression-based thresholds.
```

opennms-doc/guide-admin/src/asciidoc/text/thresholds/thresholding.adoc
```diff
@@ -173,11 +173,10 @@ During evaluation of an expression, the following scopes are available:
...
+
 For more information on metadata and how to define it, see link:metadata[Metadata].
```
Bonrob2, line 178: The metadata link is broken.

Bonrob2 (changes requested):
Just a typo and broken link, otherwise looks good!

opennms-doc/guide-admin/src/asciidoc/text/thresholds/thresholding.adoc
cgorantla: Guess I updated it right

Bonrob2 (approved):
LGTM!

cgorantla:
Will address validation issues on https://issues.opennms.org/browse/NMS-12689.

PR #3158, opened 2020-09-16
NMS-10554: Added foreignId to the event definition

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-10554

Small correction to the event definition to include also the foreignId.


fooker (approved):
Astonishing

PR #3159, opened 2020-09-16
NMS-12818: Handle Snmp response when number of responses are less than requested oids.

cgorantla:
Handle the case where number of responses are less than requested oids
by filling the missing responses as SNMP_NULL

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12818


j-white (approved):
LGTM

PR #3161, opened 2020-09-17
NMS-12814: Interfaces incorrectly marked as having flows resulting in no data via Helm

patrick-schweizer:
Jira https://issues.opennms.org/browse/NMS-12814

### DIFF
--- a/features/flows/itests/src/test/java/org/opennms/netmgt/flows/elastic/MarkerCacheIT.java
+++ b/features/flows/itests/src/test/java/org/opennms/netmgt/flows/elastic/MarkerCacheIT.java
@@ -32,14 +32,18 @@ import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
 import static com.github.tomakehurst.wiremock.client.WireMock.post;
 import static com.github.tomakehurst.wiremock.client.WireMock.stubFor;
 import static org.hamcrest.Matchers.contains;
-import static org.hamcrest.Matchers.containsInAnyOrder;
 import static org.hamcrest.Matchers.empty;
 import static org.hamcrest.Matchers.hasProperty;
 import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertEquals;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
+import java.util.Arrays;
+import java.util.List;
 import java.util.Optional;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import org.junit.Assert;
 import org.junit.Before;
@@ -178,9 +182,8 @@ public class MarkerCacheIT {
             elasticFlowRepository.persist(Lists.newArrayList(getMockFlow(Flow.Direction.INGRESS)), getMockFlowSource());
 
             Assert.assertThat(nodeDao.findAllHavingFlows(), contains(hasProperty("id", is(1))));
-            Assert.assertThat(snmpInterfaceDao.findAllHavingFlows(1), containsInAnyOrder(
-                    hasProperty("ifIndex", is(2)),
-                    hasProperty("ifIndex", is(3))));
+            Assert.assertThat(snmpInterfaceDao.findAllHavingFlows(1), contains(
+                    hasProperty("ifIndex", is(2))));
         }
     }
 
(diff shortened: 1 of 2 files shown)


features/flows/itests/src/test/java/org/opennms/netmgt/flows/elastic/MarkerCacheIT.java
```diff
@@ -221,11 +236,81 @@ public void testNMS12740() throws Exception {
...
+            checkInterfaces(1, "192.168.1.1");
+            checkInterfaces(2);
+
+            elasticFlowRepository.persist(Lists.newArrayList(getMockFlow(Flow.Direction.INGRESS)), getMockFlowSource());
+            checkInterfaces(1, "192.168.1.1", "192.168.1.1");
+            checkInterfaces(2);
+
+            elasticFlowRepository.persist(Lists.newArrayList(getMockFlow(Flow.Direction.INGRESS)), getMockFlowSource());
+            checkInterfaces(1, "192.168.1.1", "192.168.1.1");
+            checkInterfaces(2);
```
patrick-schweizer, line 293: @fooker , @christianpape, the test will succeed with this implementation. But It is not clear to me why the system behaves like it does. Why is it green when I implement it that way?

I assume the checkInterfaces method is implemented wrong. I don't have enough domain knowledge to understand what we need to test for...
Maybe you can help me out here... Thanks!

features/flows/itests/src/test/java/org/opennms/netmgt/flows/elastic/MarkerCacheIT.java
```diff
@@ -221,11 +232,97 @@ public void testNMS12740() throws Exception {
...
+
+            Integer ingress = 2;
+            Integer egress = 3;
+
+            // no flows persisted -> we shouldn't have any interfaces
+            expectAllInterfaces();
+            expectIngressInterfaces();
+            expectEgressInterfaces();
+
+            // persist egress flow
```
fooker, line 283: Comment / code mismatch (ingress vs egress)

features/flows/itests/src/test/java/org/opennms/netmgt/flows/elastic/MarkerCacheIT.java
```diff
@@ -221,11 +232,97 @@ public void testNMS12740() throws Exception {
...
+            elasticFlowRepository.persist(Lists.newArrayList(getMockFlow(Flow.Direction.INGRESS)), getMockFlowSource());
+            expectAllInterfaces(ingress);
+            expectIngressInterfaces(ingress);
+            expectEgressInterfaces();
+
+            // persist ingress flow
```
fooker, line 289: And the other way around

features/flows/itests/src/test/java/org/opennms/netmgt/flows/elastic/MarkerCacheIT.java
```diff
@@ -108,6 +116,9 @@
     @Autowired
     private InterfaceToNodeCache interfaceToNodeCache;
 
+    @Autowired
+    private TransactionTemplate transactionTemplate;
```
fooker, line 120: Use `SessionUtils` instead. It saves you from a lot of boilerplate code

features/flows/itests/src/test/java/org/opennms/netmgt/flows/elastic/MarkerCacheIT.java
```diff
@@ -221,11 +232,97 @@ public void testNMS12740() throws Exception {
...
+        factory.setHttpClientConfig(new HttpClientConfig.Builder("http://localhost:" + wireMockRule.port()).build());
+
+        try (JestClient client = factory.getObject()) {
+            final ElasticFlowRepository elasticFlowRepository = new ElasticFlowRepository(new MetricRegistry(),
+                    client, IndexStrategy.MONTHLY, documentEnricher,
+                    sessionUtils, nodeDao, snmpInterfaceDao,
+                    new MockIdentity(), new MockTracerRegistry(), new MockDocumentForwarder(), new IndexSettings(),
+                    mock(SmartQueryService.class));
+
+            Integer ingress = 2;
```
fooker, line 275: Any reason why these (and others below) are `Integer`, not `int`?

fooker: In this case: just inline them or make them real constants (same for the `nodeId` below.

fooker (changes requested):

features/flows/itests/src/test/java/org/opennms/netmgt/flows/elastic/MarkerCacheIT.java
patrick-schweizer: :+1:  fixed

features/flows/itests/src/test/java/org/opennms/netmgt/flows/elastic/MarkerCacheIT.java
patrick-schweizer: fixed

features/flows/itests/src/test/java/org/opennms/netmgt/flows/elastic/MarkerCacheIT.java
patrick-schweizer: fixed

features/flows/itests/src/test/java/org/opennms/netmgt/flows/elastic/MarkerCacheIT.java
patrick-schweizer: > Any reason why these (and others below) are `Integer`, not `int`?
nope. fixed.

patrick-schweizer: > In this case: just inline them or make them real constants (same for the `nodeId` below.

The reason I wanted to use a named variable (and not inline) is that it reads easier (2 has no meaning). 
But they are only used in the scope of the method so I didn't wanted to make it a class constant.
I made it final as a compromise. Sounds good?

christianpape (approved):
Looks good to me.

PR #3166, opened 2020-09-23
NMS-12918: Added policy to add metadata on node-level

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-12918

Updates: 
- added policy to add also interface-level metadata.
- changed key and value to required parameters
- changed naming so that metadata parameters are grouped together

### DIFF
--- a/opennms-doc/guide-admin/src/asciidoc/text/provisioning/examples.adoc
+++ b/opennms-doc/guide-admin/src/asciidoc/text/provisioning/examples.adoc
@@ -321,6 +321,11 @@ node.setLocation(new OnmsMonitoringLocation("Minneapolis", ""));
 return node;
 ----
 
+.Setting Metadata Policy
+
+The Setting Metadata policy allows you to set node-level metadata in the context `requisition` for provisioned nodes.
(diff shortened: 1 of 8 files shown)


fooker (approved):

mfuhrmann:
@christianpape I've seen a lot of spellings yet that why I want to ask if there is an official spelling for meta-data? Or metadata? Or Meta-Data? Or MetaData?

fooker (approved):

fooker:
> @christianpape I've seen a lot of spellings yet that why I want to ask if there is an official spelling for meta-data? Or metadata? Or Meta-Data? Or MetaData?

It's "metadata" according to the docs. But code uses `{m,M}etaData` as it would've been two words.

mfuhrmann:
Then we might should correct it in other documentation parts.
For example: https://docs.opennms.org/opennms/releases/26.2.1/guide-admin/guide-admin.html#ga-provisioning-meta-data
But of course this should not be discussed here. I just wanted to make sure, the docs in this PR spell it correct, so no correction afterwards is required.

mfuhrmann (approved):

PR #3167, opened 2020-09-23
NMS-12915: Support gNMI for OpenConfig

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12915


cgorantla:
@fooker 
With current config, if we need different paths with different sampling rates to be subscribed, we can do that with multiple packages but that opens multiple gRPC streams with the target ( router).

I'm thinking of  multiple parameter groups would be good solution for this where multiple paths can use one Stream connection to subscribe. Following can be added to `PackageConfig`. Let me know what do you think

         <package name="OpenConfig-Default">
            <filter>IPADDR != '0.0.0.0'</filter>
            <parameter-groups>
                            <parameter-group>
					<parameter key="port" value="${requisition:oc.port|9000}"/>
					<parameter key="paths" value="/network-instances/network-instance[instance-name='master']"/>
					<parameter key="frequency" value= "300000"/>
					<parameter key="mode" value="jti"/>
				</parameter-group>
				<parameter-group>
					<parameter key="port" value="${requisition:oc.port|9000}"/>
					<parameter key="paths" value="/protocols/protocol/bgp"/>
					<parameter key="mode" value="jti"/>
				        <parameter key="frequency" value= "5000"/>
				</parameter-group>
            </parameter-groups>
         </package>

cgorantla:
@j-white suggested using some context in parameters to separate them in groups.

<parameter context="routers" key="port" value="${requisition:oc.port|9000}"/>
<parameter context="bgp"      key="port" value="${requisition:oc.port|9000}"/>

cgorantla:
Updated PR with support for multiple paths with different sampling rates.
Still there is no easy way to determine correct RRD as the output stream doesn't have sampling rate info.
will open another issue to tackle this.

fooker:
> @fooker
> With current config, if we need different paths with different sampling rates to be subscribed, we can do that with multiple packages but that opens multiple gRPC streams with the target ( router).
> 
> I'm thinking of multiple parameter groups would be good solution for this where multiple paths can use one Stream connection to subscribe. Following can be added to `PackageConfig`. Let me know what do you think

Wouldn't it be much easier to use multiple packages and aggregate the connections if required?
Allowing multiple packages allows to be more flexible with the filters. By aggregating the connections we can still be more performant.

And we avoid to extend the config syntax which needs to be reflected by the key-value implementation for minions, too.

cgorantla:
> > @fooker
> > With current config, if we need different paths with different sampling rates to be subscribed, we can do that with multiple packages but that opens multiple gRPC streams with the target ( router).
> > I'm thinking of multiple parameter groups would be good solution for this where multiple paths can use one Stream connection to subscribe. Following can be added to `PackageConfig`. Let me know what do you think
> 
> Wouldn't it be much easier to use multiple packages and aggregate the connections if required?
> Allowing multiple packages allows to be more flexible with the filters. By aggregating the connections we can still be more performant.
> 
> And we avoid to extend the config syntax which needs to be reflected by the key-value implementation for minions, too.

Added group to the package which seems to fit into what we need here.
parameters without group will be common params and paths/frequencies can be grouped into different groups.

        <package name="OpenConfig-Default">
            <filter>IPADDR != '0.0.0.0'</filter>
            <parameter key="port" value="${requisition:oc.port|9000}"/>
            <parameter key="mode" value="jti"/>
            <parameter group="group1" key="paths" value="/network-instances/network-instance[instance-name='master'],/protocols/protocol/bgp"/>
            <parameter group="group1" key="frequency" value="5000"/>
            <parameter group="group2" key="paths" value="/protocols/protocol/bgp"/>
            <parameter group="group2" key="frequency" value="30000"/>
            <parameter group="group3" key="paths" value="/interfaces/interface[id=25]"/>
            <parameter group="group3" key="frequency" value="60000"/>
        </package>

cgorantla:
@fooker As we discussed, we will keep `groups` handling as it is. We can revisit this implementation when `Connectors` are supported on Minion.

features/telemetry/daemon/src/main/java/org/opennms/netmgt/telemetry/daemon/ConnectorManager.java
```diff
@@ -88,28 +91,43 @@ private void startStreamingFor(ConnectorConfig connectorConfig, PackageConfig pa
...
+            interpolatedMapList.add(parmMap);
+
+            // Interpolate grouped parameters
+            Map<String, Map<String, String>> parmMapByGroup = packageConfig.getParameters().stream()
+                    .filter(parameter -> !Strings.isNullOrEmpty(parameter.getGroup()))
+                    .collect(Collectors.groupingBy(Parameter::getGroup, Collectors.toMap(Parameter::getKey, Parameter::getValue)));
+
+            parmMapByGroup.forEach((group, parmeterMap) -> {
+                interpolatedMapList.add(getInterpolated(parmeterMap, serviceRef));
+            });
```
fooker, line 113: Looks like the whole block just interpolates all keys regardless of the group.

Maybe something like this does the job:
```java
List<Map<String, String>> interpolatedMapList = packageConfig.getParameters().stream()
    .map(parmeterMap -> getInterpolated(parmeterMap, serviceRef))
    .collect(Collectors.toList());
```

features/openconfig/telemetry-client/src/main/java/org/opennms/features/openconfig/telemetry/OpenConfigClientImpl.java
```diff
@@ -78,16 +87,24 @@
...
     private ScheduledExecutorService scheduledExecutor = Executors.newSingleThreadScheduledExecutor();
 
-    public OpenConfigClientImpl(InetAddress host, Map<String, String> params) {
+    public OpenConfigClientImpl(InetAddress host, List<Map<String, String>> paramList) {
         this.host = Objects.requireNonNull(host);
-        this.port = Objects.requireNonNull(StringUtils.parseInt(params.get("port"), null));
-        parameters.putAll(params);
+        this.paramList.addAll(paramList);
+        // Extract port and mode which are global.
+        this.paramList.stream().filter(entry -> entry.get("port") != null)
```
fooker, line 103: I think it would be better to enforce en empty group on the global parameters. (And maybe print an error log message if they are specified with a group)

fooker (changes requested):
Looks good. Just minor changes.

features/telemetry/daemon/src/main/java/org/opennms/netmgt/telemetry/daemon/ConnectorManager.java
cgorantla: This is to put parameters that belong to one group into one set of params. Modified this and added test case to show how grouping is done.

features/openconfig/telemetry-client/src/main/java/org/opennms/features/openconfig/telemetry/OpenConfigClientImpl.java
cgorantla: At this point, Client doesn't have any group info. It has a set of parameters.

cgorantla:
@Bonrob2 Update docs part, please take a look once.

opennms-doc/guide-admin/src/asciidoc/text/telemetryd/protocols/openconfig.adoc
```diff
@@ -128,7 +137,16 @@ The script will be invoked for every OpenConfig stream data that is received and
...
+.Globals passed to the script for gnmi (default) mode.
+[options="header, autowidth"]
+|===
+| Parameter  | Description                                                    | Type
+| `agent`    | The agent (node) against which the metrics will be associated  | `org.opennms.netmgt.collection.api.CollectionAgent`
+| `builder`  | Builder in which the resources and metrics should be added     | `org.opennms.netmgt.collection.support.builder.CollectionSetBuilder`
+| `msg`      | Decoded message from which the metrics should be extracted     | `org.opennms.features.openconfig.proto.gnmi.Gnmi.SubscribeResponse`
+|===
+
+.Globals passed to the script for jti mode.
```
Bonrob2, line 149: > .Globals passed to the script for JTI mode.

Bonrob2 (changes requested):
A few minor changes.

cgorantla:
@fooker @Bonrob2 Handled review comments. Please take a look again

opennms-doc/guide-admin/src/asciidoc/text/telemetryd/protocols/openconfig.adoc
```diff
@@ -89,8 +89,8 @@ Use multiple packages to customize the parameter's subsets of nodes.
 
 The `filter` element is optional; if missing, all services with the given name will be considered.
 
-Parameters can also grouped into multiple groups when different paths need to be streamed at different frequencies using
-`group` element in parameter. Global parameters like port and connection specific parameters doesn't need any group.
+Use the group element to group parameters into multiple groups when you need to stream different paths at different frequencies using
+`group` element in parameter. Global parameters like port, and connection-specific parameters, don't need any group.
```
Bonrob2, line 93: @cgorantla ... Sorry my edit was confusing, but it went over two lines. This section should read like this:

Use the `group` element to group parameters into multiple groups when you need to stream different paths at different frequencies. Global parameters like port, and connection-specific parameters, don't need any group.

Bonrob2 (changes requested):
One edit, otherwise looks great!

opennms-doc/guide-admin/src/asciidoc/text/telemetryd/protocols/openconfig.adoc
cgorantla: my bad, didn't read the sentence properly.

Bonrob2 (approved):
LGTM!

RangerRick:
@fooker @cgorantla so is this good to go? If it's gonna make it into Horizon 27, this pretty much needs to be resolved today.

cgorantla:
Waiting for @fooker  to take one more look.

cgorantla:
@fooker  Guess I addressed the review comments. If there is anything else, I will address them in another PR.
Merging this for now since this is needed before `release-27`

PR #3168, opened 2020-09-23
NMS-12692: Hostname resolution for flow queries

fooker:
This implements hostname querying for aggregated flows processed by nephron
and fixes hostname querying for row flows.

This is accompanied by https://github.com/OpenNMS/nephron/pull/3.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12692

### DIFF
--- a/features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/agg/AggregatedFlowQueryService.java
+++ b/features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/agg/AggregatedFlowQueryService.java
@@ -31,11 +31,14 @@ package org.opennms.netmgt.flows.elastic.agg;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import org.opennms.features.jest.client.SearchResultUtils;
 import org.opennms.features.jest.client.index.IndexSelector;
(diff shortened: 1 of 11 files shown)


features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/agg/AggregatedFlowQueryService.java
```diff
@@ -84,22 +89,30 @@ public AggregatedFlowQueryService(JestClient client, IndexSelector indexSelector
 
     @Override
     public CompletableFuture<List<TrafficSummary<Conversation>>> getTopNConversationSummaries(int N, boolean includeOther, List<Filter> filters) {
-        return getTopNSummary(N, includeOther, filters, GroupedBy.EXPORTER_INTERFACE_CONVERSATION, Types.CONVERSATION );
+        return getTopNSummary(N, includeOther, filters, GroupedBy.EXPORTER_INTERFACE_CONVERSATION, Types.CONVERSATION);
```
j-white, line 92: Conversations need to be mapped as well.

j-white (commented):
Looks like this covers all the required functionality now - thanks for putting this together.

A couple comments:
1. Can we bring the `Types` class back? This is much nicer than using string literals everywhere.
2. Is it feasible to make one query to ES to resolve multiple hostnames instead of one query per host? When building a table of Top-10 conversations for example, this now requires 20 separate queries to ES with most of the same filter criteria.
3. Have we been able to see this work end-to-end w/ Nephron?

christianpape (approved):
Awesome!

fooker:
@j-white 
>     1. Can we bring the `Types` class back? This is much nicer than using string literals everywhere.
Will try to. But it will be less fancy. It needs to store `String` only because most of the processing happens on the string type, not on the resulting object type.

>     2. Is it feasible to make one query to ES to resolve multiple hostnames instead of one query per host? When building a table of Top-10 conversations for example, this now requires 20 separate queries to ES with most of the same filter criteria.
Don't think there is a way to do this. AFAIK there is no way to specify a query for `host=lowerIp OR host=upperIp` while limiting the result to one answer each.

>     3. Have we been able to see this work end-to-end w/ Nephron?
I'm working on it...

christianpape (approved):
Looks good.

PR #3170, opened 2020-09-28
NMS-10351: handle HTTP servers that do not response with a reason phrase

RangerRick:
This changes the HTTP header regex matcher to account for missing reason phrase.

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-10351


patrick-schweizer (approved):
Looks good to me

cgorantla (approved):

PR #3194, opened 2020-10-21
NMS-12961: "what's new" high-level release notes for Horizon 27

RangerRick:
Release Notes for Horizon 27

JIRA: https://issues.opennms.org/browse/NMS-12961


opennms-doc/releasenotes/src/asciidoc/releasenotes/whatsnew.adoc
```diff
@@ -1,205 +1,101 @@
...
+Documentation improvements and additions have been made in a number of areas:
+
+* Application Perspective Monitoring
+* Data Collection Configuration
+* Metadata
+* Mail Transport Monitor
+* Minion/Kafka RPC Tuning
+* Thresholding
+* Users Configuration, Duty Schedules, and Security Roles
+
```
Bonrob2, line 96: * Provisioning

Bonrob2 (changes requested):
Looks great. Minor changes.
Do we need to include gNMI/OpenConfig support? It's something we are highlighting in the marketing material.

RangerRick:
Oh yeah, I probably missed the gNMI thing, since we already had some level of OpenConfig support.

I did notice there's no mention of it elsewhere in the docs.  Do we need to add documentation to the openconfig docs section for gNMI?  @cgorantla?

RangerRick:
Or is that the same thing as what's labeled `gRPC` in the docs?

RangerRick:
Ah, digging around a bit and I guess `gNMI` is short for [`gRPC Network Management Interface`](https://github.com/openconfig/gnmi) so I wonder if we need to settle on one term to refer to it.

Bonrob2:
> Ah, digging around a bit and I guess `gNMI` is short for [`gRPC Network Management Interface`](https://github.com/openconfig/gnmi) so I wonder if we need to settle on one term to refer to it.

Yes, we should settle on one term. Since it's gRPC in the docs, we should stick with that at this point. If gNMI is more common use, then I can flag it as a ticket to update docs and other content in the next release.

opennms-doc/releasenotes/src/asciidoc/releasenotes/whatsnew.adoc
```diff
@@ -68,6 +68,7 @@ A number of provisioning, polling, and web UI changes have been made to accomoda
 * Flow DNS enrichment has been enhanced to improve performance.
 * Aggregate flow queries now support resolving hostnames if available.
 * Sink RPC APIs have been enhanced to expose metrics to Karaf for tracing and performance tuning.
+* The OpenConfig integration now supports gNMI, in addition to JTI.
```
Bonrob2, line 71: The OpenConfig integration now supports gRPC, in addition to JTI.

Bonrob2 (changes requested):
One more change. :-)

RangerRick:
The only reason I would say go with gNMI is that I _think_ technically JTI is _also_ a form of gRPC but with a different protocol/spec, so there's a level of confusion there.  @cgorantla would be able to answer this better.

Bonrob2:
> The only reason I would say go with gNMI is that I _think_ technically JTI is _also_ a form of gRPC but with a different protocol/spec, so there's a level of confusion there. @cgorantla would be able to answer this better.

Makes sense.

opennms-doc/releasenotes/src/asciidoc/releasenotes/whatsnew.adoc
```diff
@@ -1,205 +1,103 @@
...
+* `send-event`: additional options supported to match the CLI version of `send-event` more closely
+* `snmp-fit`: can now optionally send a `newSuspect` event on success
+* `snmp-get`: like `snmp-walk`, but fetches a single MIB object and prints the result
 
-=== Graph API
+=== Minion and Sentinel
 
-A new API has been added for querying graph data, inspired by link:http://graphml.graphdrawing.org[GraphML].  
-This API consolidates a number of internal APIs for accessing topology, application, BSM, link, and other data.
+* The Minion `confd` configuration support now handles a bunch more options, including setting Jetty, RMI, and ssh host and port, Secure Credentials Vault configuration, and environment variables and options for the JVM.
```
indigo423, line 39: If I recall correctly we can reference to the [CONFD_README](https://github.com/OpenNMS/opennms/blob/develop/opennms-container/minion/CONFD_README.md) which is the only place which documents what is possible with Minion and Confd.

opennms-doc/releasenotes/src/asciidoc/releasenotes/whatsnew.adoc
cgorantla: Yeah, even JTI is gRPC based so I think original statement seems to be fine.

Bonrob2: Works for me.

opennms-doc/releasenotes/src/asciidoc/releasenotes/whatsnew.adoc
RangerRick: Good point, I'll link it in the release notes.  We should probably make an issue to transition that to actual asciidoc.

PR #3200, opened 2020-10-27
cloud-473: add support for importing server certificates into trust store

swachter:
* adds confd configuration and template for server certificates
* `entrypoint.sh` imports server certificates in a copy of the default trust store and uses that trust store

### DIFF
--- a/opennms-container/minion/container-fs/entrypoint.sh
+++ b/opennms-container/minion/container-fs/entrypoint.sh
@@ -12,11 +12,13 @@ umask 002
 MINION_HOME="/opt/minion"
 MINION_CONFIG="/opt/minion/etc/org.opennms.minion.controller.cfg"
 MINION_PROCESS_ENV_CFG="/opt/minion/etc/minion-process.env"
+MINION_SERVER_CERTS_CFG="/opt/minion/etc/minion-server-certs.env"
 MINION_OVERLAY_ETC="/opt/minion-etc-overlay"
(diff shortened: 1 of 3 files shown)


opennms-container/minion/container-fs/entrypoint.sh
```diff
@@ -192,6 +194,14 @@ configure() {
       export "$assignment"
     done < "$MINION_PROCESS_ENV_CFG"
   fi
+  if [[ -f "$MINION_SERVER_CERTS_CFG" ]]; then
+    cp "$JAVA_HOME/lib/security/cacerts" "$CACERTS"
+    export JAVA_OPTS="$JAVA_OPTS -Djavax.net.ssl.trustStore=$CACERTS -Djavax.net.ssl.trustStorePassword=changeit"
```
mattixtech, line 199: Should we be using something besides `changeit`?

mattixtech (approved):
Looks good, just not sure about using the default `changeit`.

bouff (approved):
LGTM

opennms-container/minion/container-fs/entrypoint.sh
swachter: The added server certificates are public, i.e. they do not contains secrets. In this regard, security is not changed by this pull request. Securing the trust store may be worth another ticket.

swachter:
merged

PR #3208, opened 2020-11-04
NMS-12986: Fixed VMware polling and datacollection

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-12986

### DIFF
--- a/features/topology-map/plugins/org.opennms.features.topology.plugins.topo.vmware/src/main/java/org/opennms/features/topology/plugins/topo/vmware/internal/VmwareTopologyProvider.java
+++ b/features/topology-map/plugins/org.opennms.features.topology.plugins.topo.vmware/src/main/java/org/opennms/features/topology/plugins/topo/vmware/internal/VmwareTopologyProvider.java
@@ -54,6 +54,7 @@ import org.opennms.netmgt.model.OnmsNode;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Strings;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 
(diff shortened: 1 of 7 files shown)


RangerRick:
This looks like what I was imagining it would require, but tests are NPE'ing so I assume more changes are coming down the pipe.

christianpape:
> This looks like what I was imagining it would require, but tests are NPE'ing so I assume more changes are coming down the pipe.

Test was missing the added TransactionTemplate bean. Fixed now.

christianpape:
It seems, that the VMware Topology is also affected. I'll start to work on this...

features/topology-map/plugins/org.opennms.features.topology.plugins.topo.vmware/src/main/java/org/opennms/features/topology/plugins/topo/vmware/internal/VmwareTopologyProvider.java
```diff
@@ -231,6 +232,8 @@ private AbstractVertex createHostSystemVertex(String vertexId, String vertexName
         final String vmwareTopologyInfo = getTopologyInfo(onmsNode);
 
         return Arrays.stream(vmwareTopologyInfo.split(SPLIT_REGEXP))
+                .map(s -> s.trim())
+                .filter(s -> !Strings.isNullOrEmpty(s))
```
fooker, line 236:
```suggestion
                .filter(s -> !s.isEmpty())
```
should be enough

integrations/opennms-vmware/src/main/java/org/opennms/netmgt/collectd/VmwareCimCollector.java
```diff
@@ -193,6 +194,8 @@ private String applyUnitModifier(String attributeValue, String unitModifier) {
      */
     private NodeDao m_nodeDao = null;
 
+    private TransactionTemplate m_transactionTemplate = null;
```
fooker, line 197: Use `SessionUtils` instead.

fooker (approved):

RangerRick (approved):

PR #3209, opened 2020-11-05
NMS-12985: Added additional liquibase changeset for the VMware asset migration

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-12985


fooker (approved):

RangerRick (approved):
Seems like this should work...  Have we tried it with any live data?

christianpape:
> Seems like this should work... Have we tried it with any live data?

Yes, I tried it.

PR #3214, opened 2020-11-16
NMS-12930: fix timezone handling in Grafana reports

RangerRick:
This PR does significant changes to fix timezone handling in the report ReST UI and Grafana UI.

I tried to break the commits up into logical pieces so it's a bit easier to tell what's going on, since the inconsistencies come from a number of different levels.

First, parameters that come from a Jasper report are essentially independent of each other. BUT, the internal representation of a "date" in our reporting model converts a `yyyy-MM-dd HH:mm` datestring into a Java `Date` object as soon as report parameters are processed in the ReST servlet.  Since `ReportDateParm` objects are used elsewhere, I was hesitant to completely change the model, so I instead added a flag to indicate whether the stored `java.util.Date` object is timezone-adjusted or not. Now, if a `timezone` parameter is available when a report ReST request is made, it is used when converting any date parameters into `ReportDateParm` objects.

Second, when we query Grafana for a list of dashboards, the dashboard's timezone setting is not returned; instead you have to make a separate call to retrieve extended/detailed information about the dashboard.  I extended the Grafana query code to support retrieving this information.

Third, timezone calculations were wrong for a number of reasons in the `.jrxml` files themselves.  I added code to the `TimezoneHelper` utility function to correct for this, and also fixed the ordering of parameters so timezone is actually applied when making these calculations.

Finally, the UI for the reports was confusing, since timezone could be configured in both Grafana _and_ in this query interface. Now, if the Grafana dashboard has an explicitly set timezone, it is always used, and a note is put in the UI indicating this, like so:

Screen Shot 2020-11-16 at 12 04 01 PM

A number of other changes were made to the JavaScript code to make sure UI changes were handled properly before a final request is made when clicking `Create Report`.

# External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12930


pbrane:
> Finally, the UI for the reports was confusing, since timezone could be configured in both Grafana _and_ in this query interface. Now, if the Grafana dashboard has an explicitly set timezone, it is always used, and a note is put in the UI indicating this, like so:

Nicely handled here!

agalue (approved):
I tested multiple use cases, and everything works as expected; good job.

PR #3217, opened 2020-11-19
set system properties for enabling TLSv1.1, TLSv1.2, and TLSv1.3 for …

swachter:
…clients on startup

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-12733


swachter:
Wrong target branch!

PR #3223, opened 2020-11-24
NMS-13015: Fix NPE while dequeing

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13015

### DIFF
--- a/core/ipc/sink/common/src/main/java/org/opennms/core/ipc/sink/common/AsyncDispatcherImpl.java
+++ b/core/ipc/sink/common/src/main/java/org/opennms/core/ipc/sink/common/AsyncDispatcherImpl.java
@@ -224,8 +224,10 @@ public class AsyncDispatcherImpl<W, S extends Message, T extends Message> implem
             if (messageEntry.getKey() == null) {
                 return messageEntry;
             }
-
-            resultRecordedMap.remove(messageEntry.getKey()).await();
+            CountDownLatch resultRecorded = resultRecordedMap.remove(messageEntry.getKey());
+            if(resultRecorded != null) {
+                resultRecorded.await();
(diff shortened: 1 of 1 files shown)


RangerRick (approved):

core/ipc/sink/common/src/main/java/org/opennms/core/ipc/sink/common/AsyncDispatcherImpl.java
```diff
@@ -224,8 +224,10 @@ void enqueue(T message, String key, Consumer<DispatchQueue.EnqueueResult> onEnqu
             if (messageEntry.getKey() == null) {
                 return messageEntry;
             }
-
-            resultRecordedMap.remove(messageEntry.getKey()).await();
+            CountDownLatch resultRecorded = resultRecordedMap.remove(messageEntry.getKey());
+            if(resultRecorded != null) {
+                resultRecorded.await();
+            }
```
mattixtech, line 230: I guess this should only be the case when the result from enqueuing was deferred so we did not persist a latch to track it. I wonder why the above condition on line 224 wasn't written to handle this originally. Hope we are not missing anything, been a while since I looked at this.

mattixtech (approved):

core/ipc/sink/common/src/main/java/org/opennms/core/ipc/sink/common/AsyncDispatcherImpl.java
cgorantla: yeah, `DEFERRED` seems to be the one case this could happen. Think this is corner case when default Offheap Queue fills up.

PR #3224, opened 2020-11-24
NMS-13017: Use index patterns for events / situation feedback

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13017


agalue (approved):
On a test environment, and then in the affected environment at the client-side, I retrieved the current template through The Elasticsearch API, then fixed it to include `index_patterns` instead of `template` as instructed by this PR. Then, I removed the indices created with the wrong mapping. I manually created a new index in my test environment that matches the pattern, and it got created correctly. Due to the massive amount of incoming events, a new index was immediately created on the client-side, the events started to populate it, and the exceptions reported in the logs were gone.

So, the proposed solution works as expected.

RangerRick (approved):

cgorantla:
For now, targeting this to `release-27.x`  
Need to evaluate if this is safe to back port to `foundation-2020`

PR #3225, opened 2020-11-26
Proposal: Add lombok

swachter:
Lombok is a well established project and allows massive cuts of boilerplate.

I had to upgrade the Felix maven-bundle-plugin from 3.0.1 (Nov 2015) to 3.5.1 (June 2018) to make it work. (cf. [maven-bundle-plugin-fails-with-invalid-class-file-module-info-class](https://stackoverflow.com/questions/50530927/maven-bundle-plugin-fails-with-invalid-class-file-module-info-class))

In a former pull request Patrick has already shown some of its benefits: https://github.com/OpenNMS/opennms/pull/2156

I want to point out another one that I like most: The type of local variables can be inferred using the [`val`](https://projectlombok.org/features/val) identifier.

I already added many reviewers because the introduction of Lombok should be agreed upon by the team. If someone is missing as a reviewer then please add she or he.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -151,6 +151,14 @@
     <module>opennms-asterisk</module>
     <module>opennms-config-dao</module>
   </modules>
+  <dependencies>
+    <dependency>
+      <groupId>org.projectlombok</groupId>
+      <artifactId>lombok</artifactId>
+      <version>1.18.16</version>
+      <scope>provided</scope>
+    </dependency>
+  </dependencies>
   <build>
   <extensions>
     <extension>
@@ -223,7 +231,7 @@
       <plugin>
         <groupId>org.apache.felix</groupId>
         <artifactId>maven-bundle-plugin</artifactId>
-        <version>3.0.1</version>
+        <version>3.5.1</version>
         <extensions>true</extensions>
         <configuration>
           <obrRepository>NONE</obrRepository>
@@ -1306,7 +1314,7 @@
     <install.perl.bin>/usr/bin/perl</install.perl.bin>
 
     <runMailTests>true</runMailTests>
-  
+
     <!-- rpm specific -->
     <rpm.install.dir>/opt/opennms</rpm.install.dir>
     <rpm.build.executable>rpmbuild</rpm.build.executable>


indigo423:
I've removed myself from the reviewer. I can't really add value to the discussion as I'm not involved with the development. I'll be happy with any decision is made :)

fooker:
Please push to branch named `lombok-smoke` to get full smoke test run coverage to rule out implications on build system.

fooker:
I have some mixed feelings on this one.

On one side I'm a fan of more fancy language constructs and would advocate for things like `val`. I like `final`izing as much as possible and would encourage to build similar patterns as generated by lombok.

On the other hand, this is not something that should be decided by personal preference. Introducing such syntax would higher the burden for review, debugging and maintenance as every contributor has to learn the syntax and relay on IDE integration to make something useful out of the generated stuff.

I don't have enough experience with lombok to come to a good conclusion here. I feel like we should, at least, do some kind of experiment and come up with some things that every developer should test in their setup / workflow.

To decide this, I figure that we should find a strong consent as this is going to impact our work for a long time and will be hard to revert as introduced once.

patrick-schweizer:
I am in favor of using Lombok but I am with @fooker we should get a broad consensus.
Maybe a good way to go about it would be to have a little presentation where @swachter could show us some of the benefits of Lombok? Maybe part of the Wednesday meeting? I think it's much easier to show this instead of just discussing in a pull request about it.

But let me address a few points Dustin made and add a few more of my own.

I have used Lombok in all my projects (except OpenNMS) over the last ~10 years and never had any issues. I think its a very mature and un-intrusive library. It helps produce better code and less boilerplate code.

> Introducing such syntax would higher 
> the burden for review, debugging and maintenance as every contributor has to learn the syntax and relay on IDE integration to 
> make something useful out of the generated stuff.
While I agree on a theoretical level I don't think it will be an impediment in practice. Lombok has only a couple of annotations and integrates very transparently into the IDE. There is not much to learn about it.
OpenNMS has so many different frameworks, concepts and libraries that makes it a super complex system. It strikes me odd that we are willing to add more and more of these to this system but are reluctant to add a library that reduces complexity and boilerplate code.
As for the review: I think it makes the review easier instead of more complex.
Imagine a class with some attributes. Each of those attribute will have a getter and setter, we have hashCode(), equals(), toString() etc. That's a lot of code to review that tells us nothing since it is standard (boilerplate).
Imagine now the same class with lombok: it will have only the attributes and an annotation @Data. I can review it very quickly and see that there is nothing special about it. But now imagine we have a special setter hat does something more than a "standard" setter. In the normal case this would be hidden between all the normal stuff and hard to spot. In the lombok case this would be the only setter actually implemented manually => very easy to spot. 

> I don't have enough experience with lombok to come to a good conclusion here. I feel like we should, at least, do some kind of > experiment and come up with some things that every developer should test in their setup / workflow.
+1
Maybe we could use it only in one module at first so everyone can take a look and experience a bit with it. 

> To decide this, I figure that we should find a strong consent as this is going to impact our work for a long time and will be hard 
> to revert as introduced once.
+1 on the consensus
I don't think there will be along term impact if we don't like it. It can be removed with a command:
https://projectlombok.org/features/delombok
The annotations will vanish and the generated code will appear. You can try this in the IDE on a class level.

indigo423:
* https://medium.com/@vgonzalo/dont-use-lombok-672418daa819
* https://github.com/rzwitserloot/lombok/issues?q=is%3Aissue+is%3Aopen+%5BBUG%5D

mvrueden:
I have some experience with lombok and with my overall experience in the OpenNMS code base I would not recommend using lombok as of now.

The main issues I know exist:

**1. You don't see what you get.**
Most of the time it works, however when you really need to debug deeply into some code segments, the `lombok` annotated classes are not what you will debug. Therefore you get a "byte code differs from source code" message. This might be fine in some cases, however I prefer to debug the source code, because it is easier.

**2. it doesn't help you that much**
The IDE support is quite decent, however especially with `@Getter` and `@Setter` you still have to manually fix things (Still talking about IntelliJ). Let's say you rename a property from `state` to `status` and have a `@Getter`  annotation set. All references to `getState()` have to be manually changed to `getStatus()`, whereas that will automatically happen when using the refactor rename mechanism from your IDE

**3. m_ weirdness**
The old legacy code of OpenNMS still uses `m_` as property prefixes, which will either require weirdness in lombok or renaming those properties.

In my opinion the whole lombok introduction is not really worth it. 
Sure the source code has a lot of boiler plate, however most of the code can be generated by the IDE very nicely.
In addition if you really want to get it done right, you have to either adapt existing lombok stuff or write your own code anyways. So for me, it does not really add enough usefulness to counter the risks.

Also I believe that in future Java releases most of the lombok features will make it into the native Java language.

As I don't have much stakes into this anymore, I believe this adds unnecessary complexity without introducing too much "it makes my live easier". As soon as you have to really debug into generated classes and not the source code you will regret it, I always do in that case (-:

swachter:
I must completely agree with Patrick. I also used Lombok in various projects for many years without any problem.

I can not follow the arguments given by @mvrueden. I seems to me a very rare exception that one has to look into byte code for tracking down issues. In addition, the generated methods are really bullet proofed, I do not expect that generated equals, hashCode, setters, or getters are ever the source of a bug.

IMHO avoiding boilerplate really helps a lot. As Patrick mentioned, unimportant parts do not distract you from the important ones. IDEs only help in writing code. Empirical studies show however, that reading code is way more important. (https://www.goodreads.com/quotes/835238-indeed-the-ratio-of-time-spent-reading-versus-writing-is).

Finally, if you can not use Lombok for parts of your code because of some constraints then simply do not use it there.

fooker:
> While I agree on a theoretical level I don't think it will be an impediment in practice. Lombok has only a couple of annotations and integrates very transparently into the IDE. There is not much to learn about it.

That's not what I experienced so far. As an example, today i stumbled upon a situation where I have to debug into an external library (testcontainers) which uses lombok. I was not able to set up may IDE (inetllij with lombok plugin and annotation processing enabled) in a way that the code of dependency was usefull in any way. Breakpoints are not working, variables are missing and the IDE can't find references. This may be a setup problem - but that's not what I understand under easy setup and good integration.

> OpenNMS has so many different frameworks, concepts and libraries that makes it a super complex system. It strikes me odd that we are willing to add more and more of these to this system but are reluctant to add a library that reduces complexity and boilerplate code.

That's not really a good point. The Project has already realized (after some mistakes) that adding new frameworks is not always a good idea. And using this as an excuse seems wrong to me.

> As for the review: I think it makes the review easier instead of more complex. [...]

Therefore you have to know what the `@Data` Annotation does. Having that knowledge is what I mean by increasing the burden.

> The IDE support is quite decent, however especially with `@Getter` and `@Setter` you still have to manually fix things.

For me, this would be a deal-breaker. Having good reference tracking is essential for me. This is not only about getters/setters but also about constructors and any other kind of method signatures.

> IMHO avoiding boilerplate really helps a lot. As Patrick mentioned, unimportant parts do not distract you from the important ones. IDEs only help in writing code. Empirical studies show however, that reading code is way more important.

I would use almost the same argument against lombok. I don't really see a problem with boilerplate if I don't have to write it (That's why I use an IDE). But for reading code I feel more comfortable with scrolling over boilerplate than to build the generated structures in my head to make sense out of code that I'm reading.

I also can't follow the argument that code like "special" setters are easy to spot. There is a living contract that we all agree on when it comes to these patterns. If there is somethings called a setter but does more then setting a property, then I would call this bad interface design and it requires a comment at least.

> Finally, if you can not use Lombok for parts of your code because of some constraints then simply do not use it there.

The thing is, code interacts with other code. And current mode of work on the project requires that all developers touch/understand all parts of the code. So, enabling that for a certain part of the project just for personal preference does not work out here.

mvrueden:
> I can not follow the arguments given by @mvrueden. I seems to me a very rare exception that one has to look into byte code for tracking down issues. In addition, the generated methods are really bullet proofed, I do not expect that generated equals, hashCode, setters, or getters are ever the source of a bug.

That is not what I am saying.
You almost never just have getter/setter in your objects. The various lombok annotations are proof of that in itself.
This means you are most likely (not rarely) to add custom code to those objects.
That in turn means, that there is code which contains bug, because I wrote the code and therefore may have to debug it. Then I am left with decompiled classes. Yay. 

> I completely agree with Patrick. I also used Lombok in various projects for many years without any problem.

I pointed out concrete issues I always face when using lombok. There are no counter-arguments provided which neglect my arguments, thus one example is enough to prove something wrong.

> IMHO avoiding boilerplate really helps a lot. As Patrick mentioned, unimportant parts do not distract you from the important ones. IDEs only help in writing code. Empirical studies show however, that reading code is way more important. (https://www.goodreads.com/quotes/835238-indeed-the-ratio-of-time-spent-reading-versus-writing-is).

I would also agree with @fooker on this one. The absence of the boilerplate does not reduce the complexity, in fact it increases it. The reason for this is, that your brain has to "manually generate" those segments in your head, therefore more complexity.

> OpenNMS has so many different frameworks, concepts and libraries that makes it a super complex system. It strikes me odd that we are willing to add more and more of these to this system but are reluctant to add a library that reduces complexity and boilerplate code.

As mentioned above, I do not think it reduces complexity, but increases it instead.

> Finally, if you can not use Lombok for parts of your code because of some constraints then simply do not use it there.

That simply does not work.
OpenNMS is too big for that, or phrased differently: How are you going to enforce these constraints? 
Do you want to add checkstyle rules or manually enforce those during review?
I thought lombok is supposed to make my life easier, but now I have to consider all these things, where I just wanted to write a simple Data object. Doesn't sound easier to me.

mvrueden:
I tried refactoring a field of java classes again and it is actually working as expected and renaming the getter/setter accordingly.

I don't know why this never worked for me before.

Maybe IntelliJ was in a weird state.

mattixtech:
Lots of good discussion so far. I would lean towards not bringing it in for some of the same reasons @fooker and @mvrueden have already alluded to.

I think it is really difficult to justify bringing something like this into a project such as OpenNMS that has been around for so long and accumulated so much already. I'm not sure the additional cognitive load added by such a library is worth it.

From my perspective Lombok is identifying legitimate problems with developing in Java but in doing so it is fundamentally changing the look and feel of the code to not resemble Java anymore. For better or worse the project consists of a ton of close to vanilla Java code (some quite old!). Adding Lombok now means contributors have to understand Lombok in addition to Java. Also that they have to understand the next framework that comes along that proposes to fix other perceived issues not covered by Lombok, the floodgates are open!.

If we are really committed to going down this path I would throw Kotlin up against Lombok as an alternative to consider. At the end of the day its just another 3rd party dep with some compile time considerations just like Lombok and it fixes a lot of the same problems and more (while having better IDE support, better debugging support, large community etc..). See this is what I mean about opening the floodgates 😄 

Also keep in mind newer version of Java address some of these issues as well. Type inference + records are really nice in upcoming Java versions and it would be kind of a shame to use Lombok to address that when a Java version upgrade will add another way of doing the same thing.

RangerRick:
> If we are really committed to going down this path I would throw Kotlin up against Lombok as an alternative to consider. At the end of the day its just another 3rd party dep with some compile time considerations just like Lombok and it fixes a lot of the same problems and more (while having better IDE support, better debugging support, large community etc..). See this is what I mean about opening the floodgates 😄

I hadn't gotten around to commenting yet but this is pretty much what I was going to say.  While Lombok does address some shortcomings in Java, Kotlin covers plenty of them as and we're _already_ using Kotlin in other OpenNMS projects (plus it has WAY more momentum behind tooling and IDE support, being An Official Way™️ to do Android programming).  I was _already_ assuming it was just a matter of time before Kotlin would end up on the OpenNMS side and we just hadn't gotten to it yet.

> Also keep in mind newer version of Java address some of these issues as well. Type inference + records are really nice in upcoming Java versions and it would be kind of a shame to use Lombok to address that when a Java version upgrade will add another way of doing the same thing.

This too.

In the end, if we're adding something, it should be Kotlin IMHO since it's less dissonance and has a lot of the same advantages.

fooker:
> In the end, if we're adding something, it should be Kotlin IMHO since it's less dissonance and has a lot of the same advantages.

I'm preferring to ad kotlin (if any), too. The concepts just seem to be more mature.

I've played some more wit lombok during the last days. My main pain point is still unsolved: Having a project using lombok as a dependency screws up debugging.

Therefore I'm currently having a strong opinion for voting against lombok.

swachter:
> In the end, if we're adding something, it should be Kotlin IMHO since it's less dissonance and has a lot of the same advantages.

Shall I open a pull request that adds Kotlin support? Would that have a chance?

mattixtech:
> > In the end, if we're adding something, it should be Kotlin IMHO since it's less dissonance and has a lot of the same advantages.
> 
> Shall I open a pull request that adds Kotlin support? Would that have a chance?

Personally I would love to see the option of writing in Kotlin on OpenNMS. However that is just my personal opinion I don't want to push that on any of our contributors. It comes with the same burdens of Lombok in that people need to understand a new technology.

If we are really pining for a way to write more expressive code without some of the annoyances of Java then it is a good fit. We have to be pretty bullish on it as a group to go in that direction though IMO. I can speak more in depth about pros & cons and our experience using it on some of the other projects we have on the go if people want to seriously consider it.

If we did start using Kotlin, one of the nicest things would be that we get "modern" language features without having to upgrade the JDK. That is one of Kotlin's strongest selling points for me is that you can target JDK 1.6 or 1.8 and still get all the modern languages features. Then the fact that we are still on JDK 1.8 doesn't become as much of an annoyance.

There would be implications on the OSGi level, we would have to make sure we have the bundle wired up to any of the modules that end up using it. Not sure if we have a precedent for that yet where we would have some "core library" available to most or all of the bundles running (see https://kotlinlang.org/docs/reference/kotlin-osgi.html).

swachter:
I thank all reviewers for their opinion. As of now, there is a clear reluctance of some of you to introduce Lombok. The interesting result though, was that we might introduce Kotlin.

I think having this discussion as a review of a pull request was appropriate. Therefore I will prepare a corresponding pull request introducing Kotlin eventually.

RangerRick:
> There would be implications on the OSGi level, we would have to make sure we have the bundle wired up to any of the modules that end up using it. Not sure if we have a precedent for that yet where we would have some "core library" available to most or all of the bundles running (see https://kotlinlang.org/docs/reference/kotlin-osgi.html).

In the short term, it would probably go into the "main" OpenNMS classpath in `lib/` so as long as we expose it in `custom.properties` we shouldn't have any major issues sharing it with OSGi.  Long-term, we have so many other issues to solve getting everything _into_ Karaf that it will be the least of our troubles. :D

PR #3226, opened 2020-11-26
NMS-12973: Adding metrics and checking sequences

fooker:
This is groundwork used to analyze the situation where we maybe drop flows  during processing with hostname resolution enabled.

This enables a variety of counters related to flow processing. Mostly watching out for error situations but also adding a throughput counter to every stage.

In addition, the packet sequence numbers are check and a log message is generated if a packet is detected missing.

Oh, and it fixes a memory leak if flows are considered illegal.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12973

### DIFF
--- a/features/flows/kafka-persistence/src/main/java/org/opennms/netmgt/flows/persistence/KafkaFlowForwarder.java
+++ b/features/flows/kafka-persistence/src/main/java/org/opennms/netmgt/flows/persistence/KafkaFlowForwarder.java
@@ -48,6 +48,9 @@ import org.osgi.service.cm.ConfigurationAdmin;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.codahale.metrics.Counter;
+import com.codahale.metrics.Meter;
+import com.codahale.metrics.MetricRegistry;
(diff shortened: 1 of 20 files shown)


features/flows/kafka-persistence/src/main/java/org/opennms/netmgt/flows/persistence/KafkaFlowForwarder.java
```diff
@@ -63,14 +66,27 @@
             .build();
     private Properties producerConfig;
 
-    public KafkaFlowForwarder(ConfigurationAdmin configAdmin) {
+    private final Meter forwarded;
+    private final Meter persisted;
+    private final Counter skipped;
+    private final Counter failed;
+
+    public KafkaFlowForwarder(ConfigurationAdmin configAdmin,
```
christianpape, line 74:
```suggestion
    public KafkaFlowForwarder(final ConfigurationAdmin configAdmin,
```

christianpape (approved):
Looks good.

j-white (approved):

PR #3233, opened 2020-12-03
NMS-13034: Expose telemetryd metrics with full names

fooker:
Build and expose metrics in telemetryd for parsers and adapters using a
unique, full qualified name. The name is build using the parent element
to allow duplicated names in parsers and adapters.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13034

### DIFF
--- a/features/telemetry/config/api/src/main/java/org/opennms/netmgt/telemetry/config/api/AdapterDefinition.java
+++ b/features/telemetry/config/api/src/main/java/org/opennms/netmgt/telemetry/config/api/AdapterDefinition.java
@@ -36,15 +36,11 @@ import java.util.List;
 public interface AdapterDefinition extends TelemetryBeanDefinition {
 
     /**
-     * The name of the adapter.
+     * Returns the full qualified (unique) name of the adapter.
      *
-     * This is used as a suffix for any associated queues that are created and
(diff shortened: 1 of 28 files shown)


christianpape (approved):
LGTM

RangerRick:
Should this branch be a `-smoke` one so it runs through the full test suite?  Or do you think the `SinglePortFlowsIT` is enough to be sure this didn't break anything?

fooker:
> Should this branch be a `-smoke` one so it runs through the full test suite? Or do you think the `SinglePortFlowsIT` is enough to be sure this didn't break anything?

full smoke test: https://app.circleci.com/pipelines/github/OpenNMS/opennms?branch=jira%2FNMS-13034-smoke

RangerRick (approved):

cgorantla:
May be it's good to update `telemetryd-configuration.xml` in smoke-test overlay to have two listeners with the same parser just to make sure that it works.

RangerRick:
So the tests are failing, and it appears there's probably another place that requires us to change `threshd-configuration.xml` -- but that means we're breaking this in a micro-release in a way that requires users to encode the queue name into their config, is that correct?  Is there a way to do this that hides it from the configs and just does it internally?

PR #3234, opened 2020-12-03
add -T start option for debug and suspend

swachter:
At the moment it is difficult to debug code that is immediately run on startup. The suspend option helps here.


RangerRick (approved):
Looks reasonable to me... you should probably add `[-T]` to the Usage: bit through.

j-white:
Yay, no more hacking the `bin/opennms` script every time I want to do this.

PR #3235, opened 2020-12-03
NMS-13035: revert to bouncycastle 1.66 for now, 1.67 broke things

RangerRick:
see https://github.com/AgNO3/jcifs-ng/issues/250#issuecomment-706700044
for details

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13035


rcsnogm:
I ran into this as well.  It appears that in my case the org.bouncycastle.asn1.DEROutputStream was public in bc 1.66 with a note that it would be removed from the public api.  In bc 1.67 they fullfilled that promise.  When I tried 1.67 jcifs.spnego.NegTokenInit on line 158 uses that class in version 2.1.5.  I reverted bc for now to 1.66 but we might be under pressure to move forward soon.

PR #3237, opened 2020-12-04
NMS-13042: Handle error scenarios

cgorantla:
Currently Snmp4JStrategy doesn't throw specific exception for timeout and
error scenarios. This results in null response which cause ArrayIndexBound exception

This PR should handle those error scenarios.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13042

### DIFF
--- a/core/snmp/api/src/main/java/org/opennms/netmgt/snmp/SnmpAgentTimeoutException.java
+++ b/core/snmp/api/src/main/java/org/opennms/netmgt/snmp/SnmpAgentTimeoutException.java
@@ -36,4 +36,8 @@ public class SnmpAgentTimeoutException extends Exception {
     public SnmpAgentTimeoutException(String name, InetAddress agentAddress) {
         super(String.format("Timeout retrieving '%s' for %s.", name, InetAddrUtils.str(agentAddress)));
     }
+
+    public SnmpAgentTimeoutException(InetAddress agentAddress) {
+        super(String.format("Timeout retrieving response for %s.", InetAddrUtils.str(agentAddress)));
+    }
 }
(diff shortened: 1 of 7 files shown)


RangerRick (approved):
This _seems_ good to me... it does scare me adding thrown exceptions to something as core as the SNMP4J strategy though... are we sure there aren't any side-effects elsewhere?  Feel like at the least you should run this in a `-smoke` branch.

cgorantla:
> This _seems_ good to me... it does scare me adding thrown exceptions to something as core as the SNMP4J strategy though... are we sure there aren't any side-effects elsewhere? Feel like at the least you should run this in a `-smoke` branch.

The calling method catches all the exceptions. So I  think should be safe.

cgorantla:
> > This _seems_ good to me... it does scare me adding thrown exceptions to something as core as the SNMP4J strategy though... are we sure there aren't any side-effects elsewhere? Feel like at the least you should run this in a `-smoke` branch.
> 
> The calling method catches all the exceptions. So I think should be safe.

Also SNMP4JWalker seems to do similar thing.  Odd that this error was not caught earlier as it is simple timeout issue. Going to do some digging.

cgorantla:
> > > This _seems_ good to me... it does scare me adding thrown exceptions to something as core as the SNMP4J strategy though... are we sure there aren't any side-effects elsewhere? Feel like at the least you should run this in a `-smoke` branch.
> > 
> > 
> > The calling method catches all the exceptions. So I think should be safe.
> 
> Also SNMP4JWalker seems to do similar thing. Odd that this error was not caught earlier as it is simple timeout issue. Going to do some digging.

Okay. Looks like SNMPInterfacePoller is the only place that uses get with multiple oids.  All others are get with one oid which would not throw this exception.

PR #3239, opened 2020-12-07
NMS-12963: Enhance routes with ASN/Route info

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12963


opennms-bamboo:
Welcome CI/CD System (cicd-system@opennms.com).

I could not find a signed OpenNMS Contributor Agreement (OCA) assigned to the username CI/CD System (cicd-system@opennms.com).

christianpape (changes requested):
I had only a few small comments...

features/telemetry/protocols/bmp/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/adapter/AsnInfo.java
```diff
@@ -0,0 +1,200 @@
...
+
+    public void setRemarks(String remarks) {
+        this.remarks = remarks;
+    }
+
+    public static AsnInfo parseOutput(Long asn, String source, String rawOutput) {
+        AsnInfo asnInfo = new AsnInfo(asn, source, rawOutput);
+        // Split output into lines
+        String[] lines = rawOutput.split("\\r?\\n");
+        for (String line : lines) {
```
christianpape, line 154: As far as I know, there can be entries can appear more than once. For instance there can be multiple address entries if the address is more than one line. The OpenBMP-implementation concatenates these strings before persisting them. So, this is missing here.

christianpape (changes requested):
Found another problem

features/telemetry/protocols/bmp/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/adapter/AsnInfo.java
cgorantla: yeah, OpenBMP has this handling for RIPE. I m looking for an example AS number to properly implement this.  Currently all the responses I'm getting are from ARIN and RTT.

cgorantla: I updated few things but still didn't get an example for having Address in multiple lines even with RIPE. We can address this later if we find such scenarios.

cgorantla:
@christianpape Thanks for the review. Can you take a look again.

features/telemetry/protocols/bmp/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/adapter/AsnInfo.java
christianpape: Ok

christianpape (approved):
From my point of view it's fine, but I would like Dustin to take another look at it to make sure.

features/telemetry/protocols/bmp/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/adapter/AsnInfo.java
fooker: @cgorantla https://www.ripe.net/manage-ips-and-asns/db/support/querying-the-ripe-database has some examples showing multiple address lines. What do you need for testing? Maybe I can provide something.

cgorantla: I was looking for an AS number that can give us output just like above example.  
We need to split multiple address lines into  address, city,  state and country. I'm looking if there is any standard way to do that.

fooker: My understanding of the code above is that it will overwrite with the last encountered address value. Maybe it would be more suitable to concatenate the multiple lines separated by `\n`. Looks like that's what openbmp is doing: https://github.com/SNAS/openbmp-mysql-consumer/blob/master/cron_scripts/gen-whois/gen-whois-asn.py#L186 .

The reference code calls the `whois` command, which by default also fetches related entries. Maybe `org.apache.commons.net.whois.WhoisClient` does not do this?

`whois -h whois.ripe.net AS8319` shows some output with multiple address lines.

This is not only true for the `address` field but for almost all fields (RIPE marks them as `multiple` in the link above).

cgorantla: Removing optional a with the query `a AS8319` seems to give the output identical to `whois` client.  Updated code to append multiple lines in address/remarks/orgId as per link above.

cgorantla: @fooker  This is taken care with dc3c5b6.  Can you take a look

features/telemetry/protocols/bmp/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/adapter/BmpMessagePersister.java
```diff
@@ -540,6 +600,120 @@ private BmpGlobalIpRib buildGlobalIpRib(PrefixByAS prefixByAS) {
...
+                bmpRouteInfo.setOriginAs(routeInfo.getOriginAs());
+                bmpRouteInfo.setSource(routeInfo.getSource());
+            }
+            bmpRouteInfo.setLastUpdated(Date.from(Instant.now()));
+            return bmpRouteInfo;
+        }
+        return null;
+    }
+
+    private List<BmpAsnPathAnalysis> buildBmpAsnPath(String asnPath) {
```
fooker, line 647: Can we have a test for this?

features/telemetry/protocols/bmp/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/adapter/BmpMessagePersister.java
```diff
@@ -540,6 +600,120 @@ private BmpGlobalIpRib buildGlobalIpRib(PrefixByAS prefixByAS) {
...
+                    } catch (NumberFormatException e) {
+                        e.printStackTrace();
+                        break;
+                    }
+
+                    if (rightAsn.equals(asn)) {
+                        continue;
+                    }
+
+                    Boolean isPeeringAsn = (i == 0 || i == 1) ? TRUE : FALSE;
```
fooker, line 684: This code path is never taken if `i == 0` because `asn` will always be `0` on the first iteration.

features/telemetry/protocols/bmp/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/adapter/BmpMessagePersister.java
```diff
@@ -540,6 +600,120 @@ private BmpGlobalIpRib buildGlobalIpRib(PrefixByAS prefixByAS) {
...
+            bmpRouteInfo.setLastUpdated(Date.from(Instant.now()));
+            return bmpRouteInfo;
+        }
+        return null;
+    }
+
+    private List<BmpAsnPathAnalysis> buildBmpAsnPath(String asnPath) {
+
+        List<BmpAsnPathAnalysis> bmpAsnPathAnalyses = new ArrayList<>();
+        String[] asnArray = asnPath.split(" ");
```
fooker, line 650: Why not parse the ASNs into `long`s here? this would save some error handling in the loop.
By doing so, we can skip the whole update if the path in invalid.

features/telemetry/protocols/bmp/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/adapter/BmpMessagePersister.java
```diff
@@ -540,6 +600,120 @@ private BmpGlobalIpRib buildGlobalIpRib(PrefixByAS prefixByAS) {
...
+
+        Long leftAsn = 0L;
+        Long rightAsn = 0L;
+        Long asn = 0L;
+        for (int i = 0; i < asnArray.length; i++) {
```
fooker, line 655: The loop can be simplified by iterating from `i = 1` to `i = asnArray.length - `. The  you can use `i-1` and `i+1` for left and right ASN inside the loop.

In addition, the Origin AS handling can be done out of loop at the end.

fooker (changes requested):
Beside the tuning of the AS-Path calculation, this looks very good.

features/telemetry/protocols/bmp/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/adapter/BmpMessagePersister.java
cgorantla: Simplified this and added a test case for expected output.

fooker (approved):

PR #3242, opened 2020-12-16
NMS-13055: Fix Snmp karaf commands

cgorantla:
Don't break for Timeout Exception as it's designed to be responsive to user to give output every 1 sec.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13055


dino2gnt:
Nice!

RangerRick (approved):

PR #3248, opened 2020-12-18
NMS-8959: "recommend" haveged package in OpenNMS install

RangerRick:
As of RPM 4.13, it now supports `Recommends:` and `Suggests:` like Debian does.  This PR adds a `Recommends:` to both the debian and RPM packages.  Also, I have added `haveged` to the RPM repositories in stable, so it's available without requiring the EPEL repository to be enabled.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-8959


indigo423 (approved):
👍 I deploy it in my environments as well. Totally make sense to me.

PR #3253, opened 2021-01-04
NMS-13069: remove all references and deps to nsis

RangerRick:
With the removal of Windows support and the remote poller installer, we should no longer need anything related to `makensis`.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13069


indigo423 (approved):
Tears of Joy :) LGTM! 👍

PR #3261, opened 2021-01-13
NMS-13088: Adapt invalid flows instead of dropping

fooker:
Flows with invalid start and end values are asumed to exist in a short
timespan before the flow was exported. Therefore we accept the
inaccuracy in time but take the flow into account.

No flow is illegal!

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13088

### DIFF
--- a/features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/ParserBase.java
+++ b/features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/ParserBase.java
@@ -34,6 +34,7 @@ import java.nio.ByteBuffer;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Date;
+import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
@@ -54,8 +55,10 @@ import org.opennms.netmgt.model.events.EventBuilder;
 import org.opennms.netmgt.telemetry.api.receiver.Parser;
 import org.opennms.netmgt.telemetry.api.receiver.TelemetryMessage;
 import org.opennms.netmgt.telemetry.protocols.netflow.parser.ie.RecordProvider;
-import org.opennms.netmgt.telemetry.protocols.netflow.parser.ie.Value;
 import org.opennms.netmgt.telemetry.protocols.netflow.parser.session.Session;
+import org.opennms.netmgt.telemetry.protocols.netflow.parser.transport.MessageBuilder;
+import org.opennms.netmgt.telemetry.protocols.netflow.transport.FlowMessage;
+import org.opennms.netmgt.telemetry.protocols.netflow.transport.FlowMessageOrBuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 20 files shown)


Naicisum:
I would not remove the event, users need to be aware of a device with this condition. Newer software/firmware could correct this condition.

A rate limited warn would suffice, once every 24 hours by default sounds good.

cgorantla:
Looks good to me. Yeah, we can keep the event and keep the `invalidFlows` metric.

fooker:
@Naicisum: this makes it more complicated - but here you go

@cgorantla, @christianpape: The firs commit just shuffles things around but keeps the (pretty good) tests untouched to make sure nothing is screwed up. The second commit add the wanted functionality on top.

features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/transport/IpFixMessageBuilder.java
```diff
@@ -283,52 +513,50 @@ public IpFixMessageBuilder(Iterable<Value<?>> values, RecordEnrichment enrichmen
...
+                        double samplingIntervalValue = population / size;
                         builder.setSamplingInterval(setDoubleValue(samplingIntervalValue));
                         break;
-                    case 4:
-                        Double samplingProbability = this.samplingProbability != null ?
-                                this.samplingProbability.doubleValue() : 1.0;
-                        builder.setSamplingInterval(setDoubleValue(1.0 / samplingProbability));
+                    }
+                    case 4: {
+                        Double probability = samplingProbability != null ? samplingProbability.doubleValue() : 1.0;
```
christianpape, line 546: Any reason to have a Double instead of a double here?

features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/ParserBase.java
```diff
@@ -323,20 +343,19 @@ public void setThreads(int threads) {
...
                                         .addParam("monitoringSystemLocation", identity.getLocation())
-                                        .setParam("cause", e.getMessage())
+                                        .setParam("cause", Joiner.on('\n').join(corrections))
                                         .setParam("protocol", protocol.name())
                                         .setParam("illegalFlowEventRate", (int) getIllegalFlowEventRate())
                                         .getEvent());
 
-                                LOG.warn("Illegal flow detected from exporter {}", session.getRemoteAddress().getAddress(), e);
+                                for (final String correction : corrections) {
+                                    LOG.warn("Illegal flow detected from exporter {}: \n{}", session.getRemoteAddress().getAddress(), correction);
```
Naicisum, line 352: Is this supposed to be `LOG` and not `CORRECTION_LOGGER`?

Naicisum (changes requested):

christianpape (changes requested):
Looks good. Only the LOG must be changed to CORRECTION_LOGGER.

features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/transport/IpFixMessageBuilder.java
fooker: I've just moved that code and didn't bother to make it nice :). The whole section needs reworking, TBH.

features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/ParserBase.java
fooker: This is already rate-limited (as the event is). The `CORRECTION_LOGGER` is a left-over an can be removed.

christianpape (approved):
NIce

cgorantla (approved):

Naicisum (approved):

PR #3265, opened 2021-01-14
NMS-13065: Resolve hostname only once

fooker:
Querying aggregated data sotred by nephron involves to resolve hostnames
for queries by host and by conversation. When querying for time-series,
the hostnames are now queryied only once by key instead of every
time-bucket.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13065

### DIFF
--- a/features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/ElasticFlowQueryService.java
+++ b/features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/ElasticFlowQueryService.java
@@ -28,6 +28,7 @@
 
 package org.opennms.netmgt.flows.elastic;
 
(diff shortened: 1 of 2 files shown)


patrick-schweizer:
I am not sure I understand the code change and it's implications. I wonder if he have a test that makes sure the behavior is as before?

christianpape (approved):
LGTM

fooker:
> I am not sure I understand the code change and it's implications. I wonder if he have a test that makes sure the behavior is as before?

The functionality itself is covered by a bunch of tests already (`FlowQueryIT#canGetTopNHostSummaries` et al). If you have an idea on how to verify the number of queries executed, I'm all ears.

PR #3268, opened 2021-01-18
NMS-13094: Move ServiceTracker beans to Eventd

cgorantla:
Any module that needs ServiceTracker need to depend on Eventd 
or need to retrieve it as OSGi Service 

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13094


agalue (approved):
I tested the fix via Docker using the OCI image from CircleCI, and it works. I verified by checking that the `opennms:filter` command is available when having Telemetryd disabled.

j-white (approved):

PR #3269, opened 2021-01-19
NMS-12944: Improved fix

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-12944


opennms-wmi/src/main/java/org/opennms/netmgt/collectd/WmiCollector.java
```diff
@@ -273,20 +262,6 @@ private boolean isGroupAvailable(final WmiAgentState agentState, final Wpm wpm)
         return true;
     }
 
-    private Resource getWmiResource(final Map<String, ResourceType> resourceTypesMap, final CollectionAgent agent, final String resourceType, final NodeLevelResource nodeResource, final String instance) {
-        ResourceType rt = resourceTypesMap.get(resourceType);
-        if (rt == null) {
-            LOG.debug("getWmiResourceType: using default WMI resource type strategy - index / all");
```
j-white, line 279: One thing to watch out for here is that this fallback behavior is no longer preserved when using the `DeferredGenericTypeResource`.  See https://github.com/OpenNMS/opennms/blob/304b9a3ede9765579e627afa13359d4b580a3657/features/collection/api/src/main/java/org/opennms/netmgt/collection/support/builder/DeferredGenericTypeResource.java#L110.

This means that referencing invalid/missing resource types in the WMI collector configuration will no longer work.

I think this is ultimately better, but it may be a breaking change for some.

christianpape: How about adding a fallback Callable<ResourceType> to the DeferredGenericTypeResource that can be used to construct a ResourceType if all other has failed?

j-white: Challenge there is that the `DeferredGenericTypeResource` gets created on the Minion, serialized/deserialized, and then resolved on the core - so we would need to serialize the reference to the callable.

Turns out there was a similar issue with the XML collector that was solved this way:
1) Define a default resource type in etc:
https://github.com/OpenNMS/opennms/commit/b94939fafd0f26843147054076446138ba4c6131#diff-8720003b6108a087e7c4e73135adf61564d74e84b714b9d7738aa54b681200c9R1
2) Use this resource type as a fallback:
https://github.com/OpenNMS/opennms/commit/b94939fafd0f26843147054076446138ba4c6131#diff-041473e230ac153e735489fcbea477456acfd622e8c3c68b1f27d51149f27fdbR288

christianpape: Yes, that sound better. I'll modify the PR accordingly.

j-white (approved):

PR #3272, opened 2021-01-20
Parsers are actually children of listeners not siblings

mattixtech:
The schema described flows configuration incorrectly. This PR updates the schema to the correct form. This schema is only a reference right now and isn't used directly.

This PR also updates the key paths for SSH host port as they were incorrect.

No Jira.


bouff (approved):
LGTM 👍

PR #3274, opened 2021-01-21
NMS-13070: fix timezone handling in scheduled reports

RangerRick:
I took a two-pronged approach to handling this.

1. Always perform all operations from the point-of-view of the report time zone. Schedule reports based on that zone, render dates in the Grafana reports based on that zone, etc.
2. Give the user visibility into the time zone when scheduling the report.  This means:
   * show the time zone right next to the "time" input when scheduling
   * based on the schedule, show the next time it would run, in the report's time zone, the server's time zone, and the user's time zone

Screen Shot 2021-01-20 at 4 43 33 PM

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13070


dino2gnt:
This is fantastic.

Naicisum (commented):
Tested ad-hoc and new scheduled report generation. No issues.

PR #3276, opened 2021-01-25
NMS-13104: Enable single topic by default for Kafka RPC

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13104


j-white (commented):
LGTM. Would need to note this as a breaking change since it will affect anyone using Kafka RPC.

RangerRick (approved):

cgorantla:
Updated Breaking changes.  Full suite of smoke tests passing.

PR #3277, opened 2021-01-26
NMS-13072: Add JMX prometheus exporter

indigo423:
Add Prometheus JMX exporter for out of band management of the Minion Java Virtual Machine.
Set yaml formatting and add Prometheus JMX docs.

### All Contributors

Added JMX prometheus exporter agent which is disabled by default. It has to be explicitly enabled with the Java option `-javaagent:/opt/prom-jmx-exporter/jmx_prometheus_javaagent.jar=9299:/opt/prom-jmx-exporter/config.yaml`. The port 9299/tcp needs to be published has well.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13072

### DIFF
--- a/opennms-container/minion/CONFD_README.md
+++ b/opennms-container/minion/CONFD_README.md
@@ -9,7 +9,7 @@ specific configuration files will overwrite the corresponding config provided by
 ## Contents
 The following describes the keys that can be specified in `minion-config.yaml` to configure the Minion via confd.
 ### Minion Controller Config
-```
+```yaml
 --- 
 broker-url: "<broker url>"
 http-url: "<http url>"
@@ -22,14 +22,14 @@ Config specified will be written to `etc/org.opennms.minion.controller.cfg`.
 Supplying the http or broker username/password via yaml file for configuration via confd is not supported.
 
 ### Instance Id
-```
+```yaml
 ---
 org.opennms.instance.id: "<instance id>"
 ```
 Config specified will be written to `etc/instance-id.properties`.
 
 ### AWS SQS
-```
+```yaml
 ---
 aws:
     aws_region: "us-east-1"
@@ -48,7 +48,7 @@ ipc:
 Config specified will be written to `etc/org.opennms.core.ipc.aws.sqs.cfg`.
 
 ### Kafka RPC
-```
+```yaml
 --- 
 ipc:
     rpc:
@@ -62,7 +62,7 @@ Config specified will be written to `etc/org.opennms.core.ipc.rpc.kafka.cfg`. Ad
 `bootstrap.servers` key is specified, `etc/featuresBoot.d/kafka-rpc.boot` will also be updated.
 
 ### Kafka SINK
-```
+```yaml
 --- 
 ipc:
     sink:
@@ -79,7 +79,7 @@ Config specified will be written to `etc/org.opennms.core.ipc.sink.kafka.cfg`. A
 `bootstrap.servers` key is specified, `etc/featuresBoot.d/kafka-sink.boot` will also be updated.
 
 ### Sink Off Heap
-```
+```yaml
 --- 
 ipc:
     sink:
@@ -93,7 +93,7 @@ Config specified will be written to `etc/org.opennms.core.ipc.sink.offheap.cfg`.
 ### Single Port Flows
 To configure flows on a single port, set the following `enabled` key to `true`. Optionally parameters can be provided
 that will be included in the generated config.
-```
+```yaml
 --- 
 telemetry:
     flows:
@@ -110,7 +110,7 @@ Config specified will be written to `etc/org.opennms.features.telemtry.listeners
 ### Telemetry Flow Listeners
 Individual flow listeners can be configured. See the example below for how to specify parameters and parsers. Any number
 of uniquely named listeners can be defined.
-```
+```yaml
 --- 
 telemetry:
     flows:
@@ -130,7 +130,7 @@ telemetry:
 Config specified will be written to `etc/org.opennms.features.telemtry.listeners-<Listener-Name>.cfg`.
 
 ### Syslog
-```
+```yaml
 --- 
 netmgt:
     syslog:
@@ -141,7 +141,7 @@ netmgt:
 Config specified will be written to `etc/org.opennms.netmgt.syslog.cfg`.
 
 ### Traps
-```
+```yaml
 --- 
 netmgt:
     traps:
@@ -152,7 +152,7 @@ netmgt:
 Config specified will be written to `etc/org.opennms.netmgt.trapd.cfg`.
 
 ### System Properties
-```
+```yaml
 --- 
 system:
     properties:
@@ -164,7 +164,7 @@ Config specified will be written to `etc/confd.system.properties` which gets aut
 `jaeger-agent-host` key is specified, `etc/featuresBoot.d/jaeger.boot` will also be updated.
 
 ### Karaf Properties
-```
+```yaml
 ---
 karaf:
     shell:
@@ -185,7 +185,7 @@ Config specified will be written to:
 - `etc/org.apache.karaf.management.cfg` for content under `management`.
 
 ### Jetty Properties
-```
+```yaml
 ---
 jetty:
     web:
@@ -195,7 +195,7 @@ jetty:
 Config specified will be written to `etc/org.ops4j.pax.web.cfg`
 
 ### Secure Credentials Vault Provider
-```
+```yaml
 --- 
 scv:
     provider: "dominion"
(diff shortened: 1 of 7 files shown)


opennms-container/minion/Dockerfile
```diff
@@ -8,11 +8,21 @@ ARG BASE_IMAGE="opennms/deploy-base:jre-1.2.0.b54"
 
 FROM ${BASE_IMAGE} as minion-base
 
+ARG PROM_JMX_EXPORTER_VERSION="0.14.0"
```
j-white, line 11: Looks like 0.15.0 was just released :D

https://mobile.twitter.com/roidelapluie/status/1353748899882741775

opennms-container/minion/container-fs/confd/templates/prom-jmx-exporter.yaml.tmpl
```diff
@@ -0,0 +1,25 @@
+#
+# DON'T EDIT THIS FILE :: GENERATED WITH CONFD
+#
+{{$jmxExporterPath := "/java/agent/prom-jmx-exporter" -}}
```
mattixtech, line 4: Would it be better to simply not generate the file in the case they haven't configured the exporter rather than have default values here?

opennms-container/minion/CONFD_README.md
```diff
@@ -238,3 +238,29 @@ services:
...
       - ${PWD}/opennms/opennms-container/minion/container-fs/entrypoint.sh:/entrypoint.sh
 ```
+## Prometheus JMX Exporter
+
+To provide an out of band management of the JVM with the Minion process, the Prometheus JMX exporter is shipped with this container image.
+The default configuration is set to the following values and can be set in the `minion-config.yaml` file:
+
+```yaml
+---
+java
```
mattixtech, line 248: Could you update the `minion-config-schema.yaml` file to include the new options?

mattixtech (changes requested):

opennms-container/minion/Dockerfile
indigo423: I just saw it right after creating the PR :) I have it running in my environment and haven't seen any issues. Thanks for the heads up :)

opennms-container/minion/container-fs/confd/templates/prom-jmx-exporter.yaml.tmpl
indigo423: Yes, agree. I'll look into this. 👍

opennms-container/minion/container-fs/confd/templates/prom-jmx-exporter.yaml.tmpl
```diff
@@ -1,6 +1,8 @@
 #
```
mattixtech, line 1: You'll need to include the comments inside the if condition otherwise the file will not be empty and won't be removed.

mattixtech (commented):
Looking good now, we should update the minion-config-schema.yaml for reference.

Btw I plan to retire the CONFD_README file in favour of the schema + some generated doc file at some point in the future.

bouff:
Looks good from my perspective also. 👍

indigo423:
@mattixtech I've added the Minion configuration schema. Can you please double-check if this makes sense. Thank you for the review.

opennms-container/minion/container-fs/confd/templates/prom-jmx-exporter.yaml.tmpl
```diff
@@ -0,0 +1,29 @@
...
+# DON'T EDIT THIS FILE :: GENERATED WITH CONFD
+#
+
+{{$jmxExporterPath := "/java/agent/prom-jmx-exporter" -}}
+jmxUrl: '{{getv (print $jmxExporterPath "/jmxUrl") "service:jmx:rmi:///jndi/rmi://127.0.0.1:1299/karaf-minion"}}'
+username: {{getv (print $jmxExporterPath "/username") "admin"}}
+password: {{getv (print $jmxExporterPath "/password") "admin"}}
+lowercaseOutputName: {{getv (print $jmxExporterPath "/lowercaseOutputName") "true"}}
+lowercaseOutputLabelNames: {{getv (print $jmxExporterPath "/lowercaseOutputLabelNames") "true"}}
+whitelistObjectNames: {{getv (print $jmxExporterPath "/whitelistObjectNames") `[ "org.opennms.core.ipc.sink.producer:*", "org.opennms.netmgt.dnsresolver.netty:*", "org.opennms.netmgt.telemetry:*" ]`}}
```
mattixtech, line 12: Could you move the `[]` directly into the template, omitting them from the default value specified here? If we do that, we can use the `set` type in the schema rather than `string` which will render the config form a bit nicer.

See the `bootstrap-servers` for an example where we are using a `set` and concating together with `,` which we could do here.

In the future we can specify validation for the elements to ensure they are quoted in the correct form, but that can wait.

mattixtech (changes requested):
Looking good, got one comment.

opennms-container/minion/minion-config-schema.yml
```diff
@@ -120,6 +120,64 @@ categories:
...
+              - name: "lowercaseOutputLabelNames"
+                type:
+                  name: "boolean"
+                label: "Lowercase output label names"
+                description: "Lowercase the output metric label names."
+                key: "/java/agent/prom-jmx-exporter/lowercaseOutputLabelNames"
+                example: "true"
+              - name: "whitelistObjectNames"
+                type:
+                  name: "string"
```
mattixtech, line 176: This is what I'm proposing changing to `set` (and then adding the strategy and element information).

indigo423:
@mattixtech Ok thanks for your help and was able to get the array list working. While testing I've noticed something which I don't have a good solution. This line here enables the JMX exporter:

```yaml
process-env:
   java-opts:
     - -javaagent:/opt/prom-jmx-exporter/jmx_prometheus_javaagent.jar=9299:/opt/prom-jmx-exporter/config.yaml
```

It will run with a working default configuration. When I surround the template with an if statement, no default configuration is created and it will fail? What do you think should be the behavior.

mattixtech:
> @mattixtech Ok thanks for your help and was able to get the array list working. While testing I've noticed something which I don't have a good solution. This line here enables the JMX exporter:
> 
> ```yaml
> process-env:
>    java-opts:
>      - -javaagent:/opt/prom-jmx-exporter/jmx_prometheus_javaagent.jar=9299:/opt/prom-jmx-exporter/config.yaml
> ```
> 
> It will run with a working default configuration. When I surround the template with an if statement, no default configuration is created and it will fail? What do you think should be the behavior.

What part of the template are you surrounding with an if statement and why? If all the templated parameters have defaults then we probably want to always generate the file with whatever defaults / specified values we end up with right?

indigo423:
> > @mattixtech Ok thanks for your help and was able to get the array list working. While testing I've noticed something which I don't have a good solution. This line here enables the JMX exporter:
> > ```yaml
> > process-env:
> >    java-opts:
> >      - -javaagent:/opt/prom-jmx-exporter/jmx_prometheus_javaagent.jar=9299:/opt/prom-jmx-exporter/config.yaml
> > ```
> > 
> > 
> > It will run with a working default configuration. When I surround the template with an if statement, no default configuration is created and it will fail? What do you think should be the behavior.
> 
> What part of the template are you surrounding with an if statement and why? If all the templated parameters have defaults then we probably want to always generate the file with whatever defaults / specified values we end up with right?

You have suggested to create an if statement around the [config.yaml](opennms-container/minion/container-fs/confd/templates/prom-jmx-exporter.yaml.tmpl) for the JMX exporter. A user could just set the Java agent opts and would be fine. With the if statement and not providing a config.yml with defaults, it would require a user to configure it.

mattixtech:
> > > @mattixtech Ok thanks for your help and was able to get the array list working. While testing I've noticed something which I don't have a good solution. This line here enables the JMX exporter:
> > > ```yaml
> > > process-env:
> > >    java-opts:
> > >      - -javaagent:/opt/prom-jmx-exporter/jmx_prometheus_javaagent.jar=9299:/opt/prom-jmx-exporter/config.yaml
> > > ```
> > > 
> > > 
> > > It will run with a working default configuration. When I surround the template with an if statement, no default configuration is created and it will fail? What do you think should be the behavior.
> > 
> > 
> > What part of the template are you surrounding with an if statement and why? If all the templated parameters have defaults then we probably want to always generate the file with whatever defaults / specified values we end up with right?
> 
> You have suggested creating an if statement around the [config.yaml](https://github.com/OpenNMS/opennms/pull/3277/files#diff-1694c162707961844a8b6747153faf03a4d03e83ac59ee8700c284818cb99214) for the JMX exporter. A user could just set the Java agent opts and would be fine. With the if statement and it would require a user to configure it and just setting the Java options the Minion won't start and fails with an error. I don't have a strong preference here, just want to make sure we are on the same page and the consequences.

If you are happy with writing the defaults then we don't need a global if statement. Is that what you are referring to? It looks like we aren't doing `remove-if-empty` anyways so the file will get written regardless.

mattixtech (approved):

PR #3284, opened 2021-02-02
NMS-13106: Allows flows to be received out of order

fooker:
This allows flows to be recieved out of order, either because of network related packet reordering or parallel processing.
    
Flow sequence checks now are more patient. If a flow is missing, it is reported only after the flow sequence has advanced for a certain amount.

The patience of flow sequence tracking has a default of `32` which is sane for almost all use-cases. This makes the patience parameter per parser definition in telemetryd configuration allowing compensate for more paralelized receivers or crazy network.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13106

### Remarks
For reviewers: The first commit improves the sequence tracking mechanism while the second makes the patience parameter configurable.

### DIFF
--- a/features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/session/SequenceNumberTracker.java
+++ b/features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/session/SequenceNumberTracker.java
@@ -28,19 +28,115 @@
 
 package org.opennms.netmgt.telemetry.protocols.netflow.parser.session;
 
+import java.util.Arrays;
+
+/**
+ * Tracks sequence numbers and verify completeness.
+ *
+ * Sequence numbers passed in to the tracker using the {@link #verify(long)} method are checked for completeness. I.e.
+ * every sequence number has been passed in exactly once.
+ *
+ * While checking for completeness, the tracker allows for sequence numbers to be passed in out of order. Therefore it
+ * keeps track of a short history. The {@code patience} parameter specifies the length of this history controlling how
+ * much the expected sequence number can advance before the missing sequence number is reported as missing. If the
+ * missing sequence number is passed in before the the expected sequence number advances to much, the element is
+ * considered out-of-order but not missing and history is updated.
+ *
+ * To allow re-initialisation of sequence numbers, the tracker is lenient for huge sequence number jumps. If the passed
+ * in sequence number differs from the expected sequence number by more than what {@code patience} parameters allows,
+ * the tracker is reset and the element is considered valid.
+ */
 public class SequenceNumberTracker {
 
-    private long expected;
+    /**
+     * The highest seen sequence number.
+     */
+    private long current;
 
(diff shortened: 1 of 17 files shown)


features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/session/SequenceNumberTracker.java
```diff
@@ -28,19 +28,115 @@
...
+    /**
+     * The history of seen sequence numbers relative to the current sequence number.
+     *
+     * The history is stored in a ring with the size of the expected {@code patience}. By doing so, marking a sequence
+     * number as gives the status of of the expected sequence number - patience.
+     */
+    private final Ring seen;
+
+    public SequenceNumberTracker(final int patience) {
+        if (patience < 0) {
```
christianpape, line 65: Even 1 would make no sense, right?

christianpape (approved):
Noice!

Naicisum:
Looks nice, my only concern is around performance under significant load situations.

features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/session/SequenceNumberTracker.java
```diff
@@ -28,19 +28,115 @@
...
+    /**
+     * The highest seen sequence number.
+     */
+    private long current;
+
+    /**
+     * The history of seen sequence numbers relative to the current sequence number.
+     *
+     * The history is stored in a ring with the size of the expected {@code patience}. By doing so, marking a sequence
+     * number as gives the status of of the expected sequence number - patience.
```
swachter, line 60: > The history is stored in a ring with the size of the expected {@code patience}. By doing so, marking a sequence
> number as gives the status of of the expected sequence number - patience.

These two lines of comment need clarification.

The implementation was somewhat difficult to grasp for me. I developed the following mental model to understand the code and to see its correctness (maybe this could help other code readers too): The `Ring` is like a fifo queue. It tracks if sequence numbers need not to be seen (because they have already be seen or because the tracking started over).

swachter (approved):
LGTM

fooker:
> Looks nice, my only concern is around performance under significant load situations.

@Naicisum I've updated the code to short-circuit the flow tracking algorithm if the patience is configured to be `0`.

features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/session/SequenceNumberTracker.java
fooker: I've updated the comment to explain some more what this does - If this still needs explanation, I can work on details some more.

christianpape (approved):
Good

swachter (approved):
LGTM

PR #3296, opened 2021-02-10
NMS-13064: fix timezone editing for existing reports

RangerRick:
It turns out there were a couple of weird things going on with how this was initializing.  I've made some changes to mitigate them:

1. use a straight AJAX call rather than AngularJS `$resource` which can be problematic
2. when the upstream dashboard timezone is set to `undefined` (ie, "default") we were overwriting the already-configured zone with that undefined value
3. we weren't always updating the `parametersByName['timezone']` property -- this is probably fine, but belt & suspenders I made it happen always when updating the parameter list
4. when checking for malformed zones (like `utc` rather than `UTC`) we were accepting undefined values and overwriting it
5. bonus: when editing an existing report, we don't re-initialize the "endpoint" list; the grafana endpoint should never change on edit so I hid it from the UI

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13064

### DIFF
--- a/core/web-assets/src/main/assets/js/apps/onms-reports/ReportDetails.js
+++ b/core/web-assets/src/main/assets/js/apps/onms-reports/ReportDetails.js
@@ -187,20 +187,21 @@ export default class ReportDetails {
         this.parameters.filter((parameter) => {
             return parameter.type === 'timezone';
          }).forEach((parameter) => {
-             if (this.timezoneEditable) {
-                if (this.supportedTimezones.indexOf(parameter.value) >= 0) {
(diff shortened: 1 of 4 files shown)


RangerRick:
Welp, spoke too soon... looks like the `GRAFANA_ENDPOINT_UID` and `GRAFANA_DASHBOARD_UID` parameters are not initializing properly.  Converting this to draft for now.

christianpape (approved):
Looks good to me. I tested it and it seem to work fine now.

PR #3297, opened 2021-02-15
NMS-13117: Added smoke test to verify the reloading of daemons

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-13117

### DIFF
--- a/core/lib/src/main/java/org/opennms/core/daemon/DaemonReloadEnum.java
+++ b/core/lib/src/main/java/org/opennms/core/daemon/DaemonReloadEnum.java
@@ -30,6 +30,8 @@ package org.opennms.core.daemon;
 
 public enum DaemonReloadEnum {
 
+    ALARMD("alarmd"),
+    COLLECTD("Collectd"),
     EVENTD("Eventd"),
     NOTIFD("Notifd"),
     POLLERD("Pollerd"),
--- a/smoke-test/src/test/java/org/opennms/smoketest/ReloadDaemonIT.java
+++ b/smoke-test/src/test/java/org/opennms/smoketest/ReloadDaemonIT.java
@@ -1,0 +1,94 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2021 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2021 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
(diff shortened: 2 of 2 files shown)


cgorantla:
Looks good to me. 
Can you consider extending [DaemonReloadEnum](https://github.com/OpenNMS/opennms/blob/6afe2c7a121d32687765c2ff827dbdc8b15ca62f/core/lib/src/main/java/org/opennms/core/daemon/DaemonReloadEnum.java#L31) to include all missing Daemons like Alarmd/Collectd ?

Then I would opt for different Enum in Tests to pick and choose which daemons to run in smoke test.

christianpape:
> Looks good to me.
> Can you consider extending [DaemonReloadEnum](https://github.com/OpenNMS/opennms/blob/6afe2c7a121d32687765c2ff827dbdc8b15ca62f/core/lib/src/main/java/org/opennms/core/daemon/DaemonReloadEnum.java#L31) to include all missing Daemons like Alarmd/Collectd ?
> 
> Then I would opt for different Enum in Tests to pick and choose which daemons to run in smoke test.

Ok, I’ll add Alarm and Collectd to the enum, but do we really need a different enum for the smoke-test? All daemons except Telemetryd reload instantly. I think Telemetryd is important to test. So, testing all does not really add much delay to the test...

core/lib/src/main/java/org/opennms/core/daemon/DaemonReloadEnum.java
```diff
@@ -30,6 +30,8 @@
 
 public enum DaemonReloadEnum {
 
+    ALARMD("Alarmd"),
```
cgorantla, line 33: Looks like the names for these Daemons are in lower case `alarmd` and `collectd`.

christianpape: You're right

cgorantla:
Tests are failing.  May be One of `collectd` or `alarmd` reload daemon is broken ?

christianpape:
> Tests are failing. May be One of `collectd` or `alarmd` reload daemon is broken ?

Fixed.

cgorantla:
> > Tests are failing. May be One of `collectd` or `alarmd` reload daemon is broken ?
> 
> Fixed.

Okay, got it. [This](https://github.com/OpenNMS/opennms/blob/d716e549b975c7fec4bcf25e399c7fd381963a47/opennms-services/src/main/java/org/opennms/netmgt/collectd/Collectd.java#L1120) forces it to `Collectd` instead of name.

cgorantla (approved):

PR #3298, opened 2021-02-16
NMS-13129: Added CDP/LLDP search option

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-13129


RangerRick (approved):
Seems pretty straightforward to me.

PR #3303, opened 2021-02-17
NMS-13156: Enable newts cache priming if not disabled

fooker:
The setting to disable the newts cache priming is now interpreted
correctly.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13156


RangerRick:
Could this affect older versions? Does this need fixing in older Meridians?

RangerRick (approved):

j-white:
This also affects `foundation-2018`. Here's the patch that was applied while *trying* to turn cache priming on by default: https://github.com/OpenNMS/opennms/commit/df0b845fad08801afca6bd8a16f1f570faaf7cba

christianpape (approved):
Yay!

fooker:
Rebased to foundation-2018.

PR #3310, opened 2021-02-22
NMS-13167: Don't drop messages, resend the record for events.

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13167

### DIFF
--- a/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
+++ b/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
@@ -49,6 +49,7 @@ import java.util.function.Consumer;
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.clients.producer.RecordMetadata;
+import org.apache.kafka.common.errors.TimeoutException;
 import org.apache.kafka.common.serialization.ByteArraySerializer;
 import org.joda.time.Duration;
 import org.opennms.core.ipc.common.kafka.Utils;
@@ -408,6 +409,10 @@ public class OpennmsKafkaProducer implements AlarmLifecycleListener, EventListen
                     producer.send(producerRecord, (recordMetadata, e) -> {
                         if (e != null) {
                             LOG.warn("Failed to send record to producer: {}.", producerRecord, e);
+                            if(e instanceof TimeoutException) {
+                                // If Kafka is Offline, buffer the record again.
+                                kafkaSendQueue.offer(kafkaRecord);
+                            }
                             return;
                         }
                         if (consumer != null) {
--- a/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
+++ b/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
@@ -129,7 +129,7 @@ public class OpennmsKafkaProducer implements AlarmLifecycleListener, EventListen
     private final AlarmCallbackStateTracker stateTracker = new AlarmCallbackStateTracker();
     private final OnmsTopologyDao topologyDao;
     private int kafkaSendQueueCapacity;
-    private BlockingDeque<KafkaRecord> kafkaSendQueue;
+    private BlockingDeque<KafkaRecord> kafkaSendDeque;
     private final ExecutorService kafkaSendQueueExecutor =
             Executors.newSingleThreadExecutor(runnable -> new Thread(runnable, "KafkaSendQueueProcessor"));
 
@@ -167,7 +167,7 @@ public class OpennmsKafkaProducer implements AlarmLifecycleListener, EventListen
             LOG.info("Defaulted the 'kafkaSendQueueCapacity' to 1000 since no property was set");
         }
 
-        kafkaSendQueue = new LinkedBlockingDeque<>(kafkaSendQueueCapacity);
+        kafkaSendDeque = new LinkedBlockingDeque<>(kafkaSendQueueCapacity);
         kafkaSendQueueExecutor.execute(this::processKafkaSendQueue);
 
         if (forwardEvents) {
@@ -392,7 +392,7 @@ public class OpennmsKafkaProducer implements AlarmLifecycleListener, EventListen
 
         // Any offer that fails due to capacity overflow will simply be dropped and will have to wait until the next
         // sync to be processed so this is just a best effort attempt
-        if (!kafkaSendQueue.offer(new KafkaRecord(record, callback))) {
+        if (!kafkaSendDeque.offer(new KafkaRecord(record, callback))) {
             RATE_LIMITED_LOGGER.warn("Dropped a Kafka record due to queue capacity being full.");
         }
     }
@@ -401,7 +401,7 @@ public class OpennmsKafkaProducer implements AlarmLifecycleListener, EventListen
         //noinspection InfiniteLoopStatement
         while (true) {
             try {
-                KafkaRecord kafkaRecord = kafkaSendQueue.take();
+                KafkaRecord kafkaRecord = kafkaSendDeque.take();
                 ProducerRecord<byte[], byte[]> producerRecord = kafkaRecord.getProducerRecord();
                 Consumer<RecordMetadata> consumer = kafkaRecord.getConsumer();
 
@@ -413,8 +413,8 @@ public class OpennmsKafkaProducer implements AlarmLifecycleListener, EventListen
                                 // If Kafka is Offline, buffer the record again for events.
                                 // This is best effort to keep the order although in-flight elements may still miss the order.
                                 if (producerRecord != null &&
-                                        this.eventTopic.equalsIgnoreCase(kafkaRecord.producerRecord.topic())) {
-                                    if(!kafkaSendQueue.offerFirst(kafkaRecord)) {
+                                        this.eventTopic.equalsIgnoreCase(producerRecord.topic())) {
+                                    if(!kafkaSendDeque.offerFirst(kafkaRecord)) {
                                         RATE_LIMITED_LOGGER.warn("Dropped a Kafka record due to queue capacity being full.");
                                     }
                                 }
--- a/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
+++ b/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
@@ -37,13 +37,13 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
-import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.BlockingDeque;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.LinkedBlockingDeque;
 import java.util.function.Consumer;
 
 import org.apache.kafka.clients.producer.KafkaProducer;
@@ -129,7 +129,7 @@ public class OpennmsKafkaProducer implements AlarmLifecycleListener, EventListen
     private final AlarmCallbackStateTracker stateTracker = new AlarmCallbackStateTracker();
     private final OnmsTopologyDao topologyDao;
     private int kafkaSendQueueCapacity;
-    private BlockingQueue<KafkaRecord> kafkaSendQueue;
+    private BlockingDeque<KafkaRecord> kafkaSendQueue;
     private final ExecutorService kafkaSendQueueExecutor =
             Executors.newSingleThreadExecutor(runnable -> new Thread(runnable, "KafkaSendQueueProcessor"));
 
@@ -167,7 +167,7 @@ public class OpennmsKafkaProducer implements AlarmLifecycleListener, EventListen
             LOG.info("Defaulted the 'kafkaSendQueueCapacity' to 1000 since no property was set");
         }
         
-        kafkaSendQueue = new LinkedBlockingQueue<>(kafkaSendQueueCapacity);
+        kafkaSendQueue = new LinkedBlockingDeque<>(kafkaSendQueueCapacity);
         kafkaSendQueueExecutor.execute(this::processKafkaSendQueue);
 
         if (forwardEvents) {
@@ -409,9 +409,14 @@ public class OpennmsKafkaProducer implements AlarmLifecycleListener, EventListen
                     producer.send(producerRecord, (recordMetadata, e) -> {
                         if (e != null) {
                             LOG.warn("Failed to send record to producer: {}.", producerRecord, e);
-                            if(e instanceof TimeoutException) {
-                                // If Kafka is Offline, buffer the record again.
-                                kafkaSendQueue.offer(kafkaRecord);
+                            if (e instanceof TimeoutException) {
+                                // If Kafka is Offline, buffer the record again for events.
+                                // This is best effort to keep the order although in-flight elements may still miss the order.
+                                if (kafkaRecord.producerRecord != null &&
+                                        kafkaRecord.producerRecord.topic() != null &&
+                                        this.eventTopic.equalsIgnoreCase(kafkaRecord.producerRecord.topic())) {
+                                    kafkaSendQueue.offerFirst(kafkaRecord);
+                                }
                             }
                             return;
                         }
--- a/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
+++ b/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
@@ -104,7 +104,7 @@ public class OpennmsKafkaProducer implements AlarmLifecycleListener, EventListen
     private String alarmFeedbackTopic;
     private String topologyVertexTopic;
     private String topologyEdgeTopic;
-        
+
     private boolean forwardEvents;
     private boolean forwardAlarms;
     private boolean forwardAlarmFeedback;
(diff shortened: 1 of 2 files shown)


features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
```diff
@@ -408,6 +409,10 @@ private void processKafkaSendQueue() {
                     producer.send(producerRecord, (recordMetadata, e) -> {
                         if (e != null) {
                             LOG.warn("Failed to send record to producer: {}.", producerRecord, e);
+                            if(e instanceof TimeoutException) {
```
mattixtech, line 412:
```suggestion
                            if (e instanceof TimeoutException) {
```

mattixtech, line 412: Is it only `TimeoutException` in particular we want to retry for? Any other type of failure we are ok with dropping the message?

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
```diff
@@ -408,6 +409,10 @@ private void processKafkaSendQueue() {
...
+                                // If Kafka is Offline, buffer the record again.
+                                kafkaSendQueue.offer(kafkaRecord);
```
mattixtech, line 414: I'm assuming this ends up reordering the queue as it will be put at the back of the queue right, is that ok for our case?

Also since we are `offer()`ing the message here, we may still lose the message if the queue is full right now, is that ok as well?

features/kafka/producer/src/test/java/org/opennms/features/kafka/producer/KafkaForwarderIT.java
```diff
@@ -420,6 +424,27 @@ public void testSameAlarmIsNotSynced() {
...
+        if (!kafkaProducer.getEventForwardedLatch().await(1, TimeUnit.MINUTES)) {
+            throw new Exception("No events were successfully forwarded in time!");
+        }
+        kafkaConsumer = startConsumer();
+        await().atMost(1, TimeUnit.MINUTES).until(this::getUeisForConsumedEvents, hasItems(
+                EventConstants.NODE_DOWN_EVENT_UEI));
+        kafkaServer.stopKafkaServer();
+        eventdIpcMgr.sendNow(MockEventUtil.createNodeUpEventBuilder("test", databasePopulator.getNode2()).getEvent());
+        // Sleep long enough to account for delivery.timeout.ms = 3 secs.
+        Thread.sleep(10000);
```
mattixtech, line 441: Is there any way we can avoid sleeping for 10 seconds in this test?

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
cgorantla: Callback can get below retrievable Exceptions.

`     *                  CorruptRecordException
     *                  InvalidMetadataException
     *                  NotEnoughReplicasAfterAppendException
     *                  NotEnoughReplicasException
     *                  OffsetOutOfRangeException
     *                  TimeoutException
     *                  UnknownTopicOrPartitionException`
     
Out of these  we get `TimeoutException` when Kafka is offline for short period. That is the only case we want to handle here I think.

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
cgorantla: That's tradeoff I think worth debating. We probably need to make a choice.

features/kafka/producer/src/test/java/org/opennms/features/kafka/producer/KafkaForwarderIT.java
cgorantla: There is nothing to check for. So any other mechanism need to wait for a bit to account for `delivery.timeout.ms`

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
cgorantla: IMO not dropping events/alarms may be better solution. The trade-off is out of order events/alarms. Given that Kafka also doesn't guarantee order, think this is okay since most likely when Kafka gets connected, they will be delivered immediately small difference in delivery time may not matter. 

Also considered having synchronous delivery which would keep the order. But we have dependency on starting AlarmStore based on callback from events/alarms delivery. May not be worth tinkering that.

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
mattixtech: Gotcha, makes sense!

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
cgorantla: @mattixtech  With new change, Now producer sends messages synchronously which would allow us to keep the order.

j-white: Alarms should remain in order and Kafka does guarantee ordering for the same key/partition. The alarm topic was built so that we could consume the topic and build a KTable for the track the state of alarms in external systems - changing the order breaks this.

Re-ordering is okay for other payloads such as events & flows though.

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
```diff
@@ -391,31 +374,41 @@ private void sendRecord(Callable<ProducerRecord<byte[], byte[]>> callable, Consu
...
-                        if (e != null) {
-                            LOG.warn("Failed to send record to producer: {}.", producerRecord, e);
-                            return;
-                        }
-                        if (consumer != null) {
-                            consumer.accept(recordMetadata);
-                        }
-                    });
+                    Future<RecordMetadata> future = producer.send(producerRecord);
+                    future.get();
```
j-white, line 397: This is going to kill performance.

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
cgorantla: Got it.

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
cgorantla: I think there is really no way to keep the performance in async mode and maintain the order.  
May be other ways to mitigate this are having large timeout and having more retries.

mattixtech: Do we need to keep order? That part isn't clear to me based on the previous comments.

cgorantla: I think we need to keep the order at least for Alarms. This change would allow it but would impact performance since we are blocking for every message.

mattixtech: If we need to enforce orderly dispatch from the queue, then yeah we have no choice but to wait until we know the  message didn't timeout (in the case we are doing retry on timeout). That will kill performance as discussed.

Is it sufficient to ignore ordering for timeouts and just send them ASAP instead? We could do that by putting the message at the front of the queue rather than the back. In this case we have best effort ordering while maintaining the original performance characteristics.

cgorantla: I fear best effort may not be sufficient. Given that Alarms are sent with with `reductionKey` as key,  reversing the order for null alarm ( deleted) / regular alarm may be bad for maintaining correct state at consumer side.

mattixtech: Yep, thats a good point. I don't think we can have our cake and eat it too in this case.

We can try other strategies, but I can't think of one off top of my head that is optimal.

Since we get a callback on ack, we can keep a separate data structure of ordered messages in flight and remove them on ack. With this, we can now track state while sending and upon receiving a timeout we can switch to sending sync. Now we know which in flight messages were unack'd and could attempt to wait until we flush those all out and receive acks before we transition back into the original async state. This is an attempt to keep us running async in steady state but switching to sync in failure scenario. I don't think it can guarantee ordering for all messages though.

I think this comes with additional problems, dealing with transient timeouts would be a huge pain. You would still fail to guarantee ordering in the case you sent `A` -> `B` -> `C` where `B` timed out but `A` and `C` did not, since you wouldn't know about B's timeout right away.

---

Do we care about the performance of emptying this queue? Maybe it is ok to do it sync?

cgorantla: Yeah, I m thinking may be just for Alarms, we can do sync ( that actually allows it more linear delivery) if we think Alarms may not be that high volume.

cgorantla: Chatted with @j-white on this. Since nodes/alarms will be consistent after a while since they will be synced from DB again even if they are timed out.

Only for events, we can re-insert messages that are timed out. I will make that change.

agalue:
I believe it worth mentioning on the docs, especially after that last comment, that if someone wants to implement a custom integration based on the Kafka Producer's output, a simple consumer application should not be used for alarm (and node) entities. Instead, you'd need a Kafka Stream application based on KTable so you can have the benefits of the behavior and avoid unwanted side effects. For events, metrics, and other entities, a simple consumer is allowed.

I mean, the docs say, "Publishing records in this fashion allows the topic to be used as a KTable.", but I think it is not that you "could" use KTable, is that you "should" use KTable; otherwise, you might be getting out of order alarms if Kafka is temporarily unreachable (and please correct me if I'm wrong).

Those changes can be added in another PR, but I wanted to share my thoughts.

cgorantla:
> I believe it worth mentioning on the docs, especially after that last comment, that if someone wants to implement a custom integration based on the Kafka Producer's output, a simple consumer application should not be used for alarm (and node) entities. Instead, you'd need a Kafka Stream application based on KTable so you can have the benefits of the behavior and avoid unwanted side effects. For events, metrics, and other entities, a simple consumer is allowed.
> 
> I mean, the docs say, "Publishing records in this fashion allows the topic to be used as a KTable.", but I think it is not that you "could" use KTable, is that you "should" use KTable; otherwise, you might be getting out of order alarms if Kafka is temporarily unreachable (and please correct me if I'm wrong).
> 
> Those changes can be added in another PR, but I wanted to share my thoughts.

As far as I understand,  KafkaAlarmDataSync that  synchronizes alarms there by nodes with DB is always there by default if forwarding of alarms is enabled. There is no extra need for consumers to do this by themselves.

Note that we are optimizing this for a case where Kafka is not available, there is good chance that even `kafkaSendQueueCapacity` is not sufficient in that case. 

Ideally we should depend on Kafka producer timeouts to not miss any messages.  Increasing `delivery.timeout.ms` and `retries`  would be first defense.

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
cgorantla: @mattixtech Can you review this again. When the message times out, we add element to the beginning of queue just for events.

mattixtech: Will take a look

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
```diff
@@ -408,6 +409,15 @@ private void processKafkaSendQueue() {
...
                         if (e != null) {
                             LOG.warn("Failed to send record to producer: {}.", producerRecord, e);
+                            if (e instanceof TimeoutException) {
+                                // If Kafka is Offline, buffer the record again for events.
+                                // This is best effort to keep the order although in-flight elements may still miss the order.
+                                if (kafkaRecord.producerRecord != null &&
+                                        kafkaRecord.producerRecord.topic() != null &&
+                                        this.eventTopic.equalsIgnoreCase(kafkaRecord.producerRecord.topic())) {
+                                    kafkaSendQueue.offerFirst(kafkaRecord);
+                                }
```
mattixtech, line 419:
```suggestion
                                if (producerRecord != null && this.eventTopic.equalsIgnoreCase(producerRecord.topic())) {
                                    kafkaSendQueue.offerFirst(kafkaRecord);
                                }
```

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
```diff
@@ -408,6 +409,15 @@ private void processKafkaSendQueue() {
                     producer.send(producerRecord, (recordMetadata, e) -> {
                         if (e != null) {
                             LOG.warn("Failed to send record to producer: {}.", producerRecord, e);
+                            if (e instanceof TimeoutException) {
+                                // If Kafka is Offline, buffer the record again for events.
+                                // This is best effort to keep the order although in-flight elements may still miss the order.
+                                if (kafkaRecord.producerRecord != null &&
+                                        kafkaRecord.producerRecord.topic() != null &&
+                                        this.eventTopic.equalsIgnoreCase(kafkaRecord.producerRecord.topic())) {
+                                    kafkaSendQueue.offerFirst(kafkaRecord);
```
mattixtech, line 418: Should we log in the case the offer fails?

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
```diff
@@ -408,6 +409,15 @@ private void processKafkaSendQueue() {
                     producer.send(producerRecord, (recordMetadata, e) -> {
```
mattixtech, line 409: Do we need to worry about the exceptions that `send()` itself might throw (caught below)? Or are we only concerned about the exceptions that will be given to us in the callback.

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
mattixtech, line 418: This is a blocking call, this may be a non-issue but it might not be desirable to block inside the callback depending on how the caller is implemented.

mattixtech: I should clarify, its blocking while acquiring the lock to the queue which looks like it isn't fair. I suspect it is probably fine though, maybe in practice that will always be quick.

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
cgorantla: Most of them are not recoverable exceptions and all of them are extended from RuntimeException so I would assume that's okay.

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
cgorantla: The callback is just for ack. Don't think caller will have any dependency on it. 
From javadoc `offer` : Inserts the specified element into this queue if it is possible to do so immediately without violating capacity restrictions.
I think it should be similar for `offerFIrst`. It shouldn't wait for any resources to be available.

mattixtech: My point was just that operations on the blockingqueue have to take the lock so they are blocking by nature. Have to lock the queue to check if there is capacity and then to insert. Its probably not an issue here I've just seen it be a problem before calling blocking operations in a callback depending on what thread it ends up executing on.

The queue is also not fair so it could block for a while potentially. Just something to keep in mind.

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java
```diff
@@ -408,6 +409,16 @@ private void processKafkaSendQueue() {
...
                         if (e != null) {
                             LOG.warn("Failed to send record to producer: {}.", producerRecord, e);
+                            if (e instanceof TimeoutException) {
+                                // If Kafka is Offline, buffer the record again for events.
+                                // This is best effort to keep the order although in-flight elements may still miss the order.
+                                if (producerRecord != null &&
+                                        this.eventTopic.equalsIgnoreCase(kafkaRecord.producerRecord.topic())) {
```
mattixtech, line 416:
```suggestion
                                        this.eventTopic.equalsIgnoreCase(producerRecord.topic())) {
```

mattixtech (approved):
Looks ok. There is some formatting churn that probably doesn't need to be in here, can clean that up if able.

So I guess in the end what we did was for failed messages, best-effort put them first in line in the queue and send them but only if they are on the event topic right?

cgorantla:
> Looks ok. There is some formatting churn that probably doesn't need to be in here, can clean that up if able.
> 
> So I guess in the end what we did was for failed messages, best-effort put them first in line in the queue and send them but only if they are on the event topic right?

That's right.  We are mostly concerned about Alarms/Events. For alarms, there is already mechanism to synchronize with DB. For events, we try to put them back in queue if they are timed out.

PR #3318, opened 2021-02-24
NMS-13174: Use perl from environment

fooker:
Do not use a hard-coded path to perl (either specified during build time
or by using a /usr/bin/perl) but query the environment for the perl
binary to use.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13174


RangerRick (approved):
fine with me

RangerRick:
Ahh, but how do we use `env` from the environment?!  Riddle me that!  ;)

fooker:
> Ahh, but how do we use `env` from the environment?! Riddle me that! ;)

Would love to have something that does that. But we have to rely on standards for this one ;-)

PR #3320, opened 2021-02-24
NMS-13125: user ID and group ID must not contain any HTML markup

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-13125


RangerRick (commented):
It doesn't appear you ended up using any of the `import org.opennms.core.utils.WebSecurityUtils` bits that were added, is that right?

christianpape:
> It doesn't appear you ended up using any of the `import org.opennms.core.utils.WebSecurityUtils` bits that were added, is that right?

You're right. I removed these unused imports.

RangerRick (approved):
LGTM, I like that you're doing the validation in both the UI _and_ the backend.

RangerRick:
I wonder if we should backport this to earlier foundation branches though, seems like this stuff is likely relevant there too.

PR #3324, opened 2021-03-02
NMS-13123: Prevent HTML markup in notification wizard

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-13123

### DIFF
--- a/opennms-webapp/src/main/webapp/WEB-INF/jsp/admin/notification/noticeWizard/eventNotices.jsp
+++ b/opennms-webapp/src/main/webapp/WEB-INF/jsp/admin/notification/noticeWizard/eventNotices.jsp
@@ -36,6 +36,7 @@
 %>
 <%@ taglib uri="http://java.sun.com/jsp/jstl/core" prefix="c"%>
 <%@ taglib uri="http://java.sun.com/jsp/jstl/functions" prefix="fn"%>
+<%@ taglib uri="https://www.owasp.org/index.php/OWASP_Java_Encoder_Project" prefix="e"%>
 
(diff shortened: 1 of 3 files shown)


christianpape:
I escaped the strings for the forms and buttons. The question is: should I still prevent the user from submitting the form with HTML/JS characters in the name?

opennms-webapp/src/main/webapp/WEB-INF/jsp/admin/notification/noticeWizard/eventNotices.jsp
```diff
@@ -109,31 +109,31 @@
           <c:forEach items="${notifications}" var="notification">
           <tr>
             <td>
-              <input type="button" class="btn btn-secondary" value="Edit" onclick="javascript:editNotice('${notification.escapedName}')"/>
+              <input type="button" class="btn btn-secondary" value="Edit" onclick="javascript:editNotice('${fn:escapeXml(notification.escapedName)}')"/>
```
j-white, line 112: Escaping for JavaScript vs XML would be safer here. See https://github.com/OpenNMS/opennms/pull/1763/files#diff-22f9d3c942102d20d541afa9a9aa46ea3a31f641449162efa61ad774684f33e4R101

j-white (commented):
I don't think preventing submissions with special characters is necessary if everything is escaped property on render.

If we do enforce validation rules, then other endpoints like REST APIs should enforce those as well.

j-white:
Patch looks good. Should we backport this to `foundation-2020` so that there is a Meridian release with the patch?

PR #3325, opened 2021-03-02
NMS-13185: Add Interface instance as String attribute

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13185

### DIFF
--- a/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/collection/CollectionSetMapper.java
+++ b/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/collection/CollectionSetMapper.java
@@ -97,6 +97,14 @@ public class CollectionSetMapper {
                             // Skip elements without instance ( aliased resources will be skipped)
                             collectionSetResourceBuilder.setInterface(interfaceResourceBuilder);
                         }
+                        // Add interface label as a String attribute.
+                        if (!Strings.isNullOrEmpty(resource.getInterfaceLabel())) {
+                            CollectionSetProtos.StringAttribute.Builder attributeBuilder = CollectionSetProtos.StringAttribute
+                                    .newBuilder();
+                            attributeBuilder.setName("label");
+                            attributeBuilder.setValue(resource.getInterfaceLabel());
+                            collectionSetResourceBuilder.addString(attributeBuilder);
+                        }
                     }
                 } else if (resource.getResourceTypeName().equals(CollectionResource.RESOURCE_TYPE_LATENCY)) {
                     CollectionSetProtos.ResponseTimeResource.Builder responseTimeResource = buildResponseTimeResource(
--- a/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/collection/CollectionSetMapper.java
+++ b/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/collection/CollectionSetMapper.java
@@ -92,19 +92,15 @@ public class CollectionSetMapper {
                         CollectionSetProtos.NodeLevelResource.Builder nodeResourceBuilder = buildNodeLevelResourceForProto(
                                 nodeCriteria);
                         interfaceResourceBuilder.setNode(nodeResourceBuilder);
+                        interfaceResourceBuilder.setInstance(resource.getInterfaceLabel());
                         if (!Strings.isNullOrEmpty(resource.getInstance())) {
-                            interfaceResourceBuilder.setInstance(resource.getInstance());
-                            // Skip elements without instance ( aliased resources will be skipped)
-                            collectionSetResourceBuilder.setInterface(interfaceResourceBuilder);
-                        }
-                        // Add interface label as a String attribute.
-                        if (!Strings.isNullOrEmpty(resource.getInterfaceLabel())) {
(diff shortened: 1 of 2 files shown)


cgorantla:
Don't remember why we used `instanceLabel` before but I think `instance` seems to be right one as it gives `ifIndex` as instance for Snmp Interfaces.
Upgrading on `meridian-2019` might give different instance name so it may look like breaking change. May need to add it to release notes.

agalue:
I can see your point for Meridian 2019, but the customer who triggered this effort expects the solution on Meridian 2020 (they don't use Horizon).

Would this be possible to retarget this to `foundation-2020`?

cgorantla:
> I can see your point for Meridian 2019, but the customer who triggered this effort expects the solution on Meridian 2020 (they don't use Horizon).
> 
> Would this be possible to retarget this to `foundation-2020`?

Since the issue exists in `foundation-2019` as well I think this may be right branch. Do you see any issues targeting this to `foundation-2019` ?

agalue:
Not at all. Having this fixed in `foundation-2019` makes total sense; perhaps I was confused by your comment related to M2019.

agalue (approved):

agalue (commented):
It is not clear what to expect on M2019, M2020, or H27, so once I understand how the solution will work on each of those versions, I'll approve this PR (so I can tell the customer what to do with their code).

RangerRick (changes requested):
It seems like changing this in 2019 is essentially changing a "public" API in an incompatible way.  Is there a way to provide this extra metadata for customers without changing existing instance entries?

RangerRick:
(To be clear, if it makes sense to do it this way long-term, that's great -- we can do that in `develop` for Horizon 28 -- but it seems like a bad idea to change this in 2019 IMHO if people are relying on the Kafka producer as a stable interface.)

agalue:
I understand the reasoning behind changing a "stable" or public API. I added the term between quotes as the "current" context is incorrect. Yet, I understand your point.

The problem I see with adding the missing content as "extra" string properties is how can we guarantee that if someone defined a MibObj as a string and use that same name as its alias, that won't override what we're going to enforce as a "workaround" (as I cannot consider it a solution) to deal with the incorrect or missing information (i.e., the ifIndex in this case).

Here is an example. The customer who has opened a support ticket and triggered this effort has 10 OpenNMS servers pushing metrics only to Kafka (different topic names per instance). They developed a solution based on our recommendations precisely because the current solution is incorrect. That solution involves forcing the ifIndex as a string via mibObj, using the alias "Index" for it (the same name proposed in this PR). Unfortunately, for some reason, that solution is not reliable (hence the support ticket). They see multiple interfaces (not all of them) from the same node (i.e., the same collection-set) with a missing Index, something that in theory should not happen. That is why I thought that "fixing" the Protobuf content is easier and faster to fix the problem they have in their production environment.

If we go the way you suggest, I think we should use a name that a user would never collide like "__index" or something like that. Or, clearly define and explain what to do if there is already a string property with the same name's going to inject (which might involve writing some warning to a log).

The customer would still need to reimplement their custom code if they decide to upgrade to M2021 in the future, but I think there is no way to avoid that. Well, there is, but that contradicts what you propose.

agalue:
So, to avoid confusion, considering that my expectations are different from what we might end up having for M2019, M2020, H27, and develop, I'd need to wait until the "final" changes, test each version independently, and understand the behavior on each of them. Then I can inform the customer about how to use it, so they know what they need to change now in their code and what they would have to change soon when they upgrade in the future.

That's why I removed my acceptance from the review.

cgorantla:
@RangerRick @agalue I have opened https://github.com/OpenNMS/opennms/pull/3328 to make the breaking change for Interface Instance.

RangerRick:
If it is this important to be actually changed in a customer environment in M2019, then maybe we leave the implementation like this, but hide the new behavior behind a property and have it off by default. Then we can just make this the default in H28+/M2022 under a big "breaking change" banner.

agalue:
Another reason why I'd like to see detailed documentation about how to use the "fix" on all the mentioned versions so I can see them working on my own, which is what I need before contacting the customer.

agalue:
It is always hard for me to understand why we need a configurable option to unlock the correct behavior and leave the wrong one mainstream. As usual, I'll adapt to whatever is decided, even if I 100% disagree with it.

RangerRick:
> It is always hard for me to understand why we need a configurable option to unlock the correct behavior and leave the wrong one mainstream. As usual, I'll adapt to whatever is decided, even if I 100% disagree with it.

It's always hard to know how all of our users (not just customers) are using these things, so it's impossible to say whether we break someone by changing it. If it's an API we publish, it's an API, and we shouldn't be changing it unless we know that it was never useful to someone as it was.

cgorantla:
Modified String attribute name to `__ifIndex` based on feedback from @agalue to make it really unique.

agalue:
Once everything goes green, and the Docker Image is available, I'll test this "workaround" against my test device to verify that the solution does what it says it does.

RangerRick (approved):

agalue (approved):
I tested it, and it works as expected. The `__ifIndex` string property is present and has the correct value for all the Interface Resources.

PR #3326, opened 2021-03-02
Update schema types in minion configuration schema

mattixtech:
- Remove set
- Add integer/decimal (not in use yet)

This PR has no functional impact on OpenNMS. Purely a schema update for minion configuration.

Once this PR is merged, I will manually propagate the schema change to the dev cloud to make it live in dev.


mattixtech:
@mikewrosey the plan is to remove `set` and just use `collection` for now as I don't think we need the added complexity of enforcing set uniqueness right now. Also this adds `integer` and `decimal` for which we don't have front end types for. However they aren't being actively used in the schema yet so there is no pressing need to implement them in the FE at this time.

bouff (approved):

mikewrosey:
> @mikewrosey the plan is to remove `set` and just use `collection` for now as I don't think we need the added complexity of enforcing set uniqueness right now. Also this adds `integer` and `decimal` for which we don't have front end types for. However they aren't being actively used in the schema yet so there is no pressing need to implement them in the FE at this time.

Sounds good. I'll create an FE task under FP Enhancements for the new types.

mikewrosey (approved):

PR #3328, opened 2021-03-03
NMS-13185: Add ifIndex for InterfaceResource

cgorantla:
Also ignore resources without instance for InterfaceResource.

This PR will correct the Interface Instance and add InterfaceLabel as String attribute.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13185

### DIFF
--- a/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/model/CollectionSetProtos.java
+++ b/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/model/CollectionSetProtos.java
@@ -1,8 +1,8 @@
 /*******************************************************************************
  * This file is part of OpenNMS(R).
  *
- * Copyright (C) 2020 The OpenNMS Group, Inc.
- * OpenNMS(R) is Copyright (C) 1999-2020 The OpenNMS Group, Inc.
+ * Copyright (C) 2021 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2021 The OpenNMS Group, Inc.
  *
  * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
  *
@@ -2727,7 +2727,7 @@ public final class CollectionSetProtos {
         getInstanceBytes();
 
     /**
-     * <code>int32 ifIndex = 3;</code>
+     * <code>int32 if_index = 3;</code>
      */
     int getIfIndex();
   }
@@ -2883,10 +2883,10 @@ public final class CollectionSetProtos {
       }
     }
 
-    public static final int IFINDEX_FIELD_NUMBER = 3;
+    public static final int IF_INDEX_FIELD_NUMBER = 3;
     private int ifIndex_;
     /**
-     * <code>int32 ifIndex = 3;</code>
+     * <code>int32 if_index = 3;</code>
      */
     public int getIfIndex() {
       return ifIndex_;
@@ -2977,7 +2977,7 @@ public final class CollectionSetProtos {
       }
       hash = (37 * hash) + INSTANCE_FIELD_NUMBER;
       hash = (53 * hash) + getInstance().hashCode();
-      hash = (37 * hash) + IFINDEX_FIELD_NUMBER;
+      hash = (37 * hash) + IF_INDEX_FIELD_NUMBER;
       hash = (53 * hash) + getIfIndex();
       hash = (29 * hash) + unknownFields.hashCode();
       memoizedHashCode = hash;
@@ -3430,13 +3430,13 @@ public final class CollectionSetProtos {
 
       private int ifIndex_ ;
       /**
-       * <code>int32 ifIndex = 3;</code>
+       * <code>int32 if_index = 3;</code>
        */
       public int getIfIndex() {
         return ifIndex_;
       }
       /**
-       * <code>int32 ifIndex = 3;</code>
+       * <code>int32 if_index = 3;</code>
        */
       public Builder setIfIndex(int value) {
 
@@ -3445,7 +3445,7 @@ public final class CollectionSetProtos {
         return this;
       }
       /**
-       * <code>int32 ifIndex = 3;</code>
+       * <code>int32 if_index = 3;</code>
        */
       public Builder clearIfIndex() {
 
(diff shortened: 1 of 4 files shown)


agalue:
Based on what we discussed, the changes here are different than the workaround for Meridian.

Also, I'm assuming that M2021 will be based on H28. If that is not going to be the case, the solution here should be ported to M2021 once we have the foundation-2021 branch.

The idea is to modify the Protobuf for [InterfaceLevelResource](https://github.com/OpenNMS/opennms/blob/develop/features/kafka/producer/src/main/proto/collectionset.proto#L29) to:

1) Add the `ifIndex` for the existing "instance" field (i.e., the output of the `getInstance()` method). If you ask me, I would change the type to be an integer instead of a string, now that newer versions can have different APIs, and the ifIndex is NOT as a string.

2) Introduce a new field called "label" with what we currently add as "instance" (i.e., the output of the `getInterfaceLabel()` method).

3) Remove the "workaround" from Meridian's solution, as the above two should be top-level attributes and not string attributes which are elements that a given user can override. That means, if we merge the Meridian solution first, this PR should take that into account and remove that solution (as it won't be needed).

As we discussed, we're going to guarantee that the aliased resources would never appear in the Collection Set sent to Kafka as those are mere duplicates of the data already provided by the Interface Resources (reducing the size of the Protobuf payload).

RangerRick:
M2021 will be based on H27.1.0, but you're right, we should port it over when 2021 is released.

cgorantla:
> Based on what we discussed, the changes here are different than the workaround for Meridian.
> 
> Also, I'm assuming that M2021 will be based on H28. If that is not going to be the case, the solution here should be ported to M2021 once we have the foundation-2021 branch.
> 
> The idea is to modify the Protobuf for [InterfaceLevelResource](https://github.com/OpenNMS/opennms/blob/develop/features/kafka/producer/src/main/proto/collectionset.proto#L29) to:
> 
> 1. Add the `ifIndex` for the existing "instance" field (i.e., the output of the `getInstance()` method). If you ask me, I would change the type to be an integer instead of a string, now that newer versions can have different APIs, and the ifIndex is NOT as a string.
> 2. Introduce a new field called "label" with what we currently add as "instance" (i.e., the output of the `getInterfaceLabel()` method).
> 3. Remove the "workaround" from Meridian's solution, as the above two should be top-level attributes and not string attributes which are elements that a given user can override. That means, if we merge the Meridian solution first, this PR should take that into account and remove that solution (as it won't be needed).
> 
> As we discussed, we're going to guarantee that the aliased resources would never appear in the Collection Set sent to Kafka as those are mere duplicates of the data already provided by the Interface Resources (reducing the size of the Protobuf payload).

All sounds right according to our discussion except for making instance as integer. I want to keep it as String as the instance that we are getting from CollectionResource seems to be String. 
For ex: [`PerspectiveResponseTimeResource`](https://github.com/OpenNMS/opennms/blob/3cb6ddce8fbf67356b69eda0f5e575501798e877/features/collection/api/src/main/java/org/opennms/netmgt/collection/support/builder/PerspectiveResponseTimeResource.java#L61) is a `RESOURCE_TYPE_IF` but it's instance is not an Integer.

agalue:
All the things that inherit from `CollectionResource` return a String, for instance, because that's an interface. That doesn't mean newer independent APIs like a Protobuf definition for things that are not related to each other (i.e., `InterfaceLevelResource`, vs. `PerspectiveResponseTimeResoure`, vs. `GenericInddexResource`, and so on) have to follow the same convention. An ifIndex is not a string; why forcing the users to convert it to an Integer instead of getting it as it should?

agalue:
Let's take the opportunity to make the Protobuf content for Collection Sets more useful and consistent with other APIs, so a given user can consume its information without the need to get content from other APIs to use it.

We spoke about not changing APIs for current versions to avoid breaking the solution that users might be currently using. In that regard, Protobuf is perfect, as an old version of a ".proto" file can parse content from a newer ".proto" (even from different versions, and vice versa) if the definition is consistent.

With that in mind, we can "keep" the implementation of the "instance" field as it is and mark it as deprecated on the newer versions of OpenNMS. That way, there is no "breaking" change to describe in the changelog.

Then, we can introduce new fields to ALL the resource entities in the ".proto" files to make them more usable and independent. Those fields are:

* id: The Resource ID, for instance node[Cisco:router1].interfaceSnmp[fa0_0]. This is required when using the Resources end-point or the Measurements API through ReST.
* name: The unique identifier used to replace the raw index which is what we use within the Resource ID. For instance, fa0_0. That'd be the name of the directory or part where the RRD for the metrics live (or the Newts Identifier).
* label: The resource label, or the human-readable content we see in the UI (which comes from the resource-type definition and what the resources end-point will show). For instance, Fa0/0.
* index: The raw identifier of a row from a tabular data like an SNMP Table (this would be the ifIndex for Interface Resources or the OID that represents the index of a table, and so on).
* type: The resource type identifier (or the resource type name). This would only apply to generic index resources, and it is already part of that entity in the ".proto" file.
* typeLabel: The label of the resource type (optional but useful, as the resources end-point provides it).

All the above fits perfectly for Interface Resources and Generic Index Resources from the SNMP perspective. I know that some of those entries might not apply to other protocols, but I believe it worth, for consistency, having all of them.

If we go this route, we would have a backward-compatible API (as instance won't be changed but marked as deprecated), and we will introduce new fields with all the necessary content to make the ".proto" content self-contained and independent; meaning a given user wouldn't have to look other sources to consume it and use it. That's because it was precisely the lack of why the old TCP Exporter was never useful, and I believe we should avoid that same mistake on the Kafka Producer.

cgorantla:
> All the things that inherit from `CollectionResource` return a String, for instance, because that's an interface. That doesn't mean newer independent APIs like a Protobuf definition for things that are not related to each other (i.e., `InterfaceLevelResource`, vs. `PerspectiveResponseTimeResoure`, vs. `GenericInddexResource`, and so on) have to follow the same convention. An ifIndex is not a string; why forcing the users to convert it to an Integer instead of getting it as it should?

Looking at [InterfaceLevelResource](https://github.com/OpenNMS/opennms/blob/1afa92d1f6cbba7a81ed855f4d39b009ecc7c5d7/features/collection/api/src/main/java/org/opennms/netmgt/collection/support/builder/InterfaceLevelResource.java#L60) it seems like `ifName` is used as both instance/label.   So I have kept the instance same as before and added `ifIndex` as Integer.  That should solve the current issue.
Also will skip resources without instance ( Aliased Resources)

We can incorporate further changes in new issue.

agalue:
> We can incorporate further changes in new issue.

https://issues.opennms.org/browse/NMS-13191

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/model/CollectionSetProtos.java
```diff
@@ -1,8 +1,8 @@
 /*******************************************************************************
  * This file is part of OpenNMS(R).
  *
- * Copyright (C) 2018 The OpenNMS Group, Inc.
- * OpenNMS(R) is Copyright (C) 1999-2018 The OpenNMS Group, Inc.
+ * Copyright (C) 2021 The OpenNMS Group, Inc.
```
RangerRick, line 4: not to nitpick but this should be 2018-2021 then  :D

RangerRick (approved):
LGTM, although it would have been nice to have that reformat done as a separate commit.  Not a huge deal though.

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/model/CollectionSetProtos.java
cgorantla: hmm, I didn't know that. I'm always using 1999-current year for all files.

RangerRick: Well, it generally should be the first time the file was made through the last time it was edited.  So in _this_ case it started at 2018 (presumably)

cgorantla: Okay.

PR #3331, opened 2021-03-05
fix NPE in linkd tests

RangerRick:
This PR fixes some NPEs we saw in CircleCI tests (example [here](https://app.circleci.com/pipelines/github/OpenNMS/opennms/9359/workflows/f706b6af-27ae-41f3-83a1-34e9f26bf620/jobs/51879))

### DIFF
--- a/opennms-services/src/test/java/org/opennms/netmgt/nb/Nms0001NetworkBuilder.java
+++ b/opennms-services/src/test/java/org/opennms/netmgt/nb/Nms0001NetworkBuilder.java
@@ -1,8 +1,8 @@
 /*******************************************************************************
  * This file is part of OpenNMS(R).
  *
- * Copyright (C) 2013-2014 The OpenNMS Group, Inc.
- * OpenNMS(R) is Copyright (C) 1999-2014 The OpenNMS Group, Inc.
+ * Copyright (C) 2013-2021 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2021 The OpenNMS Group, Inc.
  *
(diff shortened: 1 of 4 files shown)


RangerRick:
I honestly don't understand how this ever worked.

opennms-services/src/test/java/org/opennms/netmgt/nb/Nms0001NetworkBuilder.java
```diff
@@ -52,10 +52,10 @@
     FROH_IP_IF_MAP.put(InetAddressUtils.addr("10.0.1.2"), 606);
     FROH_IP_IF_MAP.put(InetAddressUtils.addr("66.3.9.1"), 613);
     FROH_IP_IF_MAP.put(InetAddressUtils.addr("10.246.16.245"), 608);
-    FROH_IP_IF_MAP.put(InetAddressUtils.addr("66-3-123-1.la-04.cvx.algx.net"), 668);
```
cgorantla, line 55: Is it just that these IPs are no longer accessible and throwing exception in static block that would make all other elements to be not added ?

cgorantla (approved):
LGTM

opennms-services/src/test/java/org/opennms/netmgt/nb/Nms0001NetworkBuilder.java
RangerRick: Yeah, they're not IPs at all, and don't parse.  It honestly makes no sense to me why this wasn't failing for all of time. The _entire_ block of map-filling doesn't happen.

rssntn67:
I wrote some notes on documentation.

PR #3333, opened 2021-03-09
NMS-13124: Fixed user deletion by renaming bug and CSRF privilege escalation issue

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-13124


j-white:
Patch looks good. Should we backport this to `foundation-2020` so that there is a Meridian release with the patch?

christianpape:
> Patch looks good. Should we backport this to `foundation-2020` so that there is a Meridian release with the patch?

Yes. Sure. I'll cherrypick this one.

PR #3336, opened 2021-03-10
NMS-13120: Make threshold state key unique for different threshold levels

cgorantla:
When there are multiple levels of thresholds for a given DS, the key that is persisted to maintain threshold state is not unique enough.
To solve this, generate stable hash for threshold values and add it to key.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13120

### DIFF
--- a/features/collection/thresholding/impl/src/main/java/org/opennms/netmgt/threshd/AbstractThresholdEvaluatorState.java
+++ b/features/collection/thresholding/impl/src/main/java/org/opennms/netmgt/threshd/AbstractThresholdEvaluatorState.java
@@ -31,6 +31,7 @@ package org.opennms.netmgt.threshd;
 import static org.opennms.core.utils.InetAddressUtils.addr;
 
 import java.io.Serializable;
+import java.nio.charset.StandardCharsets;
 import java.text.DecimalFormat;
 import java.util.Date;
 import java.util.Map;
@@ -56,6 +57,8 @@ import org.slf4j.LoggerFactory;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
+import com.google.common.hash.Hasher;
+import com.google.common.hash.Hashing;
 import com.swrve.ratelimitedlogger.RateLimitedLog;
 
 /**
@@ -98,14 +101,16 @@ public abstract class AbstractThresholdEvaluatorState<T extends AbstractThreshol
     protected final ThresholdingSession thresholdingSession;
     
     static final String THRESHOLDING_KV_CONTEXT = "thresholding";
-    
-    private final int stateTTL;
+
+    private final int stateTTL = SystemProperties.getInteger("org.opennms.netmgt.threshd.state_ttl",
+            (int) TimeUnit.SECONDS.convert(24, TimeUnit.HOURS));
     
     private Long sequenceNumber;
 
     private boolean firstEvaluation = true;
     
     private String instance;
+
 
     private static final Map<Class<? extends AbstractThresholdEvaluatorState.AbstractState>,
             SerializingBlobStore<? extends AbstractThresholdEvaluatorState.AbstractState>> serdesMap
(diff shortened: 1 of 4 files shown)


j-white (commented):
Nice work - this should do the trick.

I think we'll want to retarget this to `foundation-2020`.

dino2gnt:
What about cases where users _aren't_ using distinct UEIs?  Is it possible to include something like, for example, the threshold value itself as part of the state key to keep these distinct independent of the uei (or something else unique to a threshold)?

What about situations where a threshold is defined with the same UEI but different values in different thresholding packages? Is the state key still unique in that case?

features/collection/thresholding/impl/src/main/java/org/opennms/netmgt/threshd/AbstractThresholdEvaluatorState.java
```diff
@@ -173,7 +173,11 @@ public String toString() {
         key = String.format("%d-%s-%s-%s-%s-%s", thresholdingSession.getKey().getNodeId(),
                 thresholdingSession.getKey().getLocation(), threshold.getDsType(),
                 threshold.getDatasourceExpression(), thresholdingSession.getKey().getResource(), threshold.getType());
-
+        // Multiple threshold levels for trigger/rearm may end up with the same key.
+        // Fair to assume that threshold definitions with different threshold levels will have different uei.
+        if(threshold.getTriggeredUEI().isPresent()) {
+            key = String.format("%s-%s", key, threshold.getTriggeredUEI().get());
```
mattixtech, line 179: Is it likely these `EvaluatorState` objects get constructed frequently? If so we could potentially optimize the key to do only one String allocation instead of two. We could also cache the TTL below.

It may not be worth doing, just pointing it out in case.

mattixtech (approved):
Thanks for cleaning up after me! 🚀

cgorantla:
@mattixtech  Modified this to generate hash (non-cryptographic, stable between process/VMs) for threshold values to account for thresholds having same uei but with different levels in values/rearms.

features/collection/thresholding/impl/src/main/java/org/opennms/netmgt/threshd/AbstractThresholdEvaluatorState.java
```diff
@@ -151,15 +156,30 @@ public String toString() {
 
         this.thresholdingSession = thresholdingSession;
         kvStore = getKvStoreForType(stateType, thresholdingSession.getBlobStore());
-        key = String.format("%d-%s-%s-%s-%s-%s", thresholdingSession.getKey().getNodeId(),
+        key = String.format("%d-%s-%s-%s-%s-%s-%d", thresholdingSession.getKey().getNodeId(),
```
cgorantla, line 159: Kept other parts of key same as it may help debugging to have them descriptive.

cgorantla (commented):
Re-targeted this to `foundation-2019`

features/collection/thresholding/impl/src/main/java/org/opennms/netmgt/threshd/AbstractThresholdEvaluatorState.java
mattixtech: Makes sense 👍🏻 . I guess the downside is we won't have a fixed width key like we would with a hash. But maybe having it human readable is more valuable.

features/collection/thresholding/impl/src/main/java/org/opennms/netmgt/threshd/AbstractThresholdEvaluatorState.java
```diff
@@ -151,15 +156,30 @@ public String toString() {
...
 
-        stateTTL = SystemProperties.getInteger("org.opennms.netmgt.threshd.state_ttl",
-                (int) TimeUnit.SECONDS.convert(24, TimeUnit.HOURS));
         initializeState();
     }
 
+    // Multiple threshold levels for trigger/rearm may end up with the same key.
+    // Generate unique hashcode for threshold values.
+    private int generateHashForThresholdState(BaseThresholdDefConfigWrapper threshold) {
+        Hasher hasher = Hashing.murmur3_128().newHasher();
```
mattixtech, line 170: What happens if we didn't `put` anything to the hasher? Does it blow up or handle that gracefully?

mattixtech (changes requested):

features/collection/thresholding/impl/src/main/java/org/opennms/netmgt/threshd/AbstractThresholdEvaluatorState.java
cgorantla: That's good question.  It just returns bytes of 0x00.

features/collection/thresholding/impl/src/main/java/org/opennms/netmgt/threshd/AbstractThresholdEvaluatorState.java
```diff
@@ -151,15 +156,30 @@ public String toString() {
...
+        if (threshold.getTriggeredUEI().isPresent()) {
+            hasher.putString(threshold.getTriggeredUEI().get(), StandardCharsets.UTF_8);
+        }
+        if (threshold.getRearmedUEI().isPresent()) {
+            hasher.putString(threshold.getRearmedUEI().get(), StandardCharsets.UTF_8);
+        }
+        Optional.ofNullable(threshold.getValue()).ifPresent(hasher::putDouble);
+        Optional.ofNullable(threshold.getRearm()).ifPresent(hasher::putDouble);
+        Optional.ofNullable(threshold.getTrigger()).ifPresent(hasher::putInt);
+        return hasher.hash().asInt();
```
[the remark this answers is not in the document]
cgorantla: No specific reason. Since we are dealing with limited set of values, thought smaller would be good. Modified it to String.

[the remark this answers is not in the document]
mattixtech: Sounds good 👍🏻 , if we wanted to we could truncate to save space but unless we specifically wanted that I would just take the full hash.

features/collection/thresholding/impl/src/main/java/org/opennms/netmgt/threshd/AbstractThresholdEvaluatorState.java
mattixtech: Cool, as long as it doesn't blow up we should be ok.

mattixtech (approved):

PR #3337, opened 2021-03-10
NMS-13197: Compile OpenNMS with JDK11

j-white:
JIRA: https://issues.opennms.org/browse/NMS-13197

This patch includes build system and library upgrades to support compiling JDK11.
The tree can no longer be compiled or run with JDK8 once this is merged.

### DIFF
--- a/bin/functions.pl
+++ b/bin/functions.pl
@@ -31,7 +31,6 @@ use vars qw(
 	$TESTS
 	$SINGLE_TEST
 	$VERBOSE
-	$JDK9_OR_GT
 	@DEFAULT_GOALS
 	@ARGS
 );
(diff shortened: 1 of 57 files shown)


RangerRick:
Having not had a chance to look through the patches yet, let me say just 2 words: hell, and yeah.

bin/functions.pl
```diff
@@ -214,7 +216,7 @@ END
 	$ENV{'PATH'}      = File::Spec->catfile($JAVA_HOME, 'bin') . $PATHSEP . $ENV{'PATH'};
 
         my ($shortversion) = get_version_from_java(File::Spec->catfile($JAVA_HOME, 'bin', 'java'));
-        if ($shortversion >= 9) {
+        if ($shortversion >= 9 && $shortversion < 11) {
```
RangerRick, line 219: Doesn't this basically only set the variable if JDK is 9 or 10?  What's the reason for keeping this if minversion=11 anyway?

RangerRick (approved):
Looks pretty good, nothing jumps out at me.

bin/functions.pl
j-white: Good point. Removed.

indigo423 (approved):
LGTM! 👍  Happy you haven't upgraded the docker container from 20.10.1. They [broke](https://github.com/moby/moby/issues/41858) IPv6 in 20.10.2 and might get us in trouble with IPv6 tests. I haven't seen anything suspicious to me in the PR.

PR #3340, opened 2021-03-11
NMS-13149: fix behavior of UEI selection

RangerRick:
This fixes the UEI selection in the notice wizard to:

1. actually reset things when you hit the "Reset" button
2. always update the regular expression field whenever state changes
3. restore the full list of options if you filter the results, and then change the filter back to an empty string
4. fix the table to be wide enough to see the full events when the browser is wider

It was hard to keep myself from just rewriting this using angular or something, but I tried to do the minimal amount of changes necessary to make this work rather than replacing or cleaning up gross jQuery stuff, since it targets Meridian, and we'll probably rewrite it anyway when we start working on redoing the UI.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13149


cgorantla (approved):
LGTM

PR #3342, opened 2021-03-15
NMS-13186: separate smoke test builds

RangerRick:
This doesn't actually solve the issues of NMS-13186, but it helps mitigate them some by speeding up builds and separating things so a rebuild of a failed run doesn't use so many resources.

I will probably backport this to at least as far as `foundation-2019` if it goes well.  I'll do some checking on CircleCI costs first before committing to it anywhere else though.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13186

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -178,9 +178,9 @@ commands:
   run-smoke-tests:
     description: "Run the smoke tests"
     parameters:
-      minimal:
-        default: false
-        type: boolean
+      suite:
+        default: core
+        type: string
     steps:
       - run:
           name: Enable swap
@@ -211,7 +211,7 @@ commands:
           name: Smoke Tests
           no_output_timeout: 30m
           command: |
-            .circleci/scripts/smoke.sh << parameters.minimal >>
+            .circleci/scripts/smoke.sh << parameters.suite >>
       - run:
           name: Gather system logs
           when: always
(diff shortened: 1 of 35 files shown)


cgorantla:
Looks good to me. Would it also help to move recently failing flappers into one specific module like `smoke-test-flappers` not to ignore it but limit the re-runs to a specific module.

j-white (commented):
Logic looks good - annotations make this fairly easy to maintain going forward.

My one concern is that this is going to increase the number of credits used fairly significantly. We're going from 8 runs to 28 - each has about 10 minutes of overhead to get ready - so this adds about 200 minutes per run. Should we scale some of the parallelism levels down to mitigate this?

RangerRick:
For _now_ it probably still saves overall credits given our 50% success rate of the smoke tests currently.  But yeah, I can drop it some to make it not so bad a jump.

Naicisum:
Just a thought, what about using `circleci` to create a `test-cache` on future runs to help mitigate some of the rerun. Cache the results, parse results on subsequent loops and only run failed tests. Even create a flag somewhere to force a full-rerun.

RangerRick:
That's an interesting thought.

Honestly though, it should not _generally_ be an issue, we just have a couple of outstanding flappers that everyone's been too busy to dig into.  :D

I did this basically because it's a quick and easy way to lessen the impact of failures until we can throw some actual engineering at fixing the test flappers, and is an overall win for our build time which has ballooned quite a bit.

RangerRick:
> Logic looks good - annotations make this fairly easy to maintain going forward.

Yeah, my goal was for core to always be "everything else" just in case we forgot to annotate something, turns out the excludedGroups stuff for JUnit handles it.

> My one concern is that this is going to increase the number of credits used fairly significantly. We're going from 8 runs to 28 - each has about 10 minutes of overhead to get ready - so this adds about 200 minutes per run. Should we scale some of the parallelism levels down to mitigate this?

I dropped it to 8/6/6 so the impact shouldn't be _as_ bad.  I still want to do a little calculating on just how many more credits we spend per build this way.

Run times with 8/6/6 are:
* core: 1h45m
* minion: 58m
* sentinel: 42m

Still a little bit better than 2h12m

Thinking I could bump minion/sentinel down to 4 and still not be longer than core, the only fear is that those are the ones more likely to fail and need a reload (for now).

Maybe leave it like this for now, and drop it once we get the flappers handled.

RangerRick:
To @cgorantla's point, I'm a little hesitant to make a `smoke-test-flappers` specifically, since that would just give us license to never fix the problem.

j-white (approved):
LGTM

PR #3343, opened 2021-03-15
NMS-13120: Consider metadata in threshold values.

cgorantla:
Small update to #3336 to take care of metadata in threshold values.
Since metadata support is only added in `27.x` for threshold values this is targetted to `release-27.x`

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13120


RangerRick (approved):

mattixtech (approved):
Just checking, is it always an either or (mutual exclusion) for the "value" vs the "value string" or is there any case where we may want both?

If no, LGTM.

cgorantla:
> Just checking, is it always an either or (mutual exclusion) for the "value" vs the "value string" or is there any case where we may want both?
> 
> If no, LGTM.

Originally all values are Strings when user enters. When they are resolvable to a Number, they are considered actual values. else it's probably metadata.

So it's always either or.

PR #3350, opened 2021-03-22
NMS-13211: Improve Event forwarding performance for Kafka producer

j-white:
JIRA: https://issues.opennms.org/browse/NMS-13211

A couple changes to improve event forwarding performance:
 1) Use a threaded event listener - which allows the map from event -> protobuf to be performed in parallel
 2) Leverage the event by UEI index to speed up retrieving the event labels


cgorantla (approved):
LGTM!

PR #3351, opened 2021-03-22
NMS-13189: Trigger updatedAlarm notification for Ticket state changes

cgorantla:
Currently ticket state changes doesn't trigger any  `AlarmLifeCycleListener` changes needed to insert the updated alarm into drools again.
This results in some bugs in ticketing related workflow in Alarmd drools rules.

Hopefully this PR should fix those bugs.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13189

### DIFF
--- a/features/ticketing/daemon/src/main/java/org/opennms/netmgt/ticketd/DefaultTicketerServiceLayer.java
+++ b/features/ticketing/daemon/src/main/java/org/opennms/netmgt/ticketd/DefaultTicketerServiceLayer.java
@@ -40,6 +40,7 @@ import org.opennms.api.integration.ticketing.RelatedAlarmSummary;
 import org.opennms.api.integration.ticketing.Ticket;
 import org.opennms.api.integration.ticketing.Ticket.State;
 import org.opennms.netmgt.dao.api.AlarmDao;
+import org.opennms.netmgt.dao.api.AlarmEntityNotifier;
 import org.opennms.netmgt.events.api.EventIpcManager;
 import org.opennms.netmgt.events.api.EventIpcManagerFactory;
 import org.opennms.netmgt.model.OnmsAlarm;
@@ -72,6 +73,9 @@ public class DefaultTicketerServiceLayer implements TicketerServiceLayer, Initia
 
     @Autowired
     private AlarmDao m_alarmDao;
+
+    @Autowired
+    private AlarmEntityNotifier m_alarmEntityNotifier;
 
     private Plugin m_ticketerPlugin;
     private EventIpcManager m_eventIpcManager;
@@ -113,7 +117,7 @@ public class DefaultTicketerServiceLayer implements TicketerServiceLayer, Initia
             LOG.error("No alarm with id {} was found. Ticket with id '{}' will not be canceled.", alarmId, ticketId);
             return;
         }
-
+        TroubleTicketState previousState = alarm.getTTicketState();
         try {
             setTicketState(ticketId, Ticket.State.CANCELLED);
             alarm.setTTicketState(TroubleTicketState.CANCELLED);
@@ -125,6 +129,7 @@ public class DefaultTicketerServiceLayer implements TicketerServiceLayer, Initia
 
         m_alarmDao.saveOrUpdate(alarm);
 
+        updateNewTicketState(alarm, previousState);
     }
 
     private void setTicketState(String ticketId, State state) throws PluginException { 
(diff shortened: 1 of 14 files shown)


features/ticketing/daemon/src/main/java/org/opennms/netmgt/ticketd/DefaultTicketerServiceLayer.java
```diff
@@ -333,4 +353,8 @@ public void setEventIpcManager(EventIpcManager ipcManager) {
     public void setTicketerPlugin(Plugin plugin) {
         m_ticketerPlugin = Objects.requireNonNull(plugin, "plugin cannot be null");
     }
+
+    public void setAlarmEntityNotifier(AlarmEntityNotifier alarmEntityNotifier) {
+        this.m_alarmEntityNotifier = alarmEntityNotifier;
```
j-white, line 358: `this.m_`

j-white (approved):
Nice find! Looks good!

opennms-base-assembly/src/main/filtered/etc/alarmd/drools-rules.d/alarmd.drl
```diff
@@ -172,7 +172,6 @@ rule "closeClearedAlarmTickets"
     $sessionClock : SessionClock()
     $ticketer : AlarmTicketerService(isTicketingEnabled() == true)
     $alarm : OnmsAlarm(severity == OnmsSeverity.CLEARED &&
-                       alarmType == OnmsAlarm.PROBLEM_TYPE &&
```
cgorantla, line 175: @j-white  This was needed for UseCase3 where `clearAlarm` should close open ticket after 15 mins. But when clear alarm comes, it seems like it has `OnmsAlarm.RESOLUTION_TYPE`.  I'm not sure why this check was there before. Once I removed this check, ticket gets closed and removed properly.  Let me know if I'm missing anything.

cgorantla:
@agalue I have verified that all the usecase in the issue and they all are taken care. Can you help to validate this PR with your setup.

agalue:
Sure. Thanks for providing the fix. I'll update the PR once done.

opennms-base-assembly/src/main/filtered/etc/alarmd/drools-rules.d/alarmd.drl
j-white: Logic was copied from the old vacuumd statements: https://github.com/OpenNMS/opennms/blob/opennms-19.1.0-1/opennms-base-assembly/src/main/filtered/etc/vacuumd-configuration.xml#L185

The clear should be reduced on a problem - would need to take a closer look and see why this is not the case.

cgorantla: Looks like this was coming from [here](https://github.com/OpenNMS/opennms/blame/bcd0b266b23c458fc30e1c4b47014ff462ae3bd7/opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java#L228).

Removing that doesn't resolve the issue. That actually `unclear` alarm by this [rule](https://github.com/OpenNMS/opennms/blame/bcd0b266b23c458fc30e1c4b47014ff462ae3bd7/opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java#L228) and ticket never gets closed.

I'm puzzled how it worked with `vacuumd-configuration` in previous versions.

agalue (approved):
I verified the four use cases I described in the JIRA issue, and all are working. I compiled the source from the branch to test them and used JIRA as the Ticketing Integration implementation.

So, from a functionality perspective, it works. However, I can't comment on what has been mentioned about the configuration changes versus how Vacuumd used to do the job.

opennms-base-assembly/src/main/filtered/etc/alarmd/drools-rules.d/alarmd.drl
cgorantla: Looks like `DefaultAlarmTicketerService#createTicket`  updates `lastAutomationTime` and when clear event comes, it satisfies the condition for `unclear`.  Previous `vacuumd-configuration` doesn't seem to update `lastAutomationTime` which may be the reason it never triggers `unclear` there.

PR #3354, opened 2021-03-23
NMS-13110: Fix default minion flow config

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13110


indigo423 (approved):
LGTM! 👍 It seems like the failing smoke test is not related to your changes here.

cgorantla:
> LGTM!  It seems like the failing smoke test is not related to your changes here.

@indigo423 Think current failure is not related to this change. I will re-run that test.
But @j-white pointed  that we need to have a way to verify `conf.d` stuff.
I see that 
`single-port-listener:
      enabled: true`
creates the same routes as defined in this config and that generates `health:check` failures.

Will it be possible for `conf.d` to skip creating configuration if similar config already exists ? 

Otherwise I think we may have this problem when user creates config with `conf.d`.

indigo423:
> Will it be possible for `conf.d` to skip creating configuration if similar config already exists ?

The overlay configuration file wins and overwrites a conf.d file see here: https://github.com/OpenNMS/opennms/blob/639b8a09864a23992191f4f2d1d0195e18bc4b8e/opennms-container/minion/container-fs/entrypoint.sh#L190

cgorantla:
> > Will it be possible for `conf.d` to skip creating configuration if similar config already exists ?
> 
> The overlay configuration file wins and overwrites a conf.d file see here:
> 
> https://github.com/OpenNMS/opennms/blob/639b8a09864a23992191f4f2d1d0195e18bc4b8e/opennms-container/minion/container-fs/entrypoint.sh#L190

But that doesn't seems to be happening here.  If I don't remove conf.d, most of minion tests are failing like [this](https://app.circleci.com/pipelines/github/OpenNMS/opennms/9616/workflows/79805258-0348-4101-af57-2c151de02bd4/jobs/53544) because of Minion container startup failure.

cgorantla:
> > > Will it be possible for `conf.d` to skip creating configuration if similar config already exists ?
> > 
> > 
> > The overlay configuration file wins and overwrites a conf.d file see here:
> > https://github.com/OpenNMS/opennms/blob/639b8a09864a23992191f4f2d1d0195e18bc4b8e/opennms-container/minion/container-fs/entrypoint.sh#L190
> 
> But that doesn't seems to be happening here. If I don't remove conf.d, most of minion tests are failing like [this](https://app.circleci.com/pipelines/github/OpenNMS/opennms/9616/workflows/79805258-0348-4101-af57-2c151de02bd4/jobs/53544) because of Minion container startup failure.

I see that two files are getting created one from conf.d and other one default.
`org.opennms.features.telemetry.listeners-single-port-flows.cfg` 
 `org.opennms.features.telemetry.listeners-udp-9999.cfg`.
   
Guess here the factory name we create on the config `config:edit --alias udp-9999 --factory org.opennms.features.telemetry.listeners` should also be same otherwise it would not overwrite.

fooker (approved):

cgorantla:
Worked with @indigo423 on this. Ideally `conf.d` should replace existing telemetryd config but `conf.d` updates features files to update config not the `cfg` files in `etc` folder like [here](https://github.com/OpenNMS/opennms/blob/e55d3cd1c8ab9ebf4cd12ad250c73e31c3de09ad/opennms-container/minion/container-fs/confd/conf.d/confd-telemetry-feature.xml.toml).  This results in conflicting configuration when there is default config for Telemetryd. 
For now, we decided to drop the flow config on minion.

cgorantla:
Reported confd issue. https://issues.opennms.org/browse/NMS-13265?filter=-2

PR #3357, opened 2021-03-24
NMS-13208: suppress Kafka logging for unknown config

RangerRick:
This PR suppresses some logging from Kafka for config entries it doesn't known about (that are set for other things).

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13208


cgorantla (approved):
LGTM !

PR #3358, opened 2021-03-25
NMS-13217: make trapd config refresh time configurable

RangerRick:
Update the `restTrapConfigurationTimer` so that the interval is configurable.

Note: as soon as I started using a property for this, the `minionproperties` placeholder broke.  I suspect it wasn't even being called at runtime previously since there were no properties referenced in the `camelContext` anyway, but it didn't actively fail until we tried to resolve dynamic values.  As far as I can tell, no minion properties are used directly in the related code, except in places where those properties are already configured/defined.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13217


cgorantla (approved):
LGTM!

PR #3359, opened 2021-03-26
NMS-13210: Add flag to clear %dpname% variable

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-13210

The plan is as follows:
* add flag to clear the %dpname% in foundation-2020, default is false
* remove the flag in develop and always clear %dpname%

So, when this one is merged I'll add another PR to make the changes in develop.

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/opennms.properties
+++ b/opennms-base-assembly/src/main/filtered/etc/opennms.properties
@@ -888,4 +888,8 @@ org.opennms.newts.nan_on_counter_wrap=true
 # By default this option is disabled.
 # org.opennms.features.telemetry.ingressAndEgressRequired=false
 
+# The use of multiple Minions in one location can break the alarm life-cycle for a some OpenNMS features.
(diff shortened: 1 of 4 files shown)


opennms-base-assembly/src/main/filtered/etc/opennms.properties
```diff
@@ -888,4 +888,8 @@ org.opennms.newts.nan_on_counter_wrap=true
 # By default this option is disabled.
 # org.opennms.features.telemetry.ingressAndEgressRequired=false
 
+# The use of multiple Minions in one location can break the alarm life-cycle for a some OpenNMS features.
```
j-white, line 891: Let's avoid any changes in this file since we're targeting `foundation-2020`.

christianpape: Ok. I removed the entry.

j-white (approved):
Thanks @christianpape!

PR #3361, opened 2021-03-30
 NMS-13215: Fallback config for flow timeouts

fooker:
Some routers do not send active and inactive flow timeout configuration
values even if they are configured on the exporting device. The added
settings allow to configure a fallback for these value per
listener/parser.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13215


cgorantla (approved):
LGTM!

PR #3367, opened 2021-04-01
flip flop KafkaProducerIT

j-white:
Attempt to fix flapping KafkaProducerIT:
* Limit the amount of time we wait for the store to be ready.
* Add additional logging when the test times out - show shell output so far.

No JIRA.


j-white:
Can backport this to other affected branches if we see an improvement in `develop`+.

RangerRick (approved):
JUST DO IT! (Shia LaBeouf)

PR #3368, opened 2021-04-06
NMS-13207: Improve performance for checking salted passwords

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13207


cgorantla:
@j-white provided this fix. 
I verified that when making 1000 parallel requests, this has 2x performance improvement.  While it takes ~52 secs for checking passwords for 1000 requests without this change , it takes ~26 secs with this change.

RangerRick (approved):

PR #3372, opened 2021-04-08
NMS-13228: Sanitizing category name in Web UI

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-13228


jeffgdotorg (changes requested):
We should also sanitize the new category name string in the `CategoryController` class. That way we neutralize any attack payload before it gets stored. Otherwise we'd need to escape category names every place we display them (e.g. element/node.jsp)

christianpape:
> We should also sanitize the new category name string in the `CategoryController` class. That way we neutralize any attack payload before it gets stored. Otherwise we'd need to escape category names every place we display them (e.g. element/node.jsp)

To persist an encoded string is not a good idea. This will break other things like search, ReST and so on. I already escaped the category names on the node page (nodeCategory-box.jsp is included in node.jsp).

fooker (approved):

PR #3377, opened 2021-04-14
NMS-13224: Fix some timing issues with BMP

cgorantla:
Sometimes PeerUp notification may come up before Router initiation message.
This may result in peer not getting persisted.
Caching Peers that arrive before Router and persisting them with router should
resolve this.

Also added synchronization around persisting bmp messages.
While running multiple threads within the same session, we may have duplicates
which may result in errors.

Also add Note in doc for BMP,  When Kafka is used. `auto.offset.reset=earliest` need to be set on BMP Adapter.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13224

### DIFF
--- a/features/telemetry/protocols/bmp/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/adapter/BmpMessagePersister.java
+++ b/features/telemetry/protocols/bmp/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/adapter/BmpMessagePersister.java
@@ -35,11 +35,8 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Date;
-import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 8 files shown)


features/telemetry/protocols/bmp/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/adapter/BmpMessagePersister.java
```diff
@@ -104,8 +114,10 @@
     @Autowired
     private SessionUtils sessionUtils;
 
+    private Map<String, Set<BmpPeer>> peersByHashId = new ConcurrentHashMap<>();
```
fooker, line 117: Maybe a `SetMultimap` from guava could help here...

fooker: Does this needs to be a concurrent map if the method is synchronized?

fooker (changes requested):

features/telemetry/protocols/bmp/adapter/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/adapter/BmpMessagePersister.java
cgorantla: yeah, not needed.  Updated!

christianpape (approved):
LGTM.

cgorantla:
@fooker Handled review comments. Can you check this again

fooker (approved):

PR #3379, opened 2021-04-15
NMS-13232: Fix lag in handling minion heartbeat

cgorantla:
Move provisioning to separate thread.

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13232

### DIFF
--- a/features/minion/heartbeat/consumer/src/main/java/org/opennms/minion/heartbeat/consumer/HeartbeatConsumer.java
+++ b/features/minion/heartbeat/consumer/src/main/java/org/opennms/minion/heartbeat/consumer/HeartbeatConsumer.java
@@ -35,9 +35,11 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.RejectedExecutionHandler;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.opennms.core.ipc.sink.api.MessageConsumer;
 import org.opennms.core.ipc.sink.api.MessageConsumerManager;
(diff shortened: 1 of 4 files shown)


features/minion/heartbeat/consumer/src/main/java/org/opennms/minion/heartbeat/consumer/HeartbeatConsumer.java
```diff
@@ -142,31 +148,43 @@ public void handleMessage(MinionIdentityDTO minionHandle) {
...
-            eventBuilder.addParam(EventConstants.PARAM_MONITORING_SYSTEM_ID, minionHandle.getId());
-            eventBuilder.addParam(EventConstants.PARAM_MONITORING_SYSTEM_LOCATION, nextLocation);
-            try {
-                eventProxy.send(eventBuilder.getEvent());
-            } catch (final EventProxyException e) {
-                throw new DataAccessResourceFailureException("Unable to send event", e);
+        // Provision the minions node in a separate thread.
+        final OnmsMinion onmsMinion = minion;
+
+        executor.execute(() -> {
```
j-white, line 154: We don't actually speed up the processing of these in any way, so I would expect that these tasks will continue to pile on to the executor queue.

Can we short circuit the call to provision by verifying if the Minion node already exists? This should avoid the processing delays caused by un-marshalling potentially large requisitions.

cgorantla: Yeah, makes sense. Addressed that.

features/minion/heartbeat/consumer/src/main/java/org/opennms/minion/heartbeat/consumer/HeartbeatConsumer.java
```diff
@@ -96,6 +103,15 @@
...
     private EventSubscriptionService eventSubscriptionService;
 
+    @Autowired
+    private NodeDao nodeDao;
+
+    private final ThreadFactory threadFactory = new ThreadFactoryBuilder()
+            .setNameFormat("minion-provision-handler")
+            .build();
+
+    private final ExecutorService executor = Executors.newSingleThreadExecutor(threadFactory);
```
j-white, line 113: This executor uses an unbound queue by default, so in the event that we were receiving a large number of heartbeat messages and this thread could not keep up, we could still end up in a OOM scenario.

We could mitigate this by limiting the queue capacity to some reasonable number, which would put back-pressure on the broker when full.

j-white: I think I have a cleaner pattern to deal with this now... let me give it a whirl.

j-white (approved):

PR #3391, opened 2021-04-22
NMS-13101: consider properties files under opennms.properties.d when starting Minion

swachter:
Issue: https://issues.opennms.org/browse/NMS-13101

The properties files under `opennms.properties.d` are now appended to the `custom.system.properties` file that is loaded by Karaf.

### DIFF
--- a/opennms-container/minion/container-fs/entrypoint.sh
+++ b/opennms-container/minion/container-fs/entrypoint.sh
@@ -163,6 +163,14 @@ applyConfd() {
   fi
 }
 
+applyOpennmsPropertiesD() {
+  for filename in ${MINION_HOME}/etc/opennms.properties.d/*.properties; do
+    echo "appending to custom.system.properties: $filename"
+    echo "" >> ${MINION_HOME}/etc/custom.system.properties
+    cat "$filename" >> ${MINION_HOME}/etc/custom.system.properties
+  done
+}
+
 start() {
     export KARAF_EXEC="exec"
     cd ${MINION_HOME}/bin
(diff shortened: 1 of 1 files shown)


indigo423:
@swachter Just a heads up https://issues.opennms.org/browse/NMS-13222 is related to this issue and @jadyhou is working on this one already.

opennms-container/minion/container-fs/entrypoint.sh
```diff
@@ -163,6 +163,14 @@ applyConfd() {
   fi
 }
 
+applyOpennmsPropertiesD() {
+  for filename in ${MINION_HOME}/etc/opennms.properties.d/*.properties; do
```
RangerRick, line 167: Is it possible there are cases where this directory isn't created? Does it need to be checked for existence?

swachter: The directory is always there because [instance-id.properties.toml](https://github.com/OpenNMS/opennms/blob/develop/opennms-container/minion/container-fs/confd/conf.d/instance-id.properties.toml) is always run.

swachter: > Should we get rid of [instance-id.properties.toml](https://github.com/OpenNMS/opennms/blob/develop/opennms-container/minion/container-fs/confd/conf.d/instance-id.properties.toml) since it is just another system property ?

I do not know why `instance-id.properties` was introduced in the first place. I think @indigo423 did introduce it. If we delete it then we have to make sure that OpenNMS still gets that setting.

RangerRick: looks like it's dependency of some other external projects, we can let it go through for now rather than trying to rework it

RangerRick (approved):

PR #3398, opened 2021-04-30
NMS-12766: Fix race condition when loading event configurations

swachter:
Issue: https://issues.opennms.org/browse/NMS-12766

* Triggers configuration reloads after configuration providers get registered/unregisterd with the `ConfigReloadContainer` (this fixes a race condition between provider registration and extension binding; cf. issue)
* Fixes subtle multi-threading issues in `ConfigReloadContainer`
* Fixes a logical error in `ConfigReloadContainer.checkForUpdates`. In case the last provider got unregistered the config object gets now set to `null`

### DIFF
--- a/core/api/src/main/java/org/opennms/core/config/api/ConfigReloadContainer.java
+++ b/core/api/src/main/java/org/opennms/core/config/api/ConfigReloadContainer.java
@@ -45,7 +45,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * A container for managing configuation that sources the configuration from one or more {@link ConfigurationProvider}.
+ * A container for managing configuration that sources the configuration from one or more {@link ConfigurationProvider}.
  *
  * @param <V> type of configuration bean
  */
@@ -181,6 +181,7 @@ public class ConfigReloadContainer<V> implements ReloadingContainer<V>, Registra
                     .reduce(folder)
                     .orElse(null);
         }
+        LOG.debug("reloaded - conf object: {}", object);
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


core/api/src/main/java/org/opennms/core/config/api/ConfigReloadContainer.java
```diff
@@ -214,6 +214,7 @@ private synchronized void checkForUpdates() {
 
         if (forceReload || providers.stream().anyMatch(ConfigurationProviderState::shouldReload)) {
             reload();
+            forceReload = false;
```
j-white, line 217: Good catch.

features/api-layer/src/main/java/org/opennms/features/apilayer/config/SyslogMatchExtensionManager.java
```diff
@@ -75,6 +79,7 @@ protected void triggerReload() {
         LOG.debug("Syslog configuration changed. Triggering a reload.");
         try {
             syslogdConfig.reload();
+            eventConfDao.reload();
```
j-white, line 82: I'm not sure of this one. What's the reasoning for reloading the event conf when the syslog configuration changes?

j-white (commented):
The race condition behavior seems to apply to any configuration file that is extended via the current plugin extension mechanism, but this work around is specific to the event configuration. What happens to anything else that is extended?

features/api-layer/src/main/java/org/opennms/features/apilayer/config/SyslogMatchExtensionManager.java
swachter: In the issue, Alejandro reports that

```
show-event-config -u uei.opennms.org/vendor/cisco/syslog/ifDown
```

does not output the expected event configuration. The event configurations are cached in `DefaultEventConfDao`.

The `uei.opennms.org/vendor/cisco/syslog/ifDown` event is defined in `Cisco.ext.syslog.xml`, i.e. the `CiscoSyslogMatchExtension` that registers with the `SyslogMatchExtensionManager`.

j-white: Syslog parsing rules and event definitions are distinct.

Syslog parsing rules: https://github.com/OpenNMS/alec/blob/v1.1.1/integrations/opennms/config/src/main/resources/syslog/Cisco.ext.syslog.xml

Syslog related event definitions:
https://github.com/OpenNMS/alec/blob/v1.1.1/integrations/opennms/config/src/main/resources/events/Cisco.syslog.ext.events.xml

`CiscoSyslogMatchExtension` only extends the parsing rules and has not direct impact on the event definitions.

swachter: Ok. Will remove that line.

swachter: Ok. Will remove that line.

j-white (approved):
Looks good. Thanks for tracking this down.

May want to reconsider the target branch.

PR #3400, opened 2021-05-03
NMS-8581: Allow to specify parameters when configuring notifications

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-8581

### DIFF
--- a/opennms-webapp/src/main/java/org/opennms/web/admin/notification/noticeWizard/NotificationWizardServlet.java
+++ b/opennms-webapp/src/main/java/org/opennms/web/admin/notification/noticeWizard/NotificationWizardServlet.java
@@ -339,7 +339,11 @@ public class NotificationWizardServlet extends HttpServlet {
                 if (!Strings.isNullOrEmpty(parameterName[i])) {
                     final Parameter parameter = new Parameter();
                     parameter.setName(parameterName[i]);
-                    parameter.setValue(parameterValue[i]);
+                    if (i < parameterValue.length && parameterValue[i] != null) {
+                        parameter.setValue(parameterValue[i]);
+                    } else {
+                        parameter.setValue("");
+                    }
                     newNotice.addParameter(parameter);
                 }
             }
--- a/opennms-webapp/src/main/java/org/opennms/web/admin/notification/noticeWizard/NotificationWizardServlet.java
+++ b/opennms-webapp/src/main/java/org/opennms/web/admin/notification/noticeWizard/NotificationWizardServlet.java
@@ -58,6 +58,8 @@ import org.opennms.netmgt.filter.api.FilterDao;
 import org.opennms.netmgt.filter.api.FilterParseException;
 import org.opennms.web.api.Util;
 
+import com.google.common.base.Strings;
+
 /**
  * A servlet that handles the data comming in from the notification wizard jsps.
  *
@@ -325,6 +327,22 @@ public class NotificationWizardServlet extends HttpServlet {
         } else {
             // Must do this to allow clearing out varbind definitions
             newNotice.setVarbind(null);
+        }
+
+        final String[] parameterName = request.getParameterValues("parameterName");
+        final String[] parameterValue = request.getParameterValues("parameterValue");
+
+        newNotice.getParameters().clear();
+
+        if (parameterName != null && parameterValue != null) {
+            for (int i = 0; i < parameterName.length; i++) {
+                if (!Strings.isNullOrEmpty(parameterName[i])) {
+                    final Parameter parameter = new Parameter();
+                    parameter.setName(parameterName[i]);
+                    parameter.setValue(parameterValue[i]);
+                    newNotice.addParameter(parameter);
+                }
+            }
         }
 
         try {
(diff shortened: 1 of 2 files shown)


opennms-webapp/src/main/webapp/admin/notification/noticeWizard/choosePath.jsp
```diff
@@ -120,6 +129,32 @@
...
+    // add row
+    $(document).ready(function() {
+        $("#addRow").click(function () {
+            var html = '';
+            html+='<div class="row" id="inputFormRow">';
+            html+='  <div class="col-md-5">';
+            html+='    <label>Name:</label> <input type="text" class="form-control" size="30" name="parameterName" value=\'<%=WebSecurityUtils.sanitizeString("")%>\'/>';
+            html+='  </div>';
+            html+='  <div class="col-md-5">';
+            html+='    <label>Value:</label> <input class="form-control" type="text" size="30" name="parameterValue" value=\'<%=WebSecurityUtils.sanitizeString("")%>\'/>';
```
fooker, line 142: `WebSecurityUtils.sanitizeString("")` looks strange

fooker (changes requested):

opennms-webapp/src/main/java/org/opennms/web/admin/notification/noticeWizard/NotificationWizardServlet.java
```diff
@@ -327,6 +329,22 @@ private String processPath(final HttpServletRequest request, final HttpSession u
...
+        final String[] parameterValue = request.getParameterValues("parameterValue");
+
+        newNotice.getParameters().clear();
+
+        if (parameterName != null && parameterValue != null) {
+            for (int i = 0; i < parameterName.length; i++) {
+                if (!Strings.isNullOrEmpty(parameterName[i])) {
+                    final Parameter parameter = new Parameter();
+                    parameter.setName(parameterName[i]);
+                    parameter.setValue(parameterValue[i]);
```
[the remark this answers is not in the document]
christianpape: done

opennms-webapp/src/main/webapp/admin/notification/noticeWizard/choosePath.jsp
christianpape: done

fooker (approved):

opennms-webapp/src/main/webapp/admin/notification/noticeWizard/choosePath.jsp
```diff
@@ -152,7 +187,7 @@
         </tr>
         <tr>
           <td width="10%" valign="top" align="left">
-            <label>Parameter:</label>
+            <label>Var-Bind:</label>
```
mfuhrmann, line 190: I know you just renamed it to remove the clash between the other `parameter` we can set, but is `Var-Bind` a good name here? For what can it be used? 
I know that we can use it as some kind of filter. For example when you want to fetch only specific BSMs you can add this in your notification:

```
<varbind>
   <vbname>businessServiceName</vbname>
   <vbvalue>~^Virtual\ Host\ Cluster\ (FUL|FRA)</vbvalue>
</varbind>
```

Could make "Variable Filter" make more sense? Any other ideas?

@Bonrob2 This is definitely something we should add into the docs, since I don't think we have it already somewhere.

christianpape: I don't know. I would favor var-bind since it corresponds to the XML-structure.

opennms-webapp/src/main/webapp/admin/notification/noticeWizard/choosePath.jsp
```diff
@@ -165,7 +200,37 @@
             </div>
           </td>
         </tr>
-        <tr>
+
+          <tr>
+              <td width="10%" valign="top" align="left">
+                  <label>Parameter:</label>
```
mfuhrmann, line 206: @christianpape What are those `parameters` used for? I only use them for Slack/Mattermost. The docs only tell about notification parameters in the Slack/Mattermost section. But that doesn't mean anything ;-)
Is there an other usage possible? If it's only for Slack/Mattermost I would suggest to rename it accordingly. If it's a generic thing, we should keep the current name.

mfuhrmann (commented):
I've tested this PR successfully. I just have 2 questions :grin:

opennms-webapp/src/main/webapp/admin/notification/noticeWizard/choosePath.jsp
mfuhrmann: @christianpape Sounds legit.

Then we just should add that into the docs accordingly. The name isn't that important right now.

opennms-webapp/src/main/webapp/admin/notification/noticeWizard/choosePath.jsp
christianpape: It's a generic thing. So, parameter seems to be fine for me.

mfuhrmann: Ok. Thanks!

mfuhrmann (approved):

opennms-webapp/src/main/webapp/admin/notification/noticeWizard/choosePath.jsp
mfuhrmann: I've created an issue for the documenation part: https://issues.opennms.org/browse/NMS-13273

PR #3402, opened 2021-05-04
NMS-13259: Fixed sidebar

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-13259


fooker (approved):

agalue (approved):
I'm glad to see the fix was that easy.

I tested it, and it works!

PR #3403, opened 2021-05-04
NMS-12767: Situation feedback configuration not loaded

swachter:
Issue: [NMS-12767](https://issues.opennms.org/browse/NMS-12767)

* rename configuration file `org.opennms.features.situation-feedback.persistence.elastic` -> `org.opennms.features.situationfeedback.persistence.elastic`

Reason: A minus sign in filenames of configuration files that are loaded by file installer has special meaning (cf. [file install configurations](https://felix.apache.org/documentation/subprojects/apache-felix-file-install.html#configurations)).

### DIFF
--- a/features/situation-feedback/elastic/src/main/resources/OSGI-INF/blueprint/blueprint.xml
+++ b/features/situation-feedback/elastic/src/main/resources/OSGI-INF/blueprint/blueprint.xml
@@ -1,18 +1,13 @@
-<blueprint xmlns="http://www.osgi.org/xmlns/blueprint/v1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:cm="http://aries.apache.org/blueprint/xmlns/blueprint-cm/v1.1.0"
-           xmlns:ext="http://aries.apache.org/blueprint/xmlns/blueprint-ext/v1.1.0"
-           xsi:schemaLocation="
-		http://www.osgi.org/xmlns/blueprint/v1.0.0
+<blueprint xmlns="http://www.osgi.org/xmlns/blueprint/v1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xmlns:cm="http://aries.apache.org/blueprint/xmlns/blueprint-cm/v1.1.0"
+           xsi:schemaLocation="http://www.osgi.org/xmlns/blueprint/v1.0.0
 		http://www.osgi.org/xmlns/blueprint/v1.0.0/blueprint.xsd
 
 		http://aries.apache.org/blueprint/xmlns/blueprint-cm/v1.1.0
-		http://aries.apache.org/schemas/blueprint-cm/blueprint-cm-1.1.0.xsd
-
-		http://aries.apache.org/blueprint/xmlns/blueprint-ext/v1.1.0
-		http://aries.apache.org/schemas/blueprint-ext/blueprint-ext-1.1.xsd
-">
+		http://aries.apache.org/schemas/blueprint-cm/blueprint-cm-1.1.0.xsd">
 
     <!-- Configuration properties -->
-    <cm:property-placeholder id="elasticSituationFeedbackRepositoryProperties" persistent-id="org.opennms.features.situation-feedback.persistence.elastic" update-strategy="reload">
+    <cm:property-placeholder id="elasticSituationFeedbackRepositoryProperties" persistent-id="org.opennms.features.situationfeedback.persistence.elastic" update-strategy="reload">
         <cm:default-properties>
             <!-- Elastic Connection Settings -->
             <cm:property name="elasticUrl" value="http://localhost:9200" />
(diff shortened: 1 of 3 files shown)


agalue:
There are other configuration files for Karaf features with hyphen on their names and those never caused issues (for instance, Telemetry Listeners on Minions; or Telemetry Adapters on Sentinel), so I'd like to learn how that influences in this case, as I'm curious about it.

However, if that's actually the case, this should be mentioned on the "breaking changes" section on the next major release, as users *must* change their configuration after an upgrade.

Based on that, the customers that are currently using the feature will have to deal with the workaround I outlined in the JIRA issue, correct?

swachter:
> There are other configuration files for Karaf features with a hyphen on their names, and those never caused issues (for instance, Telemetry Listeners on Minions; or Telemetry Adapters on Sentinel), so I'd like to learn how that influences in this case, as I'm curious about it.

I am curious about these, too! Maybe the telemetry adapters use service factories in some way. Configuration files with minus signs are meant for that scenario. I will investigate that.

> 
> However, if that's actually the case, we should mention this in the "breaking changes" section on the next major release, as users _must_ change their configuration after an upgrade.

Sure.

> 
> Based on that, the customers currently using the feature will have to deal with the workaround I outlined in the JIRA issue, correct?

AFAIK: yes.

swachter:
> I am curious about these, too! Maybe the telemetry adapters use service factories in some way. Configuration files with minus signs are meant for that scenario. I will investigate that.

For adapter, service factories are used. The configuration file name identifies a service factory and a configuration alias. From the docs: The configuration in file `org.opennms.features.telemetry.adapters-bmp.cfg` is accessed by:

```
config:edit --alias BMP --factory org.opennms.features.telemetry.adapters
```

agalue:
Thanks for the details.

In that case, could it be possible to apply the same alias logic to the situations-feedback to avoid the breaking change?

swachter:
> Thanks for the details.
> 
> In that case, could it be possible to apply the same alias logic to the situations-feedback to avoid the breaking change?

IMHO changing the configuration mechanism of the situation feedback persistence to the service factory mechanism is difficult. The reason is that we use a so called property placeholder in its blueprint that gets its properties from a service configuration whose PID is specified by its `persistent-id` attribute:

```
    <!-- Configuration properties -->
    <cm:property-placeholder id="elasticSituationFeedbackRepositoryProperties" persistent-id="org.opennms.features.situation-feedback.persistence.elastic" update-strategy="reload">
        <cm:default-properties>
            <!-- Elastic Connection Settings -->
            <cm:property name="elasticUrl" value="http://localhost:9200" />
            <cm:property name="globalElasticUser" value="" />
            <cm:property name="globalElasticPassword" value="" />
            <cm:property name="elasticIndexStrategy" value="monthly" />
            <cm:property name="defaultMaxTotalConnectionPerRoute" value="-1" />
            <cm:property name="maxTotalConnection" value="-1" />
            <cm:property name="ignoreCertificates" value="false"/>
            <cm:property name="nodeDiscovery" value="false" />
            <cm:property name="nodeDiscoveryFrequency" value="0" />
            <cm:property name="proxy" value=""/>
            <cm:property name="retryCooldown" value="500" />
            <cm:property name="connTimeout" value="5000" /> <!-- 5 second timeout for Elasticsearch operations -->
            <cm:property name="readTimeout" value="30000" /> <!-- 30 second timeout for Elasticsearch socket reads -->
            <cm:property name="retries" value="0" /> <!-- Disable retries by default -->

            <!-- Bulk Action Retry settings -->
            <cm:property name="bulkRetryCount" value="5" /> <!-- Number of retries until a bulk operation is considered failed -->

            <!-- Index settings -->
            <!-- https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#index-modules-settings -->
            <cm:property name="settings.index.number_of_shards" value="" />
            <cm:property name="settings.index.number_of_replicas" value="" />
            <cm:property name="settings.index.refresh_interval" value="" />
            <cm:property name="settings.index.routing_partition_size" value="" />
            <cm:property name="indexPrefix" value="" />
        </cm:default-properties>
    </cm:property-placeholder>
```

Unfortunately, property placeholder does not support addressing configuration properties by a factory PID and an alias. In addition, having a factory PID `org.opennms.features.situation` and alias `feedback.persistence.elastic` seems unnatural to me.

When an OpenNMS migration takes place, a script could rename the `org.opennms.features.situation-feedback.persistence.elastic.cfg` file. Alternatively, that rename logic could be added to the OpenNMS launcher script.

agalue:
It is OK. I don't understand how Karaf works; I asked because I wanted to minimize the impact for the customers that rely on that feature.

The thing is, when they upgrade to a version with this fix (and customers never see the changelog before doing that), Situations Feedback won't work. Also, the workaround they know wouldn't as well. That's because both would point to the wrong file name.

Anyway, I'll keep that in mind when they upgrade so they can continue using the feature. After the upgrade, the "new" mandatory solution would be renaming the configuration file and removing the Karaf cache, I think.

j-white (approved):

agalue:
As the solution was merged into `develop` (which I think is good for H28), I have to ask, what about Meridian 2021?

swachter:
I will ask Ben what to do.

RangerRick:
Makes sense to me to at least go back to `foundation-2021` since it's a new release, so breaking changes would be OK.  It could go back earlier if we wrote an upgrade task to rename the config, couldn't it?  That would be up to you guys how far back we should try to support the fix.

agalue:
I don't recall if ALEC works on M2019, but I believe it worth fixing for foundation-2020 via upgrade task to avoid remembering the workaround if we don't for 2 more years. 

But, I think this PR should be cherry-picked on foundation-2021 as it is.

PR #3416, opened 2021-05-11
NMS-13278: Making the DNS requisition location-aware

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-13278

Allow a new `location` parameter in DNS requisition Urls. You can specify a single location that will be associated with all nodes by specifying for example something like `?location=Raleigh`. Furthermore, by preceding a `~` you can specify a properly encoded regular expression to determine the location based on the host entries. This expression must have exactly on capturing group. For instance you can associate a node to a location given by their subdomain with the expression `^(?:.\.|)(.?)\.example\.com\.$`.

### DIFF
--- a/opennms-doc/guide-admin/src/asciidoc/text/provisioning/import-handler.adoc
+++ b/opennms-doc/guide-admin/src/asciidoc/text/provisioning/import-handler.adoc
@@ -103,6 +103,8 @@ We use the regular expression `^(?:.*\.|)(.*?)\.hs-fulda\.de\.$` with exactly on
 
 This will associate nodes for the host entries `foo.e46.hs-fulda.de.` and `bar.e46.hs-fulda.de.` to the location `e46` while hosts like `aaa.g51.hs-fulda.de` and `bbb.g51.hs-fulda.de` will be assigned location `g51`.
 
+TIP: You can use online tools like `https://www.urlencoder.org` to en- and decode your parameters.
+
 .DNS Setup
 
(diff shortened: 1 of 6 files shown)


UberPinguin (approved):

opennms-doc/guide-admin/src/asciidoc/text/provisioning/import-handler.adoc
```diff
@@ -91,6 +91,18 @@ If your expression requires URL encoding (for example you need to use a `?` in t
...
 
+You can use the `location` parameter to assign a single location to all nodes in this requisition:
+
+  dns://my-dns-server/myzone.com/portland/?location=Raleigh
+
+By preceding a `~` you can specify a properly encoded regular expression to determine the location based on the host entries.
+In the following, for example, we want to derive the location based on the subdomain.
+We use the regular expression `^(?:.*\.|)(.*?)\.hs-fulda\.de\.$` with exactly one capturing group:
+
+  dns://my-dns-server/hs-fulda.de/?location=~%5E%28%3F%3A.%2A%5C.%7C%29%28.%2A%3F%29%5C.hs-fulda%5C.de%5C.%24
```
indigo423, line 102: OMG!  Srsly?! Because of this URL thingy in Provisiond, we have to URL encode the regexp?!

christianpape: Yes. That's true.

fooker (approved):

PR #3418, opened 2021-05-13
NMS-13266: Discard notifications in scheduled outage

cgorantla:
Discard notifications in scheduled outage even if corresponding event is part of `auto-ack` list.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13266


RangerRick (approved):
Makes sense to me...

PR #3439, opened 2021-05-27
NMS-13324: Store service monitor status in RRD

fooker:
The adds a flag `rrd-status` to service monitor definitions allowing to
store the status of the monitor in RRD (or other performance data
repository). The status is stored in a new repository directory called
`status` which follwos the same semantics as the `response` repository.

To query the monitor status, a new resource type is added allowing to
query the current status and the status history using the measurements
API.

The flag is disabled by default and must be enabled for each monitor.

To clarify things in UI, the labels for the status and response time
resources have been modified to show the difference.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13324


opennms-dao/src/main/java/org/opennms/netmgt/dao/support/PerspectiveServiceResourceType.java
```diff
@@ -0,0 +1,121 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2007-2020 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2020 The OpenNMS Group, Inc.
+ *
```
christianpape, line 6: use 2021 the two lines above

opennms-dao/src/main/java/org/opennms/netmgt/dao/support/PerspectiveStatusResourceType.java
```diff
@@ -0,0 +1,52 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2007-2020 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2020 The OpenNMS Group, Inc.
+ *
```
christianpape, line 6: use 2021 the two lines above

opennms-dao/src/main/java/org/opennms/netmgt/dao/support/ServiceResourceType.java
```diff
@@ -0,0 +1,150 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2007-2015 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2015 The OpenNMS Group, Inc.
+ *
```
christianpape, line 6: use 2021 the two lines above or was this file just moved around?

opennms-dao/src/main/java/org/opennms/netmgt/dao/support/ServiceStatusResourceType.java
```diff
@@ -0,0 +1,60 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2007-2015 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2015 The OpenNMS Group, Inc.
+ *
```
christianpape, line 6: use 2021 the two lines above or was this file just moved around?

opennms-services/src/main/java/org/opennms/netmgt/poller/pollables/StatusStoringServiceMonitorAdaptor.java
```diff
@@ -0,0 +1,131 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2006-2014 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2014 The OpenNMS Group, Inc.
+ *
```
christianpape, line 6: use 2021 the two lines above

opennms-services/src/test/java/org/opennms/netmgt/poller/pollables/StatusStoringServiceMonitorAdaptorPersistenceTest.java
```diff
@@ -0,0 +1,187 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2008-2015 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2015 The OpenNMS Group, Inc.
+ *
```
christianpape, line 6: use 2021 the two lines above

christianpape (changes requested):
Only headers need some more love.

opennms-dao/src/main/java/org/opennms/netmgt/dao/support/ServiceResourceType.java
fooker: Changed.

christianpape (approved):
Perfect. Great feature.

mmahacek:
There was a Mattermost message suggesting using a different name other than `rrd-status` that would be more intuitive for Newts users to know that this would apply in cases where the time series strategy is not RRD.

reissmann:
Thanks @fooker, that is exactly what I needed :)

fooker:
> I would love to see this feature, but H28 as a target seems a bit risky with a target Tuesday next week and a code freeze tomorrow. I would vote for develop as a target for the PR.

I'm fine with that.

docs/modules/operation/pages/service-assurance/polling-packages.adoc
```diff
@@ -75,6 +75,22 @@ WARNING: The RRD configuration and the service polling interval must be aligned.
...
 IMPORTANT: If you change the polling interval afterwards, you must recreate existing RRD files with the new definitions.
 
+[[ga-pollerd-packages-service-status-persistence]]
+== Service Status Persistence
+
+Each Service Monitor can be configured to store the current status of the Service in RRD files.
+This allows to query the current status and the status history similar to response times.
+
+This feature is disabled by default and must be enabled for each service individually.
+To do so, a service parameter must be added to the according services in `poller-configuration.xml`:
```
Bonrob2, line 85:
```suggestion
To do so, add a service parameter to the appropriate services in `poller-configuration.xml`:
```

Bonrob2 (changes requested):
Looks good with a few changes.

j-white:
Good idea. I'm a bit hesitant on following the semantics of the `response` repository though and it looks like the bulk of the code changes are related to this.

Could we model this a generic resource keyed by node and interface instead?

Bonrob2 (approved):
Doc changes look good!

fooker:
> Good idea. I'm a bit hesitant on following the semantics of the `response` repository though and it looks like the bulk of the code changes are related to this.
> 
> Could we model this a generic resource keyed by node and interface instead?

Moved to a separate issue: https://issues.opennms.org/browse/NMS-13363

PR #3449, opened 2021-06-01
NMS-13322: Fallback config for flow sampling interval

fooker:
Some routers do not send sampling interval configuration
values even if they are configured on the exporting device. The added
settings allow to configure a fallback for these value per
listener/parser.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13322

### DIFF
--- a/docs/modules/operation/pages/telemetryd/protocols/ipfix.adoc
+++ b/docs/modules/operation/pages/telemetryd/protocols/ipfix.adoc
@@ -30,6 +30,9 @@ The IPFIX UDP parser supports protocol detection.
 | `dnsLookupsEnabled`      | Used to enable or disable DNS resolution for flows.                                         | no       | true
 | `sequenceNumberPatience` | A value > 1 enables checking for seuqence number completeness.
                              The value gives the size of the history buffer allowing flows to be processed out of order. | no       | 32
+| `flowActiveTimeoutFallback`   | Fallback value for active flow timeout if setting value is not included in exported flows | no | null
+| `flowInactiveTimeoutFallback`   | Fallback value for inactive flow timeout if setting value is not included in exported flows | no | null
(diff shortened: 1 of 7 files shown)


fooker:
This branch also includes a cherry-picked version of #3361 which makes it not mergable. The relevant commit is only the last one. I will merge it manually after review.

christianpape (approved):
Großartig.

opennms-doc/guide-admin/src/asciidoc/text/telemetryd/protocols/netflow5.adoc
```diff
@@ -27,6 +27,8 @@ The Netflow v5 UDP Parser supports protocol detection.
 | `maxClockSkew`        | The maximum delta in seconds between exporter and Minion timestamps.       | no       | 0
 | `clockSkewEventRate`  | Used to rate-limit clock skew events in seconds.                           | no       | 3600
 | `dnsLookupsEnabled`      | Used to enable or disable DNS resolution for flows.                        | no       | true
+| `flowActiveTimeoutFallback`   | Fallback value for active flow timeout if setting value is not included in exported flows | no | null
+| `flowInactiveTimeoutFallback`   | Fallback value for inactive flow timeout if setting value is not included in exported flows | no | null
```
Naicisum, line 31: These values are defined in the documentation but are not implemented in the parser.

Naicisum (changes requested):

opennms-doc/guide-admin/src/asciidoc/text/telemetryd/protocols/netflow5.adoc
fooker: oh, good catch. This was already an oversight in the already merged version. Will fix.

Naicisum (approved):
LGTM!

PR #3451, opened 2021-06-01
NMS-13314: Prevent meta-data modification

fooker:
Modification of meta-data was only intended to be done via requisitions.
The requisitions process will create, update and delete meta-data of a node
according to the values defined in a requisition. Changing the meta-data
after a requisition synchronization was never intended conceptually. It
suffered from loosing of meta-data after re-synchronization. Keeping the
data would require manual deletion in massive scale using the UI and
leads to desync with the expected values.

This also reverts #3291.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13314


docs/modules/development/pages/rest/meta-data.adoc
```diff
@@ -1,7 +1,7 @@
 [[metadata-rest]]
 = Metadata
 
-You can qurey, add, modify, or delete the actual metadata of nodes, interfaces, and services via REST endpoints.
+You can qurey the actual metadata of nodes, interfaces, and services via REST endpoints.
 
 [[rest-api-meta-data-get]]
 == GETs (Reading Meta-Data)
```
Bonrob2, line 7:
```suggestion
== GETs (Reading Metadata)
```

Bonrob2 (changes requested):
one minor change

docs/modules/development/pages/rest/meta-data.adoc
```diff
@@ -1,7 +1,7 @@
 [[metadata-rest]]
 = Metadata
 
-You can qurey, add, modify, or delete the actual metadata of nodes, interfaces, and services via REST endpoints.
+You can qurey the actual metadata of nodes, interfaces, and services via REST endpoints.
```
Bonrob2, line 4:
```suggestion
You can query the actual metadata of nodes, interfaces, and services via REST endpoints.
```

Bonrob2 (changes requested):
one more. :-)

Bonrob2 (approved):
LGTM!

Naicisum:
Code wise, I think this should go into H29 as it's a breaking change and H28 has been released. This API has been in the wild for a while and it is unknown who would be impacted by this change.

indigo423:
> Code wise, I think this should go into H29 as it's a breaking change and H28 has been released. This API has been in the wild for a while and it is unknown who would be impacted by this change.

IIRC we try to follow semver.org and with the following guidelines:

> Given a version number MAJOR.MINOR.PATCH, increment the:
  * MAJOR version when you make incompatible API changes,
  * MINOR version when you add functionality in a backward-compatible manner, and
  * PATCH version when you make backward-compatible bug fixes.

PR #3459, opened 2021-06-07
NMS-13312:  RESTful API for local health

jadyhou:
Tag health-check and provide RESTful API for local health.

I'll update the OpenAPI doc after we decided on the RESTful APIs we want to provide. Also, need to determine the tags. 

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13312


core/health/api/src/main/java/org/opennms/core/health/api/HealthCheck.java
```diff
@@ -44,6 +44,14 @@
      */
     String getDescription();
 
+    /**
+     *
+     * A tag to indicate the category of the health-check.
+     *
+     * @return The HealthTag type, describing category of the health-check.
+     */
+    HealthTag getTag();
```
cgorantla, line 53: As we discussed before, change this to   `List<String> getTags()`.  That would give more flexibility for implementations to choose tags.

core/health/impl/src/main/java/org/opennms/core/health/impl/DefaultHealthCheckService.java
```diff
@@ -41,12 +41,7 @@
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
-import org.opennms.core.health.api.Context;
-import org.opennms.core.health.api.Health;
-import org.opennms.core.health.api.HealthCheck;
-import org.opennms.core.health.api.HealthCheckService;
-import org.opennms.core.health.api.Response;
-import org.opennms.core.health.api.Status;
+import org.opennms.core.health.api.*;
```
cgorantla, line 44: In general, use specific imports instead of wildcard.

cgorantla (changes requested):

core/health/api/src/main/java/org/opennms/core/health/api/HealthTag.java
```diff
@@ -0,0 +1,36 @@
...
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.core.health.api;
+
+public enum HealthTag {
+    LOCAL,
+    BROKER,
+    OPENNMSREST,
```
RangerRick, line 34: a tiny nitpick, but generally in enums we use `_` when something is multiple words (`OPENNMS_REST`)

jadyhou: Good point, thanks.

core/health/api/src/main/java/org/opennms/core/health/api/HealthCheck.java
jadyhou: How about use List<HealthTag> ?

cgorantla: we can stick to strings to expose those as constants here in this Interface.

cgorantla:
Consider this tags for HealthChecks : https://github.com/OpenNMS/opennms/commit/6e8e8ac7dd8ad0ad74b0611ea63281887533d6b5

core/health/impl/src/main/java/org/opennms/core/health/impl/DefaultHealthCheckService.java
jadyhou: updated

jadyhou:
> Consider these tags for HealthChecks : [6e8e8ac](https://github.com/OpenNMS/opennms/commit/6e8e8ac7dd8ad0ad74b0611ea63281887533d6b5)

Thanks, very helpful.

core/health/api/src/main/java/org/opennms/core/health/api/HealthCheck.java
jadyhou: Sure. Updated. If you don't mind. I use a HealthConstant class for the constants instead of using Constant Interface.

core/health/api/src/main/java/org/opennms/core/health/api/HealthCheck.java
```diff
@@ -44,6 +47,16 @@
      */
     String getDescription();
 
+    /**
+     *
+     * A tag to indicate the category of the health-check.
+     *
+     * @return The HealthTag type, describing category of the health-check.
+     */
+    default List<String> getTags(){
```
cgorantla, line 56: We can force implementations to adopt this. No need for default

core/health/api/src/main/java/org/opennms/core/health/api/HealthCheck.java
cgorantla: > Sure. Updated. If you don't mind. I use a HealthConstant class for the constants instead of using Constant Interface.

Not sure if there is any benefit of using Class. HealthCheckConstants would be better name for it.

core/health/rest/src/main/java/org/opennms/core/health/rest/HealthCheckRestService.java
```diff
@@ -47,4 +43,9 @@
     @GET
     @Produces(MediaType.APPLICATION_JSON)
     Response getHealth(@QueryParam("t") @DefaultValue("5000") int timeoutInMs);
+
+    @GET
+    @Path("/tags/{tag}")
+    @Produces(MediaType.APPLICATION_JSON)
```
cgorantla, line 49: Instead of adding another Rest API, we could add a query param on `getHealth` so that we can supply multiple tags. An example [here](https://github.com/OpenNMS/opennms/blob/4542129338f09206eab222cb0bbd8b0873c19f53/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/MinionRestService.java#L93)

core/health/impl/src/main/java/org/opennms/core/health/impl/DefaultHealthCheckService.java
```diff
@@ -95,8 +93,8 @@ public DefaultHealthCheckService(BundleContext bundleContext) {
         try {
             // Fail if no checks are available
             List<HealthCheck> checks = getHealthChecks();
-            if (tag != null){
-                checks = checks.stream().filter(check -> check.getTags().contains(tag)).collect(Collectors.toList());
+            if (tags != null && tags.stream().anyMatch(tag -> !Strings.isNullOrEmpty(tag))){
+                checks = checks.stream().filter(check -> check.getTags().stream().anyMatch(tags::contains)).collect(Collectors.toList());
             }
             if (checks == null || checks.isEmpty()) {
```
cgorantla, line 99: checks getting accessed before null check.

core/health/impl/src/main/java/org/opennms/core/health/impl/DefaultHealthCheckService.java
```diff
@@ -95,8 +93,8 @@ public DefaultHealthCheckService(BundleContext bundleContext) {
         try {
             // Fail if no checks are available
             List<HealthCheck> checks = getHealthChecks();
-            if (tag != null){
-                checks = checks.stream().filter(check -> check.getTags().contains(tag)).collect(Collectors.toList());
+            if (tags != null && tags.stream().anyMatch(tag -> !Strings.isNullOrEmpty(tag))){
```
cgorantla, line 96: Add a test case on DefaultHealthCheckService to verify that it can process multiple tags properly

core/health/rest/src/main/java/org/opennms/core/health/rest/HealthCheckRestService.java
```diff
@@ -42,10 +43,5 @@
 
     @GET
     @Produces(MediaType.APPLICATION_JSON)
-    Response getHealth(@QueryParam("t") @DefaultValue("5000") int timeoutInMs);
-
-    @GET
-    @Path("/tags/{tag}")
-    @Produces(MediaType.APPLICATION_JSON)
-    Response getHealthByTag(@QueryParam("t") @DefaultValue("5000") int timeoutInMs, @PathParam("tag") String tag);
+    Response getHealth(@QueryParam("t") @DefaultValue("5000") int timeoutInMs,@QueryParam("tag") @DefaultValue("") List<String> tags);
```
cgorantla, line 46: Using UriInfo would be more appropriate.

jadyhou: why?

core/health/rest/src/main/java/org/opennms/core/health/rest/HealthCheckRestService.java
jadyhou: Okay, didn't know we plan to support querying multiple tags. I've updated the code.

core/health/api/src/main/java/org/opennms/core/health/api/HealthCheck.java
jadyhou: Okay, either way works for me, all depend on how we want to design this.  Changed.

core/health/impl/src/main/java/org/opennms/core/health/impl/DefaultHealthCheckService.java
jadyhou: The checks.stream actually can handle the null value. But I'll add test cases after all design works were done and when we are sure no more further code changes needed.

core/health/rest/src/main/java/org/opennms/core/health/rest/HealthCheckRestService.java
jadyhou: Hopefully I'm not missing anything here. I see both way would work. But if we use the UriInfo, then that means user can input anything during the runtime so that user would have more chance to fail because the code doesn't handle it. If we define the query parameter individually, user would know what we specifically want them to input, which would be more intuitive and helpful for the user. Below is the screenshot I've had so far. There could be more enhancements that can be done in this feature, I'm not sure if we want them all in this jira/pull request or we want to break them down and finished by steps. 
Screen Shot 2021-06-08 at 4 11 12 PM

core/health/impl/src/main/java/org/opennms/core/health/impl/DefaultHealthCheckService.java
cgorantla: checks == null seems unnecessary.  I'm good with implementation so far.  Also update `health-check` on Karaf.
Add `-t` option on `HealthCheckCommand`

jadyhou: Sure. I've wrote up https://issues.opennms.org/browse/NMS-13327 for this.

core/health/rest/src/main/java/org/opennms/core/health/rest/HealthCheckRestService.java
jadyhou: Yes, I've noticed this in the CI test also. It is because minion JAX RS doesn't support this. I changed the List<String> to String[], then it turned out minion jaxrx supports this, but OpenNMS JAX RS doesn't support it. Can you check the difference when you enabled the REST in minion? It seems the minion is not consistent with OpenNMS, should because of a different version or different setting. Minion and OpenNMS should have the same expected behavior with the same code since they share the same Karaf feature. 
I can also try to use UriInfo to see if they behave the same.

opennms-bamboo:
Welcome Jane Hou (jhou@opennms.com).

I could not find a signed OpenNMS Contributor Agreement (OCA) assigned to the username Jane Hou (jhou@opennms.com).

core/health/rest/src/main/java/org/opennms/core/health/rest/HealthCheckRestService.java
jadyhou: UriInfo works for OpenNMS but failed for Minion on my machine this morning, but it seems to work just now when I try it again. I pushed the code and let's give it a try in CI.

jadyhou: Great news, the Ci is passed :).  If we are good with the current implementation, I'm going to start checking the test code and docs.

core/health/rest/src/main/java/org/opennms/core/health/rest/HealthCheckRestService.java
```diff
@@ -46,5 +46,5 @@
 
     @GET
     @Produces(MediaType.APPLICATION_JSON)
-    Response getHealth(@QueryParam("t") @DefaultValue("5000") int timeoutInMs);
+    Response getHealth(@QueryParam("t") @DefaultValue("5000") int timeoutInMs, @Context final UriInfo uriInfo);
```
cgorantla, line 49: Can we derivce `@QueryParam("t")`  from UriInfo ?

cgorantla (commented):
Looks good.
Can we also extend `tags`  QueryParam on  `/probe` as well ?

core/health/rest/src/main/resources/openapi/health-check-api.yaml
```diff
@@ -39,6 +39,13 @@ paths:
             type: integer
             format: int32
             default: 5000
+        - name: tag
```
cgorantla, line 42: Does swagger work on Minion without any changes ?
From my understanding this only works on OpenNMS but the URL here in this file is hardcoded to Minion URL

jadyhou: This is only the doc. We don't have embedded Swagger UI related to this. User will need to input URL by themself when they use it. The URL we put here is just an example.

core/health/rest/src/main/java/org/opennms/core/health/rest/HealthCheckRestService.java
jadyhou: Technically we can. But I'd prefer not to do that. Typically, if we can use Query Param, we should avoid using the UriInfo. I think our code in other places does follow that pattern.

cgorantla:
> > Looks good.
> > Can we also extend `tags` QueryParam on `/probe` as well ?
> 
> Sure, we can add that to make probe support tags as well if we want. We can write a jira for that.

Created https://issues.opennms.org/browse/NMS-13369 for it.

opennms-bamboo:
Welcome Jane Hou (jhou@opennms.com).

I could not find a signed OpenNMS Contributor Agreement (OCA) assigned to the username Jane Hou (jhou@opennms.com).

opennms-bamboo:
Welcome Jane Hou (jhou@opennms.com).

I could not find a signed OpenNMS Contributor Agreement (OCA) assigned to the username Jane Hou (jhou@opennms.com).

opennms-bamboo:
Welcome Jane Hou (jhou@opennms.com).

I could not find a signed OpenNMS Contributor Agreement (OCA) assigned to the username Jane Hou (jhou@opennms.com).

core/health/rest/src/main/resources/openapi/health-check-api.yaml
cgorantla: Opened https://issues.opennms.org/browse/NMS-13386 to address docs.

cgorantla (approved):

opennms-bamboo:
Welcome @jadyhou.

I could not find a signed OpenNMS Contributor Agreement (OCA) assigned to the username @jadyhou.

opennms-bamboo:
Welcome Jane Hou (jhou@opennms.com).

I could not find a signed OpenNMS Contributor Agreement (OCA) assigned to the username Jane Hou (jhou@opennms.com).

PR #3471, opened 2021-06-07
NMS-12689: Add regex to limit threshold values

cgorantla:
Add regex validation for threshold values:
Metadata notation  OR float value for  `value` and `rearm`
Metadata notation OR positive integer for `trigger`

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12689

### DIFF
--- a/opennms-config-model/src/main/resources/xsds/thresholding.xsd
+++ b/opennms-config-model/src/main/resources/xsds/thresholding.xsd
@@ -376,41 +376,60 @@
       </annotation>
     </attribute>
 
-    <attribute name="value" type="string" use="required">
-      <annotation>
-        <documentation>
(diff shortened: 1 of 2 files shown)


cgorantla:
I couldn't get browser validation on input fields working properly. 
Also not sure if there is browser compatibility for pattern validation.  [This](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern) doesn't list any compatibility list

cgorantla:
Verified with `config-tester` validation.

jadyhou (approved):

cgorantla:
> Scary, but good

True. that's why I targeted it to develop.

PR #3473, opened 2021-06-08
NMS-13179: Reworked Interpolator results

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-13179


fooker (approved):

indigo423:
This here looks a bit weird and very rough? It seems like you can expose more information and not just the scope, right?
Screenshot 2021-06-08 at 18 16 40

indigo423:
> This here looks a bit weird and very rough? It seems like you can expose more information and not just the scope, right?
> Screenshot 2021-06-08 at 18 16 40

Suggestion: What about adding 2 columns "Match" and "Scope" instead of just dumping a series of key values?

christianpape:
Reworked the page. Now a value's tooltip contains the match and scope data.

indigo423 (approved):
I like this approach with your tooltip idea much better! Well done
Screenshot 2021-06-09 at 09 18 30

fooker:
> I like this approach with your tooltip idea much better! Well done
> Screenshot 2021-06-09 at 09 18 30

@indigo423 How does the tooltip look like if the parameter contains multiple placeholders?

indigo423:
Screenshot 2021-06-09 at 11 09 47

PR #3484, opened 2021-06-10
NMS-13354: add version to OpenAPI doc info section

jadyhou:
This fix the version part of the Jira.

### External References

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-13354

### DIFF
--- a/opennms-webapp-rest/src/main/webapp/WEB-INF/applicationContext-cxf-rest-v2.xml
+++ b/opennms-webapp-rest/src/main/webapp/WEB-INF/applicationContext-cxf-rest-v2.xml
@@ -44,6 +44,7 @@
     </bean>
     <bean id="openApiFeature" class="org.apache.cxf.jaxrs.openapi.OpenApiFeature">
         <property name="title" value="OpenNMS V2 RESTful API" />
+        <property name="version" value="1.0.0" />
         <property name="description" value="OpenNMS V2 RESTful API" />
         <property name="license" value="OpenNMS(R) Licensing" />
         <property name="licenseUrl" value="http://www.gnu.org/licenses/" />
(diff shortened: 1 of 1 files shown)


opennms-webapp-rest/src/main/webapp/WEB-INF/applicationContext-cxf-rest-v2.xml
```diff
@@ -44,6 +44,7 @@
     </bean>
     <bean id="openApiFeature" class="org.apache.cxf.jaxrs.openapi.OpenApiFeature">
         <property name="title" value="OpenNMS V2 RESTful API" />
+        <property name="version" value="1.0.0" />
```
cgorantla, line 47: Should this be independent of Horizon/Meridian version ?  
If that is standard practice, we can start with minor version like `0.1.0` since we currently don't support OpenAPI for all the Rest API

cgorantla (commented):
Also curious what is the tool that was used to reproduce this problem

opennms-webapp-rest/src/main/webapp/WEB-INF/applicationContext-cxf-rest-v2.xml
jadyhou: I'm fine with any version number we decide to start with here.

jadyhou:
> Also curious what is the tool that was used to reproduce this problem

No tool needed, it's in our "help" page :)

mmahacek:
> > Also curious what is the tool that was used to reproduce this problem
> 
> No tool needed, it's on our "help" page :)

I had tried to import into Postman, and when that failed, I ran the spec through https://apitools.dev/swagger-parser/online/ and that's what told me the version property was missing.

opennms-webapp-rest/src/main/webapp/WEB-INF/applicationContext-cxf-rest-v2.xml
jadyhou: I've updated the version number to 0.1.0

RangerRick (approved):

PR #3485, opened 2021-06-10
NMS-13016: run as non-root by default

RangerRick:
This PR changes OpenNMS to run as `opennms` by default in both the debian and RPM packages.

It does a number of things:

* adds some command-line shell utilities for fixing/changing permissions
* sets `$RUNAS` in all scripts to `opennms`
* updates the `ensure-user-ping.sh` script to handle updating the group to match the `opennms` GID
* adds logic to `$OPENNMS_HOME/bin/install` to validate ownership

By default, it _only_ changes the permissions of files _installed_ by the packages, it does _not_ recurse through the user's data, since large installs could take hours to complete a recursive chown.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13016


agalue:
I'll try to find time to test the DEB and RPM packages on a fresh install and through an upgrade to see how all this works.

However, I cannot see the changes to the `Dockerfile` for OpenNMS now that we don't need to apply the changes there:

https://github.com/OpenNMS/opennms/blob/develop/opennms-container/horizon/Dockerfile#L38-L39
https://github.com/OpenNMS/opennms/blob/develop/opennms-container/horizon/Dockerfile#L116-L120
https://github.com/OpenNMS/opennms/blob/develop/opennms-container/horizon/Dockerfile#L133-L134

The following line should be consistent with the package manager:

https://github.com/OpenNMS/opennms/blob/develop/opennms-container/horizon/Dockerfile#L161

j-white:
What is the expected behavior for `trapd` (binds to port 162) and for ICMP pinging now that the privileges are dropped?

RangerRick:
> I'll try to find time to test the DEB and RPM packages on a fresh install and through an upgrade to see how all this works.
> 
> However, I cannot see the changes to the `Dockerfile` for OpenNMS now that we don't need to apply the changes there:
> 
> https://github.com/OpenNMS/opennms/blob/develop/opennms-container/horizon/Dockerfile#L38-L39
> https://github.com/OpenNMS/opennms/blob/develop/opennms-container/horizon/Dockerfile#L116-L120
> https://github.com/OpenNMS/opennms/blob/develop/opennms-container/horizon/Dockerfile#L133-L134
> 
> The following line should be consistent with the package manager:
> 
> https://github.com/OpenNMS/opennms/blob/develop/opennms-container/horizon/Dockerfile#L161

Good point, I hadn't looked into the docker images, I thought they were already running as non-root like minion/sentinel.  I'll take a look.

> What is the expected behavior for `trapd` (binds to port 162) and for ICMP pinging now that the privileges are dropped?

ICMP should be handled by the `ensure-user-ping.sh` script, which is called in the package postinstall.  Trapd I guess we'll have to set higher and then add some documentation on port-forwarding?

indigo423 (commented):
The only thing is 162/udp for trapd as @j-white mentioned as well. In docker, we use 1162 similar to 1514/udp for Syslogd. In containers, Docker takes care of redirects. Suggestion to solve this, change from 162/udp to 1162/udp and explain in our docs, based on here: https://opennms.discourse.group/t/how-to-deal-with-privileged-ports-1024-for-network-management-functions/1572.

Ping we can follow the procedure here, especially if we set UID/GID: https://opennms.discourse.group/t/how-to-allow-unprivileged-users-to-use-icmp-ping/1573.

If we try to cover it in the docs, I would prefer the options **without** giving capabilities to the `java` binary with `set_cap`. We get a wider attack vector through any program running with `java` and can mention it as the last option with a bigger warning sign if don't have any other option.

debian/opennms-common.postinst
```diff
@@ -0,0 +1,27 @@
+#! /bin/sh
+# postinst script for opennms-common
+#
+# vim:set ai et sts=4 sw=4 tw=0:
+
+## create the OpenNMS group and user if it doesn't exist
+getent group opennms >/dev/null 2>&1 || addgroup --system opennms
+getent passwd opennms >/dev/null 2>&1 ||
+    adduser --system --home /usr/share/opennms --no-create-home \
+            --disabled-password --ingroup opennms opennms
```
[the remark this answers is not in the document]
RangerRick: On a regular system we can't really assume they'll be available.  We could try _preferring_ it but I don't think we could rely on assuming it matches.

dino2gnt:
There is a process provided by the various distros for application packagers to request allocation of a known UID/GID combo.  Do we care about that, or are we content to dynamically allocate a UID and GID from the user range instead? This could lead to inconsistent UIDs across installs.  Does that matter? 

Ex: https://docs.fedoraproject.org/en-US/packaging-guidelines/UsersAndGroups/#Soft_static_allocation

RangerRick:
> There is a process provided by the various distros for application packagers to request allocation of a known UID/GID combo. Do we care about that, or are we content to dynamically allocate a UID and GID from the user range instead? This could lead to inconsistent UIDs across installs. Does that matter?
> 
> Ex: https://docs.fedoraproject.org/en-US/packaging-guidelines/UsersAndGroups/#Soft_static_allocation

I would guess we'd be hard pressed to ask for it considering we're not even submitted as an official package, and we install in `/opt`, but I suppose we could ask?

RangerRick:
OK, the latest updates have changes to attempt to use uid/gid 1000 if available, otherwise fall back to whatever the system gives you. Also ports over some of the perm-fixing from the other packages.

opennms-base-assembly/src/main/filtered/etc/trapd-configuration.xml
```diff
@@ -1 +1 @@
-<trapd-configuration xmlns="http://xmlns.opennms.org/xsd/config/trapd" snmp-trap-address="*" snmp-trap-port="162" new-suspect-on-trap="false" include-raw-message="false" threads="0" queue-size="10000" batch-size="1000" batch-interval="500"/>
\ No newline at end of file
+<trapd-configuration xmlns="http://xmlns.opennms.org/xsd/config/trapd" snmp-trap-address="*" snmp-trap-port="1162" new-suspect-on-trap="false" include-raw-message="false" threads="0" queue-size="10000" batch-size="1000" batch-interval="500"/>
```
j-white, line 1: Minion also defaults to 1162 - so these will conflict by default when run on the same host now

opennms-install/src/main/java/org/opennms/install/Installer.java
```diff
@@ -572,6 +577,56 @@ public void verifyFilesAndDirectories() throws FileNotFoundException {
...
+        if (RUNAS_USER == null) {
+            final var opennmsConf = readOpennmsConf();
+            // use RUNAS from opennms.conf if found, fall back to -Dopennms.runas, fall back to "opennms" if nothing is found
+            RUNAS_USER = opennmsConf.getProperty("RUNAS", System.getProperty("opennms.runas", "opennms"));
+        }
+
+        return RUNAS_USER;
+    }
+
+    private Properties readOpennmsConf() throws IOException {
```
j-white, line 608: Good candidate for a unit test? Feels like there are a lot of edge cases in here.

debian/control
```diff
@@ -1,13 +1,13 @@
 Source: opennms
 Section: contrib/net
 Priority: optional
-Maintainer: Jeff Gehlbach <jeffg@opennms.org>
+Maintainer: Benjamin Reed <ranger@opennms.org>
```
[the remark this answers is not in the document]
RangerRick: I mean, it's been me for quite a while now  :)

.circleci/config.yml
```diff
@@ -653,40 +653,22 @@ workflows:
             - integration-test
           filters:
             branches:
-              only:
```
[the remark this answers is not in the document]
RangerRick: It's probably worth talking about. I'm told we have leftover credits in circle for now, so I thought it might be useful to turn them on all the time. I was gonna take that out when I merged though, it was just to make testing easier/quicker as I was iterating.

opennms-install/src/main/java/org/opennms/install/Installer.java
RangerRick: Good point, I'll add one.

opennms-base-assembly/src/main/filtered/etc/trapd-configuration.xml
RangerRick: I don't imagine running both on the same system is super common, but doesn't hurt to change it I suppose.

opennms-base-assembly/src/main/filtered/etc/trapd-configuration.xml
```diff
@@ -1 +1 @@
-<trapd-configuration xmlns="http://xmlns.opennms.org/xsd/config/trapd" snmp-trap-address="*" snmp-trap-port="1162" new-suspect-on-trap="false" include-raw-message="false" threads="0" queue-size="10000" batch-size="1000" batch-interval="500"/>
+<trapd-configuration xmlns="http://xmlns.opennms.org/xsd/config/trapd" snmp-trap-address="*" snmp-trap-port="2162" new-suspect-on-trap="false" include-raw-message="false" threads="0" queue-size="10000" batch-size="1000" batch-interval="500"/>
```
indigo423, line 1: Just for fun :) What about using 10162 than we use the same pattern as with Syslogd 10514 on core?

https://github.com/OpenNMS/opennms/blob/7ecbfee2a8ea39d250ef3982167de39eee34df99/opennms-base-assembly/src/main/filtered/etc/syslogd-configuration.xml#L9

RangerRick: sure... ¯\_(ツ)_/¯

.circleci/config.yml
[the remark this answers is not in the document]
RangerRick: (Thinking about it, we don't usually need them that often unless it's for specific stuff. I'll revert that change before merging.)

indigo423:
We have to update the docs reflecting the new SNMP Trap UDP listening port. Just a quick sanity check search for 162/udp 
gave me this here:

```
modules/deployment/pages/core/centos-rhel/initialize-core.adoc
57:     By default, the OpenNMS {page-component-title} SNMP trap daemon listens on 162/udp and the Syslog daemon listens on 10514/udp.
```

```
modules/deployment/pages/core/debian-ubuntu/initialize-core.adoc
49:TIP: By default the OpenNMS SNMP Trap daemon listens on 162/udp and the Syslog daemon listens on 10514/udp.
```

RangerRick:
> We have to update the docs reflecting the new SNMP Trap UDP listening port.

Fixed your examples, didn't find anything else obvious with a grep.

I'm inclined to go ahead and merge this unless anyone has other feedback, it's likely there are still things we're going to run into, but I'm not sure we're going to find them without getting these changes out into the wider world.

PR #3486, opened 2021-06-10
NMS-11764: Don't add SNMP interfaces will null/blank labels to the CollectionSet

j-white:
JIRA: https://issues.opennms.org/browse/NMS-11764

Change the behavior to avoid adding SNMP interfaces will null/blank labels to the CollectionSet rather than throwing an exception when these are processed.

This allows other resources in the CollectionSet to be processed.

Can backport further if we see fit.


RangerRick (approved):
Nice catch! Seems like a candidate for `foundation-2021` (and thus Meridian 2021) at the least, IMHO.

j-white:
Rebased to `foundation-2021`, will merge once green.

PR #3487, opened 2021-06-11
NMS-13357: Lever regex for indexing of resource paths (resorceId)

patrick-schweizer:
In order to test you need to build & install locally:
* https://github.com/OpenNMS/opennms-integration-api/tree/ps/NMS-13356-ts-2.0
It won't build otherwise.

As plugin you can use the in memory plugin with this branch
* https://github.com/opennms-forge/opennms-tss-plugin-inmemory/tree/NMS-13357-regex

Jira: https://issues.opennms.org/browse/NMS-13357


features/timeseries/src/main/java/org/opennms/netmgt/timeseries/resource/TimeseriesSearcher.java
```diff
@@ -64,18 +65,16 @@
 
     private static final Logger LOG = LoggerFactory.getLogger(TimeseriesSearcher.class);
 
-    private final TimeseriesStorageManager timeseriesStorageManager;
-
     private final TimeSeriesMetaDataDao metaDataDao;
 
-    private final Cache<Tag, Set<Metric>> indexMetricsByTag;
+    private final Cache<TagMatcher, Set<Metric>> indexMetricsByTagMatcher;
```
patrick-schweizer, line 70: @j-white I used the same cache logic as we did before: Cache by tag matcher instead of cache by tag.
Maybe we want to revisit that, but so far it worked good enough I guess..

features/timeseries/src/main/java/org/opennms/netmgt/timeseries/util/TimeseriesUtils.java
```diff
@@ -137,4 +112,11 @@ public static Sample createSampleForIndexingStrings(final String resourceId, Map
                 .build();
     }
 
+    public static String toSearchRegex(ResourcePath path, int depth) {
+        return "^" + // start string
+                toResourceId(path) + // exact match
+                ":[^.:]*".repeat(depth) + // colon (:) plus any chars except colon (:), repeated 'depth' times
```
j-white, line 118: There's a slight edge case here where the `:` character is part of the path. The `EscapableResourceIdSplitter` handles this by prefixing these with `\` so they are "escaped" accordingly" and not treated as separators when parsing.

Is it possible to add test cases and adjust the regex for these? i.e. don't consider a `:` if it is prefixed with a `\`

j-white (commented):
Looks good!

features/timeseries/src/main/java/org/opennms/netmgt/timeseries/resource/TimeseriesSearcher.java
j-white: Makes sense, that can work for now.

features/timeseries/src/main/java/org/opennms/netmgt/timeseries/util/TimeseriesUtils.java
patrick-schweizer: Yeah, that makes sense...
Although thinking of it: Do we need the conversion from "/" to ":" at all? I think we did that for Newts (and thus inherited it in the integration layer) but we might take some complexity out if we remove that transformation? It might be ok to use the path forward slashes?
Do you remember why we have the colons at all?

features/timeseries/src/test/java/org/opennms/netmgt/timeseries/resource/TimeseriesResourceStorageDaoTest.java
```diff
@@ -222,26 +219,15 @@ private void replay() throws StorageException {
...
+                String regex = toSearchRegex(resourcePath, depth + 1);
                 Set<Metric> metrics = new HashSet<>();
                 for (Entry<ResourcePath, Set<String>> entry : indexedPaths.entrySet()) {
-                    Map<String, String> attributes = Maps.newHashMap();
-                    // Build the indexed attributes and attempt to match them against the given query
-                    TimeseriesUtils.addIndicesToAttributes(entry.getKey(), attributes);
-                    if (value.equals(attributes.get(field))) {
-
-                        for(String name : entry.getValue()) {
+                    if (toResourceId(entry.getKey()).matches(regex)) { // find all paths that match our regex
```
j-white, line 225: Let's use RE2 for evaluating and matching any of the regular expressions: https://github.com/OpenNMS/opennms-integration-api/pull/42

features/timeseries/src/main/java/org/opennms/netmgt/timeseries/util/TimeseriesUtils.java
patrick-schweizer: I replaced the colon with a forward slash as discussed. The assumption is that a path can never have a forward slash in an element.

features/timeseries/src/test/java/org/opennms/netmgt/timeseries/resource/TimeseriesResourceStorageDaoTest.java
patrick-schweizer: fixed.

opennms-model/src/main/java/org/opennms/netmgt/model/ResourcePath.java
```diff
@@ -54,27 +57,28 @@
...
-        for (String el : path) {
-            elements.add(el);
-        }
+        this(Arrays.asList(path));
     }
 
     public ResourcePath(Iterable<String> pathElements) {
         for (String el : pathElements) {
+            if(el != null && el.contains("/")) {
+                throw new IllegalArgumentException("path elements must not contain a forward slash");
```
patrick-schweizer, line 66: @j-white I try to make here explicit that we can't have a forward slash in a path. I hope this won't have side effects?

j-white: Should be OK.

Let's include the path elements in the exception message if thrown - will allow us to trace down the culprit.

patrick-schweizer: fixed

j-white (approved):

PR #3497, opened 2021-06-21
NMS-12778: Add  more fields to event/alarm proto

cgorantla:
Add node label, location, system id to  Event/Alarm as requested.
Note that dist_poller is system id.

Also add event_snmp on Event
and last_update_time on Alarm.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12778


features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/ProtobufMapper.java
```diff
@@ -87,17 +87,17 @@ public ProtobufMapper(EventConfDao eventConfDao, HwEntityDao hwEntityDao, Sessio
...
-                                    .setForeignId(node.getForeignId())
-                                    .setForeignSource(node.getForeignSource())
-                                    .build();
-                        } else {
-                            return OpennmsModelProtos.NodeCriteria.newBuilder()
-                                    .setId(nodeId)
-                                    .build();
+                        OpennmsModelProtos.NodeCriteria.Builder builder = OpennmsModelProtos.NodeCriteria.newBuilder();
+                        if (node == null) {
+                            return builder.setId(nodeId).build();
```
RangerRick, line 92: `nodeId` should always be set to remain backwards-compatible, should it not?  Here you're only setting it if `node == null` but the other codepath never gets it.

RangerRick (changes requested):
Other than missing `nodeId` in the codepath where node exists, this looks good to me.

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/ProtobufMapper.java
cgorantla: Oh yeah, good catch !

agalue:
This might be separate, but as in terms of code changes, the effort is  the same, so I'm going to comment on it here:

We have another customer requesting to extract the Trap ID from the events (and alarms, using the underlying event). This is stored in the database inside the `eventsnmp` column, associated with the [ISnmp](https://github.com/OpenNMS/opennms/blob/7ecbfee2a8ea39d250ef3982167de39eee34df99/features/events/api/src/main/java/org/opennms/netmgt/events/api/model/ISnmp.java#L37) interface.

Do you need a new JIRA issue for that, or can we take the opportunity to make two customers happy at once on this PR?

cgorantla:
> This might be separate, but as in terms of code changes, the effort is the same, so I'm going to comment on it here:
> 
> We have another customer requesting to extract the Trap ID from the events (and alarms, using the underlying event). This is stored in the database inside the `eventsnmp` column, associated with the [ISnmp](https://github.com/OpenNMS/opennms/blob/7ecbfee2a8ea39d250ef3982167de39eee34df99/features/events/api/src/main/java/org/opennms/netmgt/events/api/model/ISnmp.java#L37) interface.
> 
> Do you need a new JIRA issue for that, or can we take the opportunity to make two customers happy at once on this PR?

Handled that here.
Also added `last_update_time` for Alarm.

agalue:
You're the best. Thanks for adding both changes!

agalue (approved):
Looks great to me!
Once CircleCI can give me a Docker Image, I can give it a try, but don't wait on that to merge it ;)

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/ProtobufMapper.java
```diff
@@ -245,6 +248,9 @@ public ProtobufMapper(EventConfDao eventConfDao, HwEntityDao hwEntityDao, Sessio
         setTimeIfNotNull(event.getTime(), builder::setTime);
         setTimeIfNotNull(event.getCreationTime(), builder::setCreateTime);
 
+        String eventSnmp = event.getSnmp() == null ? null : SnmpInfo.format(event.getSnmp(), EVENT_SNMP_FIELD_SIZE);
```
j-white, line 251: I think we'd be better off to model this with multiple fields or a new object rather than formatting this as a string - how are clients expected to parse this?

cgorantla: I lifted this from [here](https://github.com/OpenNMS/opennms/blob/7ecbfee2a8ea39d250ef3982167de39eee34df99/features/events/daemon/src/main/java/org/opennms/netmgt/eventd/processor/HibernateEventWriter.java#L293).   I was trying to match it with `eventSnmp` field from OnmsEvent.

j-white: Doesn't map well to the protobuf model IMO. If we want to include this information should be break it out as individual fields and save the client from having the parse the string.

agalue: I agree on having discrete fields like the way `ISnmp` references it within an `IEvent`.

However, the use case for the customer who raised this particular change can retrieve the Trap ID.

OpenNMS doesn't have a field for that, but it can be build by concatenating the [enterprise](https://github.com/OpenNMS/opennms/blob/master/features/events/api/src/main/java/org/opennms/netmgt/events/api/model/ISnmp.java#L40) field (named just `id` for some reason, but called `enterprise` within the Mask element) plus the [specific](https://github.com/OpenNMS/opennms/blob/master/features/events/api/src/main/java/org/opennms/netmgt/events/api/model/ISnmp.java#L42) fields from the `ISnmp` object; makes sense? Although it would be great to combine both, remember that SNMPv2 traps know one OID as the Trap ID, whereas SNMPv1 traps dissect that into "enterprise" + "generic" + "specific".

Also, the customer who requested this last change is using Meridian 2019 (even if the other customer who triggered the JIRA issue is using Horizon). Fortunately, "adding" new fields to a Protobuf definition doesn't alter the contract (the older version of the ".proto" works, and you get the new fields once you update the client code), so I believe it should be fine to rebase that to 2019.

RangerRick (approved):

RangerRick:
Oh, lol, I commented on the wrong protobuf issue about the 0-indexed thing, but hey, this one has the `nodeId` fix I requested anyways, so rescinding my review still applies. ;)

RangerRick (approved):

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/ProtobufMapper.java
cgorantla: Created SnmpInfo in Events.

````
message SnmpInfo {
  string id = 1;
  string id_text = 2;
  string version = 3;
  uint32 specific = 4;
  uint32 generic = 5;
  string community = 6;
}
```

agalue:
I downloaded the OCI, built a Docker Test environment, and used the `snmptrap` command to emulate a trap. The Producer is forwarding it, but unfortunately, the event doesn't have the SNMP fields.

I'm using a client that I wrote in Go to convert the Protobuf payload to JSON, verifying the producer's output:
https://github.com/agalue/producer-receiver

agalue:
LOL, disregard, I forgot to update the `.proto` on my app. I'll try one more time.

agalue (approved):
Now that I added the correct `.proto` to my Go app, I verified that everything works as expected, for the `snmpInfo` on events, the `node_label` and `location` for nodes, and `last_update_time` for alarms.

As a suggestion, the following is how the `snmpInfo` is rendered:
```json
"snmpInfo": {
      "id": ".1.3.6.1.4.1.9.9.276",
      "version": "v2",
      "specific": 1,
      "generic": 6,
      "community": "public"
}
```
I think `id_text` is redundant and potentially unnecessary, even though it doesn't appear in the output.

cgorantla:
> Now that I added the correct `.proto` to my Go app, I verified that everything works as expected, for the `snmpInfo` on events, the `node_label` and `location` for nodes, and `last_update_time` for alarms.
> 
> As a suggestion, the following is how the `snmpInfo` is rendered:
> 
> ```json
> "snmpInfo": {
>       "id": ".1.3.6.1.4.1.9.9.276",
>       "version": "v2",
>       "specific": 1,
>       "generic": 6,
>       "community": "public"
> }
> ```
> 
> I think `id_text` is redundant and potentially unnecessary, even though it doesn't appear in the output.

Thanks for checking.  `id_text` is present in eventconf but not really mapped when setting Snmp elements from Trap. So removed it.

RangerRick (approved):

j-white (approved):

PR #3501, opened 2021-06-23
NMS-13378: Added Meta-Data search and custom information

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-13378

* Added search provider for meta-data values. 
* Introduced a new property `org.opennms.netmgt.search.info` which allows to add a custom line to search results. For instance you can define something like `org.opennms.netmgt.search.info=${node:sys-contact} (Department: ${asset:department})` to always display the sys-contact and the department below the node match.

### DIFF
--- a/core/ipc/rpc/utils/src/main/java/org/opennms/core/rpc/utils/mate/EntityScopeProviderImpl.java
+++ b/core/ipc/rpc/utils/src/main/java/org/opennms/core/rpc/utils/mate/EntityScopeProviderImpl.java
@@ -115,7 +115,6 @@ public class EntityScopeProviderImpl implements EntityScopeProvider {
             if (node.getAssetRecord() != null) {
                 Scope assetScope = new ObjectScope<>(Scope.ScopeName.NODE, node.getAssetRecord())
                         .map(ASSET, "category", (a) -> Optional.ofNullable(a.getCategory()))
-                        .map(ASSET, "vendor", (a) -> Optional.ofNullable(a.getVendor()))
                         .map(ASSET, "manufacturer", (a) -> Optional.ofNullable(a.getManufacturer()))
                         .map(ASSET, "vendor", (a) -> Optional.ofNullable(a.getVendor()))
                         .map(ASSET, "model-number", (a) -> Optional.ofNullable(a.getModelNumber()))
--- a/core/web-assets/src/main/assets/js/apps/onms-central-search/quicksearch.html
+++ b/core/web-assets/src/main/assets/js/apps/onms-central-search/quicksearch.html
@@ -82,6 +82,8 @@
             <a id="onms-search-result-item-{{$index}}" ng-class="{'selected': item.selected}" ng-repeat-end ng-if="item.type === 'Item'" href="{{item.url}}" class="list-group-item list-group-item-action" ng-mouseover="select($index)" ng-click="resetQuery()">
                 <h6 class="mb-0"><i class="{{item.icon}}" ng-if="item.icon">&nbsp;</i>{{item.label}} <i style="-webkit-transform: scaleY(-1);transform: scaleY(-1);" class="fa fa-reply pull-right" ng-if="item.selected"></i></h6>
                 <small class="mr-4" ng-class="{'text-muted': !item.selected}" ng-repeat="match in item.matches">{{match.label}}: {{match.value}}</small>
+                <br ng-if="item.info !== ''"/>
+                <small class="mr-4" ng-class="{'text-muted': !item.selected}" ng-if="item.info !== ''">{{item.info}}</small>
             </a>
         </div>
 
--- a/features/search/api/src/main/java/org/opennms/netmgt/search/api/SearchResultItem.java
+++ b/features/search/api/src/main/java/org/opennms/netmgt/search/api/SearchResultItem.java
@@ -55,6 +55,7 @@ public class SearchResultItem {
     private final Map<String, String> properties = new HashMap<>();
     private final List<Match> matches = new ArrayList<>();
     private int weight;
+    private String info = "";
 
     public String getIdentifier() {
         return identifier;
@@ -161,5 +162,13 @@ public class SearchResultItem {
 
         // Adjust weight
         setWeight(Math.max(getWeight(), mergeMe.getWeight()));
+    }
+
+    public void setInfo(final String info) {
+        this.info = info;
+    }
+
+    public String getInfo() {
+        return this.info;
     }
 }
--- a/features/search/providers/pom.xml
+++ b/features/search/providers/pom.xml
@@ -42,5 +42,9 @@
       <artifactId>opennms-web-api</artifactId>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.opennms.core.ipc.rpc</groupId>
+      <artifactId>org.opennms.core.ipc.rpc.utils</artifactId>
+    </dependency>
   </dependencies>
 </project>
--- a/features/search/providers/src/main/java/org/opennms/netmgt/search/providers/FilterSearchProvider.java
+++ b/features/search/providers/src/main/java/org/opennms/netmgt/search/providers/FilterSearchProvider.java
@@ -35,6 +35,7 @@ import java.util.stream.Collectors;
 
 import org.opennms.core.criteria.Criteria;
 import org.opennms.core.criteria.CriteriaBuilder;
+import org.opennms.core.rpc.utils.mate.EntityScopeProvider;
 import org.opennms.netmgt.dao.api.NodeDao;
 import org.opennms.netmgt.filter.api.FilterDao;
 import org.opennms.netmgt.filter.api.FilterParseException;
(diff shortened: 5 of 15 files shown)


mfuhrmann:
@christianpape I've tried this branch without success. Here is what I've done:

Created a new properties file like this:
```
mf1045@mf1045-nb115:~/Documents/git/onms/stack-play/minimal-horizon$ cat container-fs/opt/opennms-overlay/etc/opennms.properties.d/searchbar.properties
org.opennms.netmgt.search.info=${node:stack} ${node:env} ${node:app)
```
The nodes I defined have Meta-Data like this
image

but ONMS throws an error while starting
```
	... 33 more
Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'daoContext' defined in URL [jar:file:/opt/opennms/lib/opennms-dao-28.0.1-SNAPSHOT.jar!/beanRefContext.xml]: Bean instantiation via constructor failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.context.support.ClassPathXmlApplicationContext]: Constructor threw exception; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'sessionFactory' defined in class path resource [META-INF/opennms/applicationContext-shared.xml]: Invocation of init method failed; nested exception is java.lang.IllegalArgumentException: unmatched placeholder start [${node:stack} ${node:env} ${node:app)]
	at org.springframework.beans.factory.support.ConstructorResolver.autowireConstructor(ConstructorResolver.java:275)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.autowireConstructor(AbstractAutowireCapableBeanFactory.java:1148)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBeanInstance(AbstractAutowireCapableBeanFactory.java:1050)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:512)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:482)
	at org.springframework.beans.factory.support.AbstractBeanFactory$1.getObject(AbstractBeanFactory.java:306)
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:230)
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:302)
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:197)
	at org.springframework.beans.factory.support.BeanDefinitionValueResolver.resolveReference(BeanDefinitionValueResolver.java:351)
	... 46 more
Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.context.support.ClassPathXmlApplicationContext]: Constructor threw exception; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'sessionFactory' defined in class path resource [META-INF/opennms/applicationContext-shared.xml]: Invocation of init method failed; nested exception is java.lang.IllegalArgumentException: unmatched placeholder start [${node:stack} ${node:env} ${node:app)]
	at org.springframework.beans.BeanUtils.instantiateClass(BeanUtils.java:163)
	at org.springframework.beans.factory.support.SimpleInstantiationStrategy.instantiate(SimpleInstantiationStrategy.java:122)
	at org.springframework.beans.factory.support.ConstructorResolver.autowireConstructor(ConstructorResolver.java:267)
	... 55 more
```

When I remove 2 of the Meta-Data like this
```
mf1045@mf1045-nb115:~/Documents/git/onms/stack-play/minimal-horizon$ cat container-fs/opt/opennms-overlay/etc/opennms.properties.d/searchbar.properties
org.opennms.netmgt.search.info=${node:stack}
```

OpenNMS starts but the search box is broken: 
image

christianpape:
Last bracket isn‘t a curly one in your expression 😎

core/ipc/rpc/utils/src/main/java/org/opennms/core/rpc/utils/mate/EntityScopeProviderImpl.java
```diff
@@ -115,7 +115,6 @@ public Scope getScopeForNode(final Integer nodeId) {
             if (node.getAssetRecord() != null) {
                 Scope assetScope = new ObjectScope<>(Scope.ScopeName.NODE, node.getAssetRecord())
                         .map(ASSET, "category", (a) -> Optional.ofNullable(a.getCategory()))
-                        .map(ASSET, "vendor", (a) -> Optional.ofNullable(a.getVendor()))
```
fooker, line 118: Why was this removed?

fooker (changes requested):

core/ipc/rpc/utils/src/main/java/org/opennms/core/rpc/utils/mate/EntityScopeProviderImpl.java
christianpape: Because two lines later again the vendor is set.

fooker (approved):

mfuhrmann:
@christianpape :facepalm: Sorry. Didn't see it.

Works perfectly now! Not sure where this red error message came from. Can't reproduce it.

And there is a cool side effect (or maybe designed behavior? ). The fact, that a non-existent variable won't be displayed brings up a different approach of how to combine variables.

In some cases, I have `node` and in some cases, I have `requisition` MetaData for the same matter (in this case `stack`, `env`, `app`). If you write it like this the tag always looks great :-)
`${node:stack}${requisition:stack}:${node:env}${requisition:env}:${node:app}${requisition:app}`

mfuhrmann (approved):
Should we have documentation for this new enhancement? I know there is no documentation about the search box at all. So maybe adding the docs is not an option now and within this PR.

But I can imagine, that this configuration enhancement never reaches out to the people if it is just a new property.

Let's ping @Bonrob2. Maybe she knows about plans to document the search panel in general? Or at least she now heard about this one and can refer to it somewhere.

christianpape:
> @christianpape 🤦 Sorry. Didn't see it.
> 
> Works perfectly now! Not sure where this red error message came from. Can't reproduce it.
> 
> And there is a cool side effect (or maybe designed behavior? ). The fact, that a non-existent variable won't be displayed brings up a different approach of how to combine variables.
> 
> In some cases, I have `node` and in some cases, I have `requisition` MetaData for the same matter (in this case `stack`, `env`, `app`). If you write it like this the tag always looks great :-)
> `${node:stack}${requisition:stack}:${node:env}${requisition:env}:${node:app}${requisition:app}`

You can also use something like ${node:stack|requisition:stack}.

mfuhrmann:
@christianpape Is this PR only about adding MetaData into the search panel? Or: is it also possible to remove for example the ForeignId and other default information?

mfuhrmann:
> You can also use something like ${node:stack|requisition:stack}.

@christianpape That's cool. Is that a general notation that works everywhere? If not, we definitely should have docs about it. Or at least 1-2 examples in the properties file?

fooker:
> > You can also use something like ${node:stack|requisition:stack}.
> 
> @christianpape That's cool. Is that a general notation that works everywhere? If not, we definitely should have docs about it. Or at least 1-2 examples in the properties file?

That works everywhere. See https://docs.opennms.com/horizon/28.0.0/operation/meta-data.html#ga-meta-data-dsl

mfuhrmann:
> > > You can also use something like ${node:stack|requisition:stack}.
> > 
> > 
> > @christianpape That's cool. Is that a general notation that works everywhere? If not, we definitely should have docs about it. Or at least 1-2 examples in the properties file?
> 
> That works everywhere. See https://docs.opennms.com/horizon/28.0.0/operation/meta-data.html#ga-meta-data-dsl

@fooker Great!

christianpape:
> @christianpape Is this PR only about adding MetaData into the search panel? Or: is it also possible to remove for example the ForeignId and other default information?

The central search displays matching nodes and the information which part of each node matched the given string. Each of the implemented search provider returns this information. For instance, an user can see wether an asset field, the node’s label or any of its IP addresses matched. This is a pretty nice functionality and I really don't see a reason to omit this information.

mfuhrmann:
@christianpape I understand and you are right. It's a good behavior. In my setup the foreignId and nodeLabel are identical. So you get a result like:
image
This redundancy of information in these few pixels are confusing me :joy: Maybe that's why I thought it would be a cool option to hide it.
But of course, the foreignId and nodeLabel could be different and then it behaves different.
Thanks for clarifying.

Bonrob2:
> Should we have documentation for this new enhancement? I know there is no documentation about the search box at all. So maybe adding the docs is not an option now and within this PR.
> 
> But I can imagine, that this configuration enhancement never reaches out to the people if it is just a new property.
> 
> Let's ping @Bonrob2. Maybe she knows about plans to document the search panel in general? Or at least she now heard about this one and can refer to it somewhere.

Yes, let's document the search panel in general and this in particular. I can create a Jira.

PR #3502, opened 2021-06-23
NMS-13367: allow setting -Xms as well

RangerRick:
Allow setting `-Xms` in `opennms.conf` without needing to add it to `ADDITIONAL_MANAGER_OPTIONS`

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13367


agalue (approved):
LGTM!

I agree with your comment on the JIRA issue to keep the current one as "max" and add a new one for the "initial" for compatibility purposes.

The original tag was always intended for the max heap anyway.

Naicisum (approved):
LGTM! 🤘

PR #3507, opened 2021-06-25
NMS-13191: Add ResourceId related fields to CollectionSet

cgorantla:
Adds Resource related fields to proto.

Limitations:
Responsetime resources doesn't embed any node info in CollectionResource.
So they won't have resourceId fields filled.

Also When only kafka persister is enabled without RRD/Netws/Integration, resourceIds will not be populated.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13191


agalue:
This looks fantastic. I'll try it out soon.

I have a question. The customer who motivated me to create the JIRA issue in question is running Meridian 2020. Is it possible to rebase this great effort to `foundation-2020`?

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/collection/CollectionSetMapper.java
```diff
@@ -115,12 +126,28 @@ public void visitResource(CollectionResource resource) {
...
                     }
                     genericResourceBuilder.setType(resource.getResourceTypeName());
                     genericResourceBuilder.setInstance(resource.getInstance());
                     collectionSetResourceBuilder.setGeneric(genericResourceBuilder);
                 }
+                // Response time resources doesn't embed any node info, they will not have any resource-id info.
+                if (nodeId > 0) {
+                    ResourceId resourceId = getResourceId(resource, nodeId);
+                    if (resourceId != null) {
+                        OnmsResource onmsResource = resourceDao.getResourceById(resourceId);
```
j-white, line 140: We need to be careful with calls to the `ResourceDao` in this code path - the `ResourceDao` can make calls to the database to the load the node and SNMP interfaces. Performance really matters here.

agalue: That's definitely a problem.

Most of the information is part of the `CollectionResource`, except a few things like the Resource ID. Could it be possible to extract that logic, put it somewhere so the `ResourceDao` and the Producer can share it?

cgorantla: The logic to derive ResourceId doesn't involve ResourceDao.  We can probably derive `resource_id` from CollectionResource and remove rest of fields since it could be expensive to fetch other fields from ResourceDao.

Other fields like `resource_label`, `resource_name` can be derived from Rest API once we know `resource_id`

agalue: `CollectionResource` has the "resource name", as that's essentially part of the resource ID, obtained via `getInstance()`. The resource type name is also present there (which is also required to build the resource ID). The only thing that is not there is the Resource Label, which requires having access to the resource type. Still, considering the limitations the customers currently have (like relying on ReST after processing each CollectionSet via the Producer), I guess that at least those should be useful. I mean, to use ReST, you need the Resource ID.

cgorantla: Okay,  will check and try to retrieve `resource_id`, `resource_name` and `resource_type_name` from CollectionResource without depending  on `ResourceDao`

RangerRick (approved):

opennms-dao/src/main/java/org/opennms/netmgt/dao/support/DefaultResourceDao.java
```diff
@@ -398,4 +401,33 @@ private boolean isDomainResourceTypeUsed() {
         }
         return false;
     }
+
+    @Override
+    public ResourceId getResourceId(CollectionResource resource, long nodeId) {
```
j-white, line 406: Good candidate for a unit test.

cgorantla: Added. Also validated generated `ResourceId` with Rest API.

PR #3517, opened 2021-07-03
NMS-13413: Remove storing of string attributes at a resource path without a sample

patrick-schweizer:
Jira: https://issues.opennms.org/browse/NMS-13413


patrick-schweizer:
@j-white I am not sure if we need to test the collectors manually?

j-white (approved):
Looks good. We can do further validation before merging back into `develop`.

PR #3520, opened 2021-07-06
NMS-13356: TSS 2.0: Improved Tag handling

patrick-schweizer:
This pull request

- Simplifies the approach to indexing and reduce the number of meta-data tags necessary
- Removes the need to maintain resource meta-data in PostgreSQL
- Improves visibility into time series cardinality 

See details in in Jira: https://issues.opennms.org/browse/NMS-13356


pom.xml
```diff
@@ -1441,8 +1441,8 @@
     <powermockVersion>2.0.9</powermockVersion>
     <okhttpVersion>3.10.0</okhttpVersion>
     <okioVersion>1.14.0</okioVersion>
-    <opennmsApiVersion>0.5.1</opennmsApiVersion>
-    <opennmsApiVersionOsgi>0.5.1</opennmsApiVersionOsgi>
+    <opennmsApiVersion>0.6.0-SNAPSHOT</opennmsApiVersion>
```
patrick-schweizer, line 1444: TODO: use fixed version once its published

core/schema/src/main/liquibase/29.0.0/changelog.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd">
+
+
+  <changeSet author="pschweizer" id="29.0.0-drop-timeseries_meta-table">
```
j-white, line 9: We should include a note about breaking changes here: `docs/modules/releasenotes/pages/whatsnew.adoc`

Something like: Updated time series integration layer - need to use plugins compatible with OIA 0.6.x - DB tables are no longer used - no migration support.

features/juniper-tca-collector/src/main/java/org/opennms/netmgt/collectd/tca/TcaCollectionHandler.java
```diff
@@ -102,28 +103,28 @@
...
+
 	/**
 	 * Instantiates a new TCA collection set.
 	 *
 	 * @param agent the agent
 	 * @param repository the repository
 	 */
-	public TcaCollectionHandler(SnmpCollectionAgent agent, RrdRepository repository, ResourceStorageDao resourceStorageDao,
-	        ResourceTypesDao resourceTypesDao, LocationAwareSnmpClient locationAwareSnmpClient) {
+	public TcaCollectionHandler(SnmpCollectionAgent agent, RrdRepository repository,
```
j-white, line 118: Would be good to test the modified collectors at runtime, both on the core and on the Minion.

Just want to make sure that the dependencies get wired correctly, getting the collectors to actually collect data for these will be challanging.

core/schema/src/main/liquibase/29.0.0/changelog.xml
patrick-schweizer: How do you like this? https://github.com/OpenNMS/opennms/pull/3520/commits/20add353adc4ef49e83bdcefa85a40533dcf9e4b

pom.xml
j-white: I published a release version of 0.6.0 - it should be available and indexed within the next hour or so.

patrick-schweizer: I updated it.

features/juniper-tca-collector/src/main/java/org/opennms/netmgt/collectd/tca/TcaCollectionHandler.java
patrick-schweizer: checking....

patrick-schweizer: I think we are good but I am not 100% sure. Here is what I did:

I debugged into TcaCollector and there the blobStore seems to be available:
image

I didn't get to debug into Sftp3gppVTDXmlCollectionHandler but instead managed to debug into XmlCollector (same module) and was able to obtain the blobStore:
image

j-white: Looks good - this means the new dependency is getting wired in. These collectors are rarely used, so this a low risk item.

j-white (approved):

PR #3524, opened 2021-07-06
NMS-13417: use /run/opennms as the pid directory

RangerRick:
This PR fixes PID-handling on Debian.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13417


opennms-base-assembly/src/main/filtered/etc/opennms.service
```diff
@@ -9,6 +9,7 @@ User=opennms
 Environment="OPENNMS_HOME=${install.dir}"
 
 Type=forking
+RuntimeDirectory=opennms
```
RangerRick, line 12: this is the magic juice... `systemd` will automatically create `/run/opennms` for us before the init script starts 👍

indigo423 (approved):
Sweet `RuntimeDirectory` is the way to go and takes also the user into account. Nice!

PR #3534, opened 2021-07-13
NMS-13401: Add preliminary interfaces for Object replication

cgorantla:
Define object replication as OnmsTwin
Define interfaces for OnmsTwin publisher/subscriber.
Define broker interfaces to interact with Onms publisher/subscriber.

Add Mock implementation with kafka broker and IT to test object replication.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13401

### DIFF
--- a/features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapListener.java
+++ b/features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapListener.java
@@ -33,10 +33,13 @@ import java.lang.reflect.UndeclaredThrowableException;
 import java.net.InetAddress;
 import java.net.SocketException;
 import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
 
 import org.opennms.core.ipc.sink.api.AsyncDispatcher;
 import org.opennms.core.ipc.sink.api.MessageDispatcherFactory;
 import org.opennms.core.logging.Logging;
+import org.opennms.core.twin.api.OnmsTwin;
+import org.opennms.core.twin.subscriber.api.OnmsTwinSubscriber;
 import org.opennms.core.utils.InetAddressUtils;
 import org.opennms.netmgt.config.TrapdConfig;
 import org.opennms.netmgt.config.TrapdConfigFactory;
@@ -64,6 +67,8 @@ public class TrapListener implements TrapNotificationListener {
     private TrapdConfig m_config;
 
     private AsyncDispatcher<TrapInformationWrapper> m_dispatcher;
+
+    private OnmsTwinSubscriber onmsTwinSubscriber;
 
     public TrapListener(final TrapdConfig config) throws SocketException {
         Objects.requireNonNull(config, "Config cannot be null");
@@ -94,6 +99,17 @@ public class TrapListener implements TrapNotificationListener {
     }
 
     public void start() {
+        CompletableFuture<OnmsTwin> future = onmsTwinSubscriber.getObject("trapd-config", new OnmsTwinSubscriber.SinkCallback() {
+            @Override
+            public void sinkUpdate(OnmsTwin onmsTwin) {
+                // Unmarshal and call setSnmpV3Users
+            }
+        });
+        future.whenComplete((response, ex) -> {
+           if(response != null) {
+               // Unmarshal and call setSnmpV3Users
+           }
+        });
         final int m_snmpTrapPort = m_config.getSnmpTrapPort();
         final InetAddress address = getInetAddress();
         try {
@@ -223,5 +239,9 @@ public class TrapListener implements TrapNotificationListener {
             return false;
         }
         return !newConfig.getSnmpV3Users().equals(m_config.getSnmpV3Users());
+    }
+
+    public void setOnmsTwinSubscriber(OnmsTwinSubscriber onmsTwinSubscriber) {
+        this.onmsTwinSubscriber = onmsTwinSubscriber;
     }
 }
--- a/features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapListener.java
+++ b/features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapListener.java
@@ -40,6 +40,7 @@ import org.opennms.core.ipc.sink.api.MessageDispatcherFactory;
 import org.opennms.core.logging.Logging;
 import org.opennms.core.twin.api.OnmsTwin;
 import org.opennms.core.twin.subscriber.api.OnmsTwinSubscriber;
+import org.opennms.core.twin.subscriber.api.TwinSubscriberModule;
 import org.opennms.core.utils.InetAddressUtils;
 import org.opennms.netmgt.config.TrapdConfig;
 import org.opennms.netmgt.config.TrapdConfigFactory;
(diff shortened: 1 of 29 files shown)


cgorantla:
@j-white  Jesse, can you take initial look on this.

cgorantla:
I feel there are too many callbacks in these interfaces.  May be there are better structures. Suggestions welcome!

cgorantla:
btw Twin name came from `Digital Twin`.  We are trying to replicate same object from OpenNMS to Minion 

Minion makes a RPC request to  get specific config/workflow.   Then OpenNMS provides subsequent updates through Sink.

j-white:
Can we try wiring this API into Trapd for the SNMPV3Users? Doesn't need to be functional, but enough to illustrate its use.

This would help me see what it looks like in practice and be able to make better suggestions about modelling.

features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapListener.java
```diff
@@ -94,6 +99,17 @@ public void trapError(int error, String msg) {
     }
 
     public void start() {
+        CompletableFuture<OnmsTwin> future = onmsTwinSubscriber.getObject("trapd-config", new OnmsTwinSubscriber.SinkCallback() {
```
j-white, line 102: Can we make it so that we get the `TrapdConfig` object back in a typesafe fashion and not have to unmarshal it ourselves?

features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapdConfigWrapper.java
```diff
@@ -0,0 +1,80 @@
...
+                return marshalledConfig;
+            }
+
+            @Override
+            public String getKey() {
+                return "trapd-config";
+            }
+
+            @Override
+            public String getLocation() {
```
j-white, line 65: What's the purpose of providing the location in this context?

features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapdConfigWrapper.java
```diff
@@ -0,0 +1,80 @@
...
+        // Marshal trapdconfig and convert it to byte array.
+        byte[] marshalledConfig = marshalTrapdConfig(trapdConfig);
+        OnmsTwin onmsTwin = new OnmsTwin() {
+            @Override
+            public int getVersion() {
+                return 0;
+            }
+
+            @Override
+            public byte[] getObjectValue() {
```
j-white, line 55: Hesitant about only requiring byte arrays here - how will we handle partial updates?

fooker:
Trying to sketch a proposal to remove some callbacks:

I would split the API into two packages/bundles: one for "ONMS" side and one for the "minion" side.
The names should be more like producer/consumer or publisher/subscriber as the "minion"-part may be used on the core, too.

In the "minion"-Side:

```java
class Subscription<T> implements Closable {
	Optional<T> getObject() { ... } // Not sure if this is required

	void close() { ... } // Drops the subscription
}

interface TwinListener<T> {
	void created(T object);
	void updated(T object);
	void deleted(T object);
}

interface TwinBroker {
	# This has to be aware of the location it is running on
	
	SubscriptionT> subscribe<T>(String name, Class<T> clazz, TwinListener<T> listener); // Skip the name and assume the class is unique
}
```

and on the "onms"-side:

```java
interface TwinBroker {
	Wormhole<T> register(String name, Class<T> clazz, String location); // Same as for 'name' above
}

class Wormhole<T> implements Closable {
	String getLocation() { ... }
	String getName() { ... }

	Optional<T> get() { ... } // Returns the currently published version of the object
	void set(T object) { ... } // Updates the object and fires the create/update callback
	
	void close() { ... } // Closes the wormhole and fires the deleted callback
}
```

Note: This is just a quick write-down of what I had in mind and ignores all the things that you may have already encountered.

features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapdConfigWrapper.java
j-white: I think we want to pick which location(s) the object is exposed to right?

features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapListener.java
cgorantla: I can change this to get `TrapdConfig` object by passing specific Class.

features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapdConfigWrapper.java
cgorantla: yeah, for TrapdConfig, this may not be relavant as there is no location specific config here. But when we do, we can target it to specific location.

features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapdConfigWrapper.java
cgorantla: we also expose version.   `OnmsTwinPublisher` should be able to merge with original config depending on version.
If the version is always `0` then we update the whole object always.

features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapListener.java
cgorantla: Updated API with specific types.

cgorantla:
> Trying to sketch a proposal to remove some callbacks:
> 
> I would split the API into two packages/bundles: one for "ONMS" side and one for the "minion" side.
> The names should be more like producer/consumer or publisher/subscriber as the "minion"-part may be used on the core, too.
> 
> In the "minion"-Side:
> 
> ```java
> class Subscription<T> implements Closable {
> 	Optional<T> getObject() { ... } // Not sure if this is required
> 
> 	void close() { ... } // Drops the subscription
> }
> 
> interface TwinListener<T> {
> 	void created(T object);
> 	void updated(T object);
> 	void deleted(T object);
> }
> 
> interface TwinBroker {
> 	// This has to be aware of the location it is running on
> 	
> 	Subscription<T> subscribe<T>(String name, Class<T> clazz, TwinListener<T> listener); // Skip the name and assume the class is unique
> }
> ```
> 
> and on the "onms"-side:
> 
> ```java
> interface TwinBroker {
> 	Wormhole<T> register(String name, Class<T> clazz, String location); // Same as for 'name' above
> }
> 
> class Wormhole<T> implements Closable {
> 	String getLocation() { ... }
> 	String getName() { ... }
> 
> 	Optional<T> get() { ... } // Returns the currently published version of the object
> 	void set(T object) { ... } // Updates the object and fires the create/update callback
> 	
> 	void close() { ... } // Closes the wormhole and fires the deleted callback
> }
> ```
> 
> Note: This is just a quick write-down of what I had in mind and ignores all the things that you may have already encountered.

One thing different in my understanding is that we need to have OnmsTwinPublisher/OnmsTwinSubscriber in between module and broker which can do some heavylifting:

1.  Marshal/Unmarshal from xml/json to bytes.
2. Cache config ( at least on publisher side)
3. Merge incoming updates from modules to provide merged config.

Can you take a look on new commit ( I removed most of callbacks by having separate interface for module) and see if that makes sense.

cgorantla:
Closing this in favor of #3554

PR #3536, opened 2021-07-13
NMS-13444: Added JAR-file to VMware-related scripts

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-13444


RangerRick (approved):
seems reasonable to me 👍

PR #3545, opened 2021-07-16
NMS-13452: Update Jackson 2 Dependencies

RangerRick:
This PR bumps Jackson to 2.12.x.

Tests all pass, but it's probably worth looking into some of the lesser-used bits that this touches -- especially the features.xml files -- I'm not convinced everything gets coverage.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13452


Naicisum:
So these dependencies generally are aligned with the Kafka client library version. Since the smoke tests pass it theoretically should be ok. The Zookeeper update should also be OK but it may be a good idea to ensure that ALEC is not adversely affected...

RangerRick:
> So these dependencies generally are aligned with the Kafka client library version. Since the smoke tests pass it theoretically should be ok. The Zookeeper update should also be OK but it may be a good idea to ensure that ALEC is not adversely affected...

Yeah, that was my concern. I need to do some more complicated testing.

RangerRick:
FYI, going ahead and just closing this because it touches too much stuff, and we are not (AFAICT) exercising the bad codepath.

PR #3554, opened 2021-07-21
NMS-13401: Add Twin (Object replication) API

cgorantla:
Add  TwinPublisher/TwinSubscriber API for Object replication from OpenNMS -> Minion.

Modules on Minion makes RPC request for config/scheduled-job (Twin) and also subscribes to updates to the Twin.
Modules on OpenNMS publish initial config/scheduled-job and provides updates through Reverse Sink.

This PR provides API and  implementation that is not broker specific.
By default abstract layer assumes objects can be marshalled/unmarshalled with json.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13401


core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinPublisher.java
```diff
@@ -0,0 +1,114 @@
...
+        }
+        // TODO: What if there is no value exist here ?
+        return new TwinResponseBean(twinRequest.getKey(), twinRequest.getLocation(), value);
+    }
+
+    private <T> byte[] marshalObject(T obj) throws IOException {
+        try {
+            return objectMapper.writeValueAsBytes(obj);
+        } catch (JsonProcessingException e) {
+            throw e;
```
fooker, line 79: This just re-throws. The try-catch block can be removed.

fooker: In fact, the whole method is only used once. Whereas the in-lined version is used all over the place.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinPublisher.java
```diff
@@ -0,0 +1,114 @@
...
+        }
+    }
+
+    private class SessionImpl<T> implements Session<T> {
+
+        private final String key;
```
fooker, line 85: Use the proposed `SessionKey` here or add the location as a field

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinPublisher.java
```diff
@@ -0,0 +1,114 @@
...
+                throw e;
+            }
+        }
+
+        @Override
+        public void close() throws IOException {
+            String key = keySessionMap.remove(this);
+            if (key != null) {
+                objMap.remove(key);
+            }
```
fooker, line 110: Add signaling for the subscribers here?
```java
handleSinkUpdate(new TwinResponseBean(key, location, null));
```

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
```diff
@@ -0,0 +1,98 @@
...
+import org.opennms.core.ipc.twin.api.TwinSubscriber;
+import org.opennms.distributed.core.api.MinionIdentity;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public abstract class AbstractTwinSubscriber implements TwinSubscriber {
+
+    private final Map<String, Class<?>> classesByKey = new ConcurrentHashMap<>();
+    private final Map<String, Consumer<?>> consumerMap = new ConcurrentHashMap<>();
+    private final Map<Closeable, String> closableMap = new ConcurrentHashMap<>();
```
fooker, line 46: Can we unify this by also having a `SessionImpl` in here.
Something like this:
```java
private class SessionImpl implements Closable {
    private final String key;
    private final Class<?> clazz;
    private final Consumer<?> consumer;

    public void accept(TwinResponseBean twinResponse) throws IOException {
        final var value = objectMapper.readValue(twinResponse.value, this.clazz);
        this.consumer.accpet(value);
    }

    public void close() {
        AbstractTwinSubscriber.this.sessions.remove(this.key);
    }
}
```

This can be used in a single `Map<String, Session>` whereas the `Map`s key is the session key.
`getObject` the could do a `computeIfAbsend` on this `Map` and return the object from the `Map`.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
```diff
@@ -0,0 +1,98 @@
...
+
+import org.opennms.core.ipc.twin.api.TwinSubscriber;
+import org.opennms.distributed.core.api.MinionIdentity;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public abstract class AbstractTwinSubscriber implements TwinSubscriber {
+
+    private final Map<String, Class<?>> classesByKey = new ConcurrentHashMap<>();
+    private final Map<String, Consumer<?>> consumerMap = new ConcurrentHashMap<>();
```
fooker, line 45: Should we support multiple subscribers/consumers per key?

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
```diff
@@ -0,0 +1,98 @@
...
+    private final MinionIdentity minionIdentity;
+
+    protected AbstractTwinSubscriber(MinionIdentity minionIdentity) {
+        this.minionIdentity = minionIdentity;
+    }
+
+    /**
+     * @param twinRequest Handle RpcRequest from @{@link AbstractTwinSubscriber}
+     */
+    abstract void handleRpcRequest(TwinRequestBean twinRequest);
```
fooker, line 57:
```suggestion
    abstract void sendRpcRequest(TwinRequestBean twinRequest);
```

fooker (changes requested):
Haven't looked at the tests yet. Would be nice to have a test suite that can be used for all broker implementations. But this can be added / adopted over time.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinPublisher.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+
+    protected TwinResponseBean getTwin(TwinRequestBean twinRequest) {
+        byte[] value = objMap.get(twinRequest.getKey() + twinRequest.getLocation());
+        if (value == null) {
+            value = objMap.get(twinRequest.getKey());
+        }
+        // TODO: What if there is no value exist here ?
+        return new TwinResponseBean(twinRequest.getKey(), twinRequest.getLocation(), value);
```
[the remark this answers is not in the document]
RangerRick: Should we avoid returning null in general nowadays?  Make it `Optional` instead?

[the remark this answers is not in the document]
fooker: In pure form of returning, yes. But this is passed as parameter to the consumer. There it is considered by behavior (I recently learned) as it may makes the situation even worth: The parameter may be null or empty.

cgorantla:
> Haven't looked at the tests yet. Would be nice to have a test suite that can be used for all broker implementations. But this can be added / adopted over time.

Currently IT handles publish/subscribe on few kafka topics.  It was good to get a feel for what implementations could look like. We can adopt this later if this needs in-memory implementation.

core/ipc/twin/api/src/main/java/org/opennms/core/ipc/twin/api/TwinPublisher.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+public interface TwinPublisher {
+
+    /**
+     * Session that can publish updates to T
+     *
+     * @param <T> type of object that needs replication.
+     */
+    interface Session<T> extends Closeable {
```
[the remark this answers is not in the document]
fooker: Ok, that's fine. I just wanted to clarify that there is no signaling happening if the session is closed.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
fooker: I can foresee some (hypothetical) use-cases where multiple subscribers may be needed on the same minion. But I don't think this is critical for now as it do not affect the API. We can change this afterwards if needed.

Distributing jobs across minions is out-of-scope for now. This would need some kind of cluster-management for the minions in a location which may be module specific and not part of the Twin API.

fooker (approved):

PR #3556, opened 2021-07-23
NMS-13469: fix ReST queries for OnmsIpInterface snmpPrimary

RangerRick:
This PR does a bunch of plumbing work to fix querying nodes and interfaces based on their SNMP primary status.  Note: this work is necessary to complete the work for HELM-188.

- add `OnmsIpInterface#getSnmpPrimary` and move the hibernate annotation to it to make ReST mapping happy, `#getIsSnmpPrimary` is left in place so that a bunch of code doesn't need changing (including end-user groovy code); this includes creating a `CharacterUserType` adapter for Hibernate
- change PrimaryType to not be an `Embedded` Hibernate column, instead just make it a property of `OnmsIpInterface` to avoid API changes
- refactor `PrimaryTypeAdapter` from Provisiond into `opennms-model` so it can be live with other type adapters
- fix internal HQL queries to use `.snmpPrimary` rather than `.isSnmpPrimary`
- update `PrimaryType#get` to accept a String _or_ Character/char
- add "snmpPrimary" checks to both the `NodeRestServiceIT` and `IpInterfaceRestServiceIT`
- add search properties for PrimaryType so Helm displays the choices

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13469

### DIFF
--- a/features/enlinkd/persistence/api/src/main/java/org/opennms/netmgt/enlinkd/model/IpInterfaceTopologyEntity.java
+++ b/features/enlinkd/persistence/api/src/main/java/org/opennms/netmgt/enlinkd/model/IpInterfaceTopologyEntity.java
@@ -58,7 +58,7 @@ public class IpInterfaceTopologyEntity {
     }
 
     public IpInterfaceTopologyEntity(Integer id,
-            InetAddress ipAddress, String isManaged, Character snmpPrimary, Integer nodeId,
+            InetAddress ipAddress, String isManaged, String snmpPrimary, Integer nodeId,
             Integer snmpInterfaceId){
         this(id, ipAddress, isManaged, PrimaryType.get(snmpPrimary), nodeId, snmpInterfaceId);
     }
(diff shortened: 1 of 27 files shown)


features/enlinkd/persistence/api/src/main/java/org/opennms/netmgt/enlinkd/model/IpInterfaceTopologyEntity.java
```diff
@@ -57,6 +57,17 @@ public IpInterfaceTopologyEntity(Integer id,
...
+            InetAddress ipAddress, String isManaged, Character snmpPrimary, Integer nodeId,
+            Integer snmpInterfaceId){
+        this.id = id;
+        this.ipAddress = ipAddress;
+        this.isManaged = isManaged;
+        this.isSnmpPrimary = PrimaryType.get(snmpPrimary);
+        this.nodeId = nodeId;
+        this.snmpInterfaceId = snmpInterfaceId;
+    }
+
```
fooker, line 70:
```suggestion
    public IpInterfaceTopologyEntity(Integer id,
            InetAddress ipAddress, String isManaged, Character snmpPrimary, Integer nodeId,
            Integer snmpInterfaceId){
        this(id, ipAddress, isManaged, PrimaryType.get(snmpPrimary), nodeId, snmpInterfaceId);
    }

```

fooker (changes requested):

christianpape (approved):
LGTM.

fooker (approved):

cgorantla (commented):
Looks good.
I see that we have this limitation of 1 char column on DB. 
That makes it hard to modify.

It would be good to add a test case on IpInterfaceDao for PrimaryType validation.

PR #3558, opened 2021-07-26
NMS-13436: fix handling of permissions/ownership when running as root

RangerRick:
This PR changes the Horizon Docker image to use `rsync` for all overlay operations, including using `--chown` to make the ownership match the container user.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13436

### DIFF
--- a/opennms-container/horizon/Dockerfile
+++ b/opennms-container/horizon/Dockerfile
@@ -71,7 +71,7 @@ RUN echo "installing packages for build ${REVISION} (${BUILD_DATE})"; \
     if [[ "$(ls -1 /tmp/tarball/*.tar.gz 2>/dev/null | wc -l)" != 0 ]]; then \
         mkdir -p /opt/opennms && \
         tar -xzf /tmp/tarball/opennms-*.tar.gz -C /opt/opennms && \
-        chown -R 0:0 /opt/opennms && \
+        chown -R 10001:0 /opt/opennms && \
         cp -r /opt/opennms/etc /opt/opennms/share/etc-pristine; \
     elif [[ "$(ls -1 /tmp/rpms/*.rpm 2>/dev/null | wc -l)" != 0 ]]; then \
         echo "installing local RPMs from filesystem" && \
--- a/opennms-bootstrap/src/main/java/org/opennms/bootstrap/FilesystemPermissionException.java
+++ b/opennms-bootstrap/src/main/java/org/opennms/bootstrap/FilesystemPermissionException.java
@@ -41,7 +41,7 @@ public class FilesystemPermissionException extends Exception {
     }
 
     public FilesystemPermissionException(final Path path, final String user, final Exception e) {
-        super("Failed to validate " + path + " (or its contents) for ownership by " + user, e);
+        super("Failed to validate " + path + " (or its contents) for read/write by " + user, e);
         this.path = path;
         this.user = user;
     }
--- a/opennms-bootstrap/src/main/java/org/opennms/bootstrap/FilesystemPermissionValidator.java
+++ b/opennms-bootstrap/src/main/java/org/opennms/bootstrap/FilesystemPermissionValidator.java
@@ -40,6 +40,7 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Random;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class FilesystemPermissionValidator {
     public static boolean VERBOSE = false;
@@ -65,7 +66,7 @@ public class FilesystemPermissionValidator {
         }
 
         if (VERBOSE) {
-            System.out.println("Validating: user " + user + " owns " + path + "\n");
+            System.out.println("Validating: user " + user + " can write to " + path + "\n");
         }
 
         for (final String dir : FULLDIRS) {
@@ -131,9 +132,10 @@ public class FilesystemPermissionValidator {
                 System.out.println("- checking " + dirPath);
             }
         }
+
         final List<Path> failures;
-        try {
-            failures = Files.walk(dirPath, FileVisitOption.FOLLOW_LINKS).filter(file -> {
+        try (final Stream<Path> stream = Files.walk(dirPath, FileVisitOption.FOLLOW_LINKS)) {
+            failures = stream.filter(file -> {
                 try {
                     validateFile(user, file);
                 } catch (final FilesystemPermissionException e) {
@@ -144,25 +146,23 @@ public class FilesystemPermissionValidator {
         } catch (final IOException e) {
             throw new FilesystemPermissionException(dirPath, user, e);
         }
-        if (failures.size() > 0) {
+        if (!failures.isEmpty()) {
             throw new FilesystemPermissionException(failures.get(0), user);
         }
     }
 
     private void validateFile(final String user, final Path filePath) throws FilesystemPermissionException {
         if (VERBOSE) {
-            System.out.println("- validating file " + filePath + " is owned by " + user);
+            System.out.println("- validating file " + filePath + " is writable by " + user);
         }
-        try {
-            final UserPrincipal owner = Files.getOwner(filePath);
-            if (VERBOSE) {
-                System.out.println("  - " + filePath + " is owned by " + owner);
-            }
-            if (!Objects.equals(owner.getName(), user)) {
+        var file = filePath.toFile();
+        if (!file.canRead() || !file.canWrite()) {
+            throw new FilesystemPermissionException(filePath, user);
+        }
+        if (file.isDirectory()) {
+            if (!file.canExecute()) {
                 throw new FilesystemPermissionException(filePath, user);
             }
-        } catch (final IOException e) {
-            throw new FilesystemPermissionException(filePath, user, e);
         }
     }
 
@@ -185,7 +185,7 @@ public class FilesystemPermissionValidator {
         try {
             validator.validate(user, path);
         } catch (final FilesystemPermissionException e) {
-            System.err.println("\nERROR: at least one file in " + path + " is not owned by " + user + ".");
+            System.err.println("\nERROR: at least one file in " + path + " is not writable by " + user + ".");
             System.err.println("Make sure $RUNAS in opennms.conf matches the permissions of " + path +
                     " and its subdirectories.\n");
             e.printStackTrace();
--- a/opennms-container/horizon/container-fs/entrypoint.sh
+++ b/opennms-container/horizon/container-fs/entrypoint.sh
@@ -23,6 +23,16 @@ OPENNMS_OVERLAY_JETTY_WEBINF="/opt/opennms-jetty-webinf-overlay"
 E_ILLEGAL_ARGS=126
 E_INIT_CONFIG=127
 
+MYID="$(id -u)"
+MYUSER="$(getent passwd "${MYID}" | cut -d: -f1)"
+
+export RUNAS="${MYUSER}"
+
+if [ "$MYID" -eq 0 ]; then
+  echo "RUNAS=${MYUSER}" >> /opt/opennms/etc/opennms.conf
+  chown "$MYUSER" /opt/opennms/etc/opennms.conf
+fi
+
 # Help function used in error messages and -h option
 usage() {
   echo ""
--- a/opennms-install/src/main/java/org/opennms/install/Installer.java
+++ b/opennms-install/src/main/java/org/opennms/install/Installer.java
@@ -591,7 +591,7 @@ public class Installer {
         try {
             validator.validate(user, opennmsHome);
         } catch (final FilesystemPermissionException e) {
-            LOG.error("OpenNMS is configured to run as '{}' but '{}' is not owned by that account.", user, e.path);
+            LOG.error("OpenNMS is configured to run as '{}' but '{}' is not writable by that account.", user, e.path);
             LOG.error("To fix permissions, run '{}/bin/fix-permissions' as root", m_opennms_home);
             System.out.println();
 
--- a/tools/packages/opennms/opennms.spec
+++ b/tools/packages/opennms/opennms.spec
@@ -984,7 +984,9 @@ echo "done"
 if [ ! -e "$ROOT_INST/etc/java.conf" ]; then
 	"$ROOT_INST/bin/runjava" "-s"
 fi
-"${ROOT_INST}/bin/fix-permissions" -R "${ROOT_INST}/etc/java.conf"
+if [ -e "${ROOT_INST}/etc/java.conf" ]; then
+	"${ROOT_INST}/bin/fix-permissions" -R "${ROOT_INST}/etc/java.conf"
+fi
 
 "${ROOT_INST}/bin/update-package-permissions" "%{name}-core"
 "${ROOT_INST}/bin/ensure-user-ping.sh" || echo "WARNING: Unable to enable ping by the opennms user. Try running /usr/share/opennms/bin/ensure-user-ping.sh manually."


opennms-container/horizon/Dockerfile
```diff
@@ -71,7 +71,7 @@ RUN echo "installing packages for build ${REVISION} (${BUILD_DATE})"; \
     if [[ "$(ls -1 /tmp/tarball/*.tar.gz 2>/dev/null | wc -l)" != 0 ]]; then \
         mkdir -p /opt/opennms && \
         tar -xzf /tmp/tarball/opennms-*.tar.gz -C /opt/opennms && \
-        chown -R 0:0 /opt/opennms && \
+        chown -R 10001:10001 /opt/opennms && \
```
indigo423, line 74: I remember we have introduced this with the requirement to run in OpenShift which needs support to run with arbitrary user ids. @agalue can you double-check here if this is still the case?

indigo423 (approved):
We should double-check with @agalue if we don't break the OpenShift use case he brought in. The permission changes might break it, but I'm not familiar enough with it.

agalue:
Not sure how urgent is this, but it will take a while until I find time to have an OpenShift cluster to test this out. I can definitively test it in Kubernetes, but things changed a lot in the OpenShift realm. Because I have limited resources in cloud services (for a variety of reasons not pertinent to this PR), the OpenShift tests I can offer will also be limited to version 3.x via `minishift` using the Open Source version, as 4.x is only available in the cloud (or paid RedHat subscribers), AFAIK.

agalue:
As I had some spare time today, here is what I did:

1. Start a single-instance OpenShift 3.11 cluster:
```bash
minishift start --cpus 4 --memory 16GB
```

2. Log into the private registry in OpenShift:
```bash
eval $(minishift docker-env)
docker login -u developer -p $(oc whoami -t) $(minishift openshift registry)
```

3. Download the horizon.oci from CircleCI for the branch in question.

4. Upload the Horizon OCI to your local Docker instance.
```bash
docker image load -i horizon.oci
```

5. Tag and push the image:
```bash
docker tag horizon:latest $(minishift openshift registry)/myproject/horizon:latest
docker push $(minishift openshift registry)/myproject/horizon:latest
```

7. Start a PostgreSQL Server:
```bash
oc new-app --name=postgresql --template=postgresql-persistent
```

8. Extract the PostgreSQL credentials:
```bash
DB_USER=$(oc get secret postgresql -o jsonpath='{.data.database\-user}' | base64 -d)
DB_PASS=$(oc get secret postgresql -o jsonpath='{.data.database\-password}' | base64 -d)
```

9. Start OpenNMS:
```bash
oc new-app --image-stream=horizon --name=horizon -e POSTGRES_HOST=postgresql -e POSTGRES_USER=$DB_USER -e POSTGRES_PASSWORD=$DB_PASS
```

Because the OpenNMS image does nothing without arguments, you need to manually edit the deployment and set the arguments:
```
oc edit deployment horizon
```
Then, find `image:` and add the following under it and save it:
```
args: [ -s ]
```

10. Validate
```
oc status
```
or...
```
oc get pod -l deployment=horizon
```

But, it failed :)

openshift-complains

mmahacek:
> But, it failed :)

I've been testing a number of builds, and that failure error is similar to what I've been getting, though my latest error is the container not able to connect to the postgres instance that is running on the same host.

agalue:
I can definitely try in plain Kubernetes and plain Docker just for fun. I'll report back in a few minutes.

mmahacek:
> I can definitely try in plain Kubernetes and plain Docker just for fun. I'll report back in a few minutes.

I've tried Docker on Windows and podman on Cent8.  With the latter, I get permission denied errors that it can't create each of the pristine config files.

agalue:
Using the latest Docker for Mac (with the integrated compose):

1) Upload the OCI image:
```bash
docker image load -i horizon.oci
```

2) Create the compose file (new way without versioning):
```yaml
services:

  database:
    container_name: database
    hostname: database
    image: postgres:12
    environment:
    - TZ=America/New_York
    - POSTGRES_HOST=database
    - POSTGRES_PORT=5432
    - POSTGRES_USER=postgres
    - POSTGRES_PASSWORD=postgres
    healthcheck:
      test: ["CMD-SHELL", "pg_isready -U postgres"]
      interval: 30s
      timeout: 5s
      retries: 3

  opennms:
    container_name: opennms
    hostname: opennms
    image: horizon:latest
    depends_on:
      database:
        condition: service_healthy
    command: [ -s ]
    ports:
    - 8980:8980
    - 8101:8101
    environment:
    - TZ=America/New_York
    - POSTGRES_HOST=database
    - POSTGRES_PORT=5432
    - POSTGRES_USER=postgres
    - POSTGRES_PASSWORD=postgres
    - JAVA_OPTS=-Xms4096m -Xmx4096m -XX:+UseG1GC -XX:+UseStringDeduplication
    healthcheck:
      test: curl -f -I http://localhost:8980/opennms/login.jsp
      interval: 1m
      timeout: 5s
      retries: 3
```

3) Start the environment (using the embedded compose):
```bash
docker compose up -d
```

4) It works for me:
```bash
➜  docker compose ps
NAME                COMMAND                  SERVICE             STATUS              PORTS
database            "docker-entrypoint.s…"   database            running (healthy)   5432/tcp
opennms             "/entrypoint.sh -s"      opennms             running (healthy)   1162/udp, 0.0.0.0:8101->8101/tcp, :::8101->8101/tcp, 0.0.0.0:8980->8980/tcp, :::8980->8980/tcp, 10514/udp

➜  curl -u admin:admin http://localhost:8980/opennms/rest/info 2>/dev/null | jq
{
  "displayVersion": "29.0.0-SNAPSHOT",
  "version": "29.0.0",
  "packageName": "opennms",
  "packageDescription": "OpenNMS",
  "ticketerConfig": {
    "plugin": null,
    "enabled": false
  },
  "datetimeformatConfig": {
    "zoneId": "America/New_York",
    "datetimeformat": "yyyy-MM-dd'T'HH:mm:ssxxx"
  },
  "services": {}
}
```

agalue:
Perhaps the problem is when using custom volumes for the configuration directory? That's a different use case.

mmahacek:
> Perhaps the problem is when using custom volumes for the configuration directory? That's a different use case.

I was doing your favorite thing - copying the config steps from the documentation.

agalue:
No idea as I don't trust the documentation :)
I can try that, but only if I'm convinced it makes sense.

mmahacek:
I'm trying your docker-compose.yml now

agalue:
Don't forget to change the image, as the documentation points to the latest one in Docker Hub which is NOT what you should be testing, if you want to validate the change on this PR.

agalue:
The following is what the documentation should have (with the updated image), and it works:
```yaml
---
version: '3'

volumes:
  data-postgres: {}
  data-opennms: {}

services:
  database:
    image: postgres:13
    container_name: database
    environment:
      TZ: 'America/New_York'
      POSTGRES_USER: 'postgres'
      POSTGRES_PASSWORD: 'my-postgres-password'
    volumes:
      - 'data-postgres:/var/lib/postgresql/data'
    healthcheck:
      test: [ "CMD-SHELL", "pg_isready -U postgres" ]
      interval: 10s
      timeout: 3s
      retries: 3
    ports:
      - '5432:5432/tcp'

  horizon:
    image: horizon:latest
    container_name: horizon
    environment:
      TZ: 'America/New_York'
      POSTGRES_HOST: 'database'
      POSTGRES_PORT: 5432
      POSTGRES_USER: 'postgres'
      POSTGRES_PASSWORD: 'my-postgres-password'
      OPENNMS_DBNAME: 'opennms-core-db'
      OPENNMS_DBUSER: 'opennms'
      OPENNMS_DBPASS: 'my-opennms-db-password'
    volumes:
      - data-opennms:/opennms-data
      - ./etc:/opt/opennms/etc
    command: ["-s"]
    ports:
      - '8980:8980/tcp'
      - '8101:8101/tcp'
    healthcheck:
      test: [ 'CMD', 'curl', '-f', '-I', 'http://localhost:8980/opennms/login.jsp' ]
      interval: 1m
      timeout: 5s
      retries: 3
```
**PRO TIP**: `POSTGRES_HOST` is wrong in the documentation (careful with copy/paste) :)

I'll try Kubernetes next.

mmahacek:
I can get the containers to work if I don't define any `volumes` in my docker-compose.  When I do define the overlay folders, I get:
```
09:46:51.649 [Main] ERROR org.opennms.install.Installer - OpenNMS is configured to run as 'opennms' but '/opt/opennms/etc' is not owned by that account.
```

RangerRick:
> I remember we have introduced this with the requirement to run in OpenShift which needs support to run with arbitrary user ids. @agalue can you double-check here if this is still the case?

This is not a requirement I knew about. What exactly does "run with arbitrary user ids" look like?  Is the user still called `opennms` but has a different uid?  Or the other way around?

All our `RUNAS=` requirements in the OpenNMS code technically only care about username, and I only kept referring to `10001` in my modifications to match the existing stuff.

RangerRick:
One option would be to make RUNAS and the UID configurable, and then enforce ownership-changing in `entrypoint.sh`.  Of course, doing so would probably necessitate _also_ forcing a deep-recursive chown/chmod which I've been trying to avoid since it would be looking in share/rrd and such and could potentially take a very long time.  Maybe need an option in the minion docker entrypoint.sh options to "fix ownership" as a user-initiated step?

agalue:
I can revisit the OpenShift deployment to tell you exactly what the `PodPolicy` is enforcing and to see how the effective `uid` and `gid` are being used (for the process and the volumes). I can only tell you that it was working before. I can verify it if you're interested.

If OpenNMS decides to drop OpenShift support, I'm good with it, as long as the development team will be responsible for fixing it in the future when a customer requires it.

agalue:
1) Start a local Kubernetes Cluster:
```bash
minikube start --cpus=2 --memory=8g
```

2) Point your Docker environment to the one at Minikube
```bash
eval $(minikube docker-env)
```

3) Upload the image to the local Docker within Minikube:
```bash
docker image load -i horizon.oci
docker tag horizon:latest horizon:PR3558
```
**Note**: I tagged the image to enforce the fact that we will be testing it.

4) Start a Pod for PostgreSQL
```bash
kubectl run database --image postgres:12 --port 5432 --env POSTGRES_HOST=database --env POSTGRES_PORT=5432 --env POSTGRES_USER=postgres --env POSTGRES_PASSWORD=postgres --expose
```

5) Start a Pod for OpenNMS (using the above image)
```bash
kubectl run opennms --image horizon:PR3558 --port 8980 --env POSTGRES_HOST=database --env POSTGRES_PORT=5432 --env POSTGRES_USER=postgres --env POSTGRES_PASSWORD=postgres --expose -- -s
```
**NOTE**: The above doesn't create volumes, I'll try that next.

6) Verify
```bash
kubectl get pods
```

**It works!**

Now with volumes as the documentation says, but first, delete the `opennms` pod and service:
```bash
kubectl delete pod,svc opennms
```

Then, create the YAML manifest for OpenNMS:
```bash
kubectl run opennms --image horizon:PR3558 --port 8980 --env POSTGRES_HOST=database --env POSTGRES_PORT=5432 --env POSTGRES_USER=postgres --env POSTGRES_PASSWORD=postgres --expose --dry-run=client -o yaml -- -s > PR3558.yaml
```

Then, modifiy it to have volumes for `/opt/opennms/etc` and `/opt/opennms/data` (both empty).
I added:
```yaml
...
spec:
  volumes:
  - name: data
    emptyDir: {}
  - name: config
    emptyDir: {}
  containers:
...
    volumeMounts:
    - name: data
      mountPath: /opennms-data
    - name: config
      mountPath: /opt/opennms/etc
...
```

**And, that doesn't work :(**
```bash
kubectl logs opennms
...
17:14:02.661 [Main] WARN  org.opennms.install.Installer - Could not create file: /opt/opennms/etc/libraries.properties
17:14:02.698 [Main] ERROR org.opennms.install.Installer - OpenNMS is configured to run as 'opennms' but '/opt/opennms/etc' is not owned by that account.
17:14:02.698 [Main] ERROR org.opennms.install.Installer - To fix permissions, run '/opt/opennms/bin/fix-permissions' as root
```

I *think* OpenShift would do something similar under the hood, but with restricted permissions privileges, users, groups, etc.
I'm going to do it again to extract how OpenShift configures the run-time pod and see if I can reproduce it with plain Kubernetes; so developers can see what it does, and it is up to them to decide if our image will support OpenShift or not; as I only need to know who to blame when a customer asks for it.

agalue:
So, with plain Kubernetes (without all the extra security imposed by OpenShift), it fails as it did for @mmahacek with plain Docker, and we should fix it (regardless of the OpenShift support).

agalue:
I think it is better to show with a simple example how permissions work in OpenShift. Let's say I create a single Pod like the following:
```bash
➜  ~ cat <<EOF | oc apply -f -
apiVersion: v1
kind: Pod
metadata:
  labels:
    run: test
  name: test
spec:
  volumes:
  - name: data
    emptyDir: {}
  containers:
  - image: busybox
    name: test
    args: [ 'sleep', '3600' ]
    volumeMounts:
    - name: data
      mountPath: /data
  restartPolicy: Never
EOF
```

In plain Kubernetes, that would run the `sleep` command as `root`, but in OpenShift:
```bash
➜  ~ oc exec -it test sh
/ $ ps -ef
PID   USER     TIME  COMMAND
    1 10001400  0:00 sleep 3600
    8 10001400  0:00 sh
   14 10001400  0:00 ps -ef
/ $ id
uid=1000140000 gid=0(root) groups=1000140000
/ $ ls -ld /data/
drwxrwsrwx    2 root     10001400         6 Aug  4 19:21 /data/
/ $ exit
```

As you can see, OpenShift has chosen an arbitrary `uid`. Note the "additional" group ID. That comes from the `fsGroup` assigned at runtime due to the volume attached to `/data`. In fact:
```bash
➜  ~ oc get pod test -o yaml
...
spec:
  containers:
...
    name: test
    securityContext:
      capabilities:
        drop:
        - KILL
        - MKNOD
        - SETGID
        - SETUID
      runAsUser: 1000140000
...
  securityContext:
    fsGroup: 1000140000
    seLinuxOptions:
      level: s0:c12,c4
```
The first `securityContext` section belongs to the container, whereas the second section belongs to the Pod. At runtime, both are merged. The Pod settings apply to all the containers defined, and the container settings override or expand the pod settings.

That's an example of what Pod Policies enforced by OpenShift by default, something that doesn't happen in plain Kubernetes.

You might think, well, I can set the user ID to whatever I want in the YAML manifest, right? Well not exactly, if I do the following, the `uid=1000` is accepted as it is NOT root, but not the `gid` (it will still be zero). As I didn't added a `securityContext.fsGroup` at the Pod Level look what happens to the volume:

```bash
➜  ~ cat <<EOF | oc apply -f -
apiVersion: v1
kind: Pod
metadata:
  labels:
    run: test2
  name: test2
spec:
  volumes:
  - name: data
    emptyDir: {}
  containers:
  - image: busybox
    name: test
    args: [ 'sleep', '3600' ]
    securityContext:
      runAsUser: 1000
      runAsGroup: 1000
    volumeMounts:
    - name: data
      mountPath: /data
  restartPolicy: Never
EOF
pod/test2 created

➜  ~ oc exec -it test2 sh
/ $ ps -ef
PID   USER     TIME  COMMAND
    1 1000      0:00 sleep 3600
    8 1000      0:00 sh
   17 1000      0:00 ps -ef
/ $ id
uid=1000 gid=0(root)
/ $ ls -ld /data/
drwxrwxrwx    2 root     root             6 Aug  4 19:29 /data/
```

Please note that if you dare to enforce `runAsUser: 0`, that will fail (nothing can run as root) :)

Now I think it should be more clear the differences between Kubernetes and OpenShift at runtime. Yes, you can define PodSecurityPolicies in plain Kubernetes to emulate something like that, but that falls outside the scope here.

So, be careful with the `USER` entry in `Dockerfile`, as OpenShift should inspect the images before running them, which is why if it is `root`, when you create the application it would show:

```
 * WARNING: Image "myproject/horizon:latest" runs as the 'root' user which may not be permitted by your cluster administrator
```

agalue:
One thing I never tried with Kubernetes is the overlay directory, as I handle that in a more advanced way via `initContainers` to populate properly the persistent volume with the configuration directory and be able to handle upgrades auto-magically. I only used the overlay directory via Docker.

When using Docker, you literally mount a directory on your machine inside the container. Unfortunately, that is NOT the case with Kubernetes.

Something similar is achievable mounting config-maps as volumes, but tools like `rsync` might not work as those are symlinks.

Anyway, let's try it:

```yaml
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: onms-config
data:
  rrd.properties: |
    org.opennms.rrd.storeByGroup=true
    org.opennms.rrd.storeByForeignSource=true
    org.opennms.rrd.strategyClass=org.opennms.netmgt.rrd.rrdtool.MultithreadedJniRrdStrategy
    org.opennms.rrd.interfaceJar=/usr/share/java/jrrd2.jar
    opennms.library.jrrd2=/usr/lib64/libjrrd2.so
  web.properties: |
    org.opennms.netmgt.jetty.port=80
    opennms.rtc-client.http-post.base-url=http://localhost:80/opennms/rtc/post
    opennms.map-client.http-post.url = http://localhost:80/opennms/map/post

---
apiVersion: v1
kind: Service
metadata:
  name: opennms
spec:
  ports:
  - port: 80
    protocol: TCP
    targetPort: 80
  selector:
    run: opennms

---
apiVersion: v1
kind: Pod
metadata:
  labels:
    run: opennms
  name: opennms
spec:
  volumes:
  - name: config
    configMap:
      name: onms-config
  containers:
  - name: opennms
    image: horizon:PR3558
    args: [ -s ]
    env:
    - name: POSTGRES_HOST
      value: database
    - name: POSTGRES_PORT
      value: "5432"
    - name: POSTGRES_USER
      value: postgres
    - name: POSTGRES_PASSWORD
      value: postgres
    ports:
    - containerPort: 80
    volumeMounts:
    - name: config
      mountPath: /opt/opennms-etc-overlay/opennms.properties.d
```

Note that this time, I'm not touching the configuration directory, but mounting a configMap as the overlay, meaning each entry from the configMap will be a file.

The above works with the latest image from this PR as explained before (including having the image in Minikube, and PostgreSQL running).

```bash
➜  kubectl exec -it opennms -- cat /opt/opennms/etc/opennms.properties.d/rrd.properties
org.opennms.rrd.storeByGroup=true
org.opennms.rrd.storeByForeignSource=true
org.opennms.rrd.strategyClass=org.opennms.netmgt.rrd.rrdtool.MultithreadedJniRrdStrategy
org.opennms.rrd.interfaceJar=/usr/share/java/jrrd2.jar
opennms.library.jrrd2=/usr/lib64/libjrrd2.so

➜  kubectl get pod
NAME       READY   STATUS    RESTARTS   AGE
database   1/1     Running   0          6m
opennms    1/1     Running   0          2m48s

➜  kubectl get svc
NAME         TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)    AGE
database     ClusterIP   10.105.149.204   <none>        5432/TCP   5m
kubernetes   ClusterIP   10.96.0.1        <none>        443/TCP    12m
opennms      ClusterIP   10.98.15.45      <none>        80/TCP     108s
➜  minikube ssh
                         _             _
            _         _ ( )           ( )
  ___ ___  (_)  ___  (_)| |/')  _   _ | |_      __
/' _ ` _ `\| |/' _ `\| || , <  ( ) ( )| '_`\  /'__`\
| ( ) ( ) || || ( ) || || |\`\ | (_) || |_) )(  ___/
(_) (_) (_)(_)(_) (_)(_)(_) (_)`\___/'(_,__/'`\____)

$ curl -u admin:admin http://10.98.15.45/opennms/rest/info
{"displayVersion":"29.0.0-SNAPSHOT","version":"29.0.0","packageName":"opennms","packageDescription":"OpenNMS","ticketerConfig":{"plugin":null,"enabled":false},"datetimeformatConfig":{"zoneId":"UTC","datetimeformat":"yyyy-MM-dd'T'HH:mm:ssxxx"},"services":{}}
$
```

agalue:
I should have been started with the basics.

If you follow the `minishift` procedure again, but instead of uploading the custom horizon image, you use the official one for 28.0.1 from Docker Hub, it works :)

That should serve as a reference for whatever work has to be done for H29.

Here is how I started OpenNMS that way:

```bash
oc new-app --docker-image=opennms/horizon:28.0.1 --name=horizon \
  -e POSTGRES_HOST=postgresql \
  -e POSTGRES_USER=postgres \
  -e POSTGRES_PASSWORD=postgres
```

Note there is no warning about "running as root".

Remember to edit the `deploymentconfig` to add `args: [ -s]` (we should have a default to avoid this, honestly).

H28-working-in-openshift

RangerRick:
So I _think_ this is my final (force) push. I undid a lot of the work I did, since it turns out it is unnecessary. Every bit of work I did to try harder to get the filesystem in the right state just added more things to incorrectly interact with the filesystem.  :D

After lots of testing back and forth with Alejandro, it turns out the _actual_ problem is that ownership inside the container for mounted volumes is funny -- it'll often be the wrong user (root), but marked `777` so it's still writable.  This final version of the patch does some explicit setting of `RUNAS=` just to be sure it matches the container user, and then also changes the Java validation code to look for writability rather than ownership.

mmahacek (approved):
Works well in Docker on Windows.

agalue:
It doesn't work in OpenShift :)
```
Use existing Mibs data directory.
Use existing Reports data directory.
Use existing RRD data directory.
Processing confd templates using /etc/confd/confd.toml
2021-08-16T18:54:39Z horizon-2-h2htl confd[13]: INFO Backend set to env
2021-08-16T18:54:39Z horizon-2-h2htl confd[13]: INFO Starting confd
2021-08-16T18:54:39Z horizon-2-h2htl confd[13]: INFO Backend source(s) set to
No custom config found in /opt/opennms-overlay. Use default configuration.
No custom config found in /opt/opennms-etc-overlay. Use default configuration.
No custom Jetty WEB-INF config found in /opt/opennms-jetty-webinf-overlay. Use default configuration.
Run config tester to validate existing configuration files.
18:54:40.546 [Main] INFO  org.opennms.netmgt.config.tester.ConfigTesterFilterDao - Initializing ConfigTesterFilterDao
18:54:40.642 [Main] INFO  org.springframework.context.support.ClassPathXmlApplicationContext - Refreshing org.springframework.context.support.ClassPathXmlApplicationContext@10baa53e: startup date [Mon Aug 16 18:54:40 UTC 2021]; root of context hierarchy
18:54:40.682 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/org.opennms.core.soa-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:40.815 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/opennms-ackd-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:40.835 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/opennms-alarmd-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:40.848 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/opennms-asterisk-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:40.862 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/opennms-config-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:40.873 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/opennms-config-tester-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:40.883 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/opennms-correlator-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:40.894 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/opennms-dao-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:40.904 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/opennms-jetty-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:40.913 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/opennms-provisiond-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:40.922 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/opennms-reportd-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:40.930 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/opennms-reporting-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:40.939 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/opennms-services-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:40.950 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/opennms_system_report.jar!/beanRefContext.xml]
18:54:40.958 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/org.opennms.core.daemon-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:40.966 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/org.opennms.features.bsm.daemon-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:40.974 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/org.opennms.features.discovery-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:40.981 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/org.opennms.features.enlinkd.daemon-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:40.989 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/org.opennms.features.events.daemon-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:40.997 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/org.opennms.features.events.syslog-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:41.004 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/org.opennms.features.events.traps-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:41.012 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/org.opennms.features.karaf-health-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:41.019 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/org.opennms.features.perspectivepoller-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:41.028 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/org.opennms.features.reporting.availability-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:41.035 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/org.opennms.features.reporting.core-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:41.041 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/org.opennms.features.reporting.jasper-reports-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:41.049 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/org.opennms.features.reporting.repository-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:41.055 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/org.opennms.features.telemetry.daemon-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:41.062 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from URL [jar:file:/opt/opennms/lib/org.opennms.features.ticketing.daemon-29.0.0-SNAPSHOT.jar!/beanRefContext.xml]
18:54:41.142 [Main] INFO  org.springframework.context.support.ClassPathXmlApplicationContext - Refreshing org.springframework.context.support.ClassPathXmlApplicationContext@4aefe63: startup date [Mon Aug 16 18:54:41 UTC 2021]; root of context hierarchy
18:54:41.143 [Main] INFO  org.springframework.beans.factory.xml.XmlBeanDefinitionReader - Loading XML bean definitions from class path resource [META-INF/opennms/applicationContext-configTester.xml]
18:54:41.967 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded Ackd Configuration in 596 ms
18:54:42.055 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded OpenNMS Report Configuration in 32 ms
18:54:42.070 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded Config for BSF Northbounder in 11 ms
18:54:42.182 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded Database Report Configuration in 7 ms
18:54:42.603 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded data-collection in 375 ms
18:54:42.669 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded Config for Drools Northbounder in 12 ms
18:54:44.406 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded Config for E-mail Northbounder in 5 ms
18:54:44.506 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded Javamail configuration in 46 ms
18:54:44.520 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded JDBC Data Collection Configuration in 11 ms
18:54:44.528 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded Jmx Configuration in 6 ms
18:54:44.558 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded XML Data Collection Configuration in 18 ms
18:54:44.576 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded Microblog Configuration in 4 ms
18:54:44.727 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded poll-outages in 10 ms
18:54:44.760 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded poll-outages in 2 ms
18:54:44.832 [Main] INFO  org.opennms.netmgt.poller.monitors.JDBCMonitor - JDBCmonitor class loaded
18:54:44.834 [Main] INFO  org.opennms.netmgt.poller.monitors.JDBCMonitor - JDBCmonitor class loaded
18:54:44.834 [Main] INFO  org.opennms.netmgt.poller.monitors.JDBCMonitor - JDBCmonitor class loaded
18:54:45.139 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded Provisiond Configuration in 5 ms
18:54:45.148 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded Reportd Configuration in 6 ms
18:54:45.301 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded Config for SNMP Trap Northbounder in 16 ms
18:54:45.316 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded statistics daemon configuration in 11 ms
18:54:45.328 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded surveillance view configuration in 8 ms
18:54:45.367 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded Config for Syslog Northbounder in 8 ms
18:54:45.393 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded Telemetryd configuration in 23 ms
18:54:45.412 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded threshd-configuration in 15 ms
18:54:45.442 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded threshd-configuration in 2 ms
18:54:45.461 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded thresholds in 15 ms
18:54:45.474 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded thresholds in 3 ms
18:54:45.483 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded TL1d configuration in 5 ms
18:54:45.553 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded WMI peer configuration in 7 ms
18:54:45.554 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded WMI peer configuration in 1 ms
18:54:45.577 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded Vmware Cim Data Collection Configuration in 8 ms
18:54:45.592 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded Vmware Data Collection Configuration in 12 ms
18:54:45.599 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded Vmware Configuration in 3 ms
18:54:45.609 [Main] INFO  org.opennms.core.xml.AbstractJaxbConfigDao - Loaded WS-Man Configuration in 5 ms
Test is finished, found no problems.
Find and set Java environment for running OpenNMS in /opt/opennms/etc/java.conf.
id: cannot find name for user ID 1000140000
ERROR: you must run this script as opennms, not '1000140000'.
```

RangerRick:
I bet it's because `1000140000` doesn't have an associated username, so I fall back to `opennms`

Mind opening a new issue? I might be able to figure out a way around it.

Naicisum:
Here is one solution for openshift

In the source image, in my case `Dockerfile` do the following at some point:
``` dockerfile
RUN chmod g+w /etc/passwd* && \
    chmod g+w /etc/group*
```

In your `bootstrap` script or however you start the container, do the following:
``` bash
#!/bin/bash
umask 002

# Set username in passwd
if ! whoami &> /dev/null; then
    if [ -w /etc/passwd ]; then
        echo "${USER_NAME:-default}:x:$(id -u):0:${USER_NAME:-default} user:${HOME}:/sbin/nologin" >> /etc/passwd
    fi
fi

# Set groupname in group
if ! groups &> /dev/null; then
    if [ -w /etc/passwd ]; then
        echo "${USER_NAME:-default}:x:$(id -G | cut -d' ' -f2):" >> /etc/group
    fi
fi
```
You can replace `default` with whatever name you want in the container and it will not resolve correctly for most tools.

PR #3560, opened 2021-07-28
JDK11 Fixes

RangerRick:
This PR just fixes a couple of JDK11-related issues.

* exclude `xml-apis` from dependencies since it's included in the base JDK -- Eclipse would complain about classes being in the `jaxb` module as well as outside a module
* fix the "minimum JDK" detection in `runjava -s` now that JDK8 is disallowed
* build/run smoke tests under JDK11

There are still a few `pom.xml` files that target 1.8, it's not clear if there was a reason in @j-white's original branch to leave them that way or not. I left them alone for now.


cgorantla (approved):
LGTM if it passes smoke!

RangerRick:
It passed in `develop`, realized I needed to backport it to `release-28.x`.  I'll still wait though.  :)

PR #3566, opened 2021-08-04
NMS-13478: Add bulking to elastic flow repository

fooker:
The elastic flow repository now collects flow documents for batch
insertion across multiple logs. Each worker thread has its own
bulk storage.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13478

### DIFF
--- a/features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/ElasticFlowRepository.java
+++ b/features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/ElasticFlowRepository.java
@@ -141,6 +141,7 @@ public class ElasticFlowRepository implements FlowRepository {
 
     private boolean enableFlowForwarding = false;
 
+    private int bulkSize = 0;
     private int bulkRetryCount = 1;
 
     /**
@@ -154,6 +155,11 @@ public class ElasticFlowRepository implements FlowRepository {
      * This maps a node ID to a set if snmpInterface IDs.
      */
     private final Map<Direction, Cache<Integer, Set<Integer>>> markerCache = Maps.newEnumMap(Direction.class);
+
+    /**
+     * Collect flow documents ready for persistence.
+     */
+    private final Map<Thread, List<FlowDocument>> bulk = Maps.newHashMap();
 
     public ElasticFlowRepository(MetricRegistry metricRegistry, JestClient jestClient, IndexStrategy indexStrategy,
                                  DocumentEnricher documentEnricher,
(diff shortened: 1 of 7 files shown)


fooker:
Does this need some kind of timeout mechanisms to flush out pending bulk operations when input is lacking?

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/ElasticFlowRepository.java
```diff
@@ -155,6 +156,11 @@
      */
     private final Map<Direction, Cache<Integer, Set<Integer>>> markerCache = Maps.newEnumMap(Direction.class);
 
+    /**
+     * Collect flow documents ready for persistence.
+     */
+    private final Map<Thread, List<FlowDocument>> bulk = Maps.newHashMap();
```
j-white, line 162: Should this be a concurrent hash map? We could have many threads adding new keys at the same time for example.

j-white:
> Does this need some kind of timeout mechanisms to flush out pending bulk operations when input is lacking?

Yes :)

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/ElasticFlowRepository.java
christianpape: done.

cgorantla (commented):
Can we add a test case to make sure this works reliably

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/ElasticFlowRepository.java
```diff
@@ -210,6 +221,45 @@ public ElasticFlowRepository(MetricRegistry metricRegistry, JestClient jestClien
...
+        if (bulkFlushMs > 0) {
+            int delay = Math.max(1, bulkFlushMs / 2);
+            flushTimer = new java.util.Timer("ElasticFlowRepositoryFlush");
+            flushTimer.scheduleAtFixedRate(new TimerTask() {
+                @Override
+                public void run() {
+                    final long currentTimeMillis = System.currentTimeMillis();
+                    for(final Map.Entry<Thread, ElasticFlowRepository.FlowBulk> entry : flowBulks.entrySet()) {
+                        final ElasticFlowRepository.FlowBulk flowBulk = entry.getValue();
+                        if (flowBulk.lock.tryLock()) {
```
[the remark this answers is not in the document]
j-white: Shouldn't we hold the lock before calling `flowBulk.documents.size()`?

j-white:
Looks good. I'd consider setting the default values of `bulkSize` to 1000 and `bulkFlushMs` to 500 so that we use these new code paths by default now.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/ElasticFlowRepository.java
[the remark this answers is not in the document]
cgorantla: yeah, that may be needed. i was only thinking of `persistBulk`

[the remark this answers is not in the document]
christianpape: done.

christianpape:
> Can we add a test case to make sure this works reliably

Added some tests.

features/flows/itests/src/test/java/org/opennms/netmgt/flows/elastic/BulkingIT.java
```diff
@@ -0,0 +1,235 @@
...
+            final MockDocumentEnricherFactory mockDocumentEnricherFactory = new MockDocumentEnricherFactory();
+            final DocumentEnricher documentEnricher = mockDocumentEnricherFactory.getEnricher();
+            final FlowRepository flowRepository = createFlowRepository(jestClient, documentEnricher, 1000, 1000);
+
+            final long[] persists = new long[2];
+
+            // send full bulk in order to estimate last persist
+            flowRepository.persist(Lists.newArrayList(createMockedFlows(1000)), FlowDocumentTest.getMockFlowSource());
+
+            with().pollInterval(25, MILLISECONDS).await().atMost(10, SECONDS).until(() -> {
```
cgorantla, line 116: nitpick on this test :  25 milliseconds is too small to check if it already processed. First iteration will always succeed here. May be we can increase `bulkflushMs` to 5 secs and wait for at least 2 seconds with `pollDelay(2, SECONDS)`.

For next check, we can use `.pollDelay(5, SECONDS)` to create sufficient delay.

cgorantla (approved):

features/flows/itests/src/test/java/org/opennms/netmgt/flows/elastic/BulkingIT.java
christianpape: done.

PR #3567, opened 2021-08-04
NMS-13481: Reload interface-to-node cache non-blocking

fooker:
By re-building the cache without locking, the enrichment process can
continue to work while a re-build is inprocess.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13481


j-white (commented):
Looks good to me. `dataSourceSync` is called in `init`, so calls won't be made until the cache is synchronized, and then subsequent refreshes will happen in the background.

christianpape (approved):
LGTM.

RangerRick:
Should this go back to `release-28.x` or even `foundation-2021` if it applies?

RangerRick:
Green!  Ship it! 🚀

PR #3568, opened 2021-08-04
NMS-13462: Add in-memory impl for twin IPC

fooker:
The in-memory implementation servers as a mock impl.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13462


cgorantla:
Other than couple of minor tweaks, this looks good to me. 
@fooker  Do we want to merge this now so that we can use test framework to test all other brokers.

cgorantla (approved):
LGTM!

PR #3570, opened 2021-08-05
NMS-13460: Add gRPC implementation for Twin API

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13460


cgorantla:
Currently Twin gRPC inherits all the config from existing IPC Config except for port.
Considered moving to same server for existing IPC and Twin but deferred it to another issue as it involves changes in existing IPC as well, created https://issues.opennms.org/browse/NMS-13487.

core/ipc/twin/grpc/common/src/main/proto/twin-grpc.proto
```diff
@@ -0,0 +1,40 @@
...
+  string location = 3;
+  map<string, string> tracing_info = 4;
+}
+
+/*
+    Twin Response object sent by OpenNMS for RPC response as well as for Sink update.
+ */
+message TwinResponseProto {
+  string consumer_key = 1;
+  bytes twin_object = 2;
```
j-white, line 31: Based on this model it looks like we're sending the complete object over the wire every-time there is an update.

Do we plan on changing this API later when we account for incremental updates?

cgorantla: yes. Currently API and proto doesn't include version.
Once we include version, we can use same `twin_object` for complete/partial objects.

j-white: 👍 thanks

RangerRick (approved):
Looks reasonable to me, nothing jumps out at me.  Have to resolve the conflicts, obviously. :D

PR #3580, opened 2021-08-10
NMS-13485: Resolve hostname from Syslog Messages by doing a lookup on specific minion

cgorantla:
Syslog messages that are received at Minion but parsed on OpenNMS may not resolve hostname (from SyslogMessage) properly.
By issuing a lookup request on that specific minion, we hope to resolve hostname and associate with corresponding node.

This may have some effect on syslog parsing performance. 
May be improved in future by moving Syslog parsing to Minion.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13485


cgorantla:
For improving performance, may be we can create generic DNS cache for all existing nodes whenever they are added. 
This way, we can depend on this cache for resolving hostname instead of resolving for every message.

features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/ConvertToEvent.java
```diff
@@ -148,7 +154,20 @@ public static final EventBuilder toEventBuilder(SyslogMessage message, String sy
...
-        final InetAddress hostAddress = message.getHostAddress();
+        InetAddress hostAddress = message.getHostAddress();
+        String hostIpAddress = null;
+        if (hostAddress == null &&
+                message.getHostName() != null &&
+                !LocationUtils.isDefaultLocationName(location) &&
+                locationAwareDnsLookupClient != null) {
+            CompletableFuture<String> future = locationAwareDnsLookupClient.lookup(message.getHostName(), location, systemId);
+            try {
+                hostIpAddress = future.get();
```
j-white, line 165: This is too big of a performance hit for this code as it stands.

The requirement for this has typically been that the host given in the syslog message must match the node label for them to be associated.

Naicisum: Use the same cache that is used for flow reverse lookups. That code can handle this with ease and is very performant

cgorantla: Added localCache for syslog.

cgorantla: > Use the same cache that is used for flow reverse lookups. That code can handle this with ease and is very performant

That's not designed for querying across locations at least for now.

cgorantla:
> For improving performance, may be we can create generic DNS cache for all existing nodes whenever they are added.
> This way, we can depend on this cache for resolving hostname instead of resolving for every message.

Added dns cache locally for syslog.

features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/ConvertToEvent.java
```diff
@@ -230,6 +245,39 @@ public static final EventBuilder toEventBuilder(SyslogMessage message, String sy
...
+        if (LocationUtils.isDefaultLocationName(location)) {
+            return message.getHostAddress();
+        }
+        String hostName = message.getHostName();
+        if (Strings.isNullOrEmpty(hostName) ||
+                locationAwareDnsLookupClient == null || dnsCache == null) {
+            return null;
+        }
+        // Try to find the element in the cache first.
+        String hostIpAddress = dnsCache.getIfPresent(new HostNameWithLocationKey(hostName, location));
```
RangerRick, line 261: Minor nit, you end up doing `new HostNameWithLocationKey(hostName, location)` twice if it's not in the cache.

RangerRick (approved):

agalue:
As this has been merged for 28.x, what about Meridian users?

cgorantla:
> As this has been merged for 28.x, what about Meridian users?

@agalue Which meridian version you would like this targeted to ?

agalue:
I'd say any supported Meridian version that works with Minions, so 2019 and newer.

PR #3583, opened 2021-08-13
Minion doc tweak for scv.jce location

mmahacek:
Small tweak to clarify the location of the scv.jce file in the docs.  Brought up due to a customer question to verify the file is in the application ETC folder, not the root system ETC folder.

### External References

* JIRA (Issue Tracker): none


Bonrob2 (approved):
LGTM!

indigo423:
You PR invalidates: https://github.com/OpenNMS/opennms/blob/d6cf8d30397a104b430fe1f4266a20706d3667f1/docs/modules/deployment/pages/minion/install.adoc#configure-connectivity-to-the-core-instance

indigo423 (changes requested):
Your PR invalidates this here:

Screenshot 2021-08-16 at 15 18 58

We would need to introduce MINION_HOME or SENTINEL_HOME globally, somewhere.

mmahacek:
> We would need to introduce MINION_HOME or SENTINEL_HOME globally, somewhere.

Fair point.  I don't think we define $OPENNMS_HOME in the docs, and that is used in many different locations.

mmahacek:
@indigo423 What do you think about adding a note like this to each of the install pages for the various components?
image

indigo423:
> @indigo423 What do you think about adding a note like this to each of the install pages for the various components?
> image

Yes, totally agree. I personally don't have a strong opinion on which variant. We can a) point the user to the directory and use a relative path or b) introduce for each component a `*_HOME` variable.

If we pick the path for the `*_HOME` variable it would be nice we really set them on installation so the copy & paste stuff works regardless of Debian/Ubuntu or CentOS/RHEL.

mmahacek:
> We already do this for core:

That only appears on a Cent install for me.  I don't see that file existing on a Ubuntu install I have.

indigo423:
> > We already do this for core:
> 
> That only appears on a Cent install for me. I don't see that file existing on a Ubuntu install I have.

ok sweet, so my suggestion would be then we use `OPENNMS_HOME`, `SENTINEL_HOME` and `MINION_HOME` and open an additional issue to set these variables during RPM/DEB install correctly. In the meantime we can add a NOTE on top of each component and say for example SENTINE_HOME is `/opt/sentinel` on CentOS/RHEL and `/usr/share/sentinel` on Debian/Ubuntu. As soon we have the patch with the variables merged we can remove the note entirely and rely on the variable. What do you think?

mmahacek:
I think it would be helpful to have the variables set during install.  IMO, the reference to the path could permanently live on the install page for each product, but there wouldn't need to be a reference to the path on component/feature pages.

docs/modules/deployment/pages/minion/message-broker/activemq.adoc
```diff
@@ -31,7 +31,7 @@ opennms:scv-set opennms.broker my-minion-user my-minion-password<2>
 <1> Set the credentials for the REST endpoint created in your {page-component-title} Core instance.
 <2> Set the credentials for the ActiveMQ message broker.
 
-The credentials are encrypted on disk in `etc/scv.jce`.
+The credentials are encrypted on disk in `$MINION_HOME/etc/scv.jce`.
```
indigo423, line 34:
```suggestion
The credentials are encrypted on disk in `${MINION_HOME}/etc/scv.jce`.
```

docs/modules/deployment/pages/minion/message-broker/kafka.adoc
```diff
@@ -55,7 +55,7 @@ opennms:scv-set opennms.http my-minion-user my-minion-password<1>
 ----
 <1> Set the credentials for the REST endpoint created in your {page-component-title} Core instance.
 
-The credentials are encrypted on disk in `etc/scv.jce`.
+The credentials are encrypted on disk in `$MINION_HOME/etc/scv.jce`.
```
indigo423, line 58:
```suggestion
The credentials are encrypted on disk in `${MINION_HOME}/etc/scv.jce`.
```

docs/modules/deployment/pages/sentinel/message-broker/activemq.adoc
```diff
@@ -44,7 +44,7 @@ opennms:scv-set opennms.broker my-sentinel-user my-sentinel-password<2>
 <1> Set the credentials for the REST endpoint created in your {page-component-title} Core instance
 <2> Set the credentials for the ActiveMQ message broker
 
-NOTE: The credentials are encrypted on disk in `etc/scv.jce`.
+NOTE: The credentials are encrypted on disk in `$SENTINEL_HOME/etc/scv.jce`.
```
indigo423, line 47:
```suggestion
NOTE: The credentials are encrypted on disk in `${SENTINEL_HOME}/etc/scv.jce`.
```

docs/modules/deployment/pages/sentinel/message-broker/kafka.adoc
```diff
@@ -66,7 +66,7 @@ opennms:scv-set opennms.http my-sentinel-user my-sentinel-password<1>
 ----
 <1> Set the credentials for the REST endpoint created in your {page-component-title} Core instance
 
-NOTE: The credentials are encrypted on disk in `etc/scv.jce`.
+NOTE: The credentials are encrypted on disk in `$SENTINEL_HOME/etc/scv.jce`.
```
indigo423, line 69:
```suggestion
NOTE: The credentials are encrypted on disk in `${SENTINEL_HOME}/etc/scv.jce`.
```

indigo423 (commented):
We should be defensive and should use `${}`, especially with `_` and string concatenation.

PR #3588, opened 2021-08-18
NMS-13511: Use Karaf shell commands to secure Minion Karaf shell access

indigo423:
### All Contributors

* [ ] Have you read and followed our [Contribution Guidelines](https://github.com/OpenNMS/opennms/blob/develop/CONTRIBUTING.md)?
* [ ] Have you [made an issue in the OpenNMS issue tracker](https://issues.opennms.org)?<br>If so, you should:
* [ ] Have you made a comment in that issue which points back to this PR?
* [ ] Have you updated the JIRA link at the bottom of this comment to link to your issue?

Use Karaf shell commands instead of editing files in the file system manually.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13511


mmahacek (approved):
These steps are a lot simpler and should be easier to maintain.  My only comment is that since the RHEL/Ubuntu options point to the same document, is there any reason to combine them to prevent possible confusion at the steps being the same? The end result only having a "non-docker" (exact wording TBD) and Docker options in the Securing Karaf section.

indigo423:
> These steps are a lot simpler and should be easier to maintain. My only comment is that since the RHEL/Ubuntu options point to the same document, is there any reason to combine them to prevent possible confusion at the steps being the same? The end result only having a "non-docker" (exact wording TBD) and Docker options in the Securing Karaf section.

Yeah, thanks for asking :) The Meridian docs exclude the Ubuntu/Debian. But I agree. Maybe we can do it like Grafana and just name it "Linux" and "Docker".

Screenshot 2021-08-19 at 09 40 57

indigo423:
@mmahacek Here the change I've made remove the duplicate entry for CentOS/RHEL and Ubuntu/Debian:

Screenshot 2021-08-23 at 11 49 44

agalue (approved):
LGTM

PR #3589, opened 2021-08-18
PE-26: make sure CM-Phase1 build docker image

freddy61025:
https://issues.opennms.org/browse/PE-26
make a CM-Phase1 container ready for system integration


RangerRick (approved):
yeah, I think this should be enough to get it to publish (horizon) to docker hub

PR #3590, opened 2021-08-19
NMS-13219: Now, time spans can only be added once

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-13219


RangerRick (approved):
LGTM

PR #3592, opened 2021-08-19
NMS-13479: Improve flow classification

swachter:
Issue: https://issues.opennms.org/browse/NMS-13479

This PR replaces the existing flow classification engine with a new implementation. The existing flow classification engine considers ports only when selecting rules. This results in a highly inefficient classification in case that a rule set mainly relies addresses.

The new classification engine uses a decision tree. Experiments with customer rule sets showed a speedup of 30x for a rule set that mainly used addresses and 4x for a rule set that considers ports and addresses.

Another benefit of the new implementation is that it is thread-safe even in case of concurrent reloads.


indigo423:
Just as an idea, cause we are refactoring the way our classification rule engine internally works, I would suggest putting this into H29 -> develop instead into a patch release for H28.

swachter:
When I started the work it was meant to go into 28.0.3. Maybe the extend of the change justifies to target it to something else. OTOH the observable behavior is not changed by this PR.

RangerRick:
Still, with such a large change, it's probably a good idea to bake it in 29 first, since we don't know if there are any unintended side-effects before exposing the code to more people.

RangerRick (approved):
Nothing major jumps out at me code-wise, once we know which branch to target, seems good to go.

PR #3579, opened 2021-08-10
NMS-13422: Add full trapoid for snmp trap events

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13422

PR #3612, opened 2021-08-31
NMS-13422:  Add full trapoid for snmp trap events

cgorantla:
This backports #3579 and updates the fix.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13422


cgorantla:
Since this is already reviewed and smoke-tests are passing, merging.

Verified locally on `foundation-2019`

PR #3613, opened 2021-08-31
NMS-13479: Reload classification decision tree asynchronously in DefaultClassificationService

swachter:
Issue: https://issues.opennms.org/browse/NMS-13479

While editing classification rules, reloads of the classification engine are triggered repeatedly. A reload may take a couple of seconds depending on rule set sizes. In order to avoid sluggish behavior, reloading is done asynchronously. Only when the classification engine is used calls are blocked until any ongoing reloads did finish.

These changes fix the tests in `ClassificationRulePageIT`.

### DIFF
--- a/features/flows/classification/engine/impl/src/main/java/org/opennms/netmgt/flows/classification/internal/DefaultClassificationEngine.java
+++ b/features/flows/classification/engine/impl/src/main/java/org/opennms/netmgt/flows/classification/internal/DefaultClassificationEngine.java
@@ -110,9 +110,9 @@ public class DefaultClassificationEngine implements ClassificationEngine {
                     .append("minComp  : " + tree.info.minComp).append('\n')
                     .append("maxComp  : " + tree.info.maxComp).append('\n')
                     .append("avgComp  : " + (double) tree.info.sumComp / tree.info.leaves).append('\n')
-                    .append("minRuleSetSize :" + tree.info.minRuleSetSize).append('\n')
-                    .append("maxRuleSetSize :" + tree.info.maxRuleSetSize).append('\n')
-                    .append("avgRuleSetSize :" + (double) tree.info.sumRuleSetSize / tree.info.leaves).append('\n');
+                    .append("minLeafSize : " + tree.info.minLeafSize).append('\n')
+                    .append("maxLeafSize : " + tree.info.maxLeafSize).append('\n')
+                    .append("avgLeafSize : " + (double) tree.info.sumLeafSize / tree.info.leaves).append('\n');
             LOG.info(sb.toString());
         }
 
--- a/features/flows/classification/engine/impl/src/main/java/org/opennms/netmgt/flows/classification/internal/DefaultClassificationService.java
+++ b/features/flows/classification/engine/impl/src/main/java/org/opennms/netmgt/flows/classification/internal/DefaultClassificationService.java
@@ -32,10 +32,6 @@ import java.io.InputStream;
 import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Objects;
-import java.util.concurrent.ArrayBlockingQueue;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.ThreadPoolExecutor;
-import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 
 import org.opennms.core.criteria.Criteria;
@@ -59,12 +55,8 @@ import org.opennms.netmgt.flows.classification.persistence.api.ClassificationGro
 import org.opennms.netmgt.flows.classification.persistence.api.ClassificationRuleDao;
 import org.opennms.netmgt.flows.classification.persistence.api.Group;
 import org.opennms.netmgt.flows.classification.persistence.api.Rule;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 public class DefaultClassificationService implements ClassificationService {
-
-    private static final Logger LOG = LoggerFactory.getLogger(DefaultClassificationService.class);
 
     private final ClassificationRuleDao classificationRuleDao;
 
@@ -87,7 +79,7 @@ public class DefaultClassificationService implements ClassificationService {
                                         SessionUtils sessionUtils) {
         this.classificationRuleDao = Objects.requireNonNull(classificationRuleDao);
         this.classificationGroupDao = Objects.requireNonNull(classificationGroupDao);
-        this.classificationEngine = new AsyncReloadingClassificationEngine(Objects.requireNonNull(classificationEngine));
+        this.classificationEngine = Objects.requireNonNull(classificationEngine);
         this.sessionUtils = Objects.requireNonNull(sessionUtils);
         this.ruleValidator = new RuleValidator(filterService);
         this.groupValidator = new GroupValidator(classificationRuleDao);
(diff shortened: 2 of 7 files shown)


smoke-test/src/test/java/org/opennms/smoketest/ClassificationRulePageIT.java
```diff
@@ -477,10 +480,20 @@ public String classify(ClassificationRequestDTO classificationRequest) {
...
-            sleep(DEFAULT_WAIT_TIME);
 
-            // Fiddle result out of UI
-            return execute(() -> findElementById("classification-response")).getText();
+            String textAfterClassification = "";
+
+            // wait until the classification response value change to a non-empty string that is different from the value
+            // before classification was started
+            for (int i = 0; i < 30; i++) {
+                sleep(DEFAULT_WAIT_TIME);
```
cgorantla, line 489: You can use  `await().atMost()` pattern here instead of sleep. https://github.com/awaitility/awaitility

features/flows/classification/engine/impl/src/main/java/org/opennms/netmgt/flows/classification/internal/DefaultClassificationService.java
```diff
@@ -358,4 +369,124 @@ private void checkForDuplicateName (Group group) {
...
+                }
+            }
+        }
+
+        private void doReload() {
+            // this method must not modify the state because it not synchronized
+            try {
+                LOG.debug("reload classification engine");
+                delegate.reload();
+                if (reloadSucceeded()) {
```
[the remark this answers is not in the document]
fooker: Yeah, this is a little bit confusing. Second interpretation could be that these methods are event handlers or state change handlers. Then they should be called `onReloadSucceeded()`, etc.

features/flows/classification/engine/impl/src/main/java/org/opennms/netmgt/flows/classification/internal/DefaultClassificationEngine.java
```diff
@@ -110,9 +110,9 @@ public void reload() {
                     .append("minComp  : " + tree.info.minComp).append('\n')
                     .append("maxComp  : " + tree.info.maxComp).append('\n')
                     .append("avgComp  : " + (double) tree.info.sumComp / tree.info.leaves).append('\n')
-                    .append("minRuleSetSize :" + tree.info.minRuleSetSize).append('\n')
-                    .append("maxRuleSetSize :" + tree.info.maxRuleSetSize).append('\n')
-                    .append("avgRuleSetSize :" + (double) tree.info.sumRuleSetSize / tree.info.leaves).append('\n');
+                    .append("minLeafSize : " + tree.info.minLeafSize).append('\n')
```
fooker, line 113: Unnecessary refactoring. Please extract in single commit next time (read about `git add -p -i` to split these changes in multiple commits).

features/flows/classification/engine/impl/src/main/java/org/opennms/netmgt/flows/classification/internal/DefaultClassificationService.java
```diff
@@ -79,11 +87,14 @@ public DefaultClassificationService(ClassificationRuleDao classificationRuleDao,
                                         SessionUtils sessionUtils) {
         this.classificationRuleDao = Objects.requireNonNull(classificationRuleDao);
         this.classificationGroupDao = Objects.requireNonNull(classificationGroupDao);
-        this.classificationEngine = Objects.requireNonNull(classificationEngine);
+        this.classificationEngine = new AsyncReloadingClassificationEngine(Objects.requireNonNull(classificationEngine));
```
fooker, line 90: I think the wrapping should be done in the application context wiring.

features/flows/classification/engine/impl/src/main/java/org/opennms/netmgt/flows/classification/internal/DefaultClassificationService.java
```diff
@@ -79,11 +87,14 @@ public DefaultClassificationService(ClassificationRuleDao classificationRuleDao,
...
         this.sessionUtils = Objects.requireNonNull(sessionUtils);
         this.ruleValidator = new RuleValidator(filterService);
         this.groupValidator = new GroupValidator(classificationRuleDao);
         this.csvService = new CsvServiceImpl(ruleValidator);
+        // trigger reload
+        // -> blocks classification requests until classification engine is ready
+        this.classificationEngine.reload();
```
fooker, line 97: Initial reload should be done by `AsyncReloadingClassificationEngine` itself. This would make the wrapper more encapsulated.

features/flows/classification/engine/impl/src/main/java/org/opennms/netmgt/flows/classification/internal/DefaultClassificationService.java
```diff
@@ -322,7 +331,13 @@ public void validateRule(final Rule validateMe) {
         return sessionUtils.withTransaction(supplier);
     }
 
-    private void updateRulePositionsAndReloadEngine(final List<Rule> rules) {
+    private <T> T runInTransactionAndThenReload(Supplier<T> supplier) {
```
fooker, line 334: While reading these changes, my initial thought was that `runInTransactionAndThenReload` will be extended by the async wrapper. This is clearly not the case. Therefore the whole renaming is purely cosmetic. Please avoid this kind of clutter.

features/flows/classification/engine/impl/src/main/java/org/opennms/netmgt/flows/classification/internal/decision/Tree.java
```diff
@@ -452,11 +454,11 @@ public String toString() {
         }
 
         public final static class WithClassifiers extends Leaf {
-            public final List<Classifier> classifiers;
+            public final Classifiers classifiers;
 
             public WithClassifiers(List<Classifier> classifiers) {
                 super(Info.forLeaf(classifiers.size()));
-                this.classifiers = classifiers;
+                this.classifiers = new Classifiers(true, classifiers);
```
fooker, line 461: What does this do? Why is this change required?

fooker (changes requested):
Please make sure that the commits message headers are correct before merging this (quash merging via the github UI and renaming there should be fine).

features/flows/classification/engine/impl/src/main/java/org/opennms/netmgt/flows/classification/internal/DefaultClassificationService.java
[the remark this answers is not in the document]
swachter: renamed to `onReloadSucceded` and `onReloadFailed`

features/flows/classification/engine/impl/src/main/java/org/opennms/netmgt/flows/classification/internal/decision/Tree.java
swachter: The change reduces allocations during classification.

features/flows/classification/engine/impl/src/main/java/org/opennms/netmgt/flows/classification/internal/DefaultClassificationService.java
```diff
@@ -358,4 +369,124 @@ private void checkForDuplicateName (Group group) {
...
+                    LOG.debug("another classification engine reload is required");
+                    doReload();
+                } else {
+                    LOG.debug("classification engine reloaded");
+                }
+            }
+        }
+
+        private synchronized boolean reloadSucceeded() {
+            if (state == State.NEED_ANOTHER_RELOAD) {
```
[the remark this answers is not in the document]
swachter: The delegate has no state of its own. The delegate is always ready to do classifications. When `reloadSucceded` is called no other reload is in progress. (Ignoring possible background reload triggers; but these have no impact because reloads are idempotent.)

features/flows/classification/engine/impl/src/main/java/org/opennms/netmgt/flows/classification/internal/DefaultClassificationService.java
swachter: I think it is a matter of taste. The `AsyncReloadingClassificationEngine` was only introduced to support consistent and non-sluggish front-end behavior. It allows to repeatedly trigger reloads but blocks on classification if a reload is on-going.

The other classification engine instances that are configured in the application context (`DefaultClassificationEngine` and `TimingClassificationEngine`) are of general use. In particular, these two classification are non-blocking even in case of on-going reloads.

The `AsyncReloadingClassificationEngine` is only used in the `DefaultClassificationService` that underlies the `ClassificationRestService`.

features/flows/classification/engine/impl/src/main/java/org/opennms/netmgt/flows/classification/internal/DefaultClassificationService.java
swachter: done

features/flows/classification/engine/impl/src/main/java/org/opennms/netmgt/flows/classification/internal/DefaultClassificationService.java
swachter: I disagree with your judgement. `runInTransactionAndThenReload` combines two steps that are executed at several places. The "reload" was indicated in method names earlier. Therefore I added it as a suffix to the new method.

features/flows/classification/engine/impl/src/main/java/org/opennms/netmgt/flows/classification/internal/DefaultClassificationService.java
fooker: It's currently only used that way. Please do note remove the flexibility we head before.

features/flows/classification/engine/impl/src/main/java/org/opennms/netmgt/flows/classification/internal/DefaultClassificationService.java
fooker: But why have you moved the reload from these methods in the first place? Just to get rid of the double-reload in `updateGroup`?

swachter: Yes, reload takes several seconds. Classification would be blocked 2 times as long.

cgorantla (approved):
LGTM!

features/flows/classification/engine/impl/src/main/java/org/opennms/netmgt/flows/classification/internal/DefaultClassificationService.java
swachter: AsyncReloadingClassification is now wired in app context.

fooker (approved):
Please make sure the commits are squashed and named properly.

PR #3633, opened 2021-09-09
NMS-13552: Added JVM option

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-13552

### DIFF
--- a/container/shared/src/main/resources/etc/container.init
+++ b/container/shared/src/main/resources/etc/container.init
@@ -67,7 +67,7 @@ EXTRA_JAVA_OPTS="$JAVA_OPTS $EXTRA_JAVA_OPTS"
 unset JAVA_OPTS
 
 # always dump heap if we OOM
-EXTRA_JAVA_OPTS="$EXTRA_JAVA_OPTS -XX:+HeapDumpOnOutOfMemoryError"
+EXTRA_JAVA_OPTS="$EXTRA_JAVA_OPTS -XX:+HeapDumpOnOutOfMemoryError -XX:-OmitStackTraceInFastThrow"
 
 # export any default configurable variables from sysconf
 export RUNAS JAVA_MIN_MEM JAVA_MAX_MEM MAX_FD JAVA_HOME JAVA JAVA_DEBUG_OPTS JAVA_DEBUG_PORT EXTRA_JAVA_OPTS CLASSPATH KARAF_DEBUG LD_LIBRARY_PATH
--- a/features/telemetry/listeners/pom.xml
+++ b/features/telemetry/listeners/pom.xml
@@ -37,6 +37,9 @@
       <artifactId>org.opennms.features.telemetry.common</artifactId>
       <version>${project.version}</version>
     </dependency>
-
+    <dependency>
+      <groupId>com.swrve</groupId>
+      <artifactId>rate-limited-logger</artifactId>
+    </dependency>
   </dependencies>
 </project>
--- a/features/telemetry/listeners/src/main/java/org/opennms/netmgt/telemetry/listeners/UdpListener.java
+++ b/features/telemetry/listeners/src/main/java/org/opennms/netmgt/telemetry/listeners/UdpListener.java
@@ -28,20 +28,10 @@
 
 package org.opennms.netmgt.telemetry.listeners;
 
-import java.net.InetSocketAddress;
-import java.util.List;
-import java.util.Objects;
-
-import org.opennms.netmgt.telemetry.api.receiver.Listener;
-import org.opennms.netmgt.telemetry.api.receiver.Parser;
-import org.opennms.netmgt.telemetry.listeners.utils.BufferUtils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.MetricRegistry;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
-
+import com.swrve.ratelimitedlogger.RateLimitedLog;
 import io.netty.bootstrap.Bootstrap;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelHandlerContext;
@@ -57,9 +47,24 @@ import io.netty.channel.socket.DatagramPacket;
 import io.netty.channel.socket.nio.NioDatagramChannel;
 import io.netty.util.ReferenceCountUtil;
 import io.netty.util.internal.SocketUtils;
+import org.opennms.netmgt.telemetry.api.receiver.Listener;
+import org.opennms.netmgt.telemetry.api.receiver.Parser;
+import org.opennms.netmgt.telemetry.listeners.utils.BufferUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.net.InetSocketAddress;
+import java.time.Duration;
+import java.util.List;
+import java.util.Objects;
 
 public class UdpListener implements Listener {
     private static final Logger LOG = LoggerFactory.getLogger(UdpListener.class);
+
+    public static final RateLimitedLog RATE_LIMITED_LOG = RateLimitedLog
+            .withRateLimit(LOG)
+            .maxRate(5).every(Duration.ofSeconds(30))
+            .build();
 
     private final String name;
     private final List<UdpParser> parsers;
@@ -191,7 +196,7 @@ public class UdpListener implements Listener {
                 @Override
                 public void exceptionCaught(final ChannelHandlerContext ctx, final Throwable cause) throws Exception {
                     LOG.warn("Invalid packet: {}", cause.getMessage());
-                    LOG.debug("", cause);
+                    RATE_LIMITED_LOG.debug("", cause);
                 }
             });
         }


cgorantla:
We may also want to use [RateLimitedLogger](https://github.com/OpenNMS/opennms/blob/071d95719eb91e5dc6bc3c085ec35a1f35196013/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java#L89) to limit the rate at which same logs are repeated.

RangerRick:
Also, is this useful as a general debugging thing? If so, should it go back to `foundation-2021` or even older?

RangerRick (changes requested):
This should probably at least go back to foundation-2021 since it's about debugging production Minion issues.

christianpape:
> We may also want to use [RateLimitedLogger](https://github.com/OpenNMS/opennms/blob/071d95719eb91e5dc6bc3c085ec35a1f35196013/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java#L89) to limit the rate at which same logs are repeated.

This optimization mechanism only applies to the following type of exceptions:
* NullPointerException
* ArithmeticException
* ArrayIndexOutOfBoundsException
* ArrayStoreException
* ClassCastException

In my opinion are these exceptions I want to know of.

cgorantla:
> > We may also want to use [RateLimitedLogger](https://github.com/OpenNMS/opennms/blob/071d95719eb91e5dc6bc3c085ec35a1f35196013/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java#L89) to limit the rate at which same logs are repeated.
> 
> This optimization mechanism only applies to the following type of exceptions:
> 
> * NullPointerException
> * ArithmeticException
> * ArrayIndexOutOfBoundsException
> * ArrayStoreException
> * ClassCastException
> 
> In my opinion are these exceptions I want to know of.

yes, keep the optimization but use RATE_LIMITED_LOGGER where we are logging exceptions that are repetitive. 
It will still log the exception but it will emit 1 such log instead of 100.

christianpape:
> > > We may also want to use [RateLimitedLogger](https://github.com/OpenNMS/opennms/blob/071d95719eb91e5dc6bc3c085ec35a1f35196013/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java#L89) to limit the rate at which same logs are repeated.
> > 
> > 
> > This optimization mechanism only applies to the following type of exceptions:
> > 
> > * NullPointerException
> > * ArithmeticException
> > * ArrayIndexOutOfBoundsException
> > * ArrayStoreException
> > * ClassCastException
> > 
> > In my opinion are these exceptions I want to know of.
> 
> yes, keep the optimization but use RATE_LIMITED_LOGGER where we are logging exceptions that are repetitive.
> It will still log the exception but it will emit 1 such log instead of 100.



> > > We may also want to use [RateLimitedLogger](https://github.com/OpenNMS/opennms/blob/071d95719eb91e5dc6bc3c085ec35a1f35196013/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java#L89) to limit the rate at which same logs are repeated.
> > 
> > 
> > This optimization mechanism only applies to the following type of exceptions:
> > 
> > * NullPointerException
> > * ArithmeticException
> > * ArrayIndexOutOfBoundsException
> > * ArrayStoreException
> > * ClassCastException
> > 
> > In my opinion are these exceptions I want to know of.
> 
> yes, keep the optimization but use RATE_LIMITED_LOGGER where we are logging exceptions that are repetitive.
> It will still log the exception but it will emit 1 such log instead of 100.

Since these are runtime exceptions and can appear everywhere in the code, we would then need to rate-limit each and every logging statement where an exception is logged. Before we do this, we could just close this PR and keep the behaviour as it is.

Another way is not to add this option by default and to just add documentation and write troubleshooting hints about this behaviour. So, an user is able to set this option if this hits us again.

In my opinion the exceptions this optimization handles are not domain-specific or something like that. These are nothing like timeouts, connection erros, parsing error or something like this, because for these we have out own exceptions implementations in place. These exceptions like NullPointer, ArrayIndexOutOfBounds, ClassCast, ArtihmeticException indicate erroneous code with logical errors. 

Furthermore, this optimization mechanism is in place since Java v1.5. It never hit us before. So I really don't care whether we include the option or not. I thought it was a good idea, but changing the logging code for everything that runs in the Minion doesn't make sense to me.

cgorantla:
> > > > We may also want to use [RateLimitedLogger](https://github.com/OpenNMS/opennms/blob/071d95719eb91e5dc6bc3c085ec35a1f35196013/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java#L89) to limit the rate at which same logs are repeated.
> > > 
> > > 
> > > This optimization mechanism only applies to the following type of exceptions:
> > > 
> > > * NullPointerException
> > > * ArithmeticException
> > > * ArrayIndexOutOfBoundsException
> > > * ArrayStoreException
> > > * ClassCastException
> > > 
> > > In my opinion are these exceptions I want to know of.
> > 
> > 
> > yes, keep the optimization but use RATE_LIMITED_LOGGER where we are logging exceptions that are repetitive.
> > It will still log the exception but it will emit 1 such log instead of 100.
> 
> > > > We may also want to use [RateLimitedLogger](https://github.com/OpenNMS/opennms/blob/071d95719eb91e5dc6bc3c085ec35a1f35196013/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java#L89) to limit the rate at which same logs are repeated.
> > > 
> > > 
> > > This optimization mechanism only applies to the following type of exceptions:
> > > 
> > > * NullPointerException
> > > * ArithmeticException
> > > * ArrayIndexOutOfBoundsException
> > > * ArrayStoreException
> > > * ClassCastException
> > > 
> > > In my opinion are these exceptions I want to know of.
> > 
> > 
> > yes, keep the optimization but use RATE_LIMITED_LOGGER where we are logging exceptions that are repetitive.
> > It will still log the exception but it will emit 1 such log instead of 100.
> 
> Since these are runtime exceptions and can appear everywhere in the code, we would then need to rate-limit each and every logging statement where an exception is logged. Before we do this, we could just close this PR and keep the behaviour as it is.
> 
> Another way is not to add this option by default and to just add documentation and write troubleshooting hints about this behaviour. So, an user is able to set this option if this hits us again.
> 
> In my opinion the exceptions this optimization handles are not domain-specific or something like that. These are nothing like timeouts, connection erros, parsing error or something like this, because for these we have out own exceptions implementations in place. These exceptions like NullPointer, ArrayIndexOutOfBounds, ClassCast, ArtihmeticException indicate erroneous code with logical errors.
> 
> Furthermore, this optimization mechanism is in place since Java v1.5. It never hit us before. So I really don't care whether we include the option or not. I thought it was a good idea, but changing the logging code for everything that runs in the Minion doesn't make sense to me.

I'm definitely not suggesting to change logging everywhere. Just for the case where this issue originated from. 
May be [here](https://github.com/OpenNMS/opennms/blob/develop/features/telemetry/listeners/src/main/java/org/opennms/netmgt/telemetry/listeners/UdpListener.java#L204)

christianpape:
> > > > > We may also want to use [RateLimitedLogger](https://github.com/OpenNMS/opennms/blob/071d95719eb91e5dc6bc3c085ec35a1f35196013/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java#L89) to limit the rate at which same logs are repeated.
> > > > 
> > > > 
> > > > This optimization mechanism only applies to the following type of exceptions:
> > > > 
> > > > * NullPointerException
> > > > * ArithmeticException
> > > > * ArrayIndexOutOfBoundsException
> > > > * ArrayStoreException
> > > > * ClassCastException
> > > > 
> > > > In my opinion are these exceptions I want to know of.
> > > 
> > > 
> > > yes, keep the optimization but use RATE_LIMITED_LOGGER where we are logging exceptions that are repetitive.
> > > It will still log the exception but it will emit 1 such log instead of 100.
> > 
> > 
> > > > > We may also want to use [RateLimitedLogger](https://github.com/OpenNMS/opennms/blob/071d95719eb91e5dc6bc3c085ec35a1f35196013/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/OpennmsKafkaProducer.java#L89) to limit the rate at which same logs are repeated.
> > > > 
> > > > 
> > > > This optimization mechanism only applies to the following type of exceptions:
> > > > 
> > > > * NullPointerException
> > > > * ArithmeticException
> > > > * ArrayIndexOutOfBoundsException
> > > > * ArrayStoreException
> > > > * ClassCastException
> > > > 
> > > > In my opinion are these exceptions I want to know of.
> > > 
> > > 
> > > yes, keep the optimization but use RATE_LIMITED_LOGGER where we are logging exceptions that are repetitive.
> > > It will still log the exception but it will emit 1 such log instead of 100.
> > 
> > 
> > Since these are runtime exceptions and can appear everywhere in the code, we would then need to rate-limit each and every logging statement where an exception is logged. Before we do this, we could just close this PR and keep the behaviour as it is.
> > Another way is not to add this option by default and to just add documentation and write troubleshooting hints about this behaviour. So, an user is able to set this option if this hits us again.
> > In my opinion the exceptions this optimization handles are not domain-specific or something like that. These are nothing like timeouts, connection erros, parsing error or something like this, because for these we have out own exceptions implementations in place. These exceptions like NullPointer, ArrayIndexOutOfBounds, ClassCast, ArtihmeticException indicate erroneous code with logical errors.
> > Furthermore, this optimization mechanism is in place since Java v1.5. It never hit us before. So I really don't care whether we include the option or not. I thought it was a good idea, but changing the logging code for everything that runs in the Minion doesn't make sense to me.
> 
> I'm definitely not suggesting to change logging everywhere. Just for the case where this issue originated from.
> May be [here](https://github.com/OpenNMS/opennms/blob/develop/features/telemetry/listeners/src/main/java/org/opennms/netmgt/telemetry/listeners/UdpListener.java#L204)

Ah ok. Got it. Sorry, I misunderstood you.

cgorantla (approved):
LGTM!

PR #3638, opened 2021-09-11
PE-27: Support upgrade for provisiond service (XML config)

patrick-schweizer:
This PR adds the ability to the CM System to import existing xml configurations through liquibase.

* JIRA (Issue Tracker): https://issues.opennms.org/browse/PE-27


features/config/upgrade/src/main/java/org/opennms/config/upgrade/UpgradeConfigService.java
```diff
@@ -45,16 +46,22 @@
...
+                                final boolean skipConfigUpgrades) {
         this.cm = cm;
         this.dataSource = dataSource;
+        this.skipConfigUpgrades = skipConfigUpgrades;
     }
 
     @Override
     public void afterPropertiesSet() throws Exception {
-        new LiquibaseUpgrader(cm).runChangelog("changelog-cm/changelog-cm.xml", dataSource.getConnection());
+        if (!skipConfigUpgrades) { // TODO: Patrick discuss with Jesse: ok?
```
patrick-schweizer, line 63: @j-white I need a way to disable the upgrade for Integration tests. This is what I came up with. Is that the proper way to do that or do you have a better / more elegant idea? Thanks!

features/config/upgrade/src/test/java/org/opennms/config/upgrade/LiquibaseUpgraderIT.java
```diff
@@ -66,41 +70,70 @@
...
+        "classpath*:/META-INF/opennms/component-dao.xml",
+        "classpath:/META-INF/opennms/applicationContext-soa.xml",
+        "classpath:/META-INF/opennms/applicationContext-minimal-conf.xml",
+        "classpath:/META-INF/opennms/applicationContext-config-service.xml"})
+
 @JUnitTemporaryDatabase
 public class LiquibaseUpgraderIT implements TemporaryDatabaseAware<TemporaryDatabase> {
 
+    private final static String SCHEMA_NAME = "provisiond";
+    private final static String CONFIG_ID = "provisiond";
```
freddy61025, line 82: Although it is a test, we do need to think about how to unify the schema name and config id.

freddy61025 (approved):
It is good to me. Just one minor question we need to think about it.

features/config/upgrade/src/main/java/org/opennms/config/upgrade/UpgradeConfigService.java
j-white: Tricky. I think this approach is fine.

features/config/upgrade/src/main/resources/changelog-cm/29.0.0/changelog-cm.xml
```diff
@@ -9,6 +9,7 @@
 
     <changeSet author="pschweizer" id="1.0-register-schema-provisiond">
         <cm:registerSchema id="provisiond" version="1.0.0" entityClass="org.opennms.netmgt.config.provisiond.ProvisiondConfiguration"/>
```
j-white, line 11: This may be out of scope of this PR, but will mention here since it comes to mind:

The entity class, or the source used for the schema should be "frozen" for this to work reliably.

Assume an instance is upgraded from H28 to H30 and this is run for the time there.
Now assume we make changes to the XSD/model in H31.

Now, if another instance is upgraded directly from H28 to H31 - schema version 1.0.0 will be the model from H31 and not H30.

Is this being considered in the current approach?

patrick-schweizer: as discussed we will take this into a separate discussion / workstream

features/config/upgrade/src/test/java/org/opennms/config/upgrade/LiquibaseUpgraderIT.java
patrick-schweizer: I saw you had somewhere else:
 SCHEMA_NAME = "provisiond";
CONFIG_ID = "default";
Is that what you mean? I am good with that.

PR #3356, opened 2021-03-24
NMS-13187 Add Open API doc and Swagger UI to v2 RESTful API

jadyhou:
Use the URL below can get to the standard openapi doc page:
http://localhost:8980/opennms/api/v2/openapi.json

Use the URL below can get to the swagger ui page
http://localhost:8980/opennms/api/v2/api-docs/?url=/opennms/api/v2/openapi.json#/

All v2 restful endpoints are automatically scanned:

We can make rest call on the swagger UI directly:

### All Contributors

* [ ] Have you read and followed our [Contribution Guidelines](https://github.com/OpenNMS/opennms/blob/develop/CONTRIBUTING.md)?
* [ ] Have you [made an issue in the OpenNMS issue tracker](https://issues.opennms.org)?<br>If so, you should:
  1. update the title of this PR to be of the format: `${JIRA-ISSUE-NUMBER}: subject of pull request`
  2. update the JIRA link at the bottom of this comment to refer to the real issue number
  3. prefix your commit messages with the issue number, if possible
* [ ] Have you made a comment in that issue which points back to this PR?
* [ ] Have you updated the JIRA link at the bottom of this comment to link to your issue?
* [ ] If this is a new or updated feature, is there documentation for the new behavior?
* [ ] If this is new code, are there unit and/or integration tests?
* [ ] If this PR targets a `foundation-*` branch, does it avoid changing files in `$OPENNMS_HOME/etc/`?

### Pull Request Process

One or more reviewers should be assigned to each PR.

If you know that a particular person is subject matter expert in the area your PR affects, feel free to assign one or more reviewers when you create this PR, otherwise reviewers will be assigned for you.

If you have made additions or changes to the documentation, or if you _need_ documentation for these code changes, please make sure a technical writer has looked it over.

Once the reviewer(s) accept the PR and the branch passes continuous integration, the PR is eligible for merge.

At that time, if you have commit access (are an OpenNMS Group employee or a member of the Order of the Green Polo) you are welcome to merge the PR.
Otherwise, a reviewer can merge it for you.

Thanks for taking time to contribute!

### External References

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-13187

PR #3641, opened 2021-09-14
PE-18 adding swagger UI for provisiond REST APIs

maximbrener:
* JIRA (Issue Tracker): https://issues.opennms.org/browse/PE-18
* Once the server is up, swagger UI will be accessible through http://<hostname>:8980/opennms/swagger/index.html


j-white:
Just a heads up that we already have Swagger UI in the distribution: https://github.com/OpenNMS/opennms/pull/3356

We should probably review and consolidate before this is merged back to `develop`.

maximbrener:
in #3356 the swagger UI been built by annotations only on top of controller's methods, without actually creating openapi doc or any kind of validation on the request. in our case we generate openapi doc based on XSDs with all its validations inside and the only point the swagger UI to that generated openapi doc. not sure if we can even find a common ground between both implementations. I will setup a meeting to discuss

maximbrener:
found a way to leverage existing Swagger UI package with CM openapi doc

PR #3645, opened 2021-09-16
NMS-13566: reduce flakiness of ClassificationRulePageIT

swachter:
Issue: https://issues.opennms.org/browse/NMS-13566

Flakiness seems to be caused by insufficient memory and large variances in execution times in CI. This PR

* increases `-Xms` and `-Xmx` of the Horizon container from `1536m` to `2048m`
* changes the decision tree construction such that memory requirements are decreased
* increases the timeout when waiting for classification responses

In addition, `AsyncReloadingClassificationEngine` is made more robust by catching now any kind of `Throwable` and no longer `Exception` only.

Finally, the `ClassificationEngineBenchmark` was simplified and extended to include the pre-defined rule set that is shipped with OpenNMS.


cgorantla (approved):
LGTM!

smoke-test/src/test/java/org/opennms/smoketest/ClassificationRulePageIT.java
```diff
@@ -486,7 +486,9 @@ public void classifyAndWaitUntilAsserted(ClassificationRequestDTO classification
             execute(() -> findElementById("classification-submit")).click();
 
             await()
-                    .atMost(100, TimeUnit.SECONDS)
+                    // reloading the classification engine sometimes takes insanely long on CI
+                    // (this is further aggravated by the fact that one reload may be under way and another one may be pending)
```
RangerRick, line 490: Would it be possible, then, to have a new reload cancel a previous one, other than (presumably) queue-ing after it?

RangerRick (approved):
looks reasonable to me

smoke-test/src/test/java/org/opennms/smoketest/ClassificationRulePageIT.java
swachter: Good idea. I created an issue: https://issues.opennms.org/browse/NMS-13547

RangerRick: I think you meant https://issues.opennms.org/browse/NMS-13580 but yup, that's good

PR #3646, opened 2021-09-16
NMS-13568: backport docker content trust minion builds to Foundation 2021

RangerRick:
This is the first half of the work to backport DCT to foundation 2021 for use in Meridian Minion.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13568

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,6 +63,9 @@ aliases:
         "release-"*)
            MINION_IMAGE_VERSION="release-candidate"
           ;;
+        "foundation-"20[1-2][0-9])
+           MINION_IMAGE_VERSION="${CIRCLE_BRANCH}"
+          ;;
         "develop")
            MINION_IMAGE_VERSION="bleeding"
           ;;
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,6 +52,42 @@ defaults: &defaults
       type: string
       default: release-28.x
 
+aliases:
+  - &setup_dct_env
+    name: Setup DCT environment
+    command: |
+      case "${CIRCLE_BRANCH}" in
+        "master-"*)
+           MINION_IMAGE_VERSION="$(~/project/.circleci/scripts/pom2version.sh ~/project/pom.xml)"
+          ;;
+        "release-"*)
+           MINION_IMAGE_VERSION="release-candidate"
+          ;;
+        "develop")
+           MINION_IMAGE_VERSION="bleeding"
+          ;;
+        *)
+          MINION_IMAGE_VERSION=b<< pipeline.number >>
+          ;;
+      esac
+      echo "export DOCKER_CONTENT_TRUST=1" >> $BASH_ENV
+      echo "export DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE=\"$DCT_DELEGATE_KEY_PASSPHRASE\"" >> $BASH_ENV
+      echo "export MINION_IMAGE=docker.io/opennms/minion" >> $BASH_ENV
+      echo "export MINION_IMAGE_VERSION=\"$MINION_IMAGE_VERSION\"" >> $BASH_ENV
+  - &setup_dct_key
+    name: Setup DCT key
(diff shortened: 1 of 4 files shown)


cgorantla (approved):
LGTM!

RangerRick:
FYI this doesn't run on the branch because of circle config, but I had temporarily added this branch to the minion steps for testing.  You can [see the CircleCI run here](https://app.circleci.com/pipelines/github/OpenNMS/opennms/12369/workflows/8a01b41b-1084-4796-832f-0f29dab20daa), and [the images here](https://hub.docker.com/r/opennms/minion/tags?page=1&ordering=last_updated&name=b12369).

.circleci/config.yml
```diff
@@ -356,12 +405,26 @@ commands:
                 root: ~/
                 paths:
                   - code-coverage
+      - run:
+          name: Gather system logs
+          when: always
+          command: |
+            mkdir -p ~/build-results/system-logs
+            (dmesg || :) > ~/build-results/system-logs/dmesg 2>&1
```
swachter, line 413: I am always keen on learning new things. What does `(dmesg || :)` do?

.circleci/config.yml
```diff
@@ -841,38 +932,6 @@ jobs:
...
-                     BUILD_BRANCH="${CIRCLE_BRANCH}"
-                ;;
-              "develop")
-                make DOCKER_ARCH="linux/amd64,linux/arm/v7,linux/arm64" \
-                     DOCKER_FLAGS=--push \
-                     VERSION="bleeding" \
-                     BUILD_NUMBER="${CIRCLE_BUILD_NUM}" \
-                     BUILD_URL="${CIRCLE_BUILD_URL}" \
-                     BUILD_BRANCH="${CIRCLE_BRANCH}"
-                ;;
```
swachter, line 871: The comment says that we build for multiple architectures only in release branches. The code says something other.

swachter (approved):
LGTM. I have only two minor questions.

DCT is really a complicated beast! I think that @indigo423 did some experiments with newer Docker versions that would at least allow to get rid of some of the `buildx` complications. Is there a ticket for that? We should also regularly check if `buildx` eventually supports signing of multi-arch images out of the box.

.circleci/config.yml
RangerRick: `:` is a noop in shell, it's just to make sure if `dmesg` for some reason fails, it doesn't cause the run to error out

.circleci/config.yml
```diff
@@ -52,6 +52,45 @@ defaults: &defaults
...
+           MINION_IMAGE_VERSION="release-candidate"
+          ;;
+        "foundation-"20[1-2][0-9])
+           MINION_IMAGE_VERSION="${CIRCLE_BRANCH}"
+          ;;
+        "develop")
+           MINION_IMAGE_VERSION="bleeding"
+          ;;
+        *)
+          MINION_IMAGE_VERSION=b<< pipeline.number >>
```
j-white, line 73:
```suggestion
          MINION_IMAGE_VERSION="${CIRCLE_BRANCH//\//-}"
```

Let's use the same version logic as the Horizon/Sentinel containers - I ran into this in some other branches recently.

.circleci/config.yml
RangerRick: well the comment is being deleted, so I guess it's alright now :D

.circleci/config.yml
RangerRick: works for me, I was just leaving things the way they were in 28/develop as much as possible

RangerRick:
> LGTM. I have only two minor questions.
> 
> DCT is really a complicated beast! I think that @indigo423 did some experiments with newer Docker versions that would at least allow to get rid of some of the `buildx` complications. Is there a ticket for that? We should also regularly check if `buildx` eventually supports signing of multi-arch images out of the box.

Not finding a ticket for that, but it's definitely out of scope just for this issue.
If you have a better idea of what it will take to replace what's there, feel free to write up an issue for it.

PR #3662, opened 2021-09-21
NMS-13547: adds CachingHealthCheck for Rest/Broker

swachter:
Issue: https://issues.opennms.org/browse/NMS-13547

This PR adds the following two passive health checks (i.e. checks that do not tests something of their own) to Minion:

1. Track/Check communication via the `RestClient`
2. Track/Check for received echo rpcs

The the health check command and rest API is augmented by an `maxAgeMs` parameter. Tracked communication/rpcs must not be older in order to be considered by the passive checks.


core/health/api/pom.xml
```diff
@@ -11,4 +11,10 @@
     <artifactId>org.opennms.core.health.api</artifactId>
     <packaging>bundle</packaging>
     <name>OpenNMS :: Core :: Health :: API</name>
+  <dependencies>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+    </dependency>
```
swachter, line 18: Must this dependency reflected in the feature/bundle dependencies, too?

core/health/api/src/main/java/org/opennms/core/health/api/DefaultPassiveHealthCheck.java
```diff
@@ -0,0 +1,87 @@
...
+import java.time.Duration;
+import java.time.Instant;
+import java.util.List;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A health that does no action by its own but relies on being informed about healthiness.
+ */
```
swachter, line 40:
```suggestion
/**
 * A health check that does no action by its own but relies on being informed about healthiness.
 */
```

core/ipc/rpc/common/pom.xml
```diff
@@ -40,5 +40,17 @@
...
+      <artifactId>org.opennms.core.health.api</artifactId>
+      <version>28.1.0-SNAPSHOT</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.opennms.core.health</groupId>
+      <artifactId>org.opennms.core.health.impl</artifactId>
+      <version>28.1.0-SNAPSHOT</version>
+      <scope>compile</scope>
+    </dependency>
```
swachter, line 54:
```suggestion
```

features/minion/repository/pom.xml
```diff
@@ -74,6 +74,7 @@
                                 <feature>dominion-grpc-client</feature>
                                 <feature>dominion-secure-credentials-vault</feature>
                                 <featue>minion-rest-service</featue>
+                                <feature>minion-health-check</feature>
```
cgorantla, line 77: Also need to add dependency below for <artifactId>health-check</artifactId>

core/ipc/rpc/common/src/main/java/org/opennms/core/rpc/echo/HealthTrackingEchoRpcModule.java
```diff
@@ -0,0 +1,79 @@
...
+            HealthCheckResponseCache healthCheckResponseCache
+    ) {
+        super(EchoRequest.class, EchoResponse.class);
+        this.delegate = delegate;
+        this.healthCheckResponseCache = healthCheckResponseCache;
+    }
+
+    @Override
+    public CompletableFuture<EchoResponse> execute(EchoRequest request) {
+        LOG.debug("received echo request - id: {]", request.getId());
```
cgorantla, line 62: `LOG.debug("Received Echo request with id: {}", request.getId());`

This id is actually System time in msec, we can add it as received time.

core/health/api/pom.xml
cgorantla: no need, should be available by default in karaf i think

cgorantla (commented):
Few minor comments, implementation looks pretty good to me.

Haven't tried this locally, will validate and give you feedback

core/ipc/rpc/common/pom.xml
```diff
@@ -40,5 +40,11 @@
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.opennms.core.health</groupId>
+      <artifactId>org.opennms.core.health.api</artifactId>
+      <version>28.1.0-SNAPSHOT</version>
```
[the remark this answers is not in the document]
swachter: done

features/minion/repository/pom.xml
swachter: done

core/ipc/rpc/common/src/main/java/org/opennms/core/rpc/echo/HealthTrackingEchoRpcModule.java
swachter: done

core/health/api/pom.xml
swachter: When I remove that dependency the build fails.

core/health/shell/src/main/java/org/opennms/core/health/shell/HealthCheckCommand.java
```diff
@@ -59,6 +60,9 @@
     @Option(name = "-t", description = "Maximum number of milliseconds to wait before failing when waiting for a check to complete (e.g. try to establish a JMS session.")
     public long timeout = 5L * 1000L;
 
+    @Option(name = "--maxAge", description = "Maximum age of cached health check responses in milliseconds. Defaults to zero, i.e. cached responses of health checks are not considered.")
+    public long maxAgeMs = 0;
+
```
cgorantla, line 65: Can we add an option or argument for `tag` here for the karaf command ?

core/ipc/rpc/common/src/main/java/org/opennms/core/rpc/echo/HealthTrackingEchoRpcModule.java
```diff
@@ -0,0 +1,79 @@
...
+            HealthCheckResponseCache healthCheckResponseCache
+    ) {
+        super(EchoRequest.class, EchoResponse.class);
+        this.delegate = delegate;
+        this.healthCheckResponseCache = healthCheckResponseCache;
+    }
+
+    @Override
+    public CompletableFuture<EchoResponse> execute(EchoRequest request) {
+        LOG.debug("received echo request - receivedTime: {]", request.getId());
```
cgorantla, line 62: Should be 
`LOG.debug("Received echo request - receivedTime: {}", request.getId());`

core/health/api/src/main/java/org/opennms/core/health/api/DefaultPassiveHealthCheck.java
```diff
@@ -0,0 +1,87 @@
...
+            Status initialStatus
+    ) {
+        this.description = description;
+        this.tags = tags;
+        setResponse(new Response(initialStatus));
+    }
+
+    @Override
+    public synchronized void setResponse(Response response) {
+        LOG.debug("cache response - healthCheck: {}; status: {}; msg: {}", description, response.getStatus(), response.getMessage());
```
cgorantla, line 63: nit: we usually start LOG statements with Capitals.

features/minion/core/health-check/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -0,0 +1,51 @@
+<blueprint
+    xmlns="http://www.osgi.org/xmlns/blueprint/v1.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="
+        http://www.osgi.org/xmlns/blueprint/v1.0.0 
+        https://osgi.org/xmlns/blueprint/v1.0.0/blueprint.xsd
+">
+
+    <bean id="passiveEchoRpcHealthCheck" class="org.opennms.core.health.api.DefaultPassiveHealthCheck">
```
cgorantla, line 9: Although you have included `passive` for Echo RPC as well, that doesn't seem to reflect in Rest Client response. It only gives out OpenNMS Rest Response(Passive)

Can you check this ?
Screen Shot 2021-09-26 at 10 36 04 PM
?

cgorantla (commented):
Few more comments.
Also pushed  jira/NMS-13547-smoke branch to get full covergae.

core/health/api/src/main/java/org/opennms/core/health/api/DefaultPassiveHealthCheck.java
swachter: done

features/minion/core/health-check/src/main/resources/OSGI-INF/blueprint/blueprint.xml
swachter: Fixed.
`HealthCheckRestServiceImpl` did assemble health check descriptions and health check results in a somewhat complicated and brittle way. Now, health check instances and their execution results a kept toghether.

core/ipc/rpc/common/src/main/java/org/opennms/core/rpc/echo/HealthTrackingEchoRpcModule.java
swachter: done

core/health/shell/src/main/java/org/opennms/core/health/shell/HealthCheckCommand.java
swachter: I also thought of that and already have something prepared. I had in mind to replace `HealthCheckConstants` that now defines the string values of tags by an `enum HealthCheckTag`. The `enum` can be used to support the completer. However, this would mean that all health checks have to be changed slightly because the `getTags` method now would return a list of `HealthCheckTag` instead of a `List<String>`. Having our recent discussion about PR size in mind, I was not sure if I could/should include that work here.

WDYT? The additional changes are straigthforward and rather mechanicual. I would be happy to include these changes here.

cgorantla: We can keep List<String>, let the user provide Option.

swachter: Can you clarify? Should we keep `List<String>`? In that case the possible tags would not be represented by an `enum` and the set of values would have to be repeated in the completer. This is what I wanted to avoid.

cgorantla: Enum would be more useful if it is one `String` that we represent.  Since we can have multiple tags, it would be better to use List<String>.  I'm okay with Karaf command update in a separate PR.

swachter: https://issues.opennms.org/browse/NMS-13592

cgorantla:
Can we add default `maxAgeMs` for CachedHealthChecks as `60000`

It seems like now default `health:check` may not return healthy always. This might be confusing for users.

smoke-test/src/main/java/org/opennms/smoketest/utils/KarafShellUtils.java
```diff
@@ -50,7 +50,7 @@ public static boolean testHealthCheck(InetSocketAddress sshAddr, AtomicReference
         try (final SshClient sshClient = new SshClient(sshAddr, OpenNMSContainer.ADMIN_USER, OpenNMSContainer.ADMIN_PASSWORD)) {
             // Issue the 'opennms:health-check' command
             PrintStream pipe = sshClient.openShell();
-            pipe.println("opennms:health-check");
```
cgorantla, line 53: we can revert this.

smoke-test/src/test/java/org/opennms/smoketest/health/HealthCheckIT.java
```diff
@@ -99,7 +99,7 @@ private void verifyHealthCheck(final int expectedHealthCheckServices, final Inet
                 .until(() -> {
                     try (final SshClient sshClient = new SshClient(sshAddress, OpenNMSContainer.ADMIN_USER, OpenNMSContainer.ADMIN_PASSWORD)) {
                         final PrintStream pipe = sshClient.openShell();
-                        pipe.println("opennms:health-check");
+                        pipe.println("opennms:health-check --maxAge 60000");
```
cgorantla, line 102: This one as well

cgorantla (approved):
👍

cgorantla:
Minion Rest Service query with tags local & passive with maxAgeMs=60000 seems to be good for querying Minion health Check without depending on any direct Rest/Broker checks. http://localhost:8181/minion/rest/health?tag=local&tag=passive&maxAgeMs=60000.

jadyhou (approved):

PR #3669, opened 2021-09-23
PE-13: Notify relevant services about config changes

freddy61025:
https://issues.opennms.org/browse/PE-13
Notify relevant services about config changes

It add a default callback which will update reference config entity object. It also support additional callback chain.


features/config/service/api/src/main/java/org/opennms/features/config/service/api/ConfigUpdateInfo.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * For more information contact:
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.features.config.service.api;
+
+public class ConfigUpdateInfo {
```
patrick-schweizer, line 31: I understand correct that this is a composite key for a configuration (schema + configId), correct?
If so, a few remarks:
- since it's a key we should implement equals() and hashCode() so we can use it in Maps etc.
- maybe we should call it ConfigKey to express that it is a unique key for a config?

features/config/service/impl/src/main/java/org/opennms/features/config/service/impl/ConfigurationManagerServiceImpl.java
```diff
@@ -34,22 +34,24 @@
...
+import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Consumer;
 
 @Component
 public class ConfigurationManagerServiceImpl implements ConfigurationManagerService {
     private final static Logger LOG = LoggerFactory.getLogger(ConfigurationManagerServiceImpl.class);
     private final ConfigStoreDao<JSONObject> configStoreDao;
+    // This map contains key: configName value: list of Consumer
+    private final ConcurrentHashMap<String, Collection<Consumer<ConfigUpdateInfo>>> onloadNotifyMap = new ConcurrentHashMap<>();
```
patrick-schweizer, line 54: I m not clear why the key in the map is the configName (which should be (schemaName btw.) and not ConfigUpdateInfo?
If we change a config isn't is always configName + configId?

patrick-schweizer (commented):
I have questions about the keys. I am not sure I understand the Map.

features/config/service/impl/src/main/java/org/opennms/features/config/service/impl/ConfigurationManagerServiceImpl.java
freddy61025: I am thinking make it safer to trigger reload for all config for under same configName

patrick-schweizer: ok, understand. I am good with that.

patrick-schweizer (dismissed):

PR #3680, opened 2021-09-28
NMS-13587: change how "$REVISION" is created by default

RangerRick:
This PR changes how $REVISION is set for docker containers.
Currently it's running `git describe --always` which goes back into history looking for a related tag to be considered "base", and then adds the short hash to the end. As of now, this usually matches `meridian-foundation-2021.1.4` which is technically not wrong, and is unique, but is not a very intuitive value for, say, develop snapshots.

This change makes it be the branch name + pom version + short hash.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13587


bouff (approved):
LGTM!

PR #3689, opened 2021-09-30
Jira/NMS-13542: Geo-Map: Acknowlege/Unackowlege/Escalate/Clear Alarm

jadyhou:
The main work is to implement the Acknowlege/Unackowlege/Escalate/Clear Alarm on the geomap page. Also fixed a few small bugs/typo in the mean time. 

Here is how the Acknowlege/Unackowlege/Escalate/Clear alarm function works:
1. Before any alarm is selected, the dropdown and the submit button are disabled to prevent users from clicking by accident.

Screen Shot 2021-09-30 at 3 42 03 PM

2. The dropdown and the submit button will be enabled one one or more alarms are selected
Screen Shot 2021-09-30 at 3 42 29 PM

3.  After the "Submit" button is clicked, there will be a flash message to tell the user if the request is success processed with the animation of blue color fade out.
Screen Shot 2021-09-30 at 3 43 08 PM

4. After the animation of the flash message, the page will be reloaded to update the page to the latest data.

Screen Shot 2021-09-30 at 3 43 30 PM

### External References

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-13542

### DIFF
--- a/geomap/src/types/index.ts
+++ b/geomap/src/types/index.ts
@@ -28,7 +28,7 @@ export interface MapNode {
 export interface Alarm {
   id: string
   severity: string
-  nodeId: number
+  nodeId: string
   nodeLabel: string
   uei: string
   count: number
@@ -114,4 +114,15 @@ export interface SortProps {
   rows: Number
   sortField: string
   sortOrder: 1 | -1
+}
+
+export interface AlarmQueryParameters {
+  ack?: boolean
+  clear?: boolean
+  escalate?: boolean
+}
+
+export interface AlarmModificationQueryVariable {
+  pathVariable: string
+  queryParameters:  AlarmQueryParameters
 }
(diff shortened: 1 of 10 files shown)


geomap/src/types/index.ts
```diff
@@ -28,7 +28,7 @@ export interface MapNode {
 export interface Alarm {
   id: string
   severity: string
-  nodeId: number
+  nodeId: string
```
RangerRick, line 31: Is there a reason you switched this to a string? It should be a number...

jadyhou: Yeah, both number and string will work. I just tried to make it consistent with the node id in the "Node" in the same file that is migrated from foundation-ui.

RangerRick: My suspicion is that `node` is a field that can be either a `nodeId` (number) _or_ a `foreignSource:foreignId` pair. But in this case, `nodeId` should always be the node ID, which is a number.

jadyhou: It's probably because everything UI receives is a string, so I guess it's not that matter if it's consistent with the backend. I just updated it to be number just now. Please check.

PR #3696, opened 2021-10-04
NMS-13501: Added TrustManager implementation

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-13501


features/ifttt/src/main/java/org/opennms/features/ifttt/helper/IfTttTrigger.java
```diff
@@ -81,8 +88,27 @@ public IfTttTrigger event(final String event) {
         return this;
     }
 
+    public SSLContext getSslContext() {
+        if (sslContext == null) {
+            try {
+                sslContext = SSLContext.getInstance("SSL");
```
fooker, line 94:
```suggestion
                sslContext = SSLContext.getInstance("TLS");
```

fooker (changes requested):
Don't like disabling encryption verification as this sends credentials to external servers. I would expect this to use the systems certificate store and verify the servers certs against it.

features/ifttt/src/main/java/org/opennms/features/ifttt/helper/IfTttTrigger.java
christianpape: See next commit

fooker (approved):
Wonderful!

PR #3719, opened 2021-10-07
PE-117: Merge CM-Phase1 > CM-Phase2 after CM-Phase1 sync with develop

freddy61025:
https://issues.opennms.org/browse/PE-117


freddy61025:
this is duplicated, should be removed.

PR #3721, opened 2021-10-08
NMS-13590: execute health checks asynchronously and in parallel

swachter:
Issue: https://issues.opennms.org/browse/NMS-13590

This PR:
* Executes health checks concurrently
* Updates the Karaf console as results come in
* Adds a `FutureUtils` class that may be useful at other places


swachter:
> Would it be more cleaner to implement ` CompletableFuture<Response> perform(Context context)` and combine them when you get response ?

Yes, this would be nice, then the single health checks could take advantage for example from non-blocking IO. Maybe this could be done in a seconds step?

PR #3723, opened 2021-10-08
NMS-13619: Document existing behavior of edge status for the (en)linkd topology provider.

j-white:
JIRA: https://issues.opennms.org/browse/NMS-13619


j-white:
> This file has moved to a completely different path in foundaton-2021, and will need to be updated to match. https://github.com/OpenNMS/opennms/blob/foundation-2021/docs/modules/operation/pages/topology/topology.adoc

This could be handled when merging forward. Should we do something different in this PR?

mmahacek:
I'd defer to @Bonrob2 for that answer.

PR #3729, opened 2021-10-12
NMS-13316: Sort monitoring locations in V1 endpoint

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-13316


fooker:
> > Why not use `OnmsRestService#applyQueryFilters` instead?
> 
> The V2 endpoint sorts by default. So, I just did the same for the V1 endpoint.

Ok, that's fine.

PR #3809, opened 2021-11-04
NMS-13719: Fixed NPE in DefaultProvisionService

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-13719

### DIFF
--- a/opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/DefaultProvisionService.java
+++ b/opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/DefaultProvisionService.java
@@ -262,8 +262,11 @@ public class DefaultProvisionService implements ProvisionService, InitializingBe
         node.setMetaData(dbNode.getMetaData());
         for(final OnmsIpInterface onmsIpInterface : dbNode.getIpInterfaces()) {
             final InetAddress inetAddress = onmsIpInterface.getIpAddress();
-            if (inetAddress != null && node.getIpInterfaceByIpAddress(inetAddress) != null) {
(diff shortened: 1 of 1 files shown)


opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/DefaultProvisionService.java
```diff
@@ -262,7 +262,7 @@ public void updateNode(final OnmsNode node, String rescanExisting) {
         node.setMetaData(dbNode.getMetaData());
         for(final OnmsIpInterface onmsIpInterface : dbNode.getIpInterfaces()) {
             final InetAddress inetAddress = onmsIpInterface.getIpAddress();
-            if (inetAddress != null) {
+            if (inetAddress != null && node.getIpInterfaceByIpAddress(inetAddress) != null) {
```
RangerRick, line 265: Should probably pull the results of `node.getIpInterfaceByIpAddress(inetAddress)` out into a variable so it doesn't have to iterate the interfaces twice.

RangerRick (changes requested):

indigo423 (approved):
I've tested the latest build and the NPE was fixed. So function-wise 👍 :)

opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/DefaultProvisionService.java
christianpape: Yes, sure. Makes sense.

fooker (approved):

PR #3812, opened 2021-11-04
NMS-13720: Initial Vue 3 & featherDS new UI

mikewrosey:
JIRA: https://issues.opennms.org/browse/NMS-13720


.gitignore
```diff
@@ -32,6 +32,8 @@ yarn-error.log
 /features/discovery/activemq-data
 /features/topology-map/org.opennms.features.topology.app/www-test/
 /opennms-alarms/jms-northbounder/activemq-data/
+ui/node_modules/
+ui/package-lock.json
```
RangerRick, line 36: `package-lock.json` should generally be checked into git, so that upstream changes don't break builds

.github/dependabot.yml
```diff
@@ -1,5 +1,20 @@
 version: 2
 updates:
+
+  - package-ecosystem: "npm"
+    # Look for `package.json` and `lock` files in the `root` directory
+    directory: "/web"
```
j-white, line 6: needs to be renamed to `/ui`

opennms-web-api/src/main/java/org/opennms/web/servlet/SpaRoutingFilter.java
```diff
@@ -0,0 +1,78 @@
...
+            }, response);
+            return;
+        }
+        chain.doFilter(request, response);
+    }
+
+    private boolean isClientPath(final String uri) {
+        return uri.startsWith("/ui/")
+                && !uri.startsWith("/ui/assets/")
+                // FIXME: This is here to serve the favicon, we could move it under /assets/, find another approach, or delete this comment and keep it as-is
```
j-white, line 65: can reword this comment and keep as-is

.gitignore
mikewrosey: The build uses yarn and produces a yarn.lock file, which is checked in.
This was more of a safety from some devs who were using NPM locally and creating/checking in the unneeded package-lock.

mikewrosey: I remove the line though, if we're not worried about that

.github/dependabot.yml
mikewrosey: done

opennms-web-api/src/main/java/org/opennms/web/servlet/SpaRoutingFilter.java
mikewrosey: done

.gitignore
j-white: I think it's safe to ignore given we use and check-in the `yarn.lock`.

@RangerRick do you think we should check-in both?

RangerRick: Ah, sorry, if you're using yarn then never mind, I missed that.

j-white (approved):

PR #3815, opened 2021-11-05
Add event to sentinel

cajunesque:
I will be porting over the sping content of the app context into this blueprint file so that we can actually see the service from the container


j-white:
Thanks for the test. Closing this one until we're further along :)

PR #3829, opened 2021-11-09
NMS-13268: Karaf: Globally Prevent AutoRefresh Cascade on Feature Install

swachter:
Issue: https://issues.opennms.org/browse/NMS-13268

This PR configures Karaf such that bundles are not refreshed when features are installed. Without that change a bundle is stopped / started if a newly installed feature brings an optional dependency or a newer version of a dependency of that bundle. That seems to be unnecessary in case of OpenNMS Karaf setups.

In addition, this PR reformats two lines of code where the flaky `BMSTopologyIT.canChangeFocusWhenInSimulationMode` test randomly fails with a NPE. Next time that test failes, more precise NPE information may help to fix the flakiness of that test.

### DIFF
--- a/container/karaf/src/main/filtered-resources/etc/org.apache.karaf.features.cfg
+++ b/container/karaf/src/main/filtered-resources/etc/org.apache.karaf.features.cfg
@@ -143,6 +143,11 @@ featuresBoot = ( \
 featuresBootAsynchronous=false
 
 #
+# Define if the feature service automatically refreshes bundles
+#
+autoRefresh=false
+
+#
 # Service requirements enforcement
 #
 # By default, the feature resolver checks the service requirements/capabilities of
--- a/container/shared/src/main/filtered-resources/etc/org.apache.karaf.features.cfg
+++ b/container/shared/src/main/filtered-resources/etc/org.apache.karaf.features.cfg
@@ -78,6 +78,11 @@ featuresBoot = \
 featuresBootAsynchronous=false
 
 #
+# Define if the feature service automatically refreshes bundles
+#
+autoRefresh=false
+
+#
 # Service requirements enforcement
 #
 # By default, the feature resolver checks the service requirements/capabilities of
--- a/features/container/sentinel/src/main/filtered-resources/etc/org.apache.karaf.features.cfg
+++ b/features/container/sentinel/src/main/filtered-resources/etc/org.apache.karaf.features.cfg
@@ -75,6 +75,11 @@ featuresBoot = \
 featuresBootAsynchronous=false
 
 #
+# Define if the feature service automatically refreshes bundles
+#
+autoRefresh=false
+
+#
 # Service requirements enforcement
 #
 # By default, the feature resolver checks the service requirements/capabilities of
--- a/src/assembly/source.xml
+++ b/src/assembly/source.xml
@@ -20,6 +20,7 @@
         <exclude>**/.settings/**/*</exclude>
         <exclude>**/.DS_Store</exclude>
         <exclude>**/.idea</exclude>
+        <exclude>**/*.iml</exclude>
         <exclude>**/target</exclude>
         <exclude>**/target/**/*</exclude>
       </excludes>


fooker (approved):

j-white:
May be okay, but I can't say that I'm confident this won't have some unintended side effects.

How much does this impact startup time?

RangerRick:
> May be okay, but I can't say that I'm confident this won't have some unintended side effects.

My fear as well, but I guess we have an entire release schedule to make sure it works OK... :D

RangerRick (dismissed):

swachter:
> May be okay, but I can't say that I'm confident this won't have some unintended side effects.
> 
> How much does this impact startup time?

I attached some log outputs and added a comment with some numbers to the ticket: [NMS-13268](https://issues.opennms.org/browse/NMS-13268?focusedCommentId=66317&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-66317).

TLDR: For the Alec plugin the speedup was 669 ms.

Agreeable, that may not be worth the change. OTOH having less bundles going up and down may be a benefit on its own. I also do not have experience on the number of plugins in typical OpenNMS installations.

One by-product of my investigation was that the way features are installed by the `KarafExtender` is not influenced by the `autoRefresh` configuration setting. AutoRefresh behavior of the `KarafExtender` must be set by giving a parameter when calling the `FeatureService`.

Another by-product was that it may be worth to investigate what is happening between feature installs. Startup pauses in the range of 10s of seconds.

swachter:
Some more information: Karaf 4.3.2 has a bug that prevents the usage of the `autoFlush` configuration setting (cf. [KARAF-7282](https://issues.apache.org/jira/browse/KARAF-7282)). Auto refresh can be prevented programmatically only.

RangerRick:
Sounds like we should revisit this when 4.3.4 is out.

swachter:
> Sounds like we should revisit this when 4.3.4 is out.

If we wait for 4.3.4 then auto-refresh on feature installs can be completely disabled by the global configuration setting `autoRefresh=false`. This will apply to both, feature installs on startup and manual feature installs. No code changes are necessary to `KarafExtender` and `KarDeployer`.

If we do not wait for 4.3.4 then we could disable auto-refreshs on startup only.

In summary, I also opt for waiting until we upgrade to Karaf 4.3.4.

RangerRick:
We need to update to 4.3.4 to continue to work on JDK17 support anyway, so it's gonna be a priority going forward.

RangerRick:
@swachter should be able to look at this again as soon as https://github.com/OpenNMS/opennms/pull/4112 is merged

RangerRick:
Hrm, the failures in `smoke-test-flaky` look like they could actually be from things not loading right...?

swachter:
> Hrm, the failures in `smoke-test-flaky` look like they could actually be from things not loading right...?

It is each time a different one: `ReportRestIT`, `canChangeFocusWhenInSimulationMode`, and `verifyBmpProcessing`. Not sure what that means, either. I am inclined that the change does not introduce a problem. I will have a closer look on the logs, though.

RangerRick:
> > Hrm, the failures in `smoke-test-flaky` look like they could actually be from things not loading right...?
> 
> It is each time a different one: `ReportRestIT`, `canChangeFocusWhenInSimulationMode`, and `verifyBmpProcessing`. Not sure what that means, either. I am inclined that the change does not introduce a problem. I will have a closer look on the logs, though.

I think it is unlikely it's a pre-existing condition since I don't see these kinds of errors in the main build. We _do_ get some flapping, but not in this particular way. This almost looks like some kind of race condition or missing dependency or something.

swachter:
* The failure in `ReportRestIT` relates to [NMS-13961](https://issues.opennms.org/browse/NMS-13961) which is hopefully fixed soon.

* The failure in `canChangeFocusWhenInSimulationMode` is caused by a NPE in `GraphPainter.getSourceKey(Edge)`. In principle, that method is allowed to return `null` but does no `null` check if the vertex exists before `.getKey()` is called:

```
	private String getSourceKey(Edge edge) {
		return m_graphContainer
				.getTopologyServiceClient()
				.getVertex(
						edge
								.getSource()
								.getVertex(),
						m_graphContainer.getCriteria()
				)
				.getKey();
	}
```

Admittedly, this seems to be a timing issue because sometimes that test fails and sometimes it does not. This PR changes the startup timing. However, I think that the issue should be resolved on its own. I am tempted to "fix" the issue by just adding a null check for the vertex. Maybe someone has more insight into this??? WDYT?

* Unfortunately, I have no idea about the failed `verifyBmpProcessing` test.

RangerRick:
> * In principle, that method is allowed to return `null` but does no `null` check if the vertex exists before `.getKey()` is called

Yeah, this has been flapping elsewhere as well, not just your branch. Does seem correct to null-check, and it affects all the way back to `foundation-2019` so I'll play with it there.

RangerRick (dismissed):
assuming it passes full smoke once you fix conflicts/merge the updates, LGTM

cgorantla:
> > assuming it passes full smoke once you fix conflicts/merge the updates, LGTM
> 
> There were always some smoke tests failing even after 3 "return from failed":
> 
> * rerun: `SyslogKafkaElasticsearchIT`
> * rerun: `testHeartbeatOutages`
> * initial run: `verifyStatusEnrichmentApplication`, `testSnmpV3TrapsOnMinion`, and `testRestHealthServiceOnSentinel`
> 
> Either I am extremely unlucky or this change makes our tests indeed more flaky. I am unsure about this PR but tend to just close it.

Some of these are flaky in general.
Overall, Since we are not sure what are the side effects of this change, i would vote for closing it for now.
We can revisit this later.

PR #3832, opened 2021-11-09
NMS-13652: Retry

fooker:
Add a retry while requesting a twin update.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13652


christianpape (approved):
Looks good.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
```diff
@@ -230,72 +136,183 @@ protected TwinUpdate mapTwinResponseToProto(byte[] responseBytes) {
...
+            if (this.value != null && Objects.equals(this.value.value, value.value)) {
+                return;
+            }
+
+            // Remember value
+            this.value = value;
+
+            // Cancel outstanding retry
+            if (this.retry != null) {
+                this.retry.cancel(false);
```
swachter, line 250: Is there a specific reason why `cancel(false)` is used and not `cancel(true)`?

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
```diff
@@ -230,72 +136,183 @@ protected TwinUpdate mapTwinResponseToProto(byte[] responseBytes) {
...
-        public int hashCode() {
-            return Objects.hash(key, consumer, clazz);
+        private synchronized void request() {
+            // Send a request
+            final var request = new TwinRequest(this.key, AbstractTwinSubscriber.this.identity.getLocation());
+            AbstractTwinSubscriber.this.sendRpcRequest(request);
+
+            // Cancel an outstanding retry
+            if (this.retry != null) {
+                this.retry.cancel(false);
```
swachter, line 265: Is there a specific reason why `cancel(false)` is used and not `cancel(true)`?

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
```diff
@@ -230,72 +136,183 @@ protected TwinUpdate mapTwinResponseToProto(byte[] responseBytes) {
...
+
+            // Skip if value has not changed
+            if (this.value != null && Objects.equals(this.value.value, value.value)) {
+                return;
+            }
+
+            // Remember value
+            this.value = value;
+
+            // Cancel outstanding retry
```
swachter, line 248: An outstanding retry should also be canceled in case the value did not change. Otherwise we will retry forever.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
```diff
@@ -30,177 +30,83 @@
...
-
-    private boolean isObjectUpdated(TwinTracker twinTracker, TwinUpdate twinUpdate) {
-        if (twinUpdate.isPatch()) {
-            return true;
-        }
-        return !Arrays.equals(twinTracker.getObj(), twinUpdate.getObject());
-    }
+        // Consume in thread instead of using broker's callback thread.
+        this.executorService.execute(() -> {
+            final var subscription = this.subscriptions.computeIfAbsent(twinUpdate.getKey(), Subscription::new);
```
swachter, line 101: Should a subscription (without a consumer) be created? When the last consumer of a subscription is closed then the subscription gets removed from the map. This indicates that subscriptions without consumers are not stored.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
```diff
@@ -230,72 +136,183 @@ protected TwinUpdate mapTwinResponseToProto(byte[] responseBytes) {
...
+                        // Version advanced - apply path
+                        try {
+                            final var patchObj = AbstractTwinSubscriber.this.objectMapper.readTree(update.getObject());
+                            final var patch = JsonPatch.fromJson(patchObj);
+
+                            final var value = patch.apply(this.value.value);
+
+                            this.accept(new Value(update.getSessionId(), update.getVersion(), value));
+                        } catch (JsonPatchException e) {
+                            throw new IOException("Unable to apply patch", e);
```
swachter, line 307: Instead of throwing an `IOException` you could log the problem and just call `this.request()` like in the other cases where the value could not be processed.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
```diff
@@ -30,177 +30,83 @@
...
-                    }
-                });
+            try {
+                subscription.update(twinUpdate);
+            } catch (final IOException e) {
+                LOG.error("Processing update failed: {}", twinUpdate.getKey(), e);
+                subscription.request();
             }
-        }
+        });
```
swachter, line 109: The subscription could call `request` on itself instead of throwing an `IOException`.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
```diff
@@ -230,72 +136,183 @@ protected TwinUpdate mapTwinResponseToProto(byte[] responseBytes) {
...
-            if (o == null || getClass() != o.getClass()) return false;
-            SessionImpl<?> session = (SessionImpl<?>) o;
-            return Objects.equals(key, session.key) && Objects.equals(consumer, session.consumer) && Objects.equals(clazz, session.clazz);
+        private synchronized void accept(final Value value) {
+            Objects.requireNonNull(value);
+
+            // Skip if value has not changed
+            if (this.value != null && Objects.equals(this.value.value, value.value)) {
+                return;
+            }
```
swachter, line 243: What happens if the value does not change, but the session/version? I think in that case the `Value` should be replaced but consumers need not to be notified.

swachter (changes requested):
Overall, code is in good shape. Some questions, though.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
fooker: Because this would interrupt a ongoing send action and may leave the sender in an undesired state. Changes are low this would happen (comparing retry interval with the amount of time spend sending the request). There is no harm in sending the request.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
fooker: Same as above.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
fooker: Same as above.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
fooker: The subscription is created to store incoming values before a consumer is registered.

This is a trade-off between leaking memory and early-accepting values. I see more benefit in early-accepting values to spare some requests. Therefore skipping the removal for now.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
fooker: This would multiply the logging and the call to `request` to all places where an exception is thrown.
I foresee some more advanced error handling required in the future, which can happen in this exact spot, too.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
fooker: This is clearly an error case related to an internal protocol error - not something we would expect on a regular basis. Therefore handling it the same as other decoding errors.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
fooker: Why should the value be replaced if it does not change? The equal-semantics on a JSON object are well-defined. There is no harm in keeping the old object.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
swachter: This is not the same as above. `void request` schedules new requests forever. An outstanding retry must be canceled (and the member set to `null`) when a value is accepted, irrespective if the value is equal to the current value or not.

Even if `void accept` cancels the current retry there is a slight chance that retries may go on forever:

- retry was scheduled
- accept is entered
- retry starts to execute and waits to enter the synchronized `request` method
- accept cancels the retry (but does not interrupt)
- accept is left
- retry enters `request`
- a request is sent and another retry is scheduled

I think we could use the interruption mechanism here. In `void request` the current thread should check if it is interrupted and in that case it would not send a request or schedule another retry.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
swachter: I see no benefit in splitting the processing logic of updates into two places. However, this is a matter of taste.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
swachter: The session/version could have changed.

fooker: granted.

swachter: AFAIS it is not granted. `accept` is called if the sessionId or version has changed and the update was not a patch.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
fooker: I've updated this to update the value container on version/session change.

swachter (commented):
The retry logic should be fine now. I can approve it. However, I would like you to consider an alternative implementation that seems easier to me and that does not rely on a single threaded executor:

* Subscriptions hold a counter
* Whenever an update is received that counter is increased
* When a request is sent the current value of that counter is locally remembered
* The request is repeated until the current counter differs from the locally remembered counter

cgorantla:
Thanks for cleaning up subscriber code. Looks good to me.

I think we could use different executor for `subscription.update` and `subscription.request` so that they don't block each other when we have multiple subscriptions.  Though this can be optimized at later point when we have more subscribers on Twin.

We can have some tests for retry mechanism with `JUnitKafkaServer` by stopping kafka and then restarting, this test could be Kafka specific.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
```diff
@@ -230,72 +136,173 @@ protected TwinUpdate mapTwinResponseToProto(byte[] responseBytes) {
...
+         * Adds a consumer to the subscription. Incoming values will be forwarded to the passed consumer until the
+         * returned value is closed. If there is a value already available for the subscription, the consumer will be
+         * called with this value immediately.
+         *
+         * @param clazz The class of the value to consume
+         * @param consumer The consumer accepting the values
+         * @param <T> The class of the value to consume
+         * @return a Closable, stopping the consumption when closed
+         */
+        public synchronized <T> Closeable consume(final Class<T> clazz, final Consumer<T> consumer) {
```
swachter, line 201: In case that several consumers register at around the same time and no value is available yet, multiple requests are sent. Maybe the subscriber could use information that a request is already underway. A non-null `retry` future would be an indicator for that.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
```diff
@@ -230,72 +136,173 @@ protected TwinUpdate mapTwinResponseToProto(byte[] responseBytes) {
...
+                try {
+                    // Deserialize to the final class
+                    final var value = AbstractTwinSubscriber.this.objectMapper.treeToValue(json, clazz);
+
+                    // Forward to typed consumer
+                    consumer.accept(value);
+
+                } catch (final JsonProcessingException e) {
+                    LOG.error("Deserialize twin failed: {} as {}", this.key, clazz, e);
+                }
```
swachter, line 212: Admittedly, the `JsonProcessingException` is unlikely to happen. However, if it happens shouldn't it trigger a request for a new value? If `treeToValue` is called by the subscriber then it could catch the `JsonProcessingException` and either call `request` or throw an `IOException` that also would trigger a `request`.

swachter (approved):
I approve this PR but I also see room for improvement:
1. More than necessary requests may be sent. The state that a request is already under way should be considered
2. I think the synchronization logic is correct but it is also very subtle. It relies on synchronization and using a single thread executor. Luckily, the `consume` method that is called by different threads and that also changes the synchronized state does not break invariants.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/LocalTwinSubscriberImpl.java
```diff
@@ -31,8 +31,13 @@
 
 import java.io.IOException;
 
+import org.opennms.distributed.core.api.Identity;
+
 public class LocalTwinSubscriberImpl extends AbstractTwinSubscriber implements LocalTwinSubscriber {
```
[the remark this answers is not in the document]
fooker: The forwarding call to `super.accept(...)` is required to adapt from `LocalTwinSubscriber` to `AbstractTwinSubscriber`.
This has room for further improvement by not going through serialisation/deserialisation for the local impl.

The override for `close` has been removed.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
```diff
@@ -30,177 +30,83 @@
...
-                LOG.error("Exception while sending response to consumer", e);
-            }
-        } else {
-            TwinRequest twinRequest = new TwinRequest(key, location);
-            sendRpcRequest(twinRequest);
-        }
-
-        LOG.info("Subscribed to object updates with key {}", key);
-        return session;
+    public <T> Closeable subscribe(final String key, final Class<T> clazz, final Consumer<T> consumer) {
```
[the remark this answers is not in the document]
fooker: done

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
```diff
@@ -230,72 +136,173 @@ protected TwinUpdate mapTwinResponseToProto(byte[] responseBytes) {
...
         private final String key;
-        private final Consumer<T> consumer;
-        private final Class<T> clazz;
 
-        public SessionImpl(String key, Class<T> clazz, Consumer<T> consumer) {
-            this.key = key;
-            this.clazz = clazz;
-            this.consumer = consumer;
-        }
+        private final Set<Consumer<JsonNode>> consumers = Sets.newConcurrentHashSet();
```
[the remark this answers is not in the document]
fooker: Because subscribers can have different types, as long as they match the json data.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
fooker: I would argue that en error in desertification is the subscribers fault. Even more, the consumer used to subscribe can throw an exception. Therefore I altered the exception handler to be more broad.

Forwarding the exception would mean to cancel all other consumers. Therefor logging is the only thing we can do there.

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinSubscriber.java
fooker: Moved to retry-cancel logic to `accept` to prevent loop if nothing changed and added proposed check.

fooker:
> Thanks for cleaning up subscriber code. Looks good to me.
> 
> I think we could use different executor for `subscription.update` and `subscription.request` so that they don't block each other when we have multiple subscriptions. Though this can be optimized at later point when we have more subscribers on Twin.
> 
> We can have some tests for retry mechanism with `JUnitKafkaServer` by stopping kafka and then restarting, this test could be Kafka specific.

Test added

swachter (approved):

PR #3845, opened 2021-11-11
NMS-13736: fix xargs interaction with chown and ARG_MAX

RangerRick:
@christianpape ran into a weird issue with "argument list too long" when running `chown` on the opennms share directory.
It turns out this comes from a strange interaction between argument limits when running `sudo` inside `xargs`, instead of the other way around.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13736

### DIFF
--- a/opennms-base-assembly/src/main/filtered/bin/fix-permissions
+++ b/opennms-base-assembly/src/main/filtered/bin/fix-permissions
@@ -70,7 +70,10 @@ do_sudo() {
   if [ -x "$SUDO" ]; then
     "$SUDO" "$@"
   else
-    # if no sudo, fall back to running and hope for the best
+    # if no sudo, fall back to running as the current user
+    if [ "$(id -u -n)" != "root" ]; then
+      echo "WARNING: sudo was not located, but this script is not being run as root. This will probably fail."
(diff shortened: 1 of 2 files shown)


opennms-base-assembly/src/main/filtered/bin/fix-permissions
```diff
@@ -66,6 +66,15 @@ fi
 
 SUDO="$(command -v sudo 2>/dev/null || which sudo 2>/dev/null || :)"
 
+do_sudo() {
+  if [ -x "$SUDO" ]; then
+    "$SUDO" "$@"
+  else
+    # if no sudo, fall back to running and hope for the best
+    "$@"
```
indigo423, line 74: Maybe we can give the user a warning that we run as $user here and we are in a fallback mode and expect having sudo available?

indigo423 (approved):
Besides the trivial non-blocking comment :) 👍

cgorantla (dismissed):

indigo423 (approved):
LGTM! 👍

christianpape (approved):
Looks good!

PR #3846, opened 2021-11-11
NMS-13734: don't try to force uid/gid, just make normal system accounts

RangerRick:
Simplify user-account creation.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13734

### DIFF
--- a/tools/packages/opennms/opennms.spec
+++ b/tools/packages/opennms/opennms.spec
@@ -880,18 +880,10 @@ if [ -e "${ROOT_INST}/etc/users.xml" ]; then
 fi
 
 if ! getent group opennms >/dev/null 2>&1; then \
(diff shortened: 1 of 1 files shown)


indigo423 (approved):
Yes make sense to me, the only place where we deal with UID/GID is when we create the Docker container image. To be compliant running in OpenShift we have picked a high user id 10001 and we bypass any start scripts or update scripts in docker. Depending on which persistence strategy is used, the user needs to take care of himself the persisted data is owned by the right uid/gid.

RangerRick:
We need to do some work to fix OpenShift support anyway, but I haven't been able to set aside enough time to look into it yet.  I'll take that as "good enough" for this pass then.  :D

cgorantla (approved):

PR #3848, opened 2021-11-12
NMS-13742: Fix authorization entry for Twin

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13742


indigo423 (approved):
I've tested the change in H29 and the problem is solved with this patch. 👍

PR #3852, opened 2021-11-12
NMS-13733: Prefix Kafka twin topics with OpenNMS Instance Id

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13733

### DIFF
--- a/core/ipc/twin/kafka/common/src/main/java/org/opennms/core/ipc/twin/kafka/common/Topic.java
+++ b/core/ipc/twin/kafka/common/src/main/java/org/opennms/core/ipc/twin/kafka/common/Topic.java
@@ -28,6 +28,8 @@
 
 package org.opennms.core.ipc.twin.kafka.common;
 
+import org.opennms.core.utils.SystemInfoUtils;
+
 import java.util.Objects;
 
 public class Topic {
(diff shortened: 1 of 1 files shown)


RangerRick:
Traditionally, minions on older minor versions could still communicate with upstream, but this would break that, wouldn't it? We should probably include a VERY BIG NOTICE that the Minions need to be updated in lock-step, which is unusual in a .0 to .1 release.

cgorantla:
> Traditionally, minions on older minor versions could still communicate with upstream, but this would break that, wouldn't it? We should probably include a VERY BIG NOTICE that the Minions need to be updated in lock-step, which is unusual in a .0 to .1 release.

may be we need breaking changes on .1 version

fooker (approved):
Code-wise this is fine. But the compatibility problem needs a solution. Is there some hard requirement for this or is this an improvement that can go into next major release, too?

cgorantla:
> Code-wise this is fine. But the compatibility problem needs a solution. Is there some hard requirement for this or is this an improvement that can go into next major release, too?

This is broken when same Kafka server is used with multiple OpenNMS instances ( locations should also be same across those instances).  If we need compatible solution that works on `.0` and `.1` release, we can definitely fix this in code for now ( consume from both `instance.twin.sink` and `twin.sink`} and push to both `instance.twin.sink` and `twin.sink` topics. No harm in doing that for `.1` release and revert it in `release-30.x`

agalue:
> On Nov 15, 2021, at 4:48 PM, Dustin Frisch ***@***.***> wrote:
> Code-wise this is fine. But the compatibility problem needs a solution. Is there some hard requirement for this or is this an improvement that can go into next major release, too?
> 

We have multiple customers having a central Kafka shared across multiple independent OpenNMS instances, each with its set of Minions. Without this patch, none of those customers would upgrade to H29 without some hacks, and I don't consider this feature complete without it. As an example, I stopped testing H29 as soon as I found this problem.

RangerRick:
Seems reasonable to just fix it and say "make sure you update both sides when you upgrade" in the release notes then.

RangerRick (approved):

PR #3864, opened 2021-11-15
get rid of log message initializing MockLogger

RangerRick:
I don't think this part is going to fail, given the 17 billion times it's been run, and it sure does add junk to the logs.

### DIFF
--- a/core/test-api/lib/src/main/java/org/opennms/core/test/MockLogger.java
+++ b/core/test-api/lib/src/main/java/org/opennms/core/test/MockLogger.java
@@ -123,7 +123,7 @@ public class MockLogger extends MarkerIgnoringBase {
     // Load properties file, if found.
     // Override with system properties.
     static void init() {
-        System.err.println("Initializing MockLogger.");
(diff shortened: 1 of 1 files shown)


christianpape (approved):
Haha. This will save thousands of lines!

cgorantla (approved):

core/test-api/lib/src/main/java/org/opennms/core/test/MockLogger.java
```diff
@@ -123,7 +123,7 @@ private static boolean getBooleanProperty(String name, boolean defaultValue) {
     // Load properties file, if found.
     // Override with system properties.
     static void init() {
-        System.err.println("Initializing MockLogger.");
+        // System.err.println("Initializing MockLogger.");
```
indigo423, line 126: This is clearly a stderr :)

christianpape: CircleCI can switch off one datacenter.

PR #3866, opened 2021-11-16
Fix Kafka producer client configuration

indigo423:
Trivial change that clarifies how to configure the Kafka producer client which was reported in our discussion chat.

### DIFF
--- a/docs/modules/operation/pages/kafka-producer/enable-kafka.adoc
+++ b/docs/modules/operation/pages/kafka-producer/enable-kafka.adoc
@@ -20,7 +20,7 @@ config:property-set bootstrap.servers kafka-broker-ip:9092 <1>
 config:update
 ----
 <1> Set the Kafka broker IP by replacing the kafka-broker-ip and port you want to connect to.
-    You can add multiple hosts in a comma separated list, e.g. `kafka-broker-1:9092,kafka-broker-2:9092`
+    Add multiple hosts in a comma-separated list, for example, `kafka-broker-1:9092,kafka-broker-2:9092`
 
 Next, install the `opennms-kafka-producer` feature from that same shell using:
 
--- a/docs/modules/operation/pages/kafka-producer/enable-kafka.adoc
+++ b/docs/modules/operation/pages/kafka-producer/enable-kafka.adoc
@@ -6,20 +6,26 @@ The Kafka Producer is disabled by default and can be enabled as follows.
 First, log in to the Karaf shell of your {page-component-title} instance and configure the Kafka client settings to point to your Kafka broker.
 See link:https://kafka.apache.org/10/documentation.html#producerconfigs[Producer Configs] for a complete list of available options.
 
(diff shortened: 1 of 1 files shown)


docs/modules/operation/pages/kafka-producer/enable-kafka.adoc
```diff
@@ -6,20 +6,27 @@ The Kafka Producer is disabled by default and can be enabled as follows.
...
 ----
 
+.Configure the Kafka producer client
+[source, karaf]
+----
+config:edit org.opennms.features.kafka.producer.client
+config:property-set bootstrap.servers kafka-broker-ip:9092 <1>
+config:update
+----
+<1> Set the Kafka broker IP by replacing the kafka-broker-ip and port you want to connect.
```
Bonrob2, line 22:
```suggestion
<1> Set the Kafka broker IP by replacing the kafka-broker-ip and port you want to connect to.
```

docs/modules/operation/pages/kafka-producer/enable-kafka.adoc
```diff
@@ -6,20 +6,27 @@ The Kafka Producer is disabled by default and can be enabled as follows.
...
+    You can add multiple hosts in a comma separated list, e.g. `kafka-broker-1:9092,kafka-broker-2:9092`
```
Bonrob2, line 23:
```suggestion
    Add multiple hosts in a comma-separated list, for example, `kafka-broker-1:9092,kafka-broker-2:9092`
```

Bonrob2 (changes requested):
Minor changes.

docs/modules/operation/pages/kafka-producer/enable-kafka.adoc
mmahacek: Is a preposition word we want to end a sentence with?

Bonrob2: It's not usually good practice, but in this situation it's okay. Changing it to "with which you want to connect" would be too awkward. Most style guides allow it now under these types of circumstances. 
We could say "... and connection port." if that would still make sense to users.

mmahacek: Cool cool cool

indigo423:
@Bonrob2 thank you for the improvements!

mmahacek (approved):
While the highlighter.js doesn't have syntax rules for applying color coding to `karaf` source blocks, they do make it easier for the user to know where they should be running the commands, and removing the karaf prompt makes it easier to copy/paste.

LGTM!

indigo423:
@Bonrob2 Just FYI, cause you requested changes for the PR, you're the only one with the power to unblock the merge if everything is as you expect :)

Screenshot 2021-11-17 at 00 19 45 1

Bonrob2 (approved):
LGTM!

PR #3872, opened 2021-11-16
PE-114: Return empty array in case provisiond config has no requisition-def

freddy61025:
https://issues.opennms.org/browse/PE-114

Besides handling requistion-def, it will fill up all missing config properties with default value / null.


features/config/service/impl/src/main/java/org/opennms/features/config/service/impl/ConfigurationManagerServiceImpl.java
```diff
@@ -165,14 +171,27 @@ public void updateConfiguration(String configName, String configId, JsonAsString
...
+            if (schemaName == null) { // assume if top element name is null, the top schema name is configName
+                schemaName = configName;
+            }
+            OpenAPIConfigHelper.fillWithDefaultValue(def.get().getSchema(), schemaName, configObj.get());
+        }
+        return configObj;
     }
 
     @Override
     public String getJSONStrConfiguration(String configName, String configId) throws IOException, IllegalArgumentException {
```
patrick-schweizer, line 191: I wonder if we shouldn't use here Optional<String> getJSONStrConfiguration instead of throwing a FileNotFoundException? This would be consistent to getJSONConfiguration()

features/config/test/src/test/java/org/opennms/features/config/service/util/OpenAPIConfigHelperTest.java
```diff
@@ -0,0 +1,74 @@
...
+import org.opennms.features.config.dao.api.ConfigItem;
+import org.opennms.features.config.dao.impl.util.ConfigSwaggerConverter;
+import org.opennms.features.config.dao.impl.util.OpenAPIBuilder;
+import org.skyscreamer.jsonassert.JSONAssert;
+
+import java.math.BigDecimal;
+import java.util.List;
+
+public class OpenAPIConfigHelperTest {
+    String prefix = "/rest/cm";
```
patrick-schweizer, line 47: nit: I saw we use the same prefix in different places. Maybe it makes sense to centralize that so we can change it later easier since it's just in one place?

patrick-schweizer (commented):
2 points to think about. No blockers

patrick-schweizer (dismissed):
looks good to me now

patrick-schweizer (approved):

PR #3874, opened 2021-11-17
PE-148: Migrate to new CM jmx-config.xml

dherdt:
### External References

* JIRA (Issue Tracker): https://issues.opennms.org/browse/PE-148


opennms-config-model/src/main/resources/xsds/jmx-config.xsd
```diff
@@ -0,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
patrick-schweizer, line 1: as discussed: let's restrict just to the same types as we have in the Jaxb class.

patrick-schweizer (changes requested):
Looks good! First service done :-)
As discussed please make the XSD less restrictive before we merge so we don't hurt existing customers. Thanks!

patrick-schweizer (approved):
Looks now good to me!

PR #3884, opened 2021-11-18
NMS-13706: Flow Thresholding basics

fooker:
Flow thresholding groundwork.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13706

### DIFF
--- a/features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/DocumentEnricher.java
+++ b/features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/DocumentEnricher.java
@@ -101,20 +101,10 @@ public class DocumentEnricher {
 
     private final long clockSkewCorrectionThreshold;
 
-    private final ThresholdingService thresholdingService;
-
-    private final Map<ApplicationKey, AtomicLong> applicationAccumulator = Maps.newHashMap();
-
-    private final static RrdRepository FLOW_APP_RRD_REPO = new RrdRepository();
-
-    private final CollectionAgentFactory collectionAgentFactory;
-
     public DocumentEnricher(MetricRegistry metricRegistry, NodeDao nodeDao, InterfaceToNodeCache interfaceToNodeCache,
                             SessionUtils sessionUtils, ClassificationEngine classificationEngine,
                             CacheConfig cacheConfig,
-                            final long clockSkewCorrectionThreshold,
-                            final ThresholdingService thresholdingService,
-                            final CollectionAgentFactory collectionAgentFactory) {
+                            final long clockSkewCorrectionThreshold) {
         this.nodeDao = Objects.requireNonNull(nodeDao);
         this.interfaceToNodeCache = Objects.requireNonNull(interfaceToNodeCache);
         this.sessionUtils = Objects.requireNonNull(sessionUtils);
@@ -142,9 +132,6 @@ public class DocumentEnricher {
         this.nodeLoadTimer = metricRegistry.timer("nodeLoadTime");
 
         this.clockSkewCorrectionThreshold = clockSkewCorrectionThreshold;
-
-        this.thresholdingService = Objects.requireNonNull(thresholdingService);
-        this.collectionAgentFactory = Objects.requireNonNull(collectionAgentFactory);
     }
 
     public List<FlowDocument> enrich(final Collection<Flow> flows, final FlowSource source) {
@@ -206,45 +193,6 @@ public class DocumentEnricher {
                     document.setFirstSwitched(document.getFirstSwitched() - skew);
                     document.setDeltaSwitched(document.getDeltaSwitched() - skew);
                     document.setLastSwitched(document.getLastSwitched() - skew);
-                }
-            }
-
-            if (document.getNodeExporter() != null &&
-                !Strings.isNullOrEmpty(document.getApplication())) {
-                final var key = new ApplicationKey(document.getNodeExporter().getNodeId(),
-                                                   document.getDirection() == Direction.INGRESS
-                                                   ? document.getInputSnmp()
-                                                   : document.getOutputSnmp(),
-                                                   document.getApplication());
-
-                final var accumulator = applicationAccumulator.computeIfAbsent(key, k -> new AtomicLong(0L));
-
-                final var counter = accumulator.addAndGet(document.getBytes());
-
-                try {
-                    // TODO: Cache this session along with the counter
-                    final var thresholdingSession = this.thresholdingService.createSession(key.nodeId,
-                                                           document.getDirection() == Direction.INGRESS
-                                                                       ? document.getSrcAddr()
-                                                                       : document.getDstAddr(),
-                                                           "flow-application",
-                                                           FLOW_APP_RRD_REPO,
-                                                           new ServiceParameters(Collections.emptyMap()));
-
-                    final var collectionAgent = this.collectionAgentFactory.createCollectionAgent(
-                            Integer.toString(key.nodeId),
-                            InetAddressUtils.addr(source.getSourceAddress()));
-
-                    final var nodeResource = new NodeLevelResource(key.nodeId);
-                    final var appResource = new DeferredGenericTypeResource(nodeResource, "flow-application", key.application);
-
-                    final var collectionSetBuilder = new CollectionSetBuilder(collectionAgent);
-                    collectionSetBuilder.withCounter(appResource, "flow-application", "bytes", counter);
-
-                    thresholdingSession.accept(collectionSetBuilder.build());
-
-                } catch (final ThresholdInitializationException e) {
-                    LOG.warn("Failed to create thresholding session", e);
                 }
             }
 
--- a/features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/DocumentEnricher.java
+++ b/features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/DocumentEnricher.java
@@ -29,14 +29,15 @@
 package org.opennms.netmgt.flows.elastic;
 
 import java.net.InetAddress;
-import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Collectors;
 
 import org.opennms.core.cache.Cache;
@@ -45,6 +46,15 @@ import org.opennms.core.cache.CacheConfig;
 import org.opennms.core.cache.CacheConfigBuilder;
 import org.opennms.core.rpc.utils.mate.ContextKey;
 import org.opennms.core.utils.InetAddressUtils;
+import org.opennms.netmgt.collection.api.CollectionAgentFactory;
+import org.opennms.netmgt.collection.api.ServiceParameters;
+import org.opennms.netmgt.collection.dto.CollectionSetDTO;
+import org.opennms.netmgt.collection.support.builder.CollectionSetBuilder;
+import org.opennms.netmgt.collection.support.builder.DeferredGenericTypeResource;
+import org.opennms.netmgt.collection.support.builder.GenericTypeResource;
+import org.opennms.netmgt.collection.support.builder.InterfaceLevelResource;
+import org.opennms.netmgt.collection.support.builder.NodeLevelResource;
+import org.opennms.netmgt.collection.support.builder.PerspectiveResponseTimeResource;
 import org.opennms.netmgt.dao.api.InterfaceToNodeCache;
 import org.opennms.netmgt.dao.api.NodeDao;
 import org.opennms.netmgt.dao.api.SessionUtils;
@@ -55,6 +65,10 @@ import org.opennms.netmgt.flows.classification.ClassificationRequest;
 import org.opennms.netmgt.flows.classification.persistence.api.Protocols;
 import org.opennms.netmgt.model.OnmsCategory;
 import org.opennms.netmgt.model.OnmsNode;
+import org.opennms.netmgt.poller.PollStatus;
+import org.opennms.netmgt.rrd.RrdRepository;
+import org.opennms.netmgt.threshd.api.ThresholdInitializationException;
+import org.opennms.netmgt.threshd.api.ThresholdingService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -62,6 +76,7 @@ import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
 import com.google.common.base.Strings;
 import com.google.common.cache.CacheLoader;
+import com.google.common.collect.Maps;
 
 public class DocumentEnricher {
     private static final Logger LOG = LoggerFactory.getLogger(DocumentEnricher.class);
@@ -86,10 +101,20 @@ public class DocumentEnricher {
 
     private final long clockSkewCorrectionThreshold;
 
+    private final ThresholdingService thresholdingService;
+
+    private final Map<ApplicationKey, AtomicLong> applicationAccumulator = Maps.newHashMap();
+
+    private final static RrdRepository FLOW_APP_RRD_REPO = new RrdRepository();
+
+    private final CollectionAgentFactory collectionAgentFactory;
+
     public DocumentEnricher(MetricRegistry metricRegistry, NodeDao nodeDao, InterfaceToNodeCache interfaceToNodeCache,
                             SessionUtils sessionUtils, ClassificationEngine classificationEngine,
                             CacheConfig cacheConfig,
-                            final long clockSkewCorrectionThreshold) {
+                            final long clockSkewCorrectionThreshold,
+                            final ThresholdingService thresholdingService,
+                            final CollectionAgentFactory collectionAgentFactory) {
         this.nodeDao = Objects.requireNonNull(nodeDao);
         this.interfaceToNodeCache = Objects.requireNonNull(interfaceToNodeCache);
         this.sessionUtils = Objects.requireNonNull(sessionUtils);
@@ -117,6 +142,9 @@ public class DocumentEnricher {
         this.nodeLoadTimer = metricRegistry.timer("nodeLoadTime");
 
         this.clockSkewCorrectionThreshold = clockSkewCorrectionThreshold;
+
+        this.thresholdingService = Objects.requireNonNull(thresholdingService);
+        this.collectionAgentFactory = Objects.requireNonNull(collectionAgentFactory);
     }
 
     public List<FlowDocument> enrich(final Collection<Flow> flows, final FlowSource source) {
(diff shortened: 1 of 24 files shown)


features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
```diff
@@ -0,0 +1,135 @@
...
+                // Apply thresholding
+                final var thresholdingSession = this.thresholdingSessions.get(new NodeInterfaceKey(nodeId, source.getSourceAddress()));
+
+                final var collectionAgent = this.collectionAgentFactory.createCollectionAgent(
+                        Integer.toString(nodeId),
+                        InetAddressUtils.addr(source.getSourceAddress()));
+
+                final var nodeResource = new NodeLevelResource(nodeId);
+                final var ifaceResource = new InterfaceLevelResource(nodeResource, Integer.toString(key.iface));
+                final var appResource = new DeferredGenericTypeResource(nodeResource, "flow_app", document.getApplication());
```
christianpape, line 125: Make "flow_app" a constant.

features/flows/itests/src/test/java/org/opennms/netmgt/flows/elastic/ThresholdingIT.java
```diff
@@ -0,0 +1,315 @@
...
+        }
+        return flows;
+    }
+
+    private FlowRepository createFlowRepository(final JestClient jestClient) throws InterruptedException {
+        final var config = new ThresholdingConfig();
+        final var group = new Group();
+        group.setName("test");
+        group.setRrdRepository("/opt/opennms/share/rrd/snmp/");
+        final var threshold = new Threshold();
```
christianpape, line 184: Use a config file in src/test/resources instead?

opennms-base-assembly/src/main/filtered/etc/resource-types.d/flows-resource.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0"?>
+<resource-types>
+   <resourceType name="flow_app" label="Flow Application" resourceLabel="Flows: Application ${application}">
```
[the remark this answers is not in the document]
fooker: changed

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
fooker: changed

features/flows/itests/src/test/java/org/opennms/netmgt/flows/elastic/ThresholdingIT.java
```diff
@@ -0,0 +1,315 @@
...
+    private OverrideableThreshdDao threshdDao;
+
+    @Autowired
+    private MockEventIpcManager mockEventIpcManager;
+
+    @Autowired
+    private ApplicationContext applicationContext;
+
+    @Before
+    public void before() {
```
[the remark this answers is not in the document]
fooker: not possible

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
```diff
@@ -0,0 +1,135 @@
...
+
+                final var collectionSetBuilder = new CollectionSetBuilder(collectionAgent)
+                        .withCounter(appResource, "flowappxxxx", "bytes", counter);
```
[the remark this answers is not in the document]
fooker: This is currently undecided and needs more investigation while implementing the persistence.

features/flows/itests/src/test/java/org/opennms/netmgt/flows/elastic/ThresholdingIT.java
fooker: done

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
```diff
@@ -0,0 +1,135 @@
...
+                                                   ? document.getInputSnmp()
+                                                   : document.getOutputSnmp(),
+                                                   document.getApplication());
+
+                // Update the counter
+                final var counter = this.applicationAccumulator.computeIfAbsent(key, k -> new AtomicLong(0))
+                                                               .addAndGet(document.getBytes());
+
+                // Apply thresholding
+                final var thresholdingSession = this.thresholdingSessions.get(new NodeInterfaceKey(nodeId, source.getSourceAddress()));
```
swachter, line 118: Can you document why the `NodeInterfaceKey` uses the `sourceAddress` and not the interface index? Can the same `nodeId` go together with different `sourceAddress` values? Maybe this would be a good chance to document what a `FlowSource` is.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/ElasticFlowRepository.java
```diff
@@ -139,6 +142,8 @@
 
     private final EnrichedFlowForwarder enrichedFlowForwarder;
 
+    private final FlowThresholding thresholding;
```
swachter, line 145: Flow thresholding should be an independent feature that is not part of ElasticSearch related functionality. I can imagine having flow threshold without ES persistence.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/ApplicationKey.java
```diff
@@ -0,0 +1,66 @@
...
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.netmgt.flows.elastic.thresholding;
+
+import java.util.Objects;
+
+public class ApplicationKey {
+    public final int nodeId;
+    public final int iface;
```
swachter, line 35: Maybe we can combine `nodeId` and `iface` into a single long. That could make it more performant.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
```diff
@@ -0,0 +1,135 @@
...
+                        InetAddressUtils.addr(source.getSourceAddress()));
+
+                final var nodeResource = new NodeLevelResource(nodeId);
+                final var appResource = new DeferredGenericTypeResource(nodeResource, RESOURCE_TYPE_NAME, document.getApplication());
+
+                final var collectionSetBuilder = new CollectionSetBuilder(collectionAgent)
+                        .withCounter(appResource, "flowappxxxx", "bytes", counter);
+                // TODO fooker: Set sequence number from flow to aid distributed thresholding
+
+                thresholdingSession.accept(collectionSetBuilder.build());
```
swachter, line 131: Creating a `collectionAgent`, `nodeResource`, and `appResource` each time anew seems expensive to me. Could these things be cached too? The `threasholdingSession` map could contain ready to call `consumers` that setup all these things when a new entry is added to the map.

swachter (changes requested):
It is great that so much functionality could be reused. Major concern is that flow thresholding should be independent of ElasticSearch related functionalities.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/DocumentEnricher.java
```diff
@@ -45,6 +46,15 @@
...
 import org.opennms.core.utils.InetAddressUtils;
+import org.opennms.netmgt.collection.api.CollectionAgentFactory;
+import org.opennms.netmgt.collection.api.ServiceParameters;
+import org.opennms.netmgt.collection.dto.CollectionSetDTO;
+import org.opennms.netmgt.collection.support.builder.CollectionSetBuilder;
+import org.opennms.netmgt.collection.support.builder.DeferredGenericTypeResource;
+import org.opennms.netmgt.collection.support.builder.GenericTypeResource;
+import org.opennms.netmgt.collection.support.builder.InterfaceLevelResource;
+import org.opennms.netmgt.collection.support.builder.NodeLevelResource;
+import org.opennms.netmgt.collection.support.builder.PerspectiveResponseTimeResource;
```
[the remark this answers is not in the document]
fooker: done

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/DocumentEnricher.java
```diff
@@ -29,14 +29,15 @@
...
-import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.atomic.AtomicLong;
```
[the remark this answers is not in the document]
fooker: done

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/DocumentEnricher.java
```diff
@@ -55,13 +65,18 @@
 import org.opennms.netmgt.flows.classification.persistence.api.Protocols;
 import org.opennms.netmgt.model.OnmsCategory;
 import org.opennms.netmgt.model.OnmsNode;
+import org.opennms.netmgt.poller.PollStatus;
+import org.opennms.netmgt.rrd.RrdRepository;
+import org.opennms.netmgt.threshd.api.ThresholdInitializationException;
+import org.opennms.netmgt.threshd.api.ThresholdingService;
```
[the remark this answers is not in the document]
fooker: done

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/DocumentEnricher.java
```diff
@@ -55,13 +65,18 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
 import com.google.common.base.Strings;
 import com.google.common.cache.CacheLoader;
+import com.google.common.collect.Maps;
```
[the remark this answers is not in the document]
fooker: done

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/ElasticFlowRepository.java
fooker: It should. But it depends on the enrichment process, which is only available inside the `elastic` module. Refactoring this would mean to refactor the whole flow processing pipeline and introduce a bunch of shared modules and transport data types. Therefore I will not do this right now. But I will consider this for later.

fooker: Related issue: https://issues.opennms.org/browse/NMS-13770

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
fooker: > Can you document why the `NodeInterfaceKey` uses the `sourceAddress` and not the interface index?

Simply, because that's what the `ThresholdingSession` and the `CollectionAgent` need for creation.

> Can the same `nodeId` go together with different `sourceAddress` values?

Indeed. We use the source address (and location) to reverse-lookup the node which has this interface assigned.
This takes into account that the same interface may be assigned to multiple nodes (even in the same location) and that proxies may be used to aggregate multiple exporters into a single source (hence the support for `sourceId` lookup).

This is far from ideal. In case of `sourceId` lookups, we maybe choose an interface, that maybe does not exist on the interface. I've created https://issues.opennms.org/browse/NMS-13771 to track this later.

> Maybe this would be a good chance to document what a `FlowSource` is.

Not sure if the `FlowSource` needs documentation. It's a source of flows. The address (in a location) that sends out flows.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/ApplicationKey.java
fooker: Simplicity is preferred over (dubious) performance improvements. Especially at this stage of the project and without benchmarking.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
fooker: Refactored to cache the `CollectionAgent` along with the `ThresholdingSession`. The resources are to cheap to create to bother right now. They are even subject of change in further development.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
```diff
@@ -0,0 +1,149 @@
...
+
+                // Apply thresholding
+                final var session = this.sessions.get(new NodeInterfaceKey(nodeId, source.getSourceAddress()));
+
+                final var nodeResource = new NodeLevelResource(nodeId);
+                final var appResource = new DeferredGenericTypeResource(nodeResource, RESOURCE_TYPE_NAME, document.getApplication());
+
+                final var collectionSetBuilder = new CollectionSetBuilder(session.collectionAgent)
+                        .withCounter(appResource, document.getApplication(), "bytes", counter);
+                // TODO fooker: Set sequence number from flow to aid distributed thresholding
```
j-white, line 132: :D

features/flows/itests/src/test/java/org/opennms/netmgt/flows/elastic/ThresholdingIT.java
```diff
@@ -0,0 +1,275 @@
...
+            this.transactionTemplate.execute((tx) -> {
+                try {
+                    flowRepository.persist(createMockedFlows(1000), source);
+                } catch (FlowException e) {
+                    throw new RuntimeException(e);
+                }
+                return null;
+            });
+
+            await().atMost(60, TimeUnit.SECONDS).until(eventAnticipator::getAnticipatedEvents, hasSize(0));
```
j-white, line 270: 👍

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
```diff
@@ -0,0 +1,149 @@
...
+                                                               .addAndGet(document.getBytes());
+
+                // Apply thresholding
+                final var session = this.sessions.get(new NodeInterfaceKey(nodeId, source.getSourceAddress()));
+
+                final var nodeResource = new NodeLevelResource(nodeId);
+                final var appResource = new DeferredGenericTypeResource(nodeResource, RESOURCE_TYPE_NAME, document.getApplication());
+
+                final var collectionSetBuilder = new CollectionSetBuilder(session.collectionAgent)
+                        .withCounter(appResource, document.getApplication(), "bytes", counter);
```
j-white, line 131: Would it be useful to track `bytesIn` vs `bytesOut` vs `bytesTotal`?

features/flows/elastic/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -206,6 +206,25 @@
...
     <service interface="org.opennms.netmgt.flows.api.FlowQueryService" ref="smartQueryService" />
 
+    <reference id="thresholdingService" interface="org.opennms.netmgt.threshd.api.ThresholdingService"/>
+    <reference id="collectionAgentFactory" interface="org.opennms.netmgt.collection.api.CollectionAgentFactory"/>
+
+    <!-- Thresholding -->
+    <bean id="thresholdingCacheConfig" class="org.opennms.core.cache.CacheConfig">
+        <argument value="flow-thresholding-sessions"/>
+        <property name="metricRegistry" ref="metricRegistry" />
+        <property name="enabled" value="${flowThresholdCache.enabled}" />
```
j-white, line 216: Define these with defaults above?

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
fooker: We're currently playing around with this. Seems more useful to have a distinct metric per application. This would allow to define thresholds per application.

features/flows/elastic/src/main/resources/OSGI-INF/blueprint/blueprint.xml
fooker: done

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
swachter: Can the same flow source send out flows of different nodes? Your remark on "proxies" hints for that. In that case the `sourceAddress` seems of dubious use to identify a metric for thresholding.

The relation between `nodeId`, `sourceAddress`, `interfaces`, and `itfIndex` is still not clear to me. If `sourceAddress` and `location` are used to lookup a `nodeId` then having both, `nodeId` and `sourceAddress` as part of the `NodeInterfaceKey` seems unnecessary. Shouldn't the `location` be considered? I would like to have these relationships be documented somewhere.

I was surprised to hear that "the same interface may be assigned to multiple nodes". My current understanding is that "interface" and IP address (numerical) are often used as synonyms. Is this the case here, too? In my understanding there is a 1:1 relationship between (nodeId, itfIndex) and numeric IP addresses.

Can you point me to some documentation where the involved terms are documented? If there is no such place then maybe we should create a documentation issue.

christianpape (approved):
Looks good to me.

swachter (approved):
I still not fully understand the underlying model of "node", "flow source", "location", "interface", and "interface index".  I raised some questions / remarks in a conversation. However, this is not necessarily related to the PR at hand.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
fooker: > Can the same flow source send out flows of different nodes? Your remark on "proxies" hints for that. In that case the `sourceAddress` seems of dubious use to identify a metric for thresholding.

Yes it can. Therefore we use the `nodeId` as part of the key. The `nodeId` is found using the `sourceId` and meta-data associations.

> The relation between `nodeId`, `sourceAddress`, `interfaces`, and `itfIndex` is still not clear to me. If `sourceAddress` and `location` are used to lookup a `nodeId` then having both, `nodeId` and `sourceAddress` as part of the `NodeInterfaceKey` seems unnecessary. Shouldn't the `location` be considered? I would like to have these relationships be documented somewhere.

There are two types of interfaces: IP Interfaces (identified by its IP) and SNMP Interfaces (identified by the SNMP interface index). A Node can have multiple IP Interfaces and multiple SNMP Interfaces. Two IP Interfaces with the same IP address can exists if they are assigned to different nodes, even if the nodes are in the same location.

As mentioned above, the proxy case is not implemented correctly right now. The perfect solution would be to used the database ID of the IP Interface. But lookup of this is expensive and requires larger restructuring.

The key was chosen this way to make allow creation of the cached value from the key only. Other solution would increase complexity in caching. I don't see a big problem here. The key may not be as small as it could be - but there are no duplicates possible.

This all is ages old core concept of OpenNMS. These internals are not documented AFAIK. But creating documentation is out-of-scope for this task.
 
> I was surprised to hear that "the same interface may be assigned to multiple nodes". My current understanding is that "interface" and IP address (numerical) are often used as synonyms. Is this the case here, too? In my understanding there is a 1:1 relationship between (nodeId, itfIndex) and numeric IP addresses.

There can be duplicates. Even in the same location.
In addition, "Interface" can mean two things: The IP interface on which we exchange things with and SNMP interfaces on witch we collect metrics.

> Can you point me to some documentation where the involved terms are documented? If there is no such place then maybe we should create a documentation issue.

Sorry, I can not. I don't think, this is part of the task. If you fell this needs documentation - go ahead and create such an issue.

PR #3892, opened 2021-11-19
NMS-13589: port geomap code to ui-foundation

jadyhou:
This work is to port the geo-map code from an independent project to the "UI Preview" (originally branch "ui-foundation").   
A screenshot of the layout is below.
Screen Shot 2021-11-19 at 8 49 48 AM

### External References

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-13589


ui/src/components/Map/MapNodesGrid.vue
```diff
@@ -0,0 +1,215 @@
...
+  });
+  gridColumnApi.autoSizeColumns(allColumnIds, skipHeader);
+}
+
+watch(
+  () => interestedNodesID.value,
+  () => {
+    gridApi.setRowData(
+      getGridRowDataFromInterestedNodes()
+    );
```
mikewrosey, line 94: Getting an error here when page first loads `gridApi.setRowData is not a function`. On refreshing it works though, so looks like it's just not ready on the initial load. Might only need to wrap it in a conditional to see if it's available.

ui/src/components/Map/LeafletMap.vue
```diff
@@ -0,0 +1,130 @@
...
+import {
+  LMap,
+  LTileLayer,
+  LMarker,
+  LPopup,
+  LControlLayers,
+  LPolyline,
+} from "@vue-leaflet/vue-leaflet";
+import MarkerCluster from "./MarkerCluster.vue";
+import { Vue } from "vue-class-component";
```
mikewrosey, line 57: Looks like this import can be removed?

ui/src/components/Map/LeafletMap.vue
```diff
@@ -0,0 +1,130 @@
...
+import { useStore } from "vuex";
+import { Coordinates } from "@/types";
+let leafletReady = ref(false);
+let leafletObject = ref("");
+let visible = ref(false);
```
mikewrosey, line 62: `visible` here also looks unused and can be removed.

ui/src/components/Map/LeafletMap.vue
```diff
@@ -0,0 +1,130 @@
...
+  coordinate.push(node.assetRecord.latitude);
+  coordinate.push(node.assetRecord.longitude);
+  return coordinate;
+}
+let interestedNodesID = computed(() => {
+  return store.getters['mapModule/getInterestedNodesID'];
+})
+let edges = computed(() => {
+  let ids = interestedNodesID.value;
+  let interestedNodesCoordinateMap = getInterestedNodesCoordinateMap();
```
mikewrosey, line 84: For all `let` and `var` in this file and elsewhere, we can likely switch to `const` to prevent accidental re-assignment.
Only those that are are being reassigned will need `let`. Adding properties to an object or pushing new values to an array does not qualify as re-assigning so we're good to use `const` for those as well.

ui/src/components/Map/SeverityFloatingFilter.vue
```diff
@@ -0,0 +1,54 @@
...
+            },
+
+            onParentModelChanged(parentModel) {
+                  // When the filter is empty we will receive a null value here
+                  if (!parentModel) {
+                        this.currentValue = '';
+                  } else {
+                        this.currentValue = parentModel.filter;
+                  }
+            }
```
mikewrosey, line 50: Files here generally need formatting to conform to the .prettierrc file rules. This will change the indentation among other things. The rules in that file are just my opinion about what works best so let me know if you disagree with any. Whatever rules we pick though, we'll just need to make sure it's consistent everywhere. 

If you're using vs code, you can use the Prettier formatter extension to comply with the rules. I assume other editors will have similar plugins.

ui/src/components/Map/MarkerCluster.vue
```diff
@@ -0,0 +1,131 @@
+<template>
+  <div style="display: none">
+    <slot v-if="ready"></slot>
+  </div>
+</template>
+
+<script>
```
mikewrosey, line 7: Would we be able to use `<script setup lang=ts>` here to keep consistent with the new component syntax?
Some minor refactoring will need to take place after (removing the export default / setup functions, etc)

ui/src/components/Map/SeverityFloatingFilter.vue
```diff
@@ -0,0 +1,54 @@
...
+                  :style="{ width: '70px' }"
+                  v-model="currentValue"
+                  @change="onSelectionChanged()"
+            >
+                  <option v-for="option in severities" :value="option" :key="option">{{ option }}</option>
+            </select>
+      </div>
+</template>
+
+<script>
```
mikewrosey, line 14: Same with this component, we can convert to the newer <script setup lang=ts> syntax and likely reduce it by a few lines.

ui/src/components/Map/MapAlarmsGrid.vue
```diff
@@ -0,0 +1,343 @@
...
+const GStore: any = inject('GStore');
+
+let selectedAlarmIds: any = ref([]);
+
+let hasAlarmSelected: any = ref(false);
+
+function onSelectionChanged() {
+  let selectedRows = gridApi.getSelectedNodes().map((node: any) => node.data);
+  selectedAlarmIds = selectedRows.map((alarm: any) => alarm.id);
+  hasAlarmSelected.value = selectedAlarmIds.length > 0;
```
mikewrosey, line 176: `selectedAlarmIds` is a ref so should this be using `selectedAlarmIds.value` on 175 and 176?

One way to makes sure TS catches this is to avoid the use of `any` wherever possible. 
Above we have `let selectedAlarmIds: any = ref([]);` If we want to give this ref type safety we can change it like so `const selectedAlarmIds = ref<string[]>([]); `.

ui/src/components/Map/MapNodesGrid.vue
jadyhou: Thanks for finding this. I added a condition to check if ``gridApi.setRowData `` is available.

ui/src/components/Map/LeafletMap.vue
jadyhou: Yes, removed.

ui/src/components/Map/LeafletMap.vue
jadyhou: Thanks, removed.

ui/src/components/Map/LeafletMap.vue
jadyhou: Thanks, I've updated all three vue files.

ui/src/components/Map/MapAlarmsGrid.vue
jadyhou: Good point, updated.

ui/package.json
```diff
@@ -29,11 +29,20 @@
...
     "axios": "^0.21.2",
     "dayjs": "^1.10.7",
+    "@ag-grid-community/vue3": "^26.0.0",
+    "ag-grid-community": "^26.0.0",
+    "ag-grid-vue3": "26.0.0",
+    "leaflet": "^1.7.1",
+    "leaflet.markercluster": "^1.5.0",
     "lodash": "^4.17.21",
+    "patch-package": "^6.4.7",
+    "postinstall-postinstall": "^2.1.0",
```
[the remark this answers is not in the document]
jadyhou: You are correct, removed.

j-white (approved):
Thanks for all the updates @mikewrosey. Looks great!

Lets squash merge.

PR #3898, opened 2021-11-22
NMS-13772: add file editor view, logs view

mikewrosey:
* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-13772


j-white:
Edit files:
image

XML validation:
image

j-white:
I'm not sure if this is something we will want to keep when we cut Horizon 30 next year, and whether or not we want to support this an actual feature. I do believe however that it would be very helpful to have in the tree while we develop and continue to make progress on the new UI.

I'm good to merge this, with the intent of revisiting it later in the release cycle and making a final decision then. What do other folks think?

dino2gnt:
I think a hard cutover from "config files I can view and edit" to "everything is in the database and you configure it with some radio buttons and checkboxes :smirk:" is uncomfortably abrupt, and having an editor available in the UI is a good transitory step.

indigo423:
well done @mikewrosey pretty rad :)

RangerRick:
> I'm not sure if this is something we will want to keep when we cut Horizon 30 next year, and whether or not we want to support this an actual feature. I do believe however that it would be very helpful to have in the tree while we develop and continue to make progress on the new UI.

This seems like good insurance against the case that some config _isn't_ converted over yet by H30...

> I'm good to merge this, with the intent of revisiting it later in the release cycle and making a final decision then. What do other folks think?

Seems reasonable to me.

RangerRick (dismissed):

j-white (approved):

j-white (approved):

cgorantla:
This looks awesome. Great work @mikewrosey !

mikewrosey:
thanks @indigo423 and @cgorantla !

PR #3902, opened 2021-11-24
PE-148: Migrate JMX configuration using the configuration management framework

dherdt:
All test issues are fixed

* JIRA (Issue Tracker): https://issues.opennms.org/browse/PE-148


opennms-config-tester/src/main/resources/META-INF/opennms/applicationContext-configTester.xml
```diff
@@ -238,7 +238,7 @@
     <property name="configResource" value="file:${opennms.home}/etc/jdbc-datacollection-config.xml"/>
   </bean>
 
-  <bean id="jmxConfigDao" class="org.opennms.netmgt.dao.jmx.JmxConfigDaoCm"/>
+  <bean id="jmxConfigDao" class="org.opennms.netmgt.dao.jaxb.DefaultJmxConfigDao"/>
```
freddy61025, line 242: only one minor concern about the package name. It is not jaxb related, it is better to name it cm or something else.

freddy61025 (commented):
only 1 minor concern

patrick-schweizer (approved):
looks good to me. I think you need to rebase before we can merge...

PR #3906, opened 2021-11-25
NMS-13507: Speedup node link derivations

swachter:
Issue: https://issues.opennms.org/browse/NMS-13507

The current implementation used hundreds of separate SQL selects to retrieve the information necessary to derive bridge links. On a customer database the request `http://localhost:8980/opennms/api/v2/enlinkd/bridge_links/1322` took tens of seconds.

This PR prefetches some of the required information by bulk selects and uses caches for other information that are also loaded by  bulk selects. In summary, the improvements sped up rest request to around 125 milliseconds.

### DIFF
--- a/opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/SnmpInterfaceDaoHibernate.java
+++ b/opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/SnmpInterfaceDaoHibernate.java
@@ -47,7 +47,7 @@ public class SnmpInterfaceDaoHibernate extends AbstractDaoHibernate<OnmsSnmpInte
     public OnmsSnmpInterface findByNodeIdAndIfIndex(Integer nodeId, Integer ifIndex) {
         Assert.notNull(nodeId, "nodeId may not be null");
         Assert.notNull(ifIndex, "ifIndex may not be null");
-        return findUnique("select distinct snmpIf from OnmsSnmpInterface as snmpIf where snmpIf.node.id = ? and snmpIf.ifIndex = ?", 
+        return findUnique("select snmpIf from OnmsSnmpInterface as snmpIf where snmpIf.node.id = ? and snmpIf.ifIndex = ?",
                           nodeId, 
                           ifIndex);
         
@@ -56,7 +56,7 @@ public class SnmpInterfaceDaoHibernate extends AbstractDaoHibernate<OnmsSnmpInte
     @Override
     public List<OnmsSnmpInterface> findByNodeId(Integer nodeId) {
         Assert.notNull(nodeId, "nodeId may not be null");
-        return find("select distinct snmpIf from OnmsSnmpInterface as snmpIf where snmpIf.node.id = ?",
+        return find("select snmpIf from OnmsSnmpInterface as snmpIf where snmpIf.node.id = ?",
                 nodeId);
 
     }
@@ -74,7 +74,7 @@ public class SnmpInterfaceDaoHibernate extends AbstractDaoHibernate<OnmsSnmpInte
         Assert.notNull(foreignSource, "foreignSource may not be null");
         Assert.notNull(foreignId, "foreignId may not be null");
         Assert.notNull(ifIndex, "ifIndex may not be null");
-        return findUnique("select distinct snmpIf from OnmsSnmpInterface as snmpIf join snmpIf.node as node where node.foreignSource = ? and node.foreignId = ? and node.type = 'A' and snmpIf.ifIndex = ?", 
+        return findUnique("select snmpIf from OnmsSnmpInterface as snmpIf join snmpIf.node as node where node.foreignSource = ? and node.foreignId = ? and node.type = 'A' and snmpIf.ifIndex = ?",
                           foreignSource, 
                           foreignId, 
                           ifIndex);
@@ -85,7 +85,7 @@ public class SnmpInterfaceDaoHibernate extends AbstractDaoHibernate<OnmsSnmpInte
         Assert.notNull(nodeId, "nodeId may not be null");
         Assert.notNull(description, "description may not be null");
 
-        return findUnique("SELECT DISTINCT snmpIf FROM OnmsSnmpInterface AS snmpIf WHERE snmpIf.node.id = ? AND (LOWER(snmpIf.ifDescr) = LOWER(?) OR LOWER(snmpIf.ifName) = LOWER(?))", 
+        return findUnique("SELECT snmpIf FROM OnmsSnmpInterface AS snmpIf WHERE snmpIf.node.id = ? AND (LOWER(snmpIf.ifDescr) = LOWER(?) OR LOWER(snmpIf.ifName) = LOWER(?))",
             nodeId, 
             description,
             description
--- a/opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/SnmpInterfaceDaoHibernate.java
+++ b/opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/SnmpInterfaceDaoHibernate.java
@@ -62,6 +62,14 @@ public class SnmpInterfaceDaoHibernate extends AbstractDaoHibernate<OnmsSnmpInte
     }
 
     @Override
+    public List<OnmsSnmpInterface> findByMacLinksOfNode(Integer nodeId) {
+        Assert.notNull(nodeId, "nodeId may not be null");
+        return find("from OnmsSnmpInterface snmpIf where snmpIf.physAddr in (select l.macAddress from BridgeMacLink l where l.node.id = ?)",
+                nodeId);
+
+    }
+
+    @Override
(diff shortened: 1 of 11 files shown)


opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/SnmpInterfaceDaoHibernate.java
```diff
@@ -53,6 +53,14 @@ public OnmsSnmpInterface findByNodeIdAndIfIndex(Integer nodeId, Integer ifIndex)
         
     }
 
+    @Override
+    public List<OnmsSnmpInterface> findByNodeId(Integer nodeId) {
+        Assert.notNull(nodeId, "nodeId may not be null");
+        return find("select distinct snmpIf from OnmsSnmpInterface as snmpIf where snmpIf.node.id = ?",
```
fooker, line 59: Does this really need to be distinct? As this includes the ID, distinctiveness should be inherent, right?

opennms-web-api/src/main/java/org/opennms/web/enlinkd/EnLinkdElementFactory.java
```diff
@@ -687,7 +706,13 @@ private BridgeLinkNode addBridgeRemotesNodes(Integer nodeid, String mac, BridgeL
             if (sharedmac.equals(mac)) {
                 continue;
             }
-            macsToIpNetTOMediaMap.put(sharedmac, new ArrayList<IpNetToMedia>(m_ipNetToMediaDao.findByPhysAddress(sharedmac)));
+            // half of time spent here; bulk prefetch all IpNetToMedia that can be joined ipn.physAddress = bridgeMacLink.macAddress && bridgeMacLink.node.id = nodeId
+//            final var ipNetToMedias = new ArrayList<IpNetToMedia>(m_ipNetToMediaDao.findByPhysAddress(sharedmac));
```
fooker, line 713: Looks like a remnant of profiling.

opennms-web-api/src/main/java/org/opennms/web/enlinkd/EnLinkdElementFactory.java
```diff
@@ -710,7 +735,13 @@ private BridgeLinkNode addBridgeRemotesNodes(Integer nodeid, String mac, BridgeL
 
             List<OnmsIpInterface> remipaddrs = new ArrayList<OnmsIpInterface>();
             for (IpNetToMedia ipnettomedia : macsToIpNetTOMediaMap.get(sharedmac)) {
-                remipaddrs.addAll(m_ipInterfaceDao.findByIpAddress(ipnettomedia.getNetAddress().getHostAddress()));
+                // half of time spent here; bulk prefetch all OnmsIpInterface where onmsIpInterface.ipAddress == ipnettomedia.netAddress.hostAddress && ipnettomedia.physAddress == bridgeMacLink.macAddress && bridgeMacLink.node.id = nodeId
+//                final var byIpAddress = m_ipInterfaceDao.findByIpAddress(ipnettomedia.getNetAddress().getHostAddress());
```
fooker, line 743: Looks like a remnant of profiling.

opennms-web-api/src/main/java/org/opennms/web/enlinkd/EnLinkdElementFactory.java
```diff
@@ -752,9 +783,27 @@ private BridgeLinkNode addBridgeRemotesNodes(Integer nodeid, String mac, BridgeL
...
+                bridgelinks.add(
+                        convertFromModel(
+                                nodeId,
+                                segment,
+                                bridgeElementCache,
+                                snmpInterfaceCache,
+                                physAddrToIpNetToMedia,
+                                ipAddressToInterfaces
+                        )
+                ); // most time spent here
```
fooker, line 814: Looks like a remnant of profiling.

opennms-web-api/src/main/java/org/opennms/web/enlinkd/EnLinkdElementFactory.java
```diff
@@ -932,4 +989,38 @@ private String getIpList(Set<String> ipstrings) {
...
+            if (ifIdx == null) {
+                return null;
+            } else {
+                var m = map.get(nodeId);
+                if (m == null) {
+                    m = m_snmpInterfaceDao.findByNodeId(nodeId).stream().collect(Collectors.groupingBy(itf -> itf.getIfIndex()));
+                    map.put(nodeId, m);
+                }
+                var l = m.get(ifIdx);
+                return l != null && l.size() == 1 ? l.get(0) : null;
```
fooker, line 1013: Can this be simplified using `Map.computeIfAbsent`?

In addition, the check for a single element can be done while fetching. This would safe some memory. Maybe even using a different `Collector` to store a value if there are no duplicates.

opennms-web-api/src/main/java/org/opennms/web/enlinkd/EnLinkdElementFactory.java
```diff
@@ -932,4 +989,38 @@ private String getIpList(Set<String> ipstrings) {
...
+            }
+        }
+    }
+
+    private class BridgeElementCache {
+        private Map<Integer, Map<Integer, List<BridgeElement>>> map = new HashMap<>();
```
fooker, line 1019: Same here. Why storing a list if only one element is ever needed?

opennms-web-api/src/main/java/org/opennms/web/enlinkd/EnLinkdElementFactory.java
```diff
@@ -718,7 +721,8 @@ private BridgeLinkNode addBridgeRemotesNodes(
         for (String sharedmac: macsToIpNetTOMediaMap.keySet()) {
             final BridgeLinkRemoteNode remlinknode = new BridgeLinkRemoteNode();
             if (macsToIpNetTOMediaMap.get(sharedmac).isEmpty()) {
-                OnmsSnmpInterface snmp = getFromPhysAddress(sharedmac);
+                var l = physAddrToSnmpInterfaces.get(sharedmac);
+                OnmsSnmpInterface snmp = l != null && l.size() == 1 ? l.get(0) : null;
```
fooker, line 725: Same as in the caches. This can be done while building the map. In this case we don't even need the keys with a null value but can skip the whole entry instead.

opennms-web-api/src/main/java/org/opennms/web/enlinkd/EnLinkdElementFactory.java
```diff
@@ -625,7 +626,7 @@ private BridgeLinkNode convertFromModel(
         if (bridgeElement != null) {
             linknode.setBridgeInfo(bridgeElement.getVlanname());
         }
-        return addBridgeRemotesNodes(nodeid, null, linknode, segment, bridgeElementCache, snmpInterfaceCache, physAddrToIpNetToMedias, ipAddressToInterfaces); // most time spent here
+        return addBridgeRemotesNodes(nodeid, null, linknode, segment, bridgeElementCache, snmpInterfaceCache, physAddrToIpNetToMedias, ipAddressToIpInterfaces, physAddrToSnmpInterfaces); // most time spent here
```
fooker, line 629: Looks like a remnant of profiling.

fooker (changes requested):

RangerRick (dismissed):

RangerRick:
In general LGTM, but @fooker's feedback seems reasonable to look into before merging.

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/SnmpInterfaceDaoHibernate.java
swachter: I am not familiar enough with the underlying relational model. AFAIU the same snmp interface can be referenced by several ip interfaces, a node in turn may have several ip interfaces. In summary, it seems that the same snmp interfaces can be related to a node via different ip interfaces.

swachter:
This PR focused on bridge links. Customer feedback suggests that there may be some additional work necessary on the other kinds of links.

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/SnmpInterfaceDaoHibernate.java
fooker: This is not true. The `SnmpInterface` is directly associated to a `Node` in a 1:n relation.

fooker (dismissed):
Just the one thing about using `distinct` left. If this is resolved, this is good to merge.

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/SnmpInterfaceDaoHibernate.java
swachter: You are right, `distinct` is not necessary. I was mislead by other selects in that class that include `distinct`. I removed these occurrences, too.

fooker (approved):

christianpape (approved):
Nice work.

PR #3914, opened 2021-11-26
NMS-13779: Remove links to the wiki

indigo423:
### All Contributors

We have in our global Web UI the "Help" menu which is always available and guides you to the official documentation. I don't think we have to tell on every other place where you can go to the documentation to get more detailed information :)

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13779


Bonrob2 (approved):
I agree, removing it is the better option.

mmahacek (approved):
I like how some of those "wiki" links actually took you to "www".

PR #3930, opened 2021-11-30
NMS-13508: unschedule service polling in case it no longer exists in …

swachter:
Issue: https://issues.opennms.org/browse/NMS-13508

When a polling result for a service is received that is not found in the database then polling for that service gets unscheduled.

### DIFF
--- a/opennms-services/src/main/java/org/opennms/netmgt/poller/DefaultPollContext.java
+++ b/opennms-services/src/main/java/org/opennms/netmgt/poller/DefaultPollContext.java
@@ -442,7 +442,7 @@ public class DefaultPollContext implements PollContext, EventListener {
     public void trackPoll(PollableService service, PollStatus result) {
         try {
             if (!result.isUnknown() && !DISABLE_POLL_TIMESTAMP_TRACKING) {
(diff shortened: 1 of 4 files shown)


swachter:
> If this service was removed out of band, this PR looks good to me.
> 
> Were you able to reproduce this ? IMO, this should have been handled [here](https://github.com/OpenNMS/opennms/blob/a8a58900757c25f21d2a539dab2160b3731361b7/opennms-services/src/main/java/org/opennms/netmgt/poller/PollerEventProcessor.java#L573) in normal control flow. If there is a place where we are not sending `SUSPEND_POLLING_SERVICE_EVENT_UEI`, we should probably handle it with this event.

In the error report there is a `uei.opennms.org/nodes/nodeDeleted` event for a node (id 16545) at 16:42:14. Nearly 40 minutes later there is a `nodeUp` event for the very same node. In `PollerEventProcessor` when a `nodeDeleted` event is processed it gets "deleted". This causes all service pollings to get unscheduled. I do not know why this was not the case here. It seems to be a spurious behavior. Maybe a database hickup at that time?

fooker (approved):

PR #3934, opened 2021-12-01
NMS-13799: install all OpenNMS packages in the docker image by default

RangerRick:
The horizon docker images originally only installed a subset of RPMs under the assumption they wouldn't be used in production, and should be kept slimmer.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13799

### DIFF
--- a/opennms-container/horizon/Dockerfile
+++ b/opennms-container/horizon/Dockerfile
@@ -43,7 +43,7 @@ RUN groupadd -g 10001 opennms && \
 FROM horizon-base
 
 # A default which installs the required OpenNMS Horizon packages
-ARG ONMS_PACKAGES="opennms-core opennms-webapp-jetty opennms-webapp-hawtio"
+ARG ONMS_PACKAGES="opennms opennms-core opennms-webapp-jetty opennms-webapp-hawtio"
 
 # Allow to install optional packages via YUM
 ARG ADD_YUM_PACKAGES
--- a/opennms-container/horizon/Dockerfile
+++ b/opennms-container/horizon/Dockerfile
@@ -43,7 +43,7 @@ RUN groupadd -g 10001 opennms && \
 FROM horizon-base
 
 # A default which installs the required OpenNMS Horizon packages
-ARG ONMS_PACKAGES="opennms opennms-core opennms-webapp-jetty opennms-webapp-hawtio"
+ARG ONMS_PACKAGES="opennms-core opennms-webapp-jetty opennms-webapp-hawtio opennms-plugins"
 
 # Allow to install optional packages via YUM
 ARG ADD_YUM_PACKAGES


indigo423:
What are the additional packages we install with adding the opennms meta package? Does it mean we install now in the container image also PostgreSQL? That would be something which we shouldn't do.

indigo423:
I remember you get a lot of alarms and other noise from all these plugins you have now run but aren't used. One example is the DNS provisioning adapter. They produce a lot of alarm/event and log noise when they are not installed but configured.  You can't easily get rid of them when they are baked into the container image and definitely reduce the value. To be honest, I'm not sure if that is a good idea. I think @agalue has some people running the container image in prod and has some inputs here as well.

indigo423 (approved):
Thanks for using the opennms-plugins meta-package instead of opennms. I'm not in the way here, my concerns expressed :)

RangerRick:
Maybe we should make an `opennms-plugins-recommended` and install that by default?  @agalue do you have recommendations on what should be installed default that's separate and what shouldn't?

indigo423:
> Maybe we should make an `opennms-plugins-recommended` and install that by default? @agalue do you have recommendations on what should be installed default that's separate and what shouldn't?

The idea is nice, it would force us to evaluate what plugins are mature enough and worth supporting.

j-white:
Yep, a lot of alarms will start coming from the optional provisioning adapters with this.

What's actually driving us to include these in the default container?

RangerRick:
The drive was specifically because the cloud stuff expects a plugin to be there and it's not in the Horizon default Docker image. I am suddenly drawing a blank on which adapter it was, one of the SNMP ones I think?  @maximbrener? I'm searching through mattermost and I'm not finding the specific conversation.

It was me that made the assumption that the Docker container should contain all of the possible code, since it would be ridiculous to expect people using the docker image to hand-mount individual optional jars or anything. But what this is really exposing is how much we need to move a lot of this to OIA so it is actually pluggable properly.  :D

RangerRick:
LOL found it, it was actually... the XMP plugin? There is no reason anyone should be even using the XMP plugin, it probably shouldn't even be in the source anymore. @maximbrener can you shed some light on why you'd even want it in the Docker image?

maximbrener:
The drive came from CM project (triggered by xmp for now, but many others will follow once we migrate more plugin's configs to DB). Without plugin we don't have it's XSD. If config migrated to CM it requires XSD for validation, and without XSD onms cannot start due to that.

RangerRick:
In that case we should pull the XSDs out of the plugins and have them packaged with the core as part of the CM project.

fooker (approved):

indigo423 (approved):
-0 from me
We will have some fallout on various places with this PR. If there is a really good reason to install these plugins I'm not aware of, my concerns expressed :)

bborovan (approved):
Makes sense to pull out of the plugins and package it only for CM IMHO.

j-white:
For CM, I'd rather see us push the XSD & change-log to the plugin rather than make these part of the core.

Similarly, I think we should keep the plugins optional, and not try and bake all of them into the default image.

maximbrener:
> For CM, I'd rather see us push the XSD & change-log to the plugin rather than make these part of the core.
> 
> Similarly, I think we should keep the plugins optional, and not try and bake all of them into the default image.

By default, script names are predefined in the main Liquibase xml. To support that dynamically,  we need to do a research to check if we can customize Liquibase to scan classpath for new scripts with migration instructions.

RangerRick:
Gonna go ahead and close this since it seems this needs to be revisited more as an enhancement request and part of a larger effort to determine what should be "default" in the horizon image(s)

PR #3935, opened 2021-12-01
NMS-13711: Add housekeeping for flow thresholding

fooker:
Sessions are persistent but track update state internally. Old sessions
are removed if no data is received after some time.

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13711

### DIFF
--- a/features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/ApplicationKey.java
+++ b/features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/ApplicationKey.java
@@ -30,16 +30,18 @@ package org.opennms.netmgt.flows.elastic.thresholding;
 
 import java.util.Objects;
 
+import org.opennms.netmgt.flows.elastic.Direction;
+
 public class ApplicationKey {
-    public final ExporterKey exporterKey;
     public final int iface;
+    public final Direction direction;
     public final String application;
 
-    public ApplicationKey(final ExporterKey exporterKey,
-                          final int iface,
+    public ApplicationKey(final int iface,
+                          final Direction direction,
                           final String application) {
-        this.exporterKey = Objects.requireNonNull(exporterKey);
         this.iface = iface;
+        this.direction = direction;
         this.application = Objects.requireNonNull(application);
     }
 
@@ -52,15 +54,15 @@ public class ApplicationKey {
             return false;
         }
         final ApplicationKey that = (ApplicationKey) o;
-        return Objects.equals(this.exporterKey, that.exporterKey) &&
-               Objects.equals(this.iface, that.iface) &&
+        return Objects.equals(this.iface, that.iface) &&
+               Objects.equals(this.direction, that.direction) &&
                Objects.equals(this.application, that.application);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(this.exporterKey,
-                            this.iface,
+        return Objects.hash(this.iface,
+                            this.direction,
                             this.application);
     }
 }
--- a/features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
+++ b/features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
@@ -28,19 +28,21 @@
 
 package org.opennms.netmgt.flows.elastic.thresholding;
 
+import java.io.Closeable;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.Collections;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.Timer;
 import java.util.TimerTask;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.atomic.AtomicLong;
 
-import org.opennms.core.cache.Cache;
-import org.opennms.core.cache.CacheBuilder;
-import org.opennms.core.cache.CacheConfig;
 import org.opennms.netmgt.collection.api.CollectionAgent;
 import org.opennms.netmgt.collection.api.CollectionAgentFactory;
 import org.opennms.netmgt.collection.api.ServiceParameters;
@@ -60,14 +62,15 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Strings;
-import com.google.common.cache.CacheLoader;
+import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 
-public class FlowThresholding {
+public class FlowThresholding implements Closeable {
     private static final Logger LOG = LoggerFactory.getLogger(FlowThresholding.class);
 
     public static final String SERVICE_NAME = "Flow-Threshold";
     public static final String RESOURCE_TYPE_NAME = "flowApp";
+    public static final String RESOURCE_GROUP = "application";
 
     private final static RrdRepository FLOW_APP_RRD_REPO = new RrdRepository();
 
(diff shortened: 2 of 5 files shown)


christianpape (changes requested):
Can you add at least a simple test for this?

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
```diff
@@ -76,36 +82,20 @@
...
-                .withCacheLoader(new CacheLoader<ExporterKey, Session>() {
-                    @Override
-                    public Session load(final ExporterKey key) throws Exception {
-                        throw new IllegalStateException();
-                    }
-                })
-                .build();
     }
 
     public long getStepSizeMs() {
```
swachter, line 101: I think that `setStepSize` should be synchronized. Otherwise you may miss a running timer.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
```diff
@@ -76,36 +82,20 @@
 
     private final IpInterfaceDao ipInterfaceDao;
 
-    private final Map<ApplicationKey, AtomicLong> applicationAccumulator;
-
-    private final Cache<ExporterKey, Session> sessions;
+    private final ConcurrentMap<ExporterKey, Session> sessions = Maps.newConcurrentMap();
 
     private long stepSizeMs = 0;
+    private long idleTimeoutMs = 15 * 60 * 1000;
```
swachter, line 88: Should be volatile. Otherwise timer task may not see the current value.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
```diff
@@ -133,78 +123,144 @@ public void run() {
         }, stepSizeMs, stepSizeMs);
     }
 
+    public long getIdleTimeoutMs() {
+        return this.idleTimeoutMs;
+    }
+
+    public void setIdleTimeoutMs(final long idleTimeoutMs) {
```
swachter, line 130: Can we enforce an invariant that `idleTimeoutMs` must be greater than `stepSizeMs`? If `IdleTimeoutMs` is too small sessions get removed before their values are reported.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
```diff
@@ -133,78 +123,144 @@ public void run() {
...
     private static class Session {
+        public final Map<ApplicationKey, AtomicLong> applications;
+
         public final ThresholdingSession thresholdingSession;
         public final CollectionAgent collectionAgent;
 
+        // The last time this session sees any incoming flow.
+        // This is not synchronized as reference updates are always atomic.
+        // See https://docs.oracle.com/javase/specs/jls/se7/html/jls-17.html#jls-17.7
+        private Instant lastUpdate = null;
```
swachter, line 239: Should be volatile, otherwise timer task may not see last value.

swachter (changes requested):
Some minor changes needed.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/ApplicationKey.java
```diff
@@ -30,16 +30,18 @@
 
 import java.util.Objects;
 
+import org.opennms.netmgt.flows.elastic.Direction;
+
```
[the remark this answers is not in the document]
fooker: But all three values are used to uniquely identify an application

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
fooker: It's just a setting. I don't care if a run is lost in this case. This is not intended to be changed after initial config.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
fooker: AFAIK `volatile` is only required for multiple access of the same variable in the same method. Method boundaries are memory barriers and invalidate caches anyways. This is not a tight spinning loop.

Even if this need does a run with the old value after changing the config, this is fine.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
fooker: If idle timeout is lower than step size, this would result in invalidating after doing the threshold. Which would result in strange behavior. I think that this is more a documentation issue as I can not control the order in which the setters are applied during config changes. Therefore we must allow the broken combination as temporary during config changes.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
fooker: Same as above. This is only accessed once in the method.

christianpape (approved):
Looks good to me.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/ApplicationKey.java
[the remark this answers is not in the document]
christianpape: Agree. I think ApplicationKey is good.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
swachter: Method boundaries have nothing to do with memory barriers. Volatile guaranties that changes made by a thread to a volatile variable are seen other threads accessing that variable.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
swachter: As said, `volatile` has nothing to do with methods.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
swachter: The implementation explicitly checks for a running timer. This seems to foresee that the method is called multiple time.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
swachter: The timer could be started in an init method that is called after the bean is completely configured. This would be the time when `idleTimoutMs` and `stepSize` is checked. There could also be a destroy method that stops the timer. This would make the bean a better OSGI inhabitant.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
fooker: True. But again, this is just called by explicit user interaction for configuration. Configuration is not expected to be thread-safe.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
fooker: It is expected to be user-configurable during runtime. So an `init` will not help here.

swachter: Isn't that covered by OSGI? The bean gets reloaded when its configuration is changed (`update-strategy="reload"`)? I am not sure about that, though.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
fooker: altered.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
fooker: altered.

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
fooker: I still don't see why this is a problem at all? Everything is processed at least once, right?

swachter: How shall the configuration change happen? If the configuration file is changed on disk then blueprint will reload the container. A new `ThreadThresholding` bean gets instantiated. The timer task of the previous `ThreadThresholding` bean will run for ever.

fooker: The timer task was not even changed in the PR... An this is about the idle timeout, not the interval of the timer.
In addition, the timer is canceled in the `close` method.

swachter: Where is the `close` method called? The `ElasticFlowRepository` that references the `FlowThresholding` does not. (It would also make no sense, because it does not "open/create" it either.) I can create an issue for the timer task lifecycle if you prefer.

fooker: https://github.com/OpenNMS/opennms/pull/3935/files#diff-79652c14483859c21c60d394b21ddd270fb00d23b810c3aba23d3dacbff95f08R216

fooker: Again, this is even the wrong thread...

swachter (approved):

PR #3949, opened 2021-12-03
NMS-13507: speed-up enlinkd link retrieval by bulk fetching and caching

swachter:
Issue: https://issues.opennms.org/browse/NMS-13507

This PR adds more bulk fetches and share caches for `BridgeElement` and `OnmsSnmpInterface` instances between different subtasks when all kinds of links are retrieved.

The results returned changed slightly because in the original implementation there was a bug in the `EnLinkdElementFactory.getFromCdpCacheDevicePort` method. The constructed `where` clause was not correct. The fix results in more specific cdp link information being retrieved.

### DIFF
--- a/opennms-web-api/src/main/java/org/opennms/web/enlinkd/EnLinkdElementFactory.java
+++ b/opennms-web-api/src/main/java/org/opennms/web/enlinkd/EnLinkdElementFactory.java
@@ -31,23 +31,21 @@ package org.opennms.web.enlinkd;
 import static org.opennms.core.utils.InetAddressUtils.str;
 
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
 import javax.servlet.ServletContext;
 
-import org.opennms.core.criteria.Alias.JoinType;
-import org.opennms.core.criteria.Criteria;
+import org.apache.commons.lang3.tuple.Pair;
 import org.opennms.core.criteria.CriteriaBuilder;
-import org.opennms.core.criteria.restrictions.EqRestriction;
-import org.opennms.core.criteria.restrictions.SqlRestriction.Type;
 import org.opennms.core.spring.BeanUtils;
 import org.opennms.core.sysprops.SystemProperties;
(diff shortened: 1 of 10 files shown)


rssntn67 (approved):

opennms-web-api/src/main/java/org/opennms/web/enlinkd/EnLinkdElementFactory.java
```diff
@@ -995,43 +1024,94 @@ private String getIpList(Set<String> ipstrings) {
...
+            }
+        };
+    }
+
+    /**
+     * Creates a map that contains entries for all keys derived from the values. Keys are mapped to a non-empty optionals
+     * in case the mapping is unique and to an empty optional otherwise.
      */
-    private static <K, V, I> Map<K, V> uniqueMap(List<V> list, Function<V, K> toKey, Function<V, I> toId) {
+    private static <K, V, I> Map<K, Optional<V>> uniqueMap(List<V> list, Function<V, K> toKey, Function<V, I> toId) {
```
fooker, line 1107: This usage of `Optional` looks strange. Is there a distinction between a missing value and am explicitly empty value?

Only left usage seems to be the call above, which results in a `o == null ? null : o.orElse(null)`. Which makes this even more strange.

fooker (changes requested):

fooker (changes requested):
Just a small thing about the usage of `Optional`.

Do you have any insight about the query performance? I'm uncertain about using `JOIN` vs subselects?

christianpape:
> Just a small thing about the usage of `Optional`.
> 
> Do you have any insight about the query performance? I'm uncertain about using `JOIN` vs subselects?



> Just a small thing about the usage of `Optional`.
> 
> Do you have any insight about the query performance? I'm uncertain about using `JOIN` vs subselects?

If the database is smart enough both will result in the same query plan.

opennms-web-api/src/main/java/org/opennms/web/enlinkd/EnLinkdElementFactory.java
swachter: The `toMap` collector does not distinguish between keys that are new and keys that have a `null` value. In both case the next value is added to the collected map without using the combine function. Therefore `null` can not be used as the `value` for indicating a key-value mapping is not unique.

swachter:
> Do you have any insight about the query performance? I'm uncertain about using `JOIN` vs subselects?

I did some research and found no indication that `exists` is performance wise worse than `join`. In addition, using `exists` avoids duplicates in case of a non-unique join criteria. Sure, these could be eliminated by `distinct` but this make the query more complex.

fooker (approved):

PR #3972, opened 2021-12-08
NMS-13790: Compute sequence numbers for flow thresholding

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-13790


features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
```diff
@@ -238,6 +240,7 @@ public void threshold(final List<FlowDocument> documents,
         // This is not synchronized as reference updates are always atomic.
         // See https://docs.oracle.com/javase/specs/jls/se7/html/jls-17.html#jls-17.7
         private volatile Instant lastUpdate = null;
+        private AtomicLong sequenceNumber = new AtomicLong(ThreadLocalRandom.current().nextLong(0, Integer.MAX_VALUE));
```
fooker, line 243: Can be `final`. Or does not even need to be an atomic value.

fooker (changes requested):
Just a minor nit.

fooker:
Thinking a little more about this: Can we make this even more conflict free by taking the system ID (I have faint memories that something like this exists) into account?

fooker (approved):

PR #3980, opened 2021-12-09
NMS-13441: TimescaleDB extension can't added to existing opennms DB.

aramosvizcarra:
Added changes to the Migrator class to support Timescale migration on existing Postgres opennms database.
Currently Migrator add the Timescale extension on template1 this only works when opennms database is created after the extension has been created in template1, otherwise is only available on template1. This is because postgres config is set that way on opennms-datasources.xml
In order to handle the scenario and enforce the extension to be installed on the opennms database superuser access needs to be granted to opennms temporarly.

External References
•       JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-13441


cgorantla (commented):
Nice, looks good to me code wise. 
Let me validate this!

RangerRick (approved):
seems pretty straightforward to me +1

if @cgorantla's test looks good, I say go for it

cgorantla (approved):
Looks good to me.
Validated that all the bmp stat tables get updated to hypertables now.

Updating id and `runAlways=true` on existing changelog seems to be good enough.
@RangerRick  Is this okay or should we make new hypertable changelogs instead of updating current ones.

RangerRick:
> Updating id and `runAlways=true` on existing changelog seems to be good enough. @RangerRick Is this okay or should we make new hypertable changelogs instead of updating current ones.

We talked about this in scrum but for the sake of documenting it here, since it won't break _existing_ checksums this should be fine.  There will be a difference in the liquibase manifest between a fresh install and an upgrade install, but it won't hurt anything.

RangerRick (approved):

cgorantla:
As discussed need to re-target this to `release-29.x`. 
Also if there is a way to migrate existing data to hypertables by default that would be great otherwise we need to add a breaking change doc.

aramosvizcarra:
Starting over with a new branch from release-29.x

PR #3992, opened 2021-12-13
PE-185: Resolve ${install.dir} in provisiond XML

freddy61025:
https://issues.opennms.org/browse/PE-185


patrick-schweizer (approved):
:+1:

PR #3994, opened 2021-12-13
NMS-13855: Add ifName to strings.properties

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-13855

Test can be added when data collection is done.

### DIFF
--- a/features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
+++ b/features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
@@ -161,21 +161,13 @@ public class FlowThresholding implements Closeable {
 
             for (final Map.Entry<ApplicationKey, AtomicLong> application : session.applications.entrySet()) {
                 try {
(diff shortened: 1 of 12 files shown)


RangerRick (approved):

features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/thresholding/FlowThresholding.java
```diff
@@ -158,9 +164,19 @@ private void runTimerTask() {
                     final DeferredGenericTypeResource appResource = new DeferredGenericTypeResource(nodeResource,
                                                                                                     RESOURCE_TYPE_NAME,
                                                                                                     String.format("%s:%s",
-                                                                                                                  application.getKey().iface, // TODO cpape: Find interface name
+                                                                                                                  application.getKey().iface,
```
fooker, line 167: This should use the `ifName`, too. Using the name would stabilize the name and not depend on internal IDs.

fooker (changes requested):

fooker (approved):

PR #4007, opened 2021-12-14
NMS-13649: Add Metrics for IPC Twin

cgorantla:
Add JMX metrics/graphs for IPC Twin.
Also expose them on karaf shell

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13649

### DIFF
--- a/core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinPublisher.java
+++ b/core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinPublisher.java
@@ -28,15 +28,10 @@
 
 package org.opennms.core.ipc.twin.common;
 
-import java.io.IOException;
-import java.nio.charset.StandardCharsets;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Objects;
-import java.util.function.BiConsumer;
-
+import com.codahale.metrics.Counter;
+import com.codahale.metrics.MetricRegistry;
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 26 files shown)


core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinPublisher.java
```diff
@@ -243,7 +257,12 @@ public void publish(T obj) throws IOException {
                     span.setTag(TAG_VERSION, twinUpdate.getVersion());
                     span.setTag(TAG_SESSION_ID, twinUpdate.getSessionId());
                     handleSinkUpdate(twinUpdate);
+                    String sinkUpdateMetricName = sessionKey.location != null ?
+                            MetricRegistry.name(sessionKey.location, sessionKey.key, SINK_UPDATE_SENT) :
+                            MetricRegistry.name(sessionKey.key, SINK_UPDATE_SENT);
```
fooker, line 262: As these names for a tree, maybe we should use the same number of elements in the name. is using `*` allowed?

fooker (changes requested):

core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinPublisher.java
cgorantla: > As these names for a tree, maybe we should use the same number of elements in the name. is using `*` allowed?

mbean naming should take care of it.

Screen Shot 2021-12-15 at 4 55 28 PM
.

cgorantla:
Response graphs may not give good results since it averages the metrics and twin messages are rare.

Screen Shot 2021-12-15 at 5 01 40 PM

PR #4027, opened 2021-12-17
PE-192: migrate org.opennms.features.topology.app.icons.application.cfg to CM

patrick-schweizer:
This PR migrates org.opennms.features.topology.app.icons.application.cfg to the Configuration Management System.

* JIRA (Issue Tracker): https://issues.opennms.org/browse/PE-192

### DIFF
--- a/features/config/upgrade/src/main/resources/changelog-cm/30.0.0/changelog-cm.xml
+++ b/features/config/upgrade/src/main/resources/changelog-cm/30.0.0/changelog-cm.xml
@@ -68,4 +68,12 @@
         <cm:importConfig schemaId="notifd" configId="default" filePath="notifd-configuration.xml"/>
     </changeSet>
 
+    <changeSet author="pschweizer" id="register-org.opennms.features.topology.app.icons.application">
+        <cm:registerSchema id="org.opennms.features.topology.app.icons.application"/>
+        <cm:changeSchema schemaId="org.opennms.features.topology.app.icons.application">
+            <cm:put name="application.application" type="string" default="application"/>
+            <cm:put name="application.monitored-service" type="string" default="IP_service"/>
+        </cm:changeSchema>
+        <cm:importConfig schemaId="org.opennms.features.topology.app.icons.application" configId="default" filePath="org.opennms.features.topology.app.icons.application.cfg"/>
+    </changeSet>
 </databaseChangeLog>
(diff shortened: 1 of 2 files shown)


freddy61025 (approved):

features/config/upgrade/src/main/resources/changelog-cm/30.0.0/changelog-cm.xml
```diff
@@ -68,4 +68,12 @@
         <cm:importConfig schemaId="notifd" configId="default" filePath="notifd-configuration.xml"/>
     </changeSet>
 
+    <changeSet author="pschweizer" id="register-org.opennms.features.topology.app.icons.application">
+        <cm:registerSchema id="org.opennms.features.topology.app.icons.application"/>
+        <cm:changeSchema schemaId="org.opennms.features.topology.app.icons.application">
+            <cm:put name="application.application" type="string" default="application"/>
```
j-white, line 74: This file should allow arbitrary keys, so the schema as defined here may not apply well.

Similar for the other `org.opennms.features.topology.app.icons.*` as well.

patrick-schweizer: wait, we want to allow any key? I was under the impression that we want to have a set of defined keys, their value type and potentially validation.
The ui should know about those keys as well, no?
I think for org.opennms.features.topology.app.icons.application and datachoices there shouldn't be a case where we will have additional keys (unless we make code changes), no?

j-white: Generally, we want as tight of a schema as we can define.

You're right - we would need code changes to accommodate new keys here. So we can keep the schema as is.

I was thinking more generally about the other icon files. `org.opennms.features.topology.app.icons.linkd.cfg` works a bit different in this regard where it allow arbitrary sysoids as the suffix for the keys, but we can handle that when dealing with that particular file.

patrick-schweizer: ok understand. Maybe we need something like a wildcard key definition, something like
<cm:putWildcardKeys name="application.application.*" type="string" />
but I guess that will require some discussion how we want to do that...

PR #4031, opened 2021-12-21
NMS-13869: make eventId a BIGINT

RangerRick:
To update the `eventid`, we also have to delete and recreate the `node_alarms`, `node_outages` and `node_outage_status` views since they rely on the eventid internally.  I ended up reworking how those views get created/recreated so that we don't have to keep making new changelogs deleting and remaking them whenever something changes.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13869

### DIFF
--- a/core/schema/src/main/liquibase/27.1.0/eventid-bigint.xml
+++ b/core/schema/src/main/liquibase/27.1.0/eventid-bigint.xml
@@ -1,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd" >
+
+  <changeSet author="rangerrick" id="27.1.0-eventid-drop-related-views" runAlways="true">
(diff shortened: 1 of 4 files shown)


RangerRick:
As far as I can tell, no further changes were made to views or references to eventid in H29 or develop, so this should merge forward cleanly.

core/schema/src/main/liquibase/27.1.0/eventid-bigint.xml
```diff
@@ -0,0 +1,38 @@
...
+  <changeSet author="rangerrick" id="27.1.0-eventid-drop-related-views" runAlways="true">
+    <preConditions onFail="MARK_RAN">
+      <viewExists viewName="node_outages" />
+    </preConditions>
+
+    <!-- these will get recreated by views/*.xml -->
+    <dropView viewName="node_alarms" />
+    <dropView viewName="node_outages" />
+    <dropView viewName="node_outage_status" />
+  </changeSet>
```
swachter, line 17: Minor thought: Do these views always go together? Would it be more robust to split it into 3 change sets?

swachter (approved):
LGTM! Just a minor thought.

core/schema/src/main/liquibase/27.1.0/eventid-bigint.xml
RangerRick: These views all reference `eventid` and thus have to be deleted before I can change `eventid`'s type (or else PostgreSQL complains about reference errors).  They get "normally" dropped/recreated in the other changes in `views/`

cgorantla:
Should we also update Java types to `long` in Entities ?

RangerRick:
> Should we also update Java types to `long` in Entities ?

Oof, good catch. I guess that's going to make this a `develop`-only thing, that would change serialization.

PR #4037, opened 2021-12-22
NMS-13796: Add TFTP server

swachter:
Issue: https://issues.opennms.org/browse/NMS-13796


features/minion/repository/src/main/resources/features.boot
```diff
@@ -11,6 +11,7 @@ opennms-trapd-listener
 opennms-events-sink-dispatcher
 opennms-send-event-command
 opennms-dnsresolver-netty
+opennms-deviceconfig
```
cgorantla, line 14: This should be optional dependency, doesn't need to be installed by default.

container/karaf/src/main/filtered-resources/etc/org.apache.karaf.features.cfg
```diff
@@ -93,6 +93,7 @@ featuresBoot = ( \
   opennms-measurements-shell,\
   opennms-graphml,\
   opennms-graphs,\
+  opennms-deviceconfig,\
```
cgorantla, line 96: This feature should be optional, doesn't need to be installed by default.

container/features/src/main/resources/features.xml
```diff
@@ -1839,4 +1839,12 @@
     <feature name="opennms-core-ipc-twin-shell" description="OpenNMS :: Core :: IPC :: Twin :: Shell" version="${project.version}">
         <bundle>mvn:org.opennms.core.ipc.twin/org.opennms.core.ipc.twin.shell/${project.version}</bundle>
     </feature>
+
+    <feature name="opennms-deviceconfig" description="OpenNMS :: Features :: Device Config">
```
cgorantla, line 1843: Rename it as `opennms-deviceconfig-tftp` ?

cgorantla (commented):
Basic implementation looks good to me. We can add more api changes when needed

container/features/src/main/resources/features.xml
swachter: I also wondered how fine granular we should be. I thought that a single feature would be enough.

If we decide to use a single feature (for all client related functionality) then maybe we should rename it into `opennms-deviceconfig-client`. (The corresponding `opennms-deviceconfig-server` feature would contain the `pollerd` related things.)

If we decide to use several features then the `shell` submodule (that contains a karaf command for exploring the tftp server) should also be renamed (maybe into `tftp-shell`).

smoke-test/src/main/resources/opennms-overlay/etc/org.opennms.features.deviceconfig.tftp.cfg
```diff
@@ -0,0 +1,2 @@
+port=6901
```
cgorantla, line 1: Don't need this if we are not installing `device-config` features by default.

cgorantla (approved):

PR #4061, opened 2022-01-05
NMS-13859: Sync InterfaceToNode Cache asynchronously

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13859

### DIFF
--- a/opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/InterfaceToNodeCacheDaoImpl.java
+++ b/opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/InterfaceToNodeCacheDaoImpl.java
@@ -42,7 +42,6 @@ import org.opennms.netmgt.dao.api.IpInterfaceDao;
 import org.opennms.netmgt.dao.api.NodeDao;
 import org.opennms.netmgt.model.OnmsIpInterface;
 import org.opennms.netmgt.model.OnmsNode;
-import org.opennms.netmgt.model.OnmsNode.NodeType;
 import org.opennms.netmgt.model.PrimaryType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -62,6 +61,7 @@ import java.util.Timer;
 import java.util.TimerTask;
 import java.util.TreeSet;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadFactory;
@@ -88,6 +88,7 @@ public class InterfaceToNodeCacheDaoImpl extends AbstractInterfaceToNodeCache im
             .setNameFormat("sync-interface-to-node-cache")
             .build();
     private final ExecutorService executorService = Executors.newSingleThreadExecutor(threadFactory);
+    private final CountDownLatch initialNodeSyncDone = new CountDownLatch(1);
 
     private static class Key {
         private final String location;
@@ -217,6 +218,7 @@ public class InterfaceToNodeCacheDaoImpl extends AbstractInterfaceToNodeCache im
     public void init() {
         // sync datasource asynchronously in order to not block bean initialization.
         syncDataSourceAsynchronously().whenComplete((result, ex) -> {
+            initialNodeSyncDone.countDown();
             if (refreshRate > 0) {
                 refreshTimer.schedule(new TimerTask() {
                     @Override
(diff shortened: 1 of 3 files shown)


cgorantla:
Should we also make [markerCache](https://github.com/OpenNMS/opennms/blob/opennms-28.0.0-2/features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/ElasticFlowRepository.java#L189-L204) initialization asynchronous ?

j-white:
What is the expected behavior if the cache is queried (i.e. a call to `getNodeId`) before the initial initialization is complete?

j-white:
> Should we also make [markerCache](https://github.com/OpenNMS/opennms/blob/opennms-28.0.0-2/features/flows/elastic/src/main/java/org/opennms/netmgt/flows/elastic/ElasticFlowRepository.java#L189-L204) initialization asynchronous ?

Yeah, I think that would be worthwhile.

cgorantla:
> What is the expected behavior if the cache is queried (i.e. a call to `getNodeId`) before the initial initialization is complete?

Looks like all the places that use `getNodeId` including `getFirstNodeId()` depend on it Optionally, so I believe we are good.

cgorantla:
> We could probably speed the queries up significantly if just use plain HQL to retrieve the fields we need as opposed to calling [NodeDao#findMatching](https://github.com/OpenNMS/opennms/blob/c36a3d269879ce944541a00ba0fb403aa7dd86e5/opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/InterfaceToNodeCacheDaoImpl.java#L296)

yeah, can try HQL and in-memory removal of `NodeType.DELETE` nodes.

aramosvizcarra (commented):
Looks pretty good to me!

j-white:
> Looks like all the places that use `getNodeId` including `getFirstNodeId()` depend on it Optionally, so I believe we are good.

If we think of the impact on flow processing, some flows may not be identified with the proper node until the cache is primed. It may be better to block the calls until it the cache is initialized instead of returning an empty result.

cgorantla:
> > We could probably speed the queries up significantly if just use plain HQL to retrieve the fields we need as opposed to calling [NodeDao#findMatching](https://github.com/OpenNMS/opennms/blob/c36a3d269879ce944541a00ba0fb403aa7dd86e5/opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/InterfaceToNodeCacheDaoImpl.java#L296)
> 
> yeah, can try HQL and in-memory removal of `NodeType.DELETE` nodes.

I validated HQL, criteria and in-memory removal of `NodeType.DELETE` with 25000 nodes, they all result in same performance:

```
HQL: Time took to get active nodes = 3674 
Criteria : Time took to get active nodes = 3364 
InMemory : Time took to get active nodes = 3596 
```
So didn't update query.

cgorantla:
> > Looks like all the places that use `getNodeId` including `getFirstNodeId()` depend on it Optionally, so I believe we are good.
> 
> If we think of the impact on flow processing, some flows may not be identified with the proper node until the cache is primed. It may be better to block the calls until it the cache is initialized instead of returning an empty result.

Done. also moved markerCache to aynchronous.

j-white:
Looks good - this should do the trick!

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/InterfaceToNodeCacheDaoImpl.java
```diff
@@ -206,21 +216,28 @@ public InterfaceToNodeCacheDaoImpl(long refreshRate) {
...
-                        LOG.error("An error occurred while synchronizing the datasource: {}", ex.getMessage(), ex);
+        // sync datasource asynchronously in order to not block bean initialization.
+        syncDataSourceAsynchronously().whenComplete((result, ex) -> {
+            initialNodeSyncDone.countDown();
+            if (refreshRate > 0) {
+                refreshTimer.schedule(new TimerTask() {
+                    @Override
+                    public void run() {
+                        try {
+                            dataSourceSync();
```
aramosvizcarra, line 227: I have a feeling this is part of the actual expected behavior (I'm not familiar with it) just wondering if is meant to call the dataSourceSync again when refreshRate > 0?

aramosvizcarra (approved):
Just have a question about the functionality (I'm guessing this is the expected functionality), I'm out of context here but from the code side everything else looks pretty good.

RangerRick (approved):

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/InterfaceToNodeCacheDaoImpl.java
cgorantla: yes, this will refresh every 5 mins by this [system property](https://github.com/OpenNMS/opennms/blob/5838d88eb0a0e8ad1a9011a3a142ccc9bc37e1eb/opennms-dao/src/main/resources/META-INF/opennms/applicationContext-shared.xml#L141).

PR #4062, opened 2022-01-05
NMS-13818: split docs and main build in CircleCI

RangerRick:
This PR does a few things:

1. Get rid of the alias-based `defaults:` parameters block, and just turn them into "global" parameters instead.
   This makes them easier to access from anywhere in the CircleCI config.
2. Creates some new parameters used to trigger sub-builds.
3. Creates a new "pre-build" workflow that is the only thing triggered by default.
4. Disables the weekly coverage run for now, dynamic configs + triggers is not supported.
   We should probably do this a different way anyway.

Note that I don't actually use the CircleCI-provided orb for path-filtering; it doesn't support caching the `.git` checkout, so it was taking an inordinate amount of time to run a full checkout on every new build.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13818

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,7 +112,8 @@ docker_container_config: &docker_container_config
 
 orbs:
   cloudsmith: cloudsmith/cloudsmith@1.0.3
-  path-filtering: circleci/path-filtering@0.1.1
+  continuation: circleci/continuation@0.2.0
+  # path-filtering: circleci/path-filtering@0.1.1
   sign-packages: opennms/sign-packages@2.1.3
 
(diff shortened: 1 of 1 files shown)


RangerRick:
Note that when this gets merged forward to `develop`, we should be able to adapt it for the UI build by simply adding another toggle and regex to this existing config for [NMS-13819](https://issues.opennms.org/browse/NMS-13819).

maximbrener (approved):

bborovan (approved):
Looks good Ben!

PR #4064, opened 2022-01-06
NMS-13820: move javadoc to main build

RangerRick:
Since the main `build` job already does a full build, adding on the `javadoc:aggregate` costs essentially nothing. This means we don't have to go through an entire aggregate step later during the `tarball-assembly` phase.

This technically misses javadoc from anything that's in the assembly rather than compile, but we should not be publishing APIs in there anyway.  :D

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13820

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -101,7 +101,7 @@ commands:
       steps:
         - run:
             name: Calculate cache key from pom files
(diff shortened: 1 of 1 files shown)


cgorantla (dismissed):

RangerRick:
Chandra reviewed this and the force-push was just some cleanup so I'm gonna merge this anyway.  :)

PR #4066, opened 2022-01-06
NMS-13837-smoke Smoke tests for Sentnel Health Rest

aramosvizcarra:
Added smoke tested for sentinel Health Rest API

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13837


aramosvizcarra:
Going to use original AR/jira/NMS-13837 branch instead and rename it for smoke tests

PR #4058, opened 2022-01-05
NMS-13837 Add Health Check Rest API on Sentinel

aramosvizcarra:
- Reused the Minion Rest Health check
- Moved Minion Rest from features/minion/core/rest into core/health/rest-cxf
- Both Minion and Sentinel now use the same Health Rest bundle feature

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13837

PR #4091, opened 2022-01-11
NMS-13837-smoke: Add Health Check Rest API on Sentinel

aramosvizcarra:
Same as NMS-13837. For smoke-test purpose only

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-13837


features/container/sentinel/src/main/filtered-resources/etc/org.apache.karaf.features.cfg
```diff
@@ -33,7 +33,7 @@ featuresRepositories = \
 #
```
cgorantla, line 33: Considering that full smoke seems to be more flaky than before, revert this back to old state and 
keep `sentinel-core`, `opennms-health-rest-service` optional.

cgorantla:
Should be resolved by #4058

PR #4094, opened 2022-01-12
proof-of-concept "packed" dynamic CircleCI config

RangerRick:
This is the start of an attempt to make our CircleCI config more modular and easier to maintain.
It does a number of things:

1. splits out the "pre-build" config which determines what part of the build to do from the "main" config
2. splits out the `executors:` and `parameters:` sections of the main config into their own files, since they're more likely to change
3. uses CircleCI's `pack` command to take a directory with `.yml` files and combine them into a single config, which is then processed by the continuation

Obviously there's a lot more that can be done to break things up into more manageable chunks.
Also, there's an opportunity to eventually generate completely different sets of configs depending on the type of build we're doing.
There are tons of opportunities to optimize now that we have this 2-step process that can do logic before determining what to use as a CircleCI config (entirely different workflows for "feature" branches, "release" branches, and others, etc.).


maximbrener (approved):

bborovan (approved):
Makes sense and looks good @RangerRick - will be cleaner and easier to work with going forward as you have it.

PR #4096, opened 2022-01-12
NMS-13889: Upgrade protobuf version

cgorantla:
Re-compiled JTI proto files with protoc-3.19.3 version.
 Generated java files are still protobuf2 compatible.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13889


cgorantla:
Haven't done any manual testing other than depending on JTI related integration tests and smoke test that sends jti payload.

RangerRick (approved):
So AFAICT as long as we didn't use a few specific features of protobuf, this is forwards and backwards compatible over the wire? Assuming that's the case, this seems pretty straightfoward to me.

swachter (approved):
LGTM

PR #4102, opened 2022-01-17
NMS-13913: Replace slashes in collection resource names

fooker:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/13913


christianpape (approved):

j-white:
Make sense. Assuming there was existing resources with a `/` in them, this will result in creating new ones, so maybe worth calling this out a "breaking change" for H30.

fooker:
> Make sense. Assuming there was existing resources with a `/` in them, this will result in creating new ones, so maybe worth calling this out a "breaking change" for H30.

@j-white This was already prevented by a check in `ResourcePath` (https://github.com/OpenNMS/opennms/blob/develop/opennms-model/src/main/java/org/opennms/netmgt/model/ResourcePath.java#L65).

j-white (approved):

PR #4108, opened 2022-01-17
Publish OCI-images for all feature-branches

fooker:


indigo423 (approved):
Don't see an issue having long-running feature branches published to docker hub. We probably need to think about a clean-up procedure in the DockerHub registry.

RangerRick:
Yeah, I think we should have something to clean them up before we put this in place or we'll end up never fixing it and leave a bunch of junk around.

indigo423:
Just some food for thought, DockerHub flags tags as INACTIVE when they are not pulled or pushed. If we can query them with an API call we could use these as candidates for a garbage collection job.

Screenshot 2022-01-20 at 00 03 27

RangerRick:
ooh yeah, that's a good plan

RangerRick (approved):
FYI, there's an automation for this now, so it should be safe to merge.

PR #4109, opened 2022-01-17
NMS-13902: update/replace com.jcraft:jsch:0.1.51

swachter:
Issue: https://issues.opennms.org/browse/NMS-13902

`jcraft/jsch` seems no longer be maintained. This PR replaces our outdated `jsch` dependency with a drop-in replacement that includes newer cryptographic algorithms.


cgorantla (changes requested):
We also need to update bundles in [features](https://github.com/OpenNMS/opennms/blob/1ed67493b807770df28a11a918ab66eb5f6601c1/container/features/src/main/resources/features.xml#L818) file.

RangerRick (dismissed):

RangerRick (dismissed):

cgorantla:
> It looks like this is actually breaking sentinel? or is it just a coincidence? It's been restarted a few times and still failing, might need to dig into the test logs.

yeah, that test is really flaky.  @aramosvizcarra  is looking into it.

swachter:
I also take a look into it. The `karaf.log` looks innocent. There is even a `200` response for `/health/probe`.

cgorantla (approved):
LGTM.

fooker (dismissed):
The change in the test should be reverted before merging.

cgorantla (approved):

PR #4115, opened 2022-01-19
NMS-13645 Smoke tests should use HealthCheck Rest instead of connecting to SSH

aramosvizcarra:
Updated ssh Health Check with Rest Health check for Minion and Sentinel.

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-13645

### DIFF
--- a/smoke-test/src/main/java/org/opennms/smoketest/containers/MinionContainer.java
+++ b/smoke-test/src/main/java/org/opennms/smoketest/containers/MinionContainer.java
@@ -29,6 +29,10 @@
 package org.opennms.smoketest.containers;
 
 import static java.nio.file.Files.createTempDirectory;
+import static java.util.concurrent.TimeUnit.MINUTES;
+import static java.util.concurrent.TimeUnit.SECONDS;
+import static org.awaitility.Awaitility.await;
+import static org.hamcrest.CoreMatchers.containsString;
 import static org.opennms.smoketest.utils.KarafShellUtils.awaitHealthCheckSucceeded;
 import static org.opennms.smoketest.utils.OverlayUtils.jsonMapper;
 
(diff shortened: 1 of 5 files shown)


smoke-test/src/main/java/org/opennms/smoketest/containers/MinionContainer.java
```diff
@@ -241,8 +247,17 @@ public WaitForMinion(MinionContainer container) {
         @Override
         protected void waitUntilReady() {
             LOG.info("Waiting for Minion health check...");
-            final InetSocketAddress sshAddr = container.getSshAddress();
-            awaitHealthCheckSucceeded(sshAddr, 5, "Minion");
```
swachter, line 245: Do we still need the code for `awaitHealthCheckSucceeded`?

swachter (commented):
Looks great! Only one comment.

smoke-test/src/main/java/org/opennms/smoketest/containers/MinionContainer.java
aramosvizcarra: That's a good question.  @cgorantla what do you think?.

cgorantla: May be it  will be used  in another smoke test which should still check for SSH connection on Minion/Sentinel.

cgorantla (approved):

smoke-test/src/main/java/org/opennms/smoketest/containers/MinionContainer.java
swachter: * I propose that all integration test now rely on the rest api when waiting for container startup.
* There should be an additional integration test though, that checks if the `health-check` command is available on the command line. That test could be based on the code of the current ssh-based startup check.

aramosvizcarra: Yes I agree. I'll include that in this PR

swachter (approved):
LGTM - Thanks!

cgorantla (approved):

PR #4116, opened 2022-01-19
PE-98: smoke

dherdt:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/PE-98


RangerRick (approved):
I think I get what's going on here, seems reasonable.

PR #4121, opened 2022-01-19
NMS-13923: TIMETETRA LLDP wrong portnum

rssntn67:
There is no lldpLocalPortNum index in TIMETETRA LLDP MIB.
lldpLocalPortNum is a unique index used in LLDP MIB.
So we need to find a suitable value for set up lldpLink coming from snmp walking 
tmnxLldpRemTable 

So I generate portNum using an hash of the instance id.
		public Integer getLldpLocalPortNum() {return getInstance().toString().hashCode();}

Note that nodeid and lldpLocalPortNum is a unique index in lldplink table.

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13923


features/enlinkd/tests/src/test/java/org/opennms/netmgt/nb/Nms13923NetworkBuilder.java
```diff
@@ -0,0 +1,60 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2013-2021 The OpenNMS Group, Inc.
```
christianpape, line 4:
```suggestion
 * Copyright (C) 2021 The OpenNMS Group, Inc.
```

christianpape (dismissed):
LGTM.

christianpape (approved):

PR #4155, opened 2022-01-27
Jira/nms 13914

swachter:
Issue: https://issues.opennms.org/browse/NMS-13914


features/minion/repository/pom.xml
```diff
@@ -74,6 +74,7 @@
                                 <feature>minion-api-layer</feature>
                                 <feature>minion-plugin-sample</feature>
                                 <feature>opennms-deviceconfig</feature>
+                                <feature>opennms-deviceconfig-rest</feature>
```
cgorantla, line 77: we don't need this feature on Minion

features/device-config/rest/api/src/main/java/org/opennms/features/deviceconfig/rest/api/DeviceConfigRestService.java
```diff
@@ -0,0 +1,78 @@
+/*******************************************************************************
```
cgorantla, line 1: I think we can combine rest api/impl maven modules into one. No need for both of them to exist

swachter: Now, there is only one module.

features/device-config/rest/pom.xml
```diff
@@ -0,0 +1,102 @@
...
+        <artifactId>org.opennms.features.device-config</artifactId>
+        <groupId>org.opennms.features</groupId>
+        <version>30.0.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>org.opennms.features.device-config</groupId>
+    <artifactId>org.opennms.features.device-config.rest</artifactId>
+    <packaging>bundle</packaging>
+    <name>OpenNMS :: Features :: Device Config :: ReST :: Impl</name>
```
cgorantla, line 15: nit: Can remove :: Impl

features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/api/DeviceConfigDto.java
```diff
@@ -0,0 +1,82 @@
...
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.features.deviceconfig.rest.api;
+
+import java.util.Date;
+
+import org.opennms.features.deviceconfig.persistence.api.DeviceConfig;
+
+public class DeviceConfigDto {
```
cgorantla, line 35: nit: DeviceConfigDTO would be better name ?

features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestService.java
```diff
@@ -0,0 +1,132 @@
...
+            Integer limit,
+            Integer offset,
+            String orderBy,
+            String order,
+            Integer ipInterfaceId,
+            String deviceType,
+            Long createdAfter,
+            Long createdBefore
+    ) {
+        var criteriaBuilder = new CriteriaBuilder(DeviceConfig.class);
```
cgorantla, line 63: CriteriaBuilder could be moved a different method

features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/api/DeviceConfigDto.java
swachter: I am used to "interpret" abbreviations as words in code (e.g. `Id` instead of `ID`). This seems more regular to me. However, having all other DTOs using a capitalized suffix, I changed the name.

features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestService.java
swachter: I do not understand exactly. Do you mean moving the construction logic of the criteria into a separate method? In that case all arguments would have to be passed to that method. I do not see an advantage in doing so. That logic is used on at one place.

cgorantla: It can be used in future in other methods. I'm good with either way

PR #4161, opened 2022-01-27
NMS-13945: Use current string attrs for threshold filters

fooker:
While evaluating the thresholds resource-filter, the attribute have been
read from the RRD repository (as stored in strings.properties). This
will fail if the string is not persisted (because the collection set us
purely used for thresholding) or on the first thresholding run, if the
thresholding is executed before the persistence.

This is changed to use the values from the currently processed
collection set.

### Reviewer Remarks

This PR is split in two commits. The first commit contains the actual (very small) change. The second commit is cleanup of the (now unused) RRD repository from the thresholding API.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13945


fooker:
@j-white pining you just for a quick check on possible impact of this change.

j-white:
Looks sane. This makes the assumption that the string attributes are always present in the `CollectionSet`, which is reasonable.

Nice to see the call to the `ResourceDao` and carrying the `RrdRepository` removed!

cgorantla (approved):

PR #4170, opened 2022-01-31
NMS-13729: Event configuration UI fails to persist logmsg dest changes

aramosvizcarra:
Added a converter to the field so the Combobox can bind properly with an Enum

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13729


amay94:
looks like an integration test is also failing

aramosvizcarra:
> looks like an integration test is also failing
That should be fixed now.

cgorantla (approved):

RangerRick (approved):

PR #4172, opened 2022-01-31
NMS-13957: remove EasyMock

RangerRick:
This PR removes EasyMock and converts all tests that used it to use Mockito instead. This will ease moving to JDK17.

For the most part this involves changing `EasyMock.createMock()` to `Mockito.mock()`, `EasyMock.expect()` to `Mockito.when()` and so on. There are not _too_ many places I had to do algorithmic changes to handle the way they do things differently.

I also generally added `verifyNoMoreInteractions()` in a `tearDown()` method when mocks are created in `setUp()`.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13957


cgorantla (dismissed):
LGTM!

pom.xml
```diff
@@ -3659,24 +3659,12 @@
         <artifactId>opencsv</artifactId>
         <version>${opencsvVersion}</version>
       </dependency>
-      <dependency>
```
j-white, line 3662: +1

j-white (approved):

PR #4177, opened 2022-02-01
NMS-13948: Fix '%n%n' in 'full text' system report

geraldhumphries:
Jira issue: https://issues.opennms.org/projects/NMS/issues/NMS-13948


RangerRick (approved):
lol

PR #4182, opened 2022-02-02
NMS-13647: Flow Thresholds

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-13647


indigo423:
Can we add a piece of documentation that explains the RRD step size and the default flush interval setting you can use in Karaf? I think this might be interesting to people regarding graphing and thresholding resolution.

indigo423 (approved):
Nice to have for me if we can add a piece of documentation that explains the RRD step size and the default flush interval setting you set in Karaf and is by default 5 minutes? I think this might be interesting to people regarding graphing versus thresholding resolution.

indigo423 (approved):
Thank you! Can't wait to play with this feature :)

j-white:
Looks awesome! Excited for this one too.

cgorantla (approved):

PR #4185, opened 2022-02-02
PE-162: [notifications.xml] - migrate notifications.xml to the new CM

freddy61025:
https://issues.opennms.org/browse/PE-162


features/config/upgrade/src/main/resources/changelog-cm/30.0.0/changelog-cm.xml
```diff
@@ -83,6 +83,12 @@
         <cm:importConfig schemaId="wmi" filePath="wmi-config.xml"/>
     </changeSet>
 
+    <changeSet author="uguggilam" id="1.0-register-schema-notifications">
+        <cm:registerSchema id="notifications"/>
+        <cm:importSchemaFromXsd id="notifications" xsdFileName="notifications.xsd" xsdFileHash="ef29ac614357d53f27aa75ff810e8b9ba658d48f8c725478a5b1a46311bff883" rootElement="notifications"/>
+        <cm:importConfig schemaId="notifications" configId="default" filePath="notifications.xml"/>
```
patrick-schweizer, line 89: nit: configId does not need to be specified anymore if its value is "default"

features/config/upgrade/src/test/java/org/opennms/config/upgrade/LiquibaseUpgraderIT.java
```diff
@@ -198,10 +198,10 @@ public void shouldRunChangelog() throws LiquibaseException, ValidationException,
 
             // check for org.opennms.features.datachoices.cfg
             Optional<JSONObject> config = this.cm.getJSONConfiguration("org.opennms.features.datachoices", "default");
-            assertEquals(7, config.get().keySet().size());
+            // only those field have default value will autofill
```
patrick-schweizer, line 201: maybe change to:
_only fields with a default value will auto fill_

opennms-config-model/src/main/java/org/opennms/netmgt/config/notifications/Notification.java
```diff
@@ -28,236 +28,213 @@
...
-    private String m_noticeQueue;
+    private Rule rule;
 
-    @XmlElement(name = "destinationPath", required = true)
-    private String m_destinationPath;
+    private String noticeQueue;
 
-    @XmlElement(name = "text-message", required = true)
-    private String m_textMessage;
+    @JsonProperty("destinationPath")
```
patrick-schweizer, line 60: do we need the annotation for a json fieldwith the same name as the Java attribute?

patrick-schweizer (commented):
just a few minor comments

opennms-config-model/src/main/java/org/opennms/netmgt/config/notifications/Notification.java
freddy61025: Because field in xml are mostly kebab-case. The object mapper is set to convert to kebab-case. For destinationPath it is camelCase. Unless we change the attribute case in xsd otherwise we need that hardcode name.

patrick-schweizer: got it!

patrick-schweizer (approved):
Looks good to me.

PR #4186, opened 2022-02-02
NMS-12863: Fix unexpected result from show-event-config after adding new event definitions

geraldhumphries:
- Fix the same event displaying over and over in show-event-config
- Fix events with priorities multiplying every time event config is reloaded
- Add unit tests for show-event-config and the Events XML class

https://issues.opennms.org/browse/NMS-12863


opennms-config-model/src/main/java/org/opennms/netmgt/xml/eventconf/Events.java
```diff
@@ -223,7 +199,7 @@ public void loadEventFilesIfModified(final Resource configResource, final Map<St
 
             final Events events = JaxbUtils.unmarshal(Events.class, eventResource);
             if (events.getEvents().isEmpty()) {
-                throw new IllegalStateException("Uh oh! An event file "+eventResource.getFile()+" with no events has been laoded!");
+                throw new IllegalStateException("Uh oh! An event file " + eventResource.getFile() + " with no events has been laoded!");
```
christianpape, line 202:
```suggestion
                throw new IllegalStateException("Uh oh! An event file " + eventResource.getFile() + " with no events has been loaded!");
```

opennms-config-model/src/main/java/org/opennms/netmgt/xml/eventconf/Events.java
```diff
@@ -70,10 +37,19 @@
 import org.springframework.orm.ObjectRetrievalFailureException;
 import org.springframework.util.StringUtils;
 
-@XmlRootElement(name="events")
+import javax.xml.bind.annotation.*;
```
cgorantla, line 40: Don't use wildcard imports. You can change this setting in IntelliJ. 
It's fine for now but I generally avoid formatting whole file but only format added sections that way reviewer can concentrate on what changed.

cgorantla (commented):
LGTM code wise.

@geraldhumphries Was it verified with the same usecase as described in issue description i.e. with eventconf from OIA ?

opennms-config-model/src/main/java/org/opennms/netmgt/xml/eventconf/Events.java
geraldhumphries: Got it, thank you. Will toggle that setting and avoid whole-file formatting in the future.

geraldhumphries:
@cgorantla it has not been verified with eventconf from OIA, as I was able to reproduce the issue without it. Will verify it still works with OIA.

opennms-config-model/src/main/java/org/opennms/netmgt/xml/eventconf/Events.java
cgorantla: removing wildcard imports is not optional since OpenNMS code doesn't use wildcard imports.
 formatting is fine for now.

cgorantla:
> @cgorantla it has not been verified with eventconf from OIA, as I was able to reproduce the issue without it. Will verify it still works with OIA.

You can update [these](https://github.com/OpenNMS/opennms-integration-api/blob/f85d7f0922407109e3484fa377d0f4b0fa4f37ef/sample/src/main/java/org/opennms/integration/api/sample/MyEventConfExtension.java#L41) files locally, compile and install this feature on karaf `opennms-plugin-sample`.
Let me know if you need help with this.

opennms-config-model/src/main/java/org/opennms/netmgt/xml/eventconf/Events.java
geraldhumphries: Removed them by changing IntelliJ's settings and running "optimize imports".

geraldhumphries:
@cgorantla I was able to verify that events from OIA still work, are organized by priority, and are not duplicated.

cgorantla (approved):

PR #4205, opened 2022-02-07
NMS-13937: Add device config sink

swachter:
Issue: https://issues.opennms.org/browse/NMS-13937

* adds device config sink
* changes config type to be arbitrary string
* provides additional script variables when calling `SshScriptingService` in retriever


features/device-config/persistence/api/src/main/java/org/opennms/features/deviceconfig/persistence/api/DeviceConfig.java
```diff
@@ -75,7 +76,7 @@
 
     @Column(name = "config_type", nullable = false)
     @Enumerated(EnumType.STRING)
```
cgorantla, line 78: we can remove EnumType here ?

opennms-full-assembly/pom.xml
```diff
@@ -615,8 +615,17 @@
                 <!-- Include the OpenNMS Minion features produced by container/features/pom.xml -->
                 <feature>minion-core-api</feature>
                 <feature>opennms-dao-minion</feature>
-                <feature>opennms-deviceconfig</feature>
+                <!-- Lots of deviceconfig features -->
+                <feature>opennms-deviceconfig-deps</feature>
                 <feature>opennms-deviceconfig-rest</feature>
+                <feature>opennms-deviceconfig-tftp</feature>
```
cgorantla, line 621: Shrink this to one module for sink

features/minion/repository/pom.xml
```diff
@@ -74,7 +74,13 @@
                                 <feature>minion-health-check</feature>
                                 <feature>minion-api-layer</feature>
                                 <feature>minion-plugin-sample</feature>
-                                <feature>opennms-deviceconfig</feature>
+                                <feature>opennms-deviceconfig-deps</feature>
```
cgorantla, line 77: Shrink this to one module for sink and one module for monitor

opennms-full-assembly/pom.xml
swachter: `tftp` is used both, by `sink` and by `monitor`. Shouldn't we keep it separate and reference it in these features?

features/minion/repository/pom.xml
swachter: There is now one feature for sink and one for monitor. But there are still some more (tftp and deps). Deps should be scrutinized to reduce it to a minimum.

opennms-full-assembly/pom.xml
cgorantla: Don't need `opennms-deviceconfig-monitor` here as it is not `osgi` module on OpenNMS.  Just need `opennms-deviceconfig-sink-consumer`  and `opennms-deviceconfig-sink` here.

swachter: I wasn't aware that transitive features are included automatically. Great that they can be removed here.

features/minion/repository/pom.xml
swachter: Removed the transitive features.

features/device-config/retrieval/src/main/java/org/opennms/features/deviceconfig/retrieval/impl/RetrieverImpl.java
```diff
@@ -61,16 +63,28 @@
...
+    public RetrieverImpl(SshScriptingService sshScriptingService, TftpServer tftpServer) throws Exception {
         this.sshScriptingService = sshScriptingService;
         this.tftpServer = tftpServer;
         this.executor = Executors.newCachedThreadPool(new ThreadFactoryBuilder().setNameFormat("device-config-retriever-%d").build());
+        this.tftpServerIp = determineIp();
+    }
+
+    private String determineIp() throws Exception {
+        try(final DatagramSocket socket = new DatagramSocket()) {
+            socket.connect(InetAddress.getByName("8.8.8.8"), 10002);
```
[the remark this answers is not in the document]
RangerRick: Hrm, yeah, I'm not sure this is a good idea as implemented, we have plenty of installs running in locked down environments, requiring access to resolving an outside hostname should not be assumed, especially one specifically run by Google.  :D

(I realize you don't have to be able to _connect_ to that host, just resolve it, but even _that_ should not be assumed.)

Figuring out your "real" IP is fraught with peril, so while most of the time this would work, IMHO it is dangerous to _assume_ that this will work.  At the very least, can we make it possible to explicitly configure the IP to connect to in metadata, and fall back to this behavior if it's not set? That at least leaves an out for weird environments.

I assume that's easier than trying to ask the device you connect to what it thinks your IP is (like `who -m` on linux, for example).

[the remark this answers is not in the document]
swachter: I think the operation is not critical. It is like a fire and forget (send an UDP package) and then ask what was the used sender address.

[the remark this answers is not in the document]
swachter: I don't think so. I copied these lines form Stackoverflow (cf. https://stackoverflow.com/questions/9481865/getting-the-ip-address-of-the-current-machine-using-java). However, I am not an expert in those low-level mechanics.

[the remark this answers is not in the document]
RangerRick: OK, just did a test, and I missed this the first time around reading the java docs:

>  The host name can either be a machine name, such as "java.sun.com", or a textual representation of its IP address. If a literal IP address is supplied, **_only the validity of the address format is checked_**.

So it does _not_ rely on actually resolving the address, since it is parseable as an IP address.  You can literally use any IP you want in here, so maybe it would be more clear that we don't care what it is by using `InetAddressUtils.UNPINGABLE_ADDRESS`.

```suggestion
            socket.connect(InetAddressUtils.UNPINGABLE_ADDRESS, 10002);
```

opennms-full-assembly/pom.xml
swachter: Removed `opennms-deviceconfig-monitor` OSGI feature from `opennms-full-assembly`.

features/device-config/retrieval/src/main/java/org/opennms/features/deviceconfig/retrieval/impl/RetrieverImpl.java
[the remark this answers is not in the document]
swachter: Tests by @RangerRick showed that it works even in case that DNS can not resolve anything.

container/features/src/main/resources/features.xml
```diff
@@ -1876,16 +1876,44 @@
...
+
+    <feature name="opennms-deviceconfig-tftp" description="OpenNMS :: Features :: Device Config :: TFTP">
+        <feature>opennms-deviceconfig-deps</feature>
         <bundle>mvn:org.opennms.features.device-config/org.opennms.features.device-config.shell/${project.version}</bundle>
-        <bundle>mvn:org.opennms.features.device-config/org.opennms.features.device-config.retrieval/${project.version}</bundle>
-        <bundle>mvn:org.opennms.features.device-config/org.opennms.features.device-config.ssh-scripting/${project.version}</bundle>
         <bundle>mvn:org.opennms.features.device-config/org.opennms.features.device-config.tftp/${project.version}</bundle>
     </feature>
 
+    <feature name="opennms-deviceconfig-monitor" description="OpenNMS :: Features :: Device Config :: Remote Poller">
```
cgorantla, line 1891: Name shouldn't be "Remote Poller", just say Monitor

cgorantla (approved):
I have a comment on feature name for `opennms-deviceconfig-monitor`.

Also we may need to revisit persisting `filename` as `config-type`.

LGTM for merging it into `feature/device-config`

PR #4206, opened 2022-02-07
NMS-13723: config-tester doesn't find malformed resourceTypes

freddy61025:
https://issues.opennms.org/browse/NMS-13723


RangerRick (approved):
+1

PR #4207, opened 2022-02-08
NMS-13936: Add service to trigger device backup manually

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13936


features/device-config/service/src/main/java/org/opennms/features/deviceconfig/service/DeviceConfigServiceImpl.java
```diff
@@ -0,0 +1,135 @@
...
+                return null;
+            }
+            OnmsIpInterface ipInterface = ipInterfaceOptional.get();
+            OnmsNode node = ipInterface.getNode();
+
+            return new SimpleMonitoredService(ipInterface.getIpAddress(), node.getId(), node.getLabel(), serviceName, location);
+        });
+
+        if (service == null) {
+            throw new IllegalArgumentException("No interface found with ipAddress " + ipAddress + "at location " + location);
```
christianpape, line 92:
```suggestion
            throw new IllegalArgumentException("No interface found with ipAddress " + ipAddress + " at location " + location);
```

features/device-config/service/src/main/java/org/opennms/features/deviceconfig/service/DeviceConfigServiceImpl.java
```diff
@@ -0,0 +1,135 @@
...
+            LOG.error("Couldn't find package {} in poller-config", pkg);
+            throw new IllegalArgumentException("Couldn't find package " + DEVICE_CONFIG_PACKAGE_NAME);
+        }
+        if (Strings.isNullOrEmpty(serviceName)) {
+            serviceName = DEVICE_CONFIG_SERVICE_NAME;
+        }
+        final org.opennms.netmgt.config.poller.Service svc = pollerConfig.getServiceInPackage(serviceName, pkg);
+        if (svc == null) {
+            LOG.error("Couldn't find {} service in package {}", serviceName, pkg);
+            throw new IllegalArgumentException("Couldn't find service " + serviceName + "in the package " + DEVICE_CONFIG_PACKAGE_NAME);
```
christianpape, line 121:
```suggestion
            throw new IllegalArgumentException("Couldn't find service " + serviceName + " in the package " + DEVICE_CONFIG_PACKAGE_NAME);
```

features/device-config/service/src/main/java/org/opennms/features/deviceconfig/service/DeviceConfigServiceImpl.java
```diff
@@ -0,0 +1,135 @@
...
+        }
+        final Optional<Package.ServiceMatch> service = pkg.findService(serviceName);
+        if (service.isEmpty()) {
+            LOG.error("Couldn't find {} service", serviceName);
+            throw new IllegalArgumentException("Couldn't find service " + serviceName + "in the package " + DEVICE_CONFIG_PACKAGE_NAME);
```
christianpape, line 126:
```suggestion
            throw new IllegalArgumentException("Couldn't find service " + serviceName + " in the package " + DEVICE_CONFIG_PACKAGE_NAME);
```

features/device-config/service/src/main/java/org/opennms/features/deviceconfig/service/DeviceConfigServiceImpl.java
```diff
@@ -0,0 +1,135 @@
...
+        }
+        final ServiceMonitor monitor = pollerConfig.getServiceMonitor(service.get().service.getName());
+        if (monitor == null) {
+            LOG.error("Service {} doesn't have a monitor class defined", serviceName);
+            throw new IllegalArgumentException("Service " + serviceName + "doesn't have a monitor class defined");
```
christianpape, line 131:
```suggestion
            throw new IllegalArgumentException("Service " + serviceName + " doesn't have a monitor class defined");
```

christianpape (approved):
LGTM, only some missing spaces.

swachter (commented):
Should we really use poller here? An rpc call would also do. The implementation should be easy based on the common extracted into `Retriever` and `DeviceConfig`.

cgorantla:
> Should we really use poller here? An rpc call would also do. The implementation should be easy based on the common extracted into `Retriever` and `DeviceConfig`.

With Poller, we can reuse `ServiceMonitorAdaptor` and don't need to recreate another RPC module and all the request/response types.

features/device-config/service/src/main/java/org/opennms/features/deviceconfig/service/DeviceConfigServiceImpl.java
```diff
@@ -0,0 +1,135 @@
...
+
+        String className = retrieveClassName(serviceName);
+
+        MonitoredService service = sessionUtils.withReadOnlyTransaction(() -> {
+
+            Optional<OnmsIpInterface> ipInterfaceOptional = ipInterfaceDao.findByIpAddress(ipAddress).stream().filter(ipInterface ->
+                    ipInterface.getNode().getLocation().getLocationName().equals(location)).findFirst();
+            if (ipInterfaceOptional.isEmpty()) {
+                return null;
+            }
```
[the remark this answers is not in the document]
cgorantla: sure, will update once that PR get merged.

features/device-config/service/src/main/java/org/opennms/features/deviceconfig/service/DeviceConfigServiceImpl.java
```diff
@@ -0,0 +1,145 @@
...
+    private ServiceMonitorAdaptor serviceMonitorAdaptor;
+
+
+    @Override
+    public void triggerConfigBackup(String ipAddress, String location, String configType) throws IOException {
+
+        if (!deviceConfigServiceMap.containsKey(ConfigType.valueOf(configType))) {
+            throw new IllegalArgumentException("Unknown configType " + configType);
+        }
+        String serviceName = deviceConfigServiceMap.get(ConfigType.valueOf(configType));
```
swachter, line 87: We do not need that logic if we agree that the service name = `DeviceConfig-<configType>`.

features/device-config/service/src/main/java/org/opennms/features/deviceconfig/service/DeviceConfigServiceImpl.java
```diff
@@ -0,0 +1,145 @@
...
+        if (pkg == null) {
+            LOG.error("Couldn't find package {} in poller-config", pkg);
+            throw new IllegalArgumentException("Couldn't find package " + DEVICE_CONFIG_PACKAGE_NAME);
+        }
+
+        final org.opennms.netmgt.config.poller.Service svc = pollerConfig.getServiceInPackage(serviceName, pkg);
+        if (svc == null) {
+            LOG.error("Couldn't find {} service in package {}", serviceName, pkg);
+            throw new IllegalArgumentException("Couldn't find service " + serviceName + " in the package " + DEVICE_CONFIG_PACKAGE_NAME);
+        }
```
swachter, line 132: `svc` is not used below. Do we really need that access / check?

swachter (commented):
I added some comments based on our discussion on how service name and config type may relate.

features/device-config/service/src/main/java/org/opennms/features/deviceconfig/service/DeviceConfigServiceImpl.java
cgorantla: I feel It's better to have a static map than deriving in the code.

swachter: If we don't want to rely on a convention that maps service names into config types then we should just iterate over the services contained in the `device-config` package and register a mapping from configured config types into service names.

cgorantla: For manual trigger, all the service parameters need to be fetched from service metadata [here](https://github.com/OpenNMS/opennms/blob/0fcbc5a092113a9f67103f0efcdb217d9142d7de/features/poller/client-rpc/src/main/java/org/opennms/netmgt/poller/client/rpc/PollerRequestBuilderImpl.java#L140).

Why would there be multiple services for the same `configType` on a given interface ?

features/device-config/service/src/main/java/org/opennms/features/deviceconfig/service/DeviceConfigServiceImpl.java
```diff
@@ -0,0 +1,105 @@
...
+        // All the service parameters should be loaded from metadata in PollerRequestBuilderImpl
+        // Persistence will be performed in DeviceConfigMonitorAdaptor.
+        final CompletableFuture<PollerResponse> future = locationAwarePollerClient.poll()
+                .withService(service)
+                .withAdaptor(serviceMonitorAdaptor)
+                .withMonitorClassName(DEVICE_CONFIG_MONITOR_CLASS_NAME)
+                .execute();
+        future.whenComplete(((pollerResponse, throwable) -> {
+            if (throwable != null) {
+                LOG.info("Error while manually triggering config backup for IpAddress {} at location {} for service {}", ipAddress, location, service);
```
swachter, line 101: Maybe we want the exception to be also logged? Just add it as an extra argument.

```suggestion
                LOG.info("Error while manually triggering config backup for IpAddress {} at location {} for service {}", ipAddress, location, service, throwable);
```

swachter (approved):
LGTM!

PR #4208, opened 2022-02-08
NMS-13900: fix `opennms status` when attach API is unavailable

RangerRick:
Thanks to [a bug in the JDK](https://bugs.openjdk.java.net/browse/JDK-8226919) there is literally no way to work around the attach API not working other than to do this by completely other means.

My changes are as follows:

* log service status to STDOUT or `output.log` to make it easier for debugging startup when using systemctl (the last output will show in `systemctl status opennms`)
* implement a signal handler for `SIGUSR1` to dump status to a file for consumption by CLI tools (`SIGHUP` is already taken by the thread dump listener)
* update `$OPENNMS_HOME/bin/opennms` to use `kill -USR1` to trigger the status dump, and display the output
* while I was at it, `opennms stop` has the same issue, so I changed it to just use `kill` since the `Manager` implements a shutdown handler to safely exit anyway

### All Contributors

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13900

### DIFF
--- a/core/daemon/src/main/java/org/opennms/netmgt/vmmgr/Invoker.java
+++ b/core/daemon/src/main/java/org/opennms/netmgt/vmmgr/Invoker.java
@@ -267,7 +267,11 @@ public class Invoker {
             }
         }
 
-        LOG.debug("Invoking {} on object {}", invoke.getMethod(), mbean.getObjectName());
+        if ("status".equals(invoke.getMethod())) {
+            LOG.debug("Invoking {} on object {}", invoke.getMethod(), mbean.getObjectName());
+        } else {
+            LOG.info("Invoking {} on object {}", invoke.getMethod(), mbean.getObjectName());
+        }
         
 
         Object object;
@@ -283,7 +287,11 @@ public class Invoker {
             throw t;
         }
 
-	LOG.debug("Invocation {} successful for MBean {}", invoke.getMethod(), mbean.getObjectName());
+        if ("status".equals(invoke.getMethod())) {
+            LOG.debug("Invocation {} successful for MBean {}", invoke.getMethod(), mbean.getObjectName());
+        } else {
+            LOG.info("Invocation {} successful for MBean {}", invoke.getMethod(), mbean.getObjectName());
+        }
 
         return object;
     }
--- a/core/daemon/src/main/java/org/opennms/netmgt/vmmgr/Invoker.java
+++ b/core/daemon/src/main/java/org/opennms/netmgt/vmmgr/Invoker.java
@@ -1,8 +1,8 @@
 /*******************************************************************************
  * This file is part of OpenNMS(R).
  *
- * Copyright (C) 2007-2014 The OpenNMS Group, Inc.
- * OpenNMS(R) is Copyright (C) 1999-2014 The OpenNMS Group, Inc.
+ * Copyright (C) 2007-2022 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2022 The OpenNMS Group, Inc.
  *
  * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
  *
@@ -28,8 +28,15 @@
 
 package org.opennms.netmgt.vmmgr;
 
+import static java.nio.file.StandardOpenOption.CREATE;
+import static java.nio.file.StandardOpenOption.TRUNCATE_EXISTING;
+
+import java.io.IOException;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
+import java.nio.charset.Charset;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


opennms-base-assembly/src/main/filtered/bin/opennms
```diff
@@ -634,23 +635,56 @@ doStatus(){
...
-
 	# If there is a PID file for OpenNMS, specify the PID
 	PID="$(getPid)"
 	if [ "$PID" -gt 0 ]; then
-		PARM_PID=("-p" "$PID")
+		# see org.opennms.netmgt.vmmgr.Invoker, we register a custom
+		# signal handler to look for status requests
+		kill -USR1 "$PID"
+
+		touch "/tmp/pre-check.$$"
```
amay94, line 645: is this file doing anything?

core/daemon/src/main/java/org/opennms/netmgt/vmmgr/Invoker.java
```diff
@@ -137,15 +148,55 @@ public void instantiateClasses() {
...
+
+        final var output = new StringBuilder();
+
+        for (final var invokerService : getServices()) {
+            final var serviceName = invokerService.getService().getName();
+            for (final var invoke : invokerService.getService().getInvokes()) {
+                if ("status".equals(invoke.getMethod())) {
+                    try {
+                        Object result = invoke(invoke, invokerService.getMbean());
+                        output.append(serviceName).append("=").append(result.toString()).append("\n");
```
amay94, line 182: this could potentially be a cleaner place to format the output

opennms-base-assembly/src/main/filtered/bin/opennms
RangerRick: It's there to be able to do a "newer than" check, to make sure I'm not reading an old status file.  I didn't want to _delete_ the existing status file.  I suppose I could use something along the lines of `find -newermt` but this seemed easiest.

opennms-base-assembly/src/main/filtered/bin/opennms
```diff
@@ -634,23 +635,56 @@ doStatus(){
...
+						MAX_LENGTH="${#PRE}"
+					fi
+				done < "${LOG_DIRECTORY}/status-check.txt"
+
+				# then we print them out, spacing them based on that length
+				while read -r LINE; do
+					SERVICE="$(echo "$LINE" | cut -d= -f2)"
+					STATUS="$(echo "$LINE" | cut -d= -f3- | tr '[:upper:]' '[:lower:]')"
+					printf "OpenNMS.%-${MAX_LENGTH}s : %s\n" "$SERVICE" "$STATUS"
+				done < "${LOG_DIRECTORY}/status-check.txt"
```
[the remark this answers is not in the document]
RangerRick: I'm working under the assumption that this script is not necessarily the only thing wanting to read status checks.  Perhaps I should modify the tools to write it whenever any operation is invoked.

core/daemon/src/main/java/org/opennms/netmgt/vmmgr/Invoker.java
RangerRick: true; I originally avoided it to just dump the data simply, but might as well format it when it's easily parseable in the other style

opennms-base-assembly/src/main/filtered/bin/opennms
amay94: Below this, you refer to '${LOG_DIRECTORY}/pre-check.txt' when checking the timestamp if I'm reading it correctly.

RangerRick:
I'd love for everyone to give this another look. I think I've got it pretty well set up now, and I documented the heck out of why I did things the way I did in the `opennms` script, so hopefully I didn't miss anything.

opennms-base-assembly/src/main/filtered/bin/opennms
```diff
@@ -698,7 +790,7 @@ JAVA_CMD=("$OPENNMS_HOME/bin/runjava" -r "${RUNJAVA_OPTIONS[@]}" --)
 BOOTSTRAP="$OPENNMS_HOME/lib/opennms_bootstrap.jar"
 
 MANAGER_OPTIONS=()
-#MANAGER_OPTIONS+=("-DOPENNMSLAUNCH")
+MANAGER_OPTIONS+=("-D_OPENNMS_LAUNCH")
```
amay94, line 793: did the old version of this option have any functionality tied to it?

opennms-base-assembly/src/main/filtered/bin/opennms
```diff
@@ -173,14 +174,28 @@ isRunning() {
...
 		return
 	fi
+
+	echo -1
+}
+
+findPid() {
+	local __matched_pid;
+
+	__matched_pid="$(pgrep -f _OPENNMS_LAUNCH 2>/dev/null || :)"
```
amay94, line 193: what is that final colon doing?

amay94 (approved):
Everything looks good to me

opennms-base-assembly/src/main/filtered/bin/opennms
RangerRick: "or do nothing" so that `set -e` doesn't short circuit

equivalent of doing `|| true` but it's an actual no-op rather than a "true" value

PR #4209, opened 2022-02-08
NMS-13783: Systemd startup uses legacy SysV init script

freddy61025:
https://issues.opennms.org/browse/NMS-13783

It also disable create symlink if init.d not exist.
It have been tested on CentOS 9 & 7.9.2009.

### DIFF
--- a/tools/packages/opennms/opennms.spec
+++ b/tools/packages/opennms/opennms.spec
@@ -546,7 +546,7 @@ echo "=== BUILDING ASSEMBLIES ==="
 	-Dinstall.version="%{version}-%{release}" \
 	-Ddist.name="%{name}-%{version}-%{release}.%{_arch}" \
 	-Dopennms.home="%{instprefix}" \
-	-Dinstall.init.dir="/etc/init.d" \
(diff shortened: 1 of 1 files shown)


RangerRick:
I feel like this is not the right approach, mostly for the fact that `$OPENNMS_HOME/bin` _isn't_ really an init dir.

Maybe it would be better to just get rid of `${install.init.dir}` entirely, and just reference `${install.bin.dir}` directly in the places we care about such things? In that case the `$INITDIR` in `opennms-base-assembly/src/main/filtered/bin/opennms` seems pretty pointless and can get removed altogether, too.

What do you think?

freddy61025:
Sure, I am happy to clean that. My intension is just to minimize the change. Since "-Dinstall.init.dir=/usr/share/opennms/bin" is also in ./debian/rules.

RangerRick:
> Sure, I am happy to clean that. My intension is just to minimize the change. Since "-Dinstall.init.dir=/usr/share/opennms/bin" is also in ./debian/rules.

lol, yeah, it probably shouldn't be  :D

PR #4214, opened 2022-02-09
CLOUD-475 Push signed containers to Azure CR

jb68:
### External References

* JIRA (Issue Tracker): https://issues.opennms.org/browse/CLOUD-475

Successful build, push and sign on both Azure and Docker:
https://app.circleci.com/pipelines/github/OpenNMS/opennms/16501/workflows/aef53f6a-d5a0-411e-a8d7-ceb3640bd921/jobs/114383

Please not that azurecr needs to be deployed before we can merge this commit

### DIFF
--- a/.circleci/main/@main.yml
+++ b/.circleci/main/@main.yml
@@ -1092,7 +1092,7 @@ jobs:
             # TODO function for duplication ?
             # AZURE notary
 
-            NOTARY_SERVER="https://openmspubacr.azureacr.io"
+            NOTARY_SERVER="https://opennmspubacr.azurecr.io"
             export NOTARY_AUTH="$(printf "${AZURE_SP}:${AZURE_SP_PASSWORD}" | base64 -w0)"
             export NOTARY_TARGETS_PASSPHRASE="${AZURE_DCT_CI_PASSPHRASE}"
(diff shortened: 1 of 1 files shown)


.circleci/main/@main.yml
```diff
@@ -565,7 +580,9 @@ workflows:
           matrix:
             parameters:
               architecture: [linux/amd64,linux/arm64,linux/arm/v7]
-          context: "docker-content-trust"
+          context:
+            - docker-content-trust
+            - Service Principal
```
swachter, line 585: I would prefer a context name without a space.

swachter (approved):
LGTM.

The only thing is that I would use a context name without a space. In addition, it may help if the context name `Service Prinicipal` would indicate that it is related to Azure.

.circleci/main/@main.yml
jb68: Service Principal context was there, it also is used on cloud repos so I didn't came up with the name.
Service Principal is the Azure name for a system user / account. 
I can change it to azure-cr-trust to make inline with the other context

RangerRick: no specific opinion, really, tho' I agree that having a space in there feels weird...

if other things are relying on it, it's up to you whether you want to go through the trouble of renaming it everywhere else it's used

PR #4219, opened 2022-02-10
NMS-13737: In default installation the ActiveMQ Total Enqueued Messages throw divide error exceptions

freddy61025:
https://issues.opennms.org/browse/NMS-13737

It is a minor change that if (EnqueCnt + DequeCnt = 0) it simply return 0.


dino2gnt (approved):
I had no idea that you could do that in a threshold expression.  That's awesome! There's probably multiple expressions where we could use this same solution.

PR #4228, opened 2022-02-10
NMS-13801: Create a default poller config for DeviceConfigBackup

cgorantla:
Fix some wiring on OpenNMS/Minion for Device Config Monitor

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13801


features/minion/repository/src/main/resources/features.boot
```diff
@@ -11,6 +11,7 @@ opennms-trapd-listener
 opennms-events-sink-dispatcher
 opennms-send-event-command
 opennms-dnsresolver-netty
+opennms-deviceconfig
```
swachter, line 14: Should the `opennms-deviceconfig` feature really be enabled by default on Minion?

We also have to discuss the different features we need. I think there should not be a general feature `opennms-deviceconfig`. We have poller/sink and server/client related bundles. I did some "featuring" in branch `jira/NMS-13937`.

opennms-base-assembly/src/main/filtered/etc/poller-configuration.xml
```diff
@@ -338,6 +338,33 @@
...
+         <rra>RRA:AVERAGE:0.5:288:366</rra>
+         <rra>RRA:MAX:0.5:288:366</rra>
+         <rra>RRA:MIN:0.5:288:366</rra>
+      </rrd>
+      <service name="DeviceConfig-Default" interval="30000" user-defined="false" status="on">
+         <parameter key="retry" value="3"/>
+         <parameter key="timeout" value="3000"/>
+         <parameter key="rrd-repository" value="${install.share.dir}/rrd/response"/>
+         <parameter key="rrd-base-name" value="device-config"/>
+         <parameter key="config-type" value="Default"/>
```
swachter, line 356: Maybe we should add comments on the other necessary parameters here:
* script
* user
* password

BTW: How would we configure a multi-line script. The line breaks are important.

Having the script and credentials in the service configuration means that there may be more than one service definition for a certain config type. Is the mapping configType -> service unambiguous at least for a certain IP? Otherwise we have to rethink how the service is selected in case of a manual trigger.

swachter (commented):
Looks great! However, I think we have to clarify some issues.

opennms-base-assembly/src/main/filtered/etc/poller-configuration.xml
cgorantla: If we have multi-line script, we should provide file path for that script. 
Since other parameters will mostly come from metadata, I haven't added anything but we can definitely add comment here.

features/minion/repository/src/main/resources/features.boot
cgorantla: yeah, I realized all the monitors need to be enabled by default on Minion.

opennms-base-assembly/src/main/filtered/etc/poller-configuration.xml
swachter: Where would that file path point to an absolute path on the file system?

swachter: For adding the script right into the service configuration would could also use indexed keys, e.g.:
```
         <parameter key="script.1" value="send: ..."/>
         <parameter key="script.2" value="send: atftp --put --local-file somefile.cfg --remote-file ${filename} ${tftpServerIp} ${tftpServerPort}"
         <parameter key="script.3" value="await: ..."/>
```

features/minion/repository/src/main/resources/features.boot
swachter: WDYT on the proposed renames, then?

cgorantla: `opennms-deviceconfig-monitor` for the Monitor and `opennms-deviceconfig-sink` for the Sink

opennms-base-assembly/src/main/filtered/etc/poller-configuration.xml
cgorantla: > For adding the script right into the service configuration would could also use indexed keys, e.g.:
> 
> ```
>          <parameter key="script.1" value="send: ..."/>
>          <parameter key="script.2" value="send: atftp --put --local-file somefile.cfg --remote-file ${filename} ${tftpServerIp} ${tftpServerPort}"
>          <parameter key="script.3" value="await: ..."/>
> ```

It's easier to configure a filename in the UI. But we may have to parse that script and load it in an object before sending it to Minion since minion doesn't have access to files on OpenNMS.

opennms-base-assembly/src/main/filtered/etc/poller-configuration.xml
```diff
@@ -338,6 +338,27 @@
...
+         <rra>RRA:MAX:0.5:288:366</rra>
+         <rra>RRA:MIN:0.5:288:366</rra>
+      </rrd>
+      <service name="DeviceConfig-Default" interval="30000" user-defined="false" status="on">
+         <parameter key="retry" value="3"/>
+         <parameter key="timeout" value="3000"/>
+         <parameter key="rrd-repository" value="${install.share.dir}/rrd/response"/>
+         <parameter key="rrd-base-name" value="device-config"/>
+         <!-- match service name suffix with config-type -->
+         <parameter key="config-type" value="${requisition:config-type|Default}"/>
```
swachter: I would like to see the other configuration parameter here, namely: `script`, `username`, `password`, `port`, and `timeout` (cf. `DeviceConfigMonitor`).

swachter (approved):
Just two more comments:
* I would like to see the additional parameters documented in the example configuration.
* Depending on the merge sequence the `Default` in `poller-configuration.xml` has to be changed.

swachter (approved):
LGTM. Still open the script configuration. I assume this will be done separately.

cgorantla:
Karaf ssh fails to install which prevents smoke tests to access karaf shell.  I think this is unrelated to the changes here, debugging.

```
2022-02-15T21:15:54,835 | WARN  | activator-1-thread-1 | Activator                        | 62 - org.apache.karaf.shell.ssh - 4.3.6 | Error starting activator
java.lang.ClassCastException: Cannot cast org.apache.sshd.common.util.security.eddsa.EdDSASecurityProviderRegistrar to org.apache.sshd.common.util.security.SecurityProviderRegistrar
	at java.lang.Class.cast(Class.java:3605) ~[?:?]
	at org.apache.sshd.common.util.ReflectionUtils.newInstance(ReflectionUtils.java:74) ~[!/:2.8.0]
	at org.apache.sshd.common.util.threads.ThreadUtils.createDefaultInstance(ThreadUtils.java:104) ~[!/:2.8.0]
	at org.apache.sshd.common.util.threads.ThreadUtils.createDefaultInstance(ThreadUtils.java:112) ~[!/:2.8.0]
	at org.apache.sshd.common.util.threads.ThreadUtils.createDefaultInstance(ThreadUtils.java:97) ~[!/:2.8.0]
	at org.apache.sshd.common.util.security.SecurityUtils.register(SecurityUtils.java:430) ~[!/:2.8.0]
	at org.apache.sshd.common.util.security.SecurityUtils.isBouncyCastleRegistered(SecurityUtils.java:395) ~[!/:2.8.0]
	at org.apache.sshd.common.util.security.SecurityUtils.getRandomFactory(SecurityUtils.java:548) ~[!/:2.8.0]
	at org.apache.sshd.common.BaseBuilder.fillWithDefaultValues(BaseBuilder.java:166) ~[!/:2.8.0]
	at org.apache.sshd.server.ServerBuilder.fillWithDefaultValues(ServerBuilder.java:108) ~[!/:2.8.0]
	at org.apache.sshd.server.ServerBuilder.fillWithDefaultValues(ServerBuilder.java:58) ~[!/:2.8.0]
	at org.apache.sshd.common.BaseBuilder.build(BaseBuilder.java:273) ~[!/:2.8.0]
	at org.apache.sshd.server.ServerBuilder.build(ServerBuilder.java:151) ~[!/:2.8.0]
	at org.apache.sshd.server.ServerBuilder.build(ServerBuilder.java:58) ~[!/:2.8.0]
	at org.apache.sshd.common.BaseBuilder.build(BaseBuilder.java:297) ~[!/:2.8.0]
	at org.apache.sshd.server.SshServer.setUpDefaultServer(SshServer.java:442) ~[!/:2.8.0]
	at org.apache.karaf.shell.ssh.Activator.createSshServer(Activator.java:183) ~[!/:?]
	at org.apache.karaf.shell.ssh.Activator.createAndRunSshServer(Activator.java:121) ~[!/:?]
	at org.apache.karaf.shell.ssh.Activator.doStart(Activator.java:116) ~[!/:?]
	at org.apache.karaf.util.tracker.BaseActivator.run(BaseActivator.java:312) [!/:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515) [?:?]
	at java.util.concurrent.FutureTask.run(FutureTask.java:264) [?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) [?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) [?:?]
	at java.lang.Thread.run(Thread.java:829) [?:?]
```

PR #4239, opened 2022-02-12
NMS-13835: Fix reflected XSS

geraldhumphries:
Jira: https://issues.opennms.org/browse/NMS-13835

Some notes:

1. A lot of the detected XSS vulnerabilities were actually from exception messages, and currently are not interpreted by the browser as HTML. Right now this means they aren't actually vulnerabilities, but they will be detected by pentest tools as XSS vulns regardless. If the error messages are shown to the user without being sanitized, they could trigger script execution.  I solved this by sanitizing the user's input for fields like `limit`, `orderBy`, etc. This has side effects, are they acceptable?
- The `sanitizeInteger` method strips out any non-numeric characters, so an input of `limit=<script>alert(1)</script>` will be interpreted as `1` and proceed without an error.
- Error messages will now show sanitized strings. For example: `
Severity ' &lt;script&gt;alert(document.cookie)&lt;/script&gt;' is not valid. Supported values are: [INDETERMINATE, CLEARED, NORMAL, WARNING, MINOR, MAJOR, CRITICAL]`
2. The usage of both `forJavascript` and `escapedName` meant that the notification names were being sanitized twice. This caused an issue where if a notification had any sanitized characters, the notification would be impossible to edit/delete/update. Fixed this by using the name field, and running it through the forJavascript sanitizer once. I removed the escapedName field as it is now unused.

### DIFF
--- a/features/geolocation/rest/src/main/java/org/opennms/web/rest/v2/GeolocationRestService.java
+++ b/features/geolocation/rest/src/main/java/org/opennms/web/rest/v2/GeolocationRestService.java
@@ -1,8 +1,8 @@
 /*******************************************************************************
  * This file is part of OpenNMS(R).
  *
- * Copyright (C) 2017-2017 The OpenNMS Group, Inc.
- * OpenNMS(R) is Copyright (C) 1999-2017 The OpenNMS Group, Inc.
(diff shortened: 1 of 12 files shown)


christianpape (approved):
Looks good to me.

PR #4244, opened 2022-02-14
NMS-14002: Resolve SonarCloud High priority Security Hotspots

amay94:
The sanitation in MailerServlet is a place holder until it is replaced with another mail library, see [NMS 14015](https://issues.opennms.org/browse/NMS-14015).

Updates to use PreparedStatements are fairly self-explanatory.

For the two Queries, parameter binding is not allowed for column names. To sanitize these, I used Hibernate to get a list of acceptable column names and compared the values against it.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14002


opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/OutageDaoHibernate.java
```diff
@@ -30,23 +30,19 @@
 
 import java.net.InetAddress;
 import java.sql.SQLException;
-import java.util.Collection;
-import java.util.Date;
-import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
christianpape, line 33: Do not use wildcard imports.

christianpape (changes requested):

PR #4246, opened 2022-02-14
NMS-13822: Start node "availability" graphic from the node's createTime

geraldhumphries:
- Change the "drawGreen" method to draw from where the node's createTime is, instead of just filling up the entire graphic
- Rename methods and variables to make them easier to understand

Note: If an outage exists before a node's createTime, it will appear in the non-green part of the graphic. Is this something that can happen?

Issue: https://issues.opennms.org/browse/NMS-13822

### DIFF
--- a/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/TimelineRestService.java
+++ b/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/TimelineRestService.java
@@ -53,7 +53,9 @@ import javax.ws.rs.core.UriInfo;
 import org.opennms.core.criteria.CriteriaBuilder;
 import org.opennms.core.criteria.restrictions.Restrictions;
 import org.opennms.core.utils.InetAddressUtils;
+import org.opennms.netmgt.dao.api.NodeDao;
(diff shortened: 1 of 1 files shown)


opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/TimelineRestService.java
```diff
@@ -366,6 +378,13 @@ private OnmsOutageCollection queryOutages(final UriInfo uriInfo, final int nodeI
         return onmsOutageCollection;
     }
 
+    @Autowired
+    private NodeDao m_nodeDao;
+
+    private OnmsNode queryNode(final int nodeId) {
```
geraldhumphries, line 384: This follows the same pattern as the `queryOutages` method above, but it's very simple so `image()` could just do `m_nodeDao.get()` directly. Thoughts?

RangerRick: yeah, seems a bit silly to have the indirection

RangerRick (dismissed):
seems pretty straightforward to me +1

cgorantla (commented):
LGTM!

PR #4251, opened 2022-02-15
NMS-13890: Fix modifying node/interface/service metadata through requisition after initial synchronization

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-13890


dino2gnt:
Can you talk me through the end result of this change? 

What will the behavior be in the event metadata is added via some other path (provisioning adapter, vmware integration, etc) and the requisition, containing metadata, is synchronized? 

What happens when requisitioned metadata is changed?  What happens when it is removed?

christianpape:
> Can you talk me through the end result of this change?
> 
> What will the behavior be in the event metadata is added via some other path (provisioning adapter, vmware integration, etc) and the requisition, containing metadata, is synchronized?
> 
> What happens when requisitioned metadata is changed? What happens when it is removed?

Maybe the PR title is a bit misleading. We restored the original behaviour of creating metadata. The problem was, that there were leftovers from times, where we needed to preserve values that were added through the ReST API.

Provisioning via VMware and the adapters are of course able to add metadata. If a user wants to add/change/delete metadata, he can only do it via the requisition UI or by altering the requisitions XML files.

PR #4258, opened 2022-02-15
NMS-13970: DeviceConfigBackup REST API with Schedule info

synqotik:
NMS-13970: DeviceConfigBackup REST API with Schedule info

Initial version of updated REST API that gets device config backup schedule info from Service metadata. Actual parsing of cron-style schedule pattern data will occur in subsequent PR.

'DeviceConfig' entries are retrieved as per sorting/filtering parameters, then the corresponding service schedule data is retrieved and mapped to the DTOs.

A separate integration test for the schedule part only was created as the DB setup is significantly different than the IT for the sort/filter part of the REST service.

Jira: https://issues.opennms.org/browse/NMS-13970

### DIFF
--- a/features/device-config/persistence/api/src/main/java/org/opennms/features/deviceconfig/persistence/api/DeviceConfig.java
+++ b/features/device-config/persistence/api/src/main/java/org/opennms/features/deviceconfig/persistence/api/DeviceConfig.java
@@ -35,8 +35,6 @@ import java.util.Objects;
 
 import javax.persistence.Column;
 import javax.persistence.Entity;
-import javax.persistence.EnumType;
-import javax.persistence.Enumerated;
 import javax.persistence.FetchType;
 import javax.persistence.GeneratedValue;
 import javax.persistence.GenerationType;
@@ -76,6 +74,9 @@ public class DeviceConfig implements Serializable {
 
     @Column(name = "config_type", nullable = false)
     private String configType;
+
+    @Column(name = "filename")
+    private String fileName;
 
     @Column(name = "failure_reason")
     private String failureReason;
@@ -140,9 +141,13 @@ public class DeviceConfig implements Serializable {
         return configType;
     }
 
+    public String getFileName() { return fileName; }
+
     public void setConfigType(String configType) {
         this.configType = configType;
     }
+
+    public void setFileName(String fileName) { this.fileName = fileName; }
 
     public String getFailureReason() {
         return failureReason;
@@ -181,12 +186,22 @@ public class DeviceConfig implements Serializable {
         if (this == o) return true;
         if (!(o instanceof DeviceConfig)) return false;
         DeviceConfig that = (DeviceConfig) o;
-        return ipInterface.equals(that.ipInterface) && Arrays.equals(config, that.config) && encoding.equals(that.encoding) && configType == that.configType && Objects.equals(failureReason, that.failureReason) && Objects.equals(createdTime, that.createdTime) && lastUpdated.equals(that.lastUpdated) && Objects.equals(lastFailed, that.lastFailed) && Objects.equals(lastSucceeded, that.lastSucceeded);
+        return
+            ipInterface.equals(that.ipInterface) &&
+            Arrays.equals(config, that.config) &&
+            encoding.equals(that.encoding) &&
+            configType == that.configType &&
+            fileName == that.fileName &&
+            Objects.equals(failureReason, that.failureReason) &&
+            Objects.equals(createdTime, that.createdTime) &&
+            lastUpdated.equals(that.lastUpdated) &&
+            Objects.equals(lastFailed, that.lastFailed) &&
+            Objects.equals(lastSucceeded, that.lastSucceeded);
     }
 
     @Override
     public int hashCode() {
-        int result = Objects.hash(ipInterface, encoding, configType, failureReason, createdTime, lastUpdated, lastFailed, lastSucceeded);
+        int result = Objects.hash(ipInterface, encoding, configType, fileName, failureReason, createdTime, lastUpdated, lastFailed, lastSucceeded);
         result = 31 * result + Arrays.hashCode(config);
         return result;
     }
--- a/features/device-config/persistence/api/src/main/java/org/opennms/features/deviceconfig/persistence/api/DeviceConfig.java
+++ b/features/device-config/persistence/api/src/main/java/org/opennms/features/deviceconfig/persistence/api/DeviceConfig.java
@@ -184,19 +184,9 @@ public class DeviceConfig implements Serializable {
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
-        if (!(o instanceof DeviceConfig)) return false;
+        if (o == null || getClass() != o.getClass()) return false;
         DeviceConfig that = (DeviceConfig) o;
-        return
-            ipInterface.equals(that.ipInterface) &&
-            Arrays.equals(config, that.config) &&
-            encoding.equals(that.encoding) &&
-            configType == that.configType &&
-            fileName == that.fileName &&
-            Objects.equals(failureReason, that.failureReason) &&
-            Objects.equals(createdTime, that.createdTime) &&
-            lastUpdated.equals(that.lastUpdated) &&
-            Objects.equals(lastFailed, that.lastFailed) &&
-            Objects.equals(lastSucceeded, that.lastSucceeded);
+        return Objects.equals(createdTime, that.createdTime) && Objects.equals(lastUpdated, that.lastUpdated) && Objects.equals(lastFailed, that.lastFailed) && Objects.equals(lastSucceeded, that.lastSucceeded) && Objects.equals(encoding, that.encoding) && Objects.equals(configType, that.configType) && Objects.equals(fileName, that.fileName) && Objects.equals(failureReason, that.failureReason) && Objects.equals(ipInterface, that.ipInterface) && Arrays.equals(config, that.config);
     }
 
     @Override
--- a/features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/api/DeviceConfigDTO.java
+++ b/features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/api/DeviceConfigDTO.java
@@ -55,17 +55,17 @@ public class DeviceConfigDTO {
     /**
      * Date backup last succeeded, or null if no backups succeeded.
      * If the backup process succeeded but the configuration did not change, this date could be
-     * more recent than 'configStoredDate'.
+     * more recent than 'createdTime'.
      */
     private final Date lastSucceededDate;
 
     /** Date backup last failed, or null if no backups failed. */
     private final Date lastFailedDate;
 
-    /** Encoding of the config, either 'text' or 'binary'. Other values may include 'json' or 'xml'. */
+    /** Encoding of the config, possible values include 'text', 'binary', 'ascii', 'json', 'xml'. */
     private final String encoding;
 
-    /** The config type as a String, either 'Default' or 'Running'. */
+    /** The device configuration type, either 'default' or 'running'. */
     private final String configType;
 
     /** Filename of the configuration data as received from the device. */
--- a/features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/api/DeviceConfigDTO.java
+++ b/features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/api/DeviceConfigDTO.java
@@ -30,40 +30,160 @@ package org.opennms.features.deviceconfig.rest.api;
 
 import java.util.Date;
 
+/**
+ * DTO for response value of DeviceConfigRestService.
+ * 'final' properties are retrieved directly from 'device_config' table.
+ * Other properties are calculated or from other data sources.
+ * The actual configuration data is retrieved separately using this object's 'id' field.
+ */
 public class DeviceConfigDTO {
-
+    /** Database id of 'device_config' table entry */
     private final long id;
+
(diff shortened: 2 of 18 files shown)


features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestService.java
```diff
@@ -113,13 +202,71 @@ public void deleteDeviceConfig(long id) {
...
     private static DeviceConfigDTO deviceConfigDto(DeviceConfig deviceConfig) {
         return new DeviceConfigDTO(
-                deviceConfig.getId(),
-                deviceConfig.getIpInterface().getId(),
-                deviceConfig.getConfig(),
-                deviceConfig.getEncoding(),
-                deviceConfig.getCreatedTime()
+            deviceConfig.getId(),
+            deviceConfig.getIpInterface().getId(),
+            InetAddressUtils.str(deviceConfig.getIpInterface().getIpAddress()),
```
swachter, line 262: We should check, if Hibernate does another select for each `IpInterface` separately. Maybe some `eager` annotation is necessary.

features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestService.java
```diff
@@ -56,42 +88,35 @@ public Response getDeviceConfigs(
             String orderBy,
             String order,
             Integer ipInterfaceId,
+            String configType,
             Long createdAfter,
             Long createdBefore
     ) {
-        var criteriaBuilder = new CriteriaBuilder(DeviceConfig.class);
+        var criteria = getCriteria(limit, offset, orderBy, order, ipInterfaceId, configType, createdAfter, createdBefore);
```
swachter, line 95: Maybe we need another kind of query that groups by (ipAddressId, configType) and select the maximum of each group by the lastUpdated field.

opennms-base-assembly/pom.xml
```diff
@@ -164,6 +164,10 @@
                 <groupId>eclipse</groupId>
                 <artifactId>jdtcore</artifactId>
               </exclusion>
+              <exclusion>
+                <groupId>com.lowagie</groupId>
+                <artifactId>itext</artifactId>
+              </exclusion>
```
swachter, line 170: That seems subtle. Where did this come from?

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/MonitoredServiceDaoHibernate.java
```diff
@@ -151,4 +152,16 @@ public OnmsMonitoredService get(Integer nodeId, InetAddress ipAddr, Integer ifIn
...
     }
+
+    @Override
+    public List<OnmsMonitoredService> findByServiceTypeAndIpInterfaceId(String serviceTypePrefix, List<Integer> ipInterfaceIds) {
+        var idParam = ipInterfaceIds.stream().map(String::valueOf).collect(Collectors.joining(","));
+
+        String query =
+            "from OnmsMonitoredService svc " +
+                "where svc.serviceType.name like '" + serviceTypePrefix + "%' " +
+                "and svc.ipInterface.id in (" + idParam + ")";
```
swachter, line 163: Use query parameters, to avoid SQL injection.

opennms-base-assembly/pom.xml
synqotik: I encountered issues getting this library (transitive dependency of jasper reports) due to `maven-default-http-blocker` blocking http requests. I believe this forces Maven to use a different repository to get this artifact, i.e. one of the https repositories. However could just be due to issues we were having with the OpenNMS repository. We could see if this gets resolved and remove this entry eventually.

features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestService.java
synqotik: I agree - I'm just not quite sure what data we exactly need, but hope that can be clarified soon (i.e. in next few days or week).

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/MonitoredServiceDaoHibernate.java
synqotik: Query parameters for 'in' clause did not seem to work as expected, however, I can take another look. I agree we should be using query parameters.

features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestService.java
synqotik: Good point, will take a look. Also could be resolved by doing a query as you suggested below.

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/MonitoredServiceDaoHibernate.java
swachter: An alternative would be to use Hibernate's criteria queries.

opennms-base-assembly/pom.xml
swachter: We should definitely check, in order not to break something somewhere else.

synqotik: I will remove this and see if things build on CI server.

features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/api/DeviceConfigDTO.java
```diff
@@ -30,40 +30,125 @@
...
+    private final String encoding;
+
+    /** The config type as a String, either 'Default' or 'Running'. */
+    private final String configType;
+
+    /** Failure reason description, if there was a backup failure. */
+    private final String failureReason;
+
+    /** Device name */
+    private String deviceName;
```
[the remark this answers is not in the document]
synqotik: As per discussion today, looks like from `node.nodelabel`. Looking into refactoring the DB call to get this and some other data.

features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestService.java
synqotik: May just need a different query, to get more data from node and ipinterface tables, am working on this.

opennms-base-assembly/pom.xml
RangerRick: there was an outage on the upstream maven repository, it _should_ be OK now, but if you still have problems let me know and I'll dig into it more

synqotik: Thanks will do!

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/MonitoredServiceDaoHibernate.java
synqotik: Rewritten to use CriteriaBuilder, see latest version.

opennms-base-assembly/pom.xml
synqotik: Removed this

features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestService.java
synqotik: It does look like Hibernate makes separate select calls for each `IpInterface`. I will test with `EAGER` annotation. We will almost certainly end up making additional changes once UX is more finalized and we do more testing, so may make sense to wait until then to make more changes.

features/device-config/persistence/api/src/main/java/org/opennms/features/deviceconfig/persistence/api/DeviceConfig.java
```diff
@@ -181,12 +186,22 @@ public boolean equals(Object o) {
         if (this == o) return true;
         if (!(o instanceof DeviceConfig)) return false;
         DeviceConfig that = (DeviceConfig) o;
-        return ipInterface.equals(that.ipInterface) && Arrays.equals(config, that.config) && encoding.equals(that.encoding) && configType == that.configType && Objects.equals(failureReason, that.failureReason) && Objects.equals(createdTime, that.createdTime) && lastUpdated.equals(that.lastUpdated) && Objects.equals(lastFailed, that.lastFailed) && Objects.equals(lastSucceeded, that.lastSucceeded);
+        return
```
cgorantla, line 189: If you use IntelliJ default generation, changes wouldn't look much different

features/device-config/rest/src/test/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestServiceIT.java
```diff
@@ -35,10 +35,7 @@
 import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.Matchers.is;
 
-import java.util.Collections;
-import java.util.Date;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
cgorantla, line 38: avoid wildcard imports

features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/api/DeviceConfigDTO.java
```diff
@@ -30,40 +30,160 @@
...
+     */
+    private final Date lastSucceededDate;
+
+    /** Date backup last failed, or null if no backups failed. */
+    private final Date lastFailedDate;
+
+    /** Encoding of the config, either 'text' or 'binary'. Other values may include 'json' or 'xml'. */
+    private final String encoding;
+
+    /** The config type as a String, either 'Default' or 'Running'. */
```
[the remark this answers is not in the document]
synqotik: Fixed in latest commit

features/device-config/persistence/api/src/main/java/org/opennms/features/deviceconfig/persistence/api/DeviceConfig.java
synqotik: Done

features/device-config/rest/src/test/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestServiceIT.java
synqotik: Fixed

features/device-config/persistence/api/src/main/java/org/opennms/features/deviceconfig/persistence/api/DeviceConfig.java
synqotik: Updated to remove `id` field, fixed ordering so that `ipInterface` and `config` are done last, so usually they won't need to be checked (expensive check).

cgorantla (approved):

features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestService.java
```diff
@@ -102,24 +135,188 @@ public Response getDeviceConfigs(
...
+        final var matchingService =
+            serviceList.stream()
+            .filter(x -> x.getServiceName().equalsIgnoreCase(serviceName))
+            .findFirst();
+
+        if (matchingService.isPresent()) {
+            return getScheduleMetaDataValue(matchingService.get().getMetaData());
+        }
+
+        return Optional.empty();
```
swachter, line 181:
```suggestion
        return matchingService.flatMap(service -> getScheduleMetaDataValue(service.getMetaData());
```

features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestService.java
```diff
@@ -28,25 +28,63 @@
...
 import org.opennms.web.utils.ResponseUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class DefaultDeviceConfigRestService implements DeviceConfigRestService {
+    private static final Logger LOG = LoggerFactory.getLogger(DefaultDeviceConfigRestService.class);
+    private static final String REQUISITION_CONTEXT = "requisition";
+    private static final String SCHEDULE_METADATA_KEY = "schedule";
+
+    private static final Set<String> VALID_ORDER_BY_KEYS = Set.of("lastupdated", "devicename", "createdtime", "ipaddress");
```
swachter, line 66: Maybe you could use a map: orderBy -> (hql property) here. Then the check if an `orderBy` is supported and its translation would be easy to implement.

swachter (approved):
LGTM! Two minor, things only.

features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestService.java
synqotik: I will make this change in subsequent PR

PR #4260, opened 2022-02-15
NMS-13947 Cache node artifacts

jb68:
### External References

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-13947

Result from build

```
Found a cache from build 116381 at nodejs-dependencies-v3-TeS0v4mbHi8haIv8yCeudjEOHL0qYe5eJEWhFZc31QU=-+2yB_AKHtGBvqsbqWIq9Np3cJU10qRpnWpDWQSeX2WU=
Size: 191 MiB
Cached paths:
  * /root/.npm/_cacache
  * /root/project/core/web-assets/node_modules
```

### DIFF
--- a/.circleci/README.md
+++ b/.circleci/README.md
@@ -53,12 +53,9 @@ docker image load -i sentinel.oci
 
 ### Cache node artifacts
 
-We're are caching `node_modules` in `core/web-assets` with
-`npm --prefer-offline --no-progress install`.
-In conjunction with this we added caching of the `~/.npm/_cacache`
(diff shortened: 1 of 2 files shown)


RangerRick:
Can we change it to use `ci` instead of `install` and _only_ cache the cacache stuff? Otherwise we're double-caching if we're caching `node_modules`, and we're not really using the "production" process then.

RangerRick (approved):

RangerRick (approved):

PR #4263, opened 2022-02-16
NMS-13951:  Trigger alarms when config retrieval fails

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13951


cgorantla:
Added  commits from this in #4331

PR #4287, opened 2022-02-23
NMS-14031: Add Karaf shell command to retrieve device config manually

cgorantla:
Add karaf commands to trigger and retrieve device config manually

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14031


features/device-config/service/src/main/java/org/opennms/features/deviceconfig/service/impl/DeviceConfigServiceImpl.java
```diff
@@ -69,6 +74,32 @@
...
+            PollerResponse response = future.get(timeout, TimeUnit.MILLISECONDS);
+            if (response != null && response.getPollStatus() != null && response.getPollStatus().getDeviceConfig() != null) {
+                deviceConfigFuture.complete(response.getPollStatus().getDeviceConfig());
+            }
+        } catch (InterruptedException | ExecutionException | TimeoutException e) {
+            LOG.error("Error while getting device config for IpAddress {} at location {}", ipAddress, location, e);
+            deviceConfigFuture.completeExceptionally(e);
+        }
+        return deviceConfigFuture;
+    }
```
swachter, line 100: This implementation has some issues:
* A future is returned, but in effect the future is already completed when it is returned. Returning the `byte[]` would be the same.
* The returned future is not completed in case the condition `response != null && response.getPollStatus() != null && response.getPollStatus().getDeviceConfig() != null` is not satisfied.

I propose to rewrite the method:

```suggestion
    public CompletableFuture<byte[]> getDeviceConfig(String ipAddress, String location, String configType, int timeout) throws IOException {
        return pollDeviceConfig(ipAddress, location, configType)
                .orTimeout(timeout, TimeUnit.MILLISECONDS)
                .thenApply(resp -> resp.getPollStatus().getDeviceConfig())
                .whenComplete((bytes, throwable) -> {
                    if (throwable != null) {
                        LOG.error("Error while getting device config for IpAddress {} at location {}", ipAddress, location, throwable);
                    }
                });
    }
```

features/device-config/shell/src/main/java/org/opennms/features/deviceconfig/shell/DeviceConfigTrigger.java
```diff
@@ -0,0 +1,73 @@
...
+import org.apache.karaf.shell.api.action.Argument;
+import org.apache.karaf.shell.api.action.Command;
+import org.apache.karaf.shell.api.action.Option;
+import org.apache.karaf.shell.api.action.lifecycle.Reference;
+import org.apache.karaf.shell.api.action.lifecycle.Service;
+import org.opennms.features.deviceconfig.service.DeviceConfigService;
+
+import java.io.IOException;
+
+@Command(scope = "opennms", name = "device-config-trigger", description = "Trigger device config  backu[ from a specific Interface")
```
swachter, line 41: Minor typo. In addition, maybe something like `device-config-trigger-backup` or just `device-config-backup` would be clearer.

```suggestion
@Command(scope = "opennms", name = "device-config-trigger", description = "Trigger device config  backup from a specific Interface")
```

features/device-config/shell/src/main/java/org/opennms/features/deviceconfig/shell/GetDeviceConfig.java
```diff
@@ -0,0 +1,102 @@
...
+
+
+    @Override
+    public Object execute() throws Exception {
+        try {
+            InetAddress.getByName(host);
+        } catch (UnknownHostException e) {
+            System.out.printf("Not a valid host %s \n", host);
+            return null;
+        }
```
swachter, line 81: Why do we check the host here but not in `DeviceConfigTrigger`?

swachter (changes requested):
Looks good, overall. Some comments, though.

swachter (approved):
LGTM

PR #4293, opened 2022-02-24
NMS-14017: Return device config filename when polling

swachter:
Issue: https://issues.opennms.org/browse/NMS-14017

### DIFF
--- a/features/device-config/sink/consumer/src/main/java/org/opennms/features/deviceconfig/sink/consumer/DeviceConfigConsumer.java
+++ b/features/device-config/sink/consumer/src/main/java/org/opennms/features/deviceconfig/sink/consumer/DeviceConfigConsumer.java
@@ -78,9 +78,10 @@ public class DeviceConfigConsumer implements MessageConsumer<DeviceConfigDTO, De
             if (ipInterface != null) {
                 deviceConfigDao.updateDeviceConfigContent(
                         ipInterface,
-                        message.fileName,
+                        message.fileName, // use filename as config type
                         null,
-                        message.config
+                        message.config,
+                        message.fileName
                 );
             } else {
                 LOG.warn("can not persist device config; did not find interface - location: "+ message.location + "; " + address.getHostAddress());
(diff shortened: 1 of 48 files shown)


features/device-config/sink/consumer/src/main/java/org/opennms/features/deviceconfig/sink/consumer/DeviceConfigConsumer.java
```diff
@@ -78,9 +78,10 @@ public void handleMessage(DeviceConfigDTO message) {
             if (ipInterface != null) {
                 deviceConfigDao.updateDeviceConfigContent(
                         ipInterface,
-                        message.fileName,
+                        message.fileName, // use filename as config type
```
synqotik, line 81: `filename` was added to `DeviceConfig` object and database table, this has been merged to `feature/device-config`. Should get that code and add both `configType` and `fileName` here and elsewhere.

swachter: In case of sink, we have nothing else than the filename. Maybe we need some extra configuration that maps filenames into config types. Or at least, we could allow to configure the config type to use. You guys will figure out what to do!

synqotik: Good point! Still, just want to make sure that the filename does get saved to the new `filename` field in the database, regardless of what we do about figuring out the config type.

cgorantla: i think for Sink, we can make `config-type` as `default` for now and later we can check if we can derive it from filename itself.

features/poller/api/src/main/java/org/opennms/netmgt/poller/DeviceConfig.java
```diff
@@ -0,0 +1,39 @@
...
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.netmgt.poller;
+
+public class DeviceConfig {
+    public final byte[] content;
+    public final String filename;
+
```
cgorantla, line 34: This might need default constructor ?

features/device-config/sink/consumer/src/main/java/org/opennms/features/deviceconfig/sink/consumer/DeviceConfigConsumer.java
swachter: Use `ConfigType.Default` for now.

features/poller/api/src/main/java/org/opennms/netmgt/poller/DeviceConfig.java
cgorantla: Also had to add other xml annotations to properly get it from Minion

cgorantla (approved):

PR #4307, opened 2022-03-01
NMS-1396: Make TimeTrackingMonitor work

freddy61025:
https://issues.opennms.org/browse/NMS-13969

JmxReporter
image

There is an extra rest api original debug use. If it is not acceptable it is no problem to remove it. It do have some extra features that JmxReporter cannot give. E.g. showing current scanning Node and its start time.
Rest API ouput:
image


dino2gnt:
What is the significance of the numerical value after the timestamp?  `$foreignsourceName.$date.#######.$phase` ?

Keeping in mind I am not a Java developer and can only superficially follow the commit diff, do I understand correctly that we create a new set of metrics per provisiond job run, cache them for some amount of time, and expire them later?

RangerRick (approved):

freddy61025:
> not a Java developer and can only superficially follow the commit diff, do I understand correctly that we create a new set of metrics per provisiond job run, cache them for some amou

It just simply a java object id. It use to prevent in case same task fire at the same time. I think it will be very rare but better to avoid that.

freddy61025:
> What is the significance of the numerical value after the timestamp? `$foreignsourceName.$date.#######.$phase` ?
> 
> Keeping in mind I am not a Java developer and can only superficially follow the commit diff, do I understand correctly that we create a new set of metrics per provisiond job run, cache them for some amount of time, and expire them later?

Sorry missing one of your question. Yes it is per schedule job. Currently the info will keep in memory for 3 days after last update. And the monitor info will write to provisond.log when remove from memory.

PR #4331, opened 2022-03-02
NMS-13925:  Some updates to manual trigger and add events for Device Config 

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13925


features/device-config/monitor-adaptor/src/main/java/org/opennms/features/deviceconfig/monitor/adaptor/DeviceConfigMonitorAdaptor.java
```diff
@@ -96,10 +106,28 @@ public void setIpInterfaceDao(IpInterfaceDao ipInterfaceDao) {
...
+
     private String getObjectAsString(Object object) {
         if (object instanceof String) {
             return (String) object;
         }
         return null;
     }
+
+    private void sendConfigRetrievalFailedEvent(OnmsIpInterface ipInterface, String serviceName) {
+        EventBuilder bldr = new EventBuilder(EventConstants.DEVICE_CONFIG_RETRIEVAL_FAILED_UEI, "poller");
```
synqotik, line 121: These two methods could be 1 method with just the EventContants changed.

features/device-config/service/src/main/java/org/opennms/features/deviceconfig/service/impl/DeviceConfigServiceImpl.java
```diff
@@ -81,50 +97,98 @@ public void triggerConfigBackup(String ipAddress, String location, String config
...
             }
             OnmsNode node = ipInterface.getNode();
-
+            List<OnmsMonitoredService> services =
+                    monitoredServiceDao.findSimilarServicesOnInterface(node.getId(), ipInterface.getIpAddress(), DEVICE_CONFIG_SERVICE_NAME_PREFIX);
+            if (!services.isEmpty()) {
+                Optional<OnmsMonitoredService> optional =
+                        services.stream().filter(onmsMonitoredService -> onmsMonitoredService.getServiceName().contains(configTypeName)).findFirst();
+                if (optional.isPresent()) {
+                    return new SimpleMonitoredService(ipInterface.getIpAddress(), node.getId(), node.getLabel(), optional.get().getServiceName(), location);
```
synqotik, line 129: Could set `serviceName` here and then fall through to `return new` below as they are otherwise the same.

features/device-config/service/src/main/java/org/opennms/features/deviceconfig/service/impl/DeviceConfigServiceImpl.java
```diff
@@ -81,50 +97,98 @@ public void triggerConfigBackup(String ipAddress, String location, String config
...
+                    pkgNames.contains(pollerPackage.getName())).findFirst();
+            pkg.ifPresent(aPackage -> parameters.addAll(aPackage.getService(service.getSvcName()).getParameters()));
+        }
+        return parameters;
+    }
+
+    private Map<String, Object> convertParamsToAttributes(List<Parameter> parameters) {
+        Map<String, Object> attributes = new TreeMap<>();
+        if (parameters.isEmpty()) {
+            // If we couldn't load any specific pakage, use dummy parameters
```
synqotik, line 177: pakage -> package

features/poller/shell/src/main/java/org/opennms/netmgt/poller/shell/Poll.java
```diff
@@ -195,6 +202,16 @@ public Object execute() throws Exception {
                         } else {
                             System.out.printf("(No properties were returned by the monitor.\n");
                         }
+                        if (pollStatus.getDeviceConfig() != null) {
+                            DeviceConfig deviceConfig = pollStatus.getDeviceConfig();
+                            System.out.printf("Received file %s with content .. \n\n", deviceConfig.getFilename());
```
synqotik, line 207: `..` -> `...`

features/device-config/api/src/main/java/org/opennms/features/deviceconfig/service/DeviceConfigUtil.java
```diff
@@ -0,0 +1,55 @@
...
+ *******************************************************************************/
+
+package org.opennms.features.deviceconfig.service;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.zip.GZIPInputStream;
+
+public class DeviceConfigUtil {
```
synqotik, line 36: Does it make sense to make this a class + interface instead of a static class? I have 3 methods to add to this. Perhaps with interface it's easier to mock in unit tests.

synqotik (commented):
LGTM, just some small comments and one question/suggestion re DeviceConfigUtil.

features/device-config/api/src/main/java/org/opennms/features/deviceconfig/service/DeviceConfigUtil.java
synqotik: Per conversation, leaving as static methods.

features/device-config/shell/src/main/java/org/opennms/features/deviceconfig/shell/GetDeviceConfig.java
```diff
@@ -78,13 +84,20 @@ public Object execute() throws Exception {
...
         while (true) {
             try {
                 try {
-                    byte[] dc = future.get(1, TimeUnit.SECONDS);
-                    if (dc != null) {
-                        return new String(dc, Charset.forName(encoding));
+                    DeviceConfig deviceConfig = future.get(1, TimeUnit.SECONDS);
+                    if (deviceConfig != null) {
+                        System.out.printf("Received file %s with content .. \n\n", deviceConfig.getFilename());
+                        if (deviceConfig.getFilename().contains(".gz")) {
```
fooker, line 94: Use `endsWith` instead? Or maybe even use something like real MIME-type detection?

features/device-config/shell/src/main/java/org/opennms/features/deviceconfig/shell/GetDeviceConfig.java
```diff
@@ -78,13 +84,20 @@ public Object execute() throws Exception {
...
+                            // Decompress if this is compressed file
+                            byte[] dcBytes = DeviceConfigUtil.decompressGzipToBytes(deviceConfig.getContent());
```
fooker, line 96: Shouldn't his decompressing be part of the receiver?

features/poller/shell/src/main/java/org/opennms/netmgt/poller/shell/Poll.java
```diff
@@ -195,6 +202,16 @@ public Object execute() throws Exception {
...
+                            if (deviceConfig.getFilename().contains(".gz")) {
+                                // Decompress if this is compressed file
+                                byte[] dcBytes = DeviceConfigUtil.decompressGzipToBytes(deviceConfig.getContent());
```
fooker, line 210: Same here. We need this every time we read/display the content. This should be part of the `DeviceConfig` or even done on the receiver.

opennms-dao-mock/src/main/java/org/opennms/netmgt/dao/mock/MockMonitoredServiceDao.java
```diff
@@ -188,4 +188,15 @@ public OnmsMonitoredService getPrimaryService(final Integer nodeId, final String
     public List<OnmsMonitoredService> findByServiceTypeAndIpInterfaceId(String serviceTypePrefix, List<Integer> ipInterfaceIds) {
         throw new UnsupportedOperationException("Not yet implemented!");
     }
+
+    @Override
+    public List<OnmsMonitoredService> findSimilarServicesOnInterface(Integer nodeId, InetAddress ipAddress, String svcNamePrefix) {
```
fooker, line 193: Can we please do this by class name of the associated service instead of relying on a service name/prefix.

Main concern with this id, that it would kill the poller pattern feature, which would allow for a neat integration if different config types.

features/device-config/service/src/main/java/org/opennms/features/deviceconfig/service/impl/DeviceConfigServiceImpl.java
```diff
@@ -81,50 +97,98 @@ public void triggerConfigBackup(String ipAddress, String location, String config
...
+            parameters.addAll(pkg.getService(service.getSvcName()).getParameters());
+        } else {
+            List<String> pkgNames = pollerConfig.getAllPackageMatches(service.getIpAddr());
+            List<org.opennms.netmgt.config.poller.Package> packages =
+                    pollerConfig.getPackages().stream().filter(pollerPackage ->
+                            pollerConfig.isServiceInPackageAndEnabled(service.getSvcName(), pollerPackage)).collect(Collectors.toList());
+            Optional<org.opennms.netmgt.config.poller.Package> pkg = packages.stream().filter(pollerPackage ->
+                    pkgNames.contains(pollerPackage.getName())).findFirst();
+            pkg.ifPresent(aPackage -> parameters.addAll(aPackage.getService(service.getSvcName()).getParameters()));
+        }
```
fooker, line 170: I think, having the packagae defined in metadata is a bad idea. This is super hard to debug and works differently to what the real poller does. We should strictly mimic the behavior of `Poller::findPackageForService` (ideally by moving this to the `pollerConfig`),

features/device-config/service/src/main/java/org/opennms/features/deviceconfig/service/impl/DeviceConfigServiceImpl.java
```diff
@@ -81,50 +97,98 @@ public void triggerConfigBackup(String ipAddress, String location, String config
...
+                }
+            }
             return new SimpleMonitoredService(ipInterface.getIpAddress(), node.getId(), node.getLabel(), serviceName, location);
         });
 
         if (service == null) {
             throw new IllegalArgumentException("No interface found with ipAddress " + ipAddress + " at location " + location);
         }
+        List<Parameter> parameters = getServiceParamsFromPollerConfig(service);
+        Map<String, Object> serviceAttributes = convertParamsToAttributes(parameters);
```
fooker, line 139: Should this do metadata interpolation instead?

fooker (changes requested):

features/device-config/service/src/main/java/org/opennms/features/deviceconfig/service/impl/DeviceConfigServiceImpl.java
cgorantla: yeah, can't update servicename in the lambda expression. Other alternatives doesn't look better.

features/device-config/shell/src/main/java/org/opennms/features/deviceconfig/shell/GetDeviceConfig.java
cgorantla: Tried MIME-type detection on `.gz`,  `.gz` file parses as OctetStream which couldn't be any binary.  So using `endsWith` for now

features/device-config/service/src/main/java/org/opennms/features/deviceconfig/service/impl/DeviceConfigServiceImpl.java
cgorantla: This is only needed in very limited scenario where there are multiple packages with  `DeviceConfig` service that have overlapping IPAddresses that belong to different locations. There is no easy way to determine package in this case.

features/device-config/service/src/main/java/org/opennms/features/deviceconfig/service/impl/DeviceConfigServiceImpl.java
cgorantla: `PollerRequestBuilderImpl` does interpolation in execution.
 This only needs to retrieve service configuration from Poller configuration.

opennms-dao-mock/src/main/java/org/opennms/netmgt/dao/mock/MockMonitoredServiceDao.java
cgorantla: > Can we please do this by class name of the associated service instead of relying on a service name/prefix.
> 
> Main concern with this id, that it would kill the poller pattern feature, which would allow for a neat integration if different config types.

I checked how poller pattern feature and that seems to fit what we need here.
Let's chat on this.  Currently poller patterns only work with `pattern` context which is not available on UI.
When I tried  `config-type` with poller pattern, ran into below error, may be pattern doesn't support `-`
```
Error executing command: named capturing group is missing trailing '>' near index 22
^DevicConfig-(?<config-type>.*)$
```
Also need to see how to integrate this with manual poller execution (within DeviceConfigServiceImpl).

We can tackle using poller pattern for multiple services in another issue.

features/device-config/service/src/main/java/org/opennms/features/deviceconfig/service/impl/DeviceConfigServiceImpl.java
cgorantla: Since overlapping IpAddresses is not really handled in poller packages, removed code to retrieve package from metadata.

opennms-dao-mock/src/main/java/org/opennms/netmgt/dao/mock/MockMonitoredServiceDao.java
cgorantla: We can take care of this in another issue.

PR #4350, opened 2022-03-07
NMS-14055: String Properties / External Tags not working properly

patrick-schweizer:
This pr fixes missing resource level string attributes.

### External References

* JIRA (Issue Tracker): https://issues.opennms.org/browse/DC-17


j-white (approved):
Awesome!

agalue (approved):
I tested the solution with TSS plugins, and everything seems to be working now. I can see that OpenNMS renders the placeholders correctly when visualizing the list of resources and the same when generating the graphs.

PR #4354, opened 2022-03-09
NMS-14059: Add support for pre-authorization

j-white:
JIRA: https://issues.opennms.org/browse/NMS-14059

Set roles via HTTP header, used in conjunction with pre-authentication

### DIFF
--- a/features/springframework-security/src/main/java/org/opennms/web/springframework/security/RequestHeaderPreAuthenticationProcessingFilter.java
+++ b/features/springframework-security/src/main/java/org/opennms/web/springframework/security/RequestHeaderPreAuthenticationProcessingFilter.java
@@ -28,13 +28,23 @@
 
 package org.opennms.web.springframework.security;
 
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.stream.Collectors;
+
 import javax.servlet.http.HttpServletRequest;
 
 import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.security.core.GrantedAuthority;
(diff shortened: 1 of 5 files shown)


features/springframework-security/src/main/java/org/opennms/web/springframework/security/RequestHeaderPreAuthenticationProcessingFilter.java
```diff
@@ -65,6 +76,26 @@ public void afterPropertiesSet() {
...
+        private Collection<? extends GrantedAuthority> getGrantedAuthorities(HttpServletRequest context) {
+            String roles = "";
+            if (m_authoritiesHeader != null) {
+                roles = context.getHeader(m_authoritiesHeader);
+            }
+            if (StringUtils.isBlank(roles)) {
+                return Collections.emptyList();
+            }
+            return Arrays.stream(roles.split(","))
+                    .map(SimpleGrantedAuthority::new).collect(Collectors.toList());
```
RangerRick, line 97: is it worth doing a `trim()` also in case any spaces slip through?

RangerRick (approved):
LGTM

features/springframework-security/src/main/java/org/opennms/web/springframework/security/RequestHeaderPreAuthenticationProcessingFilter.java
j-white: Good point. Will add.

fooker: And a filter for empty strings, too?

RangerRick (approved):

PR #4356, opened 2022-03-10
NMS-14047: DCB Rest API: Parse and display cron schedule info

synqotik:
Parse cron schedule info from DCB Service metadata and return `nextScheduledBackup` and `scheduledInterval` in API.

Additionally:

- moved functions requiring Apache `commons-compress` from `DeviceConfigUtil` to `rest` module `CompressionUtils` class.
- Fixed device config download to have `Content-Disposition` of `attachment` so browser will download as file instead of inline text

Jira: https://issues.opennms.org/browse/NMS-14047

### DIFF
--- a/features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestService.java
+++ b/features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestService.java
@@ -30,6 +30,7 @@ package org.opennms.features.deviceconfig.rest.impl;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.time.Duration;
 import java.util.ArrayList;
@@ -50,6 +51,7 @@ import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
 
 import com.google.common.base.Strings;
+import net.redhogs.cronparser.CronExpressionDescriptor;
 import org.apache.commons.lang3.StringUtils;
 import org.opennms.core.criteria.Criteria;
 import org.opennms.core.criteria.CriteriaBuilder;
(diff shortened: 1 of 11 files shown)


features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestService.java
```diff
@@ -352,11 +356,19 @@ private void populateScheduleInfo(List<DeviceConfigDTO> dtos, Map<Integer, List<
...
 
-        return new ScheduleInfo(nextScheduledBackup, scheduleInterval);
+        try {
+            cronExpression = new CronExpression(schedulePattern);
+            cronDescription = CronExpressionDescriptor.getDescription(schedulePattern);
+        } catch (ParseException pe) {
+            return new ScheduleInfo(null, "unknown");
+        }
+
+        final Date nextScheduledBackup = cronExpression.getNextValidTimeAfter(current);
```
cgorantla, line 369: Can you validate if both of these return same value ?https://github.com/OpenNMS/opennms/blob/5b67f532fc0243fd9fe9827cfe4c629eb9c5372f/features/device-config/monitor/src/main/java/org/opennms/features/deviceconfig/monitors/DeviceConfigMonitor.java#L110

synqotik: I looked at source code:
https://github.com/quartz-scheduler/quartz/blob/ff348f62ece41275555b45dba5a4073c910fbeab/quartz-core/src/main/java/org/quartz/impl/triggers/CronTriggerImpl.java#L530

`trigger.getFireTimeAfter()` ultimately calls `CronExpression.getTimeAfter()` and `CronExpression.getNextTimeAfter()` just calls that method as well.

Main difference is that `Trigger` code will check its start/end times and make adjustments to make sure next fire time is within it's start/end limits, but Rest API doesn't have any knowledge of that.

I'm assuming there's a bit less overhead in calling `CronExpression.getNextValidTimeAfter` vs. building a Trigger. Kind of hard to do a unit test on this as is, but I will do some testing to confirm.

synqotik: Updated to use the same code as in DeviceConfigMonitor, but code is copied (with a comment) in order not to add cross-dependencies.

cgorantla (approved):
LGTM.

PR #4374, opened 2022-03-16
NMS-14082: DCB Add 'config' as string in Rest API response

synqotik:
DCB: Add 'config' as a string in the 'device-config' Rest API response

- use 'encoding' if supplied
- basic handling of binary data (returns hex representation of binary data)
- updates Rest integration test to use hamcrest matchers

It's expected we will need to make additional adjustments after testing with different kinds of text encoding or binary data.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14082

### DIFF
--- a/features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestService.java
+++ b/features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestService.java
@@ -73,6 +73,7 @@ import com.google.common.collect.Maps;
 
 public class DefaultDeviceConfigRestService implements DeviceConfigRestService {
     private static final Logger LOG = LoggerFactory.getLogger(DefaultDeviceConfigRestService.class);
+    public static final String DEFAULT_ENCODING = StandardCharsets.UTF_8.name();
 
     private static final Map<String,String> ORDERBY_QUERY_PROPERTY_MAP = Map.of(
         "lastupdated", "lastUpdated",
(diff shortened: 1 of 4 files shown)


features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestService.java
```diff
@@ -466,7 +468,11 @@ private Date getNextRunDate(String cronSchedule, Date lastRun) {
     }
 
     private static DeviceConfigDTO createDeviceConfigDto(DeviceConfig deviceConfig) {
+        // TODO: Handle different character encodings and also binary data
         try {
+            String config =
+                deviceConfig.getConfigType() == null ? "" : new String(deviceConfig.getConfig(), StandardCharsets.UTF_8);
```
cgorantla, line 474: why check for `deviceConfig.getConfigType()` ? 
Also it is `required` field so it will always be non-null

cgorantla: Can be updated to 
```suggestion
               String encoding = deviceConfig.getEncoding != null ? deviceConfig.getEncoding : Charset.defaultCharset().name();
                String config = new String(deviceConfig.getConfig(), Charset.forName(encoding));
```

synqotik: `getConfigType` was a typo on my part! Should have been `getConfig`. Also, I agree with your suggestion, will change

fooker:
@synqotik is there an issue for handling non-text config files?

features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestService.java
synqotik: Updated

cgorantla (approved):

PR #4416, opened 2022-03-21
NMS-14057: OpenNMS points to the wrong URL 

aramosvizcarra:
OpenNMS points to the wrong URL while generating graphs

fixed the issue identifying default "null" string from jsp

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14057

### DIFF
--- a/core/web-assets/src/main/assets/js/apps/onms-resources/index.js
+++ b/core/web-assets/src/main/assets/js/apps/onms-resources/index.js
@@ -95,7 +95,7 @@ angular.module('onms-resources', [
     if (reports) {
       $scope.reports = reports;
     }
-    if (endUrl) {
+    if (endUrl && endUrl !== "null") {
       $scope.url = endUrl;
     }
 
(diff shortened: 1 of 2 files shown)


RangerRick:
FYI this should be rebased to apply to `foundation-2019`

RangerRick (approved):
Ideally the `init()` in this jsp would be called differently depending on whether the attribute is set at all so that we're not passing an empty string in here in some cases, but making that work would be a huge PITA so this works.

Just a note for people down the road, this fix essentially works because of javascript's evaluation of the truthiness of empty strings as false. The code that overwrites the scoped variables never gets called so the behavior is fixed:

```javascript
  $scope.url = 'graph/results.htm';
  $scope.reports = 'all';
  $scope.loaded = false;
  $scope.generatedId = '';

  $scope.init = function(nodeCriteria, reports, endUrl) {
    if (!nodeCriteria) {
      return;
    }
    // Update node criteria in scope.
    $scope.nodeCriteria = nodeCriteria;

    if (reports) {
      $scope.reports = reports;
    }
    if (endUrl) {
      $scope.url = endUrl;
    }
    ```

cgorantla (approved):
I think this is coming from XSS changes:

`init('${e:forJavaScript(node)}','${e:forJavaScript(reports)}','${e:forJavaScript(endUrl)}')`

I guess wasp encoding (e:forJavaScript) converts  null as "null" string that's why it appends null in the failing case.

PR #4449, opened 2022-03-28
NMS-14116:  Get monitor/collector without resolving future

cgorantla:
Small update to not use `getCollectorByClassName` and `getMonitorByClassName`.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14116

### DIFF
--- a/features/collection/api/src/main/java/org/opennms/netmgt/collection/api/ServiceCollectorRegistry.java
+++ b/features/collection/api/src/main/java/org/opennms/netmgt/collection/api/ServiceCollectorRegistry.java
@@ -37,11 +37,6 @@ import java.util.concurrent.CompletableFuture;
  * @author jwhite
  */
 public interface ServiceCollectorRegistry {
-    /**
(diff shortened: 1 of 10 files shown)


j-white:
Would changing this https://github.com/OpenNMS/opennms/blob/develop/features/collection/api/src/main/java/org/opennms/netmgt/collection/support/DefaultServiceCollectorRegistry.java#L116 from:
```
    return getCollectorFutureByClassName(className).completeOnTimeout(null, 100, TimeUnit.MILLISECONDS).join();
```

to:
```
    return getCollectorFutureByClassName(className).getNow(null);
```

have the same effect?

fooker:
Should we remove `getCollectorByClassName` completely?

cgorantla:
> https://github.com/OpenNMS/opennms/blob/develop/features/collection/api/src/main/java/org/opennms/netmgt/collection/support/DefaultServiceCollectorRegistry.java#L116

yes, it has the same effect but this method was marked as deprecated, we can remove it altogether.

fooker (approved):

PR #4464, opened 2022-03-29
NMS-14117: add karaf config with rmi hostname fix from NMS-13887

RangerRick:
This merges the fix from NMS-13887 into the Horizon container.
The original fix only applied to Minion and Sentinel.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14117


dino2gnt (approved):

christianpape (approved):
LGTM.

PR #4476, opened 2022-03-31
NMS-14098: Create OpenNMS images for the ARM processor

freddy61025:
https://issues.opennms.org/browse/NMS-14098

The PR mainly work on horizon & sentinel switching Centos base to opennms/deploy-base.
arm64 image will trigger if branch contain arm64. 

apt repo can save 1-2mins if we create an extra image to cache install for (dpkg-dev pgp busybox & pre generage gpg key)
However, since we are it is relatively simple. It seems not worth to maintain an extra image.


indigo423:
Just tested the OCI artifact from the build today :) Awesome! Thanks for digging into this topic.
Screenshot 2022-04-01 at 16 21 21

RangerRick (changes requested):
This is good for testing, but really we should also build final multi-arch images for Horizon and Sentinel like we do Minion so that people can easily pull whichever version is handy from dockerhub.

freddy61025:
Currently it is only branch with arm will do that. 
Is that we should include develop and features/* ?

RangerRick:
But the idea is to merge to `develop` so M1 users don't have to use a special branch to run against.
So in that case, the "production" version of this should make multi-arch images just like we do for Minion.

RangerRick:
We'll need it to work on feature branches too, or people with new macs won't be able to develop in features.  :)

freddy61025:
new it build for following branch
            - develop
            - /^master-.*/
            - /^release-.*/
            - /.*arm64.*/

RangerRick (approved):
Instead of making the minion image attach properly as an artifact, we should just make a new issue for publishing everything to dockerhub. This is good for now at least.

deejgregor:
> we should just make a new issue for publishing everything to dockerhub

I didn't find one already open, so I created this: https://issues.opennms.org/browse/NMS-14626

PR #4478, opened 2022-03-31
NMS-14123: Sentinel debian package fail to install

freddy61025:
[jira/NMS-14123](https://issues.opennms.org/browse/NMS-14123)

It is cause by the file remove earlier than fix permission. And, the file list is obtain from debian package. So chown will exit with non 0 status code.


cgorantla (approved):
LGTM!

But wait for Ben's approval for merging.

RangerRick (approved):
sure, this is fine

PR #4483, opened 2022-04-01
NMS-14089: DCB: Create UI for comparing 2 backup configurations

mikewrosey:
- Adds vue-diff plugin and modal content component for displaying diff
- Fixes modal position on screen, relative to parent
- Displays config in a more 'editor-like' view, with line numbers
- Fixes position of modal download btn
- Fixes view history content so that dates are within their respective tabs, as per prototype

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-14089


synqotik (approved):
LGTM

PR #4500, opened 2022-04-04
NMS-14147: DCB Endpoint Renaming

synqotik:
Update Device Config Backup Rest APIs to make their purpose more clear and to clean up the code a bit.

- `device-config` with query params will return *all* records with the given query params for filter/sort/limit/offset
- `device-config/latest` will return only the latest record per ipinterface/configType (by lastUpdated) for the given query. This is the API used by the main UI table to get the latest status across all devices
- `device-config/interface/{ipInterfaceId}` returns a history of records for the given `ipInterfaceId`, basically the history for one device. UI will need to partition between config types

Also added an integration test plus cleaned test code up a bit.

In addition, this PR updates the UI with the new endpoints.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14147

### DIFF
--- a/features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/api/DeviceConfigRestService.java
+++ b/features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/api/DeviceConfigRestService.java
@@ -55,10 +55,11 @@ public interface DeviceConfigRestService {
 
     /**
      * Gets a list of device configs along with backup schedule information.
+     *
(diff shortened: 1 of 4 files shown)


mikewrosey (approved):
FE changes look good

cgorantla (commented):
LGTM!  

we can also add another subresource `device-config/interface/{ipinterfaceid}/{configType}`

That can be addressed in another issue along with  TODO for totalCount.

cgorantla (approved):

PR #4505, opened 2022-04-05
NMS-14151: DCB fix backup in Rest API and UI

synqotik:
Device Configuration backup (triggered from UI) not working correctly due to mismatch in serviceName used.

Rest APIs should return 'serviceName'. UI should use this service name in 'device-config/backup' call

UI should accept 202 response from 'backup' API call and display 'successToast'

UI should only display "Backup Now" when a single item is selected (no multiple backups at this time)

UI should display value from 'lastBackupDate' for "Last Backup" column

Fix download file issue in UI

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14151


mikewrosey (approved):
JS changes look good to me

cgorantla (approved):
LGTM!

PR #4514, opened 2022-04-07
NMS-14016: Run as non-privileged user with ambient capabilities

amay94:
This will switch the horizon, sentinel, and minion services over to run as their non-privileged users from the start instead of dropping down from root to start the process. This allows them to use ambient capabilities defined in the service definition file.

`CAP_NET_BIND_SERVICE` allows binding to privileged ports (for syslog and snmp).
`CAP_NET_RAW` allows creating raw sockets for icmp.

`ensure-user-ping.sh` was removed from the `init.d` start script because:
1. It needs root permissions, which the script no longer has
2. With the `NET_CAP_RAW` capability, the service can now open ICMP sockets without modifying `net.ipv4.ping_group_range`
3. This script is already run when installing RPMs / debian packages. See [minion](https://github.com/OpenNMS/opennms/blob/develop/opennms-assemblies/minion/src/main/filtered/debian/opennms-minion.postinst#L31), [sentinel](https://github.com/OpenNMS/opennms/blob/develop/opennms-assemblies/sentinel/src/main/filtered/debian/opennms-sentinel.postinst#L29), and [horizon](https://github.com/OpenNMS/opennms/blob/develop/debian/opennms-common.postinst#L39) debian install scripts, as well as [minion](https://github.com/OpenNMS/opennms/blob/develop/tools/packages/minion/minion.spec#L276), [sentinel](https://github.com/OpenNMS/opennms/blob/develop/tools/packages/sentinel/sentinel.spec#L221), and [horizon](https://github.com/OpenNMS/opennms/blob/develop/tools/packages/opennms/opennms.spec#L991) RPM install specs.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14016

### DIFF
--- a/container/shared/src/main/resources/etc/container.init
+++ b/container/shared/src/main/resources/etc/container.init
@@ -24,7 +24,6 @@ SYSCONFDIR="@SYSCONFDIR@"
 RUNAS="@RUNAS@"
 STOP_RETRIES=10
 STOP_WAIT=5
-PING_REQUIRED=FALSE
 PIDFILE="@PIDFILE@"
 
 export NAME DESC PATH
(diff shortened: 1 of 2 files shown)


amay94:
I've successfully installed and started all three services in both CentOS 7 and Ubuntu 20.04.4.

RangerRick:
Did you mean to include "horizon" in your description of the PR? This change only affects Minion and Sentinel. (Also the original issue seems to only be specifically about Minion, tho' Sentinel would of course also be affected.)

Additionally, this should probably be rebased to `foundation-2022` as non-root changes affect Meridian as well.

amay94:
I saw minion and sentinel use the same `inti.d` start script and I assumed horizon, or what ever we call the 'core' that talks to both, used it as well. That's why I mentioned horizon, whoops.

I'll update the request to point to foundation-2022.

RangerRick (approved):
👍

dino2gnt:
:+1:

PR #4518, opened 2022-04-08
NMS-14163: Monitor should return poll status based on last retrieval

cgorantla:
Currently, Monitor Service status always gives out Unknown status when it doesn't fetch config.
Instead this will change monitor to reflect last known config retrieval status.

This will enable manual retrieval related success/failures to be reflected in Monitor Service Status.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14163

### DIFF
--- a/features/device-config/monitor/src/main/java/org/opennms/features/deviceconfig/monitors/DeviceConfigMonitor.java
+++ b/features/device-config/monitor/src/main/java/org/opennms/features/deviceconfig/monitors/DeviceConfigMonitor.java
@@ -41,6 +41,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
+import com.google.common.base.Strings;
 import org.opennms.core.spring.BeanUtils;
 import org.opennms.core.utils.ConfigFileConstants;
 import org.opennms.features.deviceconfig.persistence.api.ConfigType;
@@ -67,6 +68,7 @@ public class DeviceConfigMonitor extends AbstractServiceMonitor {
     public static final String SSH_TIMEOUT = "ssh-timeout";
     public static final String PASSWORD = "password";
     public static final String LAST_RETRIEVAL = "lastRetrieval";
+    public static final String FAILURE_REASON = "failureReason";
     public static final String SCRIPT_FILE = "script-file";
 
     private static final Logger LOG = LoggerFactory.getLogger(DeviceConfigMonitor.class);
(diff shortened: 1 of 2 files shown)


features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestService.java
```diff
@@ -435,41 +435,47 @@ private WebApplicationException getException(final Status status, final String m
...
-
-        dto.setIpInterfaceId(queryResult.getIpInterfaceId());
-        dto.setNodeId(queryResult.getNodeId());
-        dto.setNodeLabel(queryResult.getNodeLabel());
-        dto.setDeviceName(queryResult.getNodeLabel());
-        dto.setLocation(queryResult.getLocation());
-        dto.setOperatingSystem(queryResult.getOperatingSystem());
 
-        populateScheduleInfo(dto);
+        try {
```
fooker, line 439: This seems like a workaround for something that 1) I don't understand and 2) needs more appropriate fixing.

features/device-config/monitor/src/main/java/org/opennms/features/deviceconfig/monitors/DeviceConfigMonitor.java
```diff
@@ -129,7 +132,13 @@ public PollStatus poll(MonitoredService svc, Map<String, Object> parameters) {
         final Date nextRun = getNextRunDate(cronSchedule, lastRun);
 
         if (!triggeredPoll && !nextRun.before(new Date())) {
-            return PollStatus.unknown("Skipping. Next retrieval scheduled for " + nextRun);
+            // Send previous status back based on failure reason.
+            String failureReason = getKeyedString(parameters, FAILURE_REASON, null);
+            if (Strings.isNullOrEmpty(failureReason)) {
```
fooker, line 137: This implies that the poller will be up before the first backup has been scheduled.

fooker (changes requested):
This seems to be (almost) fine. Maybe it is easier to iron out the corner cases if this is done in the adapter?

features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestService.java
cgorantla: Encountered when I removed nodes but device config still exist for those nodes.
This may be covered by https://issues.opennms.org/browse/NMS-14038  but we can keep this for not exposing lower level exceptions in Rest API.

cgorantla:
Should be covered by #4570

PR #4520, opened 2022-04-08
DC-72: TagMatchers: Query only on node level, not above

patrick-schweizer:
* JIRA (Issue Tracker): https://issues.opennms.org/browse/DC-72

### DIFF
--- a/features/timeseries/src/main/java/org/opennms/netmgt/timeseries/samplewrite/TimeseriesWriter.java
+++ b/features/timeseries/src/main/java/org/opennms/netmgt/timeseries/samplewrite/TimeseriesWriter.java
@@ -161,12 +161,9 @@ public class TimeseriesWriter implements WorkHandler<SampleBatchEvent>, Disposab
     }
 
     public void insert(List<Sample> samples) {
-        pushToRingBuffer(samples, TRANSLATOR);
-    }
 
-    private void pushToRingBuffer(List<Sample> samples, EventTranslatorOneArg<SampleBatchEvent, List<Sample>> translator) {
         // Add the samples to the ring buffer
-        if (!ringBuffer.tryPublishEvent(translator, samples)) {
+        if (!ringBuffer.tryPublishEvent(TRANSLATOR, samples)) {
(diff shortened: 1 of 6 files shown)


features/timeseries/src/main/java/org/opennms/netmgt/timeseries/samplewrite/TimeseriesWriter.java
```diff
@@ -161,12 +161,9 @@ public void destroy() {
     }
 
     public void insert(List<Sample> samples) {
```
patrick-schweizer, line 163: this is just to simplify code, has nothing to do with the querying

features/timeseries/src/main/java/org/opennms/netmgt/timeseries/util/TimeseriesUtils.java
```diff
@@ -47,9 +47,14 @@
...
-
-    public static final int WILDCARD_INDEX_NO = 2; // => node level
+    /**
+     * This constant defines at which level in a resource path we start to query for multiple metrics.
+     * To save queries, we want to get all resources under a node.
+     * requisitions have the node at the 4th position, example: snmp/fs/xx/4/
+     * auto detected metrics have the node at 2nd position, example: snmp/4/mib2-tcp
+     * since we can't distinguish we set the level to 4.
+     */
+    public static final int WILDCARD_INDEX_NO = 4;
```
patrick-schweizer, line 57: I wonder if we can distinguish between the 2 types of resources (provision vs. auto detect).
Maybe check if the second or 4th element is a number? Not sure if that would work?

patrick-schweizer: @j-white what is your take on this?

j-white (approved):

agalue (approved):
I believe the change is better and more constrained than the original approach, therefore a lot faster.

However, based on my tests, I can see subsequent queries for each IP interface on the node, which introduces other challenges, but I believe it is worth having the change in place even with them.

That said, I'd love to hear opinions from other OpenNMS developers.

j-white (approved):

PR #4532, opened 2022-04-12
NMS-14146: Fix initial DCB schedule

fooker:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14146


christianpape (commented):
LGTM.

PR #4536, opened 2022-04-12
NMS-14165: DCB correctly return device count for 'latest' query

synqotik:
Device Config Backup, Rest API, correctly return device count for `device-config/latest` call. This is total number of devices (i.e. ipinterface ids) that would be returned if 'limit' and 'offset' were not set.

This does an HQL query similar to `getLatestConfigForEachInterface` but simplified as we only need the count and don't need to do any sorting, grouping or 'windowing' functions. Do need to take search term into consideration.

Note, there are features in newer versions of Hibernate (i.e. `Projections.countDistinct`) that would allow this query to be done in Hibernate instead of HQL.

Also, a small update to display the device count in UI.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14165


cgorantla (approved):
LGTM!

PR #4558, opened 2022-04-18
NMS-14175: Remove config-type column, add icon, rename hooks to composables, useSpinner/useSnackbar

mikewrosey:
- Updates packages
- Removes config type column
- Adds speedometer mat icon to node label
- Change 'Back to old UI' btn to 'Back to main page'
- Highlights selected date on view history modal
- Removes 'in progress' status for now, as it is not used
- Renames 'hooks' folder to 'composables' to be more in line with Vue conventions
- Removes old toast and replaces it with the new feather snackbar component
- Adds useSpinner and useSnackbar composables, to be used wherever needed and removes old spinner/notification stores as they are now no longer used

### External References

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-14175


synqotik (approved):
Looks good!

PR #4566, opened 2022-04-19
NMS-14169: Added IEs to set input and output ifIndex

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-14169


cgorantla (approved):
LGTM!

PR #4575, opened 2022-04-20
NMS-14032: do not fail on directories that do not need write

RangerRick:
This PR makes a couple of changes to how the installer does validation for writeability.

* change to using `Files.walk` and an iterator, so that it can continue even if there is an exception reading from a directory or file, as long as that directory or file is ignored
* only check `etc`, `data`, `instances`, `logs`, and `share`, since any other directories shouldn't need to be written to

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14032

### DIFF
--- a/opennms-bootstrap/src/main/java/org/opennms/bootstrap/FilesystemPermissionValidator.java
+++ b/opennms-bootstrap/src/main/java/org/opennms/bootstrap/FilesystemPermissionValidator.java
@@ -32,25 +32,37 @@ import java.io.File;
 import java.io.FileFilter;
 import java.io.IOException;
 import java.nio.file.FileVisitOption;
+import java.nio.file.FileVisitResult;
+import java.nio.file.FileVisitor;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
 import java.util.Random;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 public class FilesystemPermissionValidator {
     public static boolean VERBOSE = false;
+    private static final Random RAND = new Random();
+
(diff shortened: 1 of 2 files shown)


opennms-bootstrap/src/main/java/org/opennms/bootstrap/FilesystemPermissionValidator.java
```diff
@@ -132,25 +131,35 @@ private void validatePath(final String user, final Path dirPath) throws Filesyst
...
-                        if (uri.contains("/.git/")) return false;
-                        if (uri.contains("/lost+found")) return false;
-                        return true;
-                    }).filter(file -> {
+        final List<Path> failures = new ArrayList<>();
+        try {
+            final var iter = Files.walk(dirPath, FileVisitOption.FOLLOW_LINKS).iterator();
+            while (true) {
+                try {
+                    if (iter.hasNext()) {
```
amay94, line 139: are there cases where iter.hasNext() can throw unexpected exceptions here? If not, that can replace `while(true)`

amay94 (commented):
Looks good, but it looks like that loop can be simplified a bit.

opennms-bootstrap/src/main/java/org/opennms/bootstrap/FilesystemPermissionValidator.java
RangerRick: From what I had read, if the file or directory exists but is unreadable due to permissions, it can throw a runtime exception in `hasNext`, but I didn't validate that directly.

cgorantla (dismissed):
LGTM!

opennms-bootstrap/src/main/java/org/opennms/bootstrap/FilesystemPermissionValidator.java
RangerRick: And actually, I re-evaluated what I did and I found some thinkos in other parts of the logic, so I ended up redoing a decent amount and I'm much happier with it now.  Mind taking another look?

aramosvizcarra:
Looks good to me

cgorantla (commented):
LGTM!

PR #4582, opened 2022-04-22
NMS-14200:  Wait for roles to load before checking route access

mikewrosey:
Routes were loading before API response with roles was returned.
On a first load, this will wait until the roles are available until loading a protected route.

### External References

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-14200


synqotik (approved):
LGTM!

PR #4593, opened 2022-04-25
NMS-14206: Move container bridge logging to INFO

cgorantla:
* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14206

### DIFF
--- a/container/karaf/src/main/filtered-resources/etc/org.ops4j.pax.logging.cfg
+++ b/container/karaf/src/main/filtered-resources/etc/org.ops4j.pax.logging.cfg
@@ -66,6 +66,10 @@ log4j2.logger.kafkaAdminClientConfig.level = ERROR
 log4j2.logger.opennmsTopology.name = org.opennms.features.topology
 log4j2.logger.opennmsTopology.level = WARN
 
+# OPENNMS: Reduce container bridge logging to INFO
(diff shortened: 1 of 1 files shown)


container/karaf/src/main/filtered-resources/etc/org.ops4j.pax.logging.cfg
```diff
@@ -66,6 +66,10 @@ log4j2.logger.kafkaAdminClientConfig.level = ERROR
 log4j2.logger.opennmsTopology.name = org.opennms.features.topology
 log4j2.logger.opennmsTopology.level = WARN
 
+# OPENNMS: Reduce container bridge logging to INFO
+log4j2.logger.opennms.name = org.opennms.container.web.bridge.rest
```
synqotik, line 70: I'm not sure about the semantics of this config file, but is it correct to have the same keys (`log4j2.logger.opennms.name` and `level`) as on lines 74-75 below?

synqotik (approved):

container/karaf/src/main/filtered-resources/etc/org.ops4j.pax.logging.cfg
cgorantla: you are right, fixed!

cgorantla:
Verified locally and also on the CircleCI logs.

PR #4599, opened 2022-04-27
NMS-14218: Masked credentials should be ignored in update

cgorantla:
PUT: should ignore updating username/password whenever password is masked.
         will overwrite attributes and username/password in PUT ( should have existing credentials).
         username/password can be null in update, attributes not in the request will be deleted.

POST: will handle new credentials whenever alias doesn't exist.

masked password check ( any thing with more than two **)

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14218

### DIFF
--- a/features/scv/rest/src/main/java/org/opennms/features/scv/rest/DefaultScvRestService.java
+++ b/features/scv/rest/src/main/java/org/opennms/features/scv/rest/DefaultScvRestService.java
@@ -36,15 +36,18 @@ import org.slf4j.LoggerFactory;
 
 import javax.ws.rs.core.Response;
 import java.text.Collator;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


features/scv/rest/src/main/java/org/opennms/features/scv/rest/DefaultScvRestService.java
```diff
@@ -60,8 +63,8 @@ public Response getCredentials(String alias) {
             if (credentials.getUsername() != null && credentials.getPassword() != null) {
                 // Mask password.
                 CredentialsDTO dto = new CredentialsDTO(alias, credentials.getUsername(), "******");
-                dto.addAttributes(credentials.getAttributes());
-                
+                // Mask values in attributes.
+                credentials.getAttributes().forEach((key, value) -> dto.getAttributes().put(key, "******"));
```
synqotik, line 67: Perhaps make the mask a constant since we are now using it twice?

synqotik (approved):
LGTM

PR #4638, opened 2022-05-03
NMS-14239: Fixing new UI list log & etc fail due to symbolic link

freddy61025:
https://issues.opennms.org/browse/NMS-14239

Tested locally both standalone & docker works fine.

image

### DIFF
--- a/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/FilesystemRestService.java
+++ b/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/FilesystemRestService.java
@@ -33,6 +33,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.FileVisitOption;
 import java.nio.file.Files;
 import java.nio.file.Paths;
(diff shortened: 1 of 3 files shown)


opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/FilesystemRestService.java
```diff
@@ -98,8 +98,9 @@ public List<String> getFiles(@QueryParam("changedFilesOnly") boolean changedFile
         }
 
         try {
-            return Files.find(etcFolder, 4, (path, basicFileAttributes) -> isSupportedExtension(path))
```
RangerRick, line 101: Silly question, but can't you just change the `Files.find` to have `FOLLOW_SYMLINKS` on the end of it using the 4-arg version?  This means even if someone symlinks around other things deeper in the tree, it would still work without having to do the extra map.

```java
return Files.find(etcFolder, 4, (path, basicFileAttributes) -> isSupportedExtension(path), FileVisitOption.FOLLOW_LINKS)
```

freddy61025: tested all working fine.

cgorantla (approved):

PR #4647, opened 2022-05-04
NMS-14148: Still listen for traps without a subscribtion

amay94:
If the broker is not available :  The `opennms-trapd-listener` feature would hang indefinitely while waiting for a `TwinSubscriber` to initialize, which couldn't complete until a connection to the core was made. This prevented the feature from ever attempting to listen for traps.

These updates make this dependency optional. The TrapListener will now wait 60 seconds for trapd configuration to come from the core instance. If that is not received, it uses default settings and listens anyways.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14148

### DIFF
--- a/features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapListener.java
+++ b/features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapListener.java
@@ -102,13 +102,30 @@ public class TrapListener implements TrapNotificationListener {
     }
 
     public void start() {
-        m_twinSubscription = m_twinSubscriber.subscribe(TrapListenerConfig.TWIN_KEY, TrapListenerConfig.class, (config) ->  {
+        if (m_twinSubscriber != null) {
+            subscribe();
+        } else {
+            open(new TrapListenerConfig());
+        }
+    }
+
+    public void subscribe() {
+        m_twinSubscription = m_twinSubscriber.subscribe(TrapListenerConfig.TWIN_KEY, TrapListenerConfig.class, (config) -> {
             try (Logging.MDCCloseable mdc = Logging.withPrefixCloseable(Trapd.LOG4J_CATEGORY)) {
                 LOG.info("Got listener config update - reloading");
                 this.close();
                 this.open(config);
             }
         });
+    }
+
+    public void bind(TwinSubscriber twinSubscriber) {
+        m_twinSubscriber = twinSubscriber;
+        subscribe();
+    }
+
+    public void unbind(TwinSubscriber twinSubscriber) {
+        m_twinSubscriber = null;
     }
 
     private void open(final TrapListenerConfig config) {
@@ -186,14 +203,6 @@ public class TrapListener implements TrapNotificationListener {
 
     public void setDistPollerDao(DistPollerDao distPollerDao) {
         m_distPollerDao = Objects.requireNonNull(distPollerDao);
-    }
-
-    public TwinSubscriber getTwinSubscriber() {
-        return this.m_twinSubscriber;
-    }
-
-    public void setTwinSubscriber(final TwinSubscriber twinSubscriber) {
-        this.m_twinSubscriber = Objects.requireNonNull(twinSubscriber);
     }
 
     private void restartWithNewConfig(final TrapdConfigBean newConfig) {
--- a/features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapListener.java
+++ b/features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapListener.java
@@ -1,8 +1,8 @@
 /*******************************************************************************
  * This file is part of OpenNMS(R).
  *
- * Copyright (C) 2002-2014 The OpenNMS Group, Inc.
- * OpenNMS(R) is Copyright (C) 1999-2014 The OpenNMS Group, Inc.
+ * Copyright (C) 2002-2022 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2022 The OpenNMS Group, Inc.
  *
  * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
  *
@@ -34,6 +34,9 @@ import java.lang.reflect.UndeclaredThrowableException;
 import java.net.InetAddress;
 import java.net.SocketException;
 import java.util.Objects;
+import java.util.Timer;
+import java.util.TimerTask;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.opennms.core.ipc.sink.api.AsyncDispatcher;
 import org.opennms.core.ipc.sink.api.MessageDispatcherFactory;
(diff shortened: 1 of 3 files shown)


features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapListener.java
```diff
@@ -102,15 +108,49 @@ public void trapError(int error, String msg) {
...
-        m_twinSubscription = m_twinSubscriber.subscribe(TrapListenerConfig.TWIN_KEY, TrapListenerConfig.class, (config) ->  {
+        if (m_twinSubscriber != null) {
+            subscribe();
+        } else {
+            new Timer().schedule(new TimerTask() {
+                @Override
+                public void run() {
+                    delayedConfigurationCheck();
+                }
+            }, 60 * 1000); // wait 60 seconds for connection from core
```
cgorantla, line 119: Make this `60*1000` constant.

features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapListener.java
```diff
@@ -102,15 +108,49 @@ public void trapError(int error, String msg) {
     }
 
     public void start() {
-        m_twinSubscription = m_twinSubscriber.subscribe(TrapListenerConfig.TWIN_KEY, TrapListenerConfig.class, (config) ->  {
+        if (m_twinSubscriber != null) {
+            subscribe();
+        } else {
+            new Timer().schedule(new TimerTask() {
```
cgorantla, line 114: Could use lambda expression here

features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapListener.java
```diff
@@ -56,6 +59,9 @@
 public class TrapListener implements TrapNotificationListener {
     private static final Logger LOG = LoggerFactory.getLogger(TrapListener.class);
 
+    // True if trap configuration has been received from the core instance
+    private AtomicBoolean m_configured = new AtomicBoolean(false);
```
cgorantla, line 63: Can't we use `m_registeredForTraps` to make the decision to open or not ?

features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapListener.java
```diff
@@ -102,15 +108,49 @@ public void trapError(int error, String msg) {
...
+        }
+    }
+
+    public void subscribe() {
+        m_twinSubscription = m_twinSubscriber.subscribe(TrapListenerConfig.TWIN_KEY, TrapListenerConfig.class, (config) -> {
             try (Logging.MDCCloseable mdc = Logging.withPrefixCloseable(Trapd.LOG4J_CATEGORY)) {
                 LOG.info("Got listener config update - reloading");
-                this.close();
-                this.open(config);
+                synchronized(m_configured) {
```
cgorantla, line 136: since m_configured is already atomic boolean, do you still need synchronization here ?

features/events/traps/src/test/java/org/opennms/netmgt/trapd/TrapListenerTest.java
```diff
@@ -127,4 +127,16 @@ public static Snmpv3User createUser(String authProtocol,
         user.setSecurityName(securityName);
         return user;
     }
+
+    @Test
+    public void noSubscriberTest() throws Exception {
+        TrapListener listener = new TrapListener(initialConfig);
+        listener.start();
+        Assert.assertEquals(Boolean.FALSE, listener.isRegisteredForTraps());
+        try {
```
cgorantla, line 136: Try using  `Awaitility` in order to avoid sleeping.

If you can make `60` secs configurable it can be overwritten from the test without much wait.

cgorantla (changes requested):

features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapListener.java
amay94: TimerTask isn't actually a functional interface, so you can't use lamda functions to define them.

features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapListener.java
amay94: I think so. For the TimerTask, checking the boolean value and then opening the trap listener need to be atomic with respect to this code block. My thinking is that if the TimerTask decides to start opening the listener with a default config, we want that to finish completely before this is attempted, if they happen to occur at the same time. m_configured doesn't need to be an AtomicBoolean, I just started with it that way and I wanted an object to synchronize around. I'll switch it to a standard Boolean object

features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapListener.java
amay94: `m_registeredForTraps` is updated near the middle of the `open` method. There's a chance that the TimerTask could wake up and see `m_registeredForTraps=false` and attempt to use default settings while an update is already in progress. i didn't want to move where the value is set because it's in the correct spot, so I just made another variable.

features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapListener.java
```diff
@@ -293,4 +293,12 @@ public boolean hasSnmpV3UsersChanged(TrapdConfig newConfig) {
         }
         return !newConfig.getSnmpV3Users().equals(m_config.getSnmpV3Users());
     }
+
+    public void setSubscriberTimeoutMs(long ms) {
```
cgorantla, line 297: Both of these can be package-private

features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapListener.java
```diff
@@ -63,7 +63,7 @@ public class TrapListener implements TrapNotificationListener {
     private boolean m_configured = false;
     private Object configuredLock = new Object();
 
-    public static long SUBSCRIBER_TIMEOUT_MS = 60 * 1000;
+    private long SUBSCRIBER_TIMEOUT_MS = 60 * 1000;
```
cgorantla, line 66: member variable `camelCase` notation can be followed here

features/events/traps/src/test/java/org/opennms/netmgt/trapd/TrapListenerTest.java
```diff
@@ -127,4 +129,16 @@ public static Snmpv3User createUser(String authProtocol,
...
     }
+
+    @Test
+    public void noSubscriberTest() throws Exception {
+        TrapListener listener = new TrapListener(initialConfig);
+        listener.setSubscriberTimeoutMs(2 * 1000);
+        listener.start();
+        Assert.assertEquals(Boolean.FALSE, listener.isRegisteredForTraps());
+        
+        Awaitility.await().atMost(3 * 1000, TimeUnit.MILLISECONDS).with()
```
cgorantla, line 140: `atMost(5*1000` would be better

cgorantla (commented):
LGTM,  few comments related to tests

cgorantla (approved):

cgorantla (commented):
Probably this needs to be re-targeted to `foundation-2022`

cgorantla (approved):

PR #4651, opened 2022-05-04
NMS-14214: DCB: Check for and delete stale configs

synqotik:
Device Config Backup
Check for and delete "stale" device configs, based on a retention date

Retention date can be configured, use Java "Period" format: see [here](https://docs.oracle.com/javase/8/docs/api/java/time/Period.html#parse-java.lang.CharSequence-)

When a successful backup occurs for a device, database is queried for previous backups. This should delete from the database any records that:
- contain config data
- have a 'lastUpdated' date older than (current date - retention period)
- are the not only record with configuration data

If a retention period is not specified in metadata, it defaults to 1 year

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14214

### DIFF
--- a/features/device-config/monitor-adaptor/src/main/java/org/opennms/features/deviceconfig/monitor/adaptor/DeviceConfigMonitorAdaptor.java
+++ b/features/device-config/monitor-adaptor/src/main/java/org/opennms/features/deviceconfig/monitor/adaptor/DeviceConfigMonitorAdaptor.java
@@ -30,18 +30,27 @@ package org.opennms.features.deviceconfig.monitor.adaptor;
 
 import java.io.IOException;
 import java.nio.charset.Charset;
+import java.time.Period;
+import java.time.ZoneId;
+import java.time.format.DateTimeParseException;
+import java.util.Date;
+import java.util.List;
 import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
 
+import com.google.common.base.Strings;
 import org.opennms.features.deviceconfig.persistence.api.ConfigType;
+import org.opennms.features.deviceconfig.persistence.api.DeviceConfig;
 import org.opennms.features.deviceconfig.persistence.api.DeviceConfigDao;
 import org.opennms.features.deviceconfig.persistence.api.DeviceConfigStatus;
+import org.opennms.features.deviceconfig.service.DeviceConfigConstants;
 import org.opennms.features.deviceconfig.service.DeviceConfigUtil;
 import org.opennms.netmgt.dao.api.IpInterfaceDao;
 import org.opennms.netmgt.events.api.EventConstants;
 import org.opennms.netmgt.events.api.EventForwarder;
 import org.opennms.netmgt.model.OnmsIpInterface;
 import org.opennms.netmgt.model.events.EventBuilder;
-import org.opennms.netmgt.poller.DeviceConfig;
 import org.opennms.netmgt.poller.MonitoredService;
 import org.opennms.netmgt.poller.PollStatus;
 import org.opennms.netmgt.poller.ServiceMonitorAdaptor;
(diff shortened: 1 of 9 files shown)


features/device-config/monitor-adaptor/src/main/java/org/opennms/features/deviceconfig/monitor/adaptor/DeviceConfigMonitorAdaptor.java
```diff
@@ -151,11 +160,48 @@ private String getObjectAsString(Object object) {
...
+        // Check for stale config records and delete them
+        final Date currentDate = new Date();
+        final String retentionPeriodPattern = getObjectAsString(parameters.get(DeviceConfigConstants.RETENTION_PERIOD));
+        final Period retentionPeriod = getRetentionPeriodOrDefault(retentionPeriodPattern);
+        final Date retentionDate = Date.from(currentDate.toInstant().atZone(ZoneId.systemDefault()).minus(retentionPeriod).toInstant());
+
+        // Get history of configs with config data, see if any exist that are not the latest one and are also
+        // older than the retention date
+        List<DeviceConfig> allConfigs = deviceConfigDao.findConfigsForInterfaceSortedByDate(ipInterface, serviceName);
+        Optional<DeviceConfig> lastGoodConfig = allConfigs.stream().filter(dc -> dc.getConfig() != null).findFirst();
```
cgorantla, line 173: It may be more efficient to retrieve stale configs by SQL query itself.

features/device-config/service/src/main/java/org/opennms/features/deviceconfig/service/impl/DeviceConfigServiceImpl.java
```diff
@@ -180,6 +180,11 @@ public String getConfigType() {
                         public String getSchedule() {
                             return getKeyedString(pollerParameters, DeviceConfigConstants.SCHEDULE, DeviceConfigConstants.DEFAULT_CRON_SCHEDULE);
                         }
+
+                        @Override
+                        public String getRetentionPeriod() {
```
cgorantla, line 185: Unless we are using this in UI, this may not be needed.

cgorantla (commented):
LGTM, one comment to see if we can move retrieving stale configs to SQL query.

features/device-config/monitor-adaptor/src/main/java/org/opennms/features/deviceconfig/monitor/adaptor/DeviceConfigMonitorAdaptor.java
synqotik: Good point, will take a look tomorrow.

synqotik: Updated this to use Dao

features/device-config/service/src/main/java/org/opennms/features/deviceconfig/service/impl/DeviceConfigServiceImpl.java
synqotik: Removed

features/device-config/persistence/impl/src/main/java/org/opennms/features/deviceconfig/persistence/impl/DeviceConfigDaoImpl.java
```diff
@@ -82,6 +82,27 @@ public List<DeviceConfig> findConfigsForInterfaceSortedByDate(OnmsIpInterface ip
...
+            Date staleDate, Optional<Long> excludedId) {
+        var builder = new CriteriaBuilder(DeviceConfig.class);
+        builder
+            .isNotNull("lastUpdated")
+            .isNotNull("config")
+            .eq("ipInterface.id", ipInterface.getId())
+            .eq("serviceName", serviceName)
+            .lt("lastUpdated", staleDate);
+
+        if (excludedId.isPresent()) {
```
cgorantla, line 97: What if this is the first config that we are persisting and excludedId is not present then we don't need to make this query since we already know this is the only config for this interface and servicename.

synqotik: Good point.

In `DeviceConfigMonitorAdaptor.handlePollResult` we could check if `latestConfig.isPresent()`. If it is *not*, then we know right before `cleanupStaleConfigs` that the first `DeviceConfig` was added (in `deviceConfigDao.updateDeviceConfigContent`), and don't need to call `cleanupStaleConfigs`.

synqotik: Just added this, but need to test

cgorantla (approved):
LGTM!

PR #4662, opened 2022-05-09
create "core" and "shared" feature files

RangerRick:
One of the issues I'm running into while trying to upgrade guava is that on Minion and Sentinel, we want "real" dependencies so things get loaded, but in the Horizon core, we want to do nothing because we're exporting those dependencies from the `lib/` dir and the dependencies are already present because of `custom.properties`.

It is always necessary to use the `custom.properties` version on core if we are sharing code across the spring<->OSGi bridge or we run into classloader issues.

This change provides a way to "depend" on something, while still assuming it exists on the core side.

There is no JIRA issue associated with this, it's just a baseline bit of cleaning things up that will make other stuff simpler.


cgorantla (commented):
I don't see any changes here.
Am I missing anything  ?

Screen Shot 2022-05-10 at 4 22 00 PM
?

RangerRick:
Actually, I had an epiphany and realized what `<bundle dependency="true">` _actually_ means and am going to approach this differently.  Gonna close it without merging.

PR #4664, opened 2022-05-09
NMS-14038: Delete device configs for deleted IpInterfaces

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14038


RangerRick (approved):
seems pretty straightforward

features/device-config/persistence/impl/src/main/java/org/opennms/features/deviceconfig/persistence/impl/DeviceConfigDaoImpl.java
```diff
@@ -229,6 +229,12 @@ public int getLatestConfigCountForEachInterface(String searchTerm, Set<DeviceCon
         return count;
     }
 
+    @Override
+    public List<DeviceConfig> getAllDeviceConfigsWithAnInterfaceId(Integer ipInterfaceId) {
+        return find("from DeviceConfig dc where dc.lastUpdated is not null AND dc.ipInterface.id = ? ORDER BY lastUpdated DESC",
```
synqotik, line 234: May just need `from DeviceConfig dc where dc.ipInterface.id = ?`, i.e. don't need to sort and also this would catch entries where backup has not yet occurred.

cgorantla: don't need sort but all entries should be deleted including entries that are waiting for backup

synqotik: `lastUpdated` will be null for entries that are waiting for backup

cgorantla: good catch, will take care of that!

synqotik (approved):
LGTM!

synqotik (approved):

PR #4665, opened 2022-05-09
NMS-14248: Fix findByIpAddressAndLocation and it's usage

cgorantla:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14248


features/device-config/sink/consumer/src/main/java/org/opennms/features/deviceconfig/sink/consumer/DeviceConfigConsumer.java
```diff
@@ -106,6 +108,19 @@ public void handleMessage(DeviceConfigSinkDTO message) {
         }
     }
 
+    private OnmsIpInterface findMatchingInterface(final String ipAddress, final String location) {
```
synqotik, line 111: Perhaps make `DeviceConfigService.findMatchingInterface` public, then call it here (with `serviceName` = `null`) rather than duplicate code? Not sure if that introduces any dependency issues.

synqotik (approved):
LGTM

PR #4686, opened 2022-05-12
NMS-14266 remove tools/phonebook/pom-xml

dherdt:
### External References

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-14266


freddy61025:
@cgorantla in the issue it said "Meridian-2022.1.2" Is that means it need to target at least foundation-2022 ?

cgorantla:
> @cgorantla in the issue it said "Meridian-2022.1.2" Is that means it need to target at least foundation-2022 ?

yes. Should be re-targeted to `foundation-2022`

mmahacek:
I found the issue while looking through the `foundation-2022` branch, though it looks like the file was originally removed in a `foundation-2021` update and later added back, so I don't know if it's worth to target 2021 since it's currently still a supported release. IANAD

cgorantla (changes requested):
needs to be re-targeted to `foundation-2022`

cgorantla (approved):

PR #4689, opened 2022-05-12
NMS-14725: Don't error out early when script file is missing

cgorantla:
We want to create empty entries for all devices if they have DeviceConfig
service running.
If we throw exception in run time attributes, that will prevent us
from creating these entries. So send this error to poll and throw error there

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14275


synqotik (approved):
LGTM!

christianpape (approved):
Looks good to me.

synqotik (dismissed):
Re-approving.

synqotik (approved):

PR #4692, opened 2022-05-13
NMS-14291: make sure snapshot OIA is always updated when building

RangerRick:
This PR changes our OIA dependencies to be routed through an `oia-dependencies` package that sets the snapshot repository to always update.  This _should_ 🤞 make sure OIA is up-to-date when doing development builds.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14291


cgorantla (approved):
LGTM!

PR #4697, opened 2022-05-16
NMS-14297: Use sink as config type for push based configs 

cgorantla:
Any config that is uploaded from device will use config type sink instead of default.
Service name for these will be null

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14297


synqotik (dismissed):
LGTM!

synqotik (approved):

PR #4700, opened 2022-05-16
NMS-14304: DCB UX Updates

mikewrosey:
- Backup Now -> Backup
- Removed elevation on surface
- 24px above, & below title
- Remove last backup icon, add semibold hover state to text
- Move search to same row as title
- Added exported icons from Figma
- Added padding to backup status

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-14304


ui/src/assets/Backup.vue
```diff
@@ -0,0 +1,17 @@
+<template>
+  <div class="backup">
+    <Backup />
+  </div>
+</template>
+
+<script setup lang=ts>
```
synqotik, line 7: Should this and elsewhere have quote marks around "ts"?
`<script setup lang=ts>` -> `<script setup lang="ts">`

mikewrosey: It's not strictly necessary, but I would agree it's better to have it. Fixed on all files.

synqotik (approved):
Looks good!

PR #4711, opened 2022-05-17
NMS-14228: SCV:  Cache JCEKS credentials in memory

aramosvizcarra:
Added HashMap to cache credentials in memory

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14228:

### DIFF
--- a/features/scv/jceks-impl/src/main/java/org/opennms/features/scv/jceks/JCEKSSecureCredentialsVault.java
+++ b/features/scv/jceks-impl/src/main/java/org/opennms/features/scv/jceks/JCEKSSecureCredentialsVault.java
@@ -144,6 +144,7 @@ public class JCEKSSecureCredentialsVault implements SecureCredentialsVault {
     @Override
     public void setCredentials(String alias, Credentials credentials) {
         try {
+            loadCredentials();
             byte[] credentialBytes = toBase64EncodedByteArray(credentials);
             SecretKeyFactory factory = SecretKeyFactory.getInstance("PBE");
             SecretKey generatedSecret = factory.generateSecret(
--- a/features/scv/jceks-impl/src/main/java/org/opennms/features/scv/jceks/JCEKSSecureCredentialsVault.java
+++ b/features/scv/jceks-impl/src/main/java/org/opennms/features/scv/jceks/JCEKSSecureCredentialsVault.java
@@ -114,23 +114,23 @@ public class JCEKSSecureCredentialsVault implements SecureCredentialsVault {
             if (!m_credentialsCache.isEmpty()) {
                 return;
             }
-        }
-        try {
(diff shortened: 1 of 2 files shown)


features/scv/jceks-impl/src/main/java/org/opennms/features/scv/jceks/JCEKSSecureCredentialsVault.java
```diff
@@ -136,6 +146,7 @@ public void setCredentials(String alias, Credentials credentials) {
             KeyStore.PasswordProtection keyStorePP = new KeyStore.PasswordProtection(m_password);
             m_keystore.setEntry(alias, new KeyStore.SecretKeyEntry(generatedSecret), keyStorePP);
             writeKeystoreToDisk();
+            m_credentialsCache.put(alias, credentials);
```
cgorantla, line 149: need to do this in thread safe manner.

features/scv/jceks-impl/src/main/java/org/opennms/features/scv/jceks/JCEKSSecureCredentialsVault.java
```diff
@@ -102,29 +105,36 @@ public JCEKSSecureCredentialsVault(String keystoreFile, String password, byte[]
...
+            loadCredentials();
         } catch (KeyStoreException | NoSuchAlgorithmException | CertificateException | IOException e) {
             throw Throwables.propagate(e);
         }
     }
-    
-    @Override
-    public Credentials getCredentials(String alias) {
+
+    private void loadCredentials() {
```
cgorantla, line 114: This may increase startup time, should be moved to first call in `getCredentials`

features/scv/jceks-impl/src/main/java/org/opennms/features/scv/jceks/JCEKSSecureCredentialsVault.java
```diff
@@ -76,6 +77,7 @@ public class JCEKSSecureCredentialsVault implements SecureCredentialsVault {
     private final byte[] m_salt;
     private final int m_iterationCount;
     private final int m_keyLength;
+    private final HashMap<String, Credentials> m_credentialsCache;
```
cgorantla, line 80: Can initialize map here.

cgorantla (commented):
Need a test to validate this.

features/scv/jceks-impl/src/main/java/org/opennms/features/scv/jceks/JCEKSSecureCredentialsVault.java
aramosvizcarra: OK. I was wondering if I should use a synchronizedMap or ConcurrentHashMap. Looks like for the size of the matter  using synchronized should be enough.

cgorantla (commented):
LGTM!  Can you add a test case for this ?

PR #4749, opened 2022-05-25
NMS-14318: DCB history/compare to only display one config type

mikewrosey:
- Remove existing tabs in view history modal. Add configName prop for title.
- Remove more than one config type in compare configs modal.
- Retrieve configs for history/compare modals using the `configType` param
- Change page header to use Headline1
- Vertically center checkboxes on DCB Table
- Remove sorting for `Schedule Date` and `Schedule Interval` columns on DCB table
- Replace the default HTML tooltip (title tag) with FeatherTooltip

### External References

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-14318


mikewrosey:
@chinh-cle This updates to the latest feather version, which includes a needed patch for the tooltip. However, I had to make some small changes to Configuration to fix the TS issues.

synqotik (approved):
LGTM!

PR #4764, opened 2022-05-28
NMS-14314: MIB compiler creates datacollections that fail to validate

aramosvizcarra:
fix issue parsing type ipaddress when generating datacollection using RFC-1213

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14314

### DIFF
--- a/features/mib-compiler/src/main/java/org/opennms/features/mibcompiler/services/JsmiMibParser.java
+++ b/features/mib-compiler/src/main/java/org/opennms/features/mibcompiler/services/JsmiMibParser.java
@@ -451,6 +451,8 @@ public class JsmiMibParser implements MibParser, Serializable {
             return "string";
         if (type.equals(SmiPrimitiveType.UNSIGNED_32)) // Unsigned32 will be treated as integer.
             return "integer";
+        if (type.equals(SmiPrimitiveType.IP_ADDRESS)) // IpAddress will be treated as strings.
+            return "string";
         return type.toString().replaceAll("_", "").toLowerCase();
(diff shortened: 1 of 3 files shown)


features/mib-compiler/src/main/java/org/opennms/features/mibcompiler/services/JsmiMibParser.java
```diff
@@ -451,6 +451,8 @@ private String getMetricType(SmiPrimitiveType type) {
             return "string";
         if (type.equals(SmiPrimitiveType.UNSIGNED_32)) // Unsigned32 will be treated as integer.
             return "integer";
+        if (type.equals(SmiPrimitiveType.IP_ADDRESS)) // IpAddress will be treated as strings.
+            return "string";
         return type.toString().replaceAll("_", "").toLowerCase();
```
j-white, line 456: Will this ever return a correct type, or should we just default to "string" here instead?

aramosvizcarra: From my conversation with @jeffgdotorg  IpAddress type should be treated as string when generating collection files. I believe only those values allowed in the datacollection-config.xsd file should remain. 
The XML parser will fail to parse a type is not allowed in the datacollection-config.xsd and Jeff mentioned something like this type shouldn't be allowed because is some sort of custom type... I think.

RangerRick (approved):

cgorantla (approved):

PR #4790, opened 2022-06-01
NMS-14324: DCB Order by Location and Backup Status

synqotik:
NMS-14324: Device Config Backup, Rest API and UI, order by Location and Backup Status.

- Update Rest API `latest` to allow `orderBy` on location and status
- Remove code to determine backup status by device config dates. Backup status is now stored in the DB and status is determine by `DeviceConfigMonitorAdaptor` during polling
- Update integration tests
- Minimal changes in UI to enable sort by location
- UI sort by status is not implemented, would need changes to `DCBTableStatusDropdown`. However, filter by status works correctly
- UI updated 'status' query params for filtering by status to be uppercase, otherwise Java server code doesn't recognize parameters and returns a 404
- FIxed small bug in `getAndMergeDeviceConfigBackups` where responses with `204 - No Content` contained just empty string data, instead of expected array data

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14324

### DIFF
--- a/features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestService.java
+++ b/features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestService.java
@@ -28,6 +28,7 @@
 
 package org.opennms.features.deviceconfig.rest.impl;
 
+import org.apache.commons.lang3.NotImplementedException;
 import static org.opennms.features.deviceconfig.service.DeviceConfigService.DEVICE_CONFIG_PREFIX;
 
 import java.io.ByteArrayOutputStream;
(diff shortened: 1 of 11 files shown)


mikewrosey (dismissed):
FE changes look good

features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestService.java
```diff
@@ -28,6 +28,7 @@
 
 package org.opennms.features.deviceconfig.rest.impl;
 
+import org.apache.commons.lang3.NotImplementedException;
```
cgorantla, line 31: looks like this is not getting used

cgorantla (commented):
LGTM!   Minor comment.

features/device-config/rest/src/main/java/org/opennms/features/deviceconfig/rest/impl/DefaultDeviceConfigRestService.java
synqotik: Removed

cgorantla (approved):

PR #4810, opened 2022-06-02
NMS-14209: upgrade Guava to 25.1

RangerRick:
This PR upgrades Guava to 25.1 in `foundation-2019`. There are equivalent branches made for the other foundation and release branches that differ only in fixing conflicts and changing the pom version of "new" files.  Like my other "upgrade" PRs, each commit is meant to be digestable on its own, to make it easier to review, since it changes quite a bit of stuff.

There are a couple things that aren't directly related to the fix, one is a thing that lets me automatically find the right JDK using `asdf` on the command-line. Kind of like npm but for all kinds of things. The other removes the "range" portion of the pax-logging entries in the `overrides.properties` file. Turns out if I do that, it will _always_ use that version then, even if a _newer_ version is loaded.  Removing the range just means it will upgrade lower versions to that version.

Then:

* clean up the newts and cassandra-unit dependencies, because they are complicated and mess up guava in a bunch of ways
* clean up zookeeper, curator, and cassandra-driver to avoid conflicts with newer guava
* upgrade guava to 25.1-jre: this is the meat of the PR and fixes a number of incompatibilities by updating versions, making a few small code changes, etc.

Please note that wile 25.1-jre is still covered by a CVE related to `Files.createTempDir()`, but I've looked through the cassandra, newts, and opennms source and it is only used in test code, not runtime. I also did a heap dump against a running system and found no usage of that class, so it should be safe for our uses.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14209


RangerRick:
Also there's lots of reviewers just for the sake of eyeballs, I'd love it if everyone looks at it, but at least an approval or two should be enough to merge, it seems to be doing pretty well.

cgorantla (approved):
LGTM!   If it passes all smoke tests.

Since there are cassandra driver related changes, we may need to ensure nothing breaks on newts persistence.

If that is covered, I think we are good!

RangerRick:
I'm going to close this because I want to rework the older branches a bit in light of the changes I made in H30.

PR #4836, opened 2022-06-06
NMS-14321: Only mark alarms as 'outstanding' after successful callback

amay94:
The producer was adding all alarms it attempted to send to the `outstandingAlarms` map, regardless of if the send was successful or not. On subsequent attempts to send the alarm, it would be falsely flagged as an incremental update, and by default we don't send incremental alarm updates. 

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14321


RangerRick (approved):
Hah, good catch.

cgorantla (commented):
Good catch. 

This needs to re-targeted to  `foundation-2019`.

I won't hold this for merging but it would be good to add a test case in `KafkaForwarderIT`.
you can stop/start `JUnitKafkaServer` to reproduce this.

amay94:
Fixed the failed test and rebased to `foundation-2019`

RangerRick (approved):

PR #4837, opened 2022-06-06
NMS-14282: Enable multiline text entry for requisition metadata values

synqotik:
NMS-14282. Enable multiline text entry for requisition Metadata values.

In Requisition UI, replaces the text input control for entering the Metadata value with a textarea. Initial it displays 1 row, but expands to up to 3 rows if user enters 3+ lines of text (then it scrolls).

For now, always has this new multiline textarea input. It's possible this is only desired for certain types of metadata, e.g. SSH keys or DCB scripts. Can add that if needed due to following issue.

One issue is that hitting 'Enter' key inside the 'value' textarea no longer saves the Metadata item, but would add another line. User would need to tab out or else click 'Save' button.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14282


fooker (approved):
LGTM

PR #4844, opened 2022-06-07
Fix Smoke Test for GraphMLTopologyIT

rssntn67:
* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14374


RangerRick:
@rssntn67 does this apply to earlier branches too? can it be backported to `release-30.x` or `foundation-2022` (or earlier)?

cgorantla:
> @rssntn67 does this apply to earlier branches too? can it be backported to `release-30.x` or `foundation-2022` (or earlier)?

I think this is only broken on `develop`

PR #4853, opened 2022-06-10
NMS-14243: Fixed Heatmap

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-14243


amay94 (dismissed):
looks good to me

RangerRick (changes requested):
Looks OK, but can we make sure only `release-30.x` is affected before merging? The security fixes that broke it go all the way back to Meridian 2019 so it seems like this should be fixed in at least `foundation-2022` if not older, assuming it applies to them...

christianpape:
> Looks OK, but can we make sure only `release-30.x` is affected before merging? The security fixes that broke it go all the way back to Meridian 2019 so it seems like this should be fixed in at least `foundation-2022` if not older, assuming it applies to them...

You're right. Rebased to foundation-2019.

RangerRick (approved):

PR #4854, opened 2022-06-10
Bridge Topology Discovery Mismatch

rssntn67:
* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14322

Cleaning up a wrong condition that lead to wrong bridge topology discovery.
The Tests has been fixed accordingly.


cgorantla (approved):
LGTM!

PR #4881, opened 2022-06-14
NMS-13834: Modified opennms.spec and debian/rules to also create web-inf-pristine

amay94:
This creates a new directory `web-inf-pristine` at the same level as `etc-pristine` that contains  

- `jetty-webapps/opennms/WEB-INF/applicationContext-spring-security.xml`
- `jetty-webapps/opennms/WEB-INF/web.xml`
- `jetty-webapps/opennms/WEB-INF/spring-security.d/`
- `etc/examples/jetty.xml`

This is installed at `/var/lib/opennms/web-inf-pristine` on debian systems and at `/var/opennms/web-inf-pristine` on centos.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13834

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -64,6 +64,11 @@ install: build
 	rsync -avr --exclude=examples debian/temp/etc/ debian/opennms-common/var/lib/opennms/etc-pristine/
 	chmod -R go-w debian/opennms-common/var/lib/opennms/etc-pristine/
 
+	mkdir -p debian/opennms-common/var/lib/opennms/spring-security-pristine/spring-security.d/
+	rsync -avr --exclude=examples debian/temp/jetty-webapps/opennms/WEB-INF/spring-security.d/ debian/opennms-common/var/lib/opennms/spring-security-pristine/
+	cp debian/temp/jetty-webapps/opennms/WEB-INF/applicationContext-spring-security.xml debian/opennms-common/var/lib/opennms/spring-security-pristine/applicationContext-spring-security.xml
+	chmod -R go-w debian/opennms-common/var/lib/opennms/spring-security-pristine/
+	
 	##
 	## Setup the alarm northbounders
 	##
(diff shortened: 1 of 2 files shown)


amay94:
@christianpape The only problem in what you had done was spaces instead of tabs in `debian/rules`.

christianpape:
> @christianpape The only problem in what you had done was spaces instead of tabs in `debian/rules`.

Really? Damn!

RangerRick:
This is a good idea, but if we're doing it, why not make a `web-inf-pristine` and include `web.xml` too?

dino2gnt:
> why not make a `web-inf-pristine` and include `web.xml` too?

I was about to ask, "Is there a config where `web.xml ` should be edited by a user that i can't think of?" and then I searched my chat history, which told me that not only have customers modified this, but I've had to fix it when they botched it, and that may be why I opened the jira in the first place :joy:

amay94:
> This is a good idea, but if we're doing it, why not make a `web-inf-pristine` and include `web.xml` too?

By this do you mean a third `pristine` folder with just `web.xml` in it, or change `spring-security-pristine` to be a more generic `WEB-INF` configuration bucket?

debian/rules
```diff
@@ -64,6 +64,13 @@ install: build
 	rsync -avr --exclude=examples debian/temp/etc/ debian/opennms-common/var/lib/opennms/etc-pristine/
 	chmod -R go-w debian/opennms-common/var/lib/opennms/etc-pristine/
 
+	mkdir -p debian/opennms-common/var/lib/opennms/web-inf-pristine/spring-security.d/
+	rsync -avr --exclude=examples debian/temp/jetty-webapps/opennms/WEB-INF/spring-security.d/ debian/opennms-common/var/lib/opennms/web-inf-pristine/
```
RangerRick, line 68: If you put a `/` at the end of the "from" in rsync, then it won't create that directory in the target, so you need to include the directory name there too.  Otherwise this dumps, eg, `attribute-preauth.xml` directly into the `web-inf-pristine` directory.

```suggestion
	rsync -avr --exclude=examples debian/temp/jetty-webapps/opennms/WEB-INF/spring-security.d/ debian/opennms-common/var/lib/opennms/web-inf-pristine/spring-security.d/
```

RangerRick (approved):
+1

PR #4887, opened 2022-06-15
NMS-12981: populate ackTime field when clearing an alarm

james8hut:
** Sorry, shouldn't have created yet **

This defect occurs because when an alarm is currently cleared, none of the time fields are updated and the ackTime remains null. Therefore if the alarm is more than 5 minutes old, it will be deleted nearly immediately due to the drools rule cleanUp in alarmd.drl. After pressing Clear, the UI attempts to show the cleared alarm. There appears to then be a race condition, and the potential for the alarm to have been deleted prior to calling get.

Populating ackTime means that rule fullCleanUp will delete the alarm 24 hours after the alarm has been acknowledged.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-12981


opennms-model/src/main/java/org/opennms/netmgt/model/OnmsAlarm.java
```diff
@@ -1113,6 +1113,7 @@ public void unacknowledge(String ackUser) {
     @Override
     public void clear(String ackUser) {
         m_severity = OnmsSeverity.CLEARED;
+        acknowledge(ackUser);
```
j-white, line 1116: This solves the reported problem - but I don't think we want clears to result in automatic acknowledgements.

james8hut: So having reverted the code pressing clear on the alarm screen leaves the ackTime and ackUser fields blank on the alarms table. However there is a row placed in the acks table, and when we look at the cleared alarm in the UI it shows that an acknowledgment has taken place.

The acknowledge method just populates the ackTime and ackUser fields. Shouldn't we populate these two fields to maintain consistency?

image

image

image

image

cgorantla: I think we don't need to alter the behavior of cleared alarms.  
The issue is more about whether we can route UI page to all alarms if alarm already got deleted.

I favor current behavior which indicates that alarm got deleted and I don't think we need any changes here.
Pinging @jeffgdotorg  for his opinion.

PR #4911, opened 2022-06-21
NMS-14427: Remove XMP protocol support

j-white:
JIRA: https://issues.opennms.org/browse/NMS-14427


RangerRick (approved):
Anything that makes the codebase smaller, at this point, makes me happy. The XMP stuff should be reimplemented as an OIA plugin so we don't keep running into problems with JVM/SSL issues if it still needs to be supported.

christianpape (approved):
LGTM.

RangerRick:
@rdkgit Have you looked into making this an [OIA plugin](https://github.com/OpenNMS/opennms-integration-api)? It can probably be converted without too much trouble, and then will not need to be so tightly coupled with the OpenNMS core.

PR #4918, opened 2022-06-22
NMS-14043: Negate search terms in alarms advanced search

james8hut:
Within Alarm advanced search, it is now possible to search for the negative values of Alarm Text contains, TCP/IP address like, and NodeLabel contains.

To do this the first character in the field should be a -

If you negate the nodeLabel field, currently you only get alarms that have a different node to the negated one. You don't get alarms that aren't associated to a node. I thought I would create a pull request as I was told I was moving onto Streams after today. If required I can create another jira and explain my work to somebody else, or alternatively ask them to continue working on this branch.

Example. Alarm 1 is associated with Node 1. Alarm 2 is associated with Node 2. Alarm 3 has no node associated.
You type '-2' into the node label field. Only alarm 1 is returned in the list. I would have expected alarms 1 and 3.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14043

### DIFF
--- a/opennms-webapp/src/main/java/org/opennms/web/alarm/AlarmUtil.java
+++ b/opennms-webapp/src/main/java/org/opennms/web/alarm/AlarmUtil.java
@@ -1,8 +1,8 @@
 /*******************************************************************************
  * This file is part of OpenNMS(R).
  *
- * Copyright (C) 2002-2014 The OpenNMS Group, Inc.
- * OpenNMS(R) is Copyright (C) 1999-2014 The OpenNMS Group, Inc.
+ * Copyright (C) 2002-2022 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2022 The OpenNMS Group, Inc.
  *
  * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
  *
@@ -39,6 +39,7 @@ import org.apache.commons.lang.ArrayUtils;
 import org.hibernate.criterion.Order;
 import org.hibernate.criterion.Restrictions;
 import org.opennms.core.utils.InetAddressUtils;
+import org.opennms.core.utils.StringUtils;
 import org.opennms.core.utils.WebSecurityUtils;
 import org.opennms.netmgt.model.OnmsAlarm;
 import org.opennms.netmgt.model.OnmsCriteria;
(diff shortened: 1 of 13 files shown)


christianpape:
I would favor to also allow the severity and service fields to be negatable. Maybe we can consider checkboxes for the negations?

Regardless how the implementation looks like in the end, you need to create an issue to implement the same functionality for the event search, so that the behavior is identical and consistent at this point.

RangerRick (dismissed):
Seems reasonable to me. Would be awesome if we could negate anything, but I realize that's a whole other level of work.

mmahacek:
Is this targeting v31 or v30.next?  I noticed the JIRA references a target of "next" but this is pointing to `develop` instead of `release-30.x`.

thomasbigger584:
> Is this targeting v31 or v30.next? I noticed the JIRA references a target of "next" but this is pointing to `develop` instead of `release-30.x`.

Thanks, I was'nt totally sure were this should have been targeted. I will update this to point to release-30.x

mmahacek (commented):
Other than comments on the related docs PR https://github.com/OpenNMS/opennms/pull/4971 (specifically regarding
[comments on severity/service](https://github.com/OpenNMS/opennms/pull/4971/files#r913012253)) I don't have enough knowledge to comment on code changes here.

cgorantla (commented):
lgtm!  

  Is `-` standard for negative prefix ?
  I  think we use `!` in angular code.  @RangerRick might know better.

mmahacek:
! Is more common prefix in code though - is used more in search syntaxes. IMO

thomasbigger584:
> lgtm!
> 
> Is `-` standard for negative prefix ? I think we use `!` in angular code. @RangerRick might know better.

If we prefer the `!` symbol instead of `-` then shouldn't be a problem to change.

RangerRick:
AFAIK the angular stuff just queries the rest stuff behind the scenes? I'm not sure.

JEXL at least uses `!=` for negating, so `!` would be closer than `-`.
Not finding where we use negation elsewhere though with some quick grepping.

thomasbigger584:
- Updated `-` to `!`
- Changed Service and Severity to be Multiselect Checkboxes allowing to choose which from the list to filter the alarm list by, as per the review comment: https://github.com/OpenNMS/opennms/pull/4971#discussion_r913012253

cgorantla (approved):

RangerRick (approved):

PR #4921, opened 2022-06-22
Support ingest and querying of flows with unknown directions

j-white:
Adds the ability to ingest (marker cache) and query (elastic queries) for flows with unknown direction.


fooker (changes requested):
Conceptually, this seems to be fine.

Implementations for thresholding is missing.

fooker:
Closed in favor of multiple separate PRs

PR #4958, opened 2022-06-29
HELM-272: Fixes for Helm API properties call for affectedNodeCount and situationAlarmCount

synqotik:
HELM-272: Hard-code values for Helm Rest API properties call for affectedNodeCount and situationAlarmCount to prevent invalid HQL query. Also fix labels to be "Affected Node Count", etc.

This needs to be done in conjunction with OpenNMS/opennms-helm HELM-272, see [PR](https://github.com/OpenNMS/opennms-helm/pull/453).

These values (cf. `OnmsAlarm.java`) are derived inside the entity Java code, not from the database directly.  This was causing HQL parsing issues in `AlarmRestService.java` and `AbstractDaoRestServiceWithDTO.getPropertyValues()`, responding to the "properties" call from Grafana/Helm when user added one of these Filter fields in a Filter Panel. Also, these SearchProperties are set to orderBy = false, which also wasn't being handled in HQL query construction (cf. `AbstractDaoRestServiceWithDTO.doInHibernate`).

Having hard-coded values means the code path will return directly without attempting a Hibernate query. User should be using text boxes for these filter values anyway.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/HELM-272

### DIFF
--- a/opennms-webapp-rest/src/main/java/org/opennms/web/rest/support/SearchProperties.java
+++ b/opennms-webapp-rest/src/main/java/org/opennms/web/rest/support/SearchProperties.java
@@ -80,7 +80,13 @@ public abstract class SearchProperties {
 
 	static final SortedSet<SearchProperty> ALARM_PROPERTIES = new TreeSet<>(Arrays.asList(
 		new SearchProperty(OnmsAlarm.class, "id", "ID", INTEGER),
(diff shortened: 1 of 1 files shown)


aramosvizcarra (approved):
Looks good to me!

PR #4968, opened 2022-06-30
NMS-14467: Prefer ingressPhysicalInterface over INPUT_SNMP when processing flows

j-white:
JIRA: https://issues.opennms.org/browse/NMS-14467


christianpape (changes requested):
We try to keep the handling of the different information elements as consistent as possible. So, this behavior needs also to be adjusted in IpFixMessageBuilder.

christianpape (approved):
LGTM.

fooker (approved):

PR #4992, opened 2022-07-06
NMS-14403: Fixed conversion of query string

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-14403

Reverted the change in the Util class, since this made the query-string get double-quoted. I also went through the code and the logic handling parameters looks fine to me. So, in my opinion this revert will not introduce any security issues.

### DIFF
--- a/opennms-web-api/src/main/java/org/opennms/web/api/Util.java
+++ b/opennms-web-api/src/main/java/org/opennms/web/api/Util.java
@@ -471,7 +471,7 @@ public abstract class Util extends Object {
             buffer.deleteCharAt(0);
         }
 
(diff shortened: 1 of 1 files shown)


christianpape:
> Would it be better to fix this the other way around? Escaping at the lowest layer and removing the duplicates above...

We already escape at the lowest layer. This return statement returns the entire query string with all the parameters separated by ampersands. Encoding this will encode the ampersands between the parameters which makes no sense.

fooker (approved):

amay94 (approved):
looks good to me

RangerRick (approved):

PR #5017, opened 2022-07-09
NMS-14524: Improved pollerd startup time

j-white:
JIRA: https://issues.opennms.org/browse/NMS-14524

A few tweaks to make pollerd startup faster for on systems with large amounts of nodes and services.

This patch avoids caches being reset for every service that is being scheduled and reduces the amount of round-trips to the DB required. For a deployment in ~100k nodes and ~300k monitored services, this takes pollerd startup time from several hours to < 10 minutes.


cgorantla (approved):
LGTM!

RangerRick (approved):

christianpape (approved):
Looks good to me.

PR #5031, opened 2022-07-12
Minion test revert to smoke tests

chuklov:
Adding back Minion tests to the Smoke-test as they not flaky for the last 2 weeks

### External References

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-14365


RangerRick (approved):
interested to see if these actually stay un-flaky

PR #5049, opened 2022-07-15
NMS-14470: Don't persist resources that are marked as such

amay94:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14470

### DIFF
--- a/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/collection/KafkaPersister.java
+++ b/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/collection/KafkaPersister.java
@@ -40,6 +40,7 @@ import org.opennms.netmgt.collection.api.CollectionAttribute;
 import org.opennms.netmgt.collection.api.CollectionResource;
 import org.opennms.netmgt.collection.api.CollectionSet;
 import org.opennms.netmgt.collection.api.Persister;
+import org.opennms.netmgt.collection.api.ServiceParameters;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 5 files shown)


cgorantla (commented):
Nice!  looks good to me.

Possible to add specific test case for skipping collection sets ?

amay94:
It seems straightforward, but I'm struggling to work out a way to mock this up for a unit test. The `CollectionSetBuilder` doesn't take anything implementing `Persister` as an argument, and mocking the `CollectionSetDTO` (where `Persister` classes are actually introduced) looks like it would require some changes to the visibility of fields within that class.

cgorantla:
> It seems straightforward, but I'm struggling to work out a way to mock this up for a unit test. The `CollectionSetBuilder` doesn't take anything implementing `Persister` as an argument, and mocking the `CollectionSetDTO` (where `Persister` classes are actually introduced) looks like it would require some changes to the visibility of fields within that class.

May be achievable through `KafkaPersisterIT`  but it looks like this should work!

cgorantla (approved):

amay94:
I managed to get a unit test working. I originally had the `CollectionSetMapper` returning null if the resulting `CollectionSetProto` had no resources, but I changed that so it just returns an empty collection set. Now, the `KafkaPersister` decides to send/not send based on the number of resources within the `CollectionSet`.

thomasbigger584 (commented):
Looks good to me based on the jira requirements, although i'm not too familiar with this area of the codebase yet

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/collection/KafkaPersister.java
```diff
@@ -54,20 +56,37 @@ public class KafkaPersister implements Persister {
...
     private String topicName = "metrics";
 
+    public KafkaPersister(ServiceParameters params) {
+        m_params = params;
+    }
+
+    /**
+     * Only for Mockito, DO NOT USE OTHERWISE.
+     */
+    public KafkaPersister() {
```
cgorantla, line 72: Is this getting used somewhere ?  Why can't it be handled with other constructor ?

aramosvizcarra (dismissed):
LGTM!

cgorantla (approved):

PR #5072, opened 2022-07-19
NMS-14209: Update Guava to 25.1

RangerRick:
This PR updates Guava to 25.1.

### Guava 25 is Not the Latest Version... but...

Why not 31.1? 25 is the last version to have some level of support in the Cassandra driver, before we have to move to the next major version of Cassandra client driver, which will be a significant rewrite -- mostly in the Newts codebase, which means moving to Newts 2.x, or possibly needing to bump the Newts APIs to 3.x since we expose Cassandra driver internals there.

The only CVE that is active in 25 is a temporary-directory-creation one that doesn't apply to our code, other than a few unit tests.  I looked through our dependency jars and did some runtime heap checking to be sure that a running system does not use that API even in 3rd-party dependencies.

### What this PR does

1. cleans up the Newts and Cassandra dependencies - this wasn't _strictly_ necessary, but made the dependency tree easier to manage
2. cleans up Zookeeper, Curator, and Cassandra driver dependencies -- some of these needed to be updated to be compatible with a Guava later than 18 or 19
3. fixes up dependency conflicts/overlaps between `jsr305` and `javax.annotation-api` dependencies
4. splits out a number of "core" features from Karaf that are needed across Horizon/Minion/Sentinel into a separate feature file to simplify dependency management
5. updates Guava to 25 and changes code that relies on old Guava APIs that no longer exist (or have changed)

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14209


.tool-versions
```diff
@@ -0,0 +1 @@
+java liberica-8u332+9 zulu-8.62.0.19 corretto-8.332.08.1
```
fooker, line 1: Shouldn't this be pinning to jdk 11?

features/bsm/rest/api/src/main/java/org/opennms/web/rest/v2/bsm/model/edge/AbstractEdgeResponseDTO.java
```diff
@@ -46,6 +46,7 @@
 import org.opennms.web.rest.api.support.JsonResourceLocationSerializationProvider;
 import org.opennms.web.rest.v2.bsm.model.MapFunctionDTO;
 
+import com.google.common.base.MoreObjects;
 import com.google.common.base.Objects;
```
fooker, line 50: Remove this import, too (and replace with `MoreObjects` or the jdk `Objects` alternatives.

fooker (approved):
Not tested the cassandra stuff. But looks good code-wise.

.tool-versions
RangerRick: oop, yeah, got merged forward from older versions of this PR

fooker: I like the idea behind these kinds of tooling, but this provokes me to push a `shell.nix` on my next PR :)

fooker (approved):

container/features/src/main/resources/features-core.xml
```diff
@@ -0,0 +1,83 @@
...
+>
+
+    <!--
+         This file is for features that should be shared across Horizon, Minion, and Sentinel.
+         Note that _any_ bundle dependency that has classes exported in `custom.properties` must
+         have `dependency="true"` set, so that it won't try to re-load those classes on the
+         Karaf side.
+    -->
+
+    <feature name="cassandra-driver" version="${cassandraVersion}" description="Datastax Cassandra Driver">
```
amay94, line 15: is this a new feature? It looks like the other features here were moved from other config files, but I didn't see this anywhere else.

RangerRick: Yes, it overrides the (much older) one that's in the newts feature file (which is external to our own project).

Long term, this should move back into newts, but there's a [larger discussion to be had about that](https://issues.opennms.org/browse/NEWTS-115) elsewhere.

dependencies/newts/pom.xml
```diff
@@ -31,11 +39,29 @@
...
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>com.datastax.cassandra</groupId>
+      <artifactId>cassandra-driver-core</artifactId>
+      <version>${cassandraVersion}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.raccoonfink</groupId>
+      <artifactId>high-scale-lib</artifactId>
```
amay94, line 63: how can I find what repository this is retrieved from? Something stood out about the group ID so I was curious 😉 I can't find it though

RangerRick: :D

It's in the (currently flagged as phising) opennms repo.opennms.org site, under https://repo.opennms.org/3rdparty/com/raccoonfink/high-scale-lib/

The upstream `high-scale-lib` includes overridden versions of some JDK built-in collections, which conflicts with modern JDKs. I just forked it and re-published with a new version, without the JDK bits, since cassandra doesn't rely on them.

If I have some time, I'll probably change this to a shaded lib that only pulls in the classes it needs, but for now this gets it working.

PR #5075, opened 2022-07-19
NMS-14490: Add installed OIA plugin count to usage statistics

amay94:
There aren't any direct ways of determining if a service / feature is a plugin. To get around this, this parses feature dependencies to find all features that are dependent on the `opennms-integration-api` feature, since plugins should all be built on top of this.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14490


features/datachoices/src/main/java/org/opennms/features/datachoices/internal/UsageStatisticsReporter.java
```diff
@@ -298,6 +294,85 @@ private long getDeployedRequisitionWithModifiedFSCount() {
...
+                dependencies.add(feature);
+                return dependencies;
+            }
+        }
+        // Does this feature have a transitive dependency on OIA that we haven't mapped yet?
+        for (Dependency dep : feature.getDependencies()) {
+            try {
+                int numDependencies = dependencies.size();
+                Feature depFeature = m_featuresService.getFeature(dep.getName());
+                dependencies = recurse(depFeature, dependencies, featuresToIgnore);
```
cgorantla, line 361: Is there a chance that if  there is circular dependency between two dependencies and this goes into infinite loop ?

features/datachoices/src/main/java/org/opennms/features/datachoices/internal/UsageStatisticsReporter.java
```diff
@@ -298,6 +294,85 @@ private long getDeployedRequisitionWithModifiedFSCount() {
...
+        catch (Exception e) {
+            return "ERROR: Failed to enumerate the installed features: " + e.getMessage();
+        }
+        return featuresDependentOnOIA.stream()
+                    .map(f -> f.getName() + "/" + f.getVersion())
+                    .sorted()
+                    .collect(Collectors.joining(","));
+    }
+
+    private List<Feature> recurse(Feature feature, List<Feature> dependencies, List<String> featuresToIgnore) {
```
cgorantla, line 333:
```suggestion
    private List<Feature> recurse(Feature feature, List<Feature> oiaDependentFeatures, List<String> featuresToIgnore) {
```

features/datachoices/src/main/java/org/opennms/features/datachoices/internal/UsageStatisticsReporter.java
amay94: Does karaf allow circular dependencies? If so, I could keep track of the 'visited' dependencies to avoid that.

cgorantla: I'm just concerned on recursive nature here.  We are already checking 2 levels before this routine. That seems enough ?

amay94: If there were any circular dependencies that don't have an OIA dependency hidden somewhere within, then I think this would loop forever.

features/datachoices/src/main/java/org/opennms/features/datachoices/internal/UsageStatisticsReporter.java
```diff
@@ -317,6 +315,87 @@ private void setDatasourceInfo(UsageStatisticsReportDTO usageStatisticsReport) {
...
+        }
+        // Are any of this feature's dependencies already known to be dependent on OIA?
+        for (Dependency dep : feature.getDependencies()) {
+            if (oiaDependentFeatures.stream().map(d -> d.getName()).anyMatch(str -> str.equals(dep.getName()))) {
+                oiaDependentFeatures.add(feature);
+                return oiaDependentFeatures;
+            }
+        }
+        // Don't recheck this feature in the case of circular dependencies
+        featuresToIgnore.add(feature.getName());
```
amay94, line 379: This should protect from circular dependencies looping forever

cgorantla (approved):
Thanks for the updates.  LGTM!

aramosvizcarra (commented):
LGTM!

PR #5085, opened 2022-07-21
NMS-14476: Negate search terms in event advanced search - fixing smoke test from revert

thomasbigger584:
Fixing smoke test from reverting.

Related PR: https://github.com/OpenNMS/opennms/pull/5063

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14476


RangerRick (approved):
already approved before, just needed an update to fix the smoke test 👍
thanks for looking at it

PR #5092, opened 2022-07-22
NMS-14541 Bridge Topology Discovery, Move BridgeTopologyException

rssntn67:
* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14541

This has been the last code reformat before going to review the bbridge discovery algoritm.

the code has benn organized but technically no real change has been made to the methods moved.
The idea is that bridgetopology Exception is thrown only when the algorithm fails.


christianpape (changes requested):
Some formatting changes needed.

christianpape (approved):
LGTM.

PR #5107, opened 2022-07-25
Moving threshold back to smoke test.

chuklov:
Returning back ThresholdingIT tests to the smoke test 

### External References

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-14362


RangerRick (approved):
I guess if we don't know anything about why it was failing, we might as well let it try starting up again.

PR #5109, opened 2022-07-26
NMS-14476: Fixing query param issue with advanced search

thomasbigger584:
Fixing query param issue with advanced search. 

Unsure as to how the query parameters suddenly change from `&` or `&amp;`, have seen them come through as both.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14476

### DIFF
--- a/opennms-webapp/src/main/java/org/opennms/web/filter/FilterUtil.java
+++ b/opennms-webapp/src/main/java/org/opennms/web/filter/FilterUtil.java
@@ -57,11 +57,19 @@ public abstract class FilterUtil {
 
     public static String[] parse(String filterString) {
         String decodedString = URLDecoder.decode(filterString, StandardCharsets.UTF_8);
-        if (StringUtils.isEmpty(decodedString)) {
-            return null;
-        }
-        return Arrays.stream(decodedString.split("&amp;"))
+
+        return Arrays.stream(getFilterParameters(decodedString))
                 .map(fp -> fp.replace("filter=", ""))
                 .distinct().toArray(String[]::new);
+    }
+
+    public static String[] getFilterParameters(String filterString) {
+        if (StringUtils.isEmpty(filterString)) {
+            return new String[0];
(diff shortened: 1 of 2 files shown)


cgorantla (approved):
LGTM!

thomasbigger584:
@amay94 if you can find a more elegant solution, please feel free to suggest :)

amay94:
I don't have a better solution, but I think I've isolated what is going on, at least wrt [NMS-14542](https://issues.opennms.org/browse/NMS-14542).

[This](https://github.com/OpenNMS/opennms/blob/release-30.x/opennms-webapp/src/main/webapp/WEB-INF/jsp/event/list.jsp#L635) method here is getting the correct URL from `AbstractFilterCallback.makeLink()`, but when this is embedded in an `href` attribute, all `&amp;` strings are reduced to just `&`. You can test this locally pretty easily by modifying the file locally to return something static containing `&amp;`. 

I have no clue as to why this wasn't happening earlier, though. Or if this is intended javascript behavior.

amay94 (approved):
Tested this locally, can confirm that search queries are working for alarms, events, outages, etc.

thomasbigger584:
> I don't have a better solution, but I think I've isolated what is going on, at least wrt [NMS-14542](https://issues.opennms.org/browse/NMS-14542).
> 
> [This](https://github.com/OpenNMS/opennms/blob/release-30.x/opennms-webapp/src/main/webapp/WEB-INF/jsp/event/list.jsp#L635) method here is getting the correct URL from `AbstractFilterCallback.makeLink()`, but when this is embedded in an `href` attribute, all `&amp;` strings are reduced to just `&`. You can test this locally pretty easily by modifying the file locally to return something static containing `&amp;`.
> 
> I have no clue as to why this wasn't happening earlier, though. Or if this is intended javascript behavior.

Yeah quite strange, thanks for looking into this.

PR #5112, opened 2022-07-26
NMS-14544: fix reading panels from Grafana 8 API

RangerRick:
This PR fixes the handling of the differences between the `"panels"` response when querying dashboards in different Grafana versions.  It implements a custom Gson deserializer for the `Panel` class that handles both the single `datasource` string entry, or the `datasource` object that contains a `uri` and `type`.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14544


aramosvizcarra (commented):
Looks good!

PR #5150, opened 2022-08-05
NMS-14572: create a "minimal" assembly

RangerRick:
This PR creates a minimal assembly that matches the set of packages installed in the horizon/meridian docker containers (`opennms-core`, `opennms-webapp-jetty`, `opennms-webapp-remoting`, `opennms-webapp-hawtio`)

This will facilitate updating the Docker containers to not rely on RPM/debian packages in any way. (which will be [a separate project/PR](https://issues.opennms.org/browse/NMS-14573))

I have compared the output of the "default" and "full" assemblies before and after applying these patches, and they create an identical set of files in the `target/opennms-XX.X.X.tar.gz` file in the end.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14572


opennms-full-assembly/src/assembly/core.xml
```diff
@@ -0,0 +1,16 @@
...
+  <id>core</id>
+  <includeBaseDirectory>false</includeBaseDirectory>
+  <componentDescriptors>
+    <componentDescriptor>src/assembly/components/adapters.xml</componentDescriptor>
+    <componentDescriptor>src/assembly/components/base-assembly.xml</componentDescriptor>
+    <componentDescriptor>src/assembly/components/osgi.xml</componentDescriptor>
+    <componentDescriptor>src/assembly/components/selenium-monitor.xml</componentDescriptor>
+    <componentDescriptor>src/assembly/components/webapps.xml</componentDescriptor>
+    <componentDescriptor>src/assembly/components/xml.xml</componentDescriptor>
+  </componentDescriptors>
```
aramosvizcarra, line 15: is the xml-contrib.xml file meant to be excluded from this list?

aramosvizcarra (approved):
looks good to me I found my answer to my question btw.

opennms-full-assembly/src/assembly/core.xml
RangerRick: from core? yes, it's "bonus content" :)

PR #5165, opened 2022-08-09
NMS-14576: Add new 'P' prefix to m_collect field

amay94:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14576

### DIFF
--- a/opennms-model/src/main/java/org/opennms/netmgt/model/OnmsSnmpInterface.java
+++ b/opennms-model/src/main/java/org/opennms/netmgt/model/OnmsSnmpInterface.java
@@ -120,6 +120,8 @@ public class OnmsSnmpInterface extends OnmsEntity implements Serializable {
     public static final int MAX_FLOW_AGE = SystemProperties.getInteger("org.opennms.features.telemetry.maxFlowAgeSeconds", 604800);
     public static final boolean INGRESS_AND_EGRESS_REQUIRED = Boolean.getBoolean("org.opennms.features.telemetry.ingressAndEgressRequired");
 
+    public static enum CollectDefinitionSource {NONE, USER, POLICY};
+
     /**
      * <p>Constructor for OnmsSnmpInterface.</p>
      * @param node a {@link org.opennms.netmgt.model.OnmsNode} object.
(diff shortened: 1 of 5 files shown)


opennms-model/src/main/java/org/opennms/netmgt/model/OnmsSnmpInterface.java
```diff
@@ -766,10 +805,8 @@ public void mergeSnmpInterfaceAttributes(OnmsSnmpInterface scannedSnmpIface) {
             setLastSnmpPoll(scannedSnmpIface.getLastSnmpPoll());
         }
         
-        if(scannedSnmpIface.isCollectionUserSpecified() || !isCollectionUserSpecified()){
-            setCollectionEnabled(scannedSnmpIface.isCollectionEnabled(), scannedSnmpIface.isCollectionUserSpecified());
-        }
-        
+        setCollectionEnabled(scannedSnmpIface.isCollectionEnabled(), scannedSnmpIface.getCollectionDefinitionSource());
```
cgorantla, line 808: Shouldn't this be 
```suggestion
        setCollectionEnabled(scannedSnmpIface.isCollectionEnabled(), getCollectionDefinitionSource());
```

amay94: I don't think so. I changed the behavior of `setCollectionEnabled` so that it treats `source` as the scanned interface's source. It compares the incoming source against its own internal source, and only allows overrides if the incoming source takes priority over (or matches) its internal one.

cgorantla: yeah, I understand it now after looking it one more time.  Nice work!  This should fix the bug as per my understanding.

cgorantla (approved):

RangerRick (changes requested):
Should the provisioning tests check that this gets set properly?

mmahacek:
I don't have enough knowledge to review Java PRs.

RangerRick (dismissed):

mmahacek:
The WebUI  doesn't properly show the interface as collected when set to `PC`

mmahacek (approved):
Manually tested and confirmed USER overrides POLICY which overrides NONE.
Thanks!

PR #5174, opened 2022-08-12
NMS-14591: Fix stored XSS issues

amay94:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14591

### DIFF
--- a/opennms-webapp/src/main/java/org/opennms/web/admin/discovery/ActionDiscoveryServlet.java
+++ b/opennms-webapp/src/main/java/org/opennms/web/admin/discovery/ActionDiscoveryServlet.java
@@ -206,7 +206,7 @@ public class ActionDiscoveryServlet extends HttpServlet {
         //add an 'Include URL'
         if(action.equals(addIncludeUrlAction)){
             LOG.debug("Adding Include URL");
-            String url = request.getParameter("iuurl");
+            String url = WebSecurityUtils.sanitizeString(request.getParameter("iuurl"));
             String timeout = request.getParameter("iutimeout");
             String retries = request.getParameter("iuretries");
             String foreignSource = request.getParameter("iuforeignsource");
@@ -248,7 +248,7 @@ public class ActionDiscoveryServlet extends HttpServlet {
         //add an 'Exclude URL'
         if(action.equals(addExcludeUrlAction)){
             LOG.debug("Adding Exclude URL");
-            String url = request.getParameter("euurl");
+            String url = WebSecurityUtils.sanitizeString(request.getParameter("euurl"));
             String foreignSource = request.getParameter("euforeignsource");
             String location = request.getParameter("eulocation");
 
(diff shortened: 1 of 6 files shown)


opennms-webapp/src/main/java/org/opennms/web/admin/discovery/ActionDiscoveryServlet.java
```diff
@@ -206,7 +206,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
         //add an 'Include URL'
         if(action.equals(addIncludeUrlAction)){
             LOG.debug("Adding Include URL");
-            String url = request.getParameter("iuurl");
+            String url = WebSecurityUtils.sanitizeString(request.getParameter("iuurl"));
             String timeout = request.getParameter("iutimeout");
             String retries = request.getParameter("iuretries");
             String foreignSource = request.getParameter("iuforeignsource");
```
RangerRick, line 212: Silly question, but is there a reason foreign source and location shouldn't be sanitized as well? Probably not XSS issue, but could be potentially writing weird things to the filesystem or something.

amay94: those are drop-downs, not text fields
image

amay94: and `Retries` only allows numbers, so nothing bad gets past the UI

RangerRick (approved):

opennms-webapp/src/main/java/org/opennms/web/admin/roles/AdminRoleServlet.java
```diff
@@ -274,18 +274,18 @@ private class SaveDetailsAction implements Action {
...
-                String roleName = request.getParameter("role");
+                String roleName = WebSecurityUtils.sanitizeString(request.getParameter("role"));
                 WebRole role = getRoleManager().getRole(roleName);
                 if (role == null) {
                     // this is a new role so create a new on and add it to the roleManager
                     role = getRoleManager().createRole();
                 }
-                role.setName(request.getParameter("roleName"));
+                role.setName(WebSecurityUtils.sanitizeString(request.getParameter("roleName")));
                 role.setDefaultUser(getUserManager().getUser(request.getParameter("roleUser")));
```
aramosvizcarra, line 284: shouldn't roleUser and roleGroup also be sanitized? I'm not really familiar with this workflow, I'm guessing thos are not avaliable for XSS but I'm just curious.

aramosvizcarra (commented):
Looks good to me just wondering about the other parameters that are not sanitized. I'm guessing they don't present an XSS issue, but I'd just like to confirm.

opennms-webapp/src/main/java/org/opennms/web/admin/roles/AdminRoleServlet.java
amay94: roleUser and roleGroup are populated via 'Supervisor' and 'Membership Group' in the web UI, which are both drop downs. I tried to use the description field as well to XSS, but it doesn't seem to allow it. 
image

aramosvizcarra (approved):

PR #5181, opened 2022-08-15
NMS-14643: generate SSH keypair on launch

RangerRick:
Change the SSH key generation used for Karaf console access to be executed on first launch.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14643


mershad-manesh (approved):
I took a look and the changes looks good to me.

PR #5195, opened 2022-08-18
only save cache during "real" build

RangerRick:
This PR changes a number of places where we check out the source tree to do a "read-only" checkout for speed; we only need to update the cache once per commit.


mershad-manesh (approved):
Looks good to me.

PR #5197, opened 2022-08-18
NMS-14573: use only tarballs for creating docker images

RangerRick:
This is a continuation of my work in [NMS-14572](https://issues.opennms.org/browse/NMS-14572), changing the Docker images to use tarballs and an updated base image.

* update to `opennms/deploy-base:jre-1.3.0.b138`
* convert the `horizon` and `sentinel` projects in `opennms-container/` to work like the Minion image build, with appropriate tarball sources -- this gets rid of the need to fire up an RPM or Debian server to pull in dependencies (yay!)
* update all images to fetch the latest Ubuntu security dependencies at build time
* rework CircleCI build
  * rename `MINION_IMAGE_VERSION` and reuse it for all containers
  * fix OCI caching to handle archive paths with slashes (because of arch `linux/XXXXX` container types)
  * support `amd64`, `arm64`, and `arm/v7` for all images
  * remove dependencies on debian builds for docker builds, only requires tarball assembly now

The next step is to clean up/separate pushes and signing, and normalize other packaging pushes, so that we can avoid pushing artifacts to production-like locations until we know all steps (including smoke) have passed.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14573


mershad-manesh (dismissed):
looks good to me.

RangerRick:
I was seeing a smoke test failure that should have been failing even in `develop`, so I've committed a fix here so it can go green in the smoke version.

🤷

PR #5198, opened 2022-08-18
NMS-14641: Add Usage Analytics counter

chuklov:
Added Usage Analytics counter for DCB successful and failed backups. 

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-14641

### DIFF
--- a/core/schema/src/main/liquibase/30.0.2/changelog.xml
+++ b/core/schema/src/main/liquibase/30.0.2/changelog.xml
@@ -7,8 +7,8 @@
     <changeSet runOnChange="false" author="achadfield" id="30.0.2-usage_analytics">
         <createTable tableName="usage_analytics">
             <!-- Unique identifier -->
-            <column name="id" type="bigint">
-                <constraints nullable="false" primaryKey="true"/>
+            <column name="id" type="bigint" autoIncrement="true" incrementBy="1" startWith="1">
+                <constraints nullable="false" primaryKey="true" unique="true"/>
             </column>
             <!-- metric name  -->
             <column name="metric_name" type="text">
@@ -24,7 +24,6 @@
 	        </column>
         </createTable>
 
-        <createSequence sequenceName="usage_analytics_nxtid"/>
     </changeSet>
 
 </databaseChangeLog>
--- a/core/schema/src/main/liquibase/30.0.2/changelog.xml
+++ b/core/schema/src/main/liquibase/30.0.2/changelog.xml
@@ -1,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd">
+
+    <changeSet runOnChange="false" author="achadfield" id="30.0.2-usage_analytics">
+        <createTable tableName="usage_analytics">
+            <!-- Unique identifier -->
+            <column name="id" type="bigint">
+                <constraints nullable="false" primaryKey="true"/>
+            </column>
+            <!-- metric name  -->
+            <column name="metric_name" type="text">
+                <constraints nullable="false" unique="true"/>
+            </column>
+            <!-- namespace  -->
+            <column name="namespace" type="text">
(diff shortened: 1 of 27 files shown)


features/usageanalytics/api/src/main/java/org/opennms/features/usageanalytics/api/UsageAnalyticMetricName.java
```diff
@@ -0,0 +1,39 @@
...
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.features.usageanalytics.api;
+
+/**
+ * Enumeration class to provide list of metric names.
+ */
+public enum UsageAnalyticMetricName {
```
cgorantla, line 34: Since this feature can be used in other places not just DCB,  enum wouldn't make sense here. Metric names can be constants.

cgorantla:
Any Dao packages that are created in features need to be added to https://github.com/OpenNMS/opennms/blob/develop/features/distributed/dao/api/pom.xml
https://github.com/OpenNMS/opennms/blob/develop/features/distributed/dao/impl/pom.xml

features/usageanalytics/impl/src/main/java/org/opennms/features/usageanalytics/impl/UsageAnalyticDaoImpl.java
```diff
@@ -0,0 +1,100 @@
...
+    public long getValueByMetricName(String metricName) {
+        LOG.info("Getting value of the object with metric name : " + metricName);
+
+        UsageAnalytic ua = null;
+
+        CriteriaBuilder cb = new CriteriaBuilder(UsageAnalytic.class);
+        cb.eq("metricName", metricName);
+        List<UsageAnalytic> uaList = findMatching(cb.toCriteria());
+
+        if (uaList.isEmpty())
```
cgorantla, line 66: nit: for consistency always use braces even if there is only one line.

core/schema/src/main/liquibase/30.0.2/changelog.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
cgorantla, line 1: If we are targeting this to `develop`, this should be in 31.0.0/changelog.xml

features/usageanalytics/api/src/main/java/org/opennms/features/usageanalytics/api/UsageAnalyticMetricName.java
chuklov: This feature is currently made for DCB counters. There are no plans to use it outside of this scope.  However, it is not limited to extending in the future.

core/schema/src/main/liquibase/30.0.2/changelog.xml
chuklov: It is targeted for the next version whatever it would be and discussed with @RangerRick that we start this with 30.0.2

cgorantla:
> > Any Dao packages that are created in features need to be added to https://github.com/OpenNMS/opennms/blob/develop/features/distributed/dao/api/pom.xml https://github.com/OpenNMS/opennms/blob/develop/features/distributed/dao/impl/pom.xml
> 
> @cgorantla Is this true even when it's only supposed to run in core and not minion/sentinel? For now it's only a monitor adaptor; there's nothing that runs standalone in a separate container (and in fact can't, because there's no RPC mechanism for the new DAO to be callable elsewhere).
> 
> But I don't know the distributed DAO stuff well so maybe I'm missing something.

I think it is only needed for sentinel purposes but for consistency we are adding all of DAOs into distributed as well.  There are other daos in distributed like grafana/classification which are also only available on core but they are still part of distributed dao.

cgorantla:
> > I think it is only needed for sentinel purposes but for consistency we are adding all of DAOs into distributed as well. There are other daos in distributed like grafana/classification which are also only available on core but they are still part of distributed dao.
> 
> Right now it literally _can't_ be loaded anywhere other than core, so it _shouldn't_ be "distributed," should it? The only reason this DAO is wired up into anything touching distributed at all is that's where the monitor adaptor happens to be. If you pull it into a feature file it'll fail to load at all. To me if something is in the distributed dao's dependencies, it should also be in the feature file that's including, ie, `org.opennms.features.distributed.dao-impl`, and that would blow up because there's no hibernate to inject into it.

distributed dao api/impl are like shaded jars which include all dao api artifacts and impl artifacts so it won't be a problem.

cgorantla (approved):
LGTM!

PR #5200, opened 2022-08-19
NMS-14665:Dynamic Yaml generation: Duplicate build entry when we enable flaky test and doc jobs

mershad-manesh:
We ran into an intermittent issue where processing the job dependencies would create an invalid YAML file.

In this PR, we are changing the logic used to generate the YAML file. 

Before this PR, we calculated the dependencies as we went through the job.

In the new approach, we will go through the list of dependencies for each job and figure out their requirements. Then we process them one by one. 

### All Contributors

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14665


RangerRick:
Only comment I have so far is that it looks like you reformatted the code, and it's really hard to tell the "real" fixes from the noise from reformatting; could you rebase this down to something that makes it easier to review?

RangerRick (dismissed):
a `git diff -b -w` made this easier to follow, seems OK to me

.circleci/pyscripts/generate_main.py
```diff
@@ -26,13 +27,13 @@ def combine_workflow_path(job_entry_spaces, workflow_path):
     _entries = []
     combined_entries = []
     for index, element in enumerate(workflow_path):
-        re_match = re.match("^"+(" "*job_entry_spaces)+"- ", element)
+        re_match = re.match("^" + (" " * job_entry_spaces) + "- ", element)
```
bborovan, line 30: You might want to consider using Python f-strings to make this cleaner to read:
https://docs.python.org/3/whatsnew/3.6.html#whatsnew36-pep498 

Not absolutely required, but just a note.

RangerRick (approved):
yes!

PR #5210, opened 2022-08-24
NMS-14669: Dynamic Yaml generation: Simplify and improve dynamic generation

mershad-manesh:
What has changed or are new with dynamic yaml generation:
 * By default the workflow we execute is `build-deploy` (Unless we are on develop, master, foundation or release branches, in this case we run `build-publish`)
 * `Workflows_v2.json` file has been updated to include all the required filters and an additional property `user_overridable`. If `user_overridable` is set to `false` the user won't be able to override the filters. *Note:* Ability to override filters by user has been disabled.
 * workflow_name has been updated. If we are building multiple items, the workflow name will be set to `combined-builds` instead of `multibuilds`
 * `example-build-triggers.override.json` file and the logic that parsed that file has been simplified.
 * The Git keywords start with `!` instead of `#` and they match the workflow names (under bundle) 
 * the docker images will be treated as one, so the logic to come up with build path has been simplified 
 * The YAML files under workflows folder have been deleted and are replaced by the Json file. 

### All Contributors

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14669


RangerRick (changes requested):
I think this is great, other than what I mentioned it chat. I would get rid of the `user_overrideable` stuff but also just get rid of the ability to turn filters on and off altogether. If someone needs to remove a filter, they should do it direct in the workflow; the only filters left should be the ones that make sure stuff doesn't sneak into "production" that shouldn't.

.circleci/pyscripts/generate_main.py
```diff
@@ -305,29 +242,32 @@ def combine_workflow_path(job_entry_spaces, workflow_path):
                 )
                 workflow_path = append_to_sample_workflow(workflow_path, workflow)
 
-            if build_components["build"]["build"]:
-                print("build> build:", circleCI.get_Workflow_dependency("build"))
+            if build_components["build-deploy"]:
```
RangerRick, line 245: long-term, this entire collection of "ifs" should get turned into something programmatic that just reads the workflow file and iterates over it, rather than hardcoded names, but for now, this gets it working

mershad-manesh: I agree. I didn't implement that for this PR as it would be a rather large change.

mershad-manesh:
> I think this is great, other than what I mentioned it chat. I would get rid of the `user_overrideable` stuff but also just get rid of the ability to turn filters on and off altogether. If someone needs to remove a filter, they should do it direct in the workflow; the only filters left should be the ones that make sure stuff doesn't sneak into "production" that shouldn't.

The user cannot override the filters but the script can under a specific situation.  This approach allows us to run the `smoke` tests even if the user haven't added `-smoke`  to their branch name.

RangerRick (approved):

RangerRick:
we talked about this out of band; removing the filter stuff is still on the table, but makes more sense _after_ I commit my changes cleaning up where/how publishing happens

PR #5212, opened 2022-08-25
NMS-14655: Removed scv-jceks-impl from featuresBoot 

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-14655

### DIFF
--- a/smoke-test/src/main/java/org/opennms/smoketest/containers/MinionContainer.java
+++ b/smoke-test/src/main/java/org/opennms/smoketest/containers/MinionContainer.java
@@ -72,6 +72,8 @@ import org.testcontainers.utility.MountableFile;
 
 import com.github.dockerjava.api.command.CreateContainerCmd;
 
+import joptsimple.internal.Strings;
+
 public class MinionContainer extends GenericContainer implements KarafContainer, TestLifecycleAware {
     private static final Logger LOG = LoggerFactory.getLogger(MinionContainer.class);
     private static final int MINION_DEBUG_PORT = 5005;
@@ -90,13 +92,13 @@ public class MinionContainer extends GenericContainer implements KarafContainer,
 
     private final String id;
     private final String location;
-    protected final GenericContainer container;
+    private final GenericContainer container;
 
     public interface WaitStrategyInterface {
         WaitStrategy getStrategy(MinionContainer container);
     }
 
-    protected MinionContainer(final StackModel model, final String id, final String location, final WaitStrategyInterface waitStrategyInterface) {
+    private MinionContainer(final StackModel model, final String id, final String location, final WaitStrategyInterface waitStrategyInterface) {
         super("minion");
         this.model = Objects.requireNonNull(model);
         this.id = Objects.requireNonNull(id);
@@ -123,7 +125,7 @@ public class MinionContainer extends GenericContainer implements KarafContainer,
     }
 
     public MinionContainer(final StackModel model, final MinionProfile profile) {
-        this(model, profile.getId(), profile.getLocation(), c -> new WaitForMinion(c));
+        this(model, profile.getId(), profile.getLocation(), profile.getWaitStrategy());
 
         container.addFileSystemBind(writeMinionConfig(profile).toString(),
                 "/opt/minion/minion-config.yaml", BindMode.READ_ONLY, SelinuxContext.SINGLE);
@@ -165,6 +167,14 @@ public class MinionContainer extends GenericContainer implements KarafContainer,
                 "\t\"broker-url\": \"failover:tcp://" + OpenNMSContainer.ALIAS + ":61616\"\n" +
                 "}";
         OverlayUtils.writeYaml(minionConfigYaml, jsonMapper.readValue(config, Map.class));
+
+        if (!Strings.isNullOrEmpty(profile.getDominionGrpcScvClientSecret())) {
+            final String scvConfig = "{\"scv\": {\"provider\": \"dominion\"}}";
+            OverlayUtils.writeYaml(minionConfigYaml, jsonMapper.readValue(scvConfig, Map.class));
+
+            final String gprcConfig = "{\"dominion\": { \"grpc\": { \"client-secret\":\"" + profile.getDominionGrpcScvClientSecret() + "\"}}}";
+            OverlayUtils.writeYaml(minionConfigYaml, jsonMapper.readValue(gprcConfig, Map.class));
+        }
 
         if (IpcStrategy.KAFKA.equals(model.getIpcStrategy())) {
             String kafkaIpc = "{\n" +
@@ -240,7 +250,7 @@ public class MinionContainer extends GenericContainer implements KarafContainer,
         return new InetSocketAddress(getContainerIpAddress(), mappedPort);
     }
 
-    private static class WaitForMinion extends org.testcontainers.containers.wait.strategy.AbstractWaitStrategy {
+    public static class WaitForMinion extends org.testcontainers.containers.wait.strategy.AbstractWaitStrategy {
         private final MinionContainer container;
 
         public WaitForMinion(MinionContainer container) {
--- a/smoke-test/src/main/java/org/opennms/smoketest/containers/MinionContainer.java
+++ b/smoke-test/src/main/java/org/opennms/smoketest/containers/MinionContainer.java
@@ -33,7 +33,6 @@ import static java.util.concurrent.TimeUnit.MINUTES;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.awaitility.Awaitility.await;
 import static org.hamcrest.CoreMatchers.containsString;
-import static org.opennms.smoketest.utils.KarafShellUtils.awaitHealthCheckSucceeded;
 import static org.opennms.smoketest.utils.OverlayUtils.jsonMapper;
 
 import java.io.File;
@@ -57,9 +56,9 @@ import org.opennms.smoketest.stacks.NetworkProtocol;
 import org.opennms.smoketest.stacks.StackModel;
 import org.opennms.smoketest.utils.DevDebugUtils;
 import org.opennms.smoketest.utils.OverlayUtils;
+import org.opennms.smoketest.utils.RestHealthClient;
 import org.opennms.smoketest.utils.SshClient;
 import org.opennms.smoketest.utils.TestContainerUtils;
-import org.opennms.smoketest.utils.RestHealthClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.testcontainers.containers.BindMode;
@@ -71,6 +70,8 @@ import org.testcontainers.lifecycle.TestLifecycleAware;
 import org.testcontainers.utility.MountableFile;
 
 import com.github.dockerjava.api.command.CreateContainerCmd;
+
+import joptsimple.internal.Strings;
 
 public class MinionContainer extends GenericContainer implements KarafContainer, TestLifecycleAware {
     private static final Logger LOG = LoggerFactory.getLogger(MinionContainer.class);
@@ -161,7 +162,16 @@ public class MinionContainer extends GenericContainer implements KarafContainer,
                 "\t\"broker-url\": \"failover:tcp://" + OpenNMSContainer.ALIAS + ":61616\"\n" +
                 "}";
         OverlayUtils.writeYaml(minionConfigYaml, jsonMapper.readValue(config, Map.class));
-        
+
+        if (!Strings.isNullOrEmpty(profile.getScvProvider())) {
+            String scvProvider = "{\n" +
+                    "\t\"scv\": {\n" +
+                    "\t\t\"provider\": \"" + profile.getScvProvider() + "\"\n" +
+                    "\t}\n" +
+                    "}";
+            OverlayUtils.writeYaml(minionConfigYaml, jsonMapper.readValue(scvProvider, Map.class));
+        }
+
         if (IpcStrategy.KAFKA.equals(model.getIpcStrategy())) {
             String kafkaIpc = "{\n" +
                     "\t\"ipc\": {\n" +
--- a/smoke-test/src/main/java/org/opennms/smoketest/containers/MinionContainer.java
+++ b/smoke-test/src/main/java/org/opennms/smoketest/containers/MinionContainer.java
@@ -47,6 +47,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.function.Function;
 
 import org.apache.commons.io.FileUtils;
 import org.awaitility.core.ConditionTimeoutException;
@@ -71,8 +72,7 @@ import org.testcontainers.lifecycle.TestLifecycleAware;
 import org.testcontainers.utility.MountableFile;
 
 import com.github.dockerjava.api.command.CreateContainerCmd;
-
-import joptsimple.internal.Strings;
+import com.google.common.base.Strings;
 
 public class MinionContainer extends GenericContainer implements KarafContainer, TestLifecycleAware {
     private static final Logger LOG = LoggerFactory.getLogger(MinionContainer.class);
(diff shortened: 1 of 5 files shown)


smoke-test/src/main/java/org/opennms/smoketest/stacks/MinionProfile.java
```diff
@@ -54,12 +56,17 @@ public class MinionProfile {
     private final boolean icmpSupportEnabled;
     private final List<OverlayFile> files;
 
+    private final String dominionGrpcScvClientSecret;
+    private MinionContainer.WaitStrategyInterface waitStrategyInterface;
```
fooker, line 60: Can be final?

smoke-test/src/main/java/org/opennms/smoketest/stacks/MinionProfile.java
```diff
@@ -54,12 +56,17 @@ public class MinionProfile {
...
+
     private MinionProfile(Builder builder) {
         location = builder.location;
         id = builder.id;
         jvmDebuggingEnabled = builder.jvmDebuggingEnabled;
         icmpSupportEnabled = builder.icmpSupportEnabled;
         files = Collections.unmodifiableList(builder.files);
+        dominionGrpcScvClientSecret = builder.dominionGrpcScvClientSecret;
+        waitStrategyInterface = builder.waitStrategyInterface;
```
fooker, line 69: Should be checked against `null`.

smoke-test/src/main/java/org/opennms/smoketest/containers/MinionContainer.java
```diff
@@ -92,7 +94,11 @@ public class MinionContainer extends GenericContainer implements KarafContainer,
     private final String location;
     private final GenericContainer container;
 
-    private MinionContainer(final StackModel model, final String id, final String location) {
+    public interface WaitStrategyInterface {
```
fooker, line 97: Why have this interface at all? Is there any need in getting the container which will have the `container` field uninitialized? Just use `WaitStrategy` instead?

smoke-test/src/main/java/org/opennms/smoketest/containers/MinionContainer.java
```diff
@@ -57,21 +56,24 @@
...
 import org.testcontainers.containers.Network;
 import org.testcontainers.containers.SelinuxContext;
+import org.testcontainers.containers.wait.strategy.WaitStrategy;
 import org.testcontainers.lifecycle.TestDescription;
 import org.testcontainers.lifecycle.TestLifecycleAware;
 import org.testcontainers.utility.MountableFile;
 
 import com.github.dockerjava.api.command.CreateContainerCmd;
 
+import joptsimple.internal.Strings;
```
fooker, line 75: Is this the `Strings` you want?

smoke-test/src/main/java/org/opennms/smoketest/containers/MinionContainer.java
fooker: If you opt for keeping it, it should be named `WaitStrategyProvider` or something.

smoke-test/src/main/java/org/opennms/smoketest/containers/MinionContainer.java
```diff
@@ -131,7 +137,7 @@ public MinionContainer(final StackModel model, final MinionProfile profile) {
     }
 
     public MinionContainer(final StackModel model, final Map<String, String> configuration) {
-        this(model, configuration.get("MINION_ID"), configuration.get("MINION_LOCATION"));
+        this(model, configuration.get("MINION_ID"), configuration.get("MINION_LOCATION"), c -> new WaitForMinion(c));
```
fooker, line 140:
```suggestion
        this(model, configuration.get("MINION_ID"), configuration.get("MINION_LOCATION"), WaitForMinion::new);
```

smoke-test/src/main/java/org/opennms/smoketest/containers/MinionContainer.java
fooker: Or maybe just a `Function<MinionContainer, WatiStrategy>`?

smoke-test/src/test/java/org/opennms/smoketest/minion/NMS14655_IT.java
```diff
@@ -0,0 +1,71 @@
...
+public class NMS14655_IT {
+    private static final String USERNAME = "admin";
+    private static final String PASSWORD = "admin";
+
+    @ClassRule
+    public static final MinionContainer MINION_CONTAINER = new MinionContainer(StackModel.newBuilder().build(), MinionProfile.newBuilder()
+            .withId("00000000-0000-0000-0000-000000ddba11")
+            .withLocation("Fulda")
+            .withDominionGrpcScvClientSecret("foobar")
+            .withWaitStrategy(c -> new AbstractWaitStrategy() {
```
fooker, line 32: Add a comment here for why we do not wait

smoke-test/src/test/java/org/opennms/smoketest/minion/NMS14655_IT.java
```diff
@@ -0,0 +1,71 @@
...
+                .pollInterval(5, SECONDS)
+                .until(() -> {
+                    final String bundleCount = ssh(MINION_CONTAINER.getSshAddress(), "bundle:list | grep SCV | wc -l");
+                    final String scvBundles = ssh(MINION_CONTAINER.getSshAddress(), "bundle:list | grep SCV");
+                    return bundleCount.contains("3") && scvBundles.contains("Dominion gRPC Impl") && !scvBundles.contains("JCEKS Impl");
+                });
+    }
+
+    private String ssh(final InetSocketAddress sshAddress, final String command) {
+        try (final SshClient sshClient = new SshClient(sshAddress, USERNAME, PASSWORD)) {
```
fooker, line 51: Use `MINION_CONTAINER.ssh()` instead

smoke-test/src/test/java/org/opennms/smoketest/minion/NMS14655_IT.java
```diff
@@ -0,0 +1,71 @@
...
+            final PrintStream pipe = sshClient.openShell();
+            if (command != null) {
+                pipe.println(command);
+            }
+            pipe.println("logout");
+            await().atMost(30, SECONDS).until(sshClient.isShellClosedCallable());
+            // Retrieve output and strip some ANSI sequences
+            String output = sshClient.getStdout().replaceAll("\u001B\\[[;\\d]*m", "");
```
fooker, line 59: There is `CommandTestUtils.stripAnsiCodes` for this

fooker (changes requested):

smoke-test/src/main/java/org/opennms/smoketest/containers/MinionContainer.java
christianpape: The container itself is passed to the strategy.

smoke-test/src/main/java/org/opennms/smoketest/containers/MinionContainer.java
christianpape: done.

smoke-test/src/main/java/org/opennms/smoketest/containers/MinionContainer.java
christianpape: done.

smoke-test/src/main/java/org/opennms/smoketest/stacks/MinionProfile.java
christianpape: done.

smoke-test/src/main/java/org/opennms/smoketest/stacks/MinionProfile.java
christianpape: done.

smoke-test/src/main/java/org/opennms/smoketest/containers/MinionContainer.java
christianpape: done.

smoke-test/src/test/java/org/opennms/smoketest/minion/NMS14655_IT.java
christianpape: done.

smoke-test/src/test/java/org/opennms/smoketest/minion/NMS14655_IT.java
christianpape: There exist a KarafShell class, which is used in the smoke-tests, but this will also trigger log:display for each command execution. I just wanted to check the command's output, nothing else.

smoke-test/src/test/java/org/opennms/smoketest/minion/NMS14655_IT.java
christianpape: done.

smoke-test/src/test/java/org/opennms/smoketest/minion/NMS14655_IT.java
fooker: Not what I'm talking about. `MINION_CONTAINER.ssh()` is just `new SshClient(getSshAddress(), OpenNMSContainer.ADMIN_USER, OpenNMSContainer.ADMIN_PASSWORD)`...

fooker: Not what I'm talking about. `MINION_CONTAINER.ssh()` is just `new SshClient(getSshAddress(), OpenNMSContainer.ADMIN_USER, OpenNMSContainer.ADMIN_PASSWORD)`...

christianpape: got it. done.

fooker (approved):

RangerRick:
So I see this special-cases loading `scv-jceks-impl` in the test infrastructure, but what is the appropriate way for it to end up load ing in the real world as a user? Right now the only solution is to put it in features boot -- which puts us back to where we started -- or to load it manually in the CLI.

If it's necessary for the SCV stuff to be there for dominion gRPC specifically, why not just add it as a dependency to the `dominion-secure-credentials-vault` feature with `<feature prerequisite="true">scv-jceks-impl</feature>` instead of `<feature>scv-api</feature>`? (in `container/features/src/main/resources/features-minion.xml`)

RangerRick (changes requested):
oops, forgot to mark that "request changes" for my comment

christianpape:
> So I see this special-cases loading `scv-jceks-impl` in the test infrastructure, but what is the appropriate way for it to end up load ing in the real world as a user? Right now the only solution is to put it in features boot -- which puts us back to where we started -- or to load it manually in the CLI.
> 
> If it's necessary for the SCV stuff to be there for dominion gRPC specifically, why not just add it as a dependency to the `dominion-secure-credentials-vault` feature with `<feature prerequisite="true">scv-jceks-impl</feature>` instead of `<feature>scv-api</feature>`? (in `container/features/src/main/resources/features-minion.xml`)

I've no idea why it was exactly build this way, but as far as I understand the KarafExtender is doing all the magic in real world based on the evaluation of the configuration files the confd stuff generates (see https://github.com/OpenNMS/opennms/blob/b166baaa34f6493dbf8dd08e27b6b8d15bb7712d/opennms-container/minion/container-fs/confd/templates/dominion-scv.boot.tmpl). This will load the dominion-secure-credentials-vault and cancels out the scv-jceks-impl.

RangerRick:
> I've no idea why it was exactly build this way, but as far as I understand the KarafExtender is doing all the magic in real world based on the evaluation of the configuration files the confd stuff generates (see https://github.com/OpenNMS/opennms/blob/b166baaa34f6493dbf8dd08e27b6b8d15bb7712d/opennms-container/minion/container-fs/confd/templates/dominion-scv.boot.tmpl). This will load the dominion-secure-credentials-vault and cancels out the scv-jceks-impl.

Aha! So it does get loaded, just only because of `confd`. Maybe @cgorantla or @aramosvizcarra know more about why it's done this way? They were the most recent folks to work on this.

cgorantla:
> > I've no idea why it was exactly build this way, but as far as I understand the KarafExtender is doing all the magic in real world based on the evaluation of the configuration files the confd stuff generates (see https://github.com/OpenNMS/opennms/blob/b166baaa34f6493dbf8dd08e27b6b8d15bb7712d/opennms-container/minion/container-fs/confd/templates/dominion-scv.boot.tmpl). This will load the dominion-secure-credentials-vault and cancels out the scv-jceks-impl.
> 
> Aha! So it does get loaded, just only because of `confd`. Maybe @cgorantla or @aramosvizcarra know more about why it's done this way? They were the most recent folks to work on this.

Don't think anything changed here.  There may be a race condition where conf.d removes `scv-jceks-impl` and  either features.boot or [org.apache.karaf.features.cfg](https://github.com/OpenNMS/opennms/pull/5212/files#diff-3402464888f988d37afaa24ee26b9b4d32f69885c51e23fd6416b5feac02b0c8) adds it again causing this bug.

@christianpape  Should we also move `scv-shell` from here to features.boot both on sentinel and minion ?

smoke-test/src/test/java/org/opennms/smoketest/minion/NMS14655_IT.java
```diff
@@ -0,0 +1,71 @@
...
+        await().atMost(5, MINUTES)
+                .pollInterval(5, SECONDS)
+                .until(() -> {
+                    final String bundleCount = ssh(MINION_CONTAINER.getSshAddress(), "bundle:list | grep SCV | wc -l");
+                    final String scvBundles = ssh(MINION_CONTAINER.getSshAddress(), "bundle:list | grep SCV");
+                    return bundleCount.contains("3") && scvBundles.contains("Dominion gRPC Impl") && !scvBundles.contains("JCEKS Impl");
+                });
+    }
+
+    private String ssh(final InetSocketAddress sshAddress, final String command) {
```
fooker, line 49: `sshAddress` is unused now.

fooker (changes requested):

smoke-test/src/test/java/org/opennms/smoketest/minion/NMS14655_IT.java
christianpape: done.

fooker:
> Don't think anything changed here. There may be a race condition where conf.d removes `scv-jceks-impl` and either features.boot or [org.apache.karaf.features.cfg](https://github.com/OpenNMS/opennms/pull/5212/files#diff-3402464888f988d37afaa24ee26b9b4d32f69885c51e23fd6416b5feac02b0c8) adds it again causing this bug.

As far as I understand, the karaf externder is installed by `features.boot` and in turn loads one of the two features. But if this is true, I can not explain, why it has ever worked (and we have verified it to work in 30.0.0).

> @christianpape Should we also move `scv-shell` from `org.apache.karaf.features.cfg` to features.boot both on sentinel and minion ?

I think this is a core feature that is always neede.

fooker:
> @christianpape Should we also move `scv-shell` from `org.apache.karaf.features.cfg` to features.boot both on sentinel and minion ?

I'm a little confused about this. `features.boot` is a property in `org.apache.karaf.features.cfg`. Are you talking about moving it `featuresBoot.d`?

cgorantla:
> > @christianpape Should we also move `scv-shell` from `org.apache.karaf.features.cfg` to features.boot both on sentinel and minion ?
> 
> I'm a little confused about this. `features.boot` is a property in `org.apache.karaf.features.cfg`. Are you talking about moving it `featuresBoot.d`?

Minion features.boot file already has `scv-shell` https://github.com/OpenNMS/opennms/blob/1a5326c916e8440455f103ce77f298b97db287c5/features/minion/repository/src/main/resources/features.boot#L27

and Sentinel also has it https://github.com/OpenNMS/opennms/blob/1a5326c916e8440455f103ce77f298b97db287c5/features/container/sentinel/src/main/filtered-resources/etc/org.apache.karaf.features.cfg#L56

so there is no need for it to be in shared.

cgorantla (approved):
LGTM!

RangerRick (approved):

fooker (approved):

PR #5213, opened 2022-08-25
DC-266: Get Environement and OpenNMS system Id and provide to access service

patrick-schweizer:
This PR allows OpenNMS to have any UUID as system id. So far it was assumed that the system id = "00000000-0000-0000-0000-000000000000" 
Since we are moving towards the cloud we want universally unique system ids.

It is unresolved how to create a UUID when installing a new system. Ideas?

### External References

* JIRA (Issue Tracker): [http://issues.opennms.org/browse/DC-266
](https://issues.opennms.org/browse/DC-266)


patrick-schweizer:
replaced by: https://github.com/OpenNMS/opennms/pull/5216

PR #5216, opened 2022-08-25
DC-266: Get Environement and OpenNMS system Id and provide to access service

patrick-schweizer:
This PR allows OpenNMS to have any UUID as system id. So far it was assumed that the system id = "00000000-0000-0000-0000-000000000000" 
Since we are moving towards the cloud we want universally unique system ids.

It is unresolved how to create a UUID when installing a new system. Ideas?

### External References

* JIRA (Issue Tracker): [http://issues.opennms.org/browse/DC-266
](https://issues.opennms.org/browse/DC-266)

### DIFF
--- a/opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/DistPollerDaoHibernate.java
+++ b/opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/DistPollerDaoHibernate.java
@@ -43,7 +43,7 @@ public class DistPollerDaoHibernate extends AbstractDaoHibernate<OnmsDistPoller,
     @Override
     public OnmsDistPoller whoami() {
         // Return the OnmsDistPoller
-        final String hql = "from OnmsMonitoringSystem where  type='OpenNMS' AND location='Default'";
+        final String hql = "from OnmsMonitoringSystem where type='OpenNMS'";
         return this.findUnique(hql);
     }
 }
--- a/opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/DistPollerDaoHibernate.java
+++ b/opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/DistPollerDaoHibernate.java
@@ -42,7 +42,8 @@ public class DistPollerDaoHibernate extends AbstractDaoHibernate<OnmsDistPoller,
 
     @Override
     public OnmsDistPoller whoami() {
-        // Return the OnmsDistPoller with the default UUID
-        return get(DEFAULT_DIST_POLLER_ID);
+        // Return the OnmsDistPoller
+        final String hql = "from OnmsMonitoringSystem where  type='OpenNMS' AND location='Default'";
+        return this.findUnique(hql);
     }
 }


cgorantla (commented):
id with `00000000-0000-0000-0000-000000000000` is coming from [here](https://github.com/OpenNMS/opennms/blob/b166baaa34f6493dbf8dd08e27b6b8d15bb7712d/core/schema/src/main/liquibase/17.0.0/changelog.xml#L191).

We need to investigate if it possible to set systemId after OpenNMS boots up. Then we can set unique systemId from OpenNMS Cloud.   we should also have unique Instance Id.  Both of these can be set from cloud either before startup or after startup ( if it's feasible)

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/DistPollerDaoHibernate.java
```diff
@@ -42,7 +42,8 @@ public DistPollerDaoHibernate() {
 
     @Override
     public OnmsDistPoller whoami() {
-        // Return the OnmsDistPoller with the default UUID
-        return get(DEFAULT_DIST_POLLER_ID);
```
fooker, line 46: IMO, the whole `DEFAULT_DIST_POLLER_ID` should be removed (or at least moved to the test scope).

patrick-schweizer:
> id with `00000000-0000-0000-0000-000000000000` is coming from [here](https://github.com/OpenNMS/opennms/blob/b166baaa34f6493dbf8dd08e27b6b8d15bb7712d/core/schema/src/main/liquibase/17.0.0/changelog.xml#L191).
> 
> We need to investigate if it possible to set systemId after OpenNMS boots up. Then we can set unique systemId from OpenNMS Cloud. we should also have unique Instance Id. Both of these can be set from cloud either before startup or after startup ( if it's feasible)

I am not sure how the round trip works here since I use that systemId to send it to the cloud...

fooker:
> > id with `00000000-0000-0000-0000-000000000000` is coming from [here](https://github.com/OpenNMS/opennms/blob/b166baaa34f6493dbf8dd08e27b6b8d15bb7712d/core/schema/src/main/liquibase/17.0.0/changelog.xml#L191).
> > We need to investigate if it possible to set systemId after OpenNMS boots up. Then we can set unique systemId from OpenNMS Cloud. we should also have unique Instance Id. Both of these can be set from cloud either before startup or after startup ( if it's feasible)
> 
> I am not sure how the round trip works here since I use that systemId to send it to the cloud...

Can we have a migration task or a liquibase changeset that alters the value to a random UUID if it is all-zeros?

cgorantla:
> > > id with `00000000-0000-0000-0000-000000000000` is coming from [here](https://github.com/OpenNMS/opennms/blob/b166baaa34f6493dbf8dd08e27b6b8d15bb7712d/core/schema/src/main/liquibase/17.0.0/changelog.xml#L191).
> > > We need to investigate if it possible to set systemId after OpenNMS boots up. Then we can set unique systemId from OpenNMS Cloud. we should also have unique Instance Id. Both of these can be set from cloud either before startup or after startup ( if it's feasible)
> > 
> > 
> > I am not sure how the round trip works here since I use that systemId to send it to the cloud...
> 
> Can we have a migration task or a liquibase changeset that alters the value to a random UUID if it is all-zeros?

practically it may be unique but if each system has different postgres, it is not guaranteed to be unique, right ?

patrick-schweizer:
> > > > id with `00000000-0000-0000-0000-000000000000` is coming from [here](https://github.com/OpenNMS/opennms/blob/b166baaa34f6493dbf8dd08e27b6b8d15bb7712d/core/schema/src/main/liquibase/17.0.0/changelog.xml#L191).
> > > > We need to investigate if it possible to set systemId after OpenNMS boots up. Then we can set unique systemId from OpenNMS Cloud. we should also have unique Instance Id. Both of these can be set from cloud either before startup or after startup ( if it's feasible)
> > > 
> > > 
> > > I am not sure how the round trip works here since I use that systemId to send it to the cloud...
> > 
> > 
> > Can we have a migration task or a liquibase changeset that alters the value to a random UUID if it is all-zeros?
> 
> practically it may be unique but if each system has different postgres, it is not guaranteed to be unique, right ?

with this it should be unique I believe:

`Update monitoringsystems set id=gen_random_uuid () where id = '00000000-0000-0000-0000-000000000000' AND type='OpenNMS' AND location='Default';`

fooker:
> > > > id with `00000000-0000-0000-0000-000000000000` is coming from [here](https://github.com/OpenNMS/opennms/blob/b166baaa34f6493dbf8dd08e27b6b8d15bb7712d/core/schema/src/main/liquibase/17.0.0/changelog.xml#L191).
> > > > We need to investigate if it possible to set systemId after OpenNMS boots up. Then we can set unique systemId from OpenNMS Cloud. we should also have unique Instance Id. Both of these can be set from cloud either before startup or after startup ( if it's feasible)
> > > 
> > > 
> > > I am not sure how the round trip works here since I use that systemId to send it to the cloud...
> > 
> > 
> > Can we have a migration task or a liquibase changeset that alters the value to a random UUID if it is all-zeros?
> 
> practically it may be unique but if each system has different postgres, it is not guaranteed to be unique, right ?

Not guaranteed, but chances are super low as it includes time, mac address and some other stuff.

RangerRick:
> Not guaranteed, but chances are super low as it includes time, mac address and some other stuff.

Assuming PostgreSQL implements them properly (and I would be shocked if they didn't), UUIDs should be essentially guaranteed unique.

deejgregor:
Not necessary for this ticket, but it would be good for us to include the systemId in every request we send upstream, even if it isn't particularly needed for the specific RPC being invoked. Say, as baggage. It's a *great* thing to include in the logs. And, actually, it would be good to do that consistently on every request (even if it is also included in the protobuf for the RPC request) so we can have common middleware (dare I use that word and remind me of the days of **heavyweight** middleware) that grabs it, like https://github.com/OpenNMS-Cloud/TSaaS/blob/main/gateway/server/server.go#L515 .

maximbrener:
>

maximbrener:
> > id with `00000000-0000-0000-0000-000000000000` is coming from [here](https://github.com/OpenNMS/opennms/blob/b166baaa34f6493dbf8dd08e27b6b8d15bb7712d/core/schema/src/main/liquibase/17.0.0/changelog.xml#L191).
> > We need to investigate if it possible to set systemId after OpenNMS boots up. Then we can set unique systemId from OpenNMS Cloud. we should also have unique Instance Id. Both of these can be set from cloud either before startup or after startup ( if it's feasible)
> 
> I am not sure how the round trip works here since I use that systemId to send it to the cloud...

onms instance needs to register with a cloud service (each such instance treated as a separate tenant in TSaaS). Sending onms unique ID allows us to do just that. I don't see an easy way on how cloud can manage UIDs for all onms instances. I think that UID should be part of our core product and be generated as part of onms instance setup.

fooker:
> > > id with `00000000-0000-0000-0000-000000000000` is coming from [here](https://github.com/OpenNMS/opennms/blob/b166baaa34f6493dbf8dd08e27b6b8d15bb7712d/core/schema/src/main/liquibase/17.0.0/changelog.xml#L191).
> > > We need to investigate if it possible to set systemId after OpenNMS boots up. Then we can set unique systemId from OpenNMS Cloud. we should also have unique Instance Id. Both of these can be set from cloud either before startup or after startup ( if it's feasible)
> > 
> > 
> > I am not sure how the round trip works here since I use that systemId to send it to the cloud...
> 
> onms instance needs to register with a cloud service (each such instance treated as a separate tenant in TSaaS). Sending onms unique ID allows us to do just that. I don't see an easy way on how cloud can manage UIDs for all onms instances. I think that UID should be part of our core product and be generated as part of onms instance setup.

That is exactly what we're talking about. Not sure why this was closed, then.

patrick-schweizer:
sooo, war we ok do:
`Update monitoringsystems set id=gen_random_uuid () where id = '00000000-0000-0000-0000-000000000000' AND type='OpenNMS' AND location='Default';`
in a liquibase update script?
Or do we think this is too dangerous and we somehow just set the UUID only for new installations (not sure how to do that)?

cgorantla:
> sooo, are we ok do: `Update monitoringsystems set id=gen_random_uuid () where id = '00000000-0000-0000-0000-000000000000' AND type='OpenNMS' AND location='Default';` in a liquibase update script? Or do we think this is too dangerous and we somehow just set the UUID only for new installations (not sure how to do that)?

I think it should be okay to update it but make sure this is updated in alarms/events as well https://github.com/OpenNMS/opennms/blob/1a5326c916e8440455f103ce77f298b97db287c5/core/schema/src/main/liquibase/17.0.0/changelog.xml#L202 
https://github.com/OpenNMS/opennms/blob/1a5326c916e8440455f103ce77f298b97db287c5/core/schema/src/main/liquibase/17.0.0/changelog.xml#L216

fooker:
> sooo, are we ok do: `Update monitoringsystems set id=gen_random_uuid () where id = '00000000-0000-0000-0000-000000000000' AND type='OpenNMS' AND location='Default';` in a liquibase update script? Or do we think this is too dangerous and we somehow just set the UUID only for new installations (not sure how to do that)?

Worth a try. But then this should go to develop, not release-30.

patrick-schweizer:
> > sooo, are we ok do: `Update monitoringsystems set id=gen_random_uuid () where id = '00000000-0000-0000-0000-000000000000' AND type='OpenNMS' AND location='Default';` in a liquibase update script? Or do we think this is too dangerous and we somehow just set the UUID only for new installations (not sure how to do that)?
> 
> Worth a try. But then this should go to develop, not release-30.

Are we ok to have the code change in release-30.x and the database change against develop? (I would create 2 merge requests).

Reason why I am asking:
The cloud plugin is developed against r30x and needs a unique systemid. This way we would have the possibility to change the systemid manually.

RangerRick:
> Are we ok to have the code change in release-30.x and the database change against develop? (I would create 2 merge requests).
> 
> Reason why I am asking: The cloud plugin is developed against r30x and needs a unique systemid. This way we would have the possibility to change the systemid manually.

Yeah, that makes sense to me.

We'll still want to make sure people know about the change in release notes, Just In Case people have done something funky, though.

RangerRick:
@patrick-schweizer FYI you'll want to merge latest release-30.x or your builds will probably keep failing

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/DistPollerDaoHibernate.java
j-white: That's a good point - would make sure there are no longer any references to it.

j-white:
re: alarms - no need for updates since this isn't included in the reduction key anymore: https://issues.opennms.org/browse/NMS-13210

re: events - I think the systemId column would need to be updated here - it's expected to reference a known system: https://github.com/OpenNMS/opennms/blob/develop/opennms-model/src/main/java/org/opennms/netmgt/model/OnmsEvent.java#L119

re: branch - a change like this would be more appropriate for develop, instead of `release-30.x`

fooker:
> re: branch - a change like this would be more appropriate for develop, instead of `release-30.x`

I'm also strongly for moving this to develop.

patrick-schweizer:
> @patrick-schweizer FYI you'll want to merge latest release-30.x or your builds will probably keep failing

:+1: thanks! Done.

patrick-schweizer:
> > re: branch - a change like this would be more appropriate for develop, instead of `release-30.x`
> 
> I'm also strongly for moving this to develop.

I split the changes into 2 pull requests:

1. query for systemId: this PR
2. database upgrade & removal of DEFAULT_DIST_POLLER_ID from production code: https://github.com/OpenNMS/opennms/pull/5225

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/DistPollerDaoHibernate.java
```diff
@@ -42,7 +42,8 @@ public DistPollerDaoHibernate() {
...
+        // Return the OnmsDistPoller
+        final String hql = "from OnmsMonitoringSystem where type='OpenNMS'";
```
RangerRick, line 46: This makes me nervous when we're not really guaranteeing one entry, but I don't suppose we're going to suddenly implement clustering of more than one "OpenNMS" system sharing a distpoller database anytime soon. :)

RangerRick (approved):

cgorantla (approved):

PR #5217, opened 2022-08-25
NMS-14542: Fix event search filtering

amay94:
There seemed to be two primary issues. 

1. We weren't handling both `&` and `&amp;` as parameter delimiters.
2. The EventQueryServlet was handling the `any` filter option correctly, but when redirecting to the EventController, it was passing that directly on to filters.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14542

### DIFF
--- a/opennms-webapp/src/main/java/org/opennms/web/filter/FilterUtil.java
+++ b/opennms-webapp/src/main/java/org/opennms/web/filter/FilterUtil.java
@@ -55,10 +55,18 @@ public abstract class FilterUtil {
 
     public static String[] parse(String filterString) {
         String decodedString = URLDecoder.decode(filterString, StandardCharsets.UTF_8);
-        String[] filterParameter = (decodedString == null) ? null : Arrays.stream(decodedString.split("&amp;"))
-                .filter(fp -> fp.startsWith("filter="))
+        return Arrays.stream(getFilterParameters(decodedString))
                 .map(fp -> fp.replace("filter=", ""))
                 .toArray(String[]::new);
-        return filterParameter;
+    }
+
+    public static String[] getFilterParameters(String filterString) {
+        if (StringUtils.isEmpty(filterString)) {
+            return new String[0];
+        }
+        if (filterString.contains("&amp;")) {
+            return filterString.split("&amp;");
+        }
+        return filterString.split("&");
     }
 }
--- a/opennms-webapp/src/main/java/org/opennms/web/filter/FilterUtil.java
+++ b/opennms-webapp/src/main/java/org/opennms/web/filter/FilterUtil.java
@@ -34,6 +34,8 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
+import org.opennms.core.utils.StringUtils;
+
 public abstract class FilterUtil {
 
     public static String toFilterURL(String[] filters) {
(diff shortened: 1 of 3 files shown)


RangerRick:
I'm really starting to get worried at how many times we're special-casing handling query string escaping in PRs over the last year, it makes me feel like we're missing something important somewhere deeper. There should be just one place we're decoding these in the servlet layer...

amay94:
> I'm really starting to get worried at how many times we're special-casing handling query string escaping in PRs over the last year, it makes me feel like we're missing something important somewhere deeper. There should be just one place we're decoding these in the servlet layer...

I don't disagree. It took me a while to work out that we're effectively parsing the filters twice. From what I understand, the basic paradigm is the same for both Events and Alarms. The `QueryServlet::doGet` method parses the filters into `Filter` objects, and then translates those back into a cleaner query string using `org.opennms.web.api.Util.makeQueryString()` which is then parsed again in the relevant controller class (`EventController`/`AlarmFilterController`) back into those same `Filter` objects, which are then used to construct Criteria for a DB query through the relevant `WebXXXRepository`.

To go a bit deeper, it looks like the `QueryServlet` classes attempt to remove any raw filter-specifc parameters from the parameter set in favor of specific filter types (i.e. `location=Default` translates into `filter=location%3DDefault`). When the parameters are translated back into a query string via `makeQueryString`, I think the intent is for the original `location=Default` to get removed with the `INGORE_LIST` stuff in there. 

There are a few issues here, though. If the default value is set for some filters (i.e. `location=any`), no special filter is constructed in the `QueryServlet`. Not necessarily a problem by itself, but on top of this, `location` isn't one of the parameters in the `iGNORE_LIST`, so the original parameter is forwarded to the `Controller`. And on top of THAT, the `Controller` removes the special `filter=` prefix in `FilterUtils.parse()`, so by the time we get to `EventUtil.getFilter`, there's no way to distinguish between the original `location` parameter and the `location` parameter sanitized by the Servlet, so we treat `any` as the location to check against, which is why most queries are not working. 

Actually, in laying all of this all out, it looks like a cleaner approach might be adding additional parameters to the `IGNORE_LIST` in the Servlets so they aren't passed onto the Controller. So, I guess unless that sounds like a bad idea, I'll go ahead and see if that works. 

I think I accidentally just did rubber duck debugging.

RangerRick (approved):

opennms-webapp/src/main/java/org/opennms/web/filter/FilterUtil.java
```diff
@@ -55,10 +57,18 @@ public static String toFilterURL(List<String> filters) {
 
     public static String[] parse(String filterString) {
         String decodedString = URLDecoder.decode(filterString, StandardCharsets.UTF_8);
-        String[] filterParameter = (decodedString == null) ? null : Arrays.stream(decodedString.split("&amp;"))
-                .filter(fp -> fp.startsWith("filter="))
+        return Arrays.stream(getFilterParameters(decodedString))
                 .map(fp -> fp.replace("filter=", ""))
```
christianpape, line 61: There may be non-filter parameters, right? Do we need a .filter(fp->fp.startsWith("filter=")) first?

christianpape (commented):
Just a question

opennms-webapp/src/main/java/org/opennms/web/filter/FilterUtil.java
amay94: Yes, there may be other parameters, so adding that would make sense. 

On the other hand, this array eventually makes its way to `EventUtil:getFilterList()` where any non-filter parameters are ignored, so they don't cause any issues. And this is already in `release-30.x`, so modifying it would create (more) merge conflicts for Ben/Morteza to sort through. 

I'm not sure if it's worth adding. Thoughts?

christianpape: Ok

christianpape (approved):
LGTM.

PR #5223, opened 2022-08-26
NMS-14688: DCB Rest API and UI updates to capture when user enters DCB

synqotik:
Updates to be able to capture when a user enters the Device Configuration Backup page in the UI Preview. This will be used to update a usage analytics counter.

- Rest API updated to have an optional 'pageEnter' query param
- Query param set to 'true' in UI when DCB Vue component is mounted

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14688


cgorantla:
If we need to capture UI page entries, shouldn't that be separate Rest service ?

synqotik:
> If we need to capture UI page entries, shouldn't that be separate Rest service ?

Possibly, we should have a discussion on this.

amay94 (approved):
I can't speak to the typescript / vue changes, but the backend changes look good.

mikewrosey (approved):

synqotik:
I'll merge this once builds are green, we can discuss separate endpoint separately and also fix this if we go that route.

PR #5232, opened 2022-08-29
NMS-14449: Rest API v2 for obtaining a list of SNMP interfaces doesn't return back node id

aramosvizcarra:
NMS-14449: added node ID to SnmpInterface json object returned by the api/v2/snmpinterfaces rest

Part of work done for HELM-334

OpenNMS/opennms-js#377
and
OpenNMS/opennms-helm#492

External References
JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14449

### DIFF
--- a/opennms-model/src/main/java/org/opennms/netmgt/model/OnmsSnmpInterface.java
+++ b/opennms-model/src/main/java/org/opennms/netmgt/model/OnmsSnmpInterface.java
@@ -527,7 +527,6 @@ public class OnmsSnmpInterface extends OnmsEntity implements Serializable {
     @XmlElement(name="nodeId")
     //@XmlIDREF
     @XmlJavaTypeAdapter(NodeIdAdapter.class)
-    @JsonIgnore
     public OnmsNode getNode() {
         return m_node;
     }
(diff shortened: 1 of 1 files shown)


opennms-model/src/main/java/org/opennms/netmgt/model/OnmsSnmpInterface.java
```diff
@@ -527,7 +527,6 @@ public boolean isPollEnabled() {
     @XmlElement(name="nodeId")
     //@XmlIDREF
     @XmlJavaTypeAdapter(NodeIdAdapter.class)
-    @JsonIgnore
```
amay94, line 530: Wouldn't removing this cause the entire node to be JSON-encoded? It wouldn't cause any circular issues because OnmsSnmpInterfaces are annotated with JsonIgnore within OnmsNode, but the XML and JSON encodings would be different. Is that an issue?

fooker: We're using `JaxbAnnotationIntrospector` which makes jackson also process the XML annotations. Therefore the adapter defined above should break the loop. But I have not tested this for this particular use case.

fooker (approved):

fooker:
Please make sure to also include the issue number as a prefix to the commit.

PR #5243, opened 2022-08-31
NMS-14626: normalize docker publishing

RangerRick:
_(NOTE: This went green in [this](https://app.circleci.com/pipelines/github/OpenNMS/opennms/23531/workflows/214ffb33-e33e-4886-b900-bbda24a8a58f) build which forced publishing to be enabled; the current branch will not run the `publish` job again, since I have force-pushed the final production version of these changes now.)_

This PR cleans up how Docker images are built in CircleCI, and moves the jobs publishing to DockerHub, Azure, and Cloudsmith to a single job.

Also, Docker images will get tagged with a number of aliases for easier tracking of in-progress releases.  _Every_ image that gets pushed will get a tag based on its branch name (eg, `opennms/horizon:jira-NMS-14626-normalize-docker-publishing` or `opennms/horizon:master-30`).

Additionally, a number of special case tags are pushed:
* `master-*`: add a tag matching the POM version number (eg, `31.0.0`), a "short" version (eg, `31.0-latest`),  major version latest tag (eg, `31-latest`), plus the default Docker `latest` tag
* `release-*`: add a tag matching the POM version number (eg, `30.0.3-SNAPSHOT`), the version number + `-rc` (eg, `30.0.3-rc`), a "short" rc tag (eg, `30.0-rc`), major version rc tag (eg, `30-latest`), as well as a generic `release-candidate` tag
* `develop`: add a tag for the POM version (`31.0.0-SNAPSHOT`), the major version + `-dev` (eg, `31-dev`), as well as a generic `bleeding` tag

Other cleanup:
* `tarball-assembly` job is gone, it wasn't being used anyway
* `publish-cloudsmith` job is now just `publish`
* the DCT-related aliases are gone, they are subsumed by the publish scripts and now only have to exist in one place
* the contexts containing secret keys and such are _only_ pulled into the `publish` job; jobs doing simpler things like building images only get enough to log in to dockerhub
* the publish step has been optimized to only do a shallow github clone, and to pull down the artifacts it needs using the `download-artifacts.pl` script from `opennms-repo`
* dockerhub login(s) are fixed to use `--password-stdin` rather than passing `-p` on the CLI
* all docker image builds use the same command(s)
* ubuntu images are all using the CircleCI `ubuntu-2204:current` version now
* docker layer caching is only enabled where images get built
* general workflows have been cleaned up so that artifacts are built sooner and don't gatekeep based on test runs that may or may not happen now (eg, rpms/debs don't wait on integration test, since they're not enabled by default in side branches anyway, and publishing of _anything_ doesn't happen until the end)

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14626


mershad-manesh (dismissed):
The changes look good to me.

PR #5270, opened 2022-09-09
NMS-14670 Newly created node failes on device backup

chuklov:
Changed DeviceConfig backup to have reference to PollarConfig to avoid created of a new object. 
This resolves issue with newly created node are not able to backup until a reboot. 

### External References

* JIRA (Issue Tracker): (https://issues.opennms.org/browse/NMS-14670)

### DIFF
--- a/features/device-config/service/src/main/java/org/opennms/features/deviceconfig/service/impl/DeviceConfigServiceImpl.java
+++ b/features/device-config/service/src/main/java/org/opennms/features/deviceconfig/service/impl/DeviceConfigServiceImpl.java
@@ -271,6 +271,10 @@ public class DeviceConfigServiceImpl implements DeviceConfigService {
         this.serviceMonitorAdaptor = serviceMonitorAdaptor;
     }
 
+    public void setPollerConfig(PollerConfig pollerConfig) {
+        this.pollerConfig = pollerConfig;
+    }
+
     public PollerConfig getPollerConfig() throws IOException {
         if (this.pollerConfig == null) {
             this.pollerConfig = ReadOnlyPollerConfigManager.create();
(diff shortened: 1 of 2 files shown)


features/device-config/service/src/main/java/org/opennms/features/deviceconfig/service/impl/DeviceConfigServiceImpl.java
```diff
@@ -271,6 +271,10 @@ public void setServiceMonitorAdaptor(ServiceMonitorAdaptor serviceMonitorAdaptor
         this.serviceMonitorAdaptor = serviceMonitorAdaptor;
     }
 
+    public void setPollerConfig(PollerConfig pollerConfig) {
+        this.pollerConfig = pollerConfig;
+    }
+
     public PollerConfig getPollerConfig() throws IOException {
         if (this.pollerConfig == null) {
             this.pollerConfig = ReadOnlyPollerConfigManager.create();
```
RangerRick, line 280: If you're changing the code to require setting a poller config anyway, does this even need to exist anymore?

RangerRick (changes requested):

features/device-config/service/src/main/java/org/opennms/features/deviceconfig/service/impl/DeviceConfigServiceImpl.java
chuklov: This getter is used in many dcb parts, not sure if it needs to be there after changes, however, I left it to be on the safe side.  But you're right, we can change it to a regular getter.

RangerRick (dismissed):

cgorantla:
Should it be re-targeted towards `release-30` ?

RangerRick (approved):

PR #5301, opened 2022-09-20
don't initialize ActiveMQ connection factory stuff unless we use it

RangerRick:
During integration test we get a TON of noise output from initializing ActiveMQ whether or not we actually use it. This should make it a bitt less crazy, and shouldn't affect runtime at all, spring will still initialize when it needs it.

We should probably look into making a _ton_ more stuff lazy-init, tbh, but this one jumped out at me while I was working on cleaning up some tests.

### DIFF
--- a/core/daemon/src/main/resources/META-INF/opennms/applicationContext-daemon.xml
+++ b/core/daemon/src/main/resources/META-INF/opennms/applicationContext-daemon.xml
@@ -32,27 +32,27 @@
   <!-- Conditionally start a local ActiveMQ broker. -->
   <bean class="org.opennms.netmgt.daemon.ConditionalActiveMQContext"/>
 
-  <bean id="jmsConnectionFactory" class="org.apache.activemq.ActiveMQConnectionFactory">
+  <bean id="jmsConnectionFactory" class="org.apache.activemq.ActiveMQConnectionFactory" lazy-init="true">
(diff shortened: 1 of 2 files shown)


amay94 (approved):
Looks good. Do you have a rough estimate of how much time this saves?

mershad-manesh (approved):

dino2gnt (commented):
Looks simple enough.  Will this stop the spurious `Create pooled connection during start failed. This exception will be ignored.` on startup?

RangerRick:
> Looks good. Do you have a rough estimate of how much time this saves?

@amay94 On my M1 mac, `mvn generate-resources` takes **31s** the first time, and **2.7s** the second time. But _also_ the web build detects if `src/` is unchanged and doesn't build it again -- _however_, it will always build if the `package*.json` files or `node_modules/` dir changes. `npm ci` _always_ overwrites `node_modules/` to be sure it's internally consistent for production builds, so essentially a full run would always do a full web build as well.

If I prime `node_modules/`, a `mvn compile` takes **2m31s** the first time, and **20s** the second.

So overall, if you're having to build a chunk of the tree regularly, it will be saving you about 3 minutes every time after the first.

RangerRick:
> Create pooled connection during start

@dino2gnt Just tried it, and I am happy to say the answer is yes! 👍

dino2gnt:
Hell yeah!   Approved for days.

PR #5304, opened 2022-09-21
NMS-14718: Add no-op message processors to chained dispatchers

amay94:
When there are duplicate configured security names, we chain dispatchers together to handle them correctly. All of these chained dispatchers process every trap received. If a dispatcher receives a trap that it doesn't have a corresponding message processor for, it logs the error message referenced in the JIRA ticket.

The additional chained dispatchers were only given SNMPv3 message processors, so any v1/v2 traps would trigger the log message. This adds no-op message processors for v1/v2 traps so the log messages stop appearing.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14718


cgorantla (approved):
If you validated sending V1/V2 traps as well, this looks good to me

aramosvizcarra (approved):
LGTM!
Although I'm not familiar to this module. Looks straight forward.

christianpape (approved):
LGTM.

PR #5307, opened 2022-09-22
NMS-14659: GraphiteAdapter, allow groovy script to set the node

synqotik:
In `GraphiteAdapter`, allow the Groovy script to set the node for the timeseries data.

Currently, a `CollectionAgent` is created in `GraphiteAdapter` based on the IP address and node of whatever process sent the Graphite message to OpenNMS. This ultimately results in the timeseries record being saved with a resource ID of e.g. `snmp/{nodeId}/...`. This may not actually correlate to a node that the data should be correlated with, which may need additional processing to determine.

This update passes a `List<CollectionAgent>` into the Groovy script which does the processing. If the script adds a `CollectionAgent` to that list, the `GraphiteAdapter` will use that one instead of the one it created. This ultimately results in the resourceId of the saved record having the correct node id.

The update also passes in a `CollectionAgentFactory` and a `NodeDao` to enable the script to determine and create a `CollectionAgent` if needed.

This functionality should be backward compatible with scripts that don't use this functionality.

Unfortunately, `CollectionSet` does not allow modification of the `CollectionAgent` and it looked like adding this could affect a number of files, so the decision was made to check for the underlying type and only update the agent in known cases.

Example script which uses this functionality can be found here: [mos-cdr-graphite-telemetry-interface.groovy](https://github.com/opennms-forge/mos-cdr-processor/blob/main/assets/opennms/etc/telemetryd-adapters/mos-cdr-graphite-telemetry-interface.groovy)

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14659

### DIFF
--- a/features/collection/api/src/main/java/org/opennms/netmgt/collection/dto/CollectionSetDTO.java
+++ b/features/collection/api/src/main/java/org/opennms/netmgt/collection/dto/CollectionSetDTO.java
@@ -145,6 +145,10 @@ public class CollectionSetDTO implements CollectionSet {
         return timestamp;
     }
 
+    public void setCollectionAgent(CollectionAgentDTO agent) {
+        this.agent = agent;
+    }
+
     private Set<CollectionResource> buildCollectionResources() {
         final Set<CollectionResource> collectionResources = new LinkedHashSet<>();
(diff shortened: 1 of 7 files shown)


RangerRick (approved):
clever, I can dig it

PR #5311, opened 2022-09-22
NMS-9334 added lines to clear default and previous test from the text…

chuklov:
Added a few lines to reset default text and remove previous text that leads to the test failure. 

### External References

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-9334


christianpape (approved):
LGTM.

PR #5313, opened 2022-09-22
NMS-14763 Add Priority Executor Classes

rssntn67:
Added classes to manage priority scheduling
 Added LegacyPriorityExecutor that use PriorityQueue
 Added Executable that is a ReadyRunnable class with a priority
 Added Tests

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14763

PR #5310, opened 2022-09-22
NMS-14771 Move Common Adapter Enlinkd classes to Core

rssntn67:
Moved SnmpGetter to core snmp api
Moved Discovery to core daemon scheduler
Renamed Discovery to Schedulable

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14771

PR #5321, opened 2022-09-24
 NMS-14773 Provide SubNetwork Classes for Enhanced Linkd

rssntn67:
Added Netmask to IpInterfaceTopologyEntity
Created Subnetwork class
Produce suitable methods to get subnetworks
from IpInterfaceTopolgyEntity
Check if network is legal or not (verify that subnetwork has no duplicated ip addresses)
Also find PointToPoint Subnetworks and Loopbacks
Finally provide a priority map based on Sunetworks.
Priority is asigned starting from the first low legal subnetwork
All the node under the Subnetwork have the same priority.
Check connected node and increase priority
Two subnetworks are connec ted if there is at least a common node (intersection is not null).

The branch is from develop but has been merged with the following branches

ar/jira/NMS-14771 https://github.com/OpenNMS/opennms/pull/5310

ar/jira/NMS-14763 https://github.com/OpenNMS/opennms/pull/5313

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14773

### DIFF
--- a/features/enlinkd/service/api/src/main/java/org/opennms/netmgt/enlinkd/service/api/SubNetwork.java
+++ b/features/enlinkd/service/api/src/main/java/org/opennms/netmgt/enlinkd/service/api/SubNetwork.java
@@ -1,0 +1,146 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2013-2014 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2014 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License,
+ * or (at your option) any later version.
+ *
+ * OpenNMS(R) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with OpenNMS(R).  If not, see:
+ *      http://www.gnu.org/licenses/
+ *
+ * For more information contact:
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.netmgt.enlinkd.service.api;
+
+import java.net.InetAddress;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+import org.opennms.core.utils.InetAddressUtils;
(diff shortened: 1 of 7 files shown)


features/enlinkd/service/api/src/main/java/org/opennms/netmgt/enlinkd/service/api/SubNetwork.java
```diff
@@ -0,0 +1,146 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+import org.opennms.core.utils.InetAddressUtils;
+import org.opennms.netmgt.enlinkd.model.IpInterfaceTopologyEntity;
+import org.springframework.util.Assert;
+
+public class SubNetwork {
```
RangerRick, line 41: This whole class doesn't look thread-safe at all, and it seems possible that these things could be accessed concurrently depending on how scanning goes. Perhaps some read/write locking should be implemented around the node/interface map access?

If they'll only be accessed internally to one thread, then feel free to mark this as resolved. :)

RangerRick (changes requested):
Would like confirmation that there won't be any threading issues with this, otherwise it looks OK to me.

features/enlinkd/service/api/src/main/java/org/opennms/netmgt/enlinkd/service/api/SubNetwork.java
rssntn67: Ben, 
 actually this class is only used by EnhancedLinkd. 
 How it works...instead of scheduling collection at startup I have created Scheduled a group for every Protocol.
 So when the Thread executes on scheduled group it is created a list of "Executable" that will perform data collection, each with his own priority and send to the LegacyExecutorService.
 
 the priority is inherited from the subnetworks. Each node on the same network will have the same priority.
 
 The priorityMap is build calling the NideTopologyService.
 
 I priority Map is created for each thread and is accessed only by the thread.

features/enlinkd/service/api/src/main/java/org/opennms/netmgt/enlinkd/service/api/SubNetwork.java
```diff
@@ -0,0 +1,146 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2013-2014 The OpenNMS Group, Inc.
```
aramosvizcarra, line 4:
```suggestion
 * Copyright (C) 2013-2022 The OpenNMS Group, Inc.
```

features/enlinkd/service/api/src/main/java/org/opennms/netmgt/enlinkd/service/api/SubNetwork.java
```diff
@@ -0,0 +1,146 @@
...
+ * OpenNMS(R) is Copyright (C) 1999-2014 The OpenNMS Group, Inc.
```
aramosvizcarra, line 5:
```suggestion
 * OpenNMS(R) is Copyright (C) 1999-2022 The OpenNMS Group, Inc.
```

aramosvizcarra (changes requested):
A couple of changes to license dates. Everything else looks good.

aramosvizcarra (approved):

RangerRick (approved):
If they're only accessed/manipulated on a single thread each, then looks OK to me.

PR #5363, opened 2022-10-06
Minion/Sentinel Assembly Improvements

RangerRick:
This PR fixes the minion and sentinel `create-*.sh` scripts to _actually_ do assembly-only builds when asked to skip compile.

It also gets rid of some vestigial build stuff that hasn't been used for quite some time that should make it slightly quicker.


christianpape (approved):
LGTM.

mershad-manesh (approved):

PR #5378, opened 2022-10-11
NMS-14622: Add KPI for startup time

aramosvizcarra:
Added getStartTimeMilliseconds method to BaseOnmsMBean.java to be able to track start time for any managed bean. This way we don't just track ONMS startup time but start time for every managed bean invoked during startup.

Some managed Beans don't extend core/daemon/src/main/java/org/opennms/netmgt/daemon/AbstractSpringContextJmxServiceDaemon.java
In those cases the implementation currently will throw a NotImplementedException.

Probably this could be handle with 
core/daemon/src/main/java/org/opennms/netmgt/daemon/AbstractServiceDaemon.java

But is not required for this ticket.

Not sure if this is too much for what is needed but I was thinking it would be nice to track more initialization times ...

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14622


christianpape (approved):
LGTM.

PR #5380, opened 2022-10-12
NMS-14805: changed redirect logic in HttpsRewriteFilter

christianpape:
* JIRA: https://issues.opennms.org/browse/NMS-14805


indigo423:
Tested with an HTTPS-only reverse proxy just on port 443 to avoid regression from [NMS-13901](https://issues.opennms.org/browse/NMS-13901) and it works, see
[lnx-desk.labmonkeys.tech_Archive [22-10-13 14-42-09].har.zip](https://github.com/OpenNMS/opennms/files/9776752/lnx-desk.labmonkeys.tech_Archive.22-10-13.14-42-09.har.zip)

indigo423 (approved):
Tested to avoid regression from [NMS-13901](https://issues.opennms.org/browse/NMS-13901) LGTM see comment in the PR and the attached HAR file.

PR #5382, opened 2022-10-12
NMS-14610: Use oia poller configuration smoke

dherdt:
* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-14610


pom.xml
```diff
@@ -1747,7 +1747,7 @@
     <powermockVersion>2.0.9</powermockVersion>
     <okhttpVersion>3.10.0</okhttpVersion>
     <okioVersion>1.14.0</okioVersion>
-    <opennmsApiVersion>1.1.0-SNAPSHOT</opennmsApiVersion>
+    <opennmsApiVersion>2.0.0-SNAPSHOT</opennmsApiVersion>
```
j-white, line 1750: Can we avoiding bumping the major here?

j-white (commented):
Nice! This looks like it should work well.

Can we validate that editing scheduled outages does not result in the "extended" configuration being saved to disk?

pom.xml
dherdt: I think yes. The changes in [opennms-integration-api](https://github.com/OpenNMS/opennms-integration-api) were made in develop branch (2.0.0) ("Next" was specified in JIRA-Ticket). The changes there need to be picked from develop into release 1.x, then the version here can be also changed.

dherdt: Version was reduced

fooker (approved):

PR #5395, opened 2022-10-18
jira/NMS-14841: Fix width of toggle switch which was truncating Yes

rabzobi:
Changed width of toggle switch which was truncating Yes to Ye

https://issues.opennms.org/browse/NMS-14841


amay94:
Looks fine to me, but I don't have context for best practices regarding stuff like this. Added Scott for his input.

mmahacek (approved):
LGTM.

synqotik (commented):
This is probably fine to just make that width change.

However I did notice the same behavior on the main Settings tab, the "On/Off" toggle. This is at `core/web-assets/src/main/assets/js/apps/onms-geoservice/views/config.html`, may as well fix this as well.

PR #5404, opened 2022-10-19
NMS-14853: Reset flags before NodeScans are run

amay94:
NodeScan objects are reused for scheduled scans, so the flags that are tracked per-scan should be reset between each scan run.

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14853

### DIFF
--- a/opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/NodeScan.java
+++ b/opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/NodeScan.java
@@ -268,6 +268,7 @@ public class NodeScan implements Scan {
     @Override
     public void run(final BatchTask parent) {
         LOG.info("Scanning node {}/{}/{}", m_nodeId, m_foreignSource, m_foreignId);
+        reset();
         if (monitor != null) {
             monitor.beginScanning(this);
         }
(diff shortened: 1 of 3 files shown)


aramosvizcarra (dismissed):
It make sense. 
I'm also trying to get familiar node scans. Where are the schedule scans defined?

amay94:
Most of it is handled in the Provisioner. All scheduled scans are scheduled in Provisioner::scheduleNodeScan() which in turn uses NodeScan::schedule(). In there, the NodeScan object creates a Runnable that it then schedules in the Provisioner's m_scheduledExecutor with a specific delay and period between scans. The actual schedule that the scans follow comes from the provisionService in scheduleRescanForExistingNodes() and a few other places.

The primary route for on-demand rescans is through the FORCE_RESCAN_EVENT_UEI event, which is handled in Provisioner::handleForceRescan(). The Runnable created there uses a ForceRescanScan object, which is executed immediately and only once in m_scheduledExecutor. There are a few other eventHandler methods that follow a similar code path.

cgorantla (dismissed):
LGTM!   

Possible to add a test case ?

cgorantla (approved):

ProjectJYL:
Can this also be merged to foundation-2021?

PR #5405, opened 2022-10-19
NMS-14801: Add Menu Rest Service

synqotik:
Add a MenuRestService. This is part of the UI Preview Foldin for Horizon 31.

This service will be used by the Vue "new UI" menubar to determine how to display the menu links. Also uses mostly the same data sources as the "old" menubar so changes to `dispatcher-servlet.xml` will be picked up by both.

Does the following:

- reads parts of `dispatcher-servlet.xml` to determine most links
- Adds `RoleBasedNavBarEntry` class to allow menu items in `dispatcher-servlet.xml` to be included/excluded based on roles (works for old and new menubars)
- Rest service exposing the `MainMenu` class as Json
- 'Notices' may get removed in the future as user notifications may be handled a different way
- `MenuRequestContext` and `HttpMenuRequestContext` are abstractions so we don't have to pass around `HttpServletRequest` explicitly

Note that this PR does not include any changes to `dispatcher-servlet.xml` or `navbar.ftl`, those will be included in subsequent PRs.

Note, this and some subsequent PRs will target `feature/ui-preview-foldin` branch (which is based on `develop`). Once all the work for this Epic has been completed, this branch will be merged into `develop` and/or Horizon 31 release branch.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14801

### DIFF
--- a/opennms-webapp/src/main/java/org/opennms/web/navigate/RoleBasedNavBarEntry.java
+++ b/opennms-webapp/src/main/java/org/opennms/web/navigate/RoleBasedNavBarEntry.java
@@ -1,0 +1,61 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2022 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2022 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License,
+ * or (at your option) any later version.
+ *
+ * OpenNMS(R) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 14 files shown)


opennms-webapp/src/main/java/org/opennms/web/navigate/RoleBasedNavBarEntry.java
```diff
@@ -0,0 +1,61 @@
...
+
+    public void setRoles(String roles) {
+        this.roles = roles;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public DisplayStatus evaluate(MenuContext context) {
+        if (!Strings.isNullOrEmpty(this.roles)) {
+            List<String> roleList = Arrays.stream(this.roles.split(",")).map(String::trim).collect(Collectors.toList());
```
RangerRick, line 52: It probably doesn't really matter from a performance POV, but is there any point in bothering to store this role list, only to turn it back into a stream and compare it one time?

synqotik: Should probably just be: 
```
boolean anyMatch = Arrays.stream(this.roles.split(",")).map(String::trim).anyMatch(context::isUserInRole)
```
Would that be ok?

RangerRick (approved):

aramosvizcarra (approved):
LGTM

PR #5415, opened 2022-10-21
NMS-14754: opennms:metadata-test command is not present in Karaf shell

aramosvizcarra:
NMS-14754: added changes to features.xml and pom.xml so metadata-test command shows in karaf shell

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14754


christianpape (approved):
LGTM.

PR #5427, opened 2022-10-25
NMS-10469: Provisiond import errors don't appear as event reason

FrancescoBuc86:
* Jira: https://issues.opennms.org/browse/NMS-10469

### DIFF
--- a/opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/Provisioner.java
+++ b/opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/Provisioner.java
@@ -28,10 +28,8 @@
 
 package org.opennms.netmgt.provision.service;
 
-import static org.apache.commons.lang.StringUtils.isNotBlank;
 import static org.opennms.core.utils.InetAddressUtils.addr;
 import static org.opennms.netmgt.provision.service.lifecycle.Lifecycles.RESOURCE;
-import static org.opennms.netmgt.provision.service.operations.RequisitionImport.isValidRequisitionImport;
 
 import java.io.File;
 import java.net.InetAddress;
@@ -544,7 +542,7 @@ public class Provisioner implements SpringServiceDaemon {
                     resource = new FileSystemResource(file);
                 } else {
                     final String filename = file.getName();
-                    if (isNotBlank(filename) && filename.contains("%20")) {
+                    if (filename.contains("%20")) {
                         resource = new FileSystemResource(new File(file.getParentFile(), filename.replace("%20", " ")));
                     } else {
                         resource = new UrlResource(url);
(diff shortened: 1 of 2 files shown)


opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/Provisioner.java
```diff
@@ -542,7 +544,7 @@ public void doImport(final String url, final String rescanExisting, ProvisionMon
                     resource = new FileSystemResource(file);
                 } else {
                     final String filename = file.getName();
-                    if (filename.contains("%20")) {
+                    if (isNotBlank(filename) && filename.contains("%20")) {
```
fooker, line 547: Isn't this redundant? Something can not be empty and contains an element.

opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/operations/RequisitionImport.java
```diff
@@ -73,4 +75,7 @@ public boolean isAborted() {
         return false;
     }
 
+    public static boolean isValidRequisitionImport(RequisitionImport ri) {
+        return ofNullable(ri).isPresent() && ofNullable(ri.getRequisition()).isPresent();
+    }
```
fooker, line 80: This is misleading. For validity of a requisition, there is more than just checking for `null`.
In addition, what's wrong with a plain `ri != null`?

fooker (changes requested):
The change is a cluttered with a bunch of unrelated refactorings. The real change is just a removal of `.getMessage()`.

opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/Provisioner.java
FrancescoBuc86: Well, just trying to avoid any NPE in case the file name is null

opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/operations/RequisitionImport.java
FrancescoBuc86: Well true, but since we are using in this project at least java 11, i think we can use some of its features; I can of course rename the method to isRequisionImportPresent

fooker: I don't got why this was refactored into a separate method at the first place.

opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/Provisioner.java
fooker: `File#getName` is guaranteed to never return `null`.

fooker (changes requested):

opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/operations/RequisitionImport.java
FrancescoBuc86: Ok, let me just remove all the refactoring things, at the end that could be a separate task

FrancescoBuc86:
Ok @fooker, I think now the PR is ready, just need to figure out why the circleci fails

fooker (approved):

PR #5432, opened 2022-10-26
NMS-14653: Add 'I' option to poll flag

amay94:
This adds a new option to the `pollFlag` value in SnmpInterface DB objects. 

Currently, if an interface matched a `MatchingSnmpInterfacePolicy` with the `ENABLE_POLLING` action in the foreign source definition, then it would be marked `P`/ `Polled`. This list of `Polled` interfaces is then compared against package definitions in `snmp-interface-poller.configuration.xml`, and any not matching the criteria of their respective package are not actually polled.

This change sets the flag for those ignored interfaces to `I`, and displays their polling status as `Not Polled` in the UI.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14653


RangerRick (approved):
seems totally reasonable to me, a nice optimization

PR #5438, opened 2022-10-27
NMS-14903: Add plugins to Search API

synqotik:
Add plugin results to Search Rest API. This PR is mostly for server-side work, a subsequent PR will add to the Vue part of UI.

This is a fairly quick implementation since we have limited time, possibly could be reimplemented in the future with a full `SearchProvider` implementation.

Also adds `baseNodeUrl` to the Menu API which is needed by the Vue UI, for example to navigate to a node from the DCB screen (UI will be added in upcoming PR).

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14903


amay94 (approved):

aramosvizcarra (approved):
LGTM

PR #5455, opened 2022-11-01
NMS-13856 Remove Vaadin Geo Map

synqotik:
Remove the vaadin-based legacy Geographical Map and all related code. The Vue-based Geo Map will be referenced instead.

There may be a couple more files that could be removed, but can be done as a subsequent item.

Note, an integration test is failing on CircleCI, this seems to be due to CircleCI cache issues. We will investigate more but are looking to merge this code in advance of the upcoming 31 release.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-13856


aramosvizcarra (approved):
Looks good to me, it would be nice to add some tests at some point.

PR #5457, opened 2022-11-02
NMS-14918: Alarms and Events: filter and advanced search / method POST is not supported

aramosvizcarra:
Issue occurred on especific alarm/query and evnt/query servlets that only supported GET requests. 
When adding CSRF parameters the method switched to POST requests.

Also added the `_csrf` parameter to the ignored parameters in the search query generated

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14918


christianpape (approved):
LGTM.

fooker (approved):

PR #5458, opened 2022-11-02
NMS-14961: fix the misuse of a metric name in an example in documentation.

bahka:
The example of snmp-property doesn't populate a graph with data because it takes data from wrong place (`bytesIn` instead of the actual column).

* JIRA (Issue Tracker): https://issues.opennms.org/browse/NMS-14961


Bonrob2 (dismissed):
Thanks for fixing this. It looks good to me, but I'm not a coder, so if there are any other issues, I can't comment.

freddy61025:
thanks fixing this mistake

bahka:
@Bonrob2 sorry for that, but we decided to merge it to `release-31.x`, instead of `develop`.

Bonrob2 (approved):
LGTM! Thanks!

PR #5460, opened 2022-11-02
NMS-14803: Remove NavRail from Vue UI

synqotik:
Remove the NavigationRail component from the Vue UI. This is no longer needed since all items for classic and new UI are found in the menubars.

Note that for developers working in vite-only mode, you may need to directly enter the URL to get to the various Vue component pages/routes since the NavRail is gone and the Menubar may point to main OpenNMS host/port.

- Removed NavigationRail component
- Removed some state and actions re setting navrail open state
- Removed margins around content, Vue UI components now take up entire width of screen
- Display Search box in menubar when Vue Geographical Map is displayed
- Updated Nodes and NodeDetails components to just display a message. These should no longer be used or reachable, but didn't want to delete the code in case they are used in the future
- Added a blank `Home` component; updated Vue router to go to this by default, removed Node items. Normally this component will never be seen except for developers 

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14803


mikewrosey (approved):
LGTM

scthompsonopen (approved):
I'm approving this, but we should consider redirects if we haven't for the final version instead of the stub pages and the 'this page has been deprecated' messages.

synqotik:
> I'm approving this, but we should consider redirects if we haven't for the final version instead of the stub pages and the 'this page has been deprecated' messages.

These stub pages should not be reachable in normal usage, they don't show up in the main menubar. Will consider cleaning this up a bit more and using redirects in a subsequent PR.

PR #5461, opened 2022-11-02
NMS-14920 and NMS-14892: Update RPM/DEB package dependencies

mershad-manesh:
Updating the RPM and DEB package dependences.

### All Contributors

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14920 and http://issues.opennms.org/browse/NMS-14892

### DIFF
--- a/makedeb.sh
+++ b/makedeb.sh
@@ -126,6 +126,12 @@ function version()
         head -n 1
 }
 
+function opa_version()
+{
+    grep '<opennmsApiVersion>' pom.xml | \
+    sed -e 's,^[^>]*>,,' -e 's,<.*$,,' -e 's,-[^-]*-SNAPSHOT$,,' -e 's,-SNAPSHOT$,,' -e 's,-testing$,,' -e 's,-,.,g' | \
+    head -n 1
+}
 function skipCompile()
 {
     if $ASSEMBLY_ONLY; then echo 1; else echo 0; fi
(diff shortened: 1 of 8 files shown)


makedeb.sh
```diff
@@ -211,7 +218,9 @@ function build_opennms()
     echo
     echo "Version: " $VERSION
     echo "Release: " $RELEASE
+    echo "OPA VERSION: " $OPA_VERSION
     echo
+    sed -i "s/OPA_VERSION/$OPA_VERSION/g" debian/control
```
RangerRick, line 223: Not real excited about changing "source" files during the build. I wonder if this could be a `control.in` file instead and we translate it in `debian/rules` or something.

makedeb.sh
```diff
@@ -233,8 +242,9 @@ function build_minion()
     echo
     echo "Version: " $VERSION
     echo "Release: " $RELEASE
+    echo "OPA VERSION: " $OPA_VERSION
     echo
-
+    sed -i "s/OPA_VERSION/$OPA_VERSION/g" $TOPDIR/opennms-assemblies/minion/src/main/filtered/debian/control
```
RangerRick, line 247: This, especially, since it's in the "filtered" directory, could just have `${opennmsApiVersion}` in it directly, it should get translated at build time.

RangerRick (commented):
Also where does the `opennms-cloud-plugin` package come from? It's not part of the OpenNMS build, AFAICT. So it's an external dependency?

RangerRick (approved):
Actually I'm gonna go ahead and approve this since it should work fine, but maybe we should make a new issue to change the build in 31.0.1 to not modify stuff in the source and do it a little differently in a more maven/git-friendly way.

PR #5463, opened 2022-11-03
NMS-14730: UI Preview Foldin Feature merge to Release 31

synqotik:
Merge of the UI Preview Foldin feature to `release-31.x` branch.

### External References

Epic link:

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14730


j-white:
Looks good - the pages are working and the geomap is nice and snappy. Both UIs for the Cloud and ALEC plugin work as well. 

The only problem is observed is that when both of these are loaded at the same time, clicking the link for the ALEC plugin ends up loading the Cloud plugin. We can address this separately and I wouldn't block the merge on this one though.

synqotik:
> Looks good - the pages are working and the geomap is nice and snappy. Both UIs for the Cloud and ALEC plugin work as well.
> 
> The only problem is observed is that when both of these are loaded at the same time, clicking the link for the ALEC plugin ends up loading the Cloud plugin. We can address this separately and I wouldn't block the merge on this one though.

Thanks - we can address that Plugin link right after the merge.

PR #5466, opened 2022-11-03
NMS-14885: Add support for rows in Grafana Dashboard Report

aramosvizcarra:
Modified grafana jasper report templates and grafana jasper datasource  to print panels type row.

Added a new field **rowTitlte** to each grafana report template (.jrxml) and setup some rules, when the there is no row panel content it will remove the textField element from the report (see image)

Screen Shot 2022-11-04 at 7 52 20 AM

Test report

Screen Shot 2022-11-04 at 7 59 22 AM

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14885


RangerRick (approved):
It would be nice if the indentation of jasperreports reformatting the thing was a separate commit, but other than that this is fine.

I am sorry, in advance, for the fact that this one commit now makes you the subject matter expert for all of our JasperReports stuff. 😅

PR #5472, opened 2022-11-04
Feature smoke test improvements upgrade selenium

deejgregor:
This is a replacement for #5444. but also upgrading Selenium to 4.6.0, now that it has been released and seleniarm-docker containers are also available.

### Overview

This is one of these situations where “let me look into this one test failure” turned into “fix ten various things on a path to getting visibility into the original problem.” This should be review commit-by-commit.

#### Arm64 support issues

Using native Docker images is pretty critical for the Selenium testing because testcontainers has a number of hard-coded timeouts. The timeouts combined with amd64 images that require QEMU emulation are completely unusable. Being able to use arm64 images on M1 macs is a requirement for anyone on that platform (read: me), hence working on the PR.

The versions of these Docker containers aren’t available for arm64:
1. selenium/standalone-firefox — use multi-arch “seleniarm” versions. These are only available for selenium 4.x, but we were on selenium 3.x, so it required upgrading selenium as well. See: https://github.com/seleniumhq-community/docker-seleniarm
2. testcontainers/vnc-recorder — I made a small tweak and it built fine on arm64: https://github.com/deejgregor/vnc-recorder/tree/feature-fix-build-on-arm64
3. elasticsearch-oss — upgraded to latest oss version (from early 2021; they aren’t releasing new elasticsearch-oss image as of 7.11)

The last image was a one line tweak. The vnc image is hard coded in testcontainers and the best way to workaround it  required using the image replacement functionality of testcontainers. The same functionality was used to substitute seleniarm. I'll work with testcontainers to get a version of vnc-recorder published, but in the meantime, people on M1 macs will need install it locally using the Dockerfile I linked to above.

#### Selenium upgrade

Needed to upgrade to Selenium 4.x to get arm64 support--seleniarm is only available in 4.x. Had to upgrade to 4.6.0 specifically to get a fix for an unreliability issue: https://github.com/SeleniumHQ/docker-selenium/issues/1507

This required making a small pile of simple long -> Duration changes.

Also had to deal with some dependency management issues with okhttp and netty. Also took care of some groovy warnings while I was there. Note: I got seemingly unrelated timeout issues when starting up the selenium container until I took care of the netty dependency issues (there were three incompatible versions of netty on the class path). This is another opportunity for a PR in testcontainers.

#### Testcontainers upgrade

I upgraded testcontainers while debugging Selenium problems. It might not be needed but it is working solid so far and is current.

This required updating how we handle UDP ports as the internals in testcontainers have changed. In particular, there is a post-container start check to make sure all configured exposed ports are actually exposed. This was very non-trivial to debug and I want to submit a PR to improve error reporting for this (in general, many timeout issues in testcontainers fail with almost no debugging context).

#### screen recording

We had screen recording disabled and for good reason. I’m going to avoid typing out a novel here, so look at the commit for details.

#### Selenium helpers

The selenium helpers were tweaked to do a few things:
1. Always scroll to checkbox elements when we need to check/uncheck. 
2. Fix how we figure out where the page where an element is. We were getting a relative location (based on the viewport of what is visible on the screen) and were treating it as an absolute offset. This was fixed to make the above change work.
3. Added support for x axis scrolling too. 

#### OpenNMSStack.NONE and LocalOpenNMS

This is very basic infrastructure for a no-op stack for local development testing with installed components running outside of Docker. This lets you avoid the multi-minute OpenNMS startup when testing things like browser changes. This is very minimal and might it work for anything other than basic Selenium tests.

### All Contributors

### Contribution Checklist

  5. update the JIRA link at the bottom of this comment to refer to the real issue number
  6. prefix your commit messages with the issue number, if possible
  7. once you've created this PR, please link to it in a comment in the JIRA issue

### What's Next?

### External References

### DIFF
--- a/smoke-test/pom.xml
+++ b/smoke-test/pom.xml
@@ -19,6 +19,10 @@
     <jersey.version>2.36</jersey.version>
     <karafVersion>4.3.6</karafVersion>
     <selenium.version>3.141.59</selenium.version>
+    <!-- seleniarm.version needs to match a date stamped release version from the docker-seleniarm releases page
+         for the same version of selenium specified in selenium.version above.
+         https://github.com/seleniumhq-community/docker-seleniarm/releases -->
+    <seleniarm.version>4.6.0-20221104</seleniarm.version>
     <testcontainers.version>1.17.5</testcontainers.version>
     <netty4Version>4.1.78.Final</netty4Version>
     <jna.version>5.8.0</jna.version>
@@ -78,7 +82,9 @@
           <version>${failsafeVersion}</version>
           <configuration>
             <!-- Don't specify argLine here or it will override the argLine property value -->
-            <systemPropertyVariables/>
+            <systemPropertyVariables>
+              <seleniarm.version>${seleniarm.version}</seleniarm.version>
+            </systemPropertyVariables>
             <!--
               Configure failsafe to put reports in the surefire unit test directory so that
               Bamboo and Sonar tally the test results properly.
--- a/smoke-test/pom.xml
+++ b/smoke-test/pom.xml
@@ -18,7 +18,7 @@
     <jackson2Version>2.12.3</jackson2Version>
     <jersey.version>2.36</jersey.version>
     <karafVersion>4.3.6</karafVersion>
-    <selenium.version>3.141.59</selenium.version>
+    <selenium.version>4.6.0</selenium.version>
     <!-- seleniarm.version needs to match a date stamped release version from the docker-seleniarm releases page
          for the same version of selenium specified in selenium.version above.
          https://github.com/seleniumhq-community/docker-seleniarm/releases -->
@@ -26,6 +26,7 @@
     <testcontainers.version>1.17.5</testcontainers.version>
     <netty4Version>4.1.78.Final</netty4Version>
     <jna.version>5.8.0</jna.version>
+    <okhttpVersion>3.10.0</okhttpVersion>
     <test.fork.count>0</test.fork.count>
     <!-- Limit ITs to 30 minutes by default -->
     <test.fork.timeout>1800</test.fork.timeout>
@@ -765,6 +766,13 @@
       <artifactId>jna</artifactId>
       <version>${jna.version}</version>
       <scope>test</scope>
+    </dependency>
+
+    <!-- for FileEditorIT -->
+    <dependency>
+      <groupId>com.squareup.okhttp3</groupId>
+      <artifactId>okhttp</artifactId>
+      <version>${okhttpVersion}</version>
     </dependency>
 
   </dependencies>
--- a/smoke-test/pom.xml
+++ b/smoke-test/pom.xml
@@ -18,7 +18,7 @@
     <jersey.version>2.36</jersey.version>
     <karafVersion>4.3.6</karafVersion>
     <selenium.version>3.141.59</selenium.version>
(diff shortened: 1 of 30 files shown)


deejgregor:
@RangerRick let me know if you want me to target this for a different branch or minimize it. There’s a mix of smaller cleanups in here along with the major work with selenium and test containers.

smoke-test/pom.xml
```diff
@@ -13,13 +13,20 @@
...
     <jackson2Version>2.12.3</jackson2Version>
     <jersey.version>2.36</jersey.version>
     <karafVersion>4.3.6</karafVersion>
-    <selenium.version>3.141.59</selenium.version>
-    <testcontainers.version>1.15.3</testcontainers.version>
+    <selenium.version>4.6.0</selenium.version>
+    <!-- seleniarm.version needs to match a date stamped release version from the docker-seleniarm releases page
+         for the same version of selenium specified in selenium.version above.
+         https://github.com/seleniumhq-community/docker-seleniarm/releases -->
+    <seleniarm.version>4.6.0-20221104</seleniarm.version>
```
RangerRick, line 25: should you make this `<seleniarm.version>${selenium.version}-20221104</seleniarm.version>` so it breaks gracefully if one gets changed an not the other?

smoke-test/src/main/java/org/opennms/smoketest/containers/OpenNMSContainer.java
```diff
@@ -467,6 +470,11 @@ public int getGeneratedUserId() {
 
     @Override
     public void afterTest(final TestDescription description, final Optional<Throwable> throwable) {
+        if (afterTestCalled) {
```
RangerRick, line 473: This is odd; should we be overriding something else instead? How does this happen?

RangerRick (approved):
A lot of changes, but looks great to me.

smoke-test/src/main/java/org/opennms/smoketest/containers/OpenNMSContainer.java
deejgregor: > This is odd

I can definitely confirm that this is odd, and it kinda drove me nuts. :)

> should we be overriding something else instead? How does this happen?

I spent a moderate amount of time digging in, and I couldn't figure out why it was happening, so I punted with this workaround. I was able to figure out that it appears the Rule that leads to this is getting called twice by JUnit: once before the container is torn down, once after, at which point we get failures, even with a successful execution.

Except, I just double-checked and this doesn't appear to be happening on `develop` and `release-31.x`. Hrm. I wonder if something else I tweaked caused this or something with the testcontainers update (or maybe it only happens on failures). If you're okay with merging this now, I can continue digging in on this on the side.

smoke-test/pom.xml
deejgregor: Eh, it's kinda six of one, half dozen of the other. The most important thing for it to match is a release in https://github.com/seleniumhq-community/docker-seleniarm/releases and it's a good idea (but not strictly mandatory, AFAICT) if that doesn't match the Selenium Java library version.

smoke-test/src/main/java/org/opennms/smoketest/containers/OpenNMSContainer.java
deejgregor: Except now, when I test, it doesn't happen at all. Hrrrrmm.... I'll call this one okay for now.

PR #5478, opened 2022-11-07
NMS-14739: Admin can't assign user to 'On-Call Role Schedule'

aramosvizcarra:
Looks like this is part of a bigger issue. When default locale is not US, ENGLISH, ROOT SimpleDateFormat might have some parsing issues.

I fixed this particular case but I'll create a new jira for the rest of OpenNMS, since we use SimpleDateFoormat in several other places.

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14739


christianpape (approved):
LGTM.

PR #5496, opened 2022-11-14
NMS-15001: Use service locator key in poller client

fooker:
The lookup of the service monitor implementation can not be done by class name, as there are delegate implementaions like the one used for OIA pollers. Therefor the pollers are now addressed by service locators only. This uses the registered poller class name, which can differ from the actual class in case of delegates. Thus allowing to find the correct implementations in the service monitor registry.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-15001

### DIFF
--- a/features/poller/shell/src/main/java/org/opennms/netmgt/poller/shell/Poll.java
+++ b/features/poller/shell/src/main/java/org/opennms/netmgt/poller/shell/Poll.java
@@ -70,6 +70,7 @@ import org.opennms.netmgt.poller.MonitoredService;
 import org.opennms.netmgt.poller.PollStatus;
 import org.opennms.netmgt.poller.PollerResponse;
 import org.opennms.netmgt.poller.ServiceMonitor;
+import org.opennms.netmgt.poller.ServiceMonitorLocator;
 import org.opennms.netmgt.poller.support.SimpleMonitoredService;
 
 @Command(scope = "opennms", name = "poll", description = "Used to invoke a monitor against a host at a specific location, or to test a service monitor definition from a given poller package.")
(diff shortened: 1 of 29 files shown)


features/poller/shell/src/main/java/org/opennms/netmgt/poller/shell/Poll.java
```diff
@@ -252,13 +253,13 @@ private String retrieveClassName(final InetAddress ipAddress, final String packa
             return null;
         }
 
-        final ServiceMonitor monitor = pollerConfig.getServiceMonitor(service.get().service.getName());
-        if (monitor == null) {
+        final Optional<ServiceMonitorLocator> monitor = pollerConfig.getServiceMonitorLocator(service.get().service.getName());
```
christianpape, line 256: should be `serviceMonitorLocator` and not `monitor`

christianpape (changes requested):

opennms-config/src/main/java/org/opennms/netmgt/config/DefaultServiceMonitorLocator.java
```diff
@@ -82,4 +84,29 @@ public String getServiceLocatorKey() {
...
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(this.m_serviceName, this.m_serviceClass);
+    }
+
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(this)
+                          .add("serviceName", this.m_serviceClass)
```
christianpape, line 108:
```suggestion
                          .add("serviceName", this.m_serviceName)
```

opennms-config/src/main/java/org/opennms/netmgt/config/PollerConfigManager.java
```diff
@@ -1171,49 +1138,40 @@ private void initializeServiceMonitors() {
         final Collection<ServiceMonitorLocator> locators = getServiceMonitorLocators();
         
         for (final ServiceMonitorLocator locator : locators) {
-            CompletableFuture<ServiceMonitor> monitorFuture = locator.getServiceMonitor(s_serviceMonitorRegistry);
-            monitorFuture.whenComplete((monitor, ex) -> {
-                m_svcMonitors.put(locator.getServiceName(), monitor);
-            });
-            if(!monitorFuture.isDone()) {
+            final var monitor = locator.getServiceMonitor(s_serviceMonitorRegistry);
+            if(monitor == null) {
```
christianpape, line 1142:
```suggestion
            if (monitor == null) {
```

christianpape (changes requested):
Only two small changes needed.

christianpape (approved):
LGTM.

PR #5511, opened 2022-11-16
NMS-15042: Use a "read" lock instead of an "upgrade" lock

j-white:
JIRA: https://issues.opennms.org/browse/NMS-15042

### DIFF
--- a/opennms-config/src/main/java/org/opennms/netmgt/config/SnmpPeerFactory.java
+++ b/opennms-config/src/main/java/org/opennms/netmgt/config/SnmpPeerFactory.java
@@ -45,6 +45,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReadWriteLock;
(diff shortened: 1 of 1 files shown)


opennms-config/src/main/java/org/opennms/netmgt/config/SnmpPeerFactory.java
```diff
@@ -109,9 +109,7 @@ public class SnmpPeerFactory implements SnmpAgentConfigFactory {
      */
     private static AtomicBoolean s_loaded = new AtomicBoolean(false);
 
-    private final ReadWriteUpdateLock m_globalLock = new ReentrantReadWriteUpdateLock();
-    private final Lock m_readLock = m_globalLock.updateLock();
```
j-white, line 113: For some reason we used an update lock here... but you can only hold 1 concurrently, so it defeats the purpose of the read lock.

RangerRick (approved):
seems reasonable to me; no idea why it was the other way before

RangerRick:
is this a thing that should go into older foundations? at least 2022?

j-white:
> is this a thing that should go into older foundations? at least 2022?

yeah, we could/should backport this as well

PR #5512, opened 2022-11-17
NMS-15044: Incorrect labels on OpenNMS-JMX collection resource types

mmahacek:
### All Contributors

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-15044


dino2gnt:
I don't have an ALEC environment handy, but i think if that is changed to ${index} instead of ${name} the right thing will happen and it won't get clobbered as part of the build.

mmahacek:
It should be `${name}` and it looks like I accidentally overwrote that value when I added the eventlog resource type.  This is just reverting that previous change.

dino2gnt (approved):

PR #5515, opened 2022-11-17
NMS-14996: Ensure Multiple Plugins Work

synqotik:
Plugins / UI Extensions did not work properly if more than one plugin was installed in OpenNMS at a time. This PR, in conjunction with PRs for ALEC and Cloud plugins, will allow all plugins to work when multiple ones are installed.

- Plugins get `window.VRouter` which is the Vue Router object used in OpenNMS Vue UI, then add their own routes to it. However, they added their routes as children of the 'Plugins' named route. Basically whichever plugin was installed first would have their routes 'win' vs. any subsequently installed plugin. In this PR, Horizon dynamically adds unique parent routes for each installed plugins on startup and plugins should be updated to add their routes under their specific parent route

- Plugin components get stored in `window[pluginId]`, but currently plugins may have the same ID which was based on plugin `moduleFileName` instead of `extensionId`. This PR will get component data out of `window[plugin.extensionId]`, so that multiple plugins are stored separately

- Plugins should have unique `extensionId` parameters in their bundle blueprint metadata for routing and component storage to work properly when multiple plugins are installed. To make things easier, `extensionId` and the filename portion of `moduleFileName` should be the same.

- Code has been added to detect if a 'legacy' plugin is installed, and then uses the 'old' behavior. This allows current ALEC and Cloud plugins (only) to work properly if they are the only UI plugin installed (current Horizon 30/31.0 behavior). Ideally this code (and the 'Plugin' route) should be eventually removed when all plugins have been updated to new pattern. Any other legacy plugins may not work.

- This only affects plugins with a UI component. Non-UI plugins (e.g. cortex) don't have any effect on this issue. 

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-14996

### DIFF
--- a/ui/src/components/Plugin/utils.ts
+++ b/ui/src/components/Plugin/utils.ts
@@ -1,10 +1,11 @@
 const externalComponent = (url: string) => {
   try {
+    // this is the extensionId part of the url
     const name = (
       url
         .split('/')
-        .reverse()[0]
-        .match(/^(.*?)\.es/) as any[]
+        .reverse()[1]
+        .match(/^([^?]+)/) as any[]
     )[1]
 
     if (window[name]) {
--- a/ui/src/components/Plugin/utils.ts
+++ b/ui/src/components/Plugin/utils.ts
@@ -1,12 +1,32 @@
+const isLegacyPluginUrl = (extensionId: string, moduleFileName: string) => {
+  if (moduleFileName == 'uiextension' &&
+     (extensionId === 'cloudUiExtension' || extensionId === 'uiextension' || extensionId === 'uiExtension')) {
+    return true
+  }
+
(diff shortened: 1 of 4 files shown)


RangerRick (approved):
I assume we're working under the assumption no end-users have ever made their own "legacy" plugins. (This is a pretty reasonable assumption.)

synqotik:
> I assume we're working under the assumption no end-users have ever made their own "legacy" plugins. (This is a pretty reasonable assumption.)

Yes definitely making this assumption for now. We also need to update documentation and any examples to make sure going forward that plugin-makers follow the new pattern.

mikewrosey (approved):

PR #5552, opened 2022-11-30
DC-407: Have a smoke test in OpenNMS repo to make sure the cloud plugin runs

FrancescoBuc86:
This PR:

- Added SmokeTest class to verify cloud plugin installation phase
- Changed RestHealthClient with additional isAvailable Method, to avoid errors responses because of the not yet started server
- Added necessaries certificates to run smoketest

JIRA (Issue Tracker): http://issues.opennms.org/browse/DC-407


FrancescoBuc86:
https://issues.opennms.org/browse/DC-407

smoke-test/src/test/resources/cert/horizon/pas_server.crt
```diff
@@ -0,0 +1,27 @@
+-----BEGIN CERTIFICATE-----
```
patrick-schweizer, line 1: just wondering: aren't the certs already part of the plugin test dependencies?

patrick-schweizer (commented):
nice that we have now a smoke test for the cloud plugin in OpenNMS :-)
I just have a few small comments / questions

smoke-test/src/main/java/org/opennms/smoketest/utils/SshClient.java
```diff
@@ -83,6 +83,10 @@ public SshClient(InetSocketAddress addr, String username, String password) {
         this.password = password;
     }
 
+    public Channel getChannel() {
```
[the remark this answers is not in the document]
FrancescoBuc86: good point: removed

smoke-test/src/test/resources/cert/horizon/pas_server.crt
FrancescoBuc86: good point, probably we don't need these files at all

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
```diff
@@ -0,0 +1,163 @@
...
+    public void setUp() {
+        this.opennmsShell = new KarafShell(stack.opennms().getSshAddress());
+        this.sentinelShell = new KarafShell(stack.sentinel().getSshAddress());
+        mockCloudContainer.start();
+    }
+
+    @Test
+    public void install_cloud_plugin_in_opennms_must_be_successful() throws Exception {
+        // Given
+        stack.opennms().copyFileToContainer(MountableFile.forHostPath(pathToFile), "/usr/share/opennms/mvn/org.opennms.plugins.cloud-assembly.kar");
```
freddy61025, line 92: why don't you copy to /usr/share/opennms/deploy ? 
It is more close to install packages. Also by put in deploy kar:install is not needed.

freddy61025 (commented):
some minor suggestions.

smoke-test/src/test/resources/cert/horizon/pas_server.crt
FrancescoBuc86: updated

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
FrancescoBuc86: Tried this, but without the kar:install command the test does not seem to work anymore

smoke-test/pom.xml
```diff
@@ -112,6 +112,52 @@
...
+            <execution>
+              <id>copy-kar</id>
+              <phase>process-resources</phase>
+              <goals><goal>copy</goal></goals>
+              <configuration>
+                <artifactItems>
+                  <artifactItem>
+                    <groupId>org.opennms.plugins.cloud.assembly</groupId>
+                    <artifactId>org.opennms.plugins.cloud.assembly.kar</artifactId>
+                    <version>1.0.0</version>
```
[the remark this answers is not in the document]
FrancescoBuc86: ok, I extracted it in a separate variable, but still inside the current pom

smoke-test/src/main/java/org/opennms/smoketest/utils/RestClient.java
```diff
@@ -172,13 +177,13 @@ public Map<String, Object> getUsageStatistics() throws Exception {
...
-                        return e.getValue();
-                    }
-                }));
+        return hashMap.entrySet().stream().collect(Collectors.toMap(e -> e.getKey(), e -> {
+            if (e.getValue() instanceof Integer) {
+                return Long.valueOf((Integer) e.getValue());
+            } else {
+                return e.getValue();
+            }
+        }));
```
fooker, line 186: Please do not always reformat.

smoke-test/src/main/java/org/opennms/smoketest/utils/RestClient.java
```diff
@@ -121,14 +122,18 @@ private static URL toUrl(InetSocketAddress addr) {
...
-
-        final ObjectMapper mapper = new ObjectMapper();
-        try {
-            JsonNode actualObj = mapper.readTree(json);
-            return actualObj.get("displayVersion").asText();
-        } catch (IOException e) {
-            LOG.debug("Failed to get displayVersion from the info REST service. (OpenNMS is probably not up yet): {}", e.getMessage());
+        boolean isReachable = checkUriAvailability(target.getUri());
+
+        if (isReachable) {
```
fooker, line 127: Can inline.

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
```diff
@@ -0,0 +1,165 @@
...
+    private static final String CONTAINER_PATH = "/usr/share/opennms/deploy/";
+    private static final String CLOUD_MOCK_JAR_NAME = "org.opennms.plugins.cloud-mock-with-dependencies.jar";
+    private static final String CLOUD_MOCK_MAIN = "org.opennms.plugins.cloud.ittest.MockCloudMain";
+    private static final String KAR_FILE_NAME = "org.opennms.plugins.cloud-assembly.kar";
+
+    private KarafShell opennmsShell;
+    private KarafShell sentinelShell;
+
+    @ClassRule
+    public static Network network = Network.newNetwork();
```
fooker, line 66: Not used?

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
```diff
@@ -0,0 +1,165 @@
...
+        install_cloud_plugin_in_sentinel_must_be_successful();
+    }
+
+    public void install_cloud_plugin_in_opennms_must_be_successful() throws Exception {
+        // Given
+        stack.opennms().copyFileToContainer(MountableFile.forHostPath(PATH_TO_FILE), format("%s%s", CONTAINER_PATH, KAR_FILE_NAME));
+        String config = createConfig();
+
+        // When
+        opennmsShell.runCommand(String.format("kar:install file:%s%s", CONTAINER_PATH, KAR_FILE_NAME));
```
fooker, line 104: There is `KarafShellUtils#withKArafShell` which ensures that better error handling and resource teardown.

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
```diff
@@ -0,0 +1,165 @@
...
+        opennmsShell.runCommand("feature:install opennms-plugin-cloud-core");
+        opennmsShell.runCommand("feature:list | grep opennms-plugin-cloud-core", output -> output.contains("Started"));
+        opennmsShell.runCommand(config);
+        opennmsShell.runCommand("opennms-cloud:init key", output -> output.contains("Initialization of cloud plugin in OPENNMS was successful."));
+    }
+
+
+    public void install_cloud_plugin_in_sentinel_must_be_successful() throws Exception {
```
fooker, line 112: Please stick to java standard method name scheme.

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
```diff
@@ -0,0 +1,165 @@
...
+    public static String classpathFileToString(final String fileInClasspath) throws IOException {
+        URL url = new URL(format("jar:file:%s!%s", PATH_TO_CLOUD_MOCK, fileInClasspath));
+        try (InputStream inputStream = url.openStream()) {
+            return new String(inputStream.readAllBytes(), StandardCharsets.UTF_8);
+        }
+    }
+
+    @AfterClass
+    public static void afterAll() {
+        if (nonNull(stack)) {
```
fooker, line 152: no need to stop explicitly. This is handled by the class rule. It may be event contra-productive as there is an option to keep the containers running intentionally for debugging purposes.

smoke-test/src/main/java/org/opennms/smoketest/utils/RestHealthClient.java
```diff
@@ -74,8 +86,11 @@ private WebTarget getTargetFor(final String path){
     }
 
     public String getProbeHealthResponse(){
+        WebTarget webTarget = getTargetFor(PROBE);
+        boolean isAvailable = checkUriAvailability(webTarget.getUri());
+        if (isAvailable) {
```
fooker, line 91: I do not understand why this is needed. There are checks to retry on exception in all callers of this method.
If there is an exception happening that is not covered/ignored in these handlers, we should examine why this happens and maybe add special handling there.

smoke-test/src/main/java/org/opennms/smoketest/utils/RestClient.java
```diff
@@ -121,14 +122,18 @@ private static URL toUrl(InetSocketAddress addr) {
...
         final WebTarget target = getTarget().path("info");
-        final String json = getBuilder(target).get(String.class);
-
-        final ObjectMapper mapper = new ObjectMapper();
-        try {
-            JsonNode actualObj = mapper.readTree(json);
-            return actualObj.get("displayVersion").asText();
-        } catch (IOException e) {
-            LOG.debug("Failed to get displayVersion from the info REST service. (OpenNMS is probably not up yet): {}", e.getMessage());
+        boolean isReachable = checkUriAvailability(target.getUri());
```
fooker, line 125: Same as for `RestClientUtils`.`

fooker (changes requested):

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
fooker: Then the "why" should be examined. Maybe it just needs some `await`ing to ensure the background process has installed the bundles.

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
FrancescoBuc86: yes you're right

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
FrancescoBuc86: ok

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
FrancescoBuc86: ok

smoke-test/src/main/java/org/opennms/smoketest/utils/RestHealthClient.java
FrancescoBuc86: This extra check will avoid a ConnectionReset exception to be thrown while initializing the docker containers: the webtarget.request tries to get a response from a server (docker container) not ready/fully created yet. Result is a lof of spam in the logs. Where are these checks done? Maybe we can add this exception there

smoke-test/src/main/java/org/opennms/smoketest/utils/RestClient.java
FrancescoBuc86: ok

smoke-test/src/main/java/org/opennms/smoketest/utils/RestHealthClient.java
fooker: `waitUntilReadyWrapped` in `OpenNMSContainer`, `MinionContainer` and `SentinelContainer`.

smoke-test/src/main/java/org/opennms/smoketest/utils/RestClient.java
FrancescoBuc86: removed

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
FrancescoBuc86: ok

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
```diff
@@ -0,0 +1,161 @@
...
+    private static final String CONTAINER_PATH = "/usr/share/opennms/deploy/";
+    private static final String CLOUD_MOCK_JAR_NAME = "org.opennms.plugins.cloud-mock-with-dependencies.jar";
+    private static final String CLOUD_MOCK_MAIN = "org.opennms.plugins.cloud.ittest.MockCloudMain";
+    private static final String KAR_FILE_NAME = "org.opennms.plugins.cloud-assembly.kar";
+
+    @ClassRule
+    public static OpenNMSStack stack = OpenNMSStack
+            .withModel(StackModel.newBuilder()
+                    .withSentinel()
+                    .build());
```
fooker, line 67: There is `OpenNMSStack.SENTINEL` for this.

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
```diff
@@ -0,0 +1,161 @@
...
+
+    @Before
+    public void setUp() {
+        mockCloudContainer.start();
+    }
+
+    @Test
+    public void installCloudPlugin() throws Exception {
+        installCloudPluginInOpennmsMustBeSuccessful();
+        installCloudPluginInSentinelMustBeSuccessful();
```
fooker, line 88: Can we make this into two distinct tests by moving the `@Test` to the methods below?

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    public void installCloudPluginInOpennmsMustBeSuccessful() throws Exception {
+        // Given
+        stack.opennms().copyFileToContainer(MountableFile.forHostPath(PATH_TO_FILE), format("%s%s", CONTAINER_PATH, KAR_FILE_NAME));
+        String config = createConfig();
+        await().atMost(20, SECONDS);
```
fooker, line 95: What are we waiting for? Try to avoid unconditional awaits / sleeps by testing for some kind of condition which signals readiness.

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
```diff
@@ -0,0 +1,161 @@
...
+                return true;
+        });
+    }
+
+
+    public void installCloudPluginInSentinelMustBeSuccessful() throws Exception {
+        // Given
+        stack.sentinel().copyFileToContainer(MountableFile.forHostPath(PATH_TO_FILE), format("%s%s", CONTAINER_PATH, KAR_FILE_NAME));
+        String config = createConfig();
+        await().atMost(20, SECONDS);
```
fooker, line 115: Same here

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
```diff
@@ -0,0 +1,161 @@
...
+
+        // When
+        KarafShellUtils.withKarafShell(stack.sentinel().getSshAddress(), Duration.ofMinutes(3), streams -> {
+            streams.stdin.printf("kar:install file:%s%s%n", CONTAINER_PATH, KAR_FILE_NAME);
+            streams.stdin.println("feature:install opennms-plugin-cloud-sentinel");
+            streams.stdin.println("feature:list | grep opennms-plugin-cloud-sentinel");
+            await().atMost(com.jayway.awaitility.Duration.ONE_MINUTE).until(() -> streams.stdout.getLines().stream().anyMatch(line -> line.contains("Started")));
+            streams.stdin.println(config);
+            try {
+                Thread.sleep(60000);
```
fooker, line 125: Same here.

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
fooker: Unconditional waiting has two drawbacks here:
- CI pipeline is billed by time
- The will flap if CI is having a slow day

fooker (changes requested):

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
FrancescoBuc86: Actually not, because the sentinel installation will not work without having the opennms one done, that's why we put the two methods in one single test

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
FrancescoBuc86: this can be probably removed

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
FrancescoBuc86: probably this too

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
FrancescoBuc86: This one not, otherwise I'm gonna get an exception ("command not found") by trying to init the opennms-cloud; I can of course try to reduce the sleep time

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
patrick-schweizer: That is not possible. The sentinel plugin requires a successful install of the core plugin first.
Reason: The core plugin does the configuration and puts the certificates into the database via KV Store.
The Sentinel Plugin needs those certificates. The reason it was implemented that way:
The user needs to configure only one plugin.

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
FrancescoBuc86: ok

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
```diff
@@ -0,0 +1,161 @@
...
+                createCmd.withEntrypoint("/bin/bash");
+                createCmd.withName("cloudMock");
+            }).withCopyFileToContainer(MountableFile.forHostPath(PATH_TO_CLOUD_MOCK), format("%s%s", CONTAINER_PATH, CLOUD_MOCK_JAR_NAME))
+            .withCommand("-c", format("java -cp %s%s %s", CONTAINER_PATH, CLOUD_MOCK_JAR_NAME, CLOUD_MOCK_MAIN)).withExposedPorts(9003)
+            .withNetwork(stack.opennms().getNetwork())
+            .withNetworkAliases("cloudMock");
+
+    @Before
+    public void setUp() {
+        mockCloudContainer.start();
```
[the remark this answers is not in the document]
FrancescoBuc86: ok

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
fooker: You should `await` the command either by listing the commands (AFAIR `help`  should do this) or by just trying the command until it succeeds.

smoke-test/src/test/java/org/opennms/smoketest/OpenNMSCloudPluginIT.java
fooker: Ok, then this is fine. But it deserves a comment describing this dependency.

FrancescoBuc86: ok

fooker (approved):

PR #5563, opened 2022-12-02
Fix logs

amay94:
Adapter name and node ID were mixed up


RangerRick (approved):
hah; I wonder if this applies back into older releases?

amay94:
Ported back to foundation-2020 as I think (?) that's the oldest version we still support.

RangerRick (approved):

PR #5465, opened 2022-11-03
Restructuring the documentation: cleanup in operational

mfuhrmann:
@Bonrob2 @emarsh1995 @mmahacek @indigo423 

With these changes, I would like to clean up the operational section first, which should be a "manual" in the end. This will improve the readability of the docs, since we have a lot good documents but the structuring is not optimal.
Configuration examples, setups of features, or third-party tools like elastic can't stay in the operational section. So, for now, I dissolved the admin section in operational and moved everything out from the operational section that in my opinion has no reason to live there. Most of these parts went into reference and a lot into reference/configuration (for now).
Once it is cleaned up (after the reviews, creating an issue and some xref fixes that are still missing), we will have a more convenient "manual". Then we can take care of the reference section. Maybe my changes here don't make 100% sense now. So a short cleanup there would be helpful. And then I could imagine starting with other sections. Maybe tutorials, setups, troubleshooting etc and move other pages into it.

I bet there will also be a lot of pages that contain text that is good for a manual and configurations/setup instructions that would be correct to have in other sections. We will have to split them at some point I guess. We will see...

But for now, it is just moving pages. This is a relatively cheap task for us.

I would like to get your feedback on this.

PR #5564, opened 2022-12-02
Moving out how-tos from Deep-dive

mfuhrmann:
@Bonrob2 As discussed the first restructure PR. 
I've orientated on @indigo423's matrix here https://github.com/OpenNMS/opennms/pull/5465#issuecomment-1302726756 and went with `how-tos`.
So I've chosen just a few pages that are more howtos than explanations.

There are for sure a lot more. Just keeping the PR small this time.
Also, there are docs like https://github.com/OpenNMS/opennms/tree/release-31.x/docs/modules/operation/pages/deep-dive/device-config-backup where the introduction section is more `explanation` but the rest feels more as `how-to`. Could make sense to split articles like this step by step.

Does that make sense?


Bonrob2:
@mfuhrmann , I think this is a good start. Titles in the How To section really should indicate specific scenarios. So, for example, instead of "Alarm Correlation", it should be "Real-time Alarm Correlation Feedback" or something more precise. 
I think the how to's need to be organized around themes. We should do some work on planning those. Perhaps we should rename the "deep-dive" section as "How To", and then organize around themes like Alarm Correlation, Elastic Search, etc. Remove the very long explanations into a "reference" or "concepts" section, as a way to streamline things. Let's meet to discuss.

PR #5573, opened 2022-12-06
NMS-14937: Set SameSite to Strict for JSESSIONID

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-14937

### DIFF
--- a/opennms-webapp/src/main/webapp/WEB-INF/web.xml
+++ b/opennms-webapp/src/main/webapp/WEB-INF/web.xml
@@ -20,6 +20,7 @@
   <session-config>
     <cookie-config>
       <http-only>true</http-only>
+      <comment>__SAME_SITE_STRICT__</comment>
     </cookie-config>
     <session-timeout>-1</session-timeout>
   </session-config>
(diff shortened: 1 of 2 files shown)


opennms-webapp/src/main/webapp/WEB-INF/web.xml
```diff
@@ -20,6 +20,7 @@
   <session-config>
     <cookie-config>
       <http-only>true</http-only>
+      <comment>__SAME_SITE_STRICT__</comment>
```
RangerRick, line 23: lol, magic strings to enable this, huh? (sigh)
(that sigh was for your need to do this, not directed at you)

RangerRick (dismissed):

RangerRick:
I guess ideally we would have a smoke test that checks this, not sure how much trouble it would be to add it to one of the simple opennms-only tests?

amay94 (approved):

christianpape:
> I guess ideally we would have a smoke test that checks this, not sure how much trouble it would be to add it to one of the simple opennms-only tests?

Added test coverage to an existing smoke-test method.

RangerRick (approved):

PR #5579, opened 2022-12-06
NMS-15008: Replace community string in poller requests

amay94:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-15008

### DIFF
--- a/features/poller/monitors/core/src/main/java/org/opennms/netmgt/poller/monitors/SnmpMonitor.java
+++ b/features/poller/monitors/core/src/main/java/org/opennms/netmgt/poller/monitors/SnmpMonitor.java
@@ -1,8 +1,8 @@
 /*******************************************************************************
  * This file is part of OpenNMS(R).
  *
- * Copyright (C) 2002-2014 The OpenNMS Group, Inc.
- * OpenNMS(R) is Copyright (C) 1999-2014 The OpenNMS Group, Inc.
+ * Copyright (C) 2002-2022 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2022 The OpenNMS Group, Inc.
  *
  * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
  *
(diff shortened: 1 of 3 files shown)


fooker (commented):
Why do this in the RPC module? Could you just override/add an attribute using the runtime-attributes mechanism?

RangerRick (dismissed):
I'm not going to pretend to understand the intricacies of SCV, but this seems pretty straightforward, and seems pretty well-bounded to not accidentally blow up.

features/poller/monitors/core/src/main/java/org/opennms/netmgt/poller/monitors/SnmpMonitor.java
```diff
@@ -107,6 +111,30 @@ public PollStatus poll(MonitoredService svc, Map<String, Object> parameters) {
         String reasonTemplate = ParameterMap.getKeyedString(parameters, "reason-template", DEFAULT_REASON_TEMPLATE);
         String hexstr = ParameterMap.getKeyedString(parameters, "hex", "false");
 
+        String location = ParameterMap.getKeyedString(parameters, "location", LocationUtils.DEFAULT_LOCATION_NAME);
```
cgorantla, line 114: I think you can get location from `scv.getNodeLocation()`

aramosvizcarra (dismissed):
They seem to do the job and the code looks good from my end

features/poller/monitors/core/src/main/java/org/opennms/netmgt/poller/monitors/SnmpMonitor.java
```diff
@@ -107,6 +111,30 @@ public PollStatus poll(MonitoredService svc, Map<String, Object> parameters) {
...
+
+        if (!LocationUtils.DEFAULT_LOCATION_NAME.equals(location) && agentConfig.getAddress().isLoopbackAddress()) {
```
cgorantla, line 116: Another way to access scv by overriding  `@Override
    public Map<String, Object> getRuntimeAttributes(MonitoredService svc, Map<String, Object> parameters)`. and get read community there and send it as a parameter.

cgorantla (commented):
LGTM, few comments.

features/poller/monitors/core/src/main/java/org/opennms/netmgt/poller/monitors/SnmpMonitor.java
cgorantla: As @amay94 pointed, we can't use `getRuntimeAttributes` for this as the scv needs to be retrieved only on Minion and `getRuntimeAttributes` only runs on OpenNMS

cgorantla (approved):
LGTM!

PR #5591, opened 2022-12-08
NMS-14988: Set additional entity expansion options

amay94:
See the [OWASP cheat sheet](https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html#spring-framework-mvcoxm-xxe-vulnerabilities) on XEE prevention.

### External References

### DIFF
--- a/core/test-api/xml/src/main/java/org/opennms/core/test/xml/XmlTest.java
+++ b/core/test-api/xml/src/main/java/org/opennms/core/test/xml/XmlTest.java
@@ -243,7 +243,7 @@ abstract public class XmlTest<T> {
         });
         try {
             final InputSource inputSource = new InputSource(getSampleXmlInputStream());
-            final XMLFilter filter = JaxbUtils.getXMLFilterForClass(getSampleClass());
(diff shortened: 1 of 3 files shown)


aramosvizcarra (commented):
LGTM, not sure if the test failing has anything todo with the rrd change.

amay94:
> LGTM, not sure if the test failing has anything todo with the rrd change.

The test was failing because we have tests that use config files with DOCTYPE declarations. It looks like RRD dumps contain those by default, so I updated the fix to only apply to XML received from rest endpoints

RangerRick:
Pretty sure the failures are real since RRD XML dumps have `DOCTYPE` set (created by `rrdtool dump`). Jasper Reports also have `DOCTYPE` defined, although Jasper _probably_ already does its own parsing.

`git grep DOCTYPE | grep src/test` should show you where potential pitfalls are.

RangerRick (approved):

christianpape (approved):
LGTM.

PR #5594, opened 2022-12-11
NMS-15134: Default to direction 'unknown' when the direction field is not present

j-white:
JIRA: https://opennms.atlassian.net/browse/NMS-15134

This allows for bidirectional flows to work properly when the exporter does not include the field.

Tested with a RB5009 on MikroTik RouterOS 7.6.

### DIFF
--- a/features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/transport/Netflow9MessageBuilder.java
+++ b/features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/transport/Netflow9MessageBuilder.java
@@ -61,6 +61,7 @@ public class Netflow9MessageBuilder implements MessageBuilder {
     @Override
     public FlowMessage.Builder buildMessage(final Iterable<Value<?>> values, final RecordEnrichment enrichment) {
         final FlowMessage.Builder builder = FlowMessage.newBuilder();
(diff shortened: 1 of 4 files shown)


features/telemetry/protocols/netflow/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/netflow/parser/transport/Netflow9MessageBuilder.java
```diff
@@ -78,7 +79,7 @@ public FlowMessage.Builder buildMessage(final Iterable<Value<?>> values, final R
         Long dstVlan = null;
         Long flowActiveTimeout = this.flowActiveTimeoutFallback;
         Long flowInActiveTimeout = this.flowInactiveTimeoutFallback;
-        Long sysUpTime = null;
+        Long sysUpTime = 0L;
```
Naicisum, line 82: My nit, why not keep this as `null` and then change `line 260` below to read as:

`        long bootTime = sysUpTime != null ? timeStampInMsecs - sysUpTime : timeStampInMsecs;`

j-white: that works too, thanks

j-white:
There are a lot of test failures here now. Will need some more work.

j-white:
Closing PR for now. May take a different approach here.

PR #5596, opened 2022-12-13
NMS-15011: Fixed scvcli commands

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-15011

### DIFF
--- a/docs/modules/deployment/pages/core/getting-started.adoc
+++ b/docs/modules/deployment/pages/core/getting-started.adoc
@@ -149,9 +149,10 @@ To achieve this, use the `scvcli` command line utility to add the credentials, a
 .Create encrypted credentials for the PostgreSQL database connections:
 [source, console]
 ----
-sudo -u opennms ${OPENNMS_HOME}/bin/scvcli set postgres opennms opennms
-sudo -u opennms ${OPENNMS_HOME}/bin/scvcli set postgres-admin postgres opennms
+sudo -u opennms ${OPENNMS_HOME}/bin/scvcli set postgres opennms password<1>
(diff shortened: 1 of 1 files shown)


emarsh1995 (commented):
Since this reflects what was suggested in the ticket, I'm assuming the commands are accurate, but I'll let @indigo423 or @mmahacek review and approve to be sure. Thank you for the update!

mmahacek (commented):
Should there be a line that suggests replacing the last "opennms" on each line with a secure password of their choosing? Otherwise users might think this is what it has to be, when in reality it shouldn't.

docs/modules/deployment/pages/core/getting-started.adoc
```diff
@@ -149,8 +149,8 @@ To achieve this, use the `scvcli` command line utility to add the credentials, a
 .Create encrypted credentials for the PostgreSQL database connections:
 [source, console]
 ----
-./bin/scvcli set postgres opennms opennms-password
-./bin/scvcli set postgres-admin postgres postgres-password
+sudo -u opennms ${OPENNMS_HOME}/bin/scvcli set postgres opennms opennms
+sudo -u opennms ${OPENNMS_HOME}/bin/scvcli set postgres-admin postgres opennms
 ----
```
mmahacek, line 154:
```suggestion
sudo -u opennms ${OPENNMS_HOME}/bin/scvcli set postgres opennms password<1>
sudo -u opennms ${OPENNMS_HOME}/bin/scvcli set postgres-admin postgres password
----
<1> Replace both `password` instances with the actual passwords for your `opennms` and `postgres` database accounts.
```

PR #5601, opened 2022-12-13
NMS-15119: Logout issue with Vue Menubar

synqotik:
There was an issue with the new Vue menubar when clicking 'Log Out', resulting in an ugly 404 page and not actually logging out. Legacy menubar actually does a form POST to `j_spring_security_logout`, Vue Menubar was doing a GET. Updated Vue Menubar to have POST behavior.

Also fixed a small null reference issue in `login.jsp`.

### External References

* JIRA (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15119

### DIFF
--- a/ui/src/components/Layout/Menubar.vue
+++ b/ui/src/components/Layout/Menubar.vue
@@ -93,6 +93,8 @@
           </FeatherDropdownItem>
         </FeatherDropdown>
 
+        <form ref="logoutForm" name="vueLogoutForm" :action="computeLogoutFormLink()" method="post"></form>
+
         <!-- Self-service menu -->
         <FeatherDropdown v-if="mainMenu.selfServiceMenu" class="menubar-dropdown"
           @mouseenter="hoverItem(SelfServiceIndex)" :modelValue="hoveredItems[SelfServiceIndex]">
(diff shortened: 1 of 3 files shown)


mikewrosey (approved):
lgtm

ui/src/components/Layout/Menubar.vue
```diff
@@ -420,6 +439,15 @@ const onMenuItemClick = (url: string, isVueLink?: boolean | null) => {
   window.location.assign(link)
 }
 
+const onSelfServiceMenuItemClick = (item: MenuItem) => {
+  if (item.id === 'logout') {
```
scthompsonopen, line 443: Are we embedding the form, generating the URL on render, and then roundabout submitting the blank form just to get the POST + redirect? I wonder if we could get the same functionality with a POST `fetch` call + a redirect to the login page.

I'll approve this regardless of my question because I assume it's mimicking how the old logout button worked to reduce integration friction and speed up time to acceptance. 

But maybe we can create a new task as tech debt for the future to create an integrated and shareable logout method that both UIs can use. It might be a good idea to try and consolidate shared functionality over time if we're going to carry both UIs into the future. (So that a future update to the Navbar logout functionality doesn't result in twice as much work to update both implementations, which would be annoying from a dev perspective I think).

scthompsonopen (approved):

ui/src/components/Layout/Menubar.vue
synqotik: Yes that's what we're doing. I agree, should be tech debt and create a better approach.

PR #5623, opened 2022-12-22
Container overlay verbose copy

deejgregor:
### All Contributors

### Contribution Checklist

  4. once you've created this PR, please link to it in a comment in the JIRA issue

### What's Next?

### External References


deejgregor:
I replaced every instance of 'cp' in the entrypoint.sh scripts with 'rsync --out-format="%n %C"` (translating other arguments appropriately).

This now returns nothing. :)
```
git grep -l -E '\scp\s' opennms-container/*/container-fs
```

mershad-manesh (dismissed):
Overall the changes looks good to me.

Before delivering, we should ask @RangerRick to take a look at the changes.

deejgregor:
> Before delivering, we should ask @RangerRick to take a look at the changes.

Thanks for taking a look, and I've requested a review from Ben.

@RangerRick If you want to save yourself some time, focus on the first commit that uses `rsync` in place of `cp` in the `entrypoint.sh` scripts: https://github.com/OpenNMS/opennms/pull/5623/commits/1c7fb278e19b3d259b4f3913d62ca46c90681cdf

The other three commits are just smoke test changes.

opennms-container/minion/container-fs/entrypoint.sh
```diff
@@ -152,7 +152,7 @@ applyOverlayConfig() {
   # Overlay etc specific config
   if [ -d "${MINION_OVERLAY_ETC}" ] && [ -n "$(ls -A ${MINION_OVERLAY_ETC})" ]; then
     echo "Apply custom etc configuration from ${MINION_OVERLAY_ETC}."
-    cp -Lr ${MINION_OVERLAY_ETC}/* ${MINION_HOME}/etc || exit ${E_INIT_CONFIG}
+    rsync -r --out-format="%n %C" ${MINION_OVERLAY_ETC}/* ${MINION_HOME}/etc || exit ${E_INIT_CONFIG}
```
RangerRick, line 155: Does this not technically change the behavior? Do you need to add `-L` to the `rsync` as well?

smoke-test/src/test/java/org/opennms/smoketest/minion/MinionOverlayIT.java
```diff
@@ -0,0 +1,57 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2016-2021 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2021 The OpenNMS Group, Inc.
```
RangerRick, line 5: whoa, you've been working on this branch a long time! <grin>

opennms-container/minion/container-fs/entrypoint.sh
RangerRick: Also anything that syncs into a directory should have `/` at the end since `cp` has somewhat different semantics. So it should be:

```bash
rsync -r -L --out-format="%n %C" ${MINION_OVERLAY_ETC}/* ${MINION_HOME}/etc/ || exit ${E_INIT_CONFIG}
```

...if we're trying to not change behavior.

(I didn't go look at the other rsync's but I bet that's the case for a lot of these.

It might just happen to work anyway this time, but I strongly recommend making sure your slashes are appropriate because it's easy for it to go wrong with rsync.

deejgregor: > Does this not technically change the behavior? Do you need to add `-L` to the `rsync` as well?

I thought `-L` in rsync did something different from `-L` in cp, but was a misunderstanding on my end and reading the **wrong** man page for cp (I was looking at macOS, which requires that `-R` be passed for `-L` to do anything, but the one that runs in the container is GNU coreutils, which doesn't seem to have such a requirement).

Reading the *detailed* docs for both, it sounds like they are equivalent, so the `-L` should be there, so I'll add that back.

GNU coreutils:
> `-L`, `--dereference`
> Follow symbolic links when copying from them. With this option, cp cannot create a symbolic link. For example, a symlink (to regular file) in the source tree will be copied to a regular file in the destination tree.

-- https://www.gnu.org/software/coreutils/manual/html_node/cp-invocation.html

rsync:
> `-L`, `--copy-links`
> When symlinks are encountered, the item that they point to (the referent) is copied, rather than the symlink. In older versions of rsync, this option also had the side-effect of telling the receiving side to follow symlinks, such as symlinks to directories. In a modern rsync such as this one, you'll need to specify --keep-dirlinks (-K) to get this extra behavior. The only exception is when sending files to an rsync that is too old to understand -K -- in that case, the -L option will still have the side-effect of -K on that older receiving rsync.

-- https://linux.die.net/man/1/rsync

deejgregor:
@RangerRick ready for re-review:

* [Ooops, -L should be here](https://github.com/OpenNMS/opennms/pull/5623/commits/86997ceec33cb9d14f4efc9c03638183e4d0f945) 
* ["/." all the things! (well, copy target directories, at least)](https://github.com/OpenNMS/opennms/pull/5623/commits/d420fa0da14632801ec194896419866b232b91d1) 
* [Fix copyright date <grin>](https://github.com/OpenNMS/opennms/pull/5623/commits/08a6b843af42be0394e6809f31deb708cd1e1d0e)

Note the middle one. It's not technically needed, but it's a nit I tend to follow when I do things to be *super* explicit. It will require that the target directory already exists--which should be the case in all of these situations. I think it's a good thing for code like this, so you don't get directories created when you thought they already existed, maybe with unexpected permissions (shouldn't be that much of an issue in containers, but you never know...).

RangerRick:
> Note the middle one. It's not technically needed, but it's a nit I tend to follow when I do things to be _super_ explicit. It will require that the target directory already exists--which should be the case in all of these situations. I think it's a good thing for code like this, so you don't get directories created when you thought they already existed, maybe with unexpected permissions (shouldn't be that much of an issue in containers, but you never know...).

This is a form I've never seen anyone do with rsync and I admit makes my brain go "wait, what?" but I can see the utility.

RangerRick (approved):

RangerRick:
Also I would say next time, do the test refactoring in a different place from just cleaning up verbose copies. :D

PR #5665, opened 2023-01-09
NMS-15124: Fix possible mismatch between business service and alarm state

amay94:
when restarting or reloading bsmd.

Defer and reduce update events until the graph is completely processed.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-15124


mmahacek:
I noticed the Jira indicading fix targeted to Meridian 2020, but this branch targets 2023.  Is there another  branch that will target the other version, or is the fix version in Jira incorrect?  Customer is on Horizon 31, and I want to at least verify if a fix will be heading to `release-31.x`.

amay94:
Good catch @mmahacek, PR for `foundation-2020` can be found here: https://github.com/OpenNMS/opennms/pull/5667

PR #5695, opened 2023-01-18
NMS-14578: Show resolving UEIs in alarm/detail.jsp

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-14578

### DIFF
--- a/opennms-config-model/src/main/java/org/opennms/netmgt/xml/eventconf/Events.java
+++ b/opennms-config-model/src/main/java/org/opennms/netmgt/xml/eventconf/Events.java
@@ -28,21 +28,6 @@
 
 package org.opennms.netmgt.xml.eventconf;
 
-import org.opennms.core.xml.JaxbUtils;
-import org.opennms.core.xml.ValidateUsing;
-import org.opennms.netmgt.config.utils.ConfigUtils;
-import org.springframework.core.io.DefaultResourceLoader;
-import org.springframework.core.io.Resource;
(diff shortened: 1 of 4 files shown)


RangerRick:
Should this really only be against 31.x? Is there any reason for doing 31.x-only commits that don't go into `foundation-2023`?

christianpape:
> Should this really only be against 31.x? Is there any reason for doing 31.x-only commits that don't go into `foundation-2023`?

right, rebased to foundation-2023

opennms-config-model/src/main/java/org/opennms/netmgt/xml/eventconf/Events.java
```diff
@@ -69,6 +54,22 @@
...
+import javax.xml.bind.annotation.XmlType;
+
+import org.opennms.core.xml.JaxbUtils;
+import org.opennms.core.xml.ValidateUsing;
+import org.opennms.netmgt.config.utils.ConfigUtils;
+import org.springframework.core.io.DefaultResourceLoader;
+import org.springframework.core.io.Resource;
+import org.springframework.dao.DataAccessResourceFailureException;
+import org.springframework.orm.ObjectRetrievalFailureException;
+import org.springframework.util.StringUtils;
```
RangerRick, line 71: the editor wars are at it again!

RangerRick (approved):
Looks OK to me, although that test failure is odd, doesn't seem like it would be related. I went ahead and reran it, we shouldn't merge until/unless it goes green.

PR #5714, opened 2023-01-19
NMS-15201: bump protobuf to latest 3.16.x

RangerRick:
Not knowing how it would affect things, I only bumped to the latest 3.16.x, rather than the latest 3.x, since 3.16.3 was sufficient to avoid known CVEs.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-15201


christianpape (approved):
LGTM.

PR #5715, opened 2023-01-19
NMS-15193: remove RANCID integration

RangerRick:
This PR removes all RANCID-related code including RANCID integration points, the RWS (RANCID Web Service), and "storage"/"inventory" code relating to RANCID.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-15193

PR #5716, opened 2023-01-19
NMS-15193: remove RANCID default configuration

RangerRick:
This is a less-invasive version of the work done for NMS-15193 in develop for #5715.
It mostly removes the default configuration files and references to them in the config tester.
I also had to make a few small changes to hide "default" versions of the rancid and rws configs in the jars because removing them completely would be too much.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-15193


integration-tests/config/src/test/java/org/opennms/netmgt/config/WillItUnmarshalIT.java
```diff
@@ -245,11 +243,11 @@ private static void addFile(final Source source, final String file, final Class<
         addFile(Source.CONFIG, "poller-configuration.xml", PollerConfiguration.class, true, null);
         // moved to CmWillItUnmarshalIT
         //addFile(Source.CLASSPATH, "/defaults/provisiond-configuration.xml", ProvisiondConfiguration.class, false, null);
-        addFile(Source.CONFIG, "rancid-configuration.xml", RancidConfiguration.class, true, null);
+        //addFile(Source.CONFIG, "rancid-configuration.xml", RancidConfiguration.class, true, null);
```
fooker, line 246: As the config file is remove, the entries here should be removed, too.

fooker (changes requested):

aramosvizcarra (dismissed):
Looks good to me

PR #5738, opened 2023-01-25
NMS-15324: Destroy ring buffer faster and more transparently

amay94:
### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-15324

### DIFF
--- a/features/newts/src/main/java/org/opennms/netmgt/newts/NewtsWriter.java
+++ b/features/newts/src/main/java/org/opennms/netmgt/newts/NewtsWriter.java
@@ -206,8 +206,6 @@ public class NewtsWriter implements WorkHandler<SampleBatchEvent>, DisposableBea
         Logging.putPrefix("collectd");
 
         List<Sample> samples = event.getSamples();
-        // Decrement our entry counter
-        m_numEntriesOnRingBuffer.decrementAndGet();
 
         // Partition the samples into collections smaller then max_batch_size
         for (List<Sample> batch : Lists.partition(samples, m_maxBatchSize)) {
@@ -231,6 +229,9 @@ public class NewtsWriter implements WorkHandler<SampleBatchEvent>, DisposableBea
                 RATE_LIMITED_LOGGER.error("An error occurred while inserting samples. Some sample may be lost.", t);
             }
         }
+
+        // Decrement our entry counter
+        m_numEntriesOnRingBuffer.decrementAndGet();
     }
 
     private static final EventTranslatorOneArg<SampleBatchEvent, List<Sample>> TRANSLATOR =
--- a/features/timeseries/src/main/java/org/opennms/netmgt/timeseries/TimeseriesStorageManagerImpl.java
+++ b/features/timeseries/src/main/java/org/opennms/netmgt/timeseries/TimeseriesStorageManagerImpl.java
@@ -53,7 +53,7 @@ public class TimeseriesStorageManagerImpl implements TimeseriesStorageManager {
 
     public TimeseriesStorageManagerImpl() {
         this(new ServiceLookupBuilder(new ServiceRegistryLookup(DefaultServiceRegistry.INSTANCE))
-                .blocking()
+                .blocking(ServiceLookupBuilder.GRACE_PERIOD_MS, ServiceLookupBuilder.LOOKUP_DELAY_MS, 0)
                 .build());
     }
 
--- a/features/timeseries/src/main/java/org/opennms/netmgt/timeseries/samplewrite/RingBufferTimeseriesWriter.java
+++ b/features/timeseries/src/main/java/org/opennms/netmgt/timeseries/samplewrite/RingBufferTimeseriesWriter.java
@@ -33,6 +33,7 @@ import java.time.Instant;
 import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.Executor;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -82,7 +83,11 @@ public class RingBufferTimeseriesWriter implements TimeseriesWriter, WorkHandler
             .build();
     private static final Duration STORAGE_GET_WARNING_DURATION = Duration.ofSeconds(5);
 
+    private static final Duration DESTROY_GRACE_PERIOD = Duration.ofSeconds(30);
+
     private WorkerPool<SampleBatchEvent> workerPool;
+
+    private ExecutorService executor;
 
     private RingBuffer<SampleBatchEvent> ringBuffer;
 
@@ -139,7 +144,7 @@ public class RingBufferTimeseriesWriter implements TimeseriesWriter, WorkHandler
         // Executor that will be used to construct new threads for consumers
         final ThreadFactory namedThreadFactory = new ThreadFactoryBuilder()
                 .setNameFormat("TimeseriesWriter-Consumer-%d").build();
-        final Executor executor = Executors.newCachedThreadPool(namedThreadFactory);
+        executor = Executors.newCachedThreadPool(namedThreadFactory);
 
         @SuppressWarnings("unchecked")
         final WorkHandler<SampleBatchEvent>[] handlers = new WorkHandler[numWriterThreads];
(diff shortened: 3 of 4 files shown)


features/timeseries/src/main/java/org/opennms/netmgt/timeseries/samplewrite/RingBufferTimeseriesWriter.java
```diff
@@ -179,6 +185,10 @@ public void destroy() {
                         break;
                     }
                 }
+                if (numEntriesOnRingBuffer.get() != 0) {
+                    LOG.warn("destroy(): WorkerPool does not want to cooperate, forcing cooperation");
```
deejgregor, line 189: <3. That is all. :)

features/timeseries/src/main/java/org/opennms/netmgt/timeseries/samplewrite/RingBufferTimeseriesWriter.java
```diff
@@ -187,28 +191,37 @@ public void destroy() {
...
-            droppedSamples.mark(samples.size());
+            insertDrop(samples, "The ring buffer is full");
             return;
         }
         // Increase our entry counter
         numEntriesOnRingBuffer.incrementAndGet();
     }
 
+    private void insertDrop(List<Sample> samples, String message) {
+        RATE_LIMITED_LOGGER.error("{}. {} samples associated with resource ids {} will be dropped.",
```
deejgregor, line 219: I would suggest changing this to include the `message` in the format string instead of pulling it in with an `{}` anchor. So, change this:
```
    RATE_LIMITED_LOGGER.error("{}. {} samples associated with resource ids {} will be dropped.",
Write
                message,
                samples.size(), new Object() {
```
to this:
```
    RATE_LIMITED_LOGGER.error(message + ". {} samples associated with resource ids {} will be dropped.",
Write
                samples.size(), new Object() {
```

Otherwise, the rate limited log messages aren't great because the message just shows up as `null`, like this:
```
2023-01-23T11:56:39,572 | ERROR | RateLimitedLogRegistry-0 | RingBufferTimeseriesWriter       | 478 - org.opennms.features.org.opennms.features.timeseries - 32.0.0.SNAPSHOT | (suppressed 2 logs similar to 'null. null samples associated with resource ids {} will be dropped.' in PT29.927S)
```

deejgregor (dismissed):
I made one suggestion, but I'm good as-is. If you push a change for the suggestion, let me know and I can quickly re-approve once I see it.

features/newts/src/main/java/org/opennms/netmgt/newts/NewtsWriter.java
```diff
@@ -202,12 +201,7 @@ public String toString() {
 
     @Override
     public void onEvent(SampleBatchEvent event) throws Exception {
-        // We'd expect the logs from this thread to be in collectd.log
-        Logging.putPrefix("collectd");
```
cgorantla, line 206: why are we removing this Logging ?

features/timeseries/src/main/java/org/opennms/netmgt/timeseries/TimeseriesStorageManagerImpl.java
```diff
@@ -53,7 +53,7 @@ public class TimeseriesStorageManagerImpl implements TimeseriesStorageManager {
 
     public TimeseriesStorageManagerImpl() {
         this(new ServiceLookupBuilder(new ServiceRegistryLookup(DefaultServiceRegistry.INSTANCE))
-                .blocking()
+                .blocking(ServiceLookupBuilder.GRACE_PERIOD_MS, ServiceLookupBuilder.LOOKUP_DELAY_MS, 0)
```
cgorantla, line 56: If this is equivalent to 0 ms wait,  simply remove blocking

features/timeseries/src/main/java/org/opennms/netmgt/timeseries/samplewrite/RingBufferTimeseriesWriter.java
```diff
@@ -154,55 +161,102 @@ private void setUpWorkerPool() {
         ringBuffer.addGatingSequences(workerPool.getWorkerSequences());
 
         workerPool.start(executor);
+
+        readyToRockAndRoll.set(true);
     }
 
     @Override
     public void destroy() {
+        readyToRockAndRoll.set(false);
```
cgorantla, line 170: Do we need to reset this back to true ?

features/newts/src/main/java/org/opennms/netmgt/newts/NewtsWriter.java
deejgregor: This isn't removing logging, it's just not messing with the prefix and leaving it to whatever it otherwise would be. When it sets the logging prefix to `collectd`, a few bad things were happening:
1. The component here isn't `collectd`, it's the newts writer (or the ring buffer which also does the same thing). The ring buffer is running inside of karaf, so its logs will go there, and I suspect the same for newts.
2. Although some of the data through here might come from `collectd`, not all of it does. Data will also commonly come from `pollerd`, as well as other sources that generate metrics.
3. Some of the logging on errors in the ring buffer/timeseriesstorage related to initialization was confusingly showing up in the collectd log, making problems harder to chase down (especially when I ran headfirst into https://opennms.atlassian.net/browse/NMS-15285 ). Eeek! These two problems made it take awhile before we noticed this: https://opennms.atlassian.net/browse/NMS-15282?focusedCommentId=69239

features/timeseries/src/main/java/org/opennms/netmgt/timeseries/TimeseriesStorageManagerImpl.java
deejgregor: There are *two* independent periods that are checked and `BlockingServiceLookup` won't give up until *both* expire. We are only changing the second one, wait period, to zero:
1. grace period -- the first argument to `ServiceLookupBuilder.blocking`, `gracePeriodInMs`.
2. wait period -- the third argument to `ServiceLookupBuilder.blocking`, `waitTimeMs`.

(the second argument to `ServiceLookupBuilder.blocking` is `sleepTimeInMs`, which is the amount of time to wait between checks to see if the service is available)

```
        // A service matching the filter is not currently available.
        // Wait until the system has finished starting up (uptime >= grace period)
        // while ensuring we've waited for at least WAIT_PERIOD_MS before aborting the search.
        final long waitUntil = System.currentTimeMillis() + this.waitTimeMs;
        while (uptimeSupplier.get() < this.gracePeriodInMs
                || System.currentTimeMillis() < waitUntil) {
            try {
                Thread.sleep(this.lookupDelayMs);
...
```
-- https://github.com/OpenNMS/opennms/blob/develop/core/soa/src/main/java/org/opennms/core/soa/lookup/BlockingServiceLookup.java#L65

features/timeseries/src/main/java/org/opennms/netmgt/timeseries/samplewrite/RingBufferTimeseriesWriter.java
deejgregor: It is only set to `true` in `setUpWorkerPool`, which is called from the constructor. Is it reasonable from a Kafka standpoint to require a new instance of the class to be constructed after `destroy` has been called?

features/timeseries/src/main/java/org/opennms/netmgt/timeseries/TimeseriesStorageManagerImpl.java
cgorantla: got it. Makes sense!

features/newts/src/main/java/org/opennms/netmgt/newts/NewtsWriter.java
cgorantla: Except for  response time (Pollerd), most of the persisting happens from Collectd. It would be good to separate logs from  karaf.log into their individual component specific logs.  

Also note that Newts may run outside Karaf when we use Newts alone . So not sure where they end up if we remove this prefix.

If we want to break this further, we can create another logging prefix for timeseries/newts but removing prefix may not be preferred solution ( especially for newts as customers may expect logs in collectd.log)

deejgregor: Thanks for the context. I removed this change from `NewtsWriter`.

I left the same change in `RingBufferTimeseriesWriter`, but it emits almost no logs in my experience except instantiation, destroy and when there are problems.

features/timeseries/src/main/java/org/opennms/netmgt/timeseries/samplewrite/RingBufferTimeseriesWriter.java
deejgregor: Change applied.

deejgregor (approved):
I'm good with the changes that Alex added. I'll still look for approval for my changes.

cgorantla (approved):

PR #5750, opened 2023-01-30
fixes and improvements for docker_gc

RangerRick:
This PR fixes the `docker_gc` automation to do a number of things:

1. write a `config.json` docker file to avoid having to deal with multiple levels of escaping around doing `expect` commands
2. add `minion`, `sentinel`, and `meridian-minion` to the list of images that can be cleaned up
3. add `jira` and `dependabot` to the list of branch types to clean up (also: increase strictness of _where_ these keywords get matched)


jb68 (approved):
LGTM

PR #5756, opened 2023-01-31
NMS-15343: remove JNLP launcher for remote poller

RangerRick:
The remote poller has been gone since `foundation-2022`. JNLP is unsupported by Oracle _and_ has essentially been untested in the OpenNMS codebase for years.

This PR leaves the CLI version of the remote poller intact, but removes JNLP/webstart and all of the infrastructure for jar signing that goes with it.

### External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/NMS-15343


christianpape (approved):
LGTM.

PR #5768, opened 2023-02-03
NMS-15385: Docker container build improvements

deejgregor:
- Add sanity check to make sure opennms.home was set correctly
- Delay copying from core-tarball to make rebuilds faster (Minion and Sentinel are already cool)
- Fix spelling of org.opennms.cicd.jobid LABEL

### External References

* JIRA (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15385


deejgregor:
I noticed this failure:
```
Sanity checking OPENNMS_HOME path in **/fix-permissions script...
-rw-r--r-- 0/0               91322502 2023-02-03 04:47 jetty-webapps/*******/source/*******-31.0.3-SNAPSHOT-source.tar.gz
found source tarball in ../../*******-full-assembly/target/*******-31.0.3-SNAPSHOT-core.tar.gz (see above)
make: *** [../common.mk:107: tarball-root/data/tmp/README] Error 1
```
-- source: https://app.circleci.com/pipelines/github/OpenNMS/opennms/27413/workflows/bd6c64f6-97b9-4181-a81e-8a2efea2d659/jobs/222251?invite=true#step-108-34

The failure is due to this check I added here: https://github.com/OpenNMS/opennms/pull/5768/files#diff-aa16ea5944b9e3379b39ad94cfca39919f3f10c3424a7a8e676e369d02cc0959R117

I checked, and it looks like we are including a ~90 MB source tarball in our Docker container images. @RangerRick Is this right? I see we include the source in a separate package for Debian and RPM. Do we want to include it in the Docker container?
```
$ docker run -it --rm --entrypoint=/bin/bash opennms/horizon:31.0.3 -c "ls -l jetty-webapps/opennms/source"
total 89080
-rw-r--r-- 1 opennms root 91213150 Jan 10 21:46 opennms-31.0.3-source.tar.gz
```

At the moment, I'm not sure why I added this check for the source tarball, though (and it isn't mentioned in the commit that added it). :-/ If it doesn't make sense, I can pull it.

deejgregor:
> At the moment, I'm not sure why I added this check for the source tarball, though (and it isn't mentioned in the commit that added it). :-/ If it doesn't make sense, I can pull it.

I went ahead and pulled it, since it didn't really fit in with the other check (and since tests failed, that change was clearly incomplete).

RangerRick (approved):
"opennme" = "open enemy?"

PR #5769, opened 2023-02-03
NMS-15386: When we fail to startup, exit with a non-zero exit code

deejgregor:
- Cleanups, take care of warnings
- Cleanup warnings
- When we fail to startup, exit with a non-zero exit code.

### External References

* JIRA (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15386

### DIFF
--- a/core/daemon/src/main/java/org/opennms/netmgt/vmmgr/Manager.java
+++ b/core/daemon/src/main/java/org/opennms/netmgt/vmmgr/Manager.java
@@ -54,28 +54,14 @@ import org.slf4j.LoggerFactory;
 
 /**
  * <p>
- * The Manager is reponsible for launching/starting all services in the VM
- * that it is started for. The Manager operates in two modes, normal and
- * server
+ * The Manager is responsible for managing running services in the VM
+ * that it is started for.
  * </p>
  * <p>
- * normal mode: In the normal mode, the Manager starts all services configured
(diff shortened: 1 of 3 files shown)


core/daemon/src/main/java/org/opennms/netmgt/vmmgr/Manager.java
```diff
@@ -54,61 +55,54 @@
...
     /**
      * <p>stop</p>
      */
     @Override
     public synchronized void stop() {
+        stop(0);
+    }
+
+    public synchronized void stop(int processExitCode) {
         if (stopInitiated.get()) {
```
RangerRick, line 101: I wonder if you should check the existing error code here, and set it if `processExitCode` is non-zero, but the current value is zero?

IE, something calls stop, but then something else calls it with an error code, I would think you'd want to propagate the error code.

RangerRick (approved):

core/daemon/src/main/java/org/opennms/netmgt/vmmgr/Manager.java
deejgregor: Good suggestion! And it made the change simpler, too. :) This is the updated commit: https://github.com/OpenNMS/opennms/pull/5769/commits/2dad56d323ecf16722cf83e4f65baa814b9783aa

RangerRick (approved):

PR #5770, opened 2023-02-03
Smoke test improvements and tweaks to help developers

deejgregor:
- Allow thread dumps on shutdown to be disabled.
- Add "restart-fast" option to opennms script.
- Disable code coverage collection by default in smoke tests
- Speedup thread dump collection and store in its own file

### External References

* JIRA (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15387

### DIFF
--- a/.circleci/scripts/smoke.sh
+++ b/.circleci/scripts/smoke.sh
@@ -75,4 +75,5 @@ fi
   -N \
   '-P!smoke.all' \
   "-Psmoke.$SUITE" \
+  -Pcoverage \
   install verify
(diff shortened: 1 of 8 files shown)


RangerRick (dismissed):
a note, right now we are not _using_ smoke coverage, I backed off that for now in the sonar bits because it wasn't working right, so you could probably turn it off for now and get a little bit more of a speed boost

deejgregor:
> a note, right now we are not _using_ smoke coverage, I backed off that for now in the sonar bits because it wasn't working right, so you could probably turn it off for now and get a little bit more of a speed boost

@RangerRick I just pushed https://github.com/OpenNMS/opennms/pull/5770/commits/68f51bcf3daf16af056802400382e261ef8c1334 to disable this for now. If you want to review that one commit, that would be great.

I think we can probably cancel the CI job, since I don't think it'll tell us anything new. ;-)

RangerRick (approved):

PR #5773, opened 2023-02-04
Move from com.jayway.awaitility (1.7.0) to org.awaitility (4.2.0)

deejgregor:
### Background

We use a mix of both old com.jayway.awaitility (1.7.0) and newer org.awaitility (4.2.0):
```
# old
REMMDGREGOR (⎈|minikube:default) opennms dgregor$ git grep awaitility | grep -i jayway| grep pom.xml | wc
      49      98    4101
# new
REMMDGREGOR (⎈|minikube:default) opennms dgregor$ git grep awaitility | grep -i org.awaitility| grep pom.xml | wc
       7      14     486
```

We’ve run into at least one case where code using the old com.jayway.awaitility (1.7.0) was copied to Horizon Stream and was then difficult to debug, where the newer version would have made the problem clear.

It’s a pretty straightforward upgrade and we only use it in test code.

### Changes

Most common changes:
1. In pom.xml files: `<groupId>com.jayway.awaitility</groupId>` -> `<groupId>org.awaitility</groupId>`
2. `import static com.jayway.awaitility.Awaitility.await;` -> `import static org.awaitility.Awaitility.await;` (and similar for other imports)

Less common changes:
1. Change `com.jayway.awaitility.Duration`s to `java.time.Duration` (slightly more than an import change, but still straightforward). -- https://github.com/OpenNMS/opennms/pull/5773/files#diff-83c224d60a19741a45c5f8ae1c3e77e8fe63943b05cf845c03c804053e0a00fbL111 (and a few more in this file)
2. `until` with a `Runnable` or lambda that doesn't return a `Boolean` -> `untilAsserted` -- https://github.com/OpenNMS/opennms/pull/5773/files#diff-83c224d60a19741a45c5f8ae1c3e77e8fe63943b05cf845c03c804053e0a00fbL111 (and a few more in this file), https://github.com/OpenNMS/opennms/pull/5773/files#diff-10602d515015513dda599ee64a46655b8d2b019142e27d29927682bae7c7794fL112 (plus a few more), https://github.com/OpenNMS/opennms/pull/5773/files#diff-10602d515015513dda599ee64a46655b8d2b019142e27d29927682bae7c7794fL112 (plus a few more)
3. `until ... pageContainsText` -> `until ... pageContainsText(...).apply(driver)`. I don't think the previous form would have ever failed, because I don't think the `pageContainsText` is evaluated until `apply` is called. -- https://github.com/OpenNMS/opennms/pull/5773/files#diff-10602d515015513dda599ee64a46655b8d2b019142e27d29927682bae7c7794fL112 (and a few more in this file)
4. Removed `<dependency>` for `com.jayway.awaitility` from top-level pom.xml. -- https://github.com/OpenNMS/opennms/pull/5773/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8L4555
5. Removed an unnecessary `<version>` for `awaitility` (it should use what is in the top-level `pom.xml`'s `dependencyManagement`). -- https://github.com/OpenNMS/opennms/pull/5773/files#diff-ec1020b8d3d40919a90742e0517ae80dba67957249bf6937b28f2c45868e0bedL178

You can see the less common changes by running:
```
git show b9b69fba5731a0edde81063e93457b393b10814d | grep '^[+-]' | grep -v -e '<groupId>' -e '^.import ' | grep -B 1 '^[+-][^+-]' | grep '^+++' | awk '{ print $2 }' | xargs -n 1 git show b9b69fba5731a0edde81063e93457b393b10814d | less
```

That returns these files, all of which are in the list of less common changes above (I'm trying to make this 166 file diff reasonably easy to review <grin>):
```
features/events/daemon/src/test/java/org/opennms/netmgt/eventd/EventIpcManagerDefaultImplTest.java
integrations/opennms-vmware/pom.xml
pom.xml
smoke-test/src/test/java/org/opennms/smoketest/ScheduledOutageIT.java
smoke-test/src/test/java/org/opennms/smoketest/graph/GraphRestServiceIT.java
smoke-test/src/test/java/org/opennms/smoketest/rest/ReportRestIT.java
```

### External References

* JIRA (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15399

### DIFF
--- a/core/api/pom.xml
+++ b/core/api/pom.xml
@@ -61,7 +61,7 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>com.jayway.awaitility</groupId>
+      <groupId>org.awaitility</groupId>
       <artifactId>awaitility</artifactId>
       <scope>test</scope>
     </dependency>
--- a/core/ipc/grpc/itests/pom.xml
+++ b/core/ipc/grpc/itests/pom.xml
@@ -66,7 +66,7 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>com.jayway.awaitility</groupId>
+      <groupId>org.awaitility</groupId>
       <artifactId>awaitility</artifactId>
       <scope>test</scope>
     </dependency>
--- a/core/ipc/grpc/itests/src/test/java/org/opennms/core/ipc/grpc/GrpcIpcRpcIT.java
+++ b/core/ipc/grpc/itests/src/test/java/org/opennms/core/ipc/grpc/GrpcIpcRpcIT.java
@@ -28,7 +28,7 @@
 
 package org.opennms.core.ipc.grpc;
 
-import static com.jayway.awaitility.Awaitility.await;
+import static org.awaitility.Awaitility.await;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.not;
 import static org.junit.Assert.assertEquals;
--- a/core/ipc/grpc/itests/src/test/java/org/opennms/core/ipc/grpc/GrpcIpcSinkIT.java
+++ b/core/ipc/grpc/itests/src/test/java/org/opennms/core/ipc/grpc/GrpcIpcSinkIT.java
@@ -28,7 +28,7 @@
 
 package org.opennms.core.ipc.grpc;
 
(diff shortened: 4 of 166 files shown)


RangerRick (approved):
green, ship it 🐿

PR #5805, opened 2023-02-19
NMS-15171: Upgrade HikariCP to 5.x (take 2)

j-white:
JIRA: https://opennms.atlassian.net/browse/NMS-15171

Revised PR for upgrading HikariCP. Updates include:
* All ITs pass cleanly now (limit # of connections in the pool)
* Make sure `HikariCP-java7` doesn't end up in `lib/` (quartz was pulling it in)
* Update log settings to suppress harmless, but possibly scary, stack trace on startup (see https://github.com/pgjdbc/pgjdbc/issues/1102 for details) (our version of Hibernate doesn't support suppress the check in the first place)

### DIFF
--- a/container/features/src/main/resources/features.xml
+++ b/container/features/src/main/resources/features.xml
@@ -307,6 +307,8 @@
     </feature>
     <feature name="quartz" version="${quartzVersion}" description="quartz">
         <feature>c3p0</feature>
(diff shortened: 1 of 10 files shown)


RangerRick (approved):
aha! nice catch

dooo eeeeet

PR #5816, opened 2023-02-21
NMS-15367: Remove elevation from menubar

synqotik:
The `FeatherAppBar` adds some "elevation" using `box-shadow`. However, the "legacy" menu bar does not have this, and also it conflicts a bit with the Meridian color scheme. Just removing this box-shadow. Will ensure this ports to Meridian, or else will open a separate PR for that.

### External References

* JIRA (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15367

### DIFF
--- a/ui/src/components/Layout/Menubar.vue
+++ b/ui/src/components/Layout/Menubar.vue
@@ -723,7 +723,6 @@ body {
 }
 
 .banner .header {
-
   .logo-link.home {
     padding-left: 0;
     margin-right: 1rem;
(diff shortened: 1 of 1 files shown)


aramosvizcarra (approved):
Looks good to me.
I'm not sure if the branch is correct. I tried a build but doesn't seem to be applied in opennms, but the issue is also not present soe I'm guessing this is only Meridian.

synqotik:
Thanks - this PR is for Horizon, I'll be applying this fix to `foundation-2023` for Meridian as well. I will follow up after this builds to make sure the CSS is correct after the build and the shadow has been removed from the Vue menubar.

PR #5817, opened 2023-02-21
fix checkstyle to not need a separate project

RangerRick:
This PR removes the need to have a separate `checkstyle` project to contain custom checkstyle config, instead opting for using the `directory-maven-plugin` to define a property pointing to the top level of the build tree, no matter where a build is happening.

This gets us out of the chicken-and-egg situation we have currently.


christianpape (approved):
LGTM.

PR #5839, opened 2023-02-28
set parameters for is_pr, use them for jira notify

RangerRick:
This PR creates (non-setup) parameters for `is_pr`, so we can decide whether to do jira notifications or not.


RangerRick:
will have to adapt this to dynamic config too once we merge this forward, but it should be easy enough

mershad-manesh (approved):

PR #5844, opened 2023-03-01
Minor EnumLookupPropertyExteneder docs update

amay94:
The second property didn't have the required 'enum-attribute' parameter. Let me know if I should create a jira for this.


Bonrob2 (commented):
Got auto tagged for this, but it looks fine to me.

RangerRick (approved):
yeah, small enough it shouldn't need an issue

PR #5748, opened 2023-01-30
NMS-15326: do not refresh bundles when features get installed

deejgregor:
This unfortunately includes a number of things, but it is all in support of preventing for Karaf bundle refreshes at startup which leads to issues like http://issues.opennms.org/browse/NMS-15326.

## Breakdown of changes

This is best reviewed commit-by-commit, and I've organized them in the groups listed below.

### Service lookup fixes/tweaks

* [BlockingServiceLookup: Log an error if we timed out waiting for a ser…](https://github.com/OpenNMS/opennms/pull/5748/commits/a194b1f9a298ff15514e20b26489760b8db9ad6f) 
* [ServiceLookupBuilder: Fix what appear to be cut-n-paste errors in wai…](https://github.com/OpenNMS/opennms/pull/5748/commits/e646dddbfbee09683cb147f8b4888d64d71f7dbc) 

### Add smoke-test check to for Karaf bundles getting destroyed at startup

* [smoke-test: Fail if we see 'Destroy container' in the karaf log](https://github.com/OpenNMS/opennms/pull/5748/commits/323026ccadc5794b3b5c482e832b7c3e6be8feb8)

### Add support for installing features at boot time of OpenNMSContainer

Note: Without this change, and features starting to use it, the 'Destroy container' check above will happen *before* plugins are loaded. The Cortex plugin checks are changed a few sections down.

* [smoke-test: OpenNMSProfile.withInstallFeature (new): install a featur…](https://github.com/OpenNMS/opennms/pull/5748/commits/2a17d3b572133059584019503aa338c1df6488b7) 

### Make sure we wait Karaf to fully come up

Otherwise, we might miss things getting destroyed late during startup.

* [smoke-test: OpenNMSContainer: Always wait for OpenNMS health check](https://github.com/OpenNMS/opennms/pull/5748/commits/c05ebf9a0b760e370f4e6280b0d66312e38c0732) 
* [smoke-test: OpenNMSContainer: Use the same RestHealthClient as Minion…](https://github.com/OpenNMS/opennms/pull/5748/commits/4bed6411e5a36df7e80ea53ee681acc664714ad2)

### Tweak the Cortex plugin checks to install at boot time and use the plugins in the container

This is our test for https://opennms.atlassian.net/browse/NMS-15326 since it reliably triggers it. Plus some infrastructure to support the changes.

[...]

PR #5851, opened 2023-03-01
NMS-15326: do not refresh bundles when features get installed

amay94:
This is a version of this PR without the karaf extender additions: https://github.com/OpenNMS/opennms/pull/5748

Those will live in develop for now and wait for 32.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15326

### DIFF
--- a/core/soa/src/main/java/org/opennms/core/soa/lookup/ServiceLookupBuilder.java
+++ b/core/soa/src/main/java/org/opennms/core/soa/lookup/ServiceLookupBuilder.java
@@ -39,7 +39,7 @@ public class ServiceLookupBuilder<C, F> {
 
 public static final long GRACE_PERIOD_MS = SystemProperties.getLong("org.opennms.core.soa.lookup.gracePeriodMs", TimeUnit.MINUTES.toMillis(5));
 
-    public static final long WAIT_PERIOD_MS = SystemProperties.getLong("org.opennms.core.soa.lookup.gracePeriodMs", TimeUnit.MINUTES.toMillis(1));
+    public static final long WAIT_PERIOD_MS = SystemProperties.getLong("org.opennms.core.soa.lookup.waitPeriodMs", TimeUnit.MINUTES.toMillis(1));
 
     public static final long LOOKUP_DELAY_MS = SystemProperties.getLong("org.opennms.core.soa.lookup.lookupDelayMs", TimeUnit.SECONDS.toMillis(5));
 
@@ -54,7 +54,7 @@ public static final long GRACE_PERIOD_MS = SystemProperties.getLong("org.opennms
     }
 
     public ServiceLookupBuilder blocking() {
-        return blocking(GRACE_PERIOD_MS, LOOKUP_DELAY_MS, GRACE_PERIOD_MS);
+        return blocking(GRACE_PERIOD_MS, LOOKUP_DELAY_MS, WAIT_PERIOD_MS);
     }
 
     public ServiceLookupBuilder blocking(long gracePeriodInMs, long sleepTimeInMs, long waitTimeMs) {
--- a/core/soa/src/main/java/org/opennms/core/soa/lookup/ServiceLookupBuilder.java
+++ b/core/soa/src/main/java/org/opennms/core/soa/lookup/ServiceLookupBuilder.java
@@ -37,9 +37,15 @@ import org.opennms.core.sysprops.SystemProperties;
 
 public class ServiceLookupBuilder<C, F> {
 
-public static final long GRACE_PERIOD_MS = SystemProperties.getLong("org.opennms.core.soa.lookup.gracePeriodMs", TimeUnit.MINUTES.toMillis(5));
+    public static final long GRACE_PERIOD_MS = SystemProperties.getLong("org.opennms.core.soa.lookup.gracePeriodMs", TimeUnit.MINUTES.toMillis(5));
 
-    public static final long WAIT_PERIOD_MS = SystemProperties.getLong("org.opennms.core.soa.lookup.waitPeriodMs", TimeUnit.MINUTES.toMillis(1));
+    public static final long WAIT_PERIOD_MS = SystemProperties.getLong("org.opennms.core.soa.lookup.waitPeriodMs",
+                                              SystemProperties.getLong("org.opennms.core.soa.lookup.gracePeriodMs",
+                                              TimeUnit.MINUTES.toMillis(1)));
+
+    public static final long BLOCKING_WAIT_PERIOD_MS = SystemProperties.getLong("org.opennms.core.soa.lookup.waitPeriodMs",
+                                                       SystemProperties.getLong("org.opennms.core.soa.lookup.gracePeriodMs",
+                                                       TimeUnit.MINUTES.toMillis(5)));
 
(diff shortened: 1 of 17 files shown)


aramosvizcarra (dismissed):
Codewise looks good to me

core/soa/src/main/java/org/opennms/core/soa/lookup/ServiceLookupBuilder.java
```diff
@@ -39,7 +39,7 @@
 
 public static final long GRACE_PERIOD_MS = SystemProperties.getLong("org.opennms.core.soa.lookup.gracePeriodMs", TimeUnit.MINUTES.toMillis(5));
 
-    public static final long WAIT_PERIOD_MS = SystemProperties.getLong("org.opennms.core.soa.lookup.gracePeriodMs", TimeUnit.MINUTES.toMillis(1));
+    public static final long WAIT_PERIOD_MS = SystemProperties.getLong("org.opennms.core.soa.lookup.waitPeriodMs", TimeUnit.MINUTES.toMillis(1));
```
RangerRick, line 42: Silly question, since I have no idea if anyone overrides `gracePeriodMs` in the real world, but couldn't this cause unexpected behavior since it's getting dropped from 5 to 1 minute?

Should this be changed to fall back to the value of gracePeriodMs if it's set, and only _then_ to 1 minute if both aren't? Or do we care?

RangerRick (approved):
Other than that little nit, I like this. Marking as "approved" even thought that Q could be considered requesting changes. If you don't think it'll impact anyone in a way that matters, then I'd say go for it.

deejgregor (approved):
It looks like the right set of changes to me.

core/soa/src/main/java/org/opennms/core/soa/lookup/ServiceLookupBuilder.java
deejgregor: **Great** catch, and you're right about "couldn't this cause unexpected behavior since it's getting dropped from 5 to 1 minute?". Looking at, I see that there are *two* interrelated issues here with the change to the definition of `WAIT_PERIOD_MS` and using it in `ServiceLookupBuilder.blocking`:
1. We are changing the default wait time of 5 minutes to 1 minute--but only for `ServiceLookupBuilder.blocking` calls, which used `GRACE_PERIOD_MS` before. Note that there are other existing users of `WAIT_PERIOD_MS` that are outside of `ServiceLookupBuilder`.
2. We are changing the property to override from `... gracePeriodMs` to `...waitPeriodMs` (which is completely new) for the wait time for all service lookups.

The first one is which worries me most. I'm not that worried about the second one, as it probably impacts very few, if any, people. 

If we wanted to be really safe, we'd want to have two different fallbacks:
1. One for `WAIT_PERIOD_MS` so that it also checks `...gracePeriodMs` before falling back to the default of 1. This will be for the *existing* consumers of `WAIT_PERIOD_MS`. E.g.: ` WAIT_PERIOD_MS = SystemProperties.getLong("org.opennms.core.soa.lookup.waitPeriodMs", SystemProperties.getLong("org.opennms.core.soa.lookup.gracePeriodMs", TimeUnit.MINUTES.toMillis(1)));`
2. Another one just for `ServiceLookupBuilder.blocking` that does the same thing as above, but with a default of 5 if neither property is found.

Or, we could pull the changes to this file the PR and just merge it separately to `develop`?

FWIW, here are the other files that use `WAIT_PERIOD_MS`.
```
features/telemetry/registry/src/main/java/org/opennms/netmgt/telemetry/protocols/registry/impl/TelemetryAdapterRegistryImpl.java:        this(ServiceLookupBuilder.GRACE_PERIOD_MS, ServiceLookupBuilder.WAIT_PERIOD_MS, ServiceLookupBuilder.LOOKUP_DELAY_MS);
features/telemetry/registry/src/main/java/org/opennms/netmgt/telemetry/protocols/registry/impl/TelemetryConnectorRegistryImpl.java:        this(ServiceLookupBuilder.GRACE_PERIOD_MS, ServiceLookupBuilder.WAIT_PERIOD_MS, ServiceLookupBuilder.LOOKUP_DELAY_MS);
features/telemetry/registry/src/main/java/org/opennms/netmgt/telemetry/protocols/registry/impl/TelemetryListenerRegistryImpl.java:        this(ServiceLookupBuilder.GRACE_PERIOD_MS, ServiceLookupBuilder.WAIT_PERIOD_MS, ServiceLookupBuilder.LOOKUP_DELAY_MS);
features/telemetry/registry/src/main/java/org/opennms/netmgt/telemetry/protocols/registry/impl/TelemetryParserRegistryImpl.java:        this(ServiceLookupBuilder.GRACE_PERIOD_MS, ServiceLookupBuilder.WAIT_PERIOD_MS, ServiceLookupBuilder.LOOKUP_DELAY_MS);
features/telemetry/registry/src/main/java/org/opennms/netmgt/telemetry/protocols/registry/impl/TelemetryServiceRegistryImpl.java:        this(staticServiceSupplier, ServiceLookupBuilder.GRACE_PERIOD_MS, ServiceLookupBuilder.WAIT_PERIOD_MS, ServiceLookupBuilder.LOOKUP_DELAY_MS);
```

amay94:
@deejgregor I think I've captured the changes you recommended in that comment. My understanding is that the intention of `org.opennms.core.soa.lookup.waitPeriodMs` is to allow overrides of how long services are waited for without changing any of the original behavior.

RangerRick (approved):

RangerRick (dismissed):
just noticed, the issue expects this fix for Meridian 2023 as well (which makes sense) -- is this going to `release-31.x` just to give it a test-run before backporting? Or does it need to be retargeted?

amay94:
I think we had discussed putting this in release-31.x first to test it out before putting it in Meridian, but if you're comfortable with the changes I can retarget it now.

deejgregor (approved):
The changes for the wait times look good to me.

deejgregor:
> I think we had discussed putting this in release-31.x first to test it out before putting it in Meridian, but if you're comfortable with the changes I can retarget it now.

I asked in support-eng a bit back:
> Any feeling on which branch to target? I think right now it's pointing at foundation-2023, but I've pulled forward some changes from develop to help with testing that we might not want there yet (the karaf extender health check, which is not a testing change).

@RangerRick said:
> Yeah, this feels like a thing to put in Horizon first and let it shake out a month or two to be certain, before backporting.

So, if this is still the proper approach, I think we're properly targeting `release-31.x` this PR. To make Jira consistent, we should probably split NMS-15326 into two tickets:
1. Get the work into release-31.x (this PR).
2. Another to backport later to foundation-2023.

amay94:
@RangerRick thoughts on where to put this? It sounds like the risky parts were the karaf extender health check changes, which are no longer a part of this.

RangerRick:
> So, if this is still the proper approach, I think we're properly targeting `release-31.x` this PR. To make Jira consistent, we should probably split NMS-15326 into two tickets:
> 
> 1. Get the work into release-31.x (this PR).
> 2. Another to backport later to foundation-2023.

Yeah, this makes sense. Sorry, I had lost the thread of previous conversation when I did my drive-by comment. 😅

PR #5895, opened 2023-03-17
NMS-15323: Make TimeseriesStorageManager throw exception if no impl is found

amay94:
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15323


deejgregor (approved):
Looks good to me!

RangerRick (changes requested):
this should be targeting `foundation-2023` shouldn't it?

RangerRick (approved):
always a fan of more specific exceptions

PR #5898, opened 2023-03-20
NMS-15352: Remove duplicate SNMP intfs from returned list

amay94:
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15352

### DIFF
--- a/opennms-model/src/main/java/org/opennms/netmgt/model/OnmsSnmpInterface.java
+++ b/opennms-model/src/main/java/org/opennms/netmgt/model/OnmsSnmpInterface.java
@@ -771,4 +771,27 @@ public class OnmsSnmpInterface extends OnmsEntity implements Serializable {
         
     }
 
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (!(o instanceof OnmsSnmpInterface)) {
+            return false;
+        }
+        OnmsSnmpInterface that = (OnmsSnmpInterface)o;
+        if (!this.getId().equals(that.getId())) {
+            return false;
+        }
+        if (!this.getPhysAddr().equals(that.getPhysAddr())) {
+            return false;
+        }
+        if (!this.getIfIndex().equals(that.getIfIndex())) {
+            return false;
+        }
+        if (!this.getNodeId().equals(that.getNodeId())) {
+            return false;
+        }
+        return true;
+    }
 }
--- a/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/SnmpInterfaceRestService.java
+++ b/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/SnmpInterfaceRestService.java
@@ -32,6 +32,7 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.ws.rs.Path;
 import javax.ws.rs.core.UriInfo;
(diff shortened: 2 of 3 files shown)


opennms-model/src/main/java/org/opennms/netmgt/model/OnmsSnmpInterface.java
```diff
@@ -771,4 +771,27 @@ public void mergeSnmpInterfaceAttributes(OnmsSnmpInterface scannedSnmpIface) {
...
+        if (!this.getPhysAddr().equals(that.getPhysAddr())) {
+            return false;
+        }
+        if (!this.getIfIndex().equals(that.getIfIndex())) {
+            return false;
+        }
+        if (!this.getNodeId().equals(that.getNodeId())) {
+            return false;
+        }
+        return true;
```
RangerRick, line 795: FYI, you don't have to change it, but java8 makes this a bit easier now with `Objects.equals()`:

```java
return Objects.equals(this.m_id), that.m_id) &&
  Objects.equals(this.m_physAddr, that.m_physAddr) &&
  Objects.equals(this.m_ifIndex, that.m_ifIndex) &&
  Objects.equals(this.m_node, that.m_node);
```

Also, doing it with direct property access means there's no logic performed in getters/setters (for example, `getNodeId()` actually has to check if `m_node` isn't null). In theory this is a tiny performance boost, although I'm not sure how much of it the compiler optimizes away nowadays anyway.

RangerRick (approved):

opennms-model/src/main/java/org/opennms/netmgt/model/OnmsSnmpInterface.java
christianpape: > FYI, you don't have to change it, but java8 makes this a bit easier now with `Objects.equals()`:
> 
> ```java
> return Objects.equals(this.m_id, that.m_id) &&
>   Objects.equals(this.m_physAddr, that.m_physAddr) &&
>   Objects.equals(this.m_ifIndex, that.m_ifIndex) &&
>   Objects.equals(this.m_node, that.m_node);
> ```
> 
> Also, doing it with direct property access means there's no logic performed in getters/setters (for example, `getNodeId()` actually has to check if `m_node` isn't null). In theory this is a tiny performance boost, although I'm not sure how much of it the compiler optimizes away nowadays anyway.

The point is that checks with Objects.equals() and properties is less prone to NPEs, e.g. `getNodeId().equals(...)` will result in a NPE if `getNodeId()` returns `null`. I would also suggest to add a `hashCode()` method in this PR as well.

christianpape (changes requested):

opennms-model/src/main/java/org/opennms/netmgt/model/OnmsSnmpInterface.java
RangerRick: @christianpape AH you're right, overriding `hashCode()` is a definite requirement if you're overriding `.equals(...)`, so might as well make these changes too.

`core/ipc/rpc/common/src/main/java/org/opennms/core/rpc/echo/EchoRequest.java` is a good example of the current idiomatic way of doing this.

```java
    @Override
    public int hashCode() {
        return Objects.hash(id, message, body, location, delay,
                shouldThrow, systemId);
    }

    @Override
    public boolean equals(Object obj) {
        if (this == obj)
            return true;
        if (obj == null)
            return false;
        if (getClass() != obj.getClass())
            return false;
        final EchoRequest other = (EchoRequest) obj;
        return Objects.equals(this.id, other.id) &&
                Objects.equals(this.message, other.message) &&
                Objects.equals(this.body, other.body) &&
                Objects.equals(this.location, other.location) &&
                Objects.equals(this.delay, other.delay) &&
                Objects.equals(this.shouldThrow, other.shouldThrow) &&
                Objects.equals(this.systemId, other.systemId);
    }
```

RangerRick (changes requested):

christianpape (approved):
LGTM.

RangerRick (approved):

RangerRick (approved):

PR #5900, opened 2023-03-20
NMS-15065: upgrade CXF to 3.5.5

RangerRick:
After doing a lot of digging, it turns out it was surprisingly difficult to move from 3.5.1 to 3.5.5 because of a bunch of stuff that just happened to work with our existing rickety configuration. :D

The biggest issue is that we were mixing versions of CXF dependencies. I'm guessing it's because newer version of the CXF OpenAPI (Swagger) stuff was more strict about the objects you pass to it. Rather than pinning to an old version, I fixed the classes that cause issues, by using annotations.

Then, while testing CXF stuff, I was ending up with some very strange classpath issues that it turns out came from a mix of changed contents in the CXF jars, and mixed Jackson dependencies. I fixed this first by cleaning up Jackson dependencies more (and bumping to the latest 2.x because it was just a micro version update).

Finally, I did the actual CXF upgrade itself. It was failing to start because of _additional_ classpath stuff, that's fixed by making sure our webapp's `WEB-INF/lib` doesn't mask anything already provided from the main (`$OPENNMS_HOME/lib`) classpath.

The _actual_ fix for using CXF 3.5.5 with latest swagger is to [revert to an older behavior in CXF which changed as of 3.5.2](https://issues.apache.org/jira/browse/CXF-8684?focusedCommentId=17701418&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17701418) because we have a number of REST classes that throw an error without setting an entity.

This issue isn't done yet, I'll be doing additional work to see how far back I can port these changes. I suspect/hope it can at least get to `foundation-2022` without much trouble, perhaps farther.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15065


aramosvizcarra (approved):
LGTM!

PR #5906, opened 2023-03-22
NMS-15471: handle pre-meridian-2019 event datetime formats

RangerRick:
This PR implements a layer of abstraction to how event datetimes are parsed. I implemented it with an interface rather than conditional code so that the JIT will optimize it away at runtime and (hopefully) not cause any changes in event parsing performance.

You can enable the old-style event format by adding `-Dorg.opennms.events.legacyFormatter=true` to `ADDITIONAL_MANAGER_OPTIONS` in `opennms.conf`.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15471

### DIFF
--- a/features/events/api/src/main/java/org/opennms/netmgt/events/api/LegacyDatetimeFormatter.java
+++ b/features/events/api/src/main/java/org/opennms/netmgt/events/api/LegacyDatetimeFormatter.java
@@ -31,6 +31,8 @@ package org.opennms.netmgt.events.api;
 import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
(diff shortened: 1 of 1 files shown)


amay94 (approved):
LGTM. Is the LegacyDatetimeFormatter an old class that you revived?

RangerRick:
> LGTM. Is the LegacyDatetimeFormatter an old class that you revived?

Yes and no. That code used to just be directly _in_ the `EventConstants` `parseDate()` and `formatDate()` methods, and we made it a breaking change. This is being introduced to allow a customer to do a smoother migration from code expecting to push events in the old format.

PR #5910, opened 2023-03-23
NMS-14589: update to newer extremecomponents with updated POI dependencies

RangerRick:
See https://github.com/OpenNMS/extremecomponents for the terrible lengths I will go to not port this silly page to JMesa.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-14589


aramosvizcarra (approved):
Looks good to me (at least from what I could see here and under the extremecomponents...)

PR #5922, opened 2023-03-28
NMS-13216: always handle reload event in Scriptd

RangerRick:
@jeffgdotorg noticed that my previous "fix" would never handle the reload event, because the processor would never get registered if there were no Scriptd scripts. :D

This fixes it to process reload events separately.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-13216


christianpape (approved):
LGTM.

PR #5998, opened 2023-04-14
NMS-15334: Catch config reloads when editing new thresholds

amay94:
I was able to recreate the issue by reloading threshd config while I had the page open for editing a new threshold / expression. We don't write the new object to disk until it's fully edited, so reloading the config wipes it from memory. This fix just creates a new default object to update if the previous one has disappeared. 

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15334

### DIFF
--- a/opennms-webapp/src/main/java/org/opennms/web/controller/admin/thresholds/ThresholdController.java
+++ b/opennms-webapp/src/main/java/org/opennms/web/controller/admin/thresholds/ThresholdController.java
@@ -225,8 +225,7 @@ public class ThresholdController extends AbstractController implements Initializ
         modelAndView.addObject("moveDownButtonTitle", MOVEDOWN_BUTTON_TITLE);
     }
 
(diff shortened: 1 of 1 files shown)


synqotik (approved):
Code and logic look good and make sense to me. Also, nice job refactoring!

RangerRick (approved):
I don't know the guts well enough to say, but is it possible the config would have changed/reloaded but _without_ the number of entries changing? Either way, this is better than before. :D

christianpape (approved):
LGTM.

amay94:
> I don't know the guts well enough to say, but is it possible the config would have changed/reloaded but without the number of entries changing? Either way, this is better than before. :D

If someone had manually updated the threshd config file and added a new threshold to this specific group while the editing page was open, this could happen. To address that, I'd want to get rid of the `thresholdIndex` parameter being passed around and only add the new threshold / expression once when editing is complete.

PR #6015, opened 2023-04-19
Jira/nms 14936 cacheable https responses cache control directive missing or misconfigured smoke

dherdt:
* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-14947


j-white:
Should we exclude static assets?

`vendor.min.js` is nearing 3MB for example - wouldn't we want this to be cached?

dherdt:
> Should we exclude static assets?
> 
> `vendor.min.js` is nearing 3MB for example - wouldn't we want this to be cached?
Is /opennms/assets/* the only path to exclude?

RangerRick (dismissed):

PR #6017, opened 2023-04-19
NMS-15606: Remove unneeded config option to avoid side effects

amay94:
This line is not needed and is causing maxMessageSize to get overwritten, see https://github.com/graylog-labs/syslog4j-graylog2/blob/master/src/main/java/org/graylog2/syslog4j/impl/AbstractSyslogConfig.java#L64

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15606


aramosvizcarra (approved):
LGTM!

christianpape (approved):
LGTM.

PR #6030, opened 2023-04-24
Asynchronous polling engine & benchmarking tool

j-white:
JIRA: https://opennms.atlassian.net/browse/NMS-15623

Here's a patch that adds support for asynchronous polling to pollerd. The behaviour is optional is done in a very unobtrusive way, allowing us to vet the solution is practice without affecting existing solutions.

When enabled, polls are triggered by the thread pools, using a configurable rate-limiter for back-pressure, and then the thread is released. When the poll is complete, execution is passed to cached thread poll (implicitly bounded by the rate limit) and the node lock  is acquired before processing the results.

### DIFF
--- a/docs/modules/operation/pages/deep-dive/service-assurance/configuration.adoc
+++ b/docs/modules/operation/pages/deep-dive/service-assurance/configuration.adoc
@@ -64,7 +64,7 @@ During evaluation of an expression, the following scopes are available:
 
 By default, the number of polls active at any given time is limited by the number of threads that are allocated to pollerd.
 Each thread is responsible for one poll, and will block until the poll is completed; this is true even when the polling is done by a Minion at a remote location.
-In this case, the thread is blocked for the duration of the RPC.
+In this case, the thread is blocked for the duration of the remote procedure call.
 
 If you have hundreds of locations and hundreds of nodes at each location, you may want to reduce the number of threads required on your {page-component-title} instance.
 Each thread allocates a fixed amount of system memory.
--- a/docs/modules/operation/pages/deep-dive/service-assurance/configuration.adoc
+++ b/docs/modules/operation/pages/deep-dive/service-assurance/configuration.adoc
@@ -58,3 +58,31 @@ During evaluation of an expression, the following scopes are available:
 * Node metadata
 * Interface metadata
 * Service metadata
+
+[[ga-pollerd-configuration-async]]
+== Asynchronous engine
+
+By default, the number of polls active at any given time is limited by the number of threads that are allocated to the daemon.
+Each thread is responsible for one poll, and will block until the poll is completed.
+This is true, even when the polling is done by a Minion at a remote location.
+In this case, the thread will block for the duration of the RPC.
+
+In scenarios where there are hundreds of different locations, and hundreds of nodes at each of these, it can be desirable to reduce the number of threads required on the core.
+Each thread allocates a fixed amount of system memory, and too many threads active at a given time can extraneous load on the system due to context switching.
+
+For these reasons, we have developed a new asynchronous mode for the poller.
+When enabled, threads are released while the polls are active, allowing Minions to process more polls than there are threads.
+
+To enable the feature, set the following attributes in the top level element of `etc/poller-configuration.xml`:
+[source, xml]
+----
+<poller-configuration threads="30"
+                      asyncPollingEngineEnabled="true" <1>
+                      maxConcurrentAsyncPolls="200" <2>
+----
+
+<1> Enable the asynchronous engine.
+<2> Used to limit the number of polls active, since each of these still consume some memory resources.
+
+In this mode, threads are only used to trigger the execution of the poll, so the number can be typically reduced.
+The actual number of polls active, or in-flight, can be observed via the `NumPollsInFlight` metrics associated with the daemon's MBeans.
--- a/docs/modules/operation/pages/deep-dive/service-assurance/configuration.adoc
+++ b/docs/modules/operation/pages/deep-dive/service-assurance/configuration.adoc
@@ -62,27 +62,27 @@ During evaluation of an expression, the following scopes are available:
 [[ga-pollerd-configuration-async]]
 == Asynchronous engine
 
-By default, the number of polls active at any given time is limited by the number of threads that are allocated to the daemon.
-Each thread is responsible for one poll, and will block until the poll is completed.
-This is true, even when the polling is done by a Minion at a remote location.
-In this case, the thread will block for the duration of the RPC.
+By default, the number of polls active at any given time is limited by the number of threads that are allocated to pollerd.
+Each thread is responsible for one poll, and will block until the poll is completed; this is true even when the polling is done by a Minion at a remote location.
+In this case, the thread is blocked for the duration of the RPC.
 
-In scenarios where there are hundreds of different locations, and hundreds of nodes at each of these, it can be desirable to reduce the number of threads required on the core.
-Each thread allocates a fixed amount of system memory, and too many threads active at a given time can extraneous load on the system due to context switching.
+If you have hundreds of locations and hundreds of nodes at each location, you may want to reduce the number of threads required on your {page-component-title} instance.
+Each thread allocates a fixed amount of system memory.
+Having too many threads active at a given time can cause extensive load on the system.
 
-For these reasons, we have developed a new asynchronous mode for the poller.
-When enabled, threads are released while the polls are active, allowing Minions to process more polls than there are threads.
+When enabled, the asynchronous polling mode mitigates this load.
+It allows threads to be released while polls are active, enabling Minions to process more polls than the number of threads available.
 
-To enable the feature, set the following attributes in the top level element of `etc/poller-configuration.xml`:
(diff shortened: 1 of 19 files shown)


docs/modules/operation/pages/deep-dive/service-assurance/configuration.adoc
```diff
@@ -58,3 +58,31 @@ During evaluation of an expression, the following scopes are available:
 * Node metadata
 * Interface metadata
 * Service metadata
+
+[[ga-pollerd-configuration-async]]
+== Asynchronous engine
+
+By default, the number of polls active at any given time is limited by the number of threads that are allocated to the daemon.
+Each thread is responsible for one poll, and will block until the poll is completed.
+This is true, even when the polling is done by a Minion at a remote location.
```
emarsh1995, line 67:
```suggestion
Each thread is responsible for one poll, and will block until the poll is completed; this is true even when the polling is done by a Minion at a remote location.
```

docs/modules/operation/pages/deep-dive/service-assurance/configuration.adoc
```diff
@@ -58,3 +58,31 @@ During evaluation of an expression, the following scopes are available:
...
+In this case, the thread will block for the duration of the RPC.
```
emarsh1995, line 68: Here, are we using RPC as an industry-standard term (i.e. Remote Procedure Call)?

```suggestion
In this case, the thread is blocked for the duration of the RPC.
```

docs/modules/operation/pages/deep-dive/service-assurance/configuration.adoc
```diff
@@ -58,3 +58,31 @@ During evaluation of an expression, the following scopes are available:
...
+
+In scenarios where there are hundreds of different locations, and hundreds of nodes at each of these, it can be desirable to reduce the number of threads required on the core.
```
emarsh1995, line 70:
```suggestion
If you have hundreds of locations and hundreds of nodes at each location, you may want to reduce the number of threads required on your {page-component-title} instance.
```

docs/modules/operation/pages/deep-dive/service-assurance/configuration.adoc
```diff
@@ -58,3 +58,31 @@ During evaluation of an expression, the following scopes are available:
...
+Each thread allocates a fixed amount of system memory, and too many threads active at a given time can extraneous load on the system due to context switching.
```
emarsh1995, line 71:
```suggestion
Each thread allocates a fixed amount of system memory.
Having too many threads active at a given time can cause extensive load on the system.
```

docs/modules/operation/pages/deep-dive/service-assurance/configuration.adoc
```diff
@@ -58,3 +58,31 @@ During evaluation of an expression, the following scopes are available:
...
+
+For these reasons, we have developed a new asynchronous mode for the poller.
+When enabled, threads are released while the polls are active, allowing Minions to process more polls than there are threads.
```
emarsh1995, line 74:
```suggestion
When enabled, the asynchronous polling mode mitigates this load.
It allows threads to be released while polls are active, enabling Minions to process more polls than the number of threads available.
```

docs/modules/operation/pages/deep-dive/service-assurance/configuration.adoc
```diff
@@ -58,3 +58,31 @@ During evaluation of an expression, the following scopes are available:
...
+
+To enable the feature, set the following attributes in the top level element of `etc/poller-configuration.xml`:
```
emarsh1995, line 76:
```suggestion
To enable the feature, set the following attributes in the top-level element of `etc/poller-configuration.xml`:

```

docs/modules/operation/pages/deep-dive/service-assurance/configuration.adoc
```diff
@@ -58,3 +58,31 @@ During evaluation of an expression, the following scopes are available:
...
+[source, xml]
+----
+<poller-configuration threads="30"
+                      asyncPollingEngineEnabled="true" <1>
+                      maxConcurrentAsyncPolls="200" <2>
+----
+
+<1> Enable the asynchronous engine.
```
emarsh1995, line 84:
```suggestion
----
<1> Enables the asynchronous polling engine.
```

docs/modules/operation/pages/deep-dive/service-assurance/configuration.adoc
```diff
@@ -58,3 +58,31 @@ During evaluation of an expression, the following scopes are available:
...
+<2> Used to limit the number of polls active, since each of these still consume some memory resources.
+
+In this mode, threads are only used to trigger the execution of the poll, so the number can be typically reduced.
```
emarsh1995, line 87:
```suggestion
In this mode, threads are used only to trigger a poll to run; this means that the number of threads can typically be reduced.
```

docs/modules/operation/pages/deep-dive/service-assurance/configuration.adoc
```diff
@@ -58,3 +58,31 @@ During evaluation of an expression, the following scopes are available:
...
+The actual number of polls active, or in-flight, can be observed via the `NumPollsInFlight` metrics associated with the daemon's MBeans.
```
emarsh1995, line 88:
```suggestion
You can see the actual number of active (in-flight) polls by reviewing the `NumPollsInFlight` metrics associated with pollerd's MBeans.
```

features/poller/shell/src/main/java/org/opennms/netmgt/poller/shell/Benchmark.java
```diff
@@ -0,0 +1,288 @@
...
+import java.util.Date;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+@Command(scope = "opennms", name = "poller-benchmark", description = "Benchmark pollerd by creating N nodes/services and waiting for them all to be polled.")
```
emarsh1995, line 73:
```suggestion
@Command(scope = "opennms", name = "poller-benchmark", description = "Benchmark pollerd by creating N nodes/services and waiting for them to be polled.")
```

emarsh1995 (dismissed):
Two minor changes

opennms-config-jaxb/src/main/java/org/opennms/netmgt/config/poller/PollerConfiguration.java
```diff
@@ -298,7 +323,9 @@ public boolean equals(Object o) {
                 && Objects.equals(getDefaultCriticalPathRetries(), that.getDefaultCriticalPathRetries())
                 && Objects.equals(getNodeOutage(), that.getNodeOutage())
                 && Objects.equals(getPackages(), that.getPackages())
-                && Objects.equals(getMonitors(), that.getMonitors());
+                && Objects.equals(getMonitors(), that.getMonitors())
+                && Objects.equals(getAsyncPollingEngineEnabled(), that.getAsyncPollingEngineEnabled())
+                && Objects.equals(getMaxConcurrentAsyncPolls(), that.getMaxConcurrentAsyncPolls());
```
RangerRick, line 328: Here's a fun question -- are config objects "equal" if one explicitly sets max concurrent to `100` and the other is null but gets it implicitly?

Arguably they are _not_ equal, because you're serializing the fields, and they wouldn't write out equivalent XML or whatever.

RangerRick (changes requested):
Other than that, I don't see anything in particular jumping out at me.

Very nice!

opennms-config-jaxb/src/main/java/org/opennms/netmgt/config/poller/PollerConfiguration.java
j-white: Agreed. This behaves the same way as other properties in the object do currently though, hesitant to change that in this scope.

RangerRick: I dunno, it's develop, seems like that's the time to do it...

docs/modules/operation/pages/deep-dive/service-assurance/configuration.adoc
```diff
@@ -58,3 +58,31 @@ During evaluation of an expression, the following scopes are available:
...
+To enable the feature, set the following attributes in the top-level element of `etc/poller-configuration.xml`:
+
+[source, xml]
+----
+<poller-configuration threads="30"
+                      asyncPollingEngineEnabled="true" <1>
+                      maxConcurrentAsyncPolls="200" <2>
+----
+<1> Enables the asynchronous polling engine.
+<2> Used to limit the number of polls active, since each of these still consume some memory resources.
```
Bonrob2, line 85:
```suggestion
<2> Used to limit the number of polls active, since each of these still consumes some memory resources.
```

Bonrob2 (commented):
minor docs change, thanks!

Bonrob2 (approved):
LGTM. Thanks!

christianpape (approved):
LGTM.

PR #6053, opened 2023-04-28
NMS-15033: Initializing asset record in the constructer since getAssetRecord() will be called anyway

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-15033


RangerRick:
Yeah, there's a bunch of model code where we have this pattern to remain "backwards-compatible" with the behavior of the old Castor-generated classes, but it seems a bit silly this far removed from that conversion. :)

RangerRick (approved):
Sorry, you branched from 31.x that I broke :(
Pretty sure those failures aren't yours.

Should still merge `release-31.x` once I've got it fixed to be sure there isn't any weird fallout.

RangerRick (approved):

PR #6071, opened 2023-05-02
NMS-15637: Fix SNMPv3 support for AES256

j-white:
JIRA: https://opennms.atlassian.net/browse/NMS-15637

Manually register the AES-256 protocol so that it *just works*.


dino2gnt:
Going to go ahead and merge this to make sure it gets into Wednesday's releases.

PR #6094, opened 2023-05-08
NMS-15636: enable third-party txt generation for packages

RangerRick:
Turns out we _were_ generating the third-party license txt file, but it only made it into the tarball version (used in ocis) and not the RPM and Debian packages.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15636


aramosvizcarra (approved):
LGTM!

PR #6103, opened 2023-05-09
NMS-15493: Usage stats, don't display notice if opted-out

synqotik:
If user already opted-out of Usage Statistics Sharing, do not display the notice.

Also fixes a bug where `initialNoticeAcknowledged` should be set to an actual `boolean` value and not `null`.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15493


aramosvizcarra (approved):
LGTM!

RangerRick (approved):

PR #6107, opened 2023-05-10
NMS-15506: update to Spring Security 4.2

RangerRick:
This PR upgrades our Spring Security to 4.2. This required a number of adaptations to our configs, _most_ of which are documented on [Spring Security's migration page](https://docs.spring.io/spring-security/site/migrate/current/3-to-4/html5/migrate-3-to-4-xml.html).

Notable changes include:

* update to spring-security `4.2.20.RELEASE` and spring-ldap `2.3.8.RELEASE`
* make it clear that the spring-security we're using in features is one of our own making, rather than what's provided in Karaf's bundles
* clean up `custom.properties` to match the current classes provided by spring and spring-security
* update all application contexts to use spring security 4.2 XSD references
* add options to `<http />` and related tags in application contexts, including putting defaults for login/logout URLs which were previously implicit and changed in spring-security 4
* add a new `XRequestedWithAwareLoginUrlAuthEntryPoint` class which is similar to the basic auth version; spring-security 4's workflow with filters and entry points changed subtly, and I needed to make sure their behavior made sense between interactive URLs vs basic-only (aka REST)
* deleted a bunch of old `loginTestWar` stuff that is part of tests that haven't been run for years
* add a `<http-firewall />` entry because the `DefaultHttpFirewall` is no longer the default 😅
* a few changes to smoke tests to handle API changes and subtle differences in how redirects are handled

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15506


RangerRick:
Since https://github.com/OpenNMS/opennms/pull/6111 got approved, I'm going to go ahead and merge this as well.

PR #6118, opened 2023-05-15
add an SSH server for junit tests

RangerRick:
My work laptop doesn't have `sshd` running, so a few ssh-related tests fail. This PR fixes that by not relying on an external sshd, instead creating one using Apache Mina.


aramosvizcarra (approved):
Nice, looks really good!

PR #6133, opened 2023-05-18
NMS-15061: shade Apache Axis to limit exposure

RangerRick:
The original version of this work ended up being pretty invasive.  Instead, a different tack was taken:

1. shade axis so that it is not in its default package, and limit it to just the 2 plugin jars it is used in
2. port the Remedy plugin to OPA and mark the existing one deprecated: https://github.com/OpenNMS/opennms-opa-plugin-ticketing-remedy

This makes the surface area of this change reasonably small, but still limits exposure.
In Horizon 32, the OTRS and Remedy plugins will be removed entirely. (see https://github.com/OpenNMS/opennms/pull/6150)

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15061


christianpape (approved):
LGTM.

PR #6147, opened 2023-05-24
NMS-15696: Remove @Transactional annotation from JdbcFilterDao

j-white:
JIRA: https://opennms.atlassian.net/browse/NMS-15696

This class doesn't use Spring/Hibernate, so injecting transaction support via AOP adds unnecessary overhead and leaves us with possible deadlock scenarios. More details in the associated JIRA.


RangerRick (approved):
The question is how far back this should go, if it could potentially affect PoweredBy. (aka, should it go back to 2018?)

I'm assuming this code hasn't changed in _ages_.

j-white:
Could be backported and should cherry-pick cleanly. It's a lot less likely to be a problem there since we don't use the FilterDao as heavily.

PR #6246, opened 2023-06-20
NMS-15108: Don't attempt to delete the service if it's not found.

j-white:
This operation may block and keep this thread alive until the node lock is aquired.

It's possible/expected that this condition will happen, and this function should not be reponsible for taking action.

Reverts code from NMS-13508.

JIRA: https://opennms.atlassian.net/browse/NMS-15108


j-white:
The patch for NMS-13508 was merged in `develop` for Horizon 30.0.0 - so we shouldn't need to backport this any further.

christianpape (approved):
LGTM.

PR #6249, opened 2023-06-21
NMS-15732: Allowing relative redirects

christianpape:
The problem occurs on page where we use relative redirects. In the Response class these redirects are converted to absolute URLs including the scheme and since this class does not know that the current location is HTTPS it is converted to the HTTP URL. Firefox and some other browsers allow the call because it is immediatly redirected to the correspondig HTTPS page. Chrome seems more picky and does not allow to query the HTTP page. The fix is to allow relative redirects in our jetty configuration. In this case relative URLs will not be converted to absolute ones, so the scheme is not affected and stays the same. 

* JIRA: https://opennms.atlassian.net/browse/NMS-15732

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/opennms.properties
+++ b/opennms-base-assembly/src/main/filtered/etc/opennms.properties
@@ -354,6 +354,9 @@ org.opennms.netmgt.jetty.sendServerVersion=false
 # Provisioning web UI)
 org.eclipse.jetty.server.Request.maxFormKeys=2000
 
+# Allow relative redirects, see NMS-15732
+#org.opennms.netmgt.jetty.relativeRedirectAllowed=true
+
 ###### JETTY HTTPS SUPPORT ######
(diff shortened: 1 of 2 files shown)


synqotik:
Looks good.

One thing to check is that all the Vue pages use the results of the Rest API call `opennms/api/v2/menu`, `baseHref` and `homeUrl` fields.

These are calculated in `MenuProvider.getMainMenu()`, and ultimately from `HttpMenuRequestContext.calculateUrlBase()` and `org.opennms.web.api.Util.calculateUrlBase()` which checks that `opennms.web.base-url` property.

So it would seem in this case if that `base-url` is set to `https` the Vue pages should all work correctly, but should double check.

synqotik (dismissed):
Looks good. See my comment, just wanted to confirm the Vue pages are working correctly in this scenario.

RangerRick (dismissed):

RangerRick (approved):

PR #6258, opened 2023-06-22
NMS-15788: switch Horizon 33 to UBI

RangerRick:
This PR merges our changes to the base container to use RedHat's UBI.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15788

### DIFF
--- a/opennms-container/core/plugins.sh
+++ b/opennms-container/core/plugins.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+set -euo pipefail
+IFS=$'\n\t'
+
 export ALEC_VERSION="latest"
 export CLOUD_VERSION="latest"
 export CORTEX_VERSION="latest"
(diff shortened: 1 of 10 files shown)


fooker (approved):
Not tested, but there is nothing that immediately stands out

opennms-container/core/plugins.sh
```diff
@@ -9,29 +12,28 @@ export DEPLOY_FOLDER="/opt/usr-plugins"
...
+microdnf -y install cpio python3-pip jq
 pip3 install --upgrade cloudsmith-cli 
 
 mkdir ~/test
 cd ~/test || exit
-artifact_urls=$(cloudsmith list packages --query="opennms-alec-plugin version:$ALEC_VERSION format:deb" opennms/common -F json  | jq -r '.data[].cdn_url')
+artifact_urls=$(cloudsmith list packages --query="opennms-alec-plugin version:$ALEC_VERSION format:rpm" opennms/common -F json  | jq -r '.data[].cdn_url')
 for url in $artifact_urls; do 
- wget "$url"
+ curl -L -O "$url"
```
deejgregor, line 22:
```suggestion
 curl -sSF -L -O "$url"
```

With curl, `-sSf` everywhere. :)

opennms-container/minion/plugins.sh
```diff
@@ -1,13 +1,14 @@
 #!/bin/bash
 
+set -euo pipefail
```
deejgregor, line 3:
```suggestion
set -euo pipefail
trap 's=$?; echo "$0: Error on line "$LINENO": $BASH_COMMAND"; exit $s' ERR
```

Whenever you use `set -e`, I **strongly** recommend something like this trap, otherwise you will get **silent failures** if a command returns a non-zero exit status but doesn't print anything, and at some point, you will really wish you know why the script exited. It might. not matter for **this** script, but you never know when someone will copy this script to create a new script.

Even if you still get some error output, if you use that tool a number of times, you might not know *which* instance of it failed. The trap will let you know which line (usually; sometimes it is buggy, but most of the time it's right or will point you at the right command).

christianpape (dismissed):
LGTM.

mershad-manesh (dismissed):
Overall, the changes look good to me

PR #6275, opened 2023-06-27
NMS-15768: Fixed scan-interval handling

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-15768

### DIFF
--- a/opennms-provision/opennms-provision-persistence/src/main/java/org/opennms/netmgt/provision/persist/StringIntervalAdapter.java
+++ b/opennms-provision/opennms-provision-persistence/src/main/java/org/opennms/netmgt/provision/persist/StringIntervalAdapter.java
@@ -57,6 +57,9 @@ public class StringIntervalAdapter extends XmlAdapter<String, Duration> {
     /** {@inheritDoc} */
     @Override
     public Duration unmarshal(final String v) {
+        if (v == null) {
+            return null;
+        }
         if ("0".equals(v.trim())) {
             return Duration.ZERO;
         }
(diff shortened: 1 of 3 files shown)


opennms-provision/opennms-provision-persistence/src/main/java/org/opennms/netmgt/provision/persist/StringIntervalAdapter.java
```diff
@@ -57,10 +57,13 @@ public String marshal(final Duration v) {
     /** {@inheritDoc} */
     @Override
     public Duration unmarshal(final String v) {
-        if ("0".equals(v)) {
+        if ("0".equals(v.trim())) {
```
RangerRick, line 60: Could `v` ever be null? This would introduce an NPE if so...

opennms-provision/opennms-provision-persistence/src/main/java/org/opennms/netmgt/provision/persist/StringIntervalPropertyEditor.java
```diff
@@ -39,10 +39,14 @@ public class StringIntervalPropertyEditor extends PropertyEditorSupport implemen
     /** {@inheritDoc} */
     @Override
     public void setAsText(final String text) throws IllegalArgumentException {
-        if ("0".equals(text)) {
+        if ("0".equals(text.trim())) {
```
RangerRick, line 42: same here with `text`

RangerRick (changes requested):
other than that, looks fine

opennms-provision/opennms-provision-persistence/src/main/java/org/opennms/netmgt/provision/persist/StringIntervalAdapter.java
christianpape: null-check added

RangerRick (approved):

PR #6282, opened 2023-06-27
NMS-15798: Cache compiled scripts to avoid Groovy classloader leak

j-white:
JIRA: https://opennms.atlassian.net/browse/NMS-15798

Here we create a cache for the compile Groovy scripts that is shared among different instances of the ScriptPolicy. This avoid having to recompile the script for every scan on every node, which reduces the impact of Groovy classloader leaks.


RangerRick (approved):

christianpape (approved):
LGTM.

PR #6288, opened 2023-06-29
NMS-15704: Securing Xml parser factory

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-15704


RangerRick (approved):
LGTM

I wonder if we should make a generic "helper" that does this in `JaxbUtils` rather than reimplementing this in many places. (not a blocker for merge, just musing...)

PR #6297, opened 2023-06-30
NMS-15809: Avoid errors in postinstall scripts when directories are empty

UberPinguin:
If /data or /data/tmp are present but empty, find throws an error when searching 'directory/*'. Instead, increase the -maxdepth argument and add a -mindepth argument to achieve the same result without risking a 'No such file or directory' error.

### All Contributors

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15809

### DIFF
--- a/tools/packages/opennms/opennms.spec
+++ b/tools/packages/opennms/opennms.spec
@@ -922,9 +922,9 @@ done
 
 printf -- "- cleaning up \$OPENNMS_HOME/data... "
 if [ -d "$ROOT_INST/data" ]; then
-	find "$ROOT_INST/data/"* -maxdepth 0 -name tmp -o -name history.txt -prune -o -print0 | xargs -0 rm -rf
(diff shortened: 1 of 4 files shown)


tools/packages/opennms/opennms.spec
```diff
@@ -922,9 +922,9 @@ done
 
 printf -- "- cleaning up \$OPENNMS_HOME/data... "
 if [ -d "$ROOT_INST/data" ]; then
-	find "$ROOT_INST/data/"* -maxdepth 0 -name tmp -o -name history.txt -prune -o -print0 | xargs -0 rm -rf
```
j-white, line 925: `-name history.txt` was removed from this one, intentional?

j-white:
See https://github.com/OpenNMS/opennms/pull/6352

PR #6321, opened 2023-07-05
NMS-15808: enlinkd enum refactor fix

RangerRick:
Note: this PR looks like a lot has changed, but actually the only real difference is that I got rid of `org.opennms.netmgt.enlinkd.service.api.ProtocolSupported` and changed everything that relied upon it to use the `TopologyProtocol` from OPA (1.6 and up) instead. I also cleaned up the Karaf feature files to use a single feature that defines the `protobuf` dependency.

These changes get rid of one of the places where code is duplicated. :)

Unfortunately, the updates _also_ needed to get put into the `opennms-kafka-producer.proto` file, which is a part of both Horizon _and_ ALEC.

Long-term, we should ideally move the proto file (and its generated classes) to OPA, and let everything rely on a single copy everywhere, and just use the enum generated as part of the protobuf file instead rather than reproducing it in multiple places, but that will have to be an OPA 2.0 thing since `TopologyProtocol` pre-existed in OPA and is relied upon already.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15808

### DIFF
--- a/container/features/src/main/resources/features-minion.xml
+++ b/container/features/src/main/resources/features-minion.xml
@@ -134,7 +134,7 @@
       <feature>opennms-dnsresolver-api</feature>
       <feature>dropwizard-metrics</feature>
       <feature>opennms-util</feature>
-      <bundle>mvn:com.google.protobuf/protobuf-java/${protobufVersion}</bundle>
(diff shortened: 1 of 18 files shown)


christianpape (approved):
LGTM.

christianpape (dismissed):
LGTM.

christianpape (approved):

PR #6336, opened 2023-07-07
NMS-15839: Measurements API: Include node metadata even if resource not found

synqotik:
The Measurements API returns metadata including information on nodes when responding to a query for resource attribute values, for example by the Grafana plugin Performance data source.

However if the requested resource is not found, for example because SNMP collection has not yet occurred on that node, then no node metadata is returned, even though the node metadata exists. When the Grafana plugin `nodeToLabel` function is used, it doesn't find the node metadata including the label name and so the label displays as the FS:FID. This causes the issue in https://opennms.atlassian.net/browse/HELM-427.

This will return metadata for the node even if the resource was not found, if the `relaxed` flag was set.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15839


RangerRick (approved):
looks good, wonder if this should go back to at least 2023 though since it would be nice to have it work for meridian users as well

PR #6351, opened 2023-07-12
fix a number of sonar warnings in recent code

RangerRick:
Our sonar reports are failing, fix a few errors that show up in recent sonar runs.


christianpape (approved):
LGTM.

PR #6369, opened 2023-07-13
fix a number of NPE issues flagged by sonar

RangerRick:
Assuming these pass fine, they should be cherry-picked back to 32 or perhaps even `foundation-2023`.

### DIFF
--- a/.circleci/build-triggers.override.json
+++ b/.circleci/build-triggers.override.json
@@ -1,0 +1,15 @@
+{
+    "build-deploy": false,
+    "coverage": true,
+    "docs": false,
+    "ui": false,
+    "integration": false,
+    "smoke": false,
+    "smoke-flaky": false,
+    "rpms": false,
+    "debs": false,
+    "oci": false,
+    "build-publish": false,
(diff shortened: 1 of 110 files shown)


opennms-webapp/src/main/java/org/opennms/web/admin/discovery/ActionDiscoveryServlet.java
```diff
@@ -326,6 +326,10 @@
                 LOG.error(me.getMessage());
             }
 
+            if (proxy == null) {
+                throw new IllegalStateException("unable to get event proxy");
```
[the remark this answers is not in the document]
RangerRick: lol, gotta fix my fix

.circleci/build-triggers.override.json
mershad-manesh, line 1: I'm not sure if this PR is ready to be delivered :). If it is, we might want to remove this file

RangerRick: dangit! always forget

christianpape (approved):
LGTM.

PR #6375, opened 2023-07-14
NMS-15780: Admin Password Change Prompt

synqotik:
Initial implementation of a password change prompt when the default admin user logs in with username `admin`, password `admin`.

- Update login page to check for “admin/admin” and redirect to Password Gate page

- Implement Password Gate page which allows user to either change password or skip

- Clicking “skip” will respect navigating to original target page (if it wasn’t just the login page)

- Refactored password change servlets to abstract base class since they contain much of the same functionality

- Fix integration and smoke tests to click “Skip” so they all still work

- Doesn’t include smoke test for this functionality and UI/UX will need review

Target branch is a feature branch. When smoke test and any UX changes have been implemented and merge to feature branch, will open a separate PR to merge to `develop`.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15780

### DIFF
--- a/opennms-webapp/src/main/webapp/account/selfService/passwordGate.jsp
+++ b/opennms-webapp/src/main/webapp/account/selfService/passwordGate.jsp
@@ -1,0 +1,226 @@
+<%--
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2006-2023 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2023 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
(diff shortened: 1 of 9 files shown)


opennms-webapp/src/main/webapp/account/selfService/passwordGate.jsp
```diff
@@ -0,0 +1,226 @@
...
+
+    oldPass.value = '';
+    pass1.value = '';
+    pass2.value = '';
+  }
+</script>
+
+<div class="login-page">
+    <div class="card login-form rounded">
+        <div style="padding-bottom: 36px; padding-top: 60px">
```
scthompsonopen, line 183: Super minor nit, but you're mixing inline styles and classes here, and also in the footer. Not a huge deal normally for a small one off but this looks like a new file and there's so much CSS just above this. :)

synqotik: Thanks for taking a look! Will wait til Ben or Christian review and approve due to the Java and server-side stuff, but wanted a FE dev to take a look as well :)

I will update this to use a CSS class here and possibly `login.jsp` in a subsequent PR.

PR #6463, opened 2023-08-01
NMS-16029: Make metadata expression the default in opennms-datasources.xml

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-16029


RangerRick (approved):
heck yeah

PR #6472, opened 2023-08-01
NMS-15345: clean up/refactor karaf dependencies to make startup more consistent

RangerRick:
The intermittent issues with starting up flow-related code appears to be a race-condition in Karaf startup that can cause it to time out.

This branch does a number of things in an attempt to clean up the startup process:

* move a number of karaf dependencies to their own features in the `features-core.xml` file and change anything using those dependencies to point to the features instead
* define a set of common start levels and retarget some bundles to match those start levels
* integrate (and override) a number of pax-web and jetty improvements to get rid of bloat and speed startup
* clean up bouncycastle dependencies as much as possible
* enable feature validation in the `container/features` project and validate a number of core and minion/sentinel features for proper dependency chains
* create a jetty "noop" extension to work around a [race condition in Jetty](https://github.com/eclipse/jetty.project/pull/10169) triggered by CXF's websocket startup, in case Jetty doesn't make another 9.4.x release
* add `Bundle-SymbolicName` and `Bundle-Version` to all wrapped dependencies, to make the bundle list in karaf easier to understand

With these changes (especially `start-level` definitions and better dependencies) I have been unable to reproduce the flow startup issue, though that doesn't guarantee it's eliminated.

That said, these updates have value anyway and are worth merging to 2023 and up. While it looks like a lot of churn, it is (for the most part) just rearranging things that already existed. The only significant updates is the removal of JASPI bits from Jetty to align with pax-web 7.4 (we don't use them and they caused exceptions in the karaf logs anyway), and an update of versions for a number of commons-* projects and asm.

There is a Horizon 32 version of this branch that I will merge upon approval of this one. It only contains minor changes relating to slightly different dependencies (newts 3 and cassandra, most notably) but is functionally the same.

I won't merge until both are green.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15345


RangerRick:
looks like the h32 version of this is [passing smoke tests](https://app.circleci.com/pipelines/github/OpenNMS/opennms/31417/workflows/e3d8ad40-d2f7-45e7-abfa-f530c316f269) as well (it just needed some fixups related to the cassandra driver 3 -> 4 move), so pretty confident this is in good shape and mergeable \o/

cgorantla (approved):
LGTM as long as we have covered enough with testing!

PR #6506, opened 2023-08-08
NMS-16057: Prevent query details in response to client for invalid Node filterRule

synqotik:
Prevent query details being returned in response to client for invalid `filterRule` V1 Node Rest API call. This is called by the Grafana Plugin.

See https://opennms.atlassian.net/browse/OCI-44.

If an invalid `filterRule` is sent in the query string, for example something starting like `filterRule=\*serviceName`, a `FilterParseException` is thrown by `JdbcFilterDao`. This was not being caught and bubbled up and the query details were being sent back to the client with a 500 error.

Changed to catch these exceptions and return a `400 Bad Request` with a message that does not reveal details. Note that `JdbcFilterDao` already logs a warning, so not doing any logging here.

Targeting `foundation-2023` since this is a security-related issue.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16057


RangerRick (approved):
Seems reasonable to me. The filter code hasn't changed in a long time, we may want to see if this can be backported further.

PR #6517, opened 2023-08-10
NMS-15806: don't just null-check, also look for empties

RangerRick:
I failed to also check for empty arrays when doing my original bounds-checking work for this.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15806


RangerRick:
I will likely cherry-pick these back to 2020 once this is merged, there are some files that don't exist in earlier branches so it's easier to start here and move back than the other way around, additively.

synqotik (approved):
Looks good!

PR #6523, opened 2023-08-11
NMS-16068: Create pinia stores, but do not yet use them

synqotik:
Initial task for moving from Vuex to Pinia, create all the pinia stores, but do not yet wire them up.

Stores will be moved in subsequent PRs, and most likely there will be additional cleanup.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16068

### DIFF
--- a/ui/src/stores/appStore.ts
+++ b/ui/src/stores/appStore.ts
@@ -1,0 +1,14 @@
+import { defineStore } from 'pinia'
+
+export const useAppStore = defineStore('appStore', () => {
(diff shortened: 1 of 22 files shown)


scthompsonopen (approved):
This looks incredible. So much less boilerplate this way (and easier to read/follow). A few of these stores are a bit long in the tooth, but I can easily say the same about some of my own stores so won't even leave it as a real comment. Sometimes they get big.

Great work Scott! LGTM.

PR #6549, opened 2023-08-17
NMS-16090: fix min and max java version

indigo423:
Set the minimum and maximum JDK version to the same range as in Horizon Core.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16090


RangerRick:
oof, good catch; this should be fixed in 2022/2023/H32 also, though (although the range should be 11-12 rather than 11-17)

RangerRick (changes requested):
this one should apply to H32 rather than develop, and there should be a similar PR for 2022/2023 pinning to only 11

PR #6632, opened 2023-09-08
Ranger/guava osgi version cleanup smoke

RangerRick:
Friday cleanup work. :)

This PR does a couple of things:

1. move the weird jest thing we make into `dependencies/` and no longer mask a real project that isn't ours (and fix its guava dependencies)
2. add bundle `<Import-Package>` entries for Guava anywhere it's used, so you're allowed to use equal-to-or-higher versions.

A quick test shows everything still starts up normally, and these fixes will be even _better_ when they come forward to H32 since it will let us move forward on Guava versions without breaking existing things, and Newts will finally be in line with our newer Guava dependency.

### DIFF
--- a/container/extender/pom.xml
+++ b/container/extender/pom.xml
@@ -18,6 +18,11 @@
                 <groupId>org.apache.felix</groupId>
                 <artifactId>maven-bundle-plugin</artifactId>
                 <extensions>true</extensions>
+                <configuration>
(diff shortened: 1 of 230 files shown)


synqotik (approved):
Looks reasonable and syntactically correct :)

christianpape (approved):
LGTM.

PR #6646, opened 2023-09-12
NMS-16028: Allow metadata in snmp-config.xml

christianpape:
JIRA: https://opennms.atlassian.net/browse/NMS-16028

### DIFF
--- a/core/snmp/api/src/main/java/org/opennms/netmgt/snmp/SnmpProfileMapper.java
+++ b/core/snmp/api/src/main/java/org/opennms/netmgt/snmp/SnmpProfileMapper.java
@@ -42,7 +42,11 @@ public interface SnmpProfileMapper {
      * @param oid         OID with which SNMP get needs to be performed.
      * @return snmpAgentConfig, a @{@link SnmpAgentConfig} from the matching profile.
      */
-    CompletableFuture<Optional<SnmpAgentConfig>> getAgentConfigFromProfiles(InetAddress inetAddress, String location, String oid);
+    default CompletableFuture<Optional<SnmpAgentConfig>> getAgentConfigFromProfiles(InetAddress inetAddress, String location, String oid) {
+        return getAgentConfigFromProfiles(inetAddress, location, oid, true);
+    }
+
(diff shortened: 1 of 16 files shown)


core/snmp/api/src/main/java/org/opennms/netmgt/snmp/SnmpProfileMapper.java
```diff
@@ -42,7 +42,11 @@ public interface SnmpProfileMapper {
      * @param oid         OID with which SNMP get needs to be performed.
      * @return snmpAgentConfig, a @{@link SnmpAgentConfig} from the matching profile.
      */
-    CompletableFuture<Optional<SnmpAgentConfig>> getAgentConfigFromProfiles(InetAddress inetAddress, String location, String oid);
+    default CompletableFuture<Optional<SnmpAgentConfig>> getAgentConfigFromProfiles(InetAddress inetAddress, String location, String oid) {
+        return getAgentConfigFromProfiles(inetAddress, location, oid, true);
+    }
```
RangerRick, line 47: such a simple addition to Java but god this makes interfaces so much nicer

RangerRick (approved):
looks good

christianpape:
1. I tested with and without encryption enabled. So, if used, the metadata expression will be stored encrypted and interpolated after decryption.
2. What do you mean by "merge" exactly? The same metadata expressions point to the same community in SCV. Of course if you specify definition with the real community and one with the reference to the SCV you cannot merge these.
3. The feature just uses the global SCV scope, so no data for nodes, interfaces or services are available. 
4. yes, sure

What do you think?

cgorantla:
(1) Good that mixing works but I feel these are two different options for user.
For (2),  we merge two definitions if they are same( by merging two successive IP addresses as range). When I use metadata in snmp-config.xml , Does this still work ?
(3) Should be added to docs.

christianpape:
> (1) Good that mixing works but I feel these are two different options for user.
> For (2), we merge two definitions if they are same( by merging two successive IP addresses as range). When I use metadata in snmp-config.xml , Does this still work ?
> (3) Should be added to docs.

See https://github.com/OpenNMS/opennms/pull/6662 for these changes.

PR #6657, opened 2023-09-13
NMS-15695: Added missing configuration files for Sentinel

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-15695


RangerRick:
@christianpape So are these basically stub configs but the important stuff ends up passing from/to core or something?

christianpape:
> @christianpape So are these basically stub configs but the important stuff ends up passing from/to core or something?

Exactly.

PR #6683, opened 2023-09-18
NMS-16136: Convert node store to pinia

synqotik:
Convert node and also event stores from `vuex` to `pinia` and removed the corresponding `vuex` code. Needed before doing more work on Node Structure.

Had to refactor the `Pagination` control a bit to use `pinia` store methods/properties instead of creating strings for `vuex`. `Pagination:query` prop takes a method that takes `QueryParameters` only; if another param like node id is needed, caller supplies it inside the referenced methods.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16136


synqotik:
Smoke test is failing due to an unrelated issue being fixed elsewhere.

mikewrosey (approved):
lgtm. Nice to remove all those vuex files!

synqotik:
> lgtm. Nice to remove all those vuex files!

hopefully will all be gone soon :)

synqotik:
Merging since smoke test error is not related to this code.

PR #6690, opened 2023-09-20
NMS-16119: Fixed poll command to also display effective parameter values

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-16119


synqotik (approved):
LGTM

PR #6691, opened 2023-09-20
NMS-15096: Add location to snmp info response

fooker:
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15096


Bonrob2 (dismissed):
docs LGTM! Thanks!

christianpape (dismissed):
LGTM.

christianpape (approved):
LGTM.

ProjectJYL:
Would API v1 support `location` and `ttl` with the addition of those elements in snmp-info.xsd? If not, can we make it to support it?

ProjectJYL (commented):
Would it make sense it make a note about GET and PUT request for a node at a location? For example, under an example, 

NOTE: To get the SNMP configuration for a node at a non `Default` location, add filter location to the url such as `http://localhost:8980/opennms/rest/snmpConfig/10.1.1.1?location=Raleigh`

NOTE: To set the SNMP configuration for a node at a non `Default` location, add element `location` to the content body.

fooker:
> Would it make sense it make a note about GET and PUT request for a node at a location? For example, under an example,
> 
> NOTE: To get the SNMP configuration for a node at a non `Default` location, add filter location to the url such as `http://localhost:8980/opennms/rest/snmpConfig/10.1.1.1?location=Raleigh`
> 
> NOTE: To set the SNMP configuration for a node at a non `Default` location, add element `location` to the content body.

IMO, the GET parameter is documented quite well. For POSTing it well leave the field empty and thus match on all nodes with the IP regardless of its location.

> Would API v1 support location and ttl with the addition of those elements in snmp-info.xsd? If not, can we make it to support it?

Yes, it is a non-breaking extension to the v1 API.

ProjectJYL:
[Here](https://github.com/OpenNMS/opennms/blob/e9984a44ae91297ace7b9d3061e2ba4255cf9271/docs/modules/development/pages/rest/snmp_configuration.adoc?plain=1#L47) it says API V1 only supports a few attributes from snmp-info.xsd. If it's supported with this PR then the docs should be updated?  

I think without specifying the location in the PUT request, it shouldn't update all nodes regardless of location. It should only update node in the `Default`. Otherwise, it's easy to make mistake and update all nodes in multiple locations in a single request. There are some use cases where duplicated IP with different locations to distinguish them part.

fooker:
> [Here](https://github.com/OpenNMS/opennms/blob/e9984a44ae91297ace7b9d3061e2ba4255cf9271/docs/modules/development/pages/rest/snmp_configuration.adoc?plain=1#L47) it says API V1 only supports a few attributes from snmp-info.xsd. If it's supported with this PR then the docs should be updated?

This is super confusing as there are multiple version involved:
- The SNMP version
- The version of the SNMP Info document
- The REST API Version

It seems like the document version is an ancient artifact from before the REST API was versioned. Nowadays we will only server what is described as version 2. The REST endpoint is still at version 1, as there is no `api/v2` SNMP config endpoint.

> I think without specifying the location in the PUT request, it shouldn't update all nodes regardless of location. It should only update node in the `Default`. Otherwise, it's easy to make mistake and update all nodes in multiple locations in a single request. There are some use cases where duplicated IP with different locations to distinguish them part.

The way PUT currently works directly reflects how the snmp-config mechanism works. It will append a definition with the `location` field left blank to the config. On a GET, the config for a the specific host is resolved and returned.

THB, the missing location may not have been an oversight at all. Replying it from the request is in the same way wrong as not having in the response. Same is true for the host address. I would expect the API to return the defined config as-is for the host I have specified in the request.

PR #6699, opened 2023-09-21
NMS-16126: Node structure, add management IP or best IP address

synqotik:
Add the management IP address, or at least the "best" IP address, to the node table.

Any time the node query changes (sorting, filters, search), we also launch an IP Address query for the returned nodes and update a map.

For each node, we try to use the IP interface where snmpPrimary is 'P', or else at least isManaged, or else the first or onlly IP. Tooltip displays managed and SNMP Primary status, clicking on link brings you to the 'classic' IP Interface page.

Currently cannot sort by IP Address, but you can configure whether it displays and in which column order. Also displays in the Node Details dialog, and it has a link to the legacy IP Interface page.

Updated default pagination to 20 nodes per page.

Updated Search to have an implicit start and end wildcard, e.g. typing `host` will do a search on `*host*` and return nodes with labels like `localhost`, `host123`, 'my_host_name`, etc.

node-table-ipaddress

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16126

### DIFF
--- a/ui/src/components/Nodes/ManagementIPTooltipCell.vue
+++ b/ui/src/components/Nodes/ManagementIPTooltipCell.vue
@@ -1,0 +1,45 @@
+<template>
+  <td :class="ipInfo.label ? 'pointer' : ''">
+    <FeatherTooltip
+      :title="tooltipTitle"
+      :alignment="PointerAlignment.left"
+      :placement="PopoverPlacement.top"
+      v-slot="{ attrs, on }">
+        <a v-bind="attrs" v-on="on" :href="computeNodeIpInterfaceLink(node.id, ipInfo.label)">
+          {{ ipInfo.label }}
+        </a>
+    </FeatherTooltip>
+  </td>
+</template>
+
+<script setup lang="ts">
+import { IpInterface, Node } from '@/types'
+import { FeatherTooltip, PointerAlignment, PopoverPlacement } from '@featherds/tooltip'
+import { PropType } from 'vue'
+import { IpInterfaceInfo, getBestIpInterfaceForNode } from './utils'
+
+const props = defineProps({
+  computeNodeIpInterfaceLink: {
+    required: true,
+    type: Function as PropType<(nodeId: number | string, ipAddress: string) => string>
+  },
+  node: {
+    required: true,
+    type: Object as PropType<Node>
+  },
+  nodeToIpInterfaceMap: {
+    required: true,
+    type: Object as PropType<Map<string, IpInterface[]>>
+  }
+})
+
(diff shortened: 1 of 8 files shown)


indigo423:
I have tested it with 16k Nodes and it's really a big improvement speed-wise, the Node list takes 12 sec. to load vs. 5 sec

Screenshot 2023-09-22 at 12 50 24

Screenshot 2023-09-22 at 12 42 23

Screenshot 2023-09-22 at 12 55 15

What I don't understand is the "MP" and "MN" behind the IP address. If that is really the way we want to tell the user if this is the "Management Type" of the IP address, like "Primary", "Secondary" and "Non-Eligible", then we should make that a column and write that out instead of introducing a new abbreviation no one will understand :)

Screenshot 2023-09-22 at 12 58 28

I would also make 20 or 50 the default instead of 10, it's like looking into a time stamp with just 10 nodes in the list and you have a lot of whitespace on the screen.

Screenshot 2023-09-22 at 12 50 24

Great work and this will be a big usability improvement!

indigo423 (commented):
Added some usability comments. Tested the functionality and also the CSV and JSON download.

As a suggestion I would rename it from "Nodes (Preview)" to something like "Node Browser".  We can introduce this as a new feature and tell users we want to replace our  "Node List" with it in H34.

indigo423:
What I realized when I used the search, it only matches from the beginning for example if you type "local" in the search you will see the node "localhost" but when you type "host" in the search you won't.

synqotik:
> I have tested it with 16k Nodes and it's really a big improvement speed-wise, the Node list takes 12 sec. to load vs. 5 sec
> 
> Great work and this will be a big usability improvement!

Thanks!

Re IP Address, I'm not sure the best way to represent this. It's supposed to be the "management IP address", so I'm trying to pick the "best" IP interface on that node. I think for now I can get rid of `(MP)` etc. and have a tooltip. If we decide to make the management type a column, that can be a separate story (I can add that). Also should be reviewed that my logic is doing the right thing.

I'll bump the default pagination to 20 per page.

I'll hold off renaming the menu item until @pbrane has a chance to look; easy to fix. And yes the idea is to replace the "legacy" Node List by H34. Need work to handle for example query strings and links that go to the old page (tickets already in Jira).

Re search, you'll need to use wildcards like `*`. So to search for `host`, use `*host` or `*host*`. We probably need some instructions somewhere on that.

synqotik:
Updates:

- default to 20 nodes per page
- IP Address column displays IP address with a tooltip to show managed / SNMP primary status
- search box now has implicit wildcards for start and end

christianpape (approved):
LGTM.

PR #6703, opened 2023-09-25
NMS-16154: Convert auth usageStats fileEditor etc stores to pinia

synqotik:
Convert more `vuex` stores to `pinia` including:

- auth
- configuration
- plugin
- usageStatisticsSharing
- ipInterfaces
- fileEditor

Most `pinia` implementation was already done, this is mostly wireup and deleting the `vuex` stores, except for more work that was needed for `fileEditor`.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16154


ui/src/components/FileEditor/Editor.vue
```diff
@@ -48,9 +48,9 @@ const theme = computed(() => {
   return 'xcode'
 })
 
-const selectedFileName = computed(() => store.state.fileEditorModule.selectedFileName)
-const isHelpOpen = computed(() => store.state.fileEditorModule.isHelpOpen)
-const fileString = computed(() => store.state.fileEditorModule.file)
+const selectedFileName = computed(() => fileEditorStore.selectedFileName)
+const isHelpOpen = computed(() => fileEditorStore.isHelpOpen)
```
mikewrosey, line 52: One other shortcut is that we should be able use these right in the template without computing them. Like `<div :class="{ 'help-open': fileEditorStore.isHelpOpen...` But again, this can be optimized after.

mikewrosey (approved):
Looks great!

ui/src/components/FileEditor/Editor.vue
synqotik: Good idea, I'll try to apply both of these suggestions in the future and eventually clean up existing code to apply them.

PR #6708, opened 2023-09-27
NMS-15774: Brief overview of the metadata changes

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-15774


Bonrob2 (approved):
LGTM! Thanks!

christianpape:
> Does this change only allow access to the SCV scope for these additional files, or to any metadata context?

In the case of monitors/collectors the node/interface/service scope is also available but in the case of general configuration only the SCV scope is available.

PR #6712, opened 2023-09-28
NMS-16156: Convert remaining items to pinia and remove vuex

synqotik:
Convert the remaining items from `vuex` to `pinia` and completely remove `vuex` from the project.

There is one test in `scv.test.ts` that wasn't working even though actual functionality works, I've skipped it for now but will go back and fix it (ensuring Add Credentials button is disabled when password is masked).

Also did some refactoring, for example in `LeafletMap`, to fix some issues and make types clearer. There seems to be a bug in Search, but will address that separately.

Apologies for the large number of files, though over a third are just deletions.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16156


ui/src/main.ts
```diff
@@ -75,7 +73,6 @@ library.add(...icons);
 // let plugins use state mngmnt / router
 (window as any).Vue = Vue;
 (window as any).Pinia = Pinia;
-(window as any).Vuex = Vuex;
```
mikewrosey, line 78: Just double checking - do we have any plugins that use Vuex? I believe Alec only used Pinia. Not sure why we needed this line to begin with.

mikewrosey (approved):
Great job!

ui/src/main.ts
synqotik: I did think of this and checked Alec, it uses pinia. It's possible it could be an issue with other plugins, but probably best if they don't rely on this and instead handle their own state management.

PR #6724, opened 2023-10-06
NMS-16159: Fixed SENTINEL_HOME in scripts

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-16159


RangerRick (changes requested):
looks fine (although the tabs don't match... grin) but this should probably apply to `foundation-2022` as well, since that's where the script got introduced

PR #6773, opened 2023-10-13
NMS-16116: Update node structure page to accept query strings

synqotik:
Updates the Node Structure page to accept query strings and convert them to query filters.

When certain query strings are included in the node Vue page link, they will be converted and added to the `nodeStructure` `pinia` store and query relaunched, and query string removed from the URL (so it's doesn't get re-processed). Query strings will come from legacy pages that currently link to the legacy node list page.

Refactored how the search works, basically search updates are saved in the `pinia` store and the `watch` in `NodesTable.vue` watches for those changes. Otherwise updates were coming from multiple places and clobbering each other.

Refactored `utils.ts` into separate hooks, it was getting unwieldy.

Still more work to be done on this but will follow up in a subsequent PR. Targeting `develop` as it's all on the new Vue page.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16116


mikewrosey (approved):
Looks great!

PR #6790, opened 2023-10-18
NMS-16116: Node structure add additional queries for snmp, flows, etc.

synqotik:
NMS-16116: Node structure add additional queries for flows, ipaddress, monitoringLocations, snmp parameters.

Parses these queries out from query string parameters, and also the user can use the UI to make these queries.

Added an `ExtendedSearchPanel` for making IP address and certain SNMP queries. For now can only search on one of these at a time, eventually can add more combinations.

As a note, when user enters a URL with a query string, or is redirected from another page, the app will "ingest" the query string filter parameters and add to the `pinia` store and update the UI, then clear the query string from the URL. Otherwise there could be issues trying to keep query string parameters and `pinia` store / UI elements all in sync. 

Also did some refactoring to break up large files, and fixed issue where IP interfaces searches were done with invalid IP addresses.

Screenshot 2023-10-18 at 18 16 50

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16116

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -58,6 +58,7 @@
     "date-fns": "^2.30.0",
     "date-fns-tz": "^2.0.0",
     "ip-regex": "^5.0.0",
+    "is-ip": "^5.0.1",
     "is-valid-domain": "^0.1.6",
     "leaflet": "^1.8.0",
     "leaflet.markercluster": "^1.5.3",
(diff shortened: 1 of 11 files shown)


RangerRick (approved):
a lot, but seems pretty reasonable to me

PR #6791, opened 2023-10-19
NMS-15790: Transport bootstrap.jsp args in context

fooker:
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15790

### Review remarks
The first and last commit is worth reviewing while the second commit is auto-generated using this script: https://gist.github.com/fooker/3097e401069e9c6db056efe98133c146

### DIFF
--- a/opennms-webapp/src/main/java/org/opennms/web/utils/Bootstrap.java
+++ b/opennms-webapp/src/main/java/org/opennms/web/utils/Bootstrap.java
@@ -1,0 +1,142 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2023 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2023 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License,
+ * or (at your option) any later version.
+ *
+ * OpenNMS(R) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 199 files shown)


opennms-webapp/src/main/java/org/opennms/web/utils/Bootstrap.java
```diff
@@ -0,0 +1,142 @@
...
+    private final PageContext pageContext;
+
+    private final List<String> headTitles = Lists.newArrayList();
+
+    private final List<Breadcrumb> breadcrumbs = Lists.newArrayList();
+
+    private String ngApp;
+    private String scrollSpy;
+
+    private Set<String> flags = Sets.newHashSet();
```
RangerRick, line 54: Guava has it's uses, but it also creates a lot of dependency headaches. Please don't use it if all you're calling is no argument `newBlahBlah()` things.

It's [basically](https://github.com/google/guava/blob/v20.0/guava/src/com/google/common/collect/Lists.java#L86) the [same](https://github.com/google/guava/blob/v20.0/guava/src/com/google/common/collect/Sets.java#L163) amount of typing. :)

opennms-webapp/src/main/webapp/includes/bootstrap.jsp
```diff
@@ -256,18 +237,25 @@
...
+      --%></c:choose><%--
+      --%><c:choose><%--
+        --%><c:when test="${breadcrumb.link != null}"><%--
+	  --%><a href="${breadcrumb.link}"><c:out value="${breadcrumb.title}" escapeXml="false"/></a><%--
+	--%></c:when><%--
+	--%><c:otherwise><%--
+	  --%><c:out value="${breadcrumb.title}" escapeXml="false"/><%--
+        --%></c:otherwise><%--
+      --%></c:choose><%--
+      --%></li>
```
RangerRick, line 256: lol, I get it, but also, ow my head 😅

RangerRick (approved):
other than the guava thing, lgtm

synqotik (dismissed):
Nice Python script!

RangerRick (approved):

PR #6805, opened 2023-10-25
NMS-15667: Interpolate Mate in threshold label and resource-filters

fooker:
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15667


RangerRick (approved):
always nice to see more deletes than adds 👍

PR #6808, opened 2023-10-25
NMS-15400: Allow multiple tile providers in Geographical Map

synqotik:
Allow multiple "tile providers" to be configured in the Vue Geographical Map, so users can add their own, for example if they have a local/private map server.

There are 2 hard-coded ones in the Vue code. An additional one can be configured in `opennms/etc/opennms.properties` under the `gwt.openlayers.url` and `gwt.openlayers.options.attribution` properties.

The additional provider, if configured, is included in the Menu Rest API, `userTileProviders` array. It will display in the tile layer dropdown. Currently name is hard-coded to `User-Defined`.

Screenshot 2023-10-25 at 13 51 50

Will open a separate Documentation ticket as any added tile provider needs to be added to `web.xml` `Content-Security-Policy` `img-src` section, or else it will be blocked by the browser.

Also added the Open TopoMap `opentopomap.org` URL to CSP, it wasn't added by default and was not usable.

This needs to be backported to `foundation-2023`, will do in separate PR as code will be a bit different due to dependencies, `pinia` and other issues.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15400

### DIFF
--- a/opennms-webapp-rest/src/main/java/org/opennms/web/rest/support/menu/MenuProvider.java
+++ b/opennms-webapp-rest/src/main/java/org/opennms/web/rest/support/menu/MenuProvider.java
@@ -28,8 +28,6 @@
 
 package org.opennms.web.rest.support.menu;
 
-import com.google.common.base.Strings;
-import com.google.common.collect.ImmutableSet;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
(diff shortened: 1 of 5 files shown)


mikewrosey (approved):
UI changes look good

PR #6838, opened 2023-11-07
NMS-16131: Added ed25519 algorithm

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-16131

### DIFF
--- a/container/karaf/src/main/filtered-resources/etc/custom.properties
+++ b/container/karaf/src/main/filtered-resources/etc/custom.properties
@@ -39,6 +39,7 @@ org.osgi.framework.system.packages.extra=org.apache.karaf.branding,\
         sun.net.dns,\
         sun.net.spi.nameservice,\
         sun.reflect,\
+        sun.security.x509,\
         javax.jms;version=1.1.0,\
         javax.mail;version=1.6.7,\
         javax.mail.internet;version=1.6.7,\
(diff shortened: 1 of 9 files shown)


RangerRick:
Hrm, looks like something going weird with jsch in the smoke tests? Could this be related to your changes? At first glance doesn't seem like it should be, but...

Bonrob2 (commented):
Got auto tagged for review, but I don't see any docs, so deferring to others.

container/karaf/src/main/filtered-resources/etc/custom.properties
```diff
@@ -39,6 +39,7 @@ org.osgi.framework.system.packages.extra=org.apache.karaf.branding,\
         sun.net.dns,\
         sun.net.spi.nameservice,\
         sun.reflect,\
+        sun.security.x509,\
```
RangerRick, line 42: You will probably have to do this in the minion and sentinel versions of `custom.properties` as well. (`features/container/sentinel/src/main/filtered-resources/etc/custom.properties` and `features/container/minion/src/main/filtered-resources/etc/custom.properties`)

RangerRick: Someday we should probably rework this stuff to be a template so we don't have to keep doing all the same things in 3 different places...

christianpape:
Yay! CircleCI is happy!

RangerRick (approved):

christianpape:
Rebased to foundation-2023.

container/karaf/src/main/filtered-resources/etc/custom.properties
christianpape: done

RangerRick (approved):
lookin' good!

PR #6844, opened 2023-11-07
NMS-16197: Node structure, more search types, fix switch

synqotik:
Node structure panel, add search types for foreign source and id as well as "system" attributes (`sysContact`, `sysLocation`, etc.). Needed to refactor a bit to clean this up.

'system' attributes use a wildcard search. SNMP and foreign source/id attributes are exact search (due to current Rest API implementation).

Used the Feather `SwitchRender` control for Match Any / Match All for Category searches.

Added more fields to the Node Info dialog.

This is probably the last set of features for the Node Structure page for the upcoming Horizon 33 release.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16197

### DIFF
--- a/ui/src/components/Nodes/ExtendedSearchPanel.vue
+++ b/ui/src/components/Nodes/ExtendedSearchPanel.vue
@@ -5,7 +5,7 @@
       :options="searchOptions"
       :textProp="'title'"
       v-model="currentSelection"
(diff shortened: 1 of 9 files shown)


mikewrosey (approved):
LGTM

PR #6848, opened 2023-11-08
NMS-15064 and NMS-16218: upgrade Camel and ActiveMQ

RangerRick:
This PR updates to Camel 2.21.5 and ActiveMQ 5.15.16.

It contains a number of changes:

1. Camel's Karaf feature file is a _mess_. Rather than trying to overcome its deficiencies, I do some terrible magic in `container/features/pom.xml` to clean it up, and attach it to our own project under the `-camel.xml` classifier.
2. There were some terrible race-conditions/ordering problems in Karaf startup that just happened to work previously and now don't. I added a bunch of features to the Karaf validation to make sure dependencies were square, and updated the features files for sentinel and base until they passed.
3. There are a number of weird conflicting issues with JMS specs, some things want an older release of it released as a beta by the Geronimo project, some want the now-official `javax.jms:jms-api` dependency. I created excludes and cleanups to force us to the latter.
4. I had to force our Spring 4.2 to load for a number of activemq/camel feature dependencies or else it tried to load 4.2 _and_ 4.3 which caused a number of wacky issues.
5. I updated all our camel files to match the newer schema. Some contexts needed `registerEndpointIdsFromRoute="true"` to retain backwards-compatibility with Camel 2.19 behavior. To move on to later Camel versions, we may need to audit our routes for missing ID references. (But we need Spring 4.3 first anyway for that to happen.)
   > Camel with Spring XML routes do not register endpoints in the Spring registry from Camel routes where \<from\> or \<to\> have endpoints assigned with an explicit id attribute. The option `registerEndpointIdsFromRoute` can be set to `true` on \<camelContext\> for backward compatibility. But this registration is deprecated and instead you should use \<endpoint\> to register Camel endpoints with id’s in Spring registry.
6. Camel's blueprint test infrastructure has a bug where it doesn't register its state service properly. Later versions of this code just don't even use it anymore, but for now I put a patched version in our test code to make it initialize properly.
7. It is not obvious, but the Camel blueprint test base class has a method called `isUseAdviceWith()` that determines whether the context should get started automatically or not. Newer versions default to `true` (which does _not_ start the context for you), but our tests assumed the context would be loaded automatically.

Once all that was done, bumping to the latest ActiveMQ 5.15.x was simple. (It had a requirement of Camel 2.20 or higher.)

This does not get us out of the woods on Camel, but resolves a few things, and sets us up to move forward farther once we can get Spring updated more.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-15064
* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16218


synqotik (approved):
I don't know much about this, but it all seems reasonable, plus the build and tests are passing :)

PR #6875, opened 2023-11-15
NMS-16242: Install node-gyp for CircleCI build-ui

synqotik:
CircleCI `build-ui` keeps failing due to not having `node-gyp` installed, I believe it's needed for `yarn install`. This ensures that it is already installed in the build environment.

Error message in NMS ticket.

There could be a different or better way to do this, but this does work.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16242


mershad-manesh (approved):
It looks good to me

PR #6887, opened 2023-11-20
NMS-16226: Clear Karaf's data directory on upgrade

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-16226

The deletion of the files in the data directory is realized by an OnmsUpgrade task since in the script itself there is no chance to decide whether an upgrade is run or just help is displayed.

### DIFF
--- a/core/upgrade/src/main/java/org/opennms/upgrade/implementations/ClearKarafCacheMigratorOffline.java
+++ b/core/upgrade/src/main/java/org/opennms/upgrade/implementations/ClearKarafCacheMigratorOffline.java
@@ -1,8 +1,8 @@
 /*******************************************************************************
  * This file is part of OpenNMS(R).
  *
- * Copyright (C) 2013-2014 The OpenNMS Group, Inc.
- * OpenNMS(R) is Copyright (C) 1999-2014 The OpenNMS Group, Inc.
+ * Copyright (C) 2023 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2023 The OpenNMS Group, Inc.
  *
  * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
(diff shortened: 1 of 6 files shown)


core/upgrade/src/main/java/org/opennms/upgrade/implementations/ClearKarafCacheMigratorOffline.java
```diff
@@ -0,0 +1,92 @@
...
+            Files.walk(this.opennmsDataPath)
+                    .filter((path) -> !path.equals(historyFilePath) && !path.equals(this.opennmsDataPath))
+                    .sorted(Comparator.reverseOrder())
+                    .map(Path::toFile)
+                    .forEach(File::delete);
+        } catch (IOException e) {
+            throw new OnmsUpgradeException(String.format("Error pruning Karaf's data directory '%s'.", this.opennmsDataPath), e);
+        }
+        throw new OnmsUpgradeException(String.format("Karaf's data directory '%s' pruned", this.opennmsDataPath));
+    }
```
RangerRick, line 74: shouldn't `preExecute()` _only_ check if it should run (which in this plugin's case, is basically always true) and the deletes should happen in `execute()` below?

RangerRick (commented):
Also, I forget the details of how upgrade scripts work, will this run every time, even after the "upgrade" task has stored it as finished?

core/upgrade/src/main/java/org/opennms/upgrade/implementations/ClearKarafCacheMigratorOffline.java
christianpape: Yes. I'll rework this, since this isn't the way it should be used.

RangerRick (approved):
looks great

PR #6927, opened 2023-12-04
NMS-16167: Only display data of matching Collectd-package

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-16167

### DIFF
--- a/opennms-webapp/src/main/webapp/WEB-INF/jsp/element/service.jsp
+++ b/opennms-webapp/src/main/webapp/WEB-INF/jsp/element/service.jsp
@@ -30,12 +30,16 @@
 --%>
 
 <%@page language="java"
-	contentType="text/html"
-	session="true"
-	import="
-        java.util.*,
-        org.opennms.netmgt.config.CollectdConfigFactory,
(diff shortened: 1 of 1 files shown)


synqotik (approved):
Looks reasonable to me.

christianpape:
> While the collection package is fixed, it looks like there was a reversion on the service parameters sections. (also may want to touch copyright dates at the top of the file) This patch: image
> 
> From a stock 32.0.5: image

I rebased to a release-31.x. Now this should look similar.

mmahacek (dismissed):
Works and looks great! (tested on a 32.0.5 instance)

mmahacek:
Can we update 2023 with the same version of this file too?  I just tested the `service.jsp` on a 2023.1.9 instance, and it works for displaying the correct collectd package as well as collectd service parameters.

christianpape:
> Can we update 2023 with the same version of this file too? I just tested the `service.jsp` on a 2023.1.9 instance, and it works for displaying the correct collectd package as well as collectd service parameters.

done.

synqotik (approved):

PR #6947, opened 2023-12-11
Karaf startup cleanups

RangerRick:
This was triggered by recent failures in the `release-32.x` and `release-33.x` branches, but is a more general cleanup/refactoring to base my 32+ fixes on.  It does a number of things:

1. fix a couple of silly bugs in karaf feature configs
2. normalize `custom.properties` to add the same `sun.*` extras to our exports everywhere
3. add a few extra features to validation to avoid future headaches
4. **update the karaf extender to also accept `features.uri` and `features.boot` files in the `system/` repository, not just in custom `repositories/*` repositories**
5. use the same `org.apache.karaf.features.cfg` file between minion and sentinel (rather than having slightly different ones) and use `features.uri` to locate the Sentinel repo
6. move the `scv-jceks-impl` to a features boot file so it gets properly excluded by confd changes


synqotik (approved):
Seems reasonable.

PR #6948, opened 2023-12-11
NMS-16253: Fix sampling frequency and origin for gnmi

cgorantla:
Previously sampling frequency was considered to be in milli seconds ( as per jti. reference)
 gNMI by default uses ns ( nano seconds).
Also set default origin if none specified.

### All Contributors

### Contribution Checklist

### What's Next?

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16253

### DIFF
--- a/features/openconfig/telemetry-client/src/main/java/org/opennms/features/openconfig/telemetry/OpenConfigClientImpl.java
+++ b/features/openconfig/telemetry-client/src/main/java/org/opennms/features/openconfig/telemetry/OpenConfigClientImpl.java
@@ -29,7 +29,22 @@
 package org.opennms.features.openconfig.telemetry;
 
 
-import static io.grpc.ConnectivityState.READY;
+import com.google.common.base.Splitter;
+import com.google.common.base.Strings;
+import io.grpc.ConnectivityState;
+import io.grpc.ManagedChannel;
+import io.grpc.stub.StreamObserver;
+import org.opennms.core.grpc.common.GrpcClientBuilder;
+import org.opennms.core.utils.InetAddressUtils;
+import org.opennms.core.utils.StringUtils;
+import org.opennms.features.openconfig.api.OpenConfigClient;
+import org.opennms.features.openconfig.proto.gnmi.Gnmi;
+import org.opennms.features.openconfig.proto.gnmi.gNMIGrpc;
+import org.opennms.features.openconfig.proto.jti.OpenConfigTelemetryGrpc;
+import org.opennms.features.openconfig.proto.jti.Telemetry;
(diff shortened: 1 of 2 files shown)


docs/modules/reference/pages/telemetryd/protocols/openconfig.adoc
```diff
@@ -115,7 +122,7 @@ Parameters are passed to the connector and are interpolated for node/interface a
 | none
 
 | frequency
-| Frequency at which OpenConfig data can be streamed.
+| Frequency at which OpenConfig data can be streamed. This is parsed as milliseconds(ms) for jti and nanoseconds(ns) for gnmi
```
fooker, line 125: Can we unify this? Always make it ms or ns and let the impl below do the conversions?

pbrane: Historically speaking, I would have leaned toward using milliseconds as most OpenNMS configs are in milliseconds, but in this case that would remove granularity from the user's control. If we could somehow make it obvious to the user that this option is in nanoseconds, I would lean toward giving them this control. Not having used this feature, I'm ambivalent to the decision.

cgorantla: we are parsing it and directly setting sampling frequency here. Not doing any conversions.  I would prefer to keep this way.   If some implementations doesn't stick to ns or ms, user has flexibility to update.

RangerRick: make it accept N.nnn ? :D

RangerRick (approved):
Still don't like it being interpreted differently; is there a case where anyone would ever need ns resolution for a config like this? If not I still think parsing and making it always ms would be better, even if it's more work for us.

But I don't feel that so strongly I'll stop you. :)

cgorantla:
> Still don't like it being interpreted differently; is there a case where anyone would ever need ns resolution for a config like this? If not I still think parsing and making it always ms would be better, even if it's more work for us.
> 
> But I don't feel that so strongly I'll stop you. :)

Yeah, that seems forced by proto definition which is different for both protocols.

gnmi : 
```
    uint64 sample_interval = 3;       // ns between samples in SAMPLE mode.
```

jti :
```
    // Time in ms between collection and transmission of the
    // specified data to the collector platform. The target device
    // will sample the corresponding data (e.g,. a counter) and
    // immediately send to the collector destination.
    //
    // If sample-frequency is set to 0, then the network device
    // must emit an update upon every datum change.
    uint32 sample_frequency                                 = 5;
```

PR #6991, opened 2024-01-08
normalize all Apache Commons usage

RangerRick:
This PR makes all apache commons packages use a top-level versioned entry in dependency management, and also updates the Karaf container(s) to use an override file that enforces using that version (or newer) in the OSGi side as well.


mershad-manesh (approved):
Changes look good to me

PR #7021, opened 2024-01-19
NMS-15850: Changed timeline URLs to use serviceId instead of serviceName

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-15850


RangerRick (approved):
teeeeeeechnically a "public" API, but I'm sure no user is consuming the timeline :)

PR #7037, opened 2024-01-24
enables async snmp interface poller on init

WilssoN93:
Enables Multi threaded scheduler for SnmpPoller on init. Also added SnmpPoller threads to jmx metrics

We previously had issues where it took about over 40 minutes to start OpenNMS since it is scanning all our interfaces at startup and we have ~3 Mil snmp interfaces. When disabling SnmpPoller service the startup takes ~5 minutes.

This PR would make it possible for the initial scan to utilize SnmpPoller threads for initial scan phase.


mmahacek:
This is neat that you added jmx metrics on the snmppoller daemon. The default configs should be updated to collect the metrics and graph them.

RangerRick:
Yeah, this is super promising. Looks like there's a failure in the tests too, might need to somehow wait on execution to complete now that it's happening asynchronously?

```
[ERROR] org.opennms.netmgt.snmpinterfacepoller.SnmpPollerIT.testIgnoredInterfaces -- Time elapsed: 0.788 s <<< FAILURE!
org.junit.ComparisonFailure: expected:<[P]> but was:<[I]>
        at org.junit.Assert.assertEquals(Assert.java:117)
        at org.junit.Assert.assertEquals(Assert.java:146)
        at org.opennms.netmgt.snmpinterfacepoller.SnmpPollerIT.testIgnoredInterfaces(SnmpPollerIT.java:109)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
```

opennms-services/src/main/java/org/opennms/netmgt/snmpinterfacepoller/SnmpPoller.java
```diff
@@ -240,10 +242,21 @@ protected void scheduleNewSnmpInterface(String ipaddr) {
...
-    	for (OnmsIpInterface iface : getNetwork().getContext().getPollableNodes()) {
-            schedulePollableInterface(iface);    		
-    	}
+        ExecutorService executor = getExecutorService();
+        for (OnmsIpInterface iface : getNetwork().getContext().getPollableNodes()) {
+            executor.execute(() -> schedulePollableInterface(iface));
+        }
+        try {
+            if (executor.awaitTermination(60, TimeUnit.MINUTES)){
+                LOG.info("All snmp interfaces were scheduled successfully!");
```
RangerRick, line 251: Don't executors only terminate if you terminate them, or shut down the JVM?  I suspect the poller `LegacyScheduler` will just keep on trucking until shutdown and this will never fire...

If you want to know if everything that was scheduled actually completed, you will probably have to wait on all its futures. I guess that kind of defeats the point of trying to do them asynchronously though, lol, unless you make a separate thread that watches for futures to complete and logs and/or handles errors.

Regardless of that, 60 minutes is a ridiculously long amount of time if these things are getting scheduled in parallel.

mmahacek:
JMX collection would be updated in the `opennms-base-assembly/src/main/filtered/etc/jmx-datacollection-config.xml` file.  Based on what I see, I _think_ you're probably looking at something like this:
```xml
         <!-- SnmpInterface Poller stats -->
         <mbean name="OpenNMS SnmpPoller" objectname="OpenNMS:Name=SnmpPoller">
            <attrib name="NumPolls" alias="ONMSIPCount" type="counter"/>
            <attrib name="NumPoolThreads" alias="ONMSSIPPoolThrd" type="gauge"/>
            <attrib name="CorePoolThreads" alias="ONMSSIPPPoolCore" type="gauge"/>
            <attrib name="MaxPoolThreads" alias="ONMSSIPPPoolMax" type="gauge"/>
            <attrib name="PeakPoolThreads" alias="ONMSSIPPPoolPeak" type="gauge"/>
            <attrib name="ActiveThreads" alias="ONMSSIPPThreadAct" type="gauge"/>
            <attrib name="TasksTotal" alias="ONMSSIPPTasksTot" type="counter"/>
            <attrib name="TasksCompleted" alias="ONMSSIPPTasksCpt" type="counter"/>
            <attrib name="TaskQueuePendingCount" alias="ONMSSIPTskQPCnt" type="gauge"/>
            <attrib name="TaskQueueRemainingCapacity" alias="ONMSSIPTskQRCap" type="gauge"/>
         </mbean>
```
You'd also need graphs to go along with this, which I'm guessing would be something like this new file. Rename to remove the txt extension.
[opennms-snmppoller-graph.properties.txt](https://github.com/OpenNMS/opennms/files/14056437/opennms-snmppoller-graph.properties.txt)

opennms-services/src/main/java/org/opennms/netmgt/snmpinterfacepoller/SnmpPoller.java
mmahacek: Especially if using the default 5 minute interval.

WilssoN93: > Don't executors only terminate if you terminate them, or shut down the JVM? I suspect the poller `LegacyScheduler` will just keep on trucking until shutdown and this will never fire...
> 
> If you want to know if everything that was scheduled actually completed, you will probably have to wait on all its futures. I guess that kind of defeats the point of trying to do them asynchronously though, lol, unless you make a separate thread that watches for futures to complete and logs and/or handles errors.
> 
> Regardless of that, 60 minutes is a ridiculously long amount of time if these things are getting scheduled in parallel.

Yeah i forgot to add the .shutdown(), When testing a small local project it works when adding the .shutdown() before awaitTermination :)
What would be a good timeout value here? I just put a high value to not accidentally abort the scheduling if it takes along time.

WilssoN93:
> JMX collection would be updated in the `opennms-base-assembly/src/main/filtered/etc/jmx-datacollection-config.xml` file. Based on what I see, I _think_ you're probably looking at something like this:
> 
> ```
>          
>          <mbean name="OpenNMS SnmpPoller" objectname="OpenNMS:Name=SnmpPoller">
>             <attrib name="NumPolls" alias="ONMSIPCount" type="counter"/>
>             <attrib name="NumPoolThreads" alias="ONMSSIPPoolThrd" type="gauge"/>
>             <attrib name="CorePoolThreads" alias="ONMSSIPPPoolCore" type="gauge"/>
>             <attrib name="MaxPoolThreads" alias="ONMSSIPPPoolMax" type="gauge"/>
>             <attrib name="PeakPoolThreads" alias="ONMSSIPPPoolPeak" type="gauge"/>
>             <attrib name="ActiveThreads" alias="ONMSSIPPThreadAct" type="gauge"/>
>             <attrib name="TasksTotal" alias="ONMSSIPPTasksTot" type="counter"/>
>             <attrib name="TasksCompleted" alias="ONMSSIPPTasksCpt" type="counter"/>
>             <attrib name="TaskQueuePendingCount" alias="ONMSSIPTskQPCnt" type="gauge"/>
>             <attrib name="TaskQueueRemainingCapacity" alias="ONMSSIPTskQRCap" type="gauge"/>
>          </mbean>
> ```
> 
> You'd also need graphs to go along with this, which I'm guessing would be something like this new file. Rename to remove the txt extension. [opennms-snmppoller-graph.properties.txt](https://github.com/OpenNMS/opennms/files/14056437/opennms-snmppoller-graph.properties.txt)

Okey thanks i will add this aswell :)

opennms-services/src/main/java/org/opennms/netmgt/snmpinterfacepoller/SnmpPoller.java
RangerRick: Right, but what I'm getting at is that I believe the scheduler injected into your poller is shared by many things, it comes from Pollerd. I think you would be shutting down the entire Pollerd scheduler as soon as init completes...

RangerRick: I think instead you just need to keep track of the futures you made, and wait for them to finish. Don't try to control the scheduler, it's not `SnmpPoller`'s place to manipulate it.

(This really is exposing a bad API, the monitor instead should be asking Pollerd to schedule something instead of being handed a complete executor.)

opennms-services/src/main/java/org/opennms/netmgt/snmpinterfacepoller/SnmpPoller.java
```diff
@@ -240,10 +242,22 @@ protected void scheduleNewSnmpInterface(String ipaddr) {
...
     protected void scheduleExistingSnmpInterface() {
-        
-    	for (OnmsIpInterface iface : getNetwork().getContext().getPollableNodes()) {
-            schedulePollableInterface(iface);    		
-    	}
+        ExecutorService executor = getExecutorService();
+        for (OnmsIpInterface iface : getNetwork().getContext().getPollableNodes()) {
+            executor.execute(() -> schedulePollableInterface(iface));
+        }
+        executor.shutdown();
```
RangerRick, line 249: I'm not doing this in an IDE so I'm not certain it's right, but I'd be doing something like...

```java
        ExecutorService executor = getExecutorService();
        List<Future<?>> futures = getNetwork().getContext().getPollableNodes()
            .stream()
            .map(iface -> executor.submit(() -> schedulePollableInterface(iface)).collect(Collectors.toList());
        CompletableFuture.allOf(futures).join();
```

RangerRick (changes requested):

opennms-services/src/main/java/org/opennms/netmgt/snmpinterfacepoller/SnmpPoller.java
RangerRick: ...which really could just be changed so `schedulePollableInterface(iface)` uses the executor and returns the future, instead, now that I think of it, since there's no reason not to do it in a thread pool for _all_ scheduled stuff, not just new existing ones on init... lol

opennms-services/src/main/java/org/opennms/netmgt/snmpinterfacepoller/SnmpPoller.java
WilssoN93: Aah! I see what you mean now. You're right, I will se what I can instead

opennms-services/src/main/java/org/opennms/netmgt/snmpinterfacepoller/SnmpPoller.java
WilssoN93: Good idea I will look into youre example and try it out!

opennms-services/src/main/java/org/opennms/netmgt/snmpinterfacepoller/SnmpPoller.java
WilssoN93: @RangerRick is this a good solution? 🙂

opennms-services/src/main/java/org/opennms/netmgt/snmpinterfacepoller/SnmpPoller.java
```diff
@@ -229,47 +233,59 @@ protected void onInit() {
...
-            if (pkgName != null) {
-                LOG.debug("Scheduling snmppolling for node: {} ip address: {} - Found package interface with name: {}", nodeid, ipaddress, pkgName);
-                scheduleSnmpCollection(getNetwork().create(nodeid,ipaddress,netmask,pkgName), pkgName, location);
-            } else if (!getPollerConfig().useCriteriaFilters()) {
-                LOG.debug("No SNMP Poll Package found for node: {} ip address: {}. - Scheduling according with default interval", nodeid, ipaddress);
-                scheduleSnmpCollection(getNetwork().create(nodeid, ipaddress,netmask, "null"), "null", location);
-            }
-        }
+    protected CompletableFuture<Void> schedulePollableInterface(OnmsIpInterface iface) {
+        CompletableFuture<Void> future = new CompletableFuture<>();
```
RangerRick, line 261: Looks to be on the right track, but this future seems extraneous... `runAsync()` already returns a completable future and you're using that as the return value for the method.

You could replace `future.complete(null);` with `return;` and then just wrap any exceptions with a `RuntimeException` (ideally create an `SnmpPollerException`) and you wouldn't have to create the extra future. Currently the exception gets eaten, because you're calling `completeExceptionally()` on a future that never gets returned anywhere.

RangerRick (changes requested):

RangerRick (approved):
This is looking pretty good to me! Just need the OCA signed and we should be able to merge this, IMHO.

RangerRick:
Thanks for this contribution, and I appreciate you working through it with us!

PR #6695, opened 2023-09-21
NMS-16143: Allow per connection pool config and split out monitoring connection

fooker:
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16143

PR #7038, opened 2024-01-24
NMS-16143: Allow per connection pool config and split out monitoring connection

fooker:
Backport of https://github.com/OpenNMS/opennms/pull/6695

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16107

### DIFF
--- a/core/db/src/main/java/org/opennms/core/db/DataSourceFactory.java
+++ b/core/db/src/main/java/org/opennms/core/db/DataSourceFactory.java
@@ -71,22 +71,22 @@ public abstract class DataSourceFactory {
     private static final List<Runnable> m_closers = new LinkedList<>();
 
     private static ClosableDataSource parseDataSource(final String dsName) {
-        String factoryClass = null;
+        final var jdbcDataSource = m_dataSourceConfigFactory.getJdbcDataSource(dsName);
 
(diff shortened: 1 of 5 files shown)


RangerRick (approved):

RangerRick (changes requested):
Wait, are we using the _admin_ connection at runtime?  Is there a reason it should be using `${install.database.admin.*}` instead of `${install.database.*}`? That seems like a huge security issue, we shouldn't be initializing any direct database connections as the admin user at runtime, only at install.

RangerRick (changes requested):
Oh wait, I guess this is for monitoring postgres itself? It doesn't appear that `foundation-2023` actually has any config that _uses_ that though.

Should we even be adding the `opennms-monitor` datasource in the backport? Or should we _also_ be backporting the selfmonitor support?

fooker:
> Oh wait, I guess this is for monitoring postgres itself? It doesn't appear that `foundation-2023` actually has any config that _uses_ that though.
> 
> Should we even be adding the `opennms-monitor` datasource in the backport? Or should we _also_ be backporting the selfmonitor support?

I don't think that we should backport self-monitoring at this point. Users still can build it themself, when it comes to DB monitoring.

RangerRick (approved):

PR #7041, opened 2024-01-25
NMS-16315: Disable User Data Collection in smoke tests

synqotik:
I *think* this will disable the User Data Collection notice from appearing in smoke tests. It sets Config Management with the user having already acknowledged the notice, and overlays any `opennms.userDataCollection.show` in the `etc/opennms.properties` file to `false` (production will have it commented-out, which enables it by default).

Will work on another PR to enable it just for a specific smoke test.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16315.


cgorantla (approved):
lgtm!

RangerRick:
> I _think_ this will disable the User Data Collection notice from appearing in smoke tests. It overlays any `opennms.userDataCollection.show` in the `etc/opennms.properties` file to `false` (production will have it commented-out, which enables it by default).

Ideally a user (and in this case, smoke test is a user) should never edit `etc/opennms.properties` nowadays.
Instead you should create a file in `etc/opennms.properties.d/`.

RangerRick:
(Oh, should have looked at the patch first, guess this is already happening.)

synqotik:
> (Oh, should have looked at the patch first, guess this is already happening.)

Yes :)

PR #7076, opened 2024-02-08
NMS-16320: Setting correct location for VMware import

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-16320


integrations/opennms-vmware/src/main/java/org/opennms/netmgt/provision/service/vmware/VmwareImporter.java
```diff
@@ -249,6 +250,11 @@ private RequisitionNode createRequisitionNode(Set<String> ipAddresses, ManagedEn
         // Foreign Id consisting of managed entity Id
         requisitionNode.setForeignId(managedEntity.getMOR().getVal());
 
+        // set the location
+        if (!Strings.isNullOrEmpty(request.getLocation())) {
```
RangerRick, line 254: FYI we have a `StringUtils` in core.lib that can do this without having to pull in Guava... (yes, I'm still on about that, lol)

RangerRick (approved):

PR #7083, opened 2024-02-12
NMS-16208: Handle exception when transaction fails

cgorantla:
Handle exception. This doesn't resolve the issue completely

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16208


opennms-alarms/daemon/src/main/java/org/opennms/netmgt/alarmd/AlarmPersisterImpl.java
```diff
@@ -120,6 +120,9 @@ public OnmsAlarm persist(Event event) {
             locks.forEach(Lock::lock);
             // Process the alarm inside a transaction
             alarm = m_transactionOperations.execute((action) -> addOrReduceEventAsAlarm(event));
+        } catch (Exception e) {
+            LOG.warn("Exception while reducing event {} to alarm", event);
```
RangerRick, line 124: could we potentially want the stack trace from this? or will it show elsewhere so it's safe to eat it here?

cgorantla: Added!

RangerRick (approved):

PR #7098, opened 2024-02-16
NMS-16356: Add procps which is required by the Karaf status scripts

indigo423:
Add procps to the required RPMS in the production container image in our Minion and Sentinel.

## Review hint

I've looked into the Core container image and we don't provide the Karaf status script in the same way as in Minion and Sentinel, which is the reason, I haven't added it to the core image.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16356

### DIFF
--- a/opennms-container/minion/Dockerfile
+++ b/opennms-container/minion/Dockerfile
@@ -59,7 +59,7 @@ RUN chmod +x /tmp/plugins.sh  && cd /tmp && ./plugins.sh && rm ./plugins.sh
 ##
 FROM ${BASE_IMAGE}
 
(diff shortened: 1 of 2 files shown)


mmahacek (approved):

RangerRick (changes requested):
You can close this, I actually added it to the [deploy base](https://github.com/opennms/deploy-base) instead.

indigo423:
Solved in https://github.com/OpenNMS/deploy-base/commit/1a47a12d163b756889e7663de0b49584a072683a and will be shipped with the container image built with 33.0.2,

PR #7178, opened 2024-04-03
build rpm and debian packages monolithically

RangerRick:
Simplify our packaging a bit by doing all the RPMs together, and all the Debian packages together. This shaves off a total of about 8 minutes runtime in CircleCI credits, and since they run faster than Integration Test anyway, it doesn't slow down the pre-publish stage any.


mershad-manesh (approved):
Took a quick look and it LGTM

PR #7220, opened 2024-04-16
NMS-16406: Allow update-package-permissions to set ownership for customized users.

UberPinguin:
Because the step that changes file ownership keys off of the RUNAS variable, it does not make sense to bail out when RUNAS does not equal {{opennms}} and it can leave the installation in an inconsistent state.

Remove the check and exit, and allow the package upgrade to continue setting file ownership.

### All Contributors

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16406

### DIFF
--- a/opennms-base-assembly/src/main/filtered/bin/update-package-permissions
+++ b/opennms-base-assembly/src/main/filtered/bin/update-package-permissions
@@ -27,12 +27,6 @@ if ! id -u "${RUNAS}" >/dev/null 2>&1; then
   exit 1
 fi
 
-if [ "${RUNAS}" != "opennms" ]; then
-  echo "\$RUNAS has been changed from the default 'opennms' user. Assuming you know what you're doing and _not_ changing ownership of any files."
-  echo ""
-  exit 0
-fi
-
 RUNAS_UID="$(id -u "${RUNAS}" 2>/dev/null || :)"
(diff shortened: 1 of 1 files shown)


opennms-base-assembly/src/main/filtered/bin/update-package-permissions
```diff
@@ -27,12 +27,6 @@ if ! id -u "${RUNAS}" >/dev/null 2>&1; then
   exit 1
 fi
 
-if [ "${RUNAS}" != "opennms" ]; then
-  echo "\$RUNAS has been changed from the default 'opennms' user. Assuming you know what you're doing and _not_ changing ownership of any files."
-  echo ""
-  exit 0
-fi
```
RangerRick, line 34: Maybe it would be better to leave the warning in, but not exit?

RangerRick (changes requested):

opennms-base-assembly/src/main/filtered/bin/update-package-permissions
UberPinguin: I thought about doing that. I'll include it in the updated PR against foundation-2024 later this morning.

RangerRick:
@UberPinguin should we close this one then?

UberPinguin:
@RangerRick Yep, I was just in meetings so I hadn't done it yet.

PR #7226, opened 2024-04-17
NMS-16406: Allow update-package-permissions to set ownership for customized users.

UberPinguin:
Because the step that changes file ownership keys off of the RUNAS variable, it does not make sense to bail out when RUNAS does not equal `opennms` and it can leave the installation in an inconsistent state.

Modify the message printed to indicate that file ownership will be changed to the RUNAS user, and do not exit immediately.

### All Contributors

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16406

### DIFF
--- a/opennms-base-assembly/src/main/filtered/bin/update-package-permissions
+++ b/opennms-base-assembly/src/main/filtered/bin/update-package-permissions
@@ -28,7 +28,7 @@ if ! id -u "${RUNAS}" >/dev/null 2>&1; then
 fi
 
 if [ "${RUNAS}" != "opennms" ]; then
-  printf '%s\n\n' "\$RUNAS has been changed from the default 'opennms' user. Assuming you know what you're doing and changing ownership of files to ${RUNAS}."
+  printf '$RUNAS has been changed from the default "opennms" user. Assuming you know what you're doing and changing ownership of files to "%s".\n\n' "${RUNAS}"
 fi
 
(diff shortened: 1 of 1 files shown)


opennms-base-assembly/src/main/filtered/bin/update-package-permissions
```diff
@@ -28,9 +28,7 @@ if ! id -u "${RUNAS}" >/dev/null 2>&1; then
 fi
 
 if [ "${RUNAS}" != "opennms" ]; then
-  echo "\$RUNAS has been changed from the default 'opennms' user. Assuming you know what you're doing and _not_ changing ownership of any files."
-  echo ""
-  exit 0
+  printf '%s\n\n' "\$RUNAS has been changed from the default 'opennms' user. Assuming you know what you're doing and changing ownership of files to ${RUNAS}."
```
RangerRick, line 31: If you're using `printf` already, you could just do:

```suggestion
  printf '$RUNAS has been changed from the default "opennms" user. Assuming you know what you're doing and changing ownership of files to "%s".\n\n' "${RUNAS}"
```

:)

RangerRick (changes requested):

RangerRick (approved):

RangerRick:
👍  I'll go ahead and merge once it goes green (can't imagine it won't)

PR #7273, opened 2024-05-11
feat: allow for faster turn-round times when developing the ui

mvrueden:
This change allow for faster ui development with some quirks while doing so.

However, I think the tradeoff is bearable, especially as this is the way modern applications are meant to be developed.

Cheers

### All Contributors

### Contribution Checklist

### What's Next?

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/${JIRA-ISSUE-NUMBER}


fooker (approved):

christianpape (approved):
LGTM.

ui/src/services/axiosInstances.ts
```diff
@@ -22,21 +22,41 @@
...
+    }
+  }
+  return null;
+}
+const authHeaders = detectAuthHeaders();
+
 const v2 = axios.create({
   baseURL: import.meta.env.VITE_BASE_V2_URL?.toString() || '/opennms/api/v2',
-  withCredentials: true
+  withCredentials: authHeaders != null,
```
synqotik, line 40: I think we need to leave this as `withCredentials: true`, it's needed for normal non-vite-server usage. It makes axios send the JWT token, etc.

synqotik (changes requested):
Thanks for doing this, I meant to but didn't get around to it, though due to how the menu works there are some issues (which you noted).

See one comment re `withCredentials`.

ui/src/services/axiosInstances.ts
mvrueden: Is this re-used somewhere? Because the only difference here is, that it does not send the cookie when in "dev mode".

synqotik: If I'm reading this correctly, `detectAuthHeaders()` will be false if either `VITE_BASE_USER` or `VITE_BASE_PASSWORD` are not defined. These are only defined in `.env.development`, so won't be present during production. So in production, `withCredentials` will be set to `false` which we don't want.

In addition, we should have some additional fail safe check like:

```
withCredentials: import.meta.env.PROD || authHeaders != null
```

I may be missing something, but want to make sure this isn't a security hole.

mvrueden: I can take another spin at this at some point. Maybe we can just leave it as is and it will work nevertheless. Then we don't have to worry about it.

Just as a side note, it should be `authHeaders == null`, as otherwise it is definitely wrong

synqotik: @mvrueden I think that last part was really throwing me off. I'll be happy to approve this with 2 changes:

- update that line to `withCredentials: import.meta.env.PROD || authHeaders === null`
- add a comment in `detectAuthHeaders()`, something like this below, just to make the intention more clear: 

> `VITE_BASE_USER`  and `VITE_BASE_PASSWORD` should only be defined in development mode, not in production

Also, make sure to test in dev and prod mode and make sure this works in either case. This would be a nice thing to have for improving the development experience.

ui/README.md
```diff
@@ -64,6 +72,22 @@ Developer workflow for development -> build -> fast deploy. There may be issues
...
 
+### Fast (with some flaws)
+
+This method assumes, that an OpenNMS instance is running with default settings, as well as a modified CORS filter.
+Refer to https://docs.opennms.com/horizon/latest/development/rest/CORS.html for details on how to set it up correctly. 
+In short, simply enable the filter in the `web.xml` of the deployed OpenNMS.
+
+In `.env.development` you can configure a `VITE_BASE_USER` and `VITE_BASE_PASSWORD`, which then configures the axios http clients to use basic auth headers.
+
+With `vite` it should not be possible to work on the OpenNMS ui and see the changes immediately.
```
synqotik, line 83: Should this be: `it should be possible` ?

Also change `ui` to `UI`

PR #7279, opened 2024-05-14
NMS-16347: Delete old primary interface when changed

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-16347


opennms-model/src/main/java/org/opennms/netmgt/model/OnmsNode.java
```diff
@@ -1644,7 +1643,7 @@ public void mergeNode(OnmsNode scannedNode, EventForwarder eventForwarder, boole
 
         mergeSnmpInterfaces(scannedNode, deleteMissing);
 
-        mergeIpInterfaces(scannedNode, eventForwarder, deleteMissing);
+        mergeIpInterfaces(scannedNode, eventForwarder, true);
```
fooker, line 1646: If this is always `true`, the parameter can be removed.

fooker (approved):

fooker (approved):

PR #7319, opened 2024-06-15
chg: remove statsd, reportd and jasper-reports

mvrueden:
I created a discussion in the OpenNMS Discourse ([here](https://opennms.discourse.group/t/the-future-of-opennms-horizon/3952/23)) with the goal to get a feeling about what can be removed from the OpenNMS Horizon source code.

I created a working branch, called `develop-headless` to be the basis for the next upcoming removals of the features mentioned in said above Discourse Discussion.

In this current Pull Request `statsd`, `reportd` and `jasper-reports` is removed.

First I wanted to remove them individually, but they are glued together pretty good, so here we go.


mvrueden:
I am not sure why the NodeDetailsPageIT is failing, as removing the reporting should not have caused any side-effects in the Topology UI. Will investigate

smoke-test/src/main/java/org/opennms/smoketest/utils/RestClient.java
```diff
@@ -419,7 +419,7 @@ public JsonNode getBackups() throws IOException {
                 .accept(MediaType.APPLICATION_JSON)
                 .get(String.class);
 
-        if (result == null) {
+        if (result == null || result.isEmpty()) {
```
mvrueden, line 422: This is necessary, as due to the removal of jasper-reports, now jersey is actually used. Before apache cxf client was used (leaked in via the opennms reporting dependencies). It turns out, jersey and cxf have slightly different behavior.

christianpape (approved):
LGTM.

PR #7353, opened 2024-07-09
NMS-16449: Display system status only if log level is set to DEBUG

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-16449


synqotik (approved):
LGTM

PR #7412, opened 2024-09-06
NMS-16508: Password gate display fix to track change made in Meridian

synqotik:
The Password Gate in Meridian had incorrect, Horizon-specific values for some images and text colors. This was fixed in NMS-16484. This is to apply similar change in Horizon so that the code is the same except for the changed values in constants.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16508


cgorantla (approved):
LGTM!

PR #7431, opened 2024-09-17
NMS-15169: Added jackson-dataformat-yaml to opennms-health-rest-service feature

christianpape:
Should fix CircleCI Sentinel smoke test errors for develop.


mershad-manesh (approved):
The change looks good; lets wait for the builds to complete before delivering the changes.

PR #7437, opened 2024-09-23
 NMS-16535: Reorganize date time and notification in main navigation

indigo423:
Clarify date and time in separate lines, to a) save some space and b) make it easier to read. Group notification components for user and team notification and the indicator if the system has notifications enabled together.

Screenshot 2024-09-23 at 16 36 48

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16535

### DIFF
--- a/core/lib/src/main/java/org/opennms/core/time/CentralizedDateTimeFormat.java
+++ b/core/lib/src/main/java/org/opennms/core/time/CentralizedDateTimeFormat.java
@@ -43,7 +43,7 @@ public class CentralizedDateTimeFormat {
 
     public final static String SYSTEM_PROPERTY_DATE_FORMAT = "org.opennms.ui.datettimeformat";
     public final static String SESSION_PROPERTY_TIMEZONE_ID = "org.opennms.ui.timezoneid";
-    public final static String DEFAULT_FORMAT_PATTERN = "yyyy-MM-dd'T'HH:mm:ssxxx";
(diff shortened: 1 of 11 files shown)


fooker (approved):

synqotik:
From a Product standpoint, we'd prefer to leave the layout as-is, except for putting date/time on 2 lines as proposed.

Preference for now is to have the "gear" icons, etc. all the way to the right as in the current design.

synqotik (changes requested):
Requesting that the layout not be changed, but only change the layout of the date/time to be 2 lines, with the new `DEFAULT_FORMAT_PATTERN`.

indigo423:
Thanks for taking a look at it.

indigo423:
> Requesting that the layout not be changed, but only change the layout of the date/time to be 2 lines, with the new `DEFAULT_FORMAT_PATTERN`.

Doesn‘t make sense to me just changing the date time format because it doesn‘t solve the problem I wanted to address. Just having it two Lines at the Same place introduces other issues like alignment. Top right corner for date/time is a common pattern in UI’s in cultures who read left to right.

Notification related components are grouped in my proposal, because they relate to each other. Menu items are grouped and the gearbox is only available for users with the role Admin, which makes it a bit hard for me to follow your argument.

Feel free take the changes from my PR to change the date/time format.

PR #7459, opened 2024-10-07
NMS-16571 : Include nodeParentId in Rest API V2 returns for Node

MusaidAli:
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16571

### DIFF
--- a/opennms-model/src/main/java/org/opennms/netmgt/model/OnmsNode.java
+++ b/opennms-model/src/main/java/org/opennms/netmgt/model/OnmsNode.java
@@ -194,7 +194,7 @@ public class OnmsNode extends OnmsEntity implements Serializable, Comparable<Onm
 
     private List<OnmsMetaData> m_requisitionedMetaData = new ArrayList<>();
 
-    private Integer parentID;
+    private Integer m_parentID;
 
     /**
      * <p>
@@ -326,18 +326,19 @@ public class OnmsNode extends OnmsEntity implements Serializable, Comparable<Onm
         m_parent = parent;
     }
     @Transient
-    @XmlElement(name = "parent.id")
-    @JsonIgnore
+    @XmlAttribute(name = "parent.id")
     public Integer getParentID() {
 
         if( m_parent != null ) {
-           return m_parent.getId();
+           this.m_parentID= m_parent.getId();
         }
-        return null;
+        return this.m_parentID;
     }
 
     public void setParentID(Integer parentID) {
-        this.parentID = parentID;
+        if( m_parent != null ) {
+            this.m_parentID= m_parent.getId();
+        }
     }
 
     @XmlEnum
@@ -1117,6 +1118,7 @@ public class OnmsNode extends OnmsEntity implements Serializable, Comparable<Onm
         .add("type", m_type == null ? null : m_type.toString())
         .add("operatingSystem", m_operatingSystem)
         .add("lastIngressFlow", m_lastIngressFlow)
+        .add("m_parentID" ,m_parentID)
         .add("lastEgressFlow", m_lastEgressFlow)
         .toString();
     }
--- a/opennms-model/src/main/java/org/opennms/netmgt/model/OnmsNode.java
+++ b/opennms-model/src/main/java/org/opennms/netmgt/model/OnmsNode.java
@@ -194,8 +194,6 @@ public class OnmsNode extends OnmsEntity implements Serializable, Comparable<Onm
 
     private List<OnmsMetaData> m_requisitionedMetaData = new ArrayList<>();
 
-    private Integer m_parentID;
-
     /**
      * <p>
      * Constructor for OnmsNode. This constructor should only be used
@@ -330,15 +328,9 @@ public class OnmsNode extends OnmsEntity implements Serializable, Comparable<Onm
     public Integer getParentID() {
 
         if( m_parent != null ) {
-           this.m_parentID= m_parent.getId();
-        }
-        return this.m_parentID;
-    }
-
-    public void setParentID(Integer parentID) {
-        if( m_parent != null ) {
-            this.m_parentID= m_parent.getId();
-        }
+          return m_parent.getId();
+         }
+          return null;
     }
 
     @XmlEnum
@@ -1118,7 +1110,6 @@ public class OnmsNode extends OnmsEntity implements Serializable, Comparable<Onm
         .add("type", m_type == null ? null : m_type.toString())
         .add("operatingSystem", m_operatingSystem)
         .add("lastIngressFlow", m_lastIngressFlow)
-        .add("m_parentID" ,m_parentID)
         .add("lastEgressFlow", m_lastEgressFlow)
         .toString();
     }
--- a/opennms-model/src/main/java/org/opennms/netmgt/model/OnmsNode.java
+++ b/opennms-model/src/main/java/org/opennms/netmgt/model/OnmsNode.java
@@ -324,7 +324,6 @@ public class OnmsNode extends OnmsEntity implements Serializable, Comparable<Onm
         m_parent = parent;
     }
 
-
     // This already maps to the existing column ,readonly
     @XmlAttribute(name = "nodeParentID")
     @Column(name = "nodeParentID",insertable = false, updatable = false)
--- a/opennms-model/src/main/java/org/opennms/netmgt/model/OnmsNode.java
+++ b/opennms-model/src/main/java/org/opennms/netmgt/model/OnmsNode.java
@@ -193,7 +193,7 @@ public class OnmsNode extends OnmsEntity implements Serializable, Comparable<Onm
     private List<OnmsMetaData> m_metaData = new ArrayList<>();
 
     private List<OnmsMetaData> m_requisitionedMetaData = new ArrayList<>();
-
+    private Integer nodeParentId;
     /**
      * <p>
      * Constructor for OnmsNode. This constructor should only be used
@@ -307,7 +307,7 @@ public class OnmsNode extends OnmsEntity implements Serializable, Comparable<Onm
      *
      * @return a {@link org.opennms.netmgt.model.OnmsNode} object.
      */
-    @XmlTransient
+
     @JsonIgnore
     @ManyToOne(fetch=FetchType.LAZY)
     @JoinColumn(name="nodeParentID")
@@ -323,14 +323,17 @@ public class OnmsNode extends OnmsEntity implements Serializable, Comparable<Onm
     public void setParent(OnmsNode parent) {
         m_parent = parent;
     }
-    @Transient
-    @XmlAttribute(name = "parent.id")
-    public Integer getParentID() {
 
-        if( m_parent != null ) {
-          return m_parent.getId();
-         }
-          return null;
+
+    // This already maps to the existing column ,readonly
+    @XmlAttribute(name = "nodeParentID")
+    @Column(name = "nodeParentID",insertable = false, updatable = false)
+    public Integer getNodeParentId() {
+        return nodeParentId;
+    }
+
+    public void setNodeParentId(Integer nodeParentId) {
+        this.nodeParentId = nodeParentId;
     }
 
     @XmlEnum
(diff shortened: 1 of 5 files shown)


opennms-model/src/main/java/org/opennms/netmgt/model/OnmsNode.java
```diff
@@ -323,6 +325,20 @@ public OnmsNode getParent() {
...
+    @JsonIgnore
+    public Integer getParentID() {
+
+        if( m_parent != null ) {
+           return m_parent.getId();
+        }
+        return null;
+    }
+
+    public void setParentID(Integer parentID) {
```
cgorantla, line 339: Who is going to set this parentId ?

Can you add a test case in NodeRestServiceV2IT  to show that when ParentId is available, we can see it Rest endpoint.

cgorantla (changes requested):

opennms-model/src/main/java/org/opennms/netmgt/model/OnmsNode.java
MusaidAli: this parentID logic is changed ,we are getting from parent.getId() and setting it.
    public void setParentID(Integer parentID) {
        if( m_parent != null ) {
            this.m_parentID= m_parent.getId();
        }
    }

MusaidAli: Screenshot from 2024-10-07 20-22-02

cgorantla: i don't see any need for m_parentId and `setParentId`, we are only getting parentId from Rest API.

Also screenshot is not a replacement for test case.

MusaidAli: Yes ,it is correct removed unnecessary parameters.Will implement test case .it is just retrieval /v2/nodes for that we have to R&D to create node with parent .

opennms-model/src/main/java/org/opennms/netmgt/model/OnmsNode.java
```diff
@@ -323,6 +323,15 @@ public OnmsNode getParent() {
     public void setParent(OnmsNode parent) {
         m_parent = parent;
     }
+    @Transient
+    @XmlAttribute(name = "parent.id")
```
synqotik, line 327: Note that `nodeparentid` is already a column in the database, `node` table. I believe we just need to return that. We don't need to access `m_parent` since we are not getting any other data from the parent node.

MusaidAli: this return m_parent whole object .

    @XmlAttribute(name = "nodeParentID")
    @JsonIgnore
    @ManyToOne(fetch=FetchType.LAZY)
    @JoinColumn(name="nodeParentID")
    public OnmsNode getParent() {
        return m_parent;
    }
if we want to return on id then the following code .

    @Transient
    @XmlAttribute(name = "parent.id")
    public Integer getParentID() {

        if( m_parent != null ) {
          return m_parent.getId();
         }
          return null;
    }
    
    I think it will be better to follow first code to return whole object.

MusaidAli: this is returning m_parent whole node .
    @JsonIgnore
    @ManyToOne(fetch=FetchType.LAZY)
    @JoinColumn(name="nodeParentID")
    @XmlAttribute(name = "nodeParentID")
    public OnmsNode getParent() {
        return m_parent;
    }
If we want id then the following code .

     @XmlAttribute(name = "parent.id")
    public Integer getParentID() {

        if( m_parent != null ) {
          return m_parent.getId();
         }
          return null;
    }

MusaidAli: I think this will work.  

  // This already maps to the existing column ,readonly
    @XmlAttribute(name = "nodeParentID")
    @Column(name = "nodeParentID",insertable = false, updatable = false)
    public Integer getNodeParentId() {
        return nodeParentId;
    }

    public void setNodeParentId(Integer nodeParentId) {
        this.nodeParentId = nodeParentId;
    }

mvrueden:
Please keep in mind, that node id is not the only criteria to address/reference a node. In some cases the `foreignSource:foreignId` is used instead or the preferred way. Therefore it might be feasible to actually expose both if you want to do it properly in my opinion.

cgorantla (changes requested):
Once you open a PR, we need commits without [skip ci] for the tests to run. otherwise  CI/CD won't run any tests

opennms-webapp-rest/src/test/java/org/opennms/web/rest/v2/AlarmRestServiceIT.java
```diff
@@ -801,4 +801,26 @@ private void executeQueryAndVerify(final String query, final int totalCount) thr
...
+        builder.addNode("Child").setForeignSource("Junit").setForeignId("Child").setType(NodeType.ACTIVE)
+                .setParent(parent).setNodeParentId(parent.getId());
+        final OnmsNode child = builder.getCurrentNode();
+        m_databasePopulator.getNodeDao().save(child);
+        m_databasePopulator.getNodeDao().flush();
+
+        List<OnmsNode> list =  m_databasePopulator.getNodeDao().findByLabel("Child");
+       if(list!=null && !list.isEmpty()) {
+
+           Assert.assertEquals(parent.getId(),list.get(0).getNodeParentId());
```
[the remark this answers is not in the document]
MusaidAli: They are not null it has id's like 4,4 in my case.

In rest api it is not reflecting ,parent is not being added like i tried in NodeRestServiceIT even when we put nodeParentID in jsob object is showing null ,In Curl command xml attribute is showing .In dao layer change is reflecting .If not using AlarmRestServiceIT where we should test in dao layer.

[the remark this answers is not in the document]
cgorantla: > They are not null it has id's like 4,4 in my case.
Validate in the test case itself,  don't assume it.

we have to test NodeRestService V2 to have parentId when it is available in DB.  Try to find a way to add a test case.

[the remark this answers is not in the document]
MusaidAli: I have validated it ,it was 4,4 while debugging .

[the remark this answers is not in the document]
cgorantla: > I have validated it ,it was 4,4 while debugging .

Validate it in code in test case that at least one of them is not null before  testing that they are equal.

synqotik:
> Please keep in mind, that node id is not the only criteria to address/reference a node. In some cases the `foreignSource:foreignId` is used instead or the preferred way. Therefore it might be feasible to actually expose both if you want to do it properly in my opinion.

@mvrueden Main reason for this PR is because a customer requested the node parent ID in the Grafana Plugin (OPG).

We already store the `nodeparentid` in the database, so this is the least expensive way to return it in the Rest API. It seems most of the time, there is no parent ID, or user doesn't need it, so it seemed expensive to fetch all the parent node data on each request.

If an OPG user wants the FS:FID, we can optionally make an additional Rest call to get it, or a Grafana dashboard can be set up so this is done automatically.

opennms-webapp-rest/src/test/resources/v1/nodes.json
```diff
@@ -79,6 +79,7 @@
             "sysObjectId": ".1.3.6.1.4.1.8072.3.2.255",
             "lastEgressFlow" : null,
             "lastIngressFlow" : null,
+            "nodeParentID" : null,
```
cgorantla, line 82: Is this needed

MusaidAli: yes ,it is being used in the following test case .It was giving error unexpected nodeParentID ,after adding here it resolved.


    @Test
    @JUnitTemporaryDatabase
    public void testAnotherNodeJson() throws Exception {

opennms-webapp-rest/src/test/java/org/opennms/web/rest/v2/NodeRestServiceIT.java
```diff
@@ -241,5 +251,26 @@ public void testAllEndPointsWithJSON() throws Exception {
...
+        final OnmsNode child = builder.getCurrentNode();
+        m_databasePopulator.getNodeDao().save(child);
+        m_databasePopulator.getNodeDao().flush();
+
+        LOG.warn(sendRequest(GET, "/nodes", 200));
+        LOG.warn(sendRequest(GET, "/nodes/1", 200)); // By ID
+
+        final String response = sendRequest(GET, "/nodes/2", 200);
+
+        final JSONObject object = new JSONObject(response);
```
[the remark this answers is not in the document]
MusaidAli: Added NotNull check.  

      Assert.assertNotNull(child.getId());
        Assert.assertNotNull(parent.getId());

opennms-webapp-rest/src/test/java/org/opennms/web/rest/v2/NodeRestServiceIT.java
```diff
@@ -241,5 +251,26 @@ public void testAllEndPointsWithJSON() throws Exception {
...
+        final OnmsNode parent = builder.getCurrentNode();
+        m_databasePopulator.getNodeDao().save(parent);
 
+        builder.addNode("Child").setForeignSource("Junit").setForeignId("Child").setType(OnmsNode.NodeType.ACTIVE)
+                .setParent(parent).setNodeParentId(parent.getId());
+        final OnmsNode child = builder.getCurrentNode();
+        m_databasePopulator.getNodeDao().save(child);
+        m_databasePopulator.getNodeDao().flush();
+
+        LOG.warn(sendRequest(GET, "/nodes", 200));
```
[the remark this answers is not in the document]
MusaidAli: this was used in other method ,I think not needed ,removed.

opennms-webapp-rest/src/test/java/org/opennms/web/rest/v2/NodeRestServiceIT.java
```diff
@@ -241,5 +251,26 @@ public void testAllEndPointsWithJSON() throws Exception {
...
+        LOG.warn(sendRequest(GET, "/nodes/1", 200)); // By ID
```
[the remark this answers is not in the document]
MusaidAli: resolved ,added node id from above.

opennms-webapp-rest/src/test/java/org/opennms/web/rest/v2/NodeRestServiceIT.java
```diff
@@ -241,5 +251,25 @@ public void testAllEndPointsWithJSON() throws Exception {
...
 
+        builder.addNode("Child").setForeignSource("Junit").setForeignId("Child").setType(OnmsNode.NodeType.ACTIVE)
+                .setParent(parent).setNodeParentId(parent.getId());
+        final OnmsNode child = builder.getCurrentNode();
+        m_databasePopulator.getNodeDao().save(child);
+        m_databasePopulator.getNodeDao().flush();
+
+        Assert.assertNotNull(child.getId());
+        Assert.assertNotNull(parent.getId());
+        final String response = sendRequest(GET, "/nodes/"+child.getId(), 200);
```
cgorantla, line 270: Can also check for  GET for parentId

cgorantla (commented):
Again, once your PR is up, don't use [skip ci] to avoid skipping test cases.

opennms-webapp-rest/src/test/java/org/opennms/web/rest/v2/NodeRestServiceIT.java
MusaidAli: Yes,added GET for parentId.

cgorantla (approved):
LGTM!
Screenshot 2024-10-17 at 6 24 32 PM
Screenshot 2024-10-17 at 6 25 22 PM


Validated this by editing requisition to add parent id and by validating rest end point.
Please see attached images

PR #7463, opened 2024-10-08
BE: NMS-16536, Provisioning fails when category has been deleted

smunir-onms:
- https://opennms.atlassian.net/browse/NMS-16536


cgorantla (commented):
Were you able to reproduce this issue on foundation-2023 without any changes to the code ?

smunir-onms:
[Screencast from 10-17-2024 01:42:44 AM.webm](https://github.com/user-attachments/assets/46f7eb8c-b29d-4d8d-bc25-da74b2d030c1)
Unit test validation screen-recording attached

opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/DefaultProvisionService.java
```diff
@@ -1033,6 +1037,11 @@ private boolean handleCategoryChanges(final OnmsNode dbNode) {
                     }
                 }
 
+                // If some categories were removed or there are new to be added, reset the cache.
+                if (m_categoryCache.get() != null && (changed || !categories.isEmpty())) {
```
cgorantla, line 1041: The logic of reloading cache looks to fine to me.  You may not need `m_categoryCache.get() != null` check here.  

Do you need  checks like `changed` or  `categories.isEmpty()` here ?  

we may have to reload cache irrespective of these flags.

cgorantla (commented):
Check if you can add a test case for this in ProvisionerIT using actual Daos instead of mocks.

opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/DefaultProvisionService.java
smunir-onms: > The logic of reloading cache looks to fine to me. You may not need `m_categoryCache.get() != null` check here.
> 
> Do you need checks like `changed` or `categories.isEmpty()` here ?
> 
> we may have to reload cache irrespective of these flags.
The m_categoryCache.get() != null check is used because the cache is reloaded in the following lines if it's null. If we skip this check, we would need to modify the subsequent logic to prevent an additional database call to reload the cache.

The changed and categories.isEmpty() checks are also important to avoid unnecessary reloads on each synchronization. If the cache is already loaded and there have been no changes (i.e., no categories have been added or removed), then there's no need to reload the cache.

Let me know in case of any confusion.

christianpape (dismissed):
LGTM.

opennms-provision/opennms-provisiond/src/test/java/org/opennms/netmgt/provision/service/ProvisionerIT.java
smunir-onms, line 1: The last three asserts should also be passed for a complete scenario, there are some IT DAO limitations;

When m_provisionService.updateNodeAttributes(n); is called from the test class

image

the last category with ID 3 was deleted
Screenshot from 2024-10-25 21-56-10
the related association was also deleted
Screenshot from 2024-10-25 22-00-36

new category created with same name with new ID 4 
Screenshot from 2024-10-25 22-05-57 

new association, with the same node and new category,  going to be created 
Screenshot from 2024-10-25 22-06-51

after this action, the category was updated with the previous ID 3 in the association
Screenshot from 2024-10-25 22-03-07

the category dao was also updated with a duplicate category with old ID 3
Screenshot from 2024-10-25 22-02-09

cgorantla:
@smunir-onms can you merge latest `foundation-2023` in this branch

smunir-onms:
> @smunir-onms can you merge latest `foundation-2023` in this branch

Done.

opennms-provision/opennms-provisiond/src/test/java/org/opennms/netmgt/provision/service/ProvisionerIT.java
```diff
@@ -1797,6 +1803,52 @@ public void testRequisitionedCategoriesThenUpdateRequisitionToRemoveCategory() t
...
+        RequisitionedCategoryAssociation deleteCategoryAssociation = m_categoryAssociationDao.findAll().stream()
+                .filter(cat -> cat.getCategory().getName().equals("NMS-16536")).findFirst().orElse(null);
+        m_categoryAssociationDao.delete(deleteCategoryAssociation);
+        m_categoryAssociationDao.flush();
+
+        assertEquals(2, m_categoryAssociationDao.findAll().size());
+
+        m_provisionService.updateNodeAttributes(n);
+
+        assertEquals(3, m_categoryAssociationDao.findAll().size());
```
cgorantla, line 1836: As I understand, `m_provisionService.updateNodeAttributes(n)` shouldn't add categories back, right ?   It should update cache based  by loading dao again. What is missing here ?

smunir-onms: As per the flow, when a new category is added via the UI, it is stored in the database if it does not already exist. In the current scenario, the category was deleted from the database but is still being sent from the UI. Therefore, this assert ensures that association is added to the requisition categories.

cgorantla (approved):
LGTM!

Thanks for the fixes.

PR #7464, opened 2024-10-08
MPLUG-21:  Add monitored service status

cgorantla:
Add status.
Consider status is `Up` only if it is in Managed state and there are no outages.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/MPLUG-21


christianpape (approved):
LGTM.

fooker (approved):

PR #7482, opened 2024-10-21
NMS-16482: Version in about page points to invalid release notes.

Junaid-Khan-Nant:
- https://opennms.atlassian.net/browse/NMS-16482

### DIFF
--- a/opennms-webapp/src/main/web-filtered/WEB-INF/version.properties
+++ b/opennms-webapp/src/main/web-filtered/WEB-INF/version.properties
@@ -28,3 +28,4 @@
 
 version.display=${display.version}
 opennms.product=${product.name}
+opennms.docs.version=${docs.product.version}
--- a/opennms-webapp/src/main/web-filtered/WEB-INF/version.properties
+++ b/opennms-webapp/src/main/web-filtered/WEB-INF/version.properties
@@ -26,6 +26,5 @@
 #     http://www.opennms.com/
 ###############################################################################
 
-version.display=${display.version}
+version.display=2024.1.3
 opennms.product=meridian
-release.version=2024
--- a/opennms-webapp/src/main/web-filtered/WEB-INF/version.properties
+++ b/opennms-webapp/src/main/web-filtered/WEB-INF/version.properties
@@ -27,3 +27,5 @@
 ###############################################################################
 
 version.display=${display.version}
+opennms.product=meridian
+release.version=2024
--- a/opennms-webapp/src/main/web-filtered/WEB-INF/version.properties
+++ b/opennms-webapp/src/main/web-filtered/WEB-INF/version.properties
@@ -26,5 +26,5 @@
 #     http://www.opennms.com/
 ###############################################################################
 
-version.display=2024.1.3
-opennms.product=meridian
+version.display=${display.version}
+opennms.product=${product.name}
--- a/opennms-webapp/src/main/webapp/about/index.jsp
+++ b/opennms-webapp/src/main/webapp/about/index.jsp
@@ -70,7 +70,7 @@
 <table class="table table-sm">
   <tr>
     <th>Version:</th>
-    <td><a href="https://docs.opennms.com/horizon/<%=Vault.getProperty("version.display")%>/releasenotes/whatsnew.html" target="_blank" title="Release Notes"><%=Vault.getProperty("version.display")%></a></td>
+    <td><a href="https://docs.opennms.com/meridian/<%="2024"%>/releasenotes/whatsnew.html" target="_blank" title="Release Notes"><%=Vault.getProperty("version.display")%></a></td>
   </tr>
 
   <tr>
--- a/opennms-webapp/src/main/webapp/about/index.jsp
+++ b/opennms-webapp/src/main/webapp/about/index.jsp
@@ -70,7 +70,7 @@
 <table class="table table-sm">
   <tr>
     <th>Version:</th>
-    <td><a href="https://docs.opennms.com/<%=Vault.getProperty("opennms.product")%>/<%=Vault.getProperty("release.version")%>/releasenotes/whatsnew.html" target="_blank" title="Release Notes"><%=Vault.getProperty("opennms.product")%> - <%=Vault.getProperty("release.version")%></a></td>
+    <td><a href="https://docs.opennms.com/<%=Vault.getProperty("opennms.product")%>/<%=Vault.getProperty("version.display")%>/releasenotes/whatsnew.html" target="_blank" title="Release Notes"><%=Vault.getProperty("opennms.product")%> - <%=Vault.getProperty("version.display")%></a></td>
   </tr>
 
   <tr>
--- a/opennms-webapp/src/main/webapp/about/index.jsp
+++ b/opennms-webapp/src/main/webapp/about/index.jsp
@@ -70,7 +70,7 @@
 <table class="table table-sm">
   <tr>
     <th>Version:</th>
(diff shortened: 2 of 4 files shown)


opennms-webapp/src/main/webapp/about/index.jsp
```diff
@@ -70,7 +70,7 @@
 <table class="table table-sm">
   <tr>
     <th>Version:</th>
-    <td><a href="https://docs.opennms.com/horizon/<%=Vault.getProperty("version.display")%>/releasenotes/whatsnew.html" target="_blank" title="Release Notes"><%=Vault.getProperty("version.display")%></a></td>
+    <td><a href="https://docs.opennms.com/meridian/<%="2024"%>/releasenotes/whatsnew.html" target="_blank" title="Release Notes"><%=Vault.getProperty("version.display")%></a></td>
```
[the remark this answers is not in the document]
smunir-onms: Right, to make it dynamic there should be a property like "opennms.product" and the version property shouldn't be replaced with hardcoded "value 2024"

indigo423 (changes requested):
This change needs to be made in the private repository https://github.com/opennms/opennms-prime. @cgorantla, yes this is a good idea having a property where we can set the "distribution" at compile time would be helpful.

Junaid-Khan-Nant:
> This change needs to be made in the private repository https://github.com/opennms/opennms-prime. @cgorantla, yes this is a good idea having a property where we can set the "opennms.product" at compile time would be helpful.

@cgorantla , don't have access for this [repository.](https://github.com/opennms/opennms-prime).

opennms-webapp/src/main/webapp/about/index.jsp
```diff
@@ -70,7 +70,7 @@
 <table class="table table-sm">
   <tr>
     <th>Version:</th>
-    <td><a href="https://docs.opennms.com/horizon/<%=Vault.getProperty("version.display")%>/releasenotes/whatsnew.html" target="_blank" title="Release Notes"><%=Vault.getProperty("version.display")%></a></td>
+    <td><a href="https://docs.opennms.com/<%=Vault.getProperty("opennms.product")%>/<%=Vault.getProperty("release.version")%>/releasenotes/whatsnew.html" target="_blank" title="Release Notes"><%=Vault.getProperty("opennms.product")%> - <%=Vault.getProperty("release.version")%></a></td>
```
smunir-onms, line 73: Don't add a new property "release.version", use "version.display" instead, docs page path and properties values would be adjusted accordingly. @cgorantla @indigo423 please correct me if am wrong. 
```suggestion 
    <td><a href="https://docs.opennms.com/<%=Vault.getProperty("opennms.product")%>/<%=Vault.getProperty("release.version")%>/releasenotes/whatsnew.html" target="_blank" title="Release Notes"><%=Vault.getProperty("opennms.product")%> - <%=Vault.getProperty("version.display")%></a></td>
```

opennms-webapp/src/main/web-filtered/WEB-INF/version.properties
```diff
@@ -26,6 +26,5 @@
 #     http://www.opennms.com/
 ###############################################################################
 
-version.display=${display.version}
-opennms.product=meridian
-release.version=2024
\ No newline at end of file
+version.display=2024.1.3
```
indigo423, line 29: We should not hardcode the version here. The `${display.version}` is set from the pom.xml using `${install.version}` and follows automatically the correct version number. Hardcoding the version number would require dealing with it manually.

opennms-webapp/src/main/webapp/about/index.jsp
```diff
@@ -70,7 +70,7 @@
 <table class="table table-sm">
   <tr>
     <th>Version:</th>
-    <td><a href="https://docs.opennms.com/<%=Vault.getProperty("opennms.product")%>/<%=Vault.getProperty("release.version")%>/releasenotes/whatsnew.html" target="_blank" title="Release Notes"><%=Vault.getProperty("opennms.product")%> - <%=Vault.getProperty("release.version")%></a></td>
+    <td><a href="https://docs.opennms.com/<%=Vault.getProperty("opennms.product")%>/<%=Vault.getProperty("version.display")%>/releasenotes/whatsnew.html" target="_blank" title="Release Notes"><%=Vault.getProperty("opennms.product")%> - <%=Vault.getProperty("version.display")%></a></td>
```
indigo423, line 73: What about doing a similar thing, setting a property in the pom.xml like `${install.product}` which is evaluated from a single place from the pom.xml which we can control at build time with maven?

indigo423 (changes requested):
We should think about a way to control these settings at build time with maven. It allows us to set it in the build jobs for Meridian / Horizon rather than hard coding it in properties.

Junaid-Khan-Nant:
Hi @indigo423,  the property related to product name has been added and tested, and it's working fine. I am facing issues on the version as it is also used in inner pom's xml. If I replace this 33.0.3-SNAPSHOT with 2024.1.3-SNAPSHOT, then initial is working fine, but after a few times it stops the server with the following logs.
log-2
log-1
Any idea about it? Or I believe there is another way to create a new tag just like we created for the product and mention the release version there.It would be in parent pom.xml.Your opinion?

pom.xml
```diff
@@ -1668,6 +1668,8 @@
     <mock.db.adminPassword/>
     <mock.runSnmpTests>false</mock.runSnmpTests>
     <mock.snmpHost>localhost</mock.snmpHost>
+    <install.product>meridian</install.product>
```
mershad-manesh, line 1671: This should be `horizon` as `Foundation-` branches are based on `horizon`.

pom.xml
```diff
@@ -1668,6 +1668,8 @@
...
+    <product.name>${install.product}</product.name>
```
cgorantla, line 1672:
```suggestion
    <product.name>meridian</product.name>
```

pom.xml
cgorantla, line 1671:
```suggestion
```

pom.xml
cgorantla: Add another property as
<docs.product.version>24</docs.product.version>

cgorantla (commented):
Also you can update this in following pages

````
git grep "docs.opennms.com"
src/main/webapp/about/index.jsp:73:    <td><a href="https://docs.opennms.com/horizon/<%=Vault.getProperty("version.display")%>/releasenotes/whatsnew.html" target="_blank" title="Release Notes"><%=Vault.getProperty("version.display")%></a></td>
src/main/webapp/includes/help-documentation.jsp:63:                            href="https://docs.opennms.com/horizon/<%=Vault.getProperty("version.display")%>/deployment/core/introduction.html"
src/main/webapp/includes/help-documentation.jsp:70:                            href="https://docs.opennms.com/horizon/<%=Vault.getProperty("version.display")%>/operation/overview/overview.html#overview"
src/main/webapp/includes/help-documentation.jsp:77:                            href="https://docs.opennms.com/horizon/<%=Vault.getProperty("version.display")%>/development/overview/overview.html#overview"
```

pom.xml
```diff
@@ -1668,6 +1668,8 @@
     <mock.db.adminPassword/>
     <mock.runSnmpTests>false</mock.runSnmpTests>
     <mock.snmpHost>localhost</mock.snmpHost>
+    <product.name>meridian</product.name>
+    <docs.product.version>2024</docs.product.version>
```
cgorantla, line 1672:
```suggestion
    <docs.product.version>33</docs.product.version>
```

cgorantla (approved):
LGTM!

PR #7489, opened 2024-10-22
NMS-16587: Update linux UBI version to 9.4

mershad-manesh:
Updating the container's base image to UBI 9.4

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16587


cgorantla (approved):
LGTM!

PR #7492, opened 2024-10-23
NMS-16367 : Grafana reports endpoint failure

MusaidAli:
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16367


features/endpoints/grafana/client/pom.xml
```diff
@@ -21,7 +21,7 @@
             <Bundle-RequiredExecutionEnvironment>JavaSE-1.8</Bundle-RequiredExecutionEnvironment>
             <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
             <Bundle-Version>${project.version}</Bundle-Version>
-            <Embed-Dependency>*;artifactId=okhttp|gson|okio</Embed-Dependency>
+            <!--<Embed-Dependency>*;artifactId=okhttp|gson|okio</Embed-Dependency>-->
```
cgorantla, line 24: It's important to understand why we are doing this. Why did this fail when we upgraded okhttp3.

container/features/src/main/resources/features.xml
```diff
@@ -1640,14 +1640,22 @@
     </feature>
 
     <feature name="opennms-endpoints-grafana" version="${project.version}" description="OpenNMS :: Features :: Endpoints :: Grafana">
+
+        <bundle dependency="true">mvn:org.jetbrains.kotlin/kotlin-osgi-bundle/${kotlinVersion}</bundle>
+        <bundle>wrap:mvn:com.google.code.gson/gson/${gsonVersion}</bundle>
+        <bundle dependency='true'>mvn:org.apache.servicemix.bundles/org.apache.servicemix.bundles.okhttp/2.7.5_1</bundle>
+        <bundle dependency='true'>mvn:org.apache.servicemix.bundles/org.apache.servicemix.bundles.okio/1.13.0_1</bundle>
+        <bundle dependency='true'>wrap:mvn:com.squareup.okhttp3/okhttp/3.5.0</bundle>
```
cgorantla, line 1648: There are multiple versions of okhttp in the same feature file.  One as wrap bundle
```
<bundle dependency='true'>mvn:org.apache.servicemix.bundles/org.apache.servicemix.bundles.okhttp/2.7.5_1</bundle>
```
one from servicemix
```
<bundle dependency='true'>mvn:org.apache.servicemix.bundles/org.apache.servicemix.bundles.okhttp/2.7.5_1</bundle>
```

And the original one on pom file is different. ` <okhttpVersion>4.9.3</okhttpVersion>` 

There is no clarity on what is getting used at runtime.

features/endpoints/grafana/client/pom.xml
MusaidAli: I have generate locally the runtime dependencies in target and okhttp3 was there .

cgorantla:
I looked into these libraries. It looks like okhttp3 > 4.x has converted all the classes into kotlin files. That may be the reason for break in the functionality.  

There are two ways we can go about this.
1.  Use [servicemix](https://mvnrepository.com/artifact/org.apache.servicemix.bundles/org.apache.servicemix.bundles.okhttp) bundles  > 4.x and resolve any issues that we encounter.
2.  Downgrade okhttp3 to 3.x only for that specific feature and see that it resolves the issues.

MusaidAli:
> 2\. Downgrade okhttp3 to 3.x only for that specific feature and see that it resolves the issues.

MusaidAli:
> There are two ways we can go about this.
> 
>     1. Use [servicemix](https://mvnrepository.com/artifact/org.apache.servicemix.bundles/org.apache.servicemix.bundles.okhttp) bundles  > 4.x and resolve any issues that we encounter.
> 
>     2. Downgrade okhttp3 to 3.x only for that specific feature and see that it resolves the issues.

``` Used option 2 . Downgrade okhttp3 to 3.x only for that specific feature and see that it resolves the issues.
okhttp3 version 3.x works fine ```

File: features/endpoints/grafana/client/pom.xml [hunk 3]
```diff
@@ -55,12 +55,12 @@
     <dependency>
       <groupId>com.squareup.okhttp3</groupId>
       <artifactId>okhttp</artifactId>
-      <version>${okhttpVersion}</version>
+      <version>3.10.0</version>
```
cgorantla, line 58: Use version [3.14.9](https://mvnrepository.com/artifact/com.squareup.okhttp3/okhttp/3.14.9) which is latest on 3.x and it's dependent version  for okio [1.17.2](https://mvnrepository.com/artifact/com.squareup.okio/okio/1.17.2)

Declare these as `okhttp3Version` and  `okioGrafanaVersion` in original pom file

You have to re-target this branch to `release-33.x` 

We also need to add a smoke test with selenium to make sure this will not break again.

cgorantla (changes requested):

mershad-manesh (changes requested):
There are invalid changes being merged from `develop` into `release-33.x`

PR #7507, opened 2024-10-30
NMS-16543 With use-address-from-varbind traps are misassigned 

Mohammad-Junid:
- https://opennms.atlassian.net/browse/NMS-16543

### DIFF
--- a/features/events/traps/src/main/java/org/opennms/netmgt/trapd/EventCreator.java
+++ b/features/events/traps/src/main/java/org/opennms/netmgt/trapd/EventCreator.java
@@ -54,7 +54,7 @@ class EventCreator {
         this.eventConfDao = Objects.requireNonNull(eventConfDao);
     }
 
-    public Event createEventFrom(final TrapDTO trapDTO, final String systemId, final String location, final InetAddress trapAddress) {
+    public Event createEventFrom(final TrapDTO trapDTO, final String systemId, final String location, final InetAddress trapAddress,boolean shouldUseAddressFromVarbind) {
         LOG.debug("{} trap - trapInterface: {}", trapDTO.getVersion(), trapDTO.getAgentAddress());
 
         // Set event data
@@ -88,7 +88,7 @@ class EventCreator {
         }
 
         // Resolve Node id and set, if known by OpenNMS
-        resolveNodeId(location, trapAddress)
+        resolveNodeId(location, shouldUseAddressFromVarbind ? trapDTO.getTrapAddress() : trapAddress)
                 .ifPresent(eventBuilder::setNodeid);
 
         // If there was no systemId in the trap message, assume that
--- a/features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapSinkConsumer.java
+++ b/features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapSinkConsumer.java
@@ -120,7 +120,7 @@ public class TrapSinkConsumer implements MessageConsumer<TrapInformationWrapper,
 						eachMessage,
 						messageLog.getSystemId(),
 						messageLog.getLocation(),
-						messageLog.getTrapAddress());
+						messageLog.getTrapAddress(),config.shouldUseAddressFromVarbind());
 				if (!shouldDiscard(event)) {
 					if (event.getSnmp() != null) {
 						trapdInstrumentation.incTrapsReceivedCount(event.getSnmp().getVersion());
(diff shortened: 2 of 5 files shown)


features/events/traps/src/test/java/org/opennms/netmgt/trapd/TrapdIT.java
```diff
@@ -283,6 +285,44 @@ public void testSnmpV2cTrapWithAddressFromVarbind() throws Exception {
         await().until(() -> m_mockEventIpcManager.getEventAnticipator().getAnticipatedEventsReceived(), hasSize(2));
     }
 
+
+@Test
+    public void testSnmpV2cTrapWithAddressFromVarbind2() throws Exception {
+        // Enable the feature (disabled by default)
+        m_trapdConfig.getConfig().setUseAddressFromVarbind(true);
```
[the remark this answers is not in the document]
Mohammad-Junid: Done

features/events/traps/src/test/java/org/opennms/netmgt/trapd/TrapdIT.java
```diff
@@ -283,6 +285,70 @@ public void testSnmpV2cTrapWithAddressFromVarbind() throws Exception {
...
+    EventBuilder secondaryTrapBuilder = new EventBuilder("uei.opennms.org/default/trap", "trapd");
+    secondaryTrapBuilder.setInterface(secondaryRemoteAddr);
+    secondaryTrapBuilder.setSnmpVersion("v2c");
+    m_mockEventIpcManager.getEventAnticipator().anticipateEvent(secondaryTrapBuilder.getEvent());
+
+    EventBuilder secondarySuspectBuilder = new EventBuilder(EventConstants.NEW_SUSPECT_INTERFACE_EVENT_UEI, "trapd");
+// The address in the secondarySuspect event should match the one specified in the varbind
+    secondarySuspectBuilder.setInterface(secondaryRemoteAddr);
+    m_mockEventIpcManager.getEventAnticipator().anticipateEvent(secondarySuspectBuilder.getEvent());
+    secondaryPdu.send(localhost, m_trapdConfig.getSnmpTrapPort(), "public");
```
cgorantla, line 345: i would prefer both pdus to be sent in one go, one after another to be very precise on timing.

features/events/traps/src/main/java/org/opennms/netmgt/trapd/TrapSinkConsumer.java
```diff
@@ -120,7 +120,7 @@ private Log toLog(TrapLogDTO messageLog) {
 						eachMessage,
 						messageLog.getSystemId(),
 						messageLog.getLocation(),
-						messageLog.getTrapAddress());
+						messageLog.getTrapAddress(),config.shouldUseAddressFromVarbind());
```
cgorantla, line 123:
```suggestion
						messageLog.getTrapAddress(),
						config.shouldUseAddressFromVarbind());
```

cgorantla, line 123: It's possible that user might have different configurations on Minion and OpenNMS ( in case they are only using Minion to receive traps, they might ignore updating trapd-configuration.xml) so instead of using `config.shouldUseAddressFromVarbind()` we should transmit the config option through TrapLogDTO and use that option here.

features/events/traps/src/main/java/org/opennms/netmgt/trapd/EventCreator.java
```diff
@@ -54,17 +54,22 @@ public EventCreator(InterfaceToNodeCache cache, EventConfDao eventConfDao) {
         this.eventConfDao = Objects.requireNonNull(eventConfDao);
     }
 
-    public Event createEventFrom(final TrapDTO trapDTO, final String systemId, final String location, final InetAddress trapAddress) {
+    public Event createEventFrom(final TrapDTO trapDTO, final String systemId, final String location, final InetAddress trapAddress,Boolean shouldUseAddressFromVarbind) {
         LOG.debug("{} trap - trapInterface: {}", trapDTO.getVersion(), trapDTO.getAgentAddress());
 
         // Set event data
+        final InetAddress sourceTrapAddress = Optional.ofNullable(shouldUseAddressFromVarbind)
```
cgorantla, line 61: I think we can simplify this by using trapAddress from TrapDTO when it is valid and fallback on trapAddress from TrapLogDTO when it is not valid.

cgorantla (commented):
This looks good to me. Good work @Mohammad-Junid !

Let me validate once and will approve this.
Can you take care of signing OCA.

cgorantla (approved):
LGTM!  Validated on OpenNMS and Minion.
As of now this can go on  develop.  

Although this doesn't change any configuration but this might need Minion/OpenNMS to be on the same minor version.

PR #7509, opened 2024-10-31
NMS-16510 : Search does not return any results for Asset Search strin…

Junaid-Khan-Nant:
- https://opennms.atlassian.net/browse/NMS-16510

### DIFF
--- a/features/search/providers/pom.xml
+++ b/features/search/providers/pom.xml
@@ -84,6 +84,7 @@
     <dependency>
       <groupId>org.opennms.core.test-api</groupId>
       <artifactId>org.opennms.core.test-api.db</artifactId>
+      <scope>test</scope>
     </dependency>
   </dependencies>
 </project>
--- a/features/search/providers/pom.xml
+++ b/features/search/providers/pom.xml
@@ -61,10 +61,29 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.opennms</groupId>
+      <artifactId>opennms-rrd-jrobin</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.opennms</groupId>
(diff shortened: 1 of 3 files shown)


cgorantla (commented):
Please add a test case either in integration or smoke tests.

cgorantla (approved):
LGTM. Please also validate if all other asset related fields have right names in the Restrictions.

Junaid-Khan-Nant:
Hi @cgorantla, All restrictions are checked; it seems fine except one that's maintContractExpires. But it is also working fine as we get value from getter; in that case, a screenshot is attached for reference. Don't need to change in it, I believe.
Screenshot from 2024-11-04 16-10-19

features/search/providers/pom.xml
```diff
@@ -60,11 +60,30 @@
...
+    </dependency>
     <dependency>
       <groupId>org.opennms.core.ipc.rpc</groupId>
       <artifactId>org.opennms.core.ipc.rpc.mock-impl</artifactId>
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.opennms.core.test-api</groupId>
+      <artifactId>org.opennms.core.test-api.db</artifactId>
```
cgorantla, line 86: <scope>test</scope>

features/search/providers/src/test/java/org/opennms/netmgt/search/providers/node/NodeAssetSearchProviderTest.java
```diff
@@ -0,0 +1,95 @@
+package org.opennms.netmgt.search.providers.node;
```
cgorantla, line 1: nit :missing copyright headers

cgorantla (approved):
LGTM!

PR #7539, opened 2024-11-21
NMS-16957: Added documentation for PTP monitor

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-16957

### DIFF
--- a/docs/modules/reference/pages/service-assurance/monitors/PtpMonitor.adoc
+++ b/docs/modules/reference/pages/service-assurance/monitors/PtpMonitor.adoc
@@ -50,20 +50,14 @@ This monitor implements the <<reference:service-assurance/introduction.adoc#ref-
 
 == Examples
 
-Example uses CentOS/RHEL path name.
-For Debian/Ubuntu, use `/var/lib/opennms/rrd/response`.
-
 Note that you must include the `monitor` section for each service in your definition.
 
 [source, xml]
 ----
  <service name="PTP" interval="300000" user-defined="false" status="on">
-     <pattern><![CDATA[^PTP-(?<clockPort>.*)-(?<portState>.*)$]]></pattern>
-     <parameter key="clock-port" value="${pattern:clockPort|requisition:ptp:clockPort|default}" />
-     <parameter key="port-state" value="${pattern:portState|requisition:ptp:portState|default}" />
-  </service>
+   <parameter key="clock-port" value="${requisition:ptp:clockPort|default}" />
+   <parameter key="port-state" value="${requisition:ptp:portState|default}" />
+ </service>
 
-  <monitor service="PTP" class-name="org.opennms.netmgt.poller.monitors.PtpMonitor" />
+ <monitor service="PTP" class-name="org.opennms.netmgt.poller.monitors.PtpMonitor" />
 ----
-
-The pattern scheme also allows a user to just associate a service name like `PTP-Ethernet2-slave` to define the clock-port `Ethernet2` and desired port-state `slave`.
--- a/docs/modules/reference/pages/service-assurance/monitors/PtpMonitor.adoc
+++ b/docs/modules/reference/pages/service-assurance/monitors/PtpMonitor.adoc
@@ -1,0 +1,69 @@
+
+= PtpMonitor
+:description: Learn how to configure and use the PtpMonitor in OpenNMS {page-component-title} to monitor the correct PTP port state for an interface.
+
+The Precision Time Protocol allows the precise synchronization of clocks in measurement and control systems implemented with packet-based networks.
+THE PRP monitor allows you to test the port state of SNMP-enabled interfaces using the https://www.rfc-editor.org/info/rfc8173[PTP MIB defined in RFC 8173].
+During the poll, it queries the device for the available clock-ports and their states and compares with a given expected state.
+If the state equals the desired state the service is considered up and down otherwise.
+The following states are available:
+
+ - initializing
+ - faulty
+ - disabled
+ - listening
+ - preMaster
+ - master
+ - passive
+ - uncalibrated
+ - slave
+
+== Monitor facts
+
+[cols="1,7"]
+|===
+| Class Name
+| `org.opennms.netmgt.poller.monitors.PtpMonitor`
+|===
+
+== Configuration and Use
+
+.Optional monitor-specific parameters for the PtpMonitor
+[options="header"]
+[cols="1,3,2"]
+|===
+| Parameter
+| Description
+| Default
+
+| clock-port
+| Name of the interface that must be in the specified state.
+| n/a
+
+| port-state
+| The desired state that the port must be in.
+| n/a
+
+|===
+
+This monitor implements the <<reference:service-assurance/introduction.adoc#ref-service-assurance-monitors-common-parameters, Common Configuration Parameters>>.
+
+== Examples
+
+Example uses CentOS/RHEL path name.
+For Debian/Ubuntu, use `/var/lib/opennms/rrd/response`.
+
+Note that you must include the `monitor` section for each service in your definition.
+
+[source, xml]
+----
+ <service name="PTP" interval="300000" user-defined="false" status="on">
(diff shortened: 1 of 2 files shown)


docs/modules/reference/pages/service-assurance/monitors/PtpMonitor.adoc
```diff
@@ -0,0 +1,60 @@
...
+For Debian/Ubuntu, use `/var/lib/opennms/rrd/response`.
+
+Note that you must include the `monitor` section for each service in your definition.
+
+[source, xml]
+----
+ <service name="PTP" interval="300000" user-defined="false" status="on">
+     <pattern><![CDATA[^PTP-(?<clockPort>.*)-(?<portState>.*)$]]></pattern>
+     <parameter key="rrd-repository" value="${install.share.dir}/rrd/response" />
+     <parameter key="rrd-base-name" value="ptp" />
```
indigo423, line 52: We can probably eliminate the problem with ${install.share.dir} removing this configuration parameter. I think there is a default set in the poller we can rely on.

We should probably also remove the configuration:

```
<parameter key="rrd-base-name" value="ptp" />
```

If you use the pattern match and have multiple services defined, they will write all in the same ptp RRD file.

docs/modules/reference/pages/service-assurance/monitors/PtpMonitor.adoc
```diff
@@ -0,0 +1,60 @@
+
+= PtpMonitor
+:description: Learn how to configure and use the PtpMonitor in OpenNMS {page-component-title} to monitor the correct PTP port state for an interface.
+
+The Precision Time Protocol allows the precise synchronization of clocks in measurement and control systems implemented with packet-based networks.
+OpenNMS can monitor the port state of SNMP-enabled interfaces by using the PTP MIB defined in RFC 8173.
```
indigo423, line 6:
```suggestion
OpenNMS can monitor the port state of SNMP-enabled interfaces by using the https://www.rfc-editor.org/info/rfc8173[PTP MIB defined in RFC 8173].
```

indigo423: Reducing the CO2 footprint by not having someone use Google to find the RFC :)

indigo423 (approved):
Minor changes, nothing blocking.

docs/modules/reference/pages/service-assurance/monitors/PtpMonitor.adoc
indigo423, line 6:
```suggestion
The PTP monitor allows you to test the port state of SNMP-enabled interfaces using the PTP MIB defined in RFC 8173.
```

indigo423: Just to remove the problem with "OpenNMS" as a project, vs. products such as Horizon, Meridian, ${powered-by-whatever}.

indigo423 (commented):
Just a hint

docs/modules/reference/pages/service-assurance/monitors/PtpMonitor.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+Example uses CentOS/RHEL path name.
+For Debian/Ubuntu, use `/var/lib/opennms/rrd/response`.
+
+Note that you must include the `monitor` section for each service in your definition.
+
+[source, xml]
+----
+ <service name="PTP" interval="300000" user-defined="false" status="on">
+     <pattern><![CDATA[^PTP-(?<clockPort>.*)-(?<portState>.*)$]]></pattern>
```
indigo423, line 50: Just a personal opinion, I would prefer to let people use MetaData key:value pairs as configuration parameters for the monitor instead of deriving it from the service monitor name. It overloads the service name and can lead to a lot of problems. Renaming a service has side effects that aren't transparent and can break things. MetaData on the service is a much cleaner way to deal with service-specific configurations. Using a pattern match to apply the service configuration to a given set of service names would make sense to me.

indigo423 (commented):
Opinions :)

docs/modules/reference/pages/service-assurance/monitors/PtpMonitor.adoc
christianpape: done.

docs/modules/reference/pages/service-assurance/monitors/PtpMonitor.adoc
christianpape: done.

docs/modules/reference/pages/service-assurance/monitors/PtpMonitor.adoc
indigo423: > We can probably eliminate the problem with ${install.share.dir} removing this configuration parameter. I think there is a default set in the poller we can rely on.

Can we remove this one here as well and rely on the default the monitor code which will always match the install.dir anyway?

indigo423: > We can probably eliminate the problem with ${install.share.dir} removing this configuration parameter. I think there is a default set in the poller we can rely on.

@christianpape Can we remove this one here as well and rely on the default monitor code which will always match the ${install.share.dir} anyway?

indigo423: > We can probably eliminate the problem with ${install.share.dir} removing this configuration parameter. I think there is a default set in the poller we can rely on.

@christianpape Can we remove this one here as well and rely on the default monitor code which will always match the ${install.share.dir} anyway?

docs/modules/reference/pages/service-assurance/monitors/PtpMonitor.adoc
indigo423: Yes I know, in my opinion suggesting all possibilities on how you can inject configuration parameters is a bit harmful in this example. People copy and paste this as is and cause more trouble to start with. I suggest making a pattern match just for the service name and removing the regexp groups deriving service config parameters from the service name. Let that be an advanced option they have to pick when users really know what they are doing.

docs/modules/reference/pages/service-assurance/monitors/PtpMonitor.adoc
christianpape: done.

docs/modules/reference/pages/service-assurance/monitors/PtpMonitor.adoc
```diff
@@ -0,0 +1,66 @@
...
+| n/a
+
+|===
+
+This monitor implements the <<reference:service-assurance/introduction.adoc#ref-service-assurance-monitors-common-parameters, Common Configuration Parameters>>.
+
+== Examples
+
+Example uses CentOS/RHEL path name.
+For Debian/Ubuntu, use `/var/lib/opennms/rrd/response`.
```
indigo423, line 54:
```suggestion

```

docs/modules/reference/pages/service-assurance/monitors/PtpMonitor.adoc
```diff
@@ -0,0 +1,66 @@
...
+| The desired state that the port must be in.
+| n/a
+
+|===
+
+This monitor implements the <<reference:service-assurance/introduction.adoc#ref-service-assurance-monitors-common-parameters, Common Configuration Parameters>>.
+
+== Examples
+
+Example uses CentOS/RHEL path name.
```
indigo423, line 53:
```suggestion

```

indigo423 (changes requested):

docs/modules/reference/pages/service-assurance/monitors/PtpMonitor.adoc
christianpape: done.

indigo423 (commented):
🦄Tears of joy 🌈🎊🏆

indigo423 (approved):
🏆🦄🌈 Tears of joy

fooker (approved):

PR #7554, opened 2024-12-05
NMS-16943 : De-Sync Bug between Web UI and SCVCLI Command.

Junaid-Khan-Nant:
- https://opennms.atlassian.net/browse/NMS-16943

### DIFF
--- a/features/scv/jceks-impl/src/main/java/org/opennms/features/scv/jceks/JCEKSSecureCredentialsVault.java
+++ b/features/scv/jceks-impl/src/main/java/org/opennms/features/scv/jceks/JCEKSSecureCredentialsVault.java
@@ -136,6 +136,7 @@ public class JCEKSSecureCredentialsVault implements SecureCredentialsVault {
     @Override
     public Credentials getCredentials(String alias) {
         loadCredentials();
+        reMapCredentials();
         synchronized (m_credentialsCache) {
             return m_credentialsCache.get(alias);
         }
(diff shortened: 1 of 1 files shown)


Junaid-Khan-Nant:
Screenshot from 2024-12-04 22-45-32
Screenshot from 2024-12-04 22-45-24
Screenshot from 2024-12-04 22-45-16

cgorantla (commented):
Don't think these changes are solving the problem properly.

Looks like this problem occurs because cache is invalid when we add credentials form scvcli.

Two ways to solve this issue:

1. We may have to watch for file updates with `FileReloadContainer` and reload the cache  when the keystore file gets updated. 

2. Add a button to reload cache in UI and implement in Rest api and service.

I would prefer to try with 1  option.

cgorantla:
Or may be try using `FileUpdateWatcher` to reload the cache.

cgorantla (changes requested):
proposed couple of solutions to solve this properly.

features/scv/jceks-impl/src/main/java/org/opennms/features/scv/jceks/JCEKSSecureCredentialsVault.java
```diff
@@ -231,4 +233,30 @@ private static String getKeystorePassword() {
...
+        try {
+            KeyStore.PasswordProtection keyStorePP = new KeyStore.PasswordProtection(m_password);
+            SecretKeyFactory factory = SecretKeyFactory.getInstance("PBE");
+            for (String aliass : getAliases()) {
+                KeyStore.SecretKeyEntry ske = (KeyStore.SecretKeyEntry) m_keystore.getEntry(aliass, keyStorePP);
+                if (ske == null) {
+                    continue;
+                }
+                PBEKeySpec keySpec = (PBEKeySpec) factory.getKeySpec(ske.getSecretKey(), PBEKeySpec.class);
+                m_credentialsCache.put(aliass, fromBase64EncodedByteArray(new String(keySpec.getPassword()).getBytes()));
```
aurang-zeb313, line 247: synchronize the block

cgorantla:
Should be handled by https://github.com/OpenNMS/opennms/pull/7566

PR #7558, opened 2024-12-10
NMS-16543:  Fix ipAddress from Traps when using  use-address-from-varbind

cgorantla:
Backports #7507  NMS-16543: With use-address-from-varbind traps are misassigned

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16543


mershad-manesh:
We should bump the minor version up once it's merged.

mershad-manesh (approved):

PR #7564, opened 2024-12-17
NMS-16943 : De-Sync Bug between Web UI and SCVCLI Command. 

Junaid-Khan-Nant:
-  https://opennms.atlassian.net/browse/NMS-16943

### DIFF
--- a/features/scv/jceks-impl/pom.xml
+++ b/features/scv/jceks-impl/pom.xml
@@ -51,5 +51,9 @@
             <artifactId>junit</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.opennms.core</groupId>
+            <artifactId>org.opennms.core.lib</artifactId>
+        </dependency>
     </dependencies>
 </project>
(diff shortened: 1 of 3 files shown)


Junaid-Khan-Nant:
Whenever I try to run the code with FileUpdateWatcher, it does not sync the credentials for both scvcli and shell, which was the issue reported. I attempted to place the code inside the constructor, but it appears that the issue is not specific to that location. I observed the same behavior when placing it in a different method, such as loadCredential().

I am not entirely sure, but I suspect that the FileUpdateWatcher may not be triggering the file update event, possibly due to the .jce extension or some other factor.

I have attached screenshots as evidence to help with further investigation. Additionally, an exception is thrown when adding multiple credentials simultaneously for scvcli and shell, which causes an issue in Karaf.

The same behavior occurs when invoking the reload method of FileUpdateWatcher forcefully.
Screenshot from 2024-12-13 18-30-57
Screenshot from 2024-12-17 16-03-14

cgorantla:
Should be handled by https://github.com/OpenNMS/opennms/pull/7566

PR #7568, opened 2024-12-20
NMS-16991: Add grpc exporter that sends monitored service status

cgorantla:
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16991


indigo423:
@cgorantla I think this is a good idea. Just managing the API compatibility along with the release of Horizon/Meridian. If we stay compatible with the semver.org versioning schema we should be fine and can reduce some management and release overhead.

cgorantla:
> @cgorantla I think this is a good idea. Just managing the API compatibility along with the release of Horizon/Meridian. If we stay compatible with the semver.org versioning schema we should be fine and can reduce some management and release overhead.

Yup. we quickly learned that this doesn't need to be separate plugin. Yeah, we have to reduce some overhead on release and repo management for lot of these plugins.

aurang-zeb313 (dismissed):
LGTM

synqotik (approved):
LGTM!

PR #7590, opened 2025-01-10
NMS-17003: Add tenant id on grpc exporter 

cgorantla:
Also simplify host ( remove port from config)

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-17003

### DIFF
--- a/docs/modules/operation/pages/deep-dive/grpc-exporter/grpc-exporter.adoc
+++ b/docs/modules/operation/pages/deep-dive/grpc-exporter/grpc-exporter.adoc
@@ -13,17 +13,20 @@ See `monitored-services.proto` in the corresponding source distribution for the 
 $ ssh -p 8101 admin@localhost
 ...
 admin@opennms()> config:edit org.opennms.features.grpc.exporter
-admin@opennms()> config:property-set host bsm.onmshs.local:1440 <1>
-admin@opennms()> config:property-set tls.cert.path /opt/opennms/etc/tls.cert <2>
-admin@opennms()> config:property-set tls.enabled false <3>
-admin@opennms()> config:property-set snapshot.interval 3600 false <4>
+admin@opennms()> config:property-set host bsm.onmshs.local:1443 <1>
+admin@opennms()> config:property-set tenant.id opennms-prime <2>
+admin@opennms()> config:property-set tls.cert.path /opt/opennms/etc/tls.cert <3>
+admin@opennms()> config:property-set tls.enabled false <4>
+admin@opennms()> config:property-set snapshot.interval 3600 false <5>
 admin@opennms()> config:update
 ----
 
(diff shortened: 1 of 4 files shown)


aurang-zeb313 (commented):
Can we make HeaderAttachingClientCall static?

aurang-zeb313:
> > Can we make HeaderAttachingClientCall static?
> 
> This does use instance data, so it needs to be non static ` headers.merge(metadata);` we can make it static if needed, what is the benefit of doing this ?
If we make HeaderAttachingClientCall static, it will no longer implicitly holds a reference to the instance of outer class. By dependency injection we can pass metadata. Loose coupling between the classes. Its my suggestion,  so if you think no need for this change we can merge it.

docs/modules/operation/pages/deep-dive/grpc-exporter/grpc-exporter.adoc
```diff
@@ -13,17 +13,20 @@ See `monitored-services.proto` in the corresponding source distribution for the
 $ ssh -p 8101 admin@localhost
 ...
 admin@opennms()> config:edit org.opennms.features.grpc.exporter
-admin@opennms()> config:property-set host bsm.onmshs.local:1440 <1>
-admin@opennms()> config:property-set tls.cert.path /opt/opennms/etc/tls.cert <2>
-admin@opennms()> config:property-set tls.enabled false <3>
-admin@opennms()> config:property-set snapshot.interval 3600 false <4>
+admin@opennms()> config:property-set host bsm.onmshs.local:1443 <1>
```
synqotik, line 16: I think this should be:

```
admin@opennms()> config:edit org.opennms.features.grpc.exporter
admin@opennms()> config:property-set host bsm.onmshs.local <1>
admin@opennms()> config:property-set port 1443 <1>
admin@opennms()> config:property-set tenant.id opennms-prime <2>
admin@opennms()> config:property-set tls.cert.path /opt/opennms/etc/tls.cert <3>
admin@opennms()> config:property-set tls.enabled false <4>
admin@opennms()> config:property-set snapshot.interval 3600 <5>
admin@opennms()> config:update
```

synqotik: never mind - looks like host:port is ok now.

Just check re that `false` in `snapshot.interval`, wasn't sure if a 2nd argument is valid.

cgorantla: good catch

aurang-zeb313 (approved):
LGTM

cgorantla:
we need to add some tests for grpc exporter. Created https://opennms.atlassian.net/browse/NMS-17064 to address this

PR #7596, opened 2025-01-16
NMS-16573 User tries to create an Alarm filter favorite, but the filter doesn't save the arguments of the filter

jahanzebdataq:
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16573


opennms-webapp/src/main/java/org/opennms/web/event/EventUtil.java
```diff
@@ -92,6 +93,16 @@ public abstract class EventUtil {
     /** Constant <code>ANY_OPTION="Any"</code> */
     public static final String ANY_OPTION = "Any";
 
+    /**
+     * Checks if a string represents an integer.
+     *
+     * @param str The string to check.
+     * @return True if the string is an integer, false otherwise.
+     */
+    public static boolean isInteger(String str) {
```
smunir-onms, line 102: Suggestion: This function seems repetitive. Can you please check if a common utility exists in opennms-webapp to accommodate this?

jahanzebdataq: I noticed duplicate code in Alarms and Event Utils too. This can be added as a TODO task for future sprints to refactor this into Generics which can handle both.

opennms-webapp/src/main/java/org/opennms/web/alarm/AlarmUtil.java
```diff
@@ -247,6 +258,30 @@ public static Filter getFilter(String[] allFilters, String filterString, Servlet
             }
         } else if (type.equals(InterfaceFilter.TYPE)) {
             filter = new InterfaceFilter(InetAddressUtils.addr(value));
+        } else if (type.equals(ConditionalFilter.TYPE)) {
+            String cleanedValue = value.substring(1, value.length() - 1); // Remove surrounding brackets
```
cgorantla, line 262: Can we check if there are surrounding brackets exists then only clean it up as the behavior can change.

cgorantla (commented):
I haven't tried to reproduce this issue, but if you tested this with the scenario mentioned in the bug, looks fine to me.

opennms-webapp/src/main/java/org/opennms/web/alarm/AlarmUtil.java
jahanzebdataq: image
ConditionalFilter always has a brackets even if their is only one parameter, because it stores data in an array. Do you still want me check?

cgorantla (approved):
LGTM!

PR #7598, opened 2025-01-17
FE: NMS-17004: Update the login events item in Usage Stats, add link to download CSV file

Shahbaz-dataq:
### Description

A new item, “CSV output of Login events for the past 60 days”, has been added to the usage stats list.

The “Latest Value” column will contain Base64-encoded file contents.

Update the column to display a download link labeled “Download CSV file”. The link should decode the Base64 content and trigger the file download.

The implementation should be generic, based on the datatype value (file|csv) provided in the metadata.

If the datatype starts with "file", split it using | to extract the file type (e.g., csv in this case) and handle the file appropriately.

Metadata:

{
    "key": "loginsPast60Days",
    "name": "CSV output of Login events for the past 60 days",
    "description": "CSV output containing the date, username login events for the past 60 days",
    "datatype": "file|csv"
}

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-17004

### DIFF
--- a/ui/src/components/UsageStatistics/UsageStatisticsTable.vue
+++ b/ui/src/components/UsageStatistics/UsageStatisticsTable.vue
@@ -52,6 +52,9 @@
             <td v-if="row.isLink">
               <a href="#" @click.prevent="() => showFullValue(row)">See full value</a>
             </td>
+            <td v-else-if="row.isFile">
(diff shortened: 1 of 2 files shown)


synqotik (approved):
LGTM!

PR #7620, opened 2025-02-07
NMS-17083: Investigate Develop branch failure when using OIA 2.0

mershad-manesh:
Enable OIA 2.0
Update Sentinel and Minion so they can build (fix dependency) with newer version of OIA

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-17083

### DIFF
--- a/features/container/minion/pom.xml
+++ b/features/container/minion/pom.xml
@@ -84,13 +84,13 @@
     </build>
 
     <dependencies>
-        <dependency>
+        <!--dependency>
           <groupId>org.opennms.integration.api.sample</groupId>
           <artifactId>sample-project</artifactId>
           <version>${opennmsApiVersion}</version>
           <type>xml</type>
           <classifier>blueprint-minion</classifier>
-        </dependency>
+        </dependency-->
         <dependency>
             <groupId>org.opennms.container</groupId>
             <artifactId>org.opennms.container.shared</artifactId>
--- a/features/container/minion/pom.xml
+++ b/features/container/minion/pom.xml
@@ -84,13 +84,25 @@
     </build>
 
     <dependencies>
-        <!--dependency>
+        <dependency>
+          <groupId>org.opennms.dependencies</groupId>
+          <artifactId>jaxb-dependencies</artifactId>
+          <version>${project.version}</version>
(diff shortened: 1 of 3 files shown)


indigo423:
How you added the dependencies doesn't seem right. Here is my suggestion to solve this issue: 

* Step 1 Make a release of OIA 2.0.0 and publish it to maven central to make it publicly available, 
* Step 2: Just bump the OIA dependency from 1.6.1 to 2.0.0

The build in CircleCI will be a bit longer, because the Maven cache is invalidated, you can update the cache later to improve the build speed.

Additional Information: As you bump a major release version regarding to https://semver.org/ we break API here. This has some significant consequences, you need to rebuild all published plugins using the OIA 2.0. This means, you have now plugins floating around which are using OIA 1.x which work with Meridian 2022-2025 and Horizon 33.x. Everything beyond Horizon 34.x and later require OIA 2.0 compatible plugins. I'm not sure if you have all the processes in place to deal with that situation accordingly.

indigo423:
@cgorantla Just some food for thought from a discussion I had with about this topic @christianpape. It would be good if could find a way to release 1.7.0 instead of 2.0.0, to make it backward compatible to Meridian 2022 - 2024 and future Horizon 34 above.

As we have increased the data type from "int" to "long", the "int" from Meridian 2022 - 2024 fits into a long without breaking the behavior.

christianpape (approved):
LGTM.

cgorantla (changes requested):
Please see my last comment.

mershad-manesh:
Closed this issue as it will be addressed in [MPLUG-70](https://opennms.atlassian.net/browse/MPLUG-70)

indigo423:
> Created https://opennms.atlassian.net/browse/MPLUG-70 to back port Event Id related changes to OIA 1.7.0 So we might not need this PR once we handle this.

@cgorantla I would suggest to put this issue in the OIA project and make sure this is publicly available. We should maintain Plugin API related topics in public, because this is what everyone would use to create a plugin. The MPLUG repository is not accessible and is tracking plugins realated to Meridian only.

PR #7637, opened 2025-02-21
NMS-17070  Log messages from Groovy are truncated

Mohammad-Junid:
Messages written to the Karaf log from a telemetry adapter groovy script appear to be truncated. How can I see the entire log message?

* Jira https://opennms.atlassian.net/browse/NMS-17070

### DIFF
--- a/docs/modules/reference/pages/telemetryd/protocols/jti.adoc
+++ b/docs/modules/reference/pages/telemetryd/protocols/jti.adoc
@@ -133,3 +133,30 @@ To increase the character size, modify the `%.-500m` part of the pattern. For ex
 ----
 log4j2.pattern = %d{ISO8601} | %-5p | %-16t | %-32c{1} | %X{bundle.id} - %X{bundle.name} - %X{bundle.version} | %encode{%.-1000m}{CRLF}%n
 ----
+
+===  Using Karaf Shell
+You can also modify the `org.ops4j.pax.logging` configuration dynamically using the Karaf shell
+
+Log in to the Karaf console in OpenNMS
+[source, console]
+----
+ssh -p 8101 admin@localhost
+----
+Use the `config:edit` command to start editing the `org.ops4j.pax.logging` configuration
+[source, console]
+----
+config:edit org.ops4j.pax.logging
+----
+Use the `config:property-set` command to update the `log4j2.pattern` property. For example, to increase the log message size to 1000 characters
+[source, console]
+----
+config:property-set log4j2.pattern '%d{ISO8601} | %-5p | %-16t | %-32c{1} | %X{bundle.id} - %X{bundle.name} - %X{bundle.version} | %encode{%.-1000m}{CRLF}%n'
+config:update
+----
+
+To verify that the changes have been applied, use the below command:
+[source, console]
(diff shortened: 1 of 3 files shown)


cgorantla (commented):
As this can be updated runtime, this can be a doc update [here](https://github.com/OpenNMS/opennms/blob/develop/docs/modules/reference/pages/telemetryd/protocols/jti.adoc) instead of updating directly here.

We should be able to update cfg file through  `config:edit org.ops4j.pax.logging` pattern.

Please validate this through config update.

dino2gnt:
> As this can be updated runtime, this can be a doc update [here](https://github.com/OpenNMS/opennms/blob/develop/docs/modules/reference/pages/telemetryd/protocols/jti.adoc) instead of updating directly here.
> 
> We should be able to update cfg file through `config:edit org.ops4j.pax.logging` pattern.
> 
> Please validate this through config update.

Why not BOTH?

christianpape (approved):
LGTM.

cgorantla:
> > As this can be updated runtime, this can be a doc update [here](https://github.com/OpenNMS/opennms/blob/develop/docs/modules/reference/pages/telemetryd/protocols/jti.adoc) instead of updating directly here.
> > We should be able to update cfg file through `config:edit org.ops4j.pax.logging` pattern.
> > Please validate this through config update.
> 
> Why not BOTH?

This file is coming from Karaf, don't want to have a custom change which needs to be taken for every karaf upgrade. Don't need to change this unless it is really a most.

For this case, we are really not sure if 2000 is sufficient or not so even if we update this, won't take care of all cases.
Good enough to have a doc update.

cgorantla (changes requested):
Only needs a doc update.

docs/modules/reference/pages/telemetryd/protocols/jti.adoc
```diff
@@ -116,3 +116,47 @@ The following globals will be passed to the script:
 | Decoded JTI message from which the metrics should be extracted.
 | org.opennms.netmgt.telemetry.adapters.jti.proto.TelemetryTop
 |===
+
+NOTE: In OpenNMS, logs are essential for monitoring and troubleshooting. By default, the log message size is limited to 500 characters. However, in some cases, such as when viewing detailed proto object logs, you may need to increase the character size to capture the full log details.
```
cgorantla, line 120:
```suggestion

```

docs/modules/reference/pages/telemetryd/protocols/jti.adoc
```diff
@@ -116,3 +116,47 @@ The following globals will be passed to the script:
...
+
+=== Steps to Increase Log Character Size
+Locate file `org.ops4j.pax.logging.cfg`  in the `${OPENNMS_HOME}/etc/` directory.
```
cgorantla, line 123: Insert following statement after ===Steps to Increase   and remove above note.

By default, the log message size is limited to 500 characters with karaf logs. However, in some cases, such as when viewing detailed proto object logs, you may need to increase the character size to capture the full log details.

cgorantla (approved):
LGTM!

PR #7646, opened 2025-03-04
NMS-16537: Forbid invalid requisition names

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-16537

### DIFF
--- a/opennms-provision/opennms-provision-persistence/src/main/java/org/opennms/netmgt/provision/persist/requisition/Requisition.java
+++ b/opennms-provision/opennms-provision-persistence/src/main/java/org/opennms/netmgt/provision/persist/requisition/Requisition.java
@@ -54,6 +54,7 @@ import javax.xml.datatype.DatatypeConfigurationException;
 import javax.xml.datatype.DatatypeFactory;
 import javax.xml.datatype.XMLGregorianCalendar;
 
+import com.google.common.collect.Sets;
 import org.apache.commons.lang.builder.CompareToBuilder;
 import org.opennms.core.xml.ValidateUsing;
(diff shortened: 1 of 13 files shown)


opennms-provision/opennms-provision-persistence/src/main/java/org/opennms/netmgt/provision/persist/requisition/Requisition.java
```diff
@@ -464,9 +466,12 @@ public void validate() throws ValidationException {
     	if (m_foreignSource == null) {
     	    throw new ValidationException("Requisition 'foreign-source' must be set!");
     	}
-    	if (m_foreignSource.contains("/")) {
-            throw new ValidationException("Foreign Source (" + m_foreignSource + ") contains invalid characters. ('/' is forbidden.)");
-    	}
+
+        for(String illegalCharacter : ILLEGAL_CHARACTERS) {
```
cgorantla, line 470: If there are requisitions already there with `imported:` as foreign source, they will not be processed now, right ?
I think we prevent requisitions with `:` in their name from web ui, but not from xml. 

May be we should add this as a breaking change. wdyt ?

cgorantla (commented):
LGTM, one comment.

opennms-provision/opennms-provision-persistence/src/main/java/org/opennms/netmgt/provision/persist/requisition/Requisition.java
christianpape: Sure.

cgorantla (approved):
LGTM!

PR #7659, opened 2025-03-13
NMS-14199: Change status timestamp if event time is set

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-14199


synqotik (approved):
LGTM!

PR #7660, opened 2025-03-13
NMS-15517: Use ReST-call to delete scheduled outage

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-15517


synqotik (approved):
LGTM!

PR #7662, opened 2025-03-13
NMS-17731: Zenith Connect UI Proof of Concept - Meridian side

synqotik:
Zenith Connect UI proof of concept (POC) - Meridian side.

Screens to connect Meridian to Zenith in order to get Keycloak access/refresh tokens which will be used by gRPC to authenticate while sending data to Zenith.

This is a POC only, but we will build on it for a full implementation. It is disabled by default, see `opennms.properties`, `opennms.zenithConnect.enabled` property.

Needed to update the `NavBarMenuEntry` code to add ability to check a system property (i.e. from the `opennms.properties` file) to determine whether to display a menu entry, or to display the "Connect to Zenith" on the Admin page.

The Menu Rest API returns whether Zenith Connect is enabled, and the base/relative URLs of the Zenith installation, which are configured in `opennms.properties`.

This doesn't get the Meridian system ID and doesn't save the registrations, that will be implemented later.

Also a fix to silence SCSS `@import` deprecation warnings, eventually we should use `@use` instead of `@import`. Also updated Feather to the latest.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-17731

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/opennms.properties
+++ b/opennms-base-assembly/src/main/filtered/etc/opennms.properties
@@ -560,16 +560,6 @@ opennms.newsFeedPanel.show=true
 #
 #opennms.newsFeedPanel.url=https://www.opennms.com/feed/
 
-# This value enables or disables display of the Zenith Connect which allows
-# you to register OpenNMS with a Zenith instance in order to send data
-opennms.zenithConnect.enabled=false
-
-# This value is the base URL of the Zenith instance for the Zenith Connect feature
-opennms.zenithConnect.zenithBaseUrl=https://onmshs.local:1443
-
-# This value is the relative URL (from the zenithBaseUrl) of the Zenith instance for the Zenith Connect feature
-opennms.zenithConnect.zenithRelativeUrl=/zenith-connect
-
 # This value enables or disables display of the Product Update Enrollment popup.
 # Set to false to disable showing the popup and to prevent any data being sent.
 # If commented-out or unset, the default value is true.


opennms-base-assembly/src/main/filtered/etc/opennms.properties
```diff
@@ -560,6 +560,16 @@ opennms.newsFeedPanel.show=true
 #
 #opennms.newsFeedPanel.url=https://www.opennms.com/feed/
 
+# This value enables or disables display of the Zenith Connect which allows
```
cgorantla, line 563: We stopped putting any new properties in opennms.properties file because it results in upgrade issues.

Defaults can be in the code. Any overrides needs to be in opennms.properties.d/ at run time.

synqotik: Redirect back into Meridian still doesn't work. There are issues with retaining the query string past login, and also redirect seems to log the user out.

After registration, the user will get a Zenith screen where they can copy the token. That screen also has a link back to Meridian which the user could paste into the location bar of an already active Meridian session, which will go to the Meridian `zenith-connect-success` endpoint which could be used to save the token.

synqotik: I removed the changes from `opennms.properties` file. I will open a separate issue to add documentation, and also to fix an issue where Connect to Zenith is still showing up in the non-Vue menu.

cgorantla (approved):
LGTM!

PR #7671, opened 2025-03-24
NMS-17766: Do not display menu item if property check fails

synqotik:
Update the `LocationBasedNavBarEntry` mechanism to not display the menu item at all if a system property check fails.

This means that the "Zenith Connect" menu entry should not appear in the legacy menu unless the feature has been enabled in the properties file.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-17766

### DIFF
--- a/opennms-webapp-rest/src/test/java/org/opennms/web/rest/support/menu/MenuProviderTest.java
+++ b/opennms-webapp-rest/src/test/java/org/opennms/web/rest/support/menu/MenuProviderTest.java
@@ -29,14 +29,15 @@ import java.util.Optional;
 
 import com.google.common.base.Strings;
 import org.junit.Assert;
+
+import static org.hamcrest.Matchers.*;
 import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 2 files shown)


opennms-webapp-rest/src/test/java/org/opennms/web/rest/support/menu/MenuProviderTest.java
```diff
@@ -29,14 +29,15 @@
 
 import com.google.common.base.Strings;
 import org.junit.Assert;
+
+import static org.hamcrest.Matchers.*;
```
cgorantla, line 33: nit: It is a test but still want to avoid wildcard imports.

synqotik: Fixed - I need to check my IntelliJ settings :)

cgorantla (approved):
LGTM!

PR #7683, opened 2025-04-01
NMS-17751: Monitoring System Rest API (foundation-2024)

synqotik:
We want to get the Meridian system ID to use in Zenith Connect to uniquely identify the system that is registering with Zenith to send data. The system ID is the MonitoringSystem id field for the `OpenNMS` `type`, a uuid.

We are providing a Rest API to get the "main" OpenNMS system.

The front end will be implemented separately.

Had to do some additional work to optionally have the `DatabasePopulator` create the original `DistPoller` (which is the `OnmsMonitoringSystem`) with a specified ID for these integration tests only. By default, the id is a uuid with all 0's, which doesn't matter for all the other integration tests.

PR includes documentation and integration tests.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-17751

### DIFF
--- a/docs/modules/development/nav.adoc
+++ b/docs/modules/development/nav.adoc
@@ -46,6 +46,7 @@
 *** xref:rest/maps.adoc[]
 *** xref:rest/measurements.adoc[]
 *** xref:rest/meta-data.adoc[]
+*** xref:rest/monitoring_systems.adoc[]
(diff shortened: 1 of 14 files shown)


cgorantla (approved):
LGTM!

PR #7686, opened 2025-04-02
NMS-15676: Detect whether OpenNMS is running inside container

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-15676

I read a lot about different cloud container engine and there are not really good ways to detect whether we are running inside a container. So, I choosed a simple but bulletproof method that at least will work in every case for the containers that are built by us. The tests regarding docker and podman are still in place, so user-built containers are also detected for these two platforms.

### DIFF
--- a/opennms-container/core/Dockerfile
+++ b/opennms-container/core/Dockerfile
@@ -126,6 +126,8 @@ COPY --chown=10001:0 ./container-fs/health.sh /
 # Allow users in the root group to access them with the same authorization as the directory and file owner
 RUN chmod -R g=u /usr/share/opennms/etc
 RUN find /usr/share/opennms/data /usr/share/opennms/system /opennms-data -type d -print0 | xargs -0 chmod g=u
+# Allow OpenNMS to detect whether it is run in a container
+RUN touch /CONTAINER_RUNNING
 
 # Arguments for labels should not invalidate caches
 ARG VERSION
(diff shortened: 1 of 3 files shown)


mershad-manesh:
Instead of using a file we can use an environment variable (This value could be overridden by the user, if they want to).

For example `OPENNMS_EXECUTION_ENVIRONMENT=container`

Shouldn't we set the same properties in Minion and Sentinel?

christianpape:
> Instead of using a file we can use an environment variable (This value could be overridden by the user, if they want to).
> 
> For example `OPENNMS_EXECUTION_ENVIRONMENT=container`
> 
> Shouldn't we set the same properties in Minion and Sentinel?

Of course, I changed it to an environment variable.

opennms-container/core/Dockerfile
```diff
@@ -126,6 +126,8 @@ COPY --chown=10001:0 ./container-fs/health.sh /
 # Allow users in the root group to access them with the same authorization as the directory and file owner
 RUN chmod -R g=u /usr/share/opennms/etc
 RUN find /usr/share/opennms/data /usr/share/opennms/system /opennms-data -type d -print0 | xargs -0 chmod g=u
+# Allow OpenNMS to detect whether it is run in a container
+RUN touch /CONTAINER_RUNNING
```
mershad-manesh, line 130: Do we still need this line?

christianpape: Yes, I missed that.

mershad-manesh (approved):

PR #7689, opened 2025-04-03
NMS-17767: Use systemId in Zenith Connect UI

synqotik:
We added a Rest API to get the main monitoring system info. This calls that API to get the Meridian `systemId` and also the `label` to populate fields used to register the Meridian instance with Zenith.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-17767


Shahbaz-dataq (approved):
LGTM

PR #7695, opened 2025-04-07
NMS-17756: Modified entrypoint.sh and disabled ActiveMQ in Kafka Smoke Tests

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-17756

I have now adapted the entrypoint.sh script. It now allows the configuration of Kafka via the IPC parameter (KAFKA_IPC_BOOTSTRAP_SERVERS) as well as via the three individual features Sink, RPC and Twin (KAFKA_SINK_BOOTSTRAP_SERVERS, KAFKA_RPC_BOOTSTRAP_SERVERS and KAFKA_TWIN_BOOTSTRAP_SERVERS). In other words, as described in the installation instructions. ActiveMQ should now be deactivated in both cases.

### DIFF
--- a/opennms-container/minion/container-fs/entrypoint.sh
+++ b/opennms-container/minion/container-fs/entrypoint.sh
@@ -95,6 +95,12 @@ function updateConfig() {
 function parseEnvironment() {
     # Configure additional features
     IFS=$'\n'
+
+    IPC_CONFIGURED="false"
+    TWIN_CONFIGURED="false"
+    RPC_CONFIGURED="false"
+    SINK_CONFIGURED="false"
+
     for VAR in $(env)
     do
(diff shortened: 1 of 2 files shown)


indigo423 (commented):
Why are we implementing a second logic in the entrypoint? All this logic exists already and even more in the confd templates? We are adding now unnecessary logic we need to maintain additionally.

christianpape:
> Why are we implementing a second logic in the entrypoint? All this logic exists already and even more in the confd templates? We are adding now unnecessary logic we need to maintain additionally.

I just fixed the existing logic to work properly. I did not add anything new.

indigo423 (approved):
LGTM!

cgorantla (approved):
LGTM!

PR #7701, opened 2025-04-14
NMS-17732:   Resolve Twin gRPC issues

cgorantla:
1. Revert previous grpc upgrades and upgrade grpc to newer version.
2. Restructure `org.opennms.core.ipc.twin.common` package so that it can be removed from custom.properties
3. Add `ServicesResourceTransformer` with the shaded plugin so that it can get dependent services
4. Remove `metricsRegistry` and `tracerRegistry` dependencies from grpc twin module 
( these are creating class loader issues, no easy way to solve)

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-17732


core/ipc/twin/common/src/main/java/org/opennms/core/ipc/twin/common/AbstractTwinPublisher.java
```diff
@@ -79,14 +84,29 @@ public abstract class AbstractTwinPublisher implements TwinPublisher {
 
     public AbstractTwinPublisher(LocalTwinSubscriber localTwinSubscriber, TracerRegistry tracerRegistry, MetricRegistry metricRegistry) {
         this.localTwinSubscriber = Objects.requireNonNull(localTwinSubscriber);
-        Objects.requireNonNull(tracerRegistry);
-        tracerRegistry.init(SystemInfoUtils.getInstanceId());
-        this.tracer = tracerRegistry.getTracer();
+        if ( tracerRegistry != null) {
```
christianpape, line 87:
```suggestion
        if (tracerRegistry != null) {
```

christianpape (approved):
Awesome work! LGTM.

PR #7709, opened 2025-04-17
NMS-17876: Limit the columns for orderBy clause

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-17876

Unfortunately, it is not possible to pass the columns for the OrderBy-clause in a prepared statement so that they are correctly escaped. This is because they are not string literals or numbers but names of table columns. We only want to allow the columns of the node entity and also the calculated severity. Unfortunately, not all property names in our node entity match the database columns (e.g. label vs. nodeLabel). Therefore, I had to do a little trickery to determine all valid column names programmatically.


synqotik (approved):
LGTM!

cgorantla (approved):
LGTM!

PR #7711, opened 2025-04-17
NMS-17873: Fix IPAddress parsing for switches that represent IPAddress as octet string

cgorantla:
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-17873


cgorantla:
This is mostly @roskens contribution from https://opennms.atlassian.net/browse/NMS-6452.

marshallmassengill:
I've asked support folks and @indigo423 to look at this and review as well.  It seems reasonable.  I'm curious if this might also resolve the issue mentioned here: https://opennms.discourse.group/t/wrong-discovery-of-ipaddresses/3949/32 or if some other fix is needed to handle that edge case as well.

indigo423:
> I've asked support folks and @indigo423 to look at this and review as well. It seems reasonable. I'm curious if this might also resolve the issue mentioned here: https://opennms.discourse.group/t/wrong-discovery-of-ipaddresses/3949/32 or if some other fix is needed to handle that edge case as well.

@marshallmassengill I'll take a look tomorrow. Thanks for the heads up.

synqotik (approved):
LGTM - not sure if any additional checking is needed that the value is a valid IP address, but I assume `InetAddressUtils.getInetAddress` will suffice.

Perhaps add another test (could be separate PR) with some bad values (letters, punctuation, control characters) and ensure it's handled properly.

opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/IPAddressTableTracker.java
```diff
@@ -111,13 +113,32 @@ public String getIpAddress() {
...
             if (addressType == TYPE_IPV4 && instanceIds.length != 6) {
-                LOG.warn("BAD AGENT: Does not conform to RFC 4001 Section 4.1 Table Indexing!!! Report them immediately.  Making a best guess!");
+                LOG.warn("IPV4: BAD AGENT: Does not conform to RFC 4001 Section 4.1 Table Indexing!!! Report them immediately.  Making a best guess!");
+                if (instanceIds.length == (addressLength + addressIndex)) {
+                    // NMS-6452: Some older Brocade Switches represent the IP address as a octet string.
+                    try {
+                        final InetAddress address = byteStringToInetAddress(instanceIds, addressIndex, addressLength);
+                        return str(address);
+                    } catch (Exception e) {
+                        LOG.debug("\"IPV4: BAD AGENT: Could not parse raw oids as octet string", e);
```
christianpape, line 126:
```suggestion
                        LOG.debug("IPV4: BAD AGENT: Could not parse raw oids as octet string", e);
```

christianpape (changes requested):
LGTM. Only one log statement to fix.

marshallmassengill (approved):
LGTM!

cgorantla:
> LGTM - not sure if any additional checking is needed that the value is a valid IP address, but I assume `InetAddressUtils.getInetAddress` will suffice.
> 
> Perhaps add another test (could be separate PR) with some bad values (letters, punctuation, control characters) and ensure it's handled properly.

Yes, `InetAddressUtils.getInetAddress` should take care of any invalid IpAddress.  Since this is very unusual corner case, feels like current test is sufficient.

PR #7714, opened 2025-04-21
NMS-17891: Update docs for Jaeger Tracing

cgorantla:
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-17891

### DIFF
--- a/docs/modules/deployment/pages/opentracing/core.adoc
+++ b/docs/modules/deployment/pages/opentracing/core.adoc
@@ -29,11 +29,11 @@ sudo vi etc/opennms.properties.d/jaeger.properties
 [source, jaeger.properties]
 ----
 org.opennms.core.tracer=jaeger<1>
-JAEGER_AGENT_HOST=my-jaeger-host<2>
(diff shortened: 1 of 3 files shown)


marshallmassengill (commented):
LGTM!  Only thing that might be worth noting somewhere is that JAEGER_ENDPOINT is for TCP and JAEGER_AGENT_HOST is for UDP... though I haven't verified if the UDP connection works - it didn't work in my lab initially so I swapped to TCP but that might have been a config issue.

marshallmassengill (approved):
LGTM!

indigo423 (commented):
LGTM

cgorantla:
> Only thing that might be worth noting somewhere is that JAEGER_ENDPOINT is for TCP and JAEGER_AGENT_HOST is for UDP... though I haven't verified if the UDP connection works - it didn't work in my lab initially so I swapped to TCP but that might have been a config issue.

I think we can go with default option, there is a link in the docs to other JAEGER properties if user wants to configure it differently.

PR #7718, opened 2025-04-25
NMS-17749: Zenith Connect UI, connect with Registration Rest API

synqotik:
Update the Zenith Connect UI to use the Rest API for storing and retrieving registration info.

Update the persistence service to exclude duplicate registrations.

Clean up the UI a bit.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-17749

### DIFF
--- a/features/zenith-connect/persistence/src/main/java/org/opennms/features/zenithconnect/persistence/impl/ZenithConnectPersistenceServiceImpl.java
+++ b/features/zenith-connect/persistence/src/main/java/org/opennms/features/zenithconnect/persistence/impl/ZenithConnectPersistenceServiceImpl.java
@@ -152,23 +152,23 @@ public class ZenithConnectPersistenceServiceImpl implements ZenithConnectPersist
         ZenithConnectRegistrations existingRegistrations = getRegistrationsImpl();
         ZenithConnectRegistration existingRegistration = existingRegistrations.first();
 
-        if (existingRegistration != null) {
-            boolean systemIdsMatch = registration.systemId != null && existingRegistration.systemId != null &&
-                    registration.systemId.equals(existingRegistration.systemId);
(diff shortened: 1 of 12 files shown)


features/zenith-connect/persistence/src/main/java/org/opennms/features/zenithconnect/persistence/impl/ZenithConnectPersistenceServiceImpl.java
```diff
@@ -136,4 +146,31 @@ private void setRegistrations(ZenithConnectRegistrations registrations) throws Z
...
+            throws ZenithConnectPersistenceException {
+        ZenithConnectRegistrations existingRegistrations = getRegistrationsImpl();
+        ZenithConnectRegistration existingRegistration = existingRegistrations.first();
+
+        if (existingRegistration != null) {
+            boolean systemIdsMatch = registration.systemId != null && existingRegistration.systemId != null &&
+                    registration.systemId.equals(existingRegistration.systemId);
+
+            if (systemIdsMatch) {
+                boolean accessTokenMatches = registration.accessToken != null &&
```
aurang-zeb313, line 160: For ZnithConnectRegistrations we are always getting first can you explain please? i don't understand it.
We have to  avoid nested if else if possible.

synqotik: I pushed a change to remove some nesting.

For now, we only ever have just 1 registration. However, in case we add more in the future (maybe user connects with multiple Zenith instances or something else), I made the underlying data structure a list instead of a single object.

Shahbaz-dataq (approved):
UI changes look good.

PR #7719, opened 2025-04-26
NMS-17883: Add DEB/RPM package dependencies and changelog

indigo423:
I have found that we need to adjust the RPM/DEB package dependencies. The package jrrd2 is now not a suggestion and required. The dependency for RRDTool is not required and satisfied by jrrd2 itself, see:
* https://github.com/OpenNMS/jrrd2/blob/f8a29dba62a566d6c2a043c546fe35a7ff0695a8/debian/control#L13
* https://github.com/OpenNMS/jrrd2/blob/f8a29dba62a566d6c2a043c546fe35a7ff0695a8/build-packages.sh#L38

I have also added changelog with some hints on this topic.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-17883

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -56,8 +56,8 @@ Description: Enterprise-grade Open-source Network Management Platform (Database)
 
 Package: opennms-server
 Architecture: all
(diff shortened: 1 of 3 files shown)


christianpape (approved):
LGTM.

PR #7720, opened 2025-04-28
NMS-17899: Remove Java 11 dependencies

indigo423:
Removed RPM/DEB package dependencies for OpenJDK 11. Changed source and compiler target from 11 to 17.

### Additional Reviewer Hint

Haveged is removed as a dependency which is obsolete on modern Kernel 5.6 forward.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-17899

### DIFF
--- a/tools/packages/minion/minion.spec
+++ b/tools/packages/minion/minion.spec
@@ -71,7 +71,6 @@ Requires(pre):  jicmp >= 3.0.0
 Requires:       jicmp6 >= 3.0.0
 Requires(pre):  jicmp6 >= 3.0.0
 Provides:	opennms-plugin-api = %{opa_version}
-Recommends:	haveged
 
 Conflicts:      %{name}-container        < %{version}-%{release}
 Conflicts:      %{name}-features-core    < %{version}-%{release}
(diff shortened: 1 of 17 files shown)


tools/packages/minion/minion.spec
```diff
@@ -71,7 +71,6 @@ Requires(pre):  jicmp >= 3.0.0
 Requires:       jicmp6 >= 3.0.0
 Requires(pre):  jicmp6 >= 3.0.0
 Provides:	opennms-plugin-api = %{opa_version}
-Recommends:	haveged
```
dino2gnt, line 74: As I recall, `haveged`  was added for entropy gathering purposes, as there was some fairly easily reproducible case where something (Karaf?) would block startup in some environments waiting for sufficient entropy to gather, and `haveged` solved that problem.  https://opennms.atlassian.net/browse/NMS-8959

indigo423: See here: https://github.com/jirka-h/haveged

dino2gnt: el9 is on 5.14 and everything older is EOL, so I can't imagine there is any reason to keep it around.

dino2gnt (approved):
LGTM!

christianpape:
Does the build executor in .circleci/executors.yml also have to be set to jdk17?

christianpape (approved):
LGTM.

PR #7721, opened 2025-04-30
NMS-16414: Reducing the risk of XSLT injection

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-16414


synqotik (approved):
LGTM!

PR #7727, opened 2025-05-06
NMS-17009: WS-Man kerberos authentication fails with Java 17 since OpenNMS 33.0.10

Mohammad-Junid:
Updated from 33.0.9 to 33.0.10 - after update WSMan datacollection doesn't succeeds anymore.
Update: seem to be a problem with Horizon 33.0.10 and Java 17 

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-17009

### DIFF
--- a/docs/modules/reference/pages/performance-data-collection/collectors/wsman.adoc
+++ b/docs/modules/reference/pages/performance-data-collection/collectors/wsman.adoc
@@ -347,3 +347,11 @@ Use this to threshold on the number results returned by an enumeration.
     <attrib name="##ElementCount##" alias="elementCount" type="Gauge"/>
 </group>
 ----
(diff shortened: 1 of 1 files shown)


marshallmassengill:
Ran into this today.  This probably should not be a docs change and shouldn't involve the user having to edit the startup options if possible.  Can we enforce this on the module?

dino2gnt:
Seconded.  This screwed us up on a customer call.  If `HTTP/1.1` needs to be forced true, then it should be the default.  Otherwise, it should be forced to 1.1 if gss-auth="true" in the `wsman-config.xml`

dino2gnt:
AH, cool cool cool.  I just found this one first.  Thanks!

PR #7739, opened 2025-05-22
NMS-17951: Postgres fails to come up on integration tests (Round 2)

mershad-manesh:
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-17951

### DIFF
--- a/.circleci/scripts/postgres.sh
+++ b/.circleci/scripts/postgres.sh
@@ -1,6 +1,6 @@
 #!/bin/bash -e
 echo "##### Starting Postgres"
-docker run --rm --name postgres-onms-itest -p 5432:5432 -e POSTGRES_HOST_AUTH_METHOD=trust -d postgres:13-alpine \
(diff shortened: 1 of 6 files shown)


cgorantla (commented):
Not sure if we really need this change.

The main drawback I see is that we need to set the same password when we run tests locally.

indigo423:
> Not sure if we really need this change.
> 
> The main drawback I see is that we need to set the same password when we run tests locally.

Here some background, with moving to 13 as the earliest version of PostgreSQL, they don't allow you to run it with an empty password for the super user. You can get around it by setting the "trust" authentication, but that is in my point of view not ideal.

With PostgreSQL 14+ the password hashing and authentication algorithm SCRAM-SHA256 and deprecated MD5. I think it would be good if we just set a password in the tests which will catch issues that are a) similar to situations as someone is using it in the real world and b) can give us some insights if we can support certain newly introduced authentication methods.

indigo423 (approved):
Looks good to me. If @cgorantla has better reasons than I have, no objections to do it differently.

cgorantla (approved):
LGTM!   Should be fine. Just need to be aware of default password when we run tests.

PR #7741, opened 2025-05-29
NMS-17968: Menu redesign initial impl for top and side menus

synqotik:
**NOTE** This PR targets `features/menu-redesign`. We will need to keep this feature branch open for awhile until we can safely replace the legacy menu.

We are redesigning the OpenNMS menu to make it easier and more intuitive to use. 

Basically we are adding a side menu that will contain most of the current menu items. Then we are redesigning the top menu to have a simpler set of options that is more geared towards getting the user up and running quickly.

The OpenNMS web UI is a mix of legacy JSP (Java Server Pages), some Vaadin pages, as well as the newer Vue.js SPA (Single Page Application) pages. Previously, we updated the Vue menu to look and act much like the legacy menu.

The current design is to update the Vue SPA (Single Page Application) menu as well as create a new project `ui-components`, which will be a Vue app embedded in JSP and Vaadin pages, and eventually remove the legacy menu.

The `ui` and `ui-components` projects will both have nearly identical files for creating the Vue-based top and side menus. At some point in the future we may need to refactor those components out to a separate library.

This PR is for an initial implementation. It is understood that there are many things to fix, many of which are documented in the epic [NMS-17906](https://opennms.atlassian.net/browse/NMS-17906).

This PR adds the new `ui-components` project and incorporates it into the build process, and has usable top and side menus on almost all new and legacy pages.

It is understood that smoke/integration tests will fail, they will need to be fixed in subsequent PRs.

Screenshot 2025-05-29 at 11 34 29

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/17968


indigo423:
> Basically we are adding a side menu that will contain most of the current menu items. Then we are redesigning the top menu to have a simpler set of options that is more geared towards getting the user up and running quickly.

I would suggest to not mix up the problem description for a) tool onboarding with new users vs. b) improving usability to access of functions. The proposal I suggested in [OEP-0009: Modernization of the Navigation Menu](https://opennms.discourse.group/t/oep-0009-modernization-of-the-navigation-menu/4404/1) is addressing the generic usability without restricting it to just "new users" in trade of making it useless to existing users who need access to various functions.

The reason I picked a different strategy in my proposal was a) you don't solve the key problem of making things hard to use and b) you make it automatically easier for new users. The proposed method to to limit it to "new user" is, a) being a new user is just a one time thing and b) it limits them to explore anything beyond that - which makes it less useful. 

These are the main reason I did the work going through the "whole content" of our current tool and restructured all of it in [Content restructuring](https://opennms.discourse.group/t/oep-0009-modernization-of-the-navigation-menu/4404/1#p-10666-content-restructuring-7). The first entry level gives users a more descriptive set of high level functions. How you want to name these and organize them is now a much more valuable discussion from a user perspective.

When we had the session, I expressed some limitations of the FeatherDS user interface design of the menu structure. The design is very limited in scaling in the sense of holding a certain number of menu items. In the proposal I made we need ~14 - 18 menu items without a scrollbar. That is nothing exceptional in applications in our domain. Competitors in our domain solved that problem successfully.

We should address the "user onboarding" journey independently and should not mix it with your menu redesign proposal.

synqotik:
> > Basically we are adding a side menu that will contain most of the current menu items. Then we are redesigning the top menu to have a simpler set of options that is more geared towards getting the user up and running quickly.
> 
> ### Feedback on the approach

@indigo423 

This is the first of probably many PRs to implement this. This is going into a feature branch and not yet into `develop`. 

The main thing this does is incorporate a side menu into most legacy and Vue pages and gets it into the build process, and also replaces the legacy menubar with Vue implementations. This validates this approach from a technical standpoint.

The exact menu layout, styling, menu behavior etc. are all very subject to change. The FeatherDS component we are using can be customized a *lot* more than the current implementation, or we can find a different component.

This is sort of an initial implementation / POC as the first step in an iterative approach. We can incorporate something more like your design in subsequent PRs (that's the intention).

indigo423:
@synqotik thank you for taking the time explaining the details, I really appreciate it and it helps me to understand where I can help contributing. Here what I read in a nutshell:

* You did the heavy lifting by getting the infrastructure in place to get us started on this topic, which is risky because of the many legacy UI stacks we have
* We take right now what's available in FeatherDS to get us going and don't make that a blocker for that improvement

Just some food for thoughts:

To avoid having a long running feature branch you have to maintain, we can come up with a version which we can deliver in 34.0.0-SNAPSHOT/35.0.0-SNAPSHOT with the following goals for  feedback from our users:

### Phase I - SNAPSHOT release
* Make the implementation stable, find elementary bugs in applications using different UI technologies, e.g. Vaadin and AngularJS
* To achieve that goal we can keep the menu structure as is and just get a technical foundation working, we just want feedback if the new menu works as we expect without issues or we miss something.
* There is basically no additional user value, the user clicks the same things in a vertical navbar than in the old navbar

### Phase II - STABLE release
* Release it in a 34.0.0 or 34.1.0 release
* Add some user value. As long we don't have a scalable Navbar component I can come up with a trimmed down version as described in [Content Restruction](https://opennms.discourse.group/t/oep-0009-modernization-of-the-navigation-menu/4404/1#p-10666-content-restructuring-7) that might work better with our current limitations
* I think what we can look for feedback from the usability point of view for the following questions:
  * Naming: Are we on the right track renaming and group things
  * Icons: Do they make sense and do they work for our users or do we add confusion
  * Positive/Negative feedback around changing the navigation
  
To add some user value regarding usability, I think this improvement is something we can look at for "Phase II" as well from here: https://github.com/OpenNMS/opennms/pull/7437
  
What do you think?

indigo423:
@synqotik The "ADD NODE" button doesn't make a lot of sense to me in the main navigation, for the following reasons:

* The main header has a global function and just highlighting a very specific item how a user can add one node to a requisition is a specific administrative task. 
* The value adding a single node in an application wide navigation bar is exaggerated

### Suggestion:
* I would just put in the left navigation bar as "Add a Node". A new user will see it quickly enough and the left navigation bar is ROLE aware and if a user is not in the role ROLE_PROVISIONING or "ROLE_ADMIN" the item isn't shown
* Having it in the main navigation is prominent enough that a user even unfamiliar with the tool will find it

indigo423:
@synqotik With the new date time formatting, does it reflect the settings we give users from here: https://github.com/OpenNMS/opennms/blob/585ac9244ed7d93092c35d815c3add2b00478b97/opennms-base-assembly/src/main/filtered/etc/opennms.properties#L365

marshallmassengill:
> ### Suggestion:
> 
>     * I would just put it in the left navigation bar as "Add a Node". A new user will see it quickly enough and the left navigation bar is ROLE aware and if a user is not in the role ROLE_PROVISIONING or "ROLE_ADMIN" the item isn't shown
> 
>     * Having it in the main navigation is prominent enough that a user even unfamiliar with the tool will find it

I agree with this.  I was recently using a different system and was expecting it in a menu where it wasn't.  I think we would benefit from that.

indigo423:
> There are things we can likely remove entirely and there is also stuff buried in the Admin page that could probably benefit from being in a menu

@marshallmassengill look at the [content restructuring proposal](https://opennms.discourse.group/t/oep-0009-modernization-of-the-navigation-menu/4404#p-10666-content-restructuring-7) and drop comments to work it out.

synqotik:
> > ### Suggestion:
> > ```
> > * I would just put it in the left navigation bar as "Add a Node". A new user will see it quickly enough and the left navigation bar is ROLE aware and if a user is not in the role ROLE_PROVISIONING or "ROLE_ADMIN" the item isn't shown
> > 

I'll bring this up in the UX/UI meeting today.

Re fixing menu items, yes we can decide on the menu structure.

Shahbaz-dataq (approved):
Overall, the changes look good. Just a couple of suggestions: we should clean up any unused code as we go. Also, I’m not sure if creating a new UI project for the old JSP pages' menu is necessary — maintaining two separate projects for menus might become difficult in the long run. Let me know your thoughts on this.

PR #7751, opened 2025-06-10
NMS-17991: Remove Nexus as a Maven shim in our dependencies

indigo423:
Refactoring how we get internal, forked and external dependencies from remote repositories. The introduced refactoring adds remote repositories with the original source where dependencies are published instead of our Maven Nexus shim. That gives the following benefits:

* transparency which module requires dependencies from external sources and where are they coming from
* when repository sources aren't available, it will fail fast and immediately, orphaned dependencies aren't longer hidden behind a caching Maven server
* Caching to build resilience is introduced in the .circleci/settings.xml using our new mvn.cdn.useast1.opennms.com server dedicated for caching purposes.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-17991

### DIFF
--- a/container/shared/src/main/resources/etc/org.ops4j.pax.url.mvn.cfg
+++ b/container/shared/src/main/resources/etc/org.ops4j.pax.url.mvn.cfg
@@ -106,7 +106,7 @@ org.ops4j.pax.url.mvn.defaultRepositories=\
 # The default list includes the following repositories:
 #    http://repo1.maven.org/maven2@id=central
 #    http://repository.apache.org/content/groups/snapshots-group@id=apache@snapshots@noreleases
-#    https://oss.sonatype.org/content/repositories/snapshots@id=sonatype.snapshots.deploy@snapshots@noreleases
+#    https://central.sonatype.com/repository/maven-snapshots@id=sonatype.snapshots.deploy@snapshots@noreleases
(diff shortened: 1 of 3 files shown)


container/bridge/rest/src/license/THIRD-PARTY.properties
```diff
@@ -0,0 +1,37 @@
+# Generated by org.codehaus.mojo.license.AddThirdPartyMojo
```
synqotik, line 1: Most/all `THIRD-PARTY.properties` don't have the `Already used licenses in project` part, I think there's a build process (in top-level `pom.xml`, `license-maven-plugin`) that creates all that so that it doesn't have to be included in each `THIRD-PARTY.properties` file. May want to just double check :)

synqotik (changes requested):
Seems reasonable to me - though not an area I'm too knowledgeable about.

See my comment re `THIRD-PARTY.properties`, should check if the `Already used licenses` need to be added to these files or not.

.circleci/build-triggers.override.json
mershad-manesh, line 1: We shouldn't check in this file :)

cgorantla:
@indigo423 @mershad-manesh Is this still relevant ?

indigo423:
> @indigo423 @mershad-manesh Is this still relevant ?

it is relevant and a capacity issue.

PR #7752, opened 2025-06-10
NMS-18018: Fix test failure with SCV

cgorantla:
Merge from foundation-2024 created these two set of properties that we need to handle.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18018

### DIFF
--- a/features/scv/api/src/main/java/org/opennms/features/scv/utils/ScvUtils.java
+++ b/features/scv/api/src/main/java/org/opennms/features/scv/utils/ScvUtils.java
@@ -52,13 +52,21 @@ public class ScvUtils {
 
         final Properties onmsProperties = new Properties();
         String keyStoreType = System.getProperty(SCV_KEYSTORE_TYPE_PROPERTY);
-        if (keyStoreType !=null && !keyStoreType.isEmpty()) {
+        String keyStoreKey = System.getProperty(KEYSTORE_KEY_PROPERTY);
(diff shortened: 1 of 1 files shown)


features/scv/api/src/main/java/org/opennms/features/scv/utils/ScvUtils.java
```diff
@@ -52,13 +52,21 @@ public static Properties loadScvProperties(String opennmsHome) {
...
-            loadProperties(Path.of(opennmsHome, "etc", OPENNMS_PROPERTIES_D_NAME).toString(), onmsProperties);
-            loadProperties(Path.of(opennmsHome, "etc", OPENNMS_PROPERTIES_NAME).toString(), onmsProperties);
         }
-
+        if (keyStoreKey != null && !keyStoreKey.isEmpty()) {
+            onmsProperties.setProperty(KEYSTORE_KEY_PROPERTY, keyStoreKey);
+        }
+        // Only load properties from files if both system properties are not set
+        if ((keyStoreType == null || keyStoreType.isEmpty()) ||
+                (keyStoreKey == null || keyStoreKey.isEmpty())) {
```
mwajahatabbasi, line 64: Use onmsProperties.isEmpty() would be better instead.

cgorantla: Even if one property is available, we need to load other property, so onmsProperties.isEmpty()  may not work here.

PR #7755, opened 2025-06-12
NMS-18002: Prevent old zookeeper version from being included

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-18001


cgorantla (commented):
Is it possible to find where the dependency is coming from and directly exclude there ?

PR #7757, opened 2025-06-12
NMS-18002: Prevent old jetty version from being loaded

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-18002


cgorantla (approved):
LGTM!

PR #7759, opened 2025-06-13
NMS-18021: Change pending Problems to pending Alarms

indigo423:
The alarm boxes for Nodes, Applications and Business services is changed from "...with Problems" to "...with Alarms" to be less confusing.

Screenshot 2025-06-13 at 14 03 14

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18021

### DIFF
--- a/core/upgrade/src/test/resources/etc2/opennms.properties
+++ b/core/upgrade/src/test/resources/etc2/opennms.properties
@@ -451,11 +451,11 @@ opennms.eventlist.showCount=false
 # page in the OpenNMS web UI. Default: true
 #opennms.nodesWithOutages.show=true
 
-# This value allows you to set the number of nodes with problems to display on the front
+# This value allows you to set the number of nodes with alarms to display on the front
 # page in the OpenNMS web UI. Default: 16
 #opennms.nodesWithProblems.count=16
 
-# This value allows you to enable/disable the nodes with problems box on the front
+# This value allows you to enable/disable the nodes with alarms box on the front
 # page in the OpenNMS web UI. Default: true
 #
 #opennms.nodesWithProblems.show=true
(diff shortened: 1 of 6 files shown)


marshallmassengill (approved):
My only comment is that I think this makes sense to target for foundation 2024 as it's non-breaking functionality and I'd be happy to see it get picked up sooner for all of our users.  I do very much appreciate this change!

indigo423:
> My only comment is that I think this makes sense to target for foundation 2024 as it's non-breaking functionality and I'd be happy to see it get picked up sooner for all of our users. I do very much appreciate this change!

Back ported to foundation-2024, we need to wait for CircleCI now.

cgorantla (approved):
LGTM!

dino2gnt:
Late to the party here, but if the alarm (or situation) is in the list, it isn't _pending_, it _exists_.

indigo423:
> That is pedantic and correct... it's probably a fair point that we could shorten all of these to "X with alarms".

"X with alarms" wouldn't be correct, because this boxes don't show alarms that are in the state "open" and "acknowledged" by a user. "X with alarms" would be to unspecific.

dino2gnt:
A raised alarm isn't "pending".  There is no "pending" state.  "Nodes with alarms" is more correct than "nodes with pending alarms".

marshallmassengill:
Not trying to pile on more but I do think this needs some additional thought into the word "pending".  We also use the word "outstanding" on the alarms page:

image

marshallmassengill:
Throwing out the idea of "X with Active Alarms"... though I realize that "Active" isn't a state either, I do think it is more descriptive than pending and/or outstanding.  Just an idea.

indigo423:
> Throwing out the idea of "X with Active Alarms"... though I realize that "Active" isn't a state either, I do think it is more descriptive than pending and/or outstanding. Just an idea.

From my non-native speaker perspective, "pending" and "outstanding" seems reasonable. What would better describe from your native speaker perspective an alarm with the condition "Unacknowledged" and "severity greater than normal"?

marshallmassengill:
"Unhandled" is probably the most descriptive English word that aligns closely with those states... "Open" might be an option as well.  I'm going to ponder this over the weekend and come at it fresh on Monday.

PR #7760, opened 2025-06-13
NMS-18022: Add tss write metrics to usage report

dino2gnt:
https://opennms.atlassian.net/browse/NMS-18022

Adds a metric to the usage report for metrics written to TSS `integration` strategies by pulling `count` from `org.opennms.timeseries:name=samples.write.integration`


synqotik (approved):
LGTM!

PR #7763, opened 2025-06-17
NMS-16375: merged foundation-2024

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-16375


christianpape:
> Not sure if we really need PR for this merge. Also I would like this to be on `develop`

PR #7768, opened 2025-06-19
NMS-17981: Added changes regarding duplicate trivy issues

MuhammadTalhaDataQHealth:
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-17981


indigo423:
Shouldn’t this an issue in the OpenNMS (NMS) project rather than Lokahi (LOKI)?

PR #7772, opened 2025-06-20
NMS-18033: Remove JMX detectors from default FS and cleanup

marshallmassengill:
Trying this again @indigo423 ;) - now with 100% less unrelated doc updates because I didn't switch branches locally.

These detectors don't seem to be commonly used and I'm not sure they belong in the default FS. I moved them into an example and a dedication Minions FS.

This is probably worth a discussion before merging.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18033


dino2gnt:
`selfmonitor` is intended for services that exist on the localhost node that represents the core.  `JMX-Minion`, `JMX-Cassandra`, and `JMX-Cassandra-Newts` will never exist on that node, so these services don't need to be here.

If you just want to pare down the DFS, then delete these and be done with it.  Move them to `examples/`, maybe.  Or don't.

`JMX-Minion` should exist in the `minions` foreign source, or the default, otherwise we're throwing away all our built-in JMX monitoring of Minion.

opennms-base-assembly/src/main/filtered/etc/foreign-sources/selfmonitor.xml
```diff
@@ -21,6 +21,41 @@
       <detector name="OpenNMS-DB-Stats" class="org.opennms.netmgt.provision.detector.loop.LoopDetector">
          <parameter key="ipMatch" value="127.0.0.1"/>
       </detector>
+      <!--The following aren't strictly for the self-monitor but are related and have been moved from the default FS-->
```
cgorantla: Align formatting as above. The comment is fine though.

marshallmassengill:
I'm going to realign this closer to how @dino2gnt has suggested above.  Will remove from default FS and then put the two Cassandra ones into an example.  The Minion JMX though is less obvious to me.  I'm not sure where that FS gets created or exists.  Can someone point me at it?

dino2gnt:
You could create one, as `./opennms-base-assembly/src/main/filtered/etc/foreign-sources/Minions.xml` with the default `Minion-SNMP-Policy`, an SNMP detector, and the `JMX-Minion` detector

dino2gnt (dismissed):
This looks good to me, but fix the tests, too.

marshallmassengill:
I added the cassandra example to the ignored files in integration-tests/config/src/test/java/org/opennms/netmgt/config/WillItUnmarshalCoverageMetaIT.java

And I addedthe Minions.xml to integration-tests/config/src/test/java/org/opennms/netmgt/config/WillItUnmarshalIT.java

dino2gnt (dismissed):
yeap, LGTM. The test failures on this run aren't related :+1:

opennms-base-assembly/src/main/filtered/etc/foreign-sources/Minions.xml
```diff
@@ -0,0 +1,24 @@
+<foreign-source xmlns="http://xmlns.opennms.org/xsd/config/foreign-source" name="Minions">
```
cgorantla: Did we validate how this effects current way of creating this in the code ?

marshallmassengill: I have not tested it but I think that gets done here: https://github.com/marshallmassengill/opennms/blob/780d7ca6c3792c2aee55bb6c130ce6b5eff756e7/features/minion/heartbeat/consumer/src/main/java/org/opennms/minion/heartbeat/consumer/HeartbeatConsumer.java#L282

I'm not certain if we overwrite an existing file or not... if we do then I would probably need to adjust this code to add the extra detector here.  If you want to take a look and let me know if that's the right place, I can see about adding it or adjusting things.

Other thing we can try is a new install and then creating the minions requisition and then trying to add a minion to see if the file gets overwritten or amended.

marshallmassengill: Well... doesn't work as it is @cgorantla - see my latest comment below.

marshallmassengill:
Closing this out for now.  I do think we need to deal with the cleanup here but it can wait.

PR #7774, opened 2025-06-23
NMS-17981: Added changes regarding duplicate trivy issues and trivy report format

MuhammadTalhaDataQHealth:
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-17981


mershad-manesh (approved):
The changes looks good.
Just a side note, I wonder if would be better to download and process all the reports into a single job.

PR #7775, opened 2025-06-23
NMS-18042: Fix BMP Aggregator parser to handle IPv6 peers correctly

tacgnol:
### All Contributors

A juniper MX240 that has a large number of IPv4 and IPv6 peers, that when connected to OpenNMS via BMP would throw the following exception and terminate the session almost immediately.

Invalid packet: java.lang.IndexOutOfBoundsException: readerIndex(2) + length(16) exceeds writerIndex(8)

This occurred in the Aggregator path attribute parser, which incorrectly assumed that the aggregator’s IP address length depended on the peer’s address family (via PeerFlags). For IPv6 peers, it attempted to read 16 bytes instead of the correct 4 bytes.

This patch should always read 2 bytes for ASN and 4 bytes for IP address.

The patch was tested on OpenNMS Horizon 33.1.7

JIRA: https://opennms.atlassian.net/browse/NMS-18042

### DIFF
--- a/features/telemetry/protocols/bmp/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/parser/proto/bgp/packets/pathattr/Aggregator.java
+++ b/features/telemetry/protocols/bmp/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/parser/proto/bgp/packets/pathattr/Aggregator.java
@@ -24,20 +24,30 @@ package org.opennms.netmgt.telemetry.protocols.bmp.parser.proto.bgp.packets.path
 import static org.opennms.netmgt.telemetry.listeners.utils.BufferUtils.uint16;
 
 import java.net.InetAddress;
+import java.net.UnknownHostException;
 
 import org.opennms.netmgt.telemetry.protocols.bmp.parser.proto.bmp.PeerFlags;
+import org.opennms.netmgt.telemetry.listeners.utils.BufferUtils;
 
 import com.google.common.base.MoreObjects;
 
 import io.netty.buffer.ByteBuf;
+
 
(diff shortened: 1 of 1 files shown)


features/telemetry/protocols/bmp/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/parser/proto/bgp/packets/pathattr/Aggregator.java
```diff
@@ -24,20 +24,30 @@
...
 
 import io.netty.buffer.ByteBuf;
 
+
 public class Aggregator implements Attribute {
     public final int as;              // uint16
     public final InetAddress address; // uint32
 
     public Aggregator(final ByteBuf buffer, final PeerFlags flags) {
+        if (buffer.readableBytes() < 6) {
```
cgorantla, line 42: May be we can add this constant for better clarity:

private static final int AGGREGATOR_ATTR_LENGTH = 6; // 2 bytes ASN + 4 bytes IP

cgorantla (changes requested):
@tacgnol Thanks for your contribution. Can you update with suggested improvement.

features/telemetry/protocols/bmp/parser/src/main/java/org/opennms/netmgt/telemetry/protocols/bmp/parser/proto/bgp/packets/pathattr/Aggregator.java
cgorantla: pushing it for now, we can update this later.

PR #7782, opened 2025-06-27
NMS-18051: Fix incorrect IP address discovery

dino2gnt:
This fixes the case where we're already handling a non-RFC4001-compliant agent, where the agent appends the interface ifIndex to the instance, causing the IP address extraction to be off-by-one.  cough fortinet cough.

Also adds some additional debug output, because I found it helpful while exploring this code.

Before:
image

After:
image

### All Contributors

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18051
https://opennms.atlassian.net/browse/SUPPORT-2975
https://opennms.discourse.group/t/wrong-discovery-of-ipaddresses/3949/30


dino2gnt:
There may be a better way to do this, I'm no expert.

marshallmassengill (approved):
LGTM!

indigo423 (commented):
@cgorantla it looks like to me we might need to get some test coverage for something like this. What do you think?

dino2gnt:
The failure makes no sense to me; the number of interfaces shouldn't have changed.

opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/IPAddressTableTracker.java
```diff
@@ -119,7 +119,14 @@ public String getIpAddress() {
                         LOG.debug("IPV4: BAD AGENT: Could not parse raw oids as octet string", e);
                     }
                 }
-                addressIndex = instanceIds.length - 4;
+                // Some popular vendors append the ifIndex to the instance value, and argue it is RFC 4001 compliant.
+                if (instanceIds[(instanceIds.length - 1)] == getIfIndex()) {
+                    LOG.debug("IPV4: BAD AGENT: Last instance value is the ifIndex, adjusting address index!");
+                    addressIndex = instanceIds.length - 5;
+                }
+                else {
```
christianpape:
```suggestion
                } else {
```

opennms-provision/opennms-provisiond/src/main/java/org/opennms/netmgt/provision/service/IPAddressTableTracker.java
```diff
@@ -132,7 +139,13 @@ public String getIpAddress() {
                         LOG.debug("IPV6: BAD AGENT: Could not parse raw oids as octet string", e);
                     }
                 }
-                addressIndex = instanceIds.length - 16;
+                if (instanceIds[(instanceIds.length - 1)] == getIfIndex()) {
+                    LOG.debug("IPV6: BAD AGENT: Last instance value is the ifIndex, adjusting address index!");
+                    addressIndex = instanceIds.length - 17;
+                }
+                else {
```
christianpape:
```suggestion
                } else {
```

dino2gnt:
> The failure makes no sense to me; the number of interfaces shouldn't have changed unless the test was broken. 🤷

Even more curious, the test asserts that there should be four SNMP interfaces, but if I load the associated snmpwalk used by this test into an snmpsim and provision a node against it, every OpenNMS version I have tested (32, 33, 2024) creates six SNMP interfaces.  I have no idea how this works.

christianpape:
> > The failure makes no sense to me; the number of interfaces shouldn't have changed unless the test was broken. 🤷
> 
> Even more curious, the test asserts that there should be four SNMP interfaces, but if I load the associated snmpwalk used by this test into an snmpsim and provision a node against it, every OpenNMS version I have tested (32, 33, 2024) creates six SNMP interfaces. I have no idea how this works.

I debugged into the code and I encountered for example the following problem: the walk contains the ipAddress 127.0.0.1 which has an ifIndex of 1, which means that this is handled by `instanceIds[(instanceIds.length - 1)] == getIfIndex()` and results in a wrong ipAdress of "1.127.0.0". This adds an additional interface since loopback addresses are normally skipped.

dino2gnt:
> I debugged into the code and I encountered for example the following problem: the walk contains the ipAddress 127.0.0.1 which has an ifIndex of 1, which means that this is handled by `instanceIds[(instanceIds.length - 1)] == getIfIndex()` and results in a wrong ipAdress of "1.127.0.0". This adds an additional interface since loopback addresses are normally skipped.

I still don't understand, because it's complaining about SNMP interfaces, not IP interfaces?  We'll see if it still fails with the new changes, though.

christianpape (dismissed):
LGTM.

dino2gnt:
> Also, would be good to add a test case in ProvisionerIT.java to reflect new behavior.

Open to suggestions on how to best add this, I don't really know how to write tests.

PR #7788, opened 2025-07-02
NMS-18043: Move Node Search input

atif1019:
Move Search Input box to the left
Prompt should be "Search List"
Should have Search icon
Add filter button

For now it will just search the same as the current input.

Separate issue for the Filter functionality.

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18043

### DIFF
--- a/ui/src/components/Nodes/NodesTable.vue
+++ b/ui/src/components/Nodes/NodesTable.vue
@@ -4,9 +4,18 @@
       <div class="feather-row title-bar">
         <span class="title">Structured Node List</span>
         <div class="action-buttons-container">
-          <NodeDownloadDropdown :onCsvDownload="onCsvDownload" :onJsonDownload="onJsonDownload" />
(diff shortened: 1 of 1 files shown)


ui/src/components/Nodes/NodesTable.vue
```diff
@@ -1,97 +1,85 @@
...
-            summary="Nodes"
-          >
-            <thead>
+        <div id="wrap" class="node-table">
+          <table :class="tableCssClasses" summary="Nodes">
+            <thead >
               <tr>
                 <th scope="column" />
                 <template v-for="column in nodeStructureStore.columns" :key="column.id">
-                  <FeatherSortHeader
```
synqotik, line 47: Please change all these back to the original formatting with one prop per line.

synqotik (changes requested):
Please change the formatting, then I can re-review.

PR #7797, opened 2025-07-12
NMS-18054, NMS-18055: Update menu as per UX design, fix MenuHeader smoke tests

synqotik:
Update the new Sidemenu menu items as per the UX design.

Fix the smoke test accordingly. Also fix smoke tests for the new logout implementation.

Update the Menu Rest API to include fields for type (can have `item`, `header`, `separator`) and to denote that a link is an external absolute URL.

Update how the central Search box is tested, since it's now the new Vue version. May eventually remove some previous code in `smoke-test//ui/framework/search` once we handle some more cases.

Various updates to the UI to make it easier to smoke test using Xpath to find elements.

UX design [Figma link](https://www.figma.com/proto/bylaeM1QWiIhQARc6GwGJE/Business-Service-Management--Service-Level-Agreement-?page-id=10531%3A33446&node-id=10531-35135&p=f&viewport=221%2C324%2C0.13&t=llNHaXUf6enbUrpQ-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=10531%3A35135)

**Note:** It is expected that some smoke tests will still fail, but the `MenuHeaderIT` ones should now pass.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18054
* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18055


christianpape:
The included pages have padding on the left but not on the right side.

christianpape:
I’m pretty confused how the menu items are categorized. Furthermore some of the icons do not really help me to understand what to expect in this category. The order, categorization and the used icons as described in the dicourse discussion were a lot more intuitive to me.

https://opennms.discourse.group/t/oep-0009-modernization-of-the-navigation-menu/4404

synqotik:
> I’m pretty confused how the menu items are categorized. Furthermore some of the icons do not really help me to understand what to expect in this category. The order, categorization and the used icons as described in the dicourse discussion were a lot more intuitive to me.
> 
> https://opennms.discourse.group/t/oep-0009-modernization-of-the-navigation-menu/4404

Agree, I can bring this up in the UX meeting today. I based the menu organization based on the UX Figma design, but may make more sense to use something closer to Ronny's version.

christianpape:
> > I’m pretty confused how the menu items are categorized. Furthermore some of the icons do not really help me to understand what to expect in this category. The order, categorization and the used icons as described in the dicourse discussion were a lot more intuitive to me.
> > https://opennms.discourse.group/t/oep-0009-modernization-of-the-navigation-menu/4404
> 
> I discussed this at the UX meeting today. Product and UX have currently signed off on this current implementation. It is expected we will go through some phases of reorganizing the menu based on user input, so it's definitely not the final version.
> 
> I am also thinking of providing an alternative menu template that looks either more like Ronny's version or like the "legacy" menu organization, so you can (fairly easily) customize your menu if desired.



> > I’m pretty confused how the menu items are categorized. Furthermore some of the icons do not really help me to understand what to expect in this category. The order, categorization and the used icons as described in the dicourse discussion were a lot more intuitive to me.
> > https://opennms.discourse.group/t/oep-0009-modernization-of-the-navigation-menu/4404
> 
> I discussed this at the UX meeting today. Product and UX have currently signed off on this current implementation. It is expected we will go through some phases of reorganizing the menu based on user input, so it's definitely not the final version.
> 
> I am also thinking of providing an alternative menu template that looks either more like Ronny's version or like the "legacy" menu organization, so you can (fairly easily) customize your menu if desired.

So Product and UI, who have never used our product and do not understand in detail what functionality lies behind a menu entry, decide how the menu should look like. That makes no sense to me. But ok, I’m looking forward to see the alternative menu template.

indigo423:
> Product and UX have currently signed off on this current implementation.

A nice way to say, I don't care what you are saying. Message received.

indigo423:
> @synqotik You did us a great favour in your initial PR here #7741 by sorting the main application header. If you could bring that back and remove that "ADD NODE" button would be great. I don't understand the iconography and the sorting and structure of the menu items of that version to be honest.
> 
> Solving the problem to provide a grouping for "Daily usage" is an extremely underestimated problem. The issue with this, it will creating a bad workflow for the most people because depending on their environment and how they use OpenNMS, daily usage will be very different. Example, people with a geographically dispersed environment or a heavy usage of application perspective monitoring might want to use a geographic map and the application monitoring "on a daily basis" but there will be a lot of people who simply never care.
> 
> @dschlenk brought up a very interesting [topic](https://opennms.discourse.group/t/oep-0009-modernization-of-the-navigation-menu/4404/14?u=indigo). I think if we could give people like him space in the menu where he can customise a "Favourite" sub menu as he likes would be a great addition. It could serve power users who use the tool in various different contexts in the long run.
> 
> I would like to vote removing the "Add Node" button in the main application header bar for the following reason:
> 
> The task adding a single node is a task which is not worth to be on a very expensive place as the main application header.
> 
> * This task is not done by every user, the user needs to have special permissions to manipulate the inventory
> * It creates confusion, because 99.9% is in the main navigation but 1 button isn't
> * It requires a user to go to "Provisioning Requisitions" anyway to create a Requisition first and sync it, so that so he can add a node in that requisition.
> * From an onboarding perspective, there is 0 value providing that button so prominently, all the user can do is adding a node in the "self-monitor" requisition which is the worst place to create your first node in.
> * The main application header is always visible and should be reserved for indication of monitoring issues people need really to take care of.
> * From an onboarding perspective, the Add Node button has only limited value because people will use it at the beginning when they evaluate the tool, but won't use it later when they use API's or other functions. It becomes less value over time and still consumes the most expensive place in an application – the application header.
> * It makes no sense to me to use the main application header as a collection of shortcuts for user onboarding purposes, a) Just adding a node is just one piece of an onboarding journey, there are also notifications and thresholds. It doesn't make sense to just dump everything which might help new users to put into the main application header. A wizard or a guided tour is much more useful.

synqotik:
Thank you for all the comments and suggestions. A big part of this PR was getting smoke tests to run with the new menu design.

We will be discussing the categorization of the menu items (which is fairly easy to change since it's mostly controlled by a json configuration file) as well as the other topics brought up.

synqotik:
New version has 3 menu templates. Also added more customization that can be done in the template, for example the icon.

To use a different template, you can literally just copy the template (in `jetty-webapps/opennms/WEB-INF`) to `menu-template.json` and it will immediately be picked up by the Menu Rest API and thus the menu UI.

- `menu-template-default.json` - New default template. This is based on Ronny's organization. This is also used as `menu-template.json`, i.e. the default template that will be active

- `menu-template-legacy.json` - A menu template that is based on the existing Horizon top menu organization. This could be helpful for users to migrate to the new side menu but with the same menu layout as before

- `menu-template-alt.json` - This is based on the UX/Figma design

synqotik:
> The included pages have padding on the left but not on the right side.

This should now be fixed.

PR #7802, opened 2025-07-18
NMS-18046 FE: Node List table updates

atif1019:
### All Contributors

Table should only display 5 columns at a time, plus Navigation and Action columns

There is a new column header which has icon buttons to show the next/previous sets of columns.

For now, use navigation/ChevronLeft and navigation/ChevronRight

These only display if there are more columns that are not displayed

These display the next or previous 5 columns.

Action columns "Edit" button should navigate to the Node details page for the node.

The vertical ellipse icon button should display a dropdown similar to current vertical ellipse button which allows user to navigate to the various pages for that node.

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18046

### DIFF
--- a/ui/src/components/Nodes/NodesTable.vue
+++ b/ui/src/components/Nodes/NodesTable.vue
@@ -100,22 +100,53 @@
           >
             <thead>
               <tr>
-                <th scope="column" />
+                <th v-if="canNavigateLeft">
+                  <div @click="navigateColumns(Direction.Left)">
+                    <FeatherIcon
+                      :icon="ChevronLeft"
+                      class="edit-icon"
+                    />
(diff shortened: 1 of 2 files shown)


ui/src/components/Nodes/NodesTable.vue
```diff
@@ -412,7 +484,7 @@ const removeItem = (item: IAutocompleteItemType, type: FilterTypeEnum) => {
       nodeStructureStore.removeFlow(item);
       break;
     case FilterTypeEnum.Location:
-      console.log("locaationparam",item)
+      console.log("locaationparam", item)
```
synqotik: Remove the `console.log`

PR #7821, opened 2025-07-28
NMS-18161 : Implement dynamic TLS certificate loading from remote host

mwajahatabbasi:
Jira : [https://opennms.atlassian.net/browse/NMS-18161](https://opennms.atlassian.net/browse/NMS-18161)


cgorantla (commented):
I don't think this is right solution. 

If you are testing this with a real TLS cert, it should be automatically validated.  We don't need to design any solution for self signed certs. They can specify TLS cert in the path.

PR #7826, opened 2025-07-29
Docs: Fix broken xref in sflow docs

dino2gnt:
DOCS DOCS DOCS

The xref is correct in 2022, but broken in 2023.


marshallmassengill (approved):
LGTM

PR #7831, opened 2025-07-31
Docs: redo the housekeeping page

dino2gnt:
DOCS DOCS DOCS DOCS DOCS DOCS

Redo the "Housekeeping" page with information actually relevant to housekeeping, instead of in the context of pre-upgrade or upgrading, which does not belong in this section.


garcia-sr (approved):
LGTM!

PR #7834, opened 2025-08-01
FE: Fix main Search input - parity with legacy

atif1019:
When we added the original Vue SPA menu, we made a new Search box component. It contains much but not all of the functionality of the “legacy” search components. Also it seems to have some bugs and styling issues. Need to fix this to be able to completely replace the legacy search box.

https://opennms.atlassian.net/browse/NMS-17975

### DIFF
--- a/ui/src/components/Menu/Search.vue
+++ b/ui/src/components/Menu/Search.vue
@@ -99,12 +99,10 @@ const hasResults = computed(() => {
     Object.keys(searchStore.searchResultsByContext).length > 0
 })
 
-// Create a unique key for each search result
 const createResultKey = (contextKey: string | number, subContextKey: string | number, itemIndex: number) => {
   return `${contextKey}-${subContextKey}-${itemIndex}`
 }
 
-// Set ref for each SearchResult component
 const setSearchResultRef = (el: any, contextKey: string | number, subContextKey: string | number, itemIndex: number) => {
   if (el) {
     const key = createResultKey(contextKey, subContextKey, itemIndex)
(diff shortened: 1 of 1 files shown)


ui/src/components/Menu/Search.vue
```diff
@@ -71,136 +77,102 @@ const props = defineProps({
...
-  if (userInput && !loading.value) {
-    searchState.dropdownOpen = true
-    searchStore.search('' + userInput)
-  } else if (!userInput) {
-    searchState.dropdownOpen = false
-  } else if (userInput && loading.value) {
-    searchState.pausedSearch = userInput
-  } 
+const onKeyDown = (event: KeyboardEvent) => {
+  if (['ArrowDown', 'ArrowUp', 'Enter', 'Escape'].includes(event.key)) {
```
synqotik, line 211: Can the user still use the arrow keys to navigate through the dropdown items?

atif1019: I also checked the previous legacy version with FeatherDropdown. When using the arrow keys to navigate options, it only moves down by one and then loses focus. Pressing Enter doesn’t trigger the route change either.

synqotik: @atif1019 we need to be able to use the keyboard to navigate up and down the list and to press Enter to go to the link, as well as type anything into the search box and retain focus.

See if you can work on that with your version. You should also try the previous version with the `FeatherDropdown` and see if there is a reason it isn't working as expected, I think it used to work better than it does now

synqotik (changes requested):
We should make sure user can use the keyboard to navigate up/down the menu items, press Enter to go to the link, and also type search terms without losing focus.

Could be done by a different component, or perhaps we can use `FeatherDropdown` but there are just some bugs in how we are using it.

PR #7843, opened 2025-08-05
Docs: Fix incorrect xreference for trap reception

dino2gnt:
DOCS DOCS DOCS DOCS DOCS DOCS 

The event-sources/snmp-traps document contained an invalid reference starting in 2023; this fixes the reference.

Also fixes some invalid variable references because I was in there.


marshallmassengill (approved):
These changes spark much joy.

PR #7848, opened 2025-08-05
NMS-18173: Fix side nav push content; fix Search box navigation and display

synqotik:
This PR fixes a few things.

Updates to the latest Feather Dock/Sidenav control which improves the behavior when the side menu is collapsed or expanded, for example to use `padding-left` instead of `margin-left`. Removed some padding that had previously been added, since the Sidenav will now correctly add it.

Slightly reduced the font size of the date.

Fixed the Search box implementation, there were some issues where Nodes were not being displayed in the search results, and some weird overwrite display behavior. Also added Search Result details which will display for Node results when you mouse over the item. Also pressing Escape when in the search input box will hide the search results dropdown but not close the side menu (though pressing Escape elsewhere will close both).

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18173


Shahbaz-dataq (dismissed):
LGTM!

cgorantla (dismissed):
LGTM!

atif1019 (dismissed):
LGTM

PR #7853, opened 2025-08-06
NMS-18140: Menu redesign documentation

synqotik:
Documentation for the new menu redesign. Also explains how to configure the menu templates.

Also fixes a couple of errors occurring on other doc pages.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18140


docs/modules/operation/pages/deep-dive/menu/introduction.adoc
```diff
@@ -0,0 +1,101 @@
+
+[[menu-overview]]
+= Menu Overview
+:description: Learn about the top and side menus in OpenNMS Horizon/Meridian, which lets users quickly navigate to most of the pages.
```
mershad-manesh: Does it make sense to say `OpenNMS {page-component-title}` instead of `OpenNMS Horizon/Meridian` ?

indigo423 (commented):
Do we want to write the documentation in reference to a previous version? I think that make sense in a release note or a change log. Just imagine you are a new user, you install 34 the first time, you don't care what it was before in the documentation.

docs/modules/operation/pages/deep-dive/menu/introduction.adoc
synqotik: yes, good catch

synqotik:
> Do we want to write the documentation in reference to a previous version? I think that make sense in a release note or a change log. Just imagine you are a new user, you install 34 the first time, you don't care what it was before in when you read the documentation.

Good point. I updated to reword things not to be in reference to a previous version.

docs/modules/operation/pages/deep-dive/menu/introduction.adoc
synqotik: Done!

PR #7864, opened 2025-08-12
NMS-18178 Manage OpenConfig Connections via TwinAPI Dynamically 

Mohammad-Junid:
Below are the configuration steps to get the OpenConfig connector receive flow on the Minion

Install  twin api on minion.

Add node with below MetaData

1.  ```
key : port
 .   value: <gRPC server port>
```

2. ```
key: hostname
 . value: <ip-address of device>
```

3. ```
key: mode
  value : <jti or gnmi>
```

4. ```
key: paths
 . value: <one or more interfaces paths sending flows >
```

. * Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18178

### DIFF
--- a/features/openconfig/telemetry-client/src/main/java/org/opennms/features/openconfig/telemetry/OpenConfigClientImpl.java
+++ b/features/openconfig/telemetry-client/src/main/java/org/opennms/features/openconfig/telemetry/OpenConfigClientImpl.java
@@ -100,6 +100,7 @@ public class OpenConfigClientImpl implements OpenConfigClient {
     private Integer retries;
     private List<Map<String,String>> paramList = new ArrayList<>();
     private AtomicBoolean closed = new AtomicBoolean(false);
+    private final AtomicBoolean enabled = new AtomicBoolean(true);
     private AtomicBoolean scheduled = new AtomicBoolean(false);
     private ExecutorService executor = Executors.newSingleThreadExecutor();
     private ScheduledExecutorService scheduledExecutor = Executors.newSingleThreadScheduledExecutor();
@@ -120,6 +121,11 @@ public class OpenConfigClientImpl implements OpenConfigClient {
 
     @Override
     public void subscribe(OpenConfigClient.Handler handler) {
+        if (!enabled.get()) {
+            LOG.info("Subscription skipped: client is disabled");
+            return;
+        }
+
         boolean succeeded = trySubscribing(handler);
         if (!succeeded) {
             close();
@@ -244,6 +250,10 @@ public class OpenConfigClientImpl implements OpenConfigClient {
 
 
     private void scheduleSubscription(OpenConfigClient.Handler handler) {
+        if (!enabled.get()) {
+            LOG.debug("Rescheduling skipped: client is disabled");
+            return;
+        }
         if (scheduled.get()) {
             // Task is already scheduled.
             return;
@@ -267,7 +277,7 @@ public class OpenConfigClientImpl implements OpenConfigClient {
 
         int retries = this.retries != null ? this.retries : DEFAULT_INTERNAL_RETRIES;
         int interval = this.interval != null ? this.interval : DEFAULT_INTERVAL_IN_SEC;
-        while (!closed.get()) {
+        while (!closed.get() && enabled.get()) {
             ScheduledFuture<Boolean> future = scheduledExecutor.schedule(() -> trySubscribing(handler), interval, TimeUnit.SECONDS);
             try {
                 succeeded = future.get();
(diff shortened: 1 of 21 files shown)


features/openconfig/telemetry-client/src/main/java/org/opennms/features/openconfig/telemetry/OpenConfigClientImpl.java
```diff
@@ -100,6 +100,7 @@ public class OpenConfigClientImpl implements OpenConfigClient {
     private Integer retries;
     private List<Map<String,String>> paramList = new ArrayList<>();
     private AtomicBoolean closed = new AtomicBoolean(false);
+    private final AtomicBoolean enabled = new AtomicBoolean(true);
```
cgorantla: why would you need disabled ? You always shutdown the client when you want to remove client for a given key.  `OpenConfigClientFactory` should be able to handle the multiple clients and shutdown when the client is not needed

features/telemetry/daemon/src/main/java/org/opennms/netmgt/telemetry/daemon/ConnectorManager.java
```diff
@@ -24,20 +24,20 @@
 import java.io.Closeable;
 import java.io.IOException;
 import java.net.InetAddress;
-import java.util.ArrayList;
-import java.util.LinkedHashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
+import java.util.*;
```
cgorantla: Avoid wild card imports

features/telemetry/daemon/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -0,0 +1,56 @@
+<blueprint
```
cgorantla, line 1: telemetry daemon is not part of Minion so this blueprint won't be loaded on Minion

DistPollerDao also doesn't gets loaded on Minion

features/telemetry/daemon/src/main/java/org/opennms/netmgt/telemetry/daemon/ConnectorState.java
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied.  See the License for the specific
+ * language governing permissions and limitations under the
+ * License.
+ */
+package org.opennms.netmgt.telemetry.daemon;
+
+public enum ConnectorState {
```
cgorantla: Don't think we need to maintain state for Connectors

features/telemetry/daemon/src/main/java/org/opennms/netmgt/telemetry/daemon/ConnectorTwinConfig.java
```diff
@@ -0,0 +1,86 @@
...
+
+public class ConnectorTwinConfig {
+
+    public static String CONNECTOR_KEY="telemetry.connector.config";
+
+    private int nodeId;
+    private String ipAddress;
+    private List<Map<String, String>> parameters;
+    private String connectionKey;
+    private ConnectorState status;
```
cgorantla: wed don't need these two fields.  Also this config should be a list of configs.

cgorantla (commented):
I think we need to revisit this workflow

features/telemetry/daemon/src/main/java/org/opennms/netmgt/telemetry/daemon/ConnectorTwinConfig.java
Mohammad-Junid: Regarding the **connectionKey**, if we remove it, how will we prevent duplicate connectors for a single node? Also, how can we identify a particular connector running for a node that has the OpenConfig service, in order to enable or disable the flows for a specific node?

features/telemetry/daemon/src/main/java/org/opennms/netmgt/telemetry/daemon/ConnectorManager.java
Mohammad-Junid: Done

features/openconfig/telemetry-client/src/main/java/org/opennms/features/openconfig/telemetry/OpenConfigClientImpl.java
Mohammad-Junid: Remove it

features/telemetry/daemon/src/main/java/org/opennms/netmgt/telemetry/daemon/ConnectorState.java
Mohammad-Junid: Remove class

PR #7878, opened 2025-08-18
NMS : 18151 : Update DefaultEventConfDao with loading from database.

Junaid-Khan-Nant:
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18151

### DIFF
--- a/opennms-config/src/main/java/org/opennms/netmgt/config/DefaultEventConfDao.java
+++ b/opennms-config/src/main/java/org/opennms/netmgt/config/DefaultEventConfDao.java
@@ -105,20 +105,12 @@ public class DefaultEventConfDao implements EventConfDao, InitializingBean {
 		}
 	}
 
-	private List<EventConfEvent>  loadFromDB(){
-		ServiceLookupFactory.withService(EventConfEventDao.class, lookup -> {
-			 confEventList = lookup.findAllEvents();
-		});
-		return confEventList;
-	}
-
 	@Override
 	public void reload() throws DataAccessException {
 		validateConfig(m_configResource);
 		try {
-			loadFromDB();
-			processEventFromDb();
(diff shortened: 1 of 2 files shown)


opennms-config/src/main/java/org/opennms/netmgt/config/DefaultEventConfDao.java
```diff
@@ -259,7 +274,9 @@ public void setConfigResource(Resource configResource) throws IOException {
 
 	@Override
 	public void afterPropertiesSet() throws DataAccessException {
-		loadConfig();
+//		loadConfig();
+		loadFromDB();
+		processEventFromDb();
```
cgorantla: we can wrap these two methods and call them asynchronously without blocking `afterPropertiesSet`

opennms-config/src/main/java/org/opennms/netmgt/config/ServiceLookupFactory.java
```diff
@@ -0,0 +1,22 @@
+package org.opennms.netmgt.config;
```
cgorantla, line 22: missing copyright headers

cgorantla:
> Hi @cgorantla, @aurang-zeb313, and @mwajahatabbasi,
> 
> As discussed, I tried persisting all EventConf XML via an API and loaded it using a synchronized service registry lookup. The events are loaded and processed in about 3 to 4 seconds. I've attached a screenshot for reference.
> 
> For now, I’ve skipped the test cases — once the implementation is finalized, we can add the tests accordingly. Screenshot from 2025-08-20 18-11-27 Screenshot from 2025-08-20 19-33-59 Screenshot from 2025-08-21 15-50-32

Thank you.

PR #7898, opened 2025-08-25
FE: Validate event XML files before uploading

atif1019:
Validate event XML files before uploading

Do some validation of any event XML files before uploading.

Validate that they use valid XML

Validate at least some of the important fields expected.

Does not have to be comprehensive validation, that will be done on server side.

This can also be used if a user edits an existing file.
* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18213

### DIFF
--- a/ui/src/components/EventConfiguration/Modal/EventConfigFilesUploadReportModal.vue
+++ b/ui/src/components/EventConfiguration/Modal/EventConfigFilesUploadReportModal.vue
@@ -82,7 +82,7 @@ const gotoViewTab = () => {
 
 <style scoped lang="scss">
 @import "@featherds/styles/themes/variables";
-
+ 
 .text-danger {
   color: var($error);
 }


ui/src/components/EventConfiguration/UploadFilesTab.vue
```diff
@@ -101,23 +108,107 @@ import { useEventConfigStore } from '@/stores/eventConfigStore'
...
+          }
+        }
+        if (eventElements.length > 0) {
+          eventElements.forEach((event, idx) => {
+            const uei = event.querySelector('uei')?.textContent?.trim()
+            const label = event.querySelector('event-label')?.textContent?.trim()
+            const severity = event.querySelector('severity')?.textContent?.trim()
+
+            const eventErrors: string[] = []
+            if (!uei) eventErrors.push('missing <uei>')
```
synqotik: Enclose all `if` blocks in curly braces.

ui/src/components/EventConfiguration/UploadFilesTab.vue
```diff
@@ -101,23 +108,107 @@ import { useEventConfigStore } from '@/stores/eventConfigStore'
...
 const isLoading = ref(false)
 
-const handleEventConfUpload = (e: Event) => {
+const handleEventConfUpload = async (e: Event) => {
   const input = e.target as HTMLInputElement
   if (input.files && input.files.length > 0) {
-    eventFiles.value = Array.from(new Set([...eventFiles.value, ...Array.from(input.files)]))    // You can add further processing of the file here, like uploading it to a server
+    const files = Array.from(input.files)
+
+    for (const file of files) {
```
synqotik, line 134: Let's move this to another file like `validation.ts`.

There can be an exported function which contains the main `for` loop, then another function to validate each specific file.

ui/src/components/EventConfiguration/eventConfigXmlValidator.ts
```diff
@@ -0,0 +1,90 @@
...
+          validationErrors.push('Empty <events> element - no content found')
+        } else {
+          const childNames = Array.from(childElements).map(el => `<${el.tagName}>`).join(', ')
+          validationErrors.push(`<events> element contains ${childNames} but no <event> elements`)
+        }
+      }
+    } else {
+      eventElements.forEach((event, idx) => {
+        const eventErrors = validateEventElement(event, idx + 1)
+        validationErrors.push(...eventErrors)
```
synqotik: Should only push `...eventErrors` if it is non-empty, otherwise it will add an empty array and later on `isValid: validationErrors.length === 0` will be incorrect.

synqotik (approved):
LGTM! 

However please fix the one item before merging, thanks.

PR #7904, opened 2025-08-26
DOCS: add a clarifying note on path outages

dino2gnt:
DOCS DOCS DOCS DOCS DOCS DOCS 

Make a note that Path Outages don't suppress the creation of any events or alarms, only notifications created by notifd.


garcia-sr (approved):
LGTM

PR #7906, opened 2025-08-26
NMS-18276: Move menu template docs to Development section

synqotik:
Move the menu templates documentation from `Operation > Deep Dive > Menus` to `Development > Menus`.

While these can be edited, we are not officially supporting editing them for users, so we want to take this out of `Operation`, but still include it in documentation for reference.

Also removed reference to moving the templates to `etc`, we are holding off on that for now.

Also fixed `userTileProviders` to denote that it is a read-only field and any user edits are ignored.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18274
* * Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18276

### DIFF
--- a/docs/modules/development/nav.adoc
+++ b/docs/modules/development/nav.adoc
@@ -18,6 +18,8 @@
 
 * xref:graph/graph.adoc[]
 
+* xref:menu/menu-templates.adoc[]
+
 * xref:rest/CORS.adoc[]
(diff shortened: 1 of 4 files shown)


cgorantla (approved):
LGTM!

PR #7910, opened 2025-08-27
NMS-18269: Fix double-shift for search input focus

synqotik:
Double-pressing a shift key should focus the main search input in the menu bar.

Just needed to fix the CSS selector used to get the input element.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18269


Shahbaz-dataq (approved):
LGTM!

PR #7914, opened 2025-08-28
FE: Display confirmation dialog when deleting event in EventConf Source

Shahbaz-dataq:
### All Contributors

Display confirmation dialog when deleting event in EventConf Source

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18273


synqotik (approved):
LGTM!

PR #7915, opened 2025-08-28
FE: Validate eventconfig.xml before uploading

atif1019:
Validate that any eventconf.xml is valid before uploading.
Should include XML validation plus schema validation.

* Jira (Issue Tracker): NMS-18260](https://opennms.atlassian.net/browse/NMS-18260)


synqotik (approved):
LGTM!

You can go ahead and merge despite build errors, they are not related to this PR.

We should add some unit tests for the XML validation - another reason to have them in separate `.ts` files.

PR #7917, opened 2025-08-28
NMS-18270 : Provide a REST API endpoint to fetch all event conf sources.

mwajahatabbasi:
Jira : [https://opennms.atlassian.net/browse/NMS-18270](https://opennms.atlassian.net/browse/NMS-18270)

### DIFF
--- a/opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/EventConfSourceDaoHibernate.java
+++ b/opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/EventConfSourceDaoHibernate.java
@@ -29,6 +29,8 @@ import org.slf4j.LoggerFactory;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.stream.Collectors;
 
 public class EventConfSourceDaoHibernate
@@ -106,6 +108,61 @@ public class EventConfSourceDaoHibernate
         LOG.info("Set enabled={} for sources {} (cascadeToEvents={})", enabled, sourceIds, cascadeToEvents);
     }
 
+    @Override
+    public Map<String, Object> filterEventConfSource(final String name, final String vendor, final String desc,
+                                                     final Integer fileOrder, final Integer eventCount,
+                                                     final Integer totalRecords, final Integer offset, Integer limit) {
+
+        int resultCount = (totalRecords != null) ? totalRecords : 0;
+        List<Object> queryParams = new ArrayList<>();
+        List<String> conditions = new ArrayList<>();
+
+        // Add filter conditions dynamically
+        if (name != null && !name.trim().isEmpty()) {
+            conditions.add("lower(s.name) like ? escape '\\'");
+            queryParams.add("%" + escapeLike(name.trim().toLowerCase()) + "%");
+        }
+
+        if (vendor != null && !vendor.trim().isEmpty()) {
+            conditions.add("lower(s.vendor) like ? escape '\\'");
+            queryParams.add("%" + escapeLike(vendor.trim().toLowerCase()) + "%");
+        }
+
+        if (desc != null && !desc.trim().isEmpty()) {
+            conditions.add("lower(s.description) like ? escape '\\'");
+            queryParams.add("%" + escapeLike(desc.trim().toLowerCase()) + "%");
+        }
+
+        if (fileOrder != null && fileOrder > 0) {
+            conditions.add("s.fileOrder = ?");
+            queryParams.add(fileOrder);
+        }
+
+        if (eventCount != null && eventCount > -1) {
+            conditions.add("s.eventCount = ?");
+            queryParams.add(eventCount);
+        }
+
+        String whereClause = conditions.isEmpty() ? "" : " where " + String.join(" AND ", conditions);
+
+        // COUNT QUERY: get total matching records if not already provided
+        if (resultCount == 0) {
+            String countQuery = "select count(s.id) from EventConfSource s " + whereClause;
+            resultCount = super.queryInt(countQuery, queryParams.toArray());
+        }
+
+        // DATA QUERY: fetch paginated results
+        List<EventConfSource> eventConfSourceList = Collections.emptyList();
+        if (resultCount > 0) {
+            String dataQuery = "from EventConfSource s " + whereClause;
+            eventConfSourceList = findWithPagination(dataQuery, queryParams.toArray(), offset, limit);
+        }
+
+        // Return map with results
+        return Map.of("totalRecords", resultCount, "eventConfSourceList", eventConfSourceList);
+
+    }
+
 
     @Override
     public void saveOrUpdate(EventConfSource source) {
@@ -115,5 +172,24 @@ public class EventConfSourceDaoHibernate
     @Override
     public void delete(EventConfSource source) {
         super.delete(source);
+    }
+
+    /**
+     * Escapes special characters (%, _, \, /, [, ]) in a string
+     * to make it safe for SQL LIKE queries.
+     *
+     * @param input the input string
+     * @return the escaped string
+     */
+    private String escapeLike(String input) {
+        return input
+                .replace("\\", "\\\\")
+                .replace("%", "\\%")
+                .replace("_", "\\_")
+                .replace("@", "\\@")
+                .replace("/", "\\/")
+                .replace("[", "\\[")
+                .replace("]", "\\]")
+                .replace(".", "\\.");
     }
 }
--- a/opennms-dao/src/test/java/org/opennms/netmgt/dao/EventConfSourceDaoIT.java
+++ b/opennms-dao/src/test/java/org/opennms/netmgt/dao/EventConfSourceDaoIT.java
@@ -352,4 +352,31 @@ public class EventConfSourceDaoIT implements InitializingBean {
     public void afterPropertiesSet() throws Exception {
         BeanUtils.assertAutowiring(this);
     }
+
+    @Test
+    public void testFilterEventConfSource_ReturnsValidMap() {
+        EventConfSource source =  new EventConfSource();
+        source.setName("testName");
+        source.setFileOrder(1);
+        source.setEventCount(5);
+        source.setEnabled(true);
+        source.setCreatedTime(new Date());
+        source.setLastModified(new Date());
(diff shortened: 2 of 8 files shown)


opennms-webapp-rest/src/test/java/org/opennms/web/rest/v2/EventConfPersistenceServiceIT.java
```diff
@@ -297,4 +298,101 @@ public void testUpdateSourceAndEventEnabled() {
...
+                .fileOrder(10)
+                .username(username)
+                .now(now)
+                .vendor("Juniper")
+                .description("Filter test source")
+                .build();
+
+        Event event = new Event();
+        event.setUei("uei.opennms.org/test/filter");
+        event.setEventLabel("Filter Event");
```
aurang-zeb313: Add test cases  for all filters to validate it properly

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/EventConfRestService.java
```diff
@@ -140,6 +140,33 @@ public Response enableDisableEventConfSources(final EventConfSrcEnableDisablePay
         }
     }
 
+    @Override
+    public Response filterEventConfSource(String name, String vendor, String desc, Integer fileOrder, Integer eventCount
```
aurang-zeb313: Can we use a filter DTO  instead of so many method  params?

aurang-zeb313 (changes requested):

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/EventConfRestService.java
mwajahatabbasi: Since this is a `GET API`, we’re passing filters as query parameters, so the method arguments are defined individually for clarity and `Swagger` documentation.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/EventConfRestService.java
```diff
@@ -40,12 +42,7 @@
 import javax.ws.rs.core.SecurityContext;
 import java.io.InputStream;
 import java.io.Serializable;
-import java.util.Set;
-import java.util.List;
-import java.util.Date;
-import java.util.ArrayList;
-import java.util.Map;
-import java.util.LinkedHashMap;
+import java.util.*;
```
cgorantla: Please don't use wildcard imports

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/EventConfRestService.java
```diff
@@ -140,6 +137,37 @@ public Response enableDisableEventConfSources(final EventConfSrcEnableDisablePay
         }
     }
 
+    @Override
+    public Response filterEventConfSource(String filter, String sortBy, String order, Integer totalRecords,
+                                          Integer offset, Integer limit, SecurityContext securityContext) {
+
+        // Return 400 Bad Request if offset < 0, limit < 1, or offset exceeds totalRecords
+        if (offset < 0 || limit < 1 ||  offset > totalRecords) {
```
cgorantla: do we need null checks for these ?

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/EventConfRestService.java
```diff
@@ -140,6 +137,37 @@ public Response enableDisableEventConfSources(final EventConfSrcEnableDisablePay
         }
     }
 
+    @Override
+    public Response filterEventConfSource(String filter, String sortBy, String order, Integer totalRecords,
```
cgorantla, line 149: why would client send us totalRecords when we can fetch it

mwajahatabbasi: Client needs `totalRecords` for pagination — if  `totalRecord`> 0, we fetch paginated results directly, otherwise we first get the count and then fetch the list.

cgorantla: Do we need UI to send us `totalRecords` or should we fetch it ourselves ?

mwajahatabbasi: We need the `totalRecords` value for pagination. If it’s 0, we fetch the `count` from the database using the applied criteria. Otherwise, there’s no need to query the count every time, when the client navigates between pages (next/previous), we only return the paginated records.

opennms-model/src/main/java/org/opennms/netmgt/model/EventConfSourceDto.java
```diff
@@ -0,0 +1,91 @@
+package org.opennms.netmgt.model;
```
cgorantla: missing copyright headers.

If it is only needed for Rest API, we can move this to opennms-webapp-rest

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/EventConfRestService.java
```diff
@@ -140,6 +143,38 @@ public Response enableDisableEventConfSources(final EventConfSrcEnableDisablePay
...
+            ||  Objects.requireNonNullElse(offset,0) > Objects.requireNonNullElse(totalRecords,0)) {
+            return Response.status(Response.Status.BAD_REQUEST)
+                    .entity(Map.of("error", "Invalid offset/limit values"))
+                    .build();
+        }
+
+        // Call service to fetch results
+        Map<String, Object> result = eventConfPersistenceService.filterEventConfSource(
+                filter, sortBy, order, totalRecords, offset, limit
+        );
```
cgorantla:
```suggestion
        Map<String, Object> result = eventConfPersistenceService.filterEventConfSource(
                filter, sortBy, order, totalRecords, offset, limit);
```

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/EventConfSourceDaoHibernate.java
```diff
@@ -106,6 +108,49 @@ public void updateEnabledFlag(Collection<Long> sourceIds, boolean enabled, boole
         LOG.info("Set enabled={} for sources {} (cascadeToEvents={})", enabled, sourceIds, cascadeToEvents);
     }
 
+    @Override
+    public Map<String, Object> filterEventConfSource(final String filter, final String sortBy, final String order,
+                                                     final Integer totalRecords, final Integer offset, Integer limit) {
+
```
cgorantla, line 115: I think this is prone to SQL injection. Can we rework this to named parameters with HQL. Or use existing CriteriaBuilder

cgorantla (commented):
Also need to merge latest changes

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/EventConfSourceDaoHibernate.java
mwajahatabbasi: You’re right, positional params `(?)` are safe but the order by was risky, so I’ve added `allowedSortFields` and order validation to fully prevent SQL injection.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/EventConfRestService.java
```diff
@@ -142,6 +145,37 @@ public Response enableDisableEventConfSources(final EventConfSrcEnableDisablePay
         }
     }
 
+    @Override
+    public Response filterEventConfSource(String filter, String sortBy, String order, Integer totalRecords,
```
cgorantla, line 149: nit: please format whole method

opennms-dao/src/test/java/org/opennms/netmgt/dao/EventConfSourceDaoIT.java
```diff
@@ -417,4 +417,31 @@ public void testLoadAndPersistMultipleEventConfFilesAndUpdateEnabledFlag() throw
...
+        source.setCreatedTime(new Date());
+        source.setLastModified(new Date());
+        source.setVendor("testVendor");
+        source.setDescription("testDesc");
+        m_dao.saveOrUpdate(source);
+        Map<String, Object> result = m_dao.filterEventConfSource(
+                "testName", "name", "Asc",0, 0, 5);
+        assertNotNull(result);
+        assertEquals(1, result.get("totalRecords"));
+        assertTrue(((List)result.get("eventConfSourceList")).size()>0);
```
cgorantla: nit: please format

opennms-dao/src/test/java/org/opennms/netmgt/dao/EventConfSourceDaoIT.java
```diff
@@ -417,4 +417,31 @@ public void testLoadAndPersistMultipleEventConfFilesAndUpdateEnabledFlag() throw
     public void afterPropertiesSet() throws Exception {
         BeanUtils.assertAutowiring(this);
     }
+
+    @Test
+    public void testFilterEventConfSource_ReturnsValidMap() {
```
cgorantla: Can we write a better comprehensive tests instead of just basic one.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/EventConfRestService.java
```diff
@@ -142,6 +145,37 @@ public Response enableDisableEventConfSources(final EventConfSrcEnableDisablePay
...
+                                          Integer offset, Integer limit, SecurityContext securityContext) {
+
+        // Return 400 Bad Request if offset < 0, limit < 1, or offset exceeds totalRecords
+        if (Objects.requireNonNullElse(offset, 0) < 0 || Objects.requireNonNullElse(limit, 0) < 1
+            || Objects.requireNonNullElse(offset, 0) > Objects.requireNonNullElse(totalRecords, 0)) {
```
cgorantla: offset > 0  and totalRecords being null  should be a valid request, right ?

mwajahatabbasi: Yes you are right, so i am removing this check.

PR #7930, opened 2025-09-03
initial

Shahbaz-dataq:
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18258

### DIFF
--- a/ui/src/components/EventConfiguration/EventConfigSourceTable.vue
+++ b/ui/src/components/EventConfiguration/EventConfigSourceTable.vue
@@ -66,8 +66,8 @@
           >
             <td>{{ config.filename }}</td>
             <td>{{ config.description }}</td>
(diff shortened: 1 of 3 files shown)


ui/src/components/EventConfiguration/EventConfigSourceTable.vue
```diff
@@ -161,8 +161,8 @@ const emptyListContent = {
 const columns = computed(() => [
   { id: 'fileName', label: 'Name' },
   { id: 'description', label: 'Description' },
-  { id: 'fileOrder', label: 'File Order' },
```
synqotik, line 164: I think here we can:

- remove `fileOrder` column entirely
- move `vendor` column between `filename` and `description`

We will only display file order in the screen where user can reorder event config sources.

synqotik (approved):
LGTM!

PR #7931, opened 2025-09-03
FE: Edit an event under a specific source by event and source ID

atif1019:
FE: Edit an event under a specific source by event and source ID

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18265


synqotik (approved):
LGTM!

PR #7939, opened 2025-09-08
NMS-18292 : Expose rest API to return names of Event Conf Sources.

aurang-zeb313:
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18292

### DIFF
--- a/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/EventConfRestService.java
+++ b/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/EventConfRestService.java
@@ -253,7 +253,7 @@ public class EventConfRestService implements EventConfRestApi {
         } catch (Exception ex) {
             return Response.status(Response.Status.INTERNAL_SERVER_ERROR).entity("Unexpected error occurred: " + ex.getMessage()).build();
         }
-
+    }
     @Override
     public Response getEventConfSourcesNames(SecurityContext securityContext) throws Exception {
         try {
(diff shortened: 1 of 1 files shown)


opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/EventConfSourceDaoHibernate.java
```diff
@@ -106,6 +106,17 @@ public void updateEnabledFlag(Collection<Long> sourceIds, boolean enabled, boole
         LOG.info("Set enabled={} for sources {} (cascadeToEvents={})", enabled, sourceIds, cascadeToEvents);
     }
 
+    @Override
+    public List<String> findAllNames() {
+        return findObjects(
+                String.class,
+                "select s.name from EventConfSource s " +
+                        "where lower(s.vendor) != 'opennms' " +
```
cgorantla: Not sure why are we omitting  `opennms` vendor. We need all file names so that user will be present with conflict name when they want to upload new files.

aurang-zeb313: As we discussed, this API will be used when users upload files. Since OpenNMS files are default files and will not be uploaded by users, I think we don’t need to fetch these files from the database.

cgorantla: Yeah, users can try by not knowing that they are already present in the system. So UI need to have all source file names.

PR #7945, opened 2025-09-09
NMS-18307: Docs: Remove references to situation feedback

dino2gnt:
DOCS DOCS DOCS DOCS DOCS

By request, remove references to Situation Feedback in the docs.  API documentation is still there.


marshallmassengill (approved):

dino2gnt:
> I think it would be good to have an issue for these things to explain reasoning.
> 
> I'm not opposed to removing it but would be better if it has some plan.
> 
> Are we going to bring back this into ALEC doc ? Are we going to remove this feature ? ( which would make sense since nobody is actively using it )

This was Marshall's suggestion so he'll have to explain his reasoning, but as far as I am aware, the feedback feature:

- Has zero users,
- has never been functional,
- does not work as documented,
- doesn't actually provide a feedback mechanism for tuning situations; and
- isn't supported

:man_shrugging:

dino2gnt:
> Any reason why are we targeting this to `foundation-2023` ?

Because it's a docs-only change, and the docs for this go back to (at least) foundation-2023.  

>  If we want to deprecate, let's start with Horizon 34

This has nothing to do with the feature itself; deprecating the feature is a separate concern not addressed here.  We're not touching any code, only documentation.

This PR follows the established process for docs-only pull requests, which is why there was no NMS issue and no one from Engineering was tagged for a review.

PR #7955, opened 2025-09-15
FE: Add validation for already uploaded event sources

Shahbaz-dataq:
### Description

Add validation for already uploaded event sources

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18306


ui/src/components/EventConfiguration/EventConfigUploadFilesTab.vue
```diff
@@ -17,9 +17,48 @@
                 <div class="file">
                   <div class="file-icon">
                     <FeatherIcon :icon="Text" />
-                    <span>{{ element.name.length > 39 ? element.name.slice(0, 36) + "..." : element.name }}</span>
+                    <span>
+                      {{ element.file.name.length > 39 ? element.file.name.slice(0, 36) + "..." : element.file.name }}
```
synqotik: This is fine as is, but could also use `ellipsify` in `src/lib/utils.ts`.

synqotik: Actually, I went back and looked at `ellipsify` and realized 2 things:

- it doesn't account for the ellipsis character(s) themselves, which you did 👍
- we should probably use the actual ellipsis character (U2026) instead of 3 dots

I would say use `ellipsify` here, and then either here or in another PR can fix that function to something like:

```typescript
export const ellipsify = (text: string, count: number) => {
  const ellipsis = '\u2026'

  if (text && count && text.length > count) {
    return text.substring(0, count - 1) + ellipsis
  }

  return text
}
```

ui/src/components/EventConfiguration/eventConfigXmlValidator.ts
```diff
@@ -95,9 +97,8 @@ const validateEventElement = (event: Element, eventNumber: number): string[] =>
 
 export const isDuplicateFile = (
   fileName: string,
-  existingFiles: File[],
-  existingInvalidFiles: { name: string; reason: string }[]
+  existingFiles: UploadEventFileType[]
 ): boolean => {
-  return existingFiles.some((f) => f.name === fileName) || existingInvalidFiles.some((f) => f.name === fileName)
+  return existingFiles.some((element) => element.file.name === fileName)
```
synqotik: We should probably do a case-insensitive check here.

```typescript
const lowerFileName = fileName.toLowerCase()

return existingFIles.some((element) => element.file.name.toLowerCase() === lowerFileName)

```

synqotik (approved):
Thanks for the changes, looks good!

Shahbaz-dataq:
Screenshot from 2025-09-17 00-44-43

synqotik (approved):
LGTM!

PR #7958, opened 2025-09-16
NMS-18198: Vaadin wallboard fix, do not display menu

synqotik:
We had to make a change in Vaadin-based pages to display the new Menu when they are the top-level page.

However, when a Vaadin page is embedded within a Wallboard, they were still displaying the menu, so there was the top-level menu and a menu inside the dashlet.

This removes display of the menu inside the dashlet in the Wallboard.

The `fromVaadinDashlet` could possibly be used in a subsequent PR to notify dashlets to reduce left margin/padding.

Also fixed a small issue where when the Wallboard was displayed it messed up the font of the date/time menu display.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18198


synqotik:
Closing this and opening under NMS-18267. Was not able to fix hiding the menu in Geographical Map inside a Wallboard, but fixed the other items.

PR #7959, opened 2025-09-16
NMS-18322: Get protobuf out of grpc shaded bundle

cgorantla:
This solves the linkage errors that we are seeing when twin grpc is getting mixed with telemetry

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18322


christianpape (approved):
LGTM.

PR #7963, opened 2025-09-18
NMS-18267: Fix whitespace in Wallboard after menu redesign

synqotik:
Fix some whitespace issues in dashlets inside a Wallboard that were related to the menu redesign.

We had to make a change in Vaadin-based pages to display the new Menu when they are the top-level page.

However, when a Vaadin page is embedded within a Wallboard, they were still displaying the menu, so there was the top-level menu and a menu inside the dashlet.

This removes display of the menu inside the dashlet in the Wallboard for all pages *except* for the Geographical Map page.

The `fromVaadinDashlet` could possibly be used in a subsequent PR to notify dashlets to reduce left margin/padding.

Removed most of left margin in Wallboard pages, basically went from `27em` to `4em`, and fixed the top margin. This isn't perfect because it is difficult to tell when a page is at the top level, or is a dashlet inside a Wallboard, but this left and top margin work pretty well in both situations.

The menu is still displayed in the Geographical Map inside a Wallboard. There are some issues with this because the geo map is now a Vue component and it is difficult to dynamically display or not display the Menubar. I tried a number of things but couldn't get it to work correctly. We can revisit this issue in the future as it is non-trivial.

Fixed a small issue where when the Wallboard was displayed it messed up the font of the date/time menu display.

See #7958 for additional info. Opened this because that specific issue is not being fixed at this time.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18267
* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18268


synqotik:
@christianpape top margin should be fixed now with the latest commit

christianpape:
> Made updates to fix the top margin in Safari 15-18.
> 
> The Chrome margins and menu bar seem ok on my system (Mac M1).
> 
> ### Chrome:
> menu-wallboard-chrome
> ### Safari:
> menu-wallboard-safari
> ### CSP Issue
> Noticed a CSP issue for the OpenNMS tile server for the geo map, `https://tile.opennms.org`, updated the `web.xml` to have correct settings (need wildcard and non-wildcard versions for part of the host name).

Yes, Safari <= 18 seems ok, upper margin with Safari 26.0 (21622.1.22.11.14) still broken.

synqotik:
It's understood there are additional issues. The pages should at least be usable after this PR is merged. We should open any additional issues under Menu Redesign: Phase 3.

PR #7971, opened 2025-09-22
NMS-18325, NMS-18533: Restrict SCV Rest to admin, fix role check in UI

synqotik:
This PR fixes 2 issues.

Updates Spring security configuration to restrict Rest API requests for SCV to admin role only.

In investigating that, I noticed that we were not properly checking role access in Vue API routing. It was missing setting a flag when we receive the role info.

If a user did not have, say, the `ROLE_ADMIN` role, they would not see a menu item for SCV or other admin-only screens, and would not be able to actually do anything on those pages (well, except for the SCV fix above), but they could still manually type in the correct URL and view the page (though it wouldn't have any real data since Rest API calls would fail). 

This fixes that issue. If a user does not have the correct role, they won't even be able to manually navigate to the UI page.

Note, we need to backport both of these fixes to probably `foundation-2023` and subsequent, will do that separately.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18325
* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18533


cgorantla (approved):
LGTM!

PR #7975, opened 2025-09-23
NMS-18541: Cannot Walk LLDP Remote Link Table for MICROSENS

rssntn67:
The LLDp Table Index has 3 values, {watermark}.{lldplocalportid}.{lldpremportid}.

Depending on the type and version of the Microsens Switch it can happen that the index is only two values.

For example the following row has  index 0.301:
.1.0.8802.1.1.2.1.4.1.1.8.0.301 = STRING: "eth0"

For proper uinderstanding let me show a proper formed row that has a three values index 76.519.4.
.1.0.8802.1.1.2.1.4.1.1.8.76.519.4 = STRING: "28"

Some devices show in the table blank entries like:
.1.0.8802.1.1.2.1.4.1.1.8.0.204 = ""

 This does not correspond to any link and should be avoided

 This fix has two points:

Fixed for managing remote port index stuff assigning default value 0 to lldpremportid
Fixed table blanck entries Avoiding Row with null remote chassisid
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18541


rssntn67:
I guess that the patch is quite simple.
I suggest to merge in the release-34.x branch too.

marshallmassengill (approved):
Changes seem reasonable to me.

features/enlinkd/tests/src/test/java/org/opennms/netmgt/nb/Nms10115NetworkBuilder.java
```diff
@@ -0,0 +1,184 @@
...
+
+import java.net.InetAddress;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * @author <a href="mailto:antonio@opennms.it">Antonio Russo</a>
+ */
+
+public class Nms10115NetworkBuilder extends NmsNetworkBuilder {
```
indigo423, line 35: Just out of curiosity, why is this named Nms10115NetworkBuilder and not Nms18541NetworkBuilder? The issue https://opennms.atlassian.net/browse/NMS-10115 is something totally different.

indigo423 (commented):
Other than the comment seems a pretty great enhancement. Thank you @rssntn67.

features/enlinkd/adapters/collectors/lldp/src/main/java/org/opennms/netmgt/enlinkd/snmp/LldpRemTableTracker.java
```diff
@@ -101,10 +101,16 @@ public LldpRemRow(int columnCount, SnmpInstId instance) {
 		}
 
 	    public Integer getLldpRemLocalPortNum() {
-	    	return getInstance().getSubIdAt(1);
+                return getInstance().getSubIdAt(1);
```
cgorantla, line 104: nit: formatting.

Do we need  `getInstance().getIds().length  >= 2` check here ?

rssntn67: the index as I have state is a formed by three values. 
It is a weird implementation if the values are 2.
So we made the fix to support MICROSENS Weird agent...
I will do a check only if we know there is a weird implementation that has on a malformed index with only one value.

features/enlinkd/tests/src/test/java/org/opennms/netmgt/nb/Nms10115NetworkBuilder.java
rssntn67: is the address of the network devices...10.115.180.*
I got the first numbers...

cgorantla: That's probably little misleading as we usually associate that with NMS issue number.

indigo423: @rssntn67 thanks for clarifying. It's not a blocker to me just something which quickly confuse someone like me :)

rssntn67: This is interesting. It could be my next choise.

rssntn67:
I decided to rebase the changees on release-34, but I have forked only develop.
So let me fork the entire repository and work on the proper branch.

PR #7980, opened 2025-09-24
NMS-18548: Restrict SCV Rest to admin, fix role check in UI

synqotik:
Ensure that the SCV Rest service can only be accessed by users with the admin (`ROLE_ADMIN`) role.

Also ensures that non-admin users cannot manually route to the SCV web UI screen.

This is the same as NMS-18325 and NMS-18533 but backported to `foundation-2024`.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18548


christianpape (approved):
LGTM.

PR #7991, opened 2025-09-29
NMS-18295 : Disable loading of events from filesystem.

mwajahatabbasi:
Jira: [https://opennms.atlassian.net/browse/NMS-18295](https://opennms.atlassian.net/browse/NMS-18295)

### DIFF
--- a/opennms-config/src/main/java/org/opennms/netmgt/config/DefaultEventConfDao.java
+++ b/opennms-config/src/main/java/org/opennms/netmgt/config/DefaultEventConfDao.java
@@ -78,9 +78,9 @@ public class DefaultEventConfDao implements EventConfDao, InitializingBean {
 	private ConfigReloadContainer<Events> m_extContainer;
 
     public DefaultEventConfDao() {
-        this.m_events = new Events();
-        this.m_partition = new EnterpriseIdPartition();
-        this.m_lastModifiedEventFiles = new LinkedHashMap<>();
+        m_events = new Events();
+        m_partition = new EnterpriseIdPartition();
+        m_lastModifiedEventFiles = new LinkedHashMap<>();
         m_events.initialize(m_partition, new EventOrdering());
     }
 
--- a/opennms-config/src/main/java/org/opennms/netmgt/config/DefaultEventConfDao.java
+++ b/opennms-config/src/main/java/org/opennms/netmgt/config/DefaultEventConfDao.java
@@ -77,6 +77,18 @@ public class DefaultEventConfDao implements EventConfDao, InitializingBean {
 
 	private ConfigReloadContainer<Events> m_extContainer;
 
+    public DefaultEventConfDao() {
+        this.m_events = new Events();
+        this.m_partition = new EnterpriseIdPartition();
+        this.m_lastModifiedEventFiles = new LinkedHashMap<>();
+        m_events.initialize(m_partition, new EventOrdering());
+
+        // Initialize extension container so it’s never null
+        this.m_extContainer = new ConfigReloadContainer.Builder<>(Events.class)
+                .withFolder((accumulator, next) -> accumulator.getEvents().addAll(next.getEvents()))
+                .build();
+    }
+
 	public String getProgrammaticStoreRelativeUrl() {
 		return m_programmaticStoreRelativePath;
 	}
@@ -102,12 +114,12 @@ public class DefaultEventConfDao implements EventConfDao, InitializingBean {
 
 	@Override
 	public void reload() throws DataAccessException {
-		validateConfig(m_configResource);
+		/*validateConfig(m_configResource);
 		try {
 		    reloadConfig();
 		} catch (Exception e) {
 			throw new DataRetrievalFailureException("Unable to load " + m_configResource, e);
-		}
+		}*/
 	}
 
 	@Override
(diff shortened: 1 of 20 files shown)


opennms-config/src/main/java/org/opennms/netmgt/config/DefaultEventConfDao.java
```diff
@@ -77,6 +77,13 @@ public class DefaultEventConfDao implements EventConfDao, InitializingBean {
 
 	private ConfigReloadContainer<Events> m_extContainer;
 
+    public DefaultEventConfDao() {
```
cgorantla, line 80: If we require this only for the test cases, we should probably remove it and initialize these separately in another method which can only be used for tests.

opennms-config/src/main/java/org/opennms/netmgt/config/DefaultEventConfDao.java
```diff
@@ -102,12 +109,12 @@ void validateConfig(final Resource resource) throws DataAccessException {
 
 	@Override
 	public void reload() throws DataAccessException {
-		validateConfig(m_configResource);
+		/*validateConfig(m_configResource);
```
cgorantla: we can properly remove this code instead of just commenting

opennms-config/src/main/java/org/opennms/netmgt/config/DefaultEventConfDao.java
```diff
@@ -320,8 +327,8 @@ public void setConfigResource(Resource configResource) throws IOException {
 
 	@Override
 	public void afterPropertiesSet() throws DataAccessException {
-		loadConfig();
```
cgorantla, line 323: we can properly remove code instead of just commenting it

opennms-config/src/main/java/org/opennms/netmgt/config/EventConfUtil.java
```diff
@@ -0,0 +1,120 @@
...
+import org.springframework.dao.DataRetrievalFailureException;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Date;
+import java.util.concurrent.atomic.AtomicLong;
+
+public final class EventConfUtil {
```
cgorantla, line 41: Should this be part of source code when it should only be used by test cases ?

mwajahatabbasi: `EventConfUtil` is used in test cases across multiple modules, so it can’t be moved to `src/test/java`. We’ll keep it here for now and consider moving it to a shared test-utilities module later.

opennms-config/src/main/java/org/opennms/netmgt/config/DefaultEventConfDao.java
mwajahatabbasi: This constructor isn’t limited to test cases; it’s also used in source classes. Removing it may cause `NullPointerException` when `DefaultEventConfDao` is autowired and `m_events` is null.

opennms-config/src/main/java/org/opennms/netmgt/config/DefaultEventConfDao.java
```diff
@@ -77,6 +77,13 @@ public class DefaultEventConfDao implements EventConfDao, InitializingBean {
 
 	private ConfigReloadContainer<Events> m_extContainer;
 
+    public DefaultEventConfDao() {
+        this.m_events = new Events();
+        this.m_partition = new EnterpriseIdPartition();
+        this.m_lastModifiedEventFiles = new LinkedHashMap<>();
+        m_events.initialize(m_partition, new EventOrdering());
```
cgorantla:
```suggestion
        m_events = new Events();
        m_partition = new EnterpriseIdPartition();
        m_events.initialize(m_partition, new EventOrdering());
```

opennms-config/src/main/java/org/opennms/netmgt/config/EventConfUtil.java
```diff
@@ -0,0 +1,120 @@
...
+
+    private static final AtomicLong sourceIdGenerator = new AtomicLong();
+    private static final AtomicLong eventIdGenerator = new AtomicLong();
+
+    public static List<EventConfEvent> parseResourcesAsEventConfEvents(Resource configResource) throws IOException {
```
cgorantla, line 48: Just annotate with @VisibleForTesting. and add a comment that this should only be used in tests.

Same thing for other methods

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/EventConfPersistenceService.java
```diff
@@ -76,7 +76,7 @@ public class EventConfPersistenceService {
     public void init() {
         // Asynchronously load events from DB in order to not to block startup
         // TODO: Uncomment when we are ready to disable loading from filesystem
-        //eventConfExecutor.execute(this::reloadEventsFromDB);
+        eventConfExecutor.execute(this::reloadEventsFromDB);
```
cgorantla, line 80: remove TODO above. Enable other instances of this in this service.

PR #7999, opened 2025-09-30
NMS-18541 Cannot Walk LLDP Remote Link Table for MICROSENS

rssntn67:
The LLDp Table Index has 3 values, {watermark}.{lldplocalportid}.{lldpremportid}.

Depending on the type and version of the Microsens Switch it can happen that the index is only two values.

For example the following row has index 0.301:
.1.0.8802.1.1.2.1.4.1.1.8.0.301 = STRING: "eth0"

For proper uinderstanding let me show a proper formed row that has a three values index 76.519.4.
.1.0.8802.1.1.2.1.4.1.1.8.76.519.4 = STRING: "28"

Some devices show in the table blank entries like:
.1.0.8802.1.1.2.1.4.1.1.8.0.204 = ""

This does not correspond to any link and should be avoided

Fixed lldpremtable two index instead of three
Fixed lldpremtable skipping rows with remote chassis id null 
Created an extended test case with real network walks

As you can see now the test case have the name of the issue to which are related

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18541


christianpape (approved):
LGTM.

PR #8001, opened 2025-09-30
NMS-18552 : Implement sorting and searching functionality for events of a source.

mwajahatabbasi:
Jira : [https://opennms.atlassian.net/browse/NMS-18552](https://opennms.atlassian.net/browse/NMS-18552)

### DIFF
--- a/opennms-dao-api/src/main/java/org/opennms/netmgt/dao/api/EventConfEventDao.java
+++ b/opennms-dao-api/src/main/java/org/opennms/netmgt/dao/api/EventConfEventDao.java
@@ -46,4 +46,6 @@ public interface EventConfEventDao extends OnmsDao<EventConfEvent, Long> {
     Map<String, Object>  findBySourceId(Long sourceId, Integer totalRecords, Integer offset, Integer limit);
 
     void updateEventEnabledFlag(Long sourceId, List<Long> eventIds, boolean enabled);
+
+    Map<String, Object> findBySourceId(Long sourceId, String eventFilter, String eventSortBy, String eventOrder, Integer totalRecords, Integer offset, Integer limit);
 }
--- a/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/EventConfPersistenceService.java
+++ b/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/EventConfPersistenceService.java
@@ -206,8 +206,10 @@ public class EventConfPersistenceService {
         }
     }
 
-    public Map<String, Object> filterConfEventsBySourceId(Long sourceId, Integer totalRecords,  Integer offset, Integer limit) {
-        return eventConfEventDao.findBySourceId(sourceId, totalRecords,  offset, limit);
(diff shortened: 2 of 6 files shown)


opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/EventConfEventDaoHibernate.java
```diff
@@ -78,22 +74,56 @@ public List<EventConfEvent> filterEventConf(final String uei, final String vendo
...
-            String orderBy = " order by e.createdTime desc";
+
+            String orderBy = "";
+            String sortField = eventSortBy;
+
+            String sortOrder = "ASC".equalsIgnoreCase(eventOrder) ? "ASC" : "DESC";
+
+            Set<String> allowedSortFields = Set.of("uei", "eventLabel", "description", "enabled");
+
+            if (!allowedSortFields.contains(eventSortBy)) {
```
cgorantla: Have to validate whether eventSortBy is null otherwise this can throw NPE

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/EventConfEventDaoHibernate.java
```diff
@@ -26,11 +26,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.List;
-import java.util.Map;
-import java.util.Collections;
+import java.util.*;
```
cgorantla: Adjust your  IDE settings, we should never do wildcard imports

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/EventConfEventDaoHibernate.java
mwajahatabbasi: Yes, I have validated it, no NPE occurs since `contains()` is null-safe; if `eventSortBy` is `null`, it simply returns `false`. Moreover, `eventSortBy` is only used in this condition and nowhere else.

cgorantla: /**
     * Returns {@code true} if this set contains the specified element.
     * More formally, returns {@code true} if and only if this set
     * contains an element {@code e} such that
     * {@code Objects.equals(o, e)}.
     *
     * @param o element whose presence in this set is to be tested
     * @return {@code true} if this set contains the specified element
     * @throws ClassCastException if the type of the specified element
     *         is incompatible with this set
     * (<a href="Collection.html#optional-restrictions">optional</a>)
     * @throws NullPointerException if the specified element is null and this
     *         set does not permit null elements
     * (<a href="Collection.html#optional-restrictions">optional</a>)
     */
    boolean contains(Object o);

Inherently contains is not null safe. It is implementation dependent.  It's always best to check for null.

PR #8016, opened 2025-10-06
NMS-18771: don't run fix-karaf-setup for no args / -h / --help

dino2gnt:
Like it says on the tin.

### All Contributors

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18771

### DIFF
--- a/opennms-base-assembly/src/main/filtered/bin/install
+++ b/opennms-base-assembly/src/main/filtered/bin/install
@@ -75,7 +75,7 @@ for i in ${@}; do
                 RUN_FIX_KARAF=false;
         fi
 done
-if [ "$RUN_FIX_KARAF" == true ]; then
+if [ "$RUN_FIX_KARAF" = true ]; then
         echo "Invoking fix-karaf-setup.sh script..."
         "$OPENNMS_BINDIR"/fix-karaf-setup.sh --yes
 fi


marshallmassengill (dismissed):
Good catch.  Ship it!

opennms-base-assembly/src/main/filtered/bin/install
```diff
@@ -67,8 +67,18 @@ for opt in $ADDITIONAL_MANAGER_OPTIONS; do
 	esac
 done
 
-echo "Invoking fix-karaf-setup.sh script..."
-"$OPENNMS_BINDIR"/fix-karaf-setup.sh --yes
+RUN_FIX_KARAF=false;
+for i in ${@}; do
+        if [ "$i" != "--help" -a $i != "-h" ]; then
```
cgorantla:
```suggestion
        if [ "$i" != "--help" -a "$i" != "-h" ]; then
```

opennms-base-assembly/src/main/filtered/bin/install
```diff
@@ -67,8 +67,18 @@ for opt in $ADDITIONAL_MANAGER_OPTIONS; do
...
+                RUN_FIX_KARAF=true;
+        else
+                RUN_FIX_KARAF=false;
+        fi
+done
+if [ $RUN_FIX_KARAF == true ]; then
```
cgorantla:
```suggestion
if [ "$RUN_FIX_KARAF" = true ]; then
```

opennms-base-assembly/src/main/filtered/bin/install
```diff
@@ -67,8 +67,18 @@ for opt in $ADDITIONAL_MANAGER_OPTIONS; do
 	esac
 done
 
-echo "Invoking fix-karaf-setup.sh script..."
-"$OPENNMS_BINDIR"/fix-karaf-setup.sh --yes
+RUN_FIX_KARAF=false;
+for i in ${@}; do
+        if [ "$i" != "--help" -a $i != "-h" ]; then
+                RUN_FIX_KARAF=true;
```
cgorantla, line 73: Is there a chance that people may use `-h`, `--help` with other arguments, then RUN_FIX_KARAF may set to true, if there are other arguments  after  `-h` or `--help`

dino2gnt: There's absolutely that chance, yes, but I don't see a straightforward route to avoid that problem.

cgorantla: add break; after `RUN_FIX_KARAF=false;`   - Wouldn't that fix the issue ?

dino2gnt: Only for the case where the args are like `-d -i -s -h -f`, but I don't use it that way, therefore no one uses it that way :)

Doing something like `"$i" =~ [.*h.*]` would likely run afoul of any switch that takes an argument.

opennms-base-assembly/src/main/filtered/bin/install
```diff
@@ -67,8 +67,18 @@ for opt in $ADDITIONAL_MANAGER_OPTIONS; do
...
-"$OPENNMS_BINDIR"/fix-karaf-setup.sh --yes
+RUN_FIX_KARAF=false;
+for i in ${@}; do
+        if [ "$i" != "--help" -a "$i" != "-h" ]; then
+                RUN_FIX_KARAF=true;
+        else
+                RUN_FIX_KARAF=false;
+        fi
+done
+if [ "$RUN_FIX_KARAF" == true ]; then
```
cgorantla:
```suggestion
if [ "$RUN_FIX_KARAF" = true ]; then
```
for consistency.

dino2gnt: At least it'll be wrong consistently ;-)

cgorantla (approved):
LGTM!

PR #8018, opened 2025-10-06
NMS-18531 :  Use Environment variables in metadata 

cgorantla:
This should work in opennms-datasources.xml and  system properties.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18531

### DIFF
--- a/core/db/src/main/java/org/opennms/netmgt/config/opennmsDataSources/ConnectionPool.java
+++ b/core/db/src/main/java/org/opennms/netmgt/config/opennmsDataSources/ConnectionPool.java
@@ -30,6 +30,13 @@ import javax.xml.bind.annotation.XmlAccessorType;
 import javax.xml.bind.annotation.XmlAttribute;
 import javax.xml.bind.annotation.XmlRootElement;
 
+import org.opennms.core.mate.api.EnvironmentScope;
+import org.opennms.core.mate.api.FallbackScope;
+import org.opennms.core.mate.api.Interpolator;
+import org.opennms.core.mate.api.SecureCredentialsVaultScope;
+import org.opennms.features.scv.api.SecureCredentialsVault;
+import org.opennms.features.scv.jceks.JCEKSSecureCredentialsVault;
+
 /**
  * Database connection pool configuration.
  * 
@@ -52,31 +59,31 @@ public class ConnectionPool implements java.io.Serializable {
      * removed.
      */
     @XmlAttribute(name = "idleTimeout")
-    private Integer idleTimeout;
+    private String rawIdleTimeout;
 
     /**
      * How long, in seconds, to attempt to make a connection to the database.
      */
     @XmlAttribute(name = "loginTimeout")
-    private Integer loginTimeout;
+    private String rawLoginTimeout;
 
     /**
      * The minimum number of pooled connections to retain.
      */
     @XmlAttribute(name = "minPool")
-    private Integer minPool;
+    private String rawMinPool;
 
     /**
      * The maximum number of pooled connections to retain.
      */
     @XmlAttribute(name = "maxPool")
-    private Integer maxPool;
+    private String rawMaxPool;
 
     /**
      * The maximum number of connections that can be created.
      */
     @XmlAttribute(name = "maxSize")
-    private Integer maxSize;
+    private String rawMaxSize;
 
     public ConnectionPool() {
         setFactory("org.opennms.core.db.HikariCPConnectionFactory");
@@ -85,36 +92,36 @@ public class ConnectionPool implements java.io.Serializable {
     /**
      */
     public void deleteIdleTimeout() {
-        this.idleTimeout= null;
+        this.rawIdleTimeout = null;
     }
 
     /**
      */
     public void deleteLoginTimeout() {
-        this.loginTimeout= null;
+        this.rawLoginTimeout = null;
     }
 
     /**
      */
     public void deleteMaxPool() {
-        this.maxPool= null;
+        this.rawMaxPool = null;
     }
 
     /**
      */
     public void deleteMaxSize() {
-        this.maxSize= null;
+        this.rawMaxSize = null;
     }
 
     /**
      */
     public void deleteMinPool() {
-        this.minPool= null;
+        this.rawMinPool = null;
     }
 
     /**
      * Overrides the Object.equals method.
-     * 
+     *
      * @param obj
      * @return true if the objects are equal.
      */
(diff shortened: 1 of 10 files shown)


core/db/src/main/java/org/opennms/netmgt/config/opennmsDataSources/ConnectionPool.java
```diff
@@ -151,112 +174,117 @@ public String getFactory() {
...
      * following description: How long, in seconds, an idle connection is kept in
      * the pool before it is removed.
-     * 
+     *
      * @return the value of field 'IdleTimeout'.
      */
     public Integer getIdleTimeout() {
-        return this.idleTimeout != null ? this.idleTimeout : Integer.valueOf("600");
+        String interpolated = interpolateAttribute(this.rawIdleTimeout);
+        return interpolated != null ? Integer.valueOf(interpolated) : Integer.valueOf("600");
```
christianpape:
```suggestion
        return interpolated != null ? Integer.valueOf(interpolated) : 600;
```

core/db/src/main/java/org/opennms/netmgt/config/opennmsDataSources/ConnectionPool.java
```diff
@@ -151,112 +174,117 @@ public String getFactory() {
...
      * following description: How long, in seconds, to attempt to make a
      * connection to the database.
-     * 
+     *
      * @return the value of field 'LoginTimeout'.
      */
     public Integer getLoginTimeout() {
-        return this.loginTimeout != null ? this.loginTimeout : Integer.valueOf("3");
+        String interpolated = interpolateAttribute(this.rawLoginTimeout);
+        return interpolated != null ? Integer.valueOf(interpolated) : Integer.valueOf("3");
```
christianpape:
```suggestion
        return interpolated != null ? Integer.valueOf(interpolated) : 3;
```

core/db/src/main/java/org/opennms/netmgt/config/opennmsDataSources/ConnectionPool.java
```diff
@@ -151,112 +174,117 @@ public String getFactory() {
...
      * Returns the value of field 'maxPool'. The field 'maxPool' has the following
      * description: The maximum number of pooled connections to retain.
-     * 
+     *
      * @return the value of field 'MaxPool'.
      */
     public Integer getMaxPool() {
-        return this.maxPool != null ? this.maxPool : Integer.valueOf("50");
+        String interpolated = interpolateAttribute(this.rawMaxPool);
+        return interpolated != null ? Integer.valueOf(interpolated) : Integer.valueOf("50");
```
christianpape:
```suggestion
        return interpolated != null ? Integer.valueOf(interpolated) : 50;
```

core/db/src/main/java/org/opennms/netmgt/config/opennmsDataSources/ConnectionPool.java
```diff
@@ -151,112 +174,117 @@ public String getFactory() {
...
      * Returns the value of field 'maxSize'. The field 'maxSize' has the following
      * description: The maximum number of connections that can be created.
-     * 
+     *
      * @return the value of field 'MaxSize'.
      */
     public Integer getMaxSize() {
-        return this.maxSize != null ? this.maxSize : Integer.valueOf("500");
+        String interpolated = interpolateAttribute(this.rawMaxSize);
+        return interpolated != null ? Integer.valueOf(interpolated) : Integer.valueOf("500");
```
christianpape:
```suggestion
        return interpolated != null ? Integer.valueOf(interpolated) : 500;
```

core/db/src/main/java/org/opennms/netmgt/config/opennmsDataSources/ConnectionPool.java
```diff
@@ -151,112 +174,117 @@ public String getFactory() {
...
      * Returns the value of field 'minPool'. The field 'minPool' has the following
      * description: The minimum number of pooled connections to retain.
-     * 
+     *
      * @return the value of field 'MinPool'.
      */
     public Integer getMinPool() {
-        return this.minPool != null ? this.minPool : Integer.valueOf("10");
+        String interpolated = interpolateAttribute(this.rawMinPool);
+        return interpolated != null ? Integer.valueOf(interpolated) : Integer.valueOf("10");
```
christianpape:
```suggestion
        return interpolated != null ? Integer.valueOf(interpolated) : 10;
```

christianpape (changes requested):

indigo423 (commented):
Can we use the following variable names, that way they fit into the existing naming pattern and we don't have to change docs and some things we used in some workarounds in our Helm Chart:

```
OPENNMS_DATABASE_CONNECTION_IDLETIMEOUT
OPENNMS_DATABASE_CONNECTION_LOGINTIMEOUT
OPENNMS_DATABASE_CONNECTION_MINPOOL
OPENNMS_DATABASE_CONNECTION_MAXPOOL
OPENNMS_DATABASE_CONNECTION_MAXSIZE
```

indigo423:
@cgorantla Alejandro added also two variables with these defaults in our Helm chart with the following defaults:

```
POSTGRES_SSL_FACTORY="org.postgresql.ssl.LibPQFactory"
POSTGRES_SSL_MODE="prefer"
```
he uses these variables in the JDBC connection URL like this:

```
  <jdbc-data-source name="opennms" 
                    database-name="${OPENNMS_DBNAME}"
                    class-name="org.postgresql.Driver"
                    url="jdbc:postgresql://${POSTGRES_HOST}:${POSTGRES_PORT}/${OPENNMS_DBNAME}?sslmode=${POSTGRES_SSL_MODE}&amp;sslfactory=${POSTGRES_SSL_FACTORY}"
                    user-name="${OPENNMS_DBUSER}"
                    password="${OPENNMS_DBPASS}" />
```

Info: https://www.postgresql.org/docs/current/libpq-ssl.html

core/db/src/main/resources/xsds/opennms-datasources.xsd
```diff
@@ -30,19 +30,19 @@
       <attribute name="factory" type="string" use="optional" default="org.opennms.core.db.HikariCPConnectionFactory">
         <annotation><documentation>The connection pool implementation to use.</documentation></annotation>
       </attribute>
-      <attribute name="idleTimeout" type="int" use="optional" default="600">
+      <attribute name="idleTimeout" type="string" use="optional" default="600">
```
indigo423, line 33: No blocker to me, just out of curiosity, is there a reason we can't validate against a more strict integer instead of a string? It would prevent us to catch errors in case a user writes something else than an integer in this field, isn't it?

cgorantla: Because we now need to allow metadata here, it has to be string type. We can only interpolate once we read the attribute.

cgorantla:
> Can we use the following variable names, that way they fit into the existing naming pattern and we don't have to change docs and some things we used in some workarounds in our Helm Chart:
> 
> ```
> OPENNMS_DATABASE_CONNECTION_IDLETIMEOUT
> OPENNMS_DATABASE_CONNECTION_LOGINTIMEOUT
> OPENNMS_DATABASE_CONNECTION_MINPOOL
> OPENNMS_DATABASE_CONNECTION_MAXPOOL
> OPENNMS_DATABASE_CONNECTION_MAXSIZE
> ```

It shouldn't matter what name we are using. It will interpolate to assigned attribute.  The change `smoke-test/src/main/resources/opennms-overlay/etc/opennms-datasources.xml`  is only used for smoke tests.

cgorantla:
> @cgorantla Alejandro added also two variables with these defaults in our Helm chart with the following defaults:
> 
> ```
> POSTGRES_SSL_FACTORY="org.postgresql.ssl.LibPQFactory"
> POSTGRES_SSL_MODE="prefer"
> ```
> 
> he uses these variables in the JDBC connection URL like this:
> 
> ```
>   <jdbc-data-source name="opennms" 
>                     database-name="${OPENNMS_DBNAME}"
>                     class-name="org.postgresql.Driver"
>                     url="jdbc:postgresql://${POSTGRES_HOST}:${POSTGRES_PORT}/${OPENNMS_DBNAME}?sslmode=${POSTGRES_SSL_MODE}&amp;sslfactory=${POSTGRES_SSL_FACTORY}"
>                     user-name="${OPENNMS_DBUSER}"
>                     password="${OPENNMS_DBPASS}" />
> ```
> 
> Info: https://www.postgresql.org/docs/current/libpq-ssl.html
> 
> Hint: He uses "require" for SSL on PostgreSQL because there is a TLS cert manager involved. I think allowing also non-SSL connections with "prefer" is less error prune and gives us some backward compatibility.

Same  thing for this one. Whatever environment variables used in url, they should resolve and map to url. No changes needed.

christianpape (approved):
LGTM.

PR #8104, opened 2025-11-06
Update page-size configuration in newts.adoc

sbernabe1:
Updating doc's to mention the following below:

Setting the page-size to 0 is recommended to disable query paging, avoiding unexpected gaps in graph data.

### DIFF
--- a/docs/modules/deployment/pages/time-series-storage/newts/newts.adoc
+++ b/docs/modules/deployment/pages/time-series-storage/newts/newts.adoc
@@ -54,7 +54,7 @@ datastax-java-driver {
     local-datacenter = datacenter1<3>
   }
   basic.request {
-    page-size = 20000<4>
+    page-size = 0<4>
   }
   advanced.auth-provider {
     class = PlainTextAuthProvider<5>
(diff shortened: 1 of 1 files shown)


dino2gnt:
Ship it

dino2gnt (dismissed):

cgorantla:
@sbernabe1  Are we ready to merge this ?   Just curious, Is there some more explanation on why this recommendation has changed.

dino2gnt:
> @sbernabe1 Are we ready to merge this ? Just curious, Is there some more explanation on why this recommendation has changed.

@cgorantla 
AFAIK, OpenNMS doesn't have any support for the pagination added in the Cassandra 4.x client.  Previously, we specified a large value for `page-size` and crossed our fingers that it was large enough to hold the entire reply. However, Support encountered a situation where our previous recommendation was not enough, and in fact even going upwards of 1million was still not holding the entire reply.  Saul discovered that `page-size=0` was available to turn off the pagination entirely, which fixed the problem for the customer, and so we're updating the documentation to reflect the new information.

marshallmassengill:
Closing as this appears to be in the docs now.

PR #8316, opened 2026-02-25
NMS-19537: Reworked Interpolator to create Scope on demand

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-19537


indigo423:
Just thinking out loud. The Metadata interpolation happens also in the service monitoring framework in Pollderd and has a similar characteristic. Would we have to re-implement a similar cache there as well?

cgorantla (commented):
It would be good to separate ScopeProvider changes from the cache changes.


Also we should avoid reordering imports for easier diff

PR #8319, opened 2026-02-26
NMS-19560: Pre populate metadata for each service

cgorantla:
Retrieve metadata once for each  Poller request instead of retrieving it for each RPC invoke.
Whenever metadata updates, we will refresh the request and interpolate those again.
Runtime attributes will still be called for each RPC request

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19560

### DIFF
--- a/opennms-services/src/main/java/org/opennms/netmgt/poller/pollables/PollableServiceConfig.java
+++ b/opennms-services/src/main/java/org/opennms/netmgt/poller/pollables/PollableServiceConfig.java
@@ -22,6 +22,7 @@
 package org.opennms.netmgt.poller.pollables;
 
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
@@ -30,6 +31,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 
 import org.opennms.core.rpc.api.RpcExceptionHandler;
+import org.opennms.core.sysprops.SystemProperties;
 import org.opennms.core.rpc.api.RpcExceptionUtils;
 import org.opennms.netmgt.collection.api.PersisterFactory;
 import org.opennms.netmgt.config.PollerConfig;
@@ -56,9 +58,21 @@ import org.slf4j.LoggerFactory;
 public class PollableServiceConfig implements PollConfig, ScheduleInterval {
     private static final Logger LOG = LoggerFactory.getLogger(PollableServiceConfig.class);
 
+    public static final String METADATA_CACHE_TTL_PROPERTY = "org.opennms.netmgt.poller.metadataCacheTtlMs";
+
+    /**
+     * Upper bound on the age of the cached, interpolated service parameters. Explicit
+     * invalidation events are the fast path; this backstop guarantees that scope changes
+     * without a corresponding event (e.g. credential rotation) are picked up eventually.
+     * A value of zero or less disables the backstop.
+     */
+    private final long m_metadataCacheTtlMs =
+            SystemProperties.getLong(METADATA_CACHE_TTL_PROPERTY, TimeUnit.MINUTES.toMillis(15));
+
     private PollerConfig m_pollerConfig;
     private PollableService m_service;
     private Map<String,Object> m_parameters = null;
+    private long m_parametersTimestamp = 0;
     private Package m_pkg;
     private Timer m_timer;
     private Service m_configService;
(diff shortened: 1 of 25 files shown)


cgorantla (commented):
I haven't tested this thoroughly that's why I kept it in Draft but I think this should work. I guess we can achieve similar pattern on Collectd/Thresholding.

opennms-services/src/main/java/org/opennms/netmgt/poller/pollables/PollableServiceConfig.java
```diff
@@ -190,11 +206,23 @@ public synchronized void refresh() {
...
+        m_parameters = null;
     }
 
     private synchronized Map<String,Object> getParameters() {
         if (m_parameters == null) {
-            m_parameters = m_configService.getParameterMap();
+            final Map<String, Object> raw = m_configService.getParameterMap();
+            if (m_entityScopeProvider != null) {
+                final Scope scope = new FallbackScope(
+                        m_entityScopeProvider.getScopeForNode(m_service.getNodeId()),
```
christianpape: Our current implementation of a Scope retrieves for instance a node entity and add calls to the entity's properties. So, it did not fill in strings but calls the entity's method on demand. This means DAO objects are kept inside the scope and additional handling for node updates isn't necessary I think. My attempt to keep MapScope objects in the cache was not as successful as I had hoped. Unfortunately, this approach created additional load on the database whenever the cache was refreshed.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/NodeIpInterfacesRestService.java
```diff
@@ -236,6 +238,7 @@ public Response deleteMetaData(@Context final UriInfo uriInfo, @PathParam("ipAdd
             }
             intf.removeMetaData(context);
             getDao().update(intf);
+            sendNodeMetadataUpdatedEvent(intf);
```
christianpape: A poller scope consist of a service scope, an interface scope and a node scope. The node scope includes asset information and stuff like sys-object-id, the interface scope things like if-description. So, just handling the change of meta-data isn't enough I think. We also need to send events when node, interfaces and services are altered via ReST. Also a simple rescan can change some of the node's fields.

opennms-services/src/main/java/org/opennms/netmgt/poller/pollables/PollableServiceConfig.java
```diff
@@ -190,11 +206,23 @@ public synchronized void refresh() {
...
+                        m_entityScopeProvider.getScopeForInterface(m_service.getNodeId(), m_service.getIpAddr()),
+                        m_entityScopeProvider.getScopeForService(m_service.getNodeId(), m_service.getAddress(), m_service.getSvcName()),
```
christianpape: Even in this case scopes and interfaces are retrieved more than neccessary. Of course, this will happen only when scheduling, but it isn't distributed over the different poller calls over time. This may have an impact on scheduled provisiond runs.

christianpape (commented):
I really don't know if it's a good idea to tackle the performance problem in this way. For me, it carries too many risks of indeterminate behaviour, which is then very difficult to investigate. 

My idea for optimisation is as follows. Currently, the interpolator methods work with scope instances. I would change this to ScopeProvider, i.e. the scope is not always generated, but only when a string actually contains a metadata expression. If it works as I imagine, this would eliminate all getScopeFor...() calls for the thresholds, as we do not have any metadata expressions there in the default configuration. And even for other spots in OpenNMS this will eliminate a lot of calls to the EntityScopeProvider.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/NodeIpInterfacesRestService.java
cgorantla: Yeah, two events I'm keeping track of, node updated and node scan completed. Both of these can take care of all the updates. If we gap somewhere in our Rest API, we can add those.

opennms-services/src/main/java/org/opennms/netmgt/poller/pollables/PollableServiceConfig.java
cgorantla: Once refresh happens, we clear current interpolated parameters and `getParameters` happens on it's own schedule. We are not fetching anything that is not necessary. It's the same code in PollerRequestBuilder that moved to here.  Instead of doing it for each service schedule, we are doing it once at the beginning. Now parameters are holding actual interpolated params instead or raw.

cgorantla:
> I really don't know if it's a good idea to tackle the performance problem in this way. For me, it carries too many risks of indeterminate behaviour, which is then very difficult to investigate.
> 
> My idea for optimisation is as follows. Currently, the interpolator methods work with scope instances. I would change this to ScopeProvider, i.e. the scope is not always generated, but only when a string actually contains a metadata expression. If it works as I imagine, this would eliminate all getScopeFor...() calls for the thresholds, as we do not have any metadata expressions there in the default configuration. And even for other spots in OpenNMS this will eliminate a lot of calls to the EntityScopeProvider.

yeah, that makes sense. That should definitely help. If we don't have metadata expressions, we don't load any scopes and DB calls. 

My current implementation in this PR moves interpolation from every poll cycle to once per service at the start. And for every node update and node scan complete, we clear the parameters. And next service schedule would do interpolation again. This doesn't introduce any new bottlenecks and DB calls are still per-service and still staggered by scheduler .  Only thing is that we need to ensure all node metadata updates are taken care by events. We already have event based lifecycle updates in Polling/Collectd and Threshold and we should be able to handle this.

We can test all of this more broadly and decide on next steps.

PR #8331, opened 2026-03-03
NMS-19573: Creating metadata cache

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-19573

### DIFF
--- a/core/mate/model/src/main/java/org/opennms/core/mate/model/CachedEntityScopeProviderImpl.java
+++ b/core/mate/model/src/main/java/org/opennms/core/mate/model/CachedEntityScopeProviderImpl.java
@@ -1,0 +1,173 @@
+/*******************************************************************************
+ * This file is part of OpenNMS(R).
+ *
+ * Copyright (C) 2019 The OpenNMS Group, Inc.
+ * OpenNMS(R) is Copyright (C) 1999-2019 The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is a registered trademark of The OpenNMS Group, Inc.
+ *
+ * OpenNMS(R) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License,
+ * or (at your option) any later version.
+ *
+ * OpenNMS(R) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with OpenNMS(R).  If not, see:
+ *      http://www.gnu.org/licenses/
+ *
+ * For more information contact:
+ *     OpenNMS(R) Licensing <license@opennms.org>
+ *     http://www.opennms.org/
+ *     http://www.opennms.com/
+ *******************************************************************************/
+
+package org.opennms.core.mate.model;
+
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+import org.opennms.core.mate.api.EntityScopeProvider;
+import org.opennms.core.mate.api.Scope;
+
+import java.net.InetAddress;
+import java.util.Objects;
+import java.util.concurrent.TimeUnit;
+
+public class CachedEntityScopeProviderImpl implements EntityScopeProvider {
+
+    private final static class Tuple<A, B> {
+        final A a;
+        final B b;
+
+        public Tuple(A a, B b) {
+            this.a = a;
+            this.b = b;
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (o == null || getClass() != o.getClass()) return false;
+            Tuple<?, ?> tuple = (Tuple<?, ?>) o;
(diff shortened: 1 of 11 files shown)


cgorantla (commented):
(review retracted)

marshallmassengill:
I updated the title to fix the typo on the NMS issue, just FYI.  I'm reviewing this one now.

marshallmassengill (changes requested):
A few things for the review:

There are some updates for F2024 that will make it so this doesn't build as it is.

Something to check is where this is wired in.  I think Core gets this from Spring (applicationContext-entity-scope-provider.xml declares plain EntityScopeProviderImpl as the primary bean).  Suggestion from Claude here:

`The cache belongs at the Spring primary bean, where it also helps pollerd's residual per-poll builds after NMS-20009 and collectd's per-cycle parameter interpolation. Also, look at NMS-20007, where thresholding will build at most one scope per service per cycle.`

It's got unbounded caches of Hibernate entity graphs by default. Default maximumSize=-1 means four unbounded caches, each value an ObjectScope pinning a loaded node, its ~60-field asset record, and metadata collections for a rolling 10 minutes. This should probably be a bounded default, and ideally the cached value should be a materialized MapScope snapshot rather than entity-wrapping scopes to also remove the detached-lazy-proxy hazard.

EntityScopeProviderImpl.getScopeForNode(null) returns EmptyScope.EMPTY; the decorator calls nodeScopes.getUnchecked(null) and Guava throws NPE on null key.

CachedEntityScopeProviderIT extends EntityScopeProviderIT and constructs the decorator with expireAfterWrite=0 — immediate expiry, caching effectively disabled. It proves delegation preserves semantics, which is worth having, but the suite passes even if the cache never caches.  This is missing a second-lookup-doesn't-invoke-the-loader test, expiry behavior, and the null-key case.

refreshAfterWrite without an async reload() override refreshes synchronously on the caller's thread — the knob implies more than it delivers. 
  
No recordStats()/metrics for what is explicitly a performance feature, so field hit-rates are unobservable. 

maximumSize=0 also means unbounded (> 0 check) — doc trap.

Check the license header.

marshallmassengill (changes requested):
Not sure if this is intentional but the cfg file only tunes the blueprint instance; the Spring instance (Collectd, poller client, core thresholding) is hardcoded 900s and can't be disabled.

These seem like they are worth looking at:
  - refreshAfterWrite=0 throws at construction (Guava requires > 0; guard is >= 0).
  - testCache is timing-sensitive (5s window across DB saves) — a Ticker constructor arg would make it deterministic.

We also may want to put something in the release notes since this is a behavior change I would think.

features/collection/thresholding/impl/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -17,13 +17,32 @@
         <property name="sessionUtils" ref="sessionUtils"/>
         <property name="scv" ref="secureCredentialsVault"/>
     </bean>
-    
+
+    <cm:property-placeholder id="org.opennms.netmgt.threshd.metadata.cache"
```
cgorantla, line 21: This is only relevant in Sentinel

core/mate/model/src/main/java/org/opennms/core/mate/model/CachedEntityScopeProviderImpl.java
```diff
@@ -0,0 +1,184 @@
...
+
+import java.net.InetAddress;
+import java.util.Objects;
+import java.util.concurrent.TimeUnit;
+
+import static org.opennms.core.mate.api.EmptyScope.EMPTY;
+
+public class CachedEntityScopeProviderImpl implements EntityScopeProvider {
+
+    private final static class Tuple<A, B> {
```
cgorantla: Can we use apache Pair/Triple if they are already on the classpath.

core/mate/model/src/main/resources/META-INF/opennms/applicationContext-entity-scope-provider.xml
```diff
@@ -13,7 +13,21 @@
...
-    <bean name="entityScopeProvider" class="org.opennms.core.mate.model.EntityScopeProviderImpl" primary="true"/>
+    <bean name="entityScopeProvider" class="org.opennms.core.mate.model.EntityScopeProviderImpl" primary="true">
+        <qualifier value="entityScopeProvider"/>
+    </bean>
 
     <onmsgi:service ref="entityScopeProvider" interface="org.opennms.core.mate.api.EntityScopeProvider"/>
+
+    <bean name="cachedEntityScopeProvider" class="org.opennms.core.mate.model.CachedEntityScopeProviderImpl">
+        <qualifier value="cachedEntityScopeProvider"/>
+        <constructor-arg ref="entityScopeProvider"/>
```
cgorantla, line 24: instead of configuring fixed values here, may be we can just configure `expireAfterWrite` as system property and other values as constants in Java itself.

cgorantla (commented):
This probably can be termed as metadata cache for all components not just for Thresholding.

core/mate/model/src/main/java/org/opennms/core/mate/model/CachedEntityScopeProviderImpl.java
christianpape: done.

core/mate/model/src/main/resources/META-INF/opennms/applicationContext-entity-scope-provider.xml
christianpape: done.

PR #8366, opened 2026-03-18
NMS-19166: Mib compiler rest implementation

aurang-zeb313:
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19166


cgorantla:
should be handled by https://github.com/OpenNMS/opennms/pull/8369

PR #8379, opened 2026-03-20
NMS-19541: Minion: Investigate if its possible to move away from Confd

mershad-manesh:
Opening this PR to gather feedback;

### All Contributors

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19541


cgorantla (commented):
1. I think we should move templates folder which has boot files to something outside etc to avoid confusion with `featuresBoot.d`.  For ex:  `${MINION_HOME}/boot-templates/` 
We can keep current `MINION_IPC=kafka` logic where we disable JMS and enable Kafka.

2.  I think we can avoid shipping with .cfg files. Let user customize. We can create all these .cfg files in smoke-test overlay for smoke tests to pass.

3. When user creates their own docker compose or helm charts, they can create their own version of cfg files with custom environment variables. We may not need to enforce this.

PR #8399, opened 2026-03-27
NMS-19137: Rerun tests and move Integration tests to large

cgorantla:
This PR back ports https://github.com/OpenNMS/opennms/pull/8362
and https://github.com/OpenNMS/opennms/pull/8382

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19612
* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19614


marshallmassengill (changes requested):
Two Blockers

  - A retried pass can hide tests that never ran. With retries on, the first pass uses --fail-at-end, which skips every module depending on a failed one. The retry re-runs only the failed classes, so the skipped modules' tests never execute and the job goes green. Reproduced with surefire 3.2.5: a genuinely broken test in a dependent module never ran, final exit 0. Fix: first pass with --fail-never, derive the failure signal from the report scan instead of Maven's exit code.
  - Flaky evidence is destroyed. The retry deletes the failing TEST-*.xml before re-running, so uploaded results and Test Insights only ever see the passing retry. Reproduced: the only report left afterwards has failures="0". Develop fixed this in PR 8404 (NMS-19648, merged 2026-04-10); foundation-2024 has none of it. Fold in the itest.sh, smoke.sh, and yml hunks from 8404 or land them immediately after.

Some other things we may want to address:
 - The enlinkd log4j2-test.xml is a dead file.
 - Systemic failures now cost a full run plus a full retry. Dropping -Dfailsafe.skipAfterFailureCount=1 means a dead Postgres or Docker runs every IT to failure, then reruns them all.  This may be intentional.

PR #8422, opened 2026-04-07
NMS-19396: Allow compilation with JDK21 

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-19396

### DIFF
--- a/features/newts-repository-converter/src/main/java/org/opennms/features/newts/converter/NewtsConverter.java
+++ b/features/newts-repository-converter/src/main/java/org/opennms/features/newts/converter/NewtsConverter.java
@@ -104,7 +104,6 @@ public class NewtsConverter implements AutoCloseable {
 
     private final static Timestamp EPOCH = Timestamp.fromEpochMillis(0);
     private final static ValueType<?> ZERO = ValueType.compose(0, MetricType.GAUGE);
-    private final static List<Runnable> EXIT_HOOKS = new ArrayList<>();
 
     private enum StorageStrategy {
         STORE_BY_METRIC,
@@ -155,6 +154,12 @@ public class NewtsConverter implements AutoCloseable {
 
     private final ExecutorService executor;
 
+    public static void run(final String... args) {
+        try (final NewtsConverter converter = new NewtsConverter(args)) {
+            converter.execute();
+        }
+    }
+
     /**
      * The main method.
      *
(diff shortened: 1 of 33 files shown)


features/newts-repository-converter/src/main/java/org/opennms/features/newts/converter/NewtsConverter.java
```diff
@@ -597,7 +614,24 @@ private void injectSamplesToNewts(final ResourcePath resourcePath,
...
+                            };
+                            t.start();
+                            while( t.isAlive() ) {
+                                try {
+                                    t.join();
+                                } catch( InterruptedException e ) {
+                                }
+                            }
+                        }
+                    }).start();
```
cgorantla: Is this change necessary ?

features/newts-repository-converter/src/main/java/org/opennms/features/newts/converter/NewtsConverter.java
```diff
@@ -183,7 +184,23 @@ public static void main(final String... args) {
...
+        runExitHooks(0);
+    }
+
+    private static void runExitHooks(final int exitCode) {
+        if (!EXIT_HOOKS.isEmpty()) {
+            for(Runnable hook : EXIT_HOOKS) {
+                hook.run();
+            }
+        } else {
+            if (exitCode != 0) {
```
cgorantla: may not need check for exitCode

cgorantla (commented):
Few comments otherwise LGTM!

features/newts-repository-converter/src/main/java/org/opennms/features/newts/converter/NewtsConverter.java
christianpape: removed.

features/newts-repository-converter/src/main/java/org/opennms/features/newts/converter/NewtsConverter.java
christianpape: This is because ‘com.github.stefanbirkner:system-rules’ is no longer maintained and no longer works with modern JDKs. I therefore had to revise it so that the tests would run successfully.

cgorantla: Can we do this without using nested thread loops ?

christianpape: I'll check...

christianpape: done.

cgorantla (approved):
LGTM!

PR #8428, opened 2026-04-08
NMS-19630: Deprecate and disable the legacy SNMP Config JSP UI

synqotik:
As we are replacing the legacy JSP SNMP Config UI with the new Vue one, we need to deprecate and hide the old one.

This removes any links to the old page, but doesn't remove the route. If user manages to get to the old page, the buttons are disabled and there's a note saying it's deprecated with a link to the new page. The servlet will throw an error if any write or lookup actions are performed on it.

Searching for `Configure SNMP Community Names by IP Address` or similar will link to the new page.

We will leave actually removing all the old code for another issue.

Also added new functionality to the new SNMP Config UI to be able to do an auto-lookup from a link. This handles an existing link from the Node Management page.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19630

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/search-actions.xml
+++ b/opennms-base-assembly/src/main/filtered/etc/search-actions.xml
@@ -87,7 +87,7 @@
     </action>
     <action>
         <label>Configure SNMP Community Names by IP Address</label>
-        <url>admin/snmpConfig?action=default</url>
+        <url>ui/index.html#/snmp-config</url>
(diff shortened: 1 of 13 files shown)


synqotik:
Converting to draft to fix smoke tests.

synqotik:
Smoke tests fixed!

christianpape (approved):
LGTM.

PR #8429, opened 2026-04-08
NMS-19639: Deprecate SNMP Config Rest Service v1

synqotik:
Deprecate the now-legacy SNMP Config v1 Rest API (`rest/snmpConfig`).

This has been replaced by the new `api/v2/snmp-config` Rest API as part of migrating SNMP config to the database.

We will remove the legacy API in a future release.

This adds `@Deprecated` annotations to the relevant Java classes and some comments, updates documentation and redirects links from the old to the new API. For example, Requisitions used the legacy API to update the SNMP read community string, this updates it to use the new API.

We will check more with the community before fully removing the legacy API, but we expect it to be removed in Horizon 37.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19639


christianpape (approved):
LGTM.

PR #8431, opened 2026-04-09
NMS-19616: Fix issue with email disappearing from Destination Paths

synqotik:
Fix issue where email addresses were disappearing from the `Notifications > Destination Paths > Path Outline > Initial Targets` edit screen. In addition, `destinationPaths.xml` was not being saved correctly as some items may have been removed as a user moves through the `Destination Wizard`.

### Additional info

`DestinationWizardServlet` saves email targets with `addCommand("javaEmail")`, but `getEmails()` and `getTargetNames()` in `chooseTargets.jsp` were only matching "email". This caused email addresses to vanish from the display when editing an existing destination path.                                                                                                                      

Both `getEmails()` and `getTargetNames()` have been updated to match either "email" or "javaEmail". The `contains("@")` guard has been added to `getEmails()` to prevent a target whose command happens to be javaEmail/email from being misclassified as an email entry.
                                                                              
Added additional checks to prevent the user from picking commands for email addresses.

Also in `DestinationWizardServlet` we fix the code to prevent `clearCommands()` from stripping the command off of email targets.

These issues resulted in corrupted `destinationPaths.xml` files being saved with items missing or invalid commands.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19616

### DIFF
--- a/opennms-webapp/src/main/java/org/opennms/web/admin/notification/DestinationWizardServlet.java
+++ b/opennms-webapp/src/main/java/org/opennms/web/admin/notification/DestinationWizardServlet.java
@@ -341,17 +341,25 @@ public class DestinationWizardServlet extends HttpServlet {
             for (int i = 0; i < targets.length; i++) {
                 String name = targets[i].getName();
                 // don't overwrite the email target command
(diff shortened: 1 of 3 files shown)


christianpape (approved):
LGTM.

PR #8432, opened 2026-04-09
NMS-19682: bump PostgreSQL max version to 18 and JDBC driver to 42.7.10

indigo423:
- Raise POSTGRESQL_MAX_VERSION_EXCLUSIVE from 17.0 to 19.0 to allow PostgreSQL 18 connections through the version gate in Migrator.java
- Upgrade postgresql JDBC driver from 42.5.5 to 42.7.10 for PostgreSQL 18 compatibility (requires Java 11+, already met)
- Update docs/antora.yml: compatible-postgresql to '10.x - 18.x' and postgresql-version to '18'
- Update debian/control: add postgresql-18 packages, remove iplike-pgsql dependency, add postgresql-client-18 to Suggests
- Update whatsnew.adoc: reflect supported range up to (and including) 18

### Reviewer Hint

I've removed the hard dependency on the in C implemented stored procedure iplike. The problem here, larger deployments justify the optimised stored procedure have a PostgreSQL database on dedicated machines. The iplike RPM or DEB package needs to be installed on the PostgreSQL database, not the Core system. Additionally, we cleanly fall back to the PL scripting language on IPLIKE when the stored procedure isn't installed in Postgres.

This change justifies making the next version a minor version bump for Meridian 2024.1.0, 2025.1.0, and Horizon 35.1.0. From a semver.org perspective, we now allow Meridian 2024.1.x, 2025.1.x and Horizon 35.1.x users to use PostgreSQL 16, 17, and 18 instead of 15.

Technically, we also need to bump the minimal PostgreSQL version 10.x to 14.x. I've specifically avoided introducing this change here in Meridian. The consequence would be that it could break people's deployment in a minor release update and would force them to upgrade PostgreSQL <14 to at least 14.x or newer. I would like to avoid that here. We should bump the minimal PostgreSQL version to 14.x in develop for Horizon 36 in a dedicated PR.

Additionally, we can ensure that we have iplike-pgsql{16,17,18}.{rpm,deb} available for Meridian and Horizon on Cloudsmith and {deb,yum}.opennms.org.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19682

### DIFF
--- a/core/schema/src/main/java/org/opennms/core/schema/Migrator.java
+++ b/core/schema/src/main/java/org/opennms/core/schema/Migrator.java
@@ -73,7 +73,7 @@ public class Migrator {
     private static final Logger LOG = LoggerFactory.getLogger(Migrator.class);
     private static final Pattern POSTGRESQL_VERSION_PATTERN = Pattern.compile("^(?:PostgreSQL|EnterpriseDB) (\\d+\\.\\d+)");
     private static final float POSTGRESQL_MIN_VERSION_INCLUSIVE = Float.parseFloat(System.getProperty("opennms.postgresql.minVersion", "10.0"));
-    private static final float POSTGRESQL_MAX_VERSION_EXCLUSIVE = Float.parseFloat(System.getProperty("opennms.postgresql.maxVersion", "17.0"));
+    private static final float POSTGRESQL_MAX_VERSION_EXCLUSIVE = Float.parseFloat(System.getProperty("opennms.postgresql.maxVersion", "19.0"));
 
     private static final String IPLIKE_SQL_RESOURCE = "iplike.sql";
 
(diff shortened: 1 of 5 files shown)


marshallmassengill:
This seems like a good change to me. We'll need to track the additional change for develop/H36.

christianpape (approved):
LGTM.

indigo423:
> This seems like a good change to me. We'll need to track the additional change for develop/H36.

PR #8434 is taking care of that.

dino2gnt (approved):
Great!

cgorantla (approved):
LGTM!

mershad-manesh (approved):
LGTM

PR #8437, opened 2026-04-10
NMS-16566: Update nodeLostService event to track failure reason in alarms

marshallmassengill:
I believe this will partially address: NMS-16566 (there is a lot more in that epic that I'm working on but this will partially address the concern raised) and I think it's a good quality of life improvement that's shockingly minimal.

I am adding this in develop as changing an alarm like this may have consequences I'm not thinking about.  Let me know if you come up with any.  I'm also unclear on the liquibase changes so those need to be reviewed here.

This adds %parm[eventReason]% to the nodeLostService logmsg template and enables description update-on-reduction in the alarm-data. This ensures the alarm's log message and description always reflect the current poll failure reason when a service stays down but the error changes (e.g., 404 to 500). Prior reasons should be preserved in the individual event history.

### DIFF
--- a/core/schema/src/main/liquibase/36.0.0/changelog.xml
+++ b/core/schema/src/main/liquibase/36.0.0/changelog.xml
@@ -19,4 +19,49 @@
         </createTable>
     </changeSet>
 
+    <!-- Update nodeLostService event to include failure reason in logmsg
+         and enable description update-on-reduction so alarm reflects
+         the current poll failure reason -->
+    <changeSet author="mmassengill" id="36.0.0-update-nodeLostService-reason-tracking">
+        <preConditions onFail="MARK_RAN">
+            <and>
+                <tableExists tableName="eventconf_events"/>
+                <sqlCheck expectedResult="1">
+                    SELECT COUNT(*) FROM eventconf_events
+                    WHERE uei = 'uei.opennms.org/nodes/nodeLostService'
+                    AND xml_content NOT LIKE '%parm[eventReason]%'
+                </sqlCheck>
+            </and>
+        </preConditions>
+
+        <comment>Add failure reason to nodeLostService logmsg and enable alarm description update on reduction</comment>
+
+        <!-- Update logmsg to include the reason parameter -->
+        <sql>
+            UPDATE eventconf_events
+            SET xml_content = REPLACE(
+                xml_content,
+                '%service% outage identified on interface %interface%.',
+                '%service% outage identified on interface %interface%. Reason: %parm[eventReason]%'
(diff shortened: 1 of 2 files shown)


christianpape (approved):
LGTM.

core/schema/src/main/liquibase/36.0.0/changelog.xml
```diff
@@ -0,0 +1,52 @@
...
+         and enable description update-on-reduction so alarm reflects
+         the current poll failure reason -->
+    <changeSet author="mmassengill" id="36.0.0-update-nodeLostService-reason-tracking">
+        <preConditions onFail="MARK_RAN">
+            <and>
+                <tableExists tableName="eventconf_events"/>
+                <sqlCheck expectedResult="1">
+                    SELECT COUNT(*) FROM eventconf_events
+                    WHERE uei = 'uei.opennms.org/nodes/nodeLostService'
+                    AND xml_content NOT LIKE '%parm[eventReason]%'
```
cgorantla, line 17: Looking at old https://github.com/OpenNMS/opennms/blob/99ac2aee0076ed1de9e0c5aee285692d1ad5ae48/opennms-base-assembly/src/main/filtered/etc/events/opennms.pollerd.events.xml#L439 -  it looks like `parm[eventReason]` is already part of xml_content. So this will be no-op

cgorantla (changes requested):

core/schema/src/main/liquibase/36.0.0/changelog.xml
marshallmassengill: Good catch.  Any suggestions on what to update this to look for?  Just leave it at the UEI and drop the AND since it's a replacement?

cgorantla: may be use `update-on-reduction` instead of `parm[eventReason]`

marshallmassengill: Will take a look and refactor it around that.

marshallmassengill:
I've addressed your comment @cgorantla.  I have a feeling this may need to be updated for the latest develop at this point to avoid the merge conflicts so let me know.  I can resolve that once you review.

cgorantla (commented):
Need to resolve conflicts. otherwise looks good to me.

marshallmassengill:
> Need to resolve conflicts. otherwise looks good to me.

Should be good to go now I think.  Let me know if not.

cgorantla (approved):
LGTM!

PR #8438, opened 2026-04-10
NMS-19685: DnsMonitor ignores retries on most failures

dino2gnt:
Currently, DnsMonitor only retries on `InterruptedIOException`, but should retry on any `IOException` if retries are possible.

Reported by Jesús Palacio over on the public Discourse.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19685
* https://opennms.discourse.group/t/dnsmonitor-retry-parameter-not-working-for-timeout-failures-horizon-35-0-4/4759

### DIFF
--- a/features/poller/monitors/core/src/main/java/org/opennms/netmgt/poller/monitors/DnsMonitor.java
+++ b/features/poller/monitors/core/src/main/java/org/opennms/netmgt/poller/monitors/DnsMonitor.java
@@ -217,6 +217,7 @@ final public class DnsMonitor extends AbstractServiceMonitor {
                 LOG.debug(reason1, e);
                 return PollStatus.unavailable(reason1);
             } catch (final IOException e) {
+                LOG.debug("IOException for address '{}', will retry if possible", addr);
                 // do nothing, retry if possible
             }
         }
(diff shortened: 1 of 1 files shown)


features/poller/monitors/core/src/main/java/org/opennms/netmgt/poller/monitors/DnsMonitor.java
```diff
@@ -217,9 +217,11 @@ private PollStatus pollDNS(final TimeoutTracker timeoutTracker, final int port,
                 LOG.debug(reason1, e);
                 return PollStatus.unavailable(reason1);
             } catch (final IOException e) {
-                String reason1 = "IOException while polling address: " + addr + " " + e.getMessage();
-                LOG.debug(reason1, e);
-                return PollStatus.unavailable(reason1);
+                if (! timeoutTracker.shouldRetry()) {
```
cgorantla: this condition is not required. It will always be false ( as the loop enters only when  timeoutTracker.shouldRetry is true).

just don't return. It will eventually fall through the for loop.

cgorantla (changes requested):

cgorantla (dismissed):
Looks good to me unless you want to add something in debug logs.

dino2gnt:
Yeah, logs are a good idea, so it's obvious what the monitor is doing.

cgorantla (approved):
LGTM!

PR #8448, opened 2026-04-14
NMS-19708: Do not display errors when SNMP config IPs are blank

synqotik:
NMS-19708: Do not display errors when SNMP config IPs are blank in the form.

On any SNMP Config details screen, error messages for IP address and ipMatch were displaying even if user hadn't entered anything in those fields. This removes those errors, until the user enters values and clicks `Add` button.

In addition, this PR revamps some of the validation logic:

- ignores IP range input fields on `Save` since what is saved is what is displayed in the IP range badges
- fixes various combinations of IP range inputs, disallow range/specific plus ipMatch
- disallows mixing range/specific and ipMatch when adding a range item to an existing definition
- disallows duplicate entries
- ensures correct error messages are displayed

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19708


cgorantla (approved):
LGTM!

PR #8452, opened 2026-04-15
NMS-19171: MIB Compiler UI for uploading files

Shahbaz-dataq:
### Description

Implement a UI for uploading MIB files via Rest API, see: https://opennms.atlassian.net/browse/NMS-19166 

Can be similar to the Event Config upload files functionality.

Files can be uploaded to either Compiled or Pending. Perhaps the easiest way to implement is to have 2 upload sections, one for each of those. But then all files the user selected, whether compiled or pending, can be uploaded at once, in one Rest API call.

The Rest API call will return a list of files and either success or an error message for each, similar to the Event Config upload response.

Files can have either a .txt or .mib extension and some basic checks for reasonable file names.

The files probably do not need any further client side validation, they are text files. Perhaps some reasonable file length limit like 5 MB. The server will parse them to ensure they are valid MIB files.

Can also do client side validation that filenames do not exist in either compiled or pending.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19171

### DIFF
--- a/ui/src/components/MibCompiler/UploadMibFiles.vue
+++ b/ui/src/components/MibCompiler/UploadMibFiles.vue
@@ -12,7 +12,7 @@
         </FeatherButton>
         <input
           type="file"
-          :accept="VALID_FILE_EXTENSION"
+          :accept="VALID_FILE_EXTENSION.join(',')"
           multiple
           @change="handleUpload"
           data-test="event-conf-upload-input"
@@ -25,7 +25,7 @@
           text
           data-test="clear-logs-button"
           @click="clear"
-          :disabled="isLoading|| mibFiles.length === 0 || logs.length === 0"
+          :disabled="isLoading || mibFiles.length === 0 || logs.length === 0"
         >
           Clear Logs
         </FeatherButton>
@@ -130,7 +130,9 @@ import Cancel from '@featherds/icon/navigation/Cancel'
 import Warning from '@featherds/icon/notification/Warning'
 import { FeatherTooltip } from '@featherds/tooltip'
 import { AxiosError } from 'axios'
-import { mibFilesValidator, VALID_FILE_EXTENSION } from './mibFilesValidator'
+import { format as fnsFormat } from 'date-fns'
+import { isValidMibExtension, mibFilesValidator, VALID_FILE_EXTENSION } from './mibFilesValidator'
+
 
 const isLoading = ref(false)
 const snackbar = useSnackbar()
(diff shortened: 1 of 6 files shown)


ui/src/components/MibCompiler/mibFilesValidator.ts
```diff
@@ -0,0 +1,23 @@
+import { UploadMibFileType } from '@/types/mibCompiler'
+
+export const mibFilesValidator = async (file: File): Promise<{ isValid: boolean; errors: string[] }> => {
+  const errors: string[] = []
+  if (!file.name.toLowerCase().endsWith('.txt')) {
```
synqotik: We should allow `.mib` files as well.

ui/src/components/MibCompiler/mibFilesValidator.ts
```diff
@@ -0,0 +1,23 @@
...
+    errors
+  }
+}
+
+export const isDuplicateFile = (fileName: string, existingFiles: UploadMibFileType[]): boolean => {
+  return existingFiles.some((file) => file.file.name === fileName)
+}
+
+export const MAX_FILE_SIZE = 5 * 1024 * 1024 // 5MB
+export const VALID_FILE_EXTENSION = '.txt,.mib'
```
synqotik: Changes this to: `export const VALID_FILE_EXTENSIONS = ['.txt', '.mib']`

Then make a function here like `isValidMibFileExtension`.

Then use it in `mibFilesValidator` above and also update `UploadMibFiles.vue` `handleUpload`.

ui/src/components/MibCompiler/UploadMibFiles.vue
```diff
@@ -1,10 +1,458 @@
...
+const logs = ref<LogEntry[]>([])
+
+const scrollLogsToBottom = async () => {
+  await nextTick()
+  if (logsContainerRef.value) {
+    logsContainerRef.value.scrollTop = logsContainerRef.value.scrollHeight
+  }
+}
+
+const formatTimestamp = (date: Date): string => {
```
synqotik: Instead of this, you can use `date-fns`.

See `ZenithConnectView.vue`, `formatRegistrationDate`.

You can use the same format string as there: `'yyyy-MM-dd HH:mm:ss'`

synqotik (changes requested):
Added a few comments.

ui/src/components/MibCompiler/UploadMibFiles.vue
Shahbaz-dataq: Fixed

ui/src/components/MibCompiler/mibFilesValidator.ts
Shahbaz-dataq: Fixed

ui/src/components/MibCompiler/mibFilesValidator.ts
Shahbaz-dataq: Fixed

synqotik (approved):
LGTM!

PR #8454, opened 2026-04-15
NMS-19174: MIB Compiler UI for deleting files

Shahbaz-dataq:
### Description

Files in either the Compiled Files or Pending Files tables should have a Delete icon button.

Button should pop up a Confirmation dialog, then call the Delete Rest API: https://opennms.atlassian.net/browse/NMS-19167 

After deletion, refresh the Get API call and refresh the tables.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19174

### DIFF
--- a/ui/src/components/MibCompiler/UploadMibFiles.vue
+++ b/ui/src/components/MibCompiler/UploadMibFiles.vue
@@ -130,6 +130,7 @@ import Cancel from '@featherds/icon/navigation/Cancel'
 import Warning from '@featherds/icon/notification/Warning'
 import { FeatherTooltip } from '@featherds/tooltip'
 import { AxiosError } from 'axios'
+import { format as fnsFormat } from 'date-fns'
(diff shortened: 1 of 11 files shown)


ui/src/components/MibCompiler/UploadMibFiles.vue
```diff
@@ -156,10 +154,21 @@ const scrollLogsToBottom = async () => {
   }
 }
 
+const formatTimestamp = (date: Date): string => {
```
synqotik: This should use `fnsFormat` instead.

synqotik (changes requested):
LGTM - just one thing to fix.

ui/src/components/MibCompiler/UploadMibFiles.vue
Shahbaz-dataq: Fixed

synqotik (approved):
LGTM!

PR #8456, opened 2026-04-16
NMS-19023: Prevent jackson-core v2.14.1 from being used

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-19023


synqotik (approved):
LGTM

PR #8460, opened 2026-04-16
NMS-19544: Fix donut graph top alignment

synqotik:
Fix the "donut" chart top/vertical alignment on the main page. We use the `c3` library to draw these charts.

When there are multiple categories / legend items, they may take up more than one line in the legend below the donut charts. This causes the chart with more items to be "pushed up" and now the 3 charts are misaligned.

Fix is to:

- load and render all charts, which also includes Rest API calls to get the data. We don't know the number of legend items in advance, so have to check after they are all done. Collect in a `Promise.all()`, then call `normalizeDonutTops`
- `normalizeDonutTops` can now via `#chart-content .c3-chart-arcs` to find the `y` transform that was applied and normalize them for all charts
- hook into `c3.onresized` to call `normalizeDonutTops` any time the browser window is resized

Also linted the file.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19544

### DIFF
--- a/core/web-assets/src/main/assets/js/apps/status-box/index.js
+++ b/core/web-assets/src/main/assets/js/apps/status-box/index.js
@@ -28,190 +28,227 @@ d3.document = window.document;
 
 // the color palette to map each severity to
 const colorPalette = {
-	Normal: '#336600',
-	Warning: '#ffcc00',
-	Minor: '#ff9900',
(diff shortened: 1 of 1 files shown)


christianpape (approved):
LGTM.

PR #8463, opened 2026-04-19
NMS-19617: Enable light/dark mode for Vue UI

synqotik:
Enable toggling light/dark mode for the Vue UI pages. Not yet enabled for legacy JSP etc. pages.

The top and side menu backgrounds remain dark regardless of mode, but individual items in it toggle light/dark mode, including the Search box and results, Notifications and User dropdowns.

These changes also work with PrimeVue.

In addition, fixed a few other issues:

- Structured Node Page pagination and Customize Columns drawer close button
- Double-shift to focus Search now works on both JSP and Vue SPA pages and the `keyup` handler is cleaned up
- `X` added to top of Search box results to close. It was confusing how to close it before.
- Some other styling fixes to Search box, and it now respects light/dark mode
- Resource graph Time popup dialog is now placed correctly (to the right of the side menu)

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19617

### DIFF
--- a/opennms-webapp/src/main/webapp/WEB-INF/jsp/element/nodeList.jsp
+++ b/opennms-webapp/src/main/webapp/WEB-INF/jsp/element/nodeList.jsp
@@ -110,7 +110,7 @@
         }
       %>
     </select>
-    <div class="btn-toolbar" role="toolbar">
+    <div class="btn-toolbar mt-2" role="toolbar">
       <span>Nodes</span>
       <span class="btn-group mr-2 ml-4"><a href="javascript:toggleClassDisplay('NLdbid', '', 'inline');"><i class="fas fa-database fa-lg" title="Toggle database IDs"></i></a>&nbsp;&nbsp;<a href="javascript:toggleClassDisplay('NLfs', '', 'inline');"><i class="fas fa-rectangle-list fa-lg" title="Toggle requisition names"></i></a>&nbsp;&nbsp;<a href="javascript:toggleClassDisplay('NLfid', '', 'inline');"><i class="fas fa-qrcode fa-lg" title="Toggle foreign IDs"></i></a>&nbsp;&nbsp;<a href="javascript:toggleClassDisplay('NLloc', '', 'inline');"><i class="fas fa-location-dot fa-lg" title="Toggle locations"></i></a> <c:if test="${anyFlows}">&nbsp;<a href="javascript:toggleClassDisplay('NLflows', '', 'inline');"><i class="fas fa-right-left fa-lg" title="Toggle flow data"></i></a></span></c:if>
     </div>
--- a/smoke-test/src/test/java/org/opennms/smoketest/MenuHeaderIT.java
+++ b/smoke-test/src/test/java/org/opennms/smoketest/MenuHeaderIT.java
@@ -94,7 +94,7 @@ public class MenuHeaderIT extends OpenNMSSeleniumIT {
         // Inventory Menu
         // Note, some items are below under Vue UI checks
         clickMenuItem("inventoryMenu", "Nodes");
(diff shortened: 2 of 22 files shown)


synqotik:
### Code review

Found 2 issues:

1. `NodesTable.vue` uses `@use`-namespaced SCSS variables without `var()`, producing invalid CSS. `background: variables.$surface` emits the literal string `--feather-surface` as the CSS value instead of `var(--feather-surface)`. The card background and icon border will silently fail to render. Compare the correct pattern in `BasicInformation.vue` which uses `var(variables.$surface)`.

https://github.com/OpenNMS/opennms/blob/18cc51e61974f76375567d4417cd5c37db7ed8d0/ui/src/components/Nodes/NodesTable.vue#L570-L572

https://github.com/OpenNMS/opennms/blob/18cc51e61974f76375567d4417cd5c37db7ed8d0/ui/src/components/Nodes/NodesTable.vue#L614-L616

2. `Search.vue` updates colors in `.search-results-dropdown` to Feather CSS variables, but the adjacent `.onms-search-input-wrapper` block (the search input itself) retains hardcoded light-mode values: `background-color: #f8f9fa`, `border: 1px solid #dee2e6`, `color: black`, `color: #6c757d`, placeholder `#0c0d0e`. The search input will be unreadable in dark mode while the dropdown below it adapts correctly.

https://github.com/OpenNMS/opennms/blob/18cc51e61974f76375567d4417cd5c37db7ed8d0/ui/src/components/Menu/Search.vue#L314-L350

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

synqotik:
Claude code review issues fixed.

Shahbaz-dataq (dismissed):
Good work. LGMT!

synqotik:
Smoke tests fixed.

Shahbaz-dataq (approved):
LGTM

indigo423:
@synqotik was too late :) But I would revert this change and would not allow light/dark mode in the UI for the following reason, because it is much too early. If you enable it now it will create now a huge bug surface in the many UI technologies variances that each comes with it's own styling variance.

Even if you through money at the problem solving it with AI tokens it will now be extremely more expensive and slows us down for the following reasons:

* You create now a large set of bugs you have to fix in components on stuff you have to rewrite anyway. In 90% of the cases, you just waste resources.
* Unify the theming from legacy to Vaadin, to GWT to FeatherDS to now something else becomes now much more expensive on tokens and time specifying and implementing it because you need to do the migration with two theming UI's with light/mode dark mode.

I would rather spent the AI tokens to get rid of the variances first, develop a theme that works in light mode and introduce the dark mode after that.

synqotik:
@indigo423 

I understand the concerns. However, here are counterpoints.

- The light/dark mode in this PR applies to Vue pages only. JSP pages should remain the same (for now)

- Feather already supports light/dark mode natively. So does PrimeVue, using basically the same mechanism (there's a PrimeVue/Feather test page that works fine in both modes, with both type components on the same page)

- We in fact supported dark mode in Horizon for the Vue pages a couple of years ago. We only disabled it when we redid the menu 2-3 years ago as the menu was implemented in both Vue and JSP and yes in that case toggling modes was more of an issue. But now that's no longer the case since menus on all pages are Vue

- I fixed the current top/side menu to look the same in both modes. Though, in dark mode, the dropdowns from the menu will be dark

- Since there are a lot more Vue pages now, with all of the SNMP-related pages, it's a nice quality of life to have light/dark mode on all those pages. Again, the support has already been there on any Feather components

- We will be spending order of magnitude more AI tokens on other things vs. light/dark mode on Vue pages :)

- And even having light/dark mode on legacy JSP/Vaadin pages is probably not as bad a lift as may be expected. But here the argument for not wasting time/tokens on support dark mode may be better.

PR #8465, opened 2026-04-21
NMS-19697: Getting ActiveMQ 5.19.4 to work

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-19697


features/container/sentinel/src/main/internal/features-processing.xml
```diff
@@ -0,0 +1,298 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
cgorantla: Are we using this file, It doesn't seems to be referenced anywhere in pom file for processing ?

christianpape: Yes, that was an leftover of things I've tried. Removed.

cgorantla (approved):
LGTM!

PR #8466, opened 2026-04-21
NMS-19718: Timeseries fetch alignment fix

marshallmassengill:
This should resolve NMS-19718.  StrafePing graphs fail with “QueryFailed” when OpenNMS is backed by a Prometheus-compatible time series store (Cortex/Mimir). Logs show java.lang.ArrayIndexOutOfBoundsException from FetchResults.asRowSortedTable. Other graphs mostly work because they only query 1–3 metrics; StrafePing queries 23.

The cause is TimeseriesFetchStrategy.fetch() builds the timestamps[] array from the first source's data points, then builds each column's double[] independently from its own data points.  Prometheus-based backends don't guarantee identical point counts across metrics.

This fix collects the union of all source timestamps into a sorted TreeSet, then aligns every source's values to that grid with Double.NaN for missing points, guaranteeing all column arrays match the timestamps array length. 

I've included test data from my lab in the test case.

Assisted-by: Claude Code:Opus 4.6/4.7

### DIFF
--- a/features/timeseries/src/main/java/org/opennms/netmgt/timeseries/sampleread/TimeseriesFetchStrategy.java
+++ b/features/timeseries/src/main/java/org/opennms/netmgt/timeseries/sampleread/TimeseriesFetchStrategy.java
@@ -29,7 +29,6 @@ import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -193,7 +192,7 @@ public class TimeseriesFetchStrategy implements MeasurementFetchStrategy {
             for (Entry<String, Future<Map<Source, List<DataPoint>>>> entry : measurementsByNewtsResourceId.entrySet()) {
                 Map<Source, List<DataPoint>> sampleList = toSampleList(entry);
                 for (Entry<Source, List<DataPoint>> column : sampleList.entrySet()) {
-                    Map<Long, Double> valuesByTimestamp = new LinkedHashMap<>();
+                    Map<Long, Double> valuesByTimestamp = new HashMap<>();
                     for (DataPoint dp : column.getValue()) {
                         long ts = dp.getTime().toEpochMilli();
                         allTimestamps.add(ts);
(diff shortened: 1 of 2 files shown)


features/timeseries/src/main/java/org/opennms/netmgt/timeseries/sampleread/TimeseriesFetchStrategy.java
```diff
@@ -181,28 +183,36 @@ public FetchResults fetch(long start, long end, long step, int maxrows, Long int
...
+            // returned by Prometheus-based backends, which don't guarantee
+            // identical timestamps across metrics, can still be aligned into
+            // equal-length arrays.
+            SortedSet<Long> allTimestamps = new TreeSet<>();
+            Map<String, Map<Long, Double>> valuesByLabel = new HashMap<>();
+            for (Entry<String, Future<Map<Source, List<DataPoint>>>> entry : measurementsByNewtsResourceId.entrySet()) {
+                Map<Source, List<DataPoint>> sampleList = toSampleList(entry);
+                for (Entry<Source, List<DataPoint>> column : sampleList.entrySet()) {
+                    Map<Long, Double> valuesByTimestamp = new HashMap<>();
+                    for (DataPoint dp : column.getValue()) {
```
dino2gnt, line 196: There is a lot of iteration going on here.  What are the upper bounds of these loops, the size of a single fetch request?  Do we need to do this for data sources that _aren't_ Prometheus-compatible?

I'm concerned about performance impact in a class that already labels itself "can be very CPU intensive"

marshallmassengill: Yeah, that's part of why I wanted you to look at this.  I don't think we can skip doing something like this here as the class doesn't know what backend it's using.  I'm open to ideas/suggestions though if there is somewhere else to put it...

So the loop on line 192 isn't new... it's in the old code, same thing with the second loop on 194.  What changes here is the last inner loop effectively.  

The original comment from j-white I think applies to the aggregation that has to be done if the backend isn't doing it.  This code doesn't touch raw samples though so I don't believe that applies here.

The new loop code is around line 196 and I think it should be the same iteration count but each iteration does a 'TreeSet add + HashMap put' instead of an 'array write' and I think that's going to be fairly minimal (O(log N)? - Jesus christ, I'm talking about O notation... I'm having flashbacks of "I'll never need this" crap sitting in a college lecture hall) so even at big numbers, I think the compute cost should be relatively small.

And then the other one is around line 208-211 and that is basically O(N), though in the case of strafeping it's like (O(23*N)... which I think is probably the worst case scenario. I think that's relatively computationally not bad.

TLDR: I don't think any of this is actually that computationally expensive given real world usage and aggregate data.  It shouldn't really have any impact on most single series data and strafeping is pretty much the absolute worst case for it given the number of series and even then, it would take all 23 series being fully disjointed to see real performance lag.  I am also not a developer and I've got PTSD from learning about Big O in college.

dino2gnt: Yeah, that makes sense. You're really adding only one additional iteration, it's just that the `toSampleList()` bit has been exploded into individual loops around it, making it look more complicated than it is.

cgorantla (approved):
LGTM!

PR #8443, opened 2026-04-14
NMS-19154: SNMP Config UI documentation

synqotik:
Documentation for the SNMP Config UI.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19154

PR #8470, opened 2026-04-23
NMS-19728: SNMP Config Lookup UI help

synqotik:
Instead of adding a lot of documentation to the docs about SNMP Config UI, which may become a burden to maintain, this PR instead adds some more help in the UI itself.

The various SNMP Config pages have Info icons that open dialogs with more help. The updates here should be enough to guide the user through the process.

Also updated the `MessageDialog` to have configurable `maxHeight` and `maxWidth`.

We can then close the [other PR](https://github.com/OpenNMS/opennms/pull/8443) and [ticket NMS-19154](https://opennms.atlassian.net/browse/NMS-19154).

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19728

### DIFF
--- a/ui/src/components/Common/MessageDialog.vue
+++ b/ui/src/components/Common/MessageDialog.vue
@@ -6,7 +6,7 @@
       hide-close
       @hidden="onClose"
     >
-      <div class="modal-body">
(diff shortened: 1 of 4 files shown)


christianpape (approved):
LGTM.

PR #8472, opened 2026-04-23
NMS-19712: Strafeping Update

marshallmassengill:
This addresses items 1,2, and 3 from NMS-19712.  Couple things in here that are all related:

1. The new StrafePing graph definition.

- Adaptive bar height: barH = maxMedian/100 (was hard-coded 0.0007145313).
- Loss-color bars now scale with observed max median RTT instead of assuming a 143 ms y-axis.
- New percentile stacking: old cp1..cp20/LIMIT + gray AREAs replaced with p1..p20 (UN-fallback chain so gaps propagate instead of becoming zero) and d1..d19 deltas rendered as STACK layers
- Transparent base + stacks: AREA:meL*#00000000 (was colorless AREA:meL*), enabling backshift v1.3.4's 8-digit-hex transparency.
- GPRINTs rewritten: explicit median avg/max/min/now and ploss avg/max/now lines replacing the single packet-loss line.

2. Removing the filters for StrafePing, I'm fairly certain these were there as examples at some point and they aren't needed and just clutter up the poller config and make using StrafePing harder.  It's not auto added to nodes via discovery anyway.
3. I've added the option for "--slope-mode" to work (as inverted from what rrdtool does with it) for rendering steps.  I've left it out of the graph definition but it was vaguely hinted at by the customer so I might need it for them when this gets in.

Some before and after images of the new graphs (with the backshift fixes in now):

Before:
image

After:
image

Before:
image

After:
image

Bonus screenshot of steps in action:
image

Assisted-by: Claude Code:Opus 4.6 & 4.7

### DIFF
--- a/core/web-assets/src/main/assets/js/apps/onms-graph/index.js
+++ b/core/web-assets/src/main/assets/js/apps/onms-graph/index.js
@@ -130,9 +130,12 @@ const drawBackshiftGraph = (el, def, dim) => {
     });
     const graphModel = rrdGraphConverter.model;
 
-    // OpenNMS-specific marker (not real rrdtool syntax) to opt into
-    // backshift's step-line rendering for this graph.
-    const stepMode = /(^|\s)--step-mode(\s|$)/.test(graphDef.command || '');
+    // Graph commands can include --slope-mode to opt into backshift's
+    // step-line rendering. The flag name is borrowed from rrdtool so the
+    // command stays rrdtool-compatible on the PNG render path, but note
+    // rrdtool's own interpretation of --slope-mode is the opposite (it
+    // enables sloped lines in PNG output).
+    const stepMode = /(^|\s)--slope-mode(\s|$)/.test(graphDef.command || '');
 
     // Build the data-source
     const ds = new Backshift.DataSource.OpenNMS({
--- a/core/web-assets/src/main/assets/js/apps/onms-graph/index.js
+++ b/core/web-assets/src/main/assets/js/apps/onms-graph/index.js
@@ -130,6 +130,10 @@ const drawBackshiftGraph = (el, def, dim) => {
     });
     const graphModel = rrdGraphConverter.model;
 
+    // OpenNMS-specific marker (not real rrdtool syntax) to opt into
+    // backshift's step-line rendering for this graph.
+    const stepMode = /(^|\s)--step-mode(\s|$)/.test(graphDef.command || '');
+
     // Build the data-source
     const ds = new Backshift.DataSource.OpenNMS({
       url: window.onmsGraphContainers.baseHref + 'rest/measurements',
@@ -147,7 +151,8 @@ const drawBackshiftGraph = (el, def, dim) => {
       model: graphModel,
       printStatements: graphModel.printStatements,
       title: graphModel.title,
-      verticalLabel: graphModel.verticalLabel
+      verticalLabel: graphModel.verticalLabel,
+      step: stepMode
     });
     graph.render();
   }).fail((jqXHR, textStatus) => {
--- a/opennms-base-assembly/src/main/filtered/etc/response-graph.properties
+++ b/opennms-base-assembly/src/main/filtered/etc/response-graph.properties
@@ -322,7 +322,7 @@ report.strafeping.command=--title="StrafePing Response Time" \
  --alt-autoscale-max \
  --alt-y-grid \
  --lower-limit 0 \
- --vertical-label Seconds \
+ --vertical-label "Seconds" \
  DEF:ping1Micro={rrd1}:ping1:AVERAGE \
  CDEF:ping1=ping1Micro,1000000,/ \
  DEF:ping2Micro={rrd1}:ping2:AVERAGE \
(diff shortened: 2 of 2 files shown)


core/web-assets/package-lock.json
```diff
@@ -3420,6 +3420,27 @@
...
+                "android"
+            ],
+            "engines": {
+                "node": ">= 10.0.0"
+            },
+            "funding": {
+                "type": "opencollective",
+                "url": "https://opencollective.com/parcel"
+            }
+        },
```
dino2gnt: AFAIK this file is generated by the build; there's n reason to modify or commit it.

core/web-assets/src/main/assets/js/apps/onms-graph/index.js
```diff
@@ -130,6 +130,10 @@ const drawBackshiftGraph = (el, def, dim) => {
     });
     const graphModel = rrdGraphConverter.model;
 
+    // OpenNMS-specific marker (not real rrdtool syntax) to opt into
```
dino2gnt: If a user sets `org.opennms.web.graphs.engine=png` does this blow up the graph?

dino2gnt (changes requested):

core/web-assets/src/main/assets/js/apps/onms-graph/index.js
marshallmassengill: I'll test.  I'm not sure.

core/web-assets/package-lock.json
marshallmassengill: Yeah, you're right.  I think this picked up a bunch of stuff that wasn't needed.  I'll remove this one.

core/web-assets/src/main/assets/js/apps/onms-graph/index.js
marshallmassengill: It does indeed fail to render the graph. (the one on top, the one below is new graph definition with png rendering).

image

I can look at some other ways that might work... not sure what yet but I'll give it a poke this morning.

My other thoughts are that I didn't include this in the updated definition so anyone using it would have to A) know about and B) intentionally set it and then C) switch to png rendering or have it enabled thinking it would work... not sure how much of a concern that is but I can remove it.

marshallmassengill: Two options... one that is stupid and one less so...

The semi-stupid option is using "--slope-mode" and it basically becomes an inversion of what it means for RRDtool... I'm not sure I like it but it can work.  It doesn't seem to do anything with the png renderer as is.

Another option is adding something to: opennms-web-api/src/main/java/org/opennms/web/svclayer/support/DefaultRrdGraphService.java to strip out the --step-mode option before it gets passed elsewhere.  That could work and keeps the option name sane.

dino2gnt: I was looking for that option, couldn't remember the name.  I'd be much more okay with co-opting `--slope-mode` for this than leaving a silent rendering breakage in the graph.

marshallmassengill: That works for me.  Update incoming.

opennms-base-assembly/src/main/filtered/etc/poller-configuration.xml
```diff
@@ -336,8 +336,6 @@
    -->
    <package name="strafer">
       <filter>IPADDR != '0.0.0.0'</filter>
```
indigo423, line 338: I wouldn't turn on the fire hose by removing the include-range filter here in this PR. It could explode in someone's face when they write now thousands and thousands additional time series data and don't know why.

If you really want to get rid of that include range, we should make sure we remove StrafePing from the default detectors and let the user control that way on which IP interfaces StrafePing is enabled or not.

opennms-base-assembly/src/main/filtered/etc/response-graph.properties
```diff
@@ -363,94 +363,121 @@ report.strafeping.command=--title="StrafePing Response Time" \
...
+ CDEF:minMedian=minMedianMicro,1000000,/ \
+ DEF:maxMedianMicro={rrd1}:median:MAX \
+ CDEF:maxMedian=maxMedianMicro,1000000,/ \
+ CDEF:barH=maxMedian,100,/ \
+ DEF:loss={rrd1}:loss:AVERAGE \
+ DEF:maxLoss={rrd1}:loss:MAX \
  CDEF:ploss=loss,20,/,100,* \
  CDEF:maxPloss=maxLoss,20,/,100,* \
- GPRINT:median:AVERAGE:"Median RTT (%.1lf %ss avg)" \
+ CDEF:p20=ping20 \
```
indigo423, line 377: Fallback to median if ping20 is unknown?
CDEF:p20=ping20,UN,median,ping20,IF

indigo423 (dismissed):
Nothing to me that blocks it from my point of view. The include filter is a [-0](https://www.apache.org/foundation/voting)

opennms-base-assembly/src/main/filtered/etc/poller-configuration.xml
dino2gnt: StrafePing isn't included in the default foreign source.

marshallmassengill: As far as I know Strafeping is not in the default detectors, which is why I think this is something we should remove.  What these filters do right now is cause a problem when you add Strafeping as a service to a node and it doesn't immediately work because the IP is out of the range.  You have to add it though.

opennms-base-assembly/src/main/filtered/etc/response-graph.properties
marshallmassengill: Nope.  StrafePing sorts results with nulls at the front... so with any packet loss, the nulls fill ping1, ping2, … from the bottom up, and ping20 always holds the slowest response.  If ping20 is UN then that means every packet is lost. In that case there's no median so we don't want the fallback chain to be here.

opennms-base-assembly/src/main/filtered/etc/poller-configuration.xml
marshallmassengill: Just to confirm, these are the default detectors today (no strafeping):
  - DNS
  - FTP
  - HTTP
  - HTTPS
  - ICMP
  - IMAP
  - LDAP
  - NRPE
  - POP3
  - SMTP
  - SNMP
  - SSH
  - WS-Man
  - MS-RDP (TCP 3389)
  - OpenNMS-JVM
  - JMX-Minion
  - JMX-Cassandra
  - JMX-Cassandra-Newts

I do think we should kill some of these but that's another PR.

opennms-base-assembly/src/main/filtered/etc/poller-configuration.xml
```diff
@@ -336,8 +336,6 @@
...
-      <include-range begin="10.1.1.1" end="10.1.1.10" />
```
cgorantla, line 339: Are we good to force config change in minor release ?

marshallmassengill: You tell me.  I think Dino and I are but I can remove that here and push it to develop.

marshallmassengill: Yeah, that's what I'm going to do.  Will submit this separately.

cgorantla (approved):
LGTM!

dino2gnt (approved):
LGTM

PR #8477, opened 2026-04-23
NMS-19712: remove default filters for strafeping

marshallmassengill:
I'm lumping this in with NMS-19712 but applying to develop based on feedback.  Strafeping is not part of the detectors so there isn't a risk of it being auto discovered and added to nodes.  This should make it easier to use.


cgorantla (approved):
LGTM!

PR #8478, opened 2026-04-23
NMS-19523: Use common dialogs in ui

synqotik:
We have created some common dialogs that we can use including `ConfirmationDialog` and `MessageDialog`.

This uses those instead of custom dialogs that did the same thing. There may be a few more cases, and in many cases we could get rid of the custom dialog file entirely and put the logic in the parent container, but this is a good first step.

This will help when moving to another UI component framework.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19523


Shahbaz-dataq (approved):
LGTM!

PR #8481, opened 2026-04-24
NMS-19731: Fix SNMP Config set default overrides

synqotik:
There was an issue with saving SNMP Config default overrides. If user cleared out values in order to set them back to the system default (i.e. in the UI, but also in Rest API), the null values would not actually replace the values in the backing store, but instead were treated as "not changed".

This PR exposes the `AbstractCmJaxbConfigDao` `updateConfig` method which takes a `boolean isReplace` to completely replace the config in the backing store. The Rest API now uses this when saving default overrides, ensuring the default override value is reset to `null` and thus the system falls back to the hard-coded default value.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19731


christianpape (approved):
LGTM.

PR #8495, opened 2026-04-30
NMS-19751: Support for Token Based Authentication with XML/JSON Collector (and HTTP collector)

marshallmassengill:
This adds support for token-based authentication against external HTTP APIs during data collection. Operators describe an API login flow once in etc/token-auth-configuration.xml and reference the resulting token from any XML/JSON or HTTP collector header via the ${token:<name>} metadata placeholder. The runtime performs the login call, caches the token, refreshes on TTL, invalidates on 401/403, and supports reload via the standard daemon-reload event.

The motivating use case is vendor APIs like Cisco Catalyst Center, F5 BIG-IP, and VMware vSphere, where the collectors themselves can't perform a login flow.

What's in here:
- JAXB model + XSD for token-auth-configuration.xml with three token-extraction modes (jsonpath, response header, body-as-token)
- TokenAcquirer, TokenCache, TokenProviderImpl runtime, plus a TokenAuthCollectorAdaptor that resolves ${token:<name>} placeholders on the controller before the collector dispatch
- Two adaptor entry points so ${token:...} works wherever auth headers live:
- beforeCollect substitutes inside the basic collection-parameter map (e.g. collectd-configuration.xml <parameter> values)
- beforeRuntimeInterpolation substitutes inside the JAXB-deserialized runtime-attribute tree (e.g. xml-datacollection-config.xml <header> values, http-datacollection-config.xml <header> values). Mate's standard pass then resolves the remaining ${node:...}, ${scv:...}, ${env:...} etc. without touching core/mate.
- Per-node placeholder support inside <token-auth> blocks: ${node:label}, ${scv:alias:key}, ${env:NAME} all resolve before the auth call. Cache identity widens to (auth-name, SHA-256 fingerprint of resolved fields) so one auth definition can serve many regional endpoints with one cache entry per distinct resolution.
- Early-refresh window of max(600s, 5% of <ttl-seconds>) on the cache so the controller never hands back a token within ten minutes of upstream expiry.
- Passive invalidation on 401/403: the failing request isn't retried in-cycle; the adaptor invalidates the matching cache entry, and the next scheduled collection re-acquires.
- Reload listener: opennms:send-event -p daemonName=TokenAuth uei.opennms.org/internal/reloadDaemonConfig flushes the cache and re-reads the file; reload failures publish reloadDaemonConfigFailed and leave the previous config in effect.
- Custom <headers> block on <url> in http-datacollection-config.xml so Authorization: Bearer ${token:foo} works in the HTTP collector too.
- Reference docs at Operation → Deep Dive → Administration → Configuration → Collector Token Authentication (docs/.../token-authentication.adoc).

Validated end-to-end against Hashicorp Vault (auth/client_token JSON), PocketBase (superuser token JSON), and WireMock — including a Cisco Catalyst Center-style stub that exchanges HTTP Basic for a 
Token JSON and gates data endpoints behind X-Auth-Token. Each test exercises SCV-resolved credentials plus per-node metadata in the auth definition (URL path, basic-auth, request body, and request 
headers).

I'd love feedback on this. We were asked by a potential customer the other day about our capability to scrape some specific endpoints that use this style of authentication and I think this fills a 
pretty big gap for us in the XML/JSON Collector today.

Assisted-by: Claude Code:Opus 4.7


marshallmassengill:
So this is passing except for the WmiPeerFactoryTest in CircleCI as far as I can tell.  There were indeed some integration test failures that I needed to fix.  They should be now.

dino2gnt:
> So this is passing except for the WmiPeerFactoryTest in CircleCI as far as I can tell.

The `WmiPeerFactoryTest` is failing to instantiate the `EntityScopeProvider`, which seems like it could be related to your metadata changes.

```
Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.context.support.ClassPathXmlApplicationContext]: Constructor threw exception; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'entityScopeProvider' defined in class path resource [test-dao-context.xml]: Instantiation of bean failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.*******.core.mate.api.EntityScopeProvider]: Specified class is an interface
```

marshallmassengill:
I'll give it a look again.

marshallmassengill:
Integration tests all seem to be passing now.  There were definitely a couple things in there.  I've made a note to myself to run those locally first now.

dino2gnt:
Smooth-brain thought:  Couldn't you accomplish this same thing using a shell script in cron that does a `curl` and feeds the token response to `scvcli`?

marshallmassengill:
> Smooth-brain thought: Couldn't you accomplish this same thing using a shell script in cron that does a `curl` and feeds the token response to `scvcli`?

I think that can/will work... but there are some misses to me...  It could definitely cover a single long lived token use case... I can't argue that really but I don't think it covers the shorter lived ones as nicely.

That being said, I think the biggest issue is that any 401/403 recoveries have to happen on the cron schedule and can't happen in-line with a response (granted, I guess you could use the failed collection in OpenNMS and have it do that with scripting but then you've got 3 things to mangle).  

There is also something to be said that this works with metadata (and SCV) whereas cron wouldn't so you could do this across a fleet of systems instead of having to probably create a separate script for each one and having to manage other secure vault stuff.

I also think this keeps everything in OpenNMS and that wayt tokens live in memory (RAM) as a result so I think that fits with the idea of tokens being short lived and not meant to be permanent (semi-permanent?) and stored in a vault.

marshallmassengill:
@cgorantla I added the minion 401/403 feature.  Wasn't too bad to add but it does alter the HTTPCollector a bit given that those need to be retries but I could not come up with a reason someone would be using that collector for testing that and there is an alternative through PSM.  Tests all seem to be passing.  Let me know about feedback.  I can adjust accordingly.

core/mate/api/src/main/java/org/opennms/core/mate/api/Interpolator.java
```diff
@@ -100,15 +100,61 @@ public static Object interpolate(final Object value, final Scope scope) {
...
+     * Returns the {@link Scope} of the interpolation call that is
+     * currently in flight on this thread, or empty if no interpolation
+     * is in progress (which is the typical case from any code that is
+     * not itself a {@link Scope} implementation reacting to a lookup).
+     */
+    public static Optional<Scope> currentCallingScope() {
+        final java.util.Deque<Scope> stack = CALLING_SCOPE_STACK.get();
+        return stack.isEmpty() ? Optional.empty() : Optional.of(stack.peek());
+    }
+
```
cgorantla: Looks like this uses Thread Local storage to resolve recursive metadata present in auth config itself.
Clever implementation but not sure if I understood this intuitively.
To really understand, I tried to check tests but there are no tests to really test this behavior.

core/mate/api/src/main/java/org/opennms/core/mate/api/Interpolator.java
```diff
@@ -100,15 +100,61 @@ public static Object interpolate(final Object value, final Scope scope) {
...
-        final ImmutableList.Builder<ResultPart> parts = ImmutableList.builder();
-        final String output = interpolateRecursive(raw, parts, scope, 1);
+        // Callers may pass a null scope when no scope context is available
+        // (e.g. WmiPeerFactory before SecureCredentialsVault is wired). The
+        // pre-dynamic-auth contract was to handle this fall-through silently
+        // for inputs that contain no placeholders. Substitute EmptyScope.EMPTY
+        // so inputs that DO contain placeholders also fall through cleanly
+        // (resolving to empty / default) instead of NPEing inside
+        // interpolateSingle when it calls scope.get(...). Skip the
+        // calling-scope-stack push since EmptyScope contributes nothing
```
cgorantla: not sure if we ned this change if it is only test that needs update.

Also it would better to avoid too much of comments in the code that doesn't add any value.

opennms-config/src/main/java/org/opennms/netmgt/config/auth/AuthConfigReloadListener.java
```diff
@@ -0,0 +1,173 @@
...
+ * </pre>
+ *
+ * <p>Self-registration: this bean lives in {@code daoContext}, which is
+ * built before {@code eventDaemonContext}. The OpenNMS
+ * {@link EventIpcManager} therefore is not yet wired when this bean is
+ * constructed. Spring publishes a {@link ContextRefreshedEvent} on every
+ * context refresh, including child contexts, and those events propagate
+ * up to listeners in the parent. We use that signal to self-register
+ * with the {@link EventIpcManagerFactory} once it is initialised; the
+ * registration is one-shot, guarded by a flag.</p>
```
cgorantla: Not sure if these assumptions works in our case. None of our code depends upon ApplicationListener<ContextRefreshedEvent> as there are multiple application contexts in OpenNMS

opennms-config/src/main/java/org/opennms/netmgt/config/auth/TokenAcquirerScopeBinder.java
```diff
@@ -0,0 +1,72 @@
...
+ * per-node placeholder resolution is enabled, or (b) standalone in test
+ * contexts and minion-light bundles that pull in opennms-config but not
+ * core/mate/model, in which case the acquirer runs in degraded mode (SCV
+ * and env scopes only).
+ *
+ * <p>Hard-referencing {@code entityScopeProvider} via a Spring
+ * {@code <property>} or {@code MethodInvokingFactoryBean} fails the
+ * Spring context build whenever one half of the pair is missing -- that
+ * was the failure mode before this binder existed.</p>
+ */
```
cgorantla: not sure if this works, this seems to be a hacky way to get around some limitation but not sure if I understood this.

features/collection/client-rpc/src/main/java/org/opennms/netmgt/collection/client/rpc/CollectorRequestBuilderImpl.java
```diff
@@ -168,8 +183,63 @@ public CompletableFuture<CollectionSet> execute() {
...
+     * the {@link CollectionSet} on. On 401/403 from a remotely-served
+     * collection, invalidates the cached dynamic-auth token(s) that
+     * carried the failed request and re-issues the RPC once. A second
+     * auth failure in a row surfaces as a {@link
+     * CollectionAuthFailureException}, matching what an unrecoverable
+     * core-local failure would produce.
+     */
+    private CompletableFuture<CollectionSet> handleResponseWithRetry(final CollectorResponseDTO response) {
+        if (response.getAuthFailureStatusCode() == null) {
+            return CompletableFuture.completedFuture(response.getCollectionSet());
```
cgorantla: This is very clever but I'm not sure if this works reliably and one off handling of these things in Core code doesn't seems to be good idea for me

cgorantla (commented):
Reviewed few modules. Need to actually validate this by running this code.

marshallmassengill:
Converted to draft for now.  Working on a revised plan based on the feedback.

core/mate/api/src/main/java/org/opennms/core/mate/api/Interpolator.java
```diff
@@ -100,15 +100,61 @@ public static Object interpolate(final Object value, final Scope scope) {
...
+    /**
+     * Per-thread stack of the {@link Scope} currently driving an
+     * interpolation pass. Updated by {@link #interpolate(String, Scope)}
+     * and queried by {@link AuthScope} so that an auth definition can
+     * resolve {@code ${node:...}}, {@code ${requisition:...}}, etc.
+     * against the same context the caller used. The stack handles
+     * legitimate nesting (e.g. a recursive interpolator pass on a
+     * fall-through default).
+     */
+    private static final ThreadLocal<java.util.Deque<Scope>> CALLING_SCOPE_STACK =
```
cgorantla: Check if it is possible to avoid changes in metadata code and restrict changes to only when Auth is involved.

features/collection/client-rpc/src/main/java/org/opennms/netmgt/collection/client/rpc/CollectorRequestBuilderImpl.java
```diff
@@ -168,8 +183,63 @@ public CompletableFuture<CollectionSet> execute() {
...
+    /**
+     * Inspects the response for an auth-failure signal. On success, hands
+     * the {@link CollectionSet} on. On 401/403 from a remotely-served
+     * collection, invalidates the cached dynamic-auth token(s) that
+     * carried the failed request and re-issues the RPC once. A second
+     * auth failure in a row surfaces as a {@link
+     * CollectionAuthFailureException}, matching what an unrecoverable
+     * core-local failure would produce.
+     */
+    private CompletableFuture<CollectionSet> handleResponseWithRetry(final CollectorResponseDTO response) {
```
cgorantla: Consider similar thing like `ServiceMonitorAdaptor` for collectors to implement this custom code.

protocols/xml/src/main/java/org/opennms/protocols/xml/collector/AbstractXmlCollectionHandler.java
```diff
@@ -448,28 +461,151 @@ protected static String parseString(final String reference, final String unforma
...
+    protected synchronized TokenProvider getTokenProvider() {
+        if (m_tokenProvider == null) {
+            try {
+                m_tokenProvider = BeanUtils.getBean("daoContext", "authTokenProvider", TokenProvider.class);
+            } catch (Throwable t) {
+                LOG.debug("Dynamic-auth TokenProvider unavailable in this context; cannot retry on auth failure", t);
+            }
+        }
+        return m_tokenProvider;
+    }
```
cgorantla: BeanUtils.getBean will not work on Minion. 

All token based code should be resolved before we invoke Collection RPC and avoid using token related code in common code.

cgorantla (commented):
I have added few more comments on what I think should be considered.

marshallmassengill:
This PR has been rewritten to address issues raised:

  - "Don't update core modules like metadata and RPC collections for the small benefit" — no changes to core/mate/api/Interpolator, EntityScopeProviderImpl, CollectorClientRpcModule, or CollectorResponseDTO.
  - "Consider similar thing like ServiceMonitorAdaptor for collectors" — new CollectorAdaptor interface in features/collection/api modeled on ServiceMonitorAdaptor. Token-auth logic lives in a single CollectorTokenAuthAdaptor class that runs on the OpenNMS controller (never on the Minion).
  - "BeanUtils.getBean will not work on Minion" — token resolution happens on the controller before RPC dispatch. The Minion sees already-substituted parameter values; no token-aware code on the Minion side.
  - "Interpolator ThreadLocal recursive lookup — clever but not intuitive" — removed entirely. The auth runtime no longer participates in the metadata DSL scope chain.
  - "ApplicationListener doesn't fit OpenNMS's multi-context model" — replaced with standard Spring DI of EventIpcManager plus an init-method, same pattern as Collectd/Statsd/etc.
  - "TokenAcquirerScopeBinder is a hacky workaround" — class deleted; the binding cycle it papered over is gone after AuthScope removal.

  Other changes:
  - Feature renamed to "Collector Token Authentication" to match the docs page title. Placeholder syntax is now ${token:<name>}. Config file is etc/collector-token-auth-configuration.xml. Daemon-reload name is CollectorTokenAuth.
  - PR base changed from develop to release-36.x
  - Passive cache invalidation on the adaptor: when the controller sees a FAILED CollectionSet whose request used a token, it invalidates that cache entry. Next cycle re-acquires naturally. No retry orchestration.

I'm going to kick off a copilot review @cgorantla but will wait to address anything until you've looked this over.

I also still need to run integration tests locally.

cgorantla (commented):
This looks much better. 

One thing that can be improved here:

Drop Collector prefix from token-auth configuration. Most of this code can be re-used for Monitors after adding a separate ServiceMonitorAdaptor.

cgorantla (commented):
Few more changes I'm thinking of 

1. Metadata resolution can be simplified further:  Auth parameters should be named, use those in collector configuration headers etc. ( which you are already doing). Then CollectorAdaptor in preRpc parses the config, gets auth parameter id, then do metadata interpolation directly on these parameters, resolve token and then update the parameters here. If auth param is also in the format of metadata, just parse that specific one but don't update any common metadata code.
3. Renaming from Collector token auth to just token auth. 
3. Can we expire `ttl` earlier ( may be at before 5 percent of time left or 10 mins of time, whichever is higher)

opennms-services/src/main/java/org/opennms/netmgt/collectd/Collectd.java
```diff
@@ -167,6 +167,16 @@ public static CollectionInstrumentation instrumentation() {
     @Autowired
     private volatile LocationAwareCollectorClient m_locationAwareCollectorClient;
 
+    /**
+     * Optional list of {@link org.opennms.netmgt.collection.api.CollectorAdaptor}
+     * beans visible in this context. Each adaptor is registered on every
+     * collector request so it can transform parameters before dispatch
+     * and inspect the result on the way back. {@code required = false}
+     * keeps this optional for deployments / tests that wire no adaptors.
+     */
```
cgorantla: omit all of this comment, not needed

opennms-services/src/main/java/org/opennms/netmgt/collectd/Collectd.java
```diff
@@ -167,6 +167,16 @@ public static CollectionInstrumentation instrumentation() {
...
+    @Autowired(required = false)
+    private volatile List<org.opennms.netmgt.collection.api.CollectorAdaptor> m_collectorAdaptors = Collections.emptyList();
```
cgorantla: remove package `org.opennms.netmgt.collection.api.`

opennms-config/src/main/resources/META-INF/opennms/applicationContext-token-auth.xml
```diff
@@ -0,0 +1,77 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns="http://www.springframework.org/schema/beans"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns:onmsgi="http://xmlns.opennms.org/xsd/spring/onms-osgi"
+       xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-4.2.xsd
+        http://xmlns.opennms.org/xsd/spring/onms-osgi http://xmlns.opennms.org/xsd/spring/onms-osgi.xsd">
+
+    <!--
```
cgorantla: we can remove all comments from this file. Most of them are not needed and repititive

opennms-config/src/main/java/org/opennms/netmgt/config/TokenAuthConfigFactory.java
```diff
@@ -0,0 +1,241 @@
...
+ */
+public final class TokenAuthConfigFactory {
+
+    private static TokenAuthConfigFactory m_singleton;
+
+    private final TokenAuthConfiguration m_config;
+
+    public TokenAuthConfigFactory(final InputStream stream) {
+        this(JaxbUtils.unmarshal(TokenAuthConfiguration.class,
+                new InputStreamReader(stream, java.nio.charset.StandardCharsets.UTF_8)));
```
cgorantla:
```suggestion
                new InputStreamReader(stream, StandardCharsets.UTF_8)));
```

opennms-config/src/main/java/org/opennms/netmgt/config/tokenauth/TokenProviderImpl.java
```diff
@@ -0,0 +1,134 @@
...
+        } catch (final IOException e) {
+            throw new RuntimeException("failed to load token-auth-configuration.xml", e);
+        } catch (final IllegalStateException e) {
+            // Race or test scenario where init() has not yet completed.
+            // Tokens silently fail to resolve until next call; surface
+            // at debug so an operator chasing a "${token:foo} resolved
+            // to empty" report can see this happened.
+            LOG.debug("TokenAuthConfigFactory not yet initialised; ${token:...} will resolve as empty until init() completes", e);
+            return null;
+        }
```
cgorantla: Don't think we need to lazily load TokenAuthConfigFactory as we ship `token-auth-configuration.xml` in etc.  There is no chance that it will be missed. 

We can wire `TokenAuthConfigFactory` in `applicationContext-token-auth.xml` the same way you did in `applicationContext-configTester.xml`.

That will be cleaner

opennms-config/src/main/java/org/opennms/netmgt/config/TokenAuthConfigFactory.java
```diff
@@ -0,0 +1,241 @@
...
+    /**
+     * Test-friendly constructor that takes an already-unmarshalled config.
+     */
+    public TokenAuthConfigFactory(final TokenAuthConfiguration config) {
+        if (config == null) {
+            throw new IllegalArgumentException("config must not be null");
+        }
+        validate(config);
+        m_config = config;
+    }
```
cgorantla: If it's only used for testing, move to the end and use @VisibleOnlyForTesting annotation

opennms-config/src/main/java/org/opennms/netmgt/config/TokenAuthConfigFactory.java
```diff
@@ -0,0 +1,241 @@
...
+     * Test-only helper to drop the active singleton so the next call
+     * to {@link #init()} or {@link #reload()} reads from disk fresh.
+     * Public visibility (in contrast to {@link #setInstance}) because
+     * resetting to null does not bypass validation -- the next load
+     * still goes through {@link #init()} and the constructor's
+     * checks. Useful from test classes outside this package.
+     */
+    public static synchronized void resetForTesting() {
+        m_singleton = null;
+    }
```
cgorantla: can avoid comments, if this is only needed for testing, move to the end and @VisibleOnlyForTesting annotation

opennms-config/src/main/java/org/opennms/netmgt/config/tokenauth/TokenAcquirer.java
```diff
@@ -0,0 +1,454 @@
...
+        // Make the EmptyKeyRelaxedTrustSSLContext algorithm available
+        // to JSSE so configureClient() can request relaxed-trust
+        // SSLContext when an auth definition opts in via
+        // disable-ssl-verification="true". HttpClientWrapper.useRelaxedSSL
+        // looks up the algorithm by name and fails with
+        // NoSuchAlgorithmException if the provider was never added.
+        // Other components (HttpCollector, PageSequenceMonitor, ...)
+        // also register this provider; java.security.Security tolerates
+        // duplicate adds and returns -1 for the second.
+        java.security.Security.addProvider(new EmptyKeyRelaxedTrustProvider());
```
cgorantla:
```suggestion
        Security.addProvider(new EmptyKeyRelaxedTrustProvider());
```
import `java.security` directly

opennms-config/src/main/java/org/opennms/netmgt/config/tokenauth/TokenAcquirer.java
```diff
@@ -0,0 +1,454 @@
...
+    }
+
+    private static final ObjectMapper JSON = new ObjectMapper();
+
+    private static final int DEFAULT_CONNECT_TIMEOUT_MS = 10_000;
+    private static final int DEFAULT_SOCKET_TIMEOUT_MS = 30_000;
+
+    private final int connectTimeoutMs;
+    private final int socketTimeoutMs;
+    private final java.time.Clock clock;
```
cgorantla:
```suggestion
    private final Clock clock;
```

opennms-config/src/main/java/org/opennms/netmgt/config/tokenauth/TokenAuthCollectorAdaptor.java
```diff
@@ -0,0 +1,401 @@
...
+            field.setAccessible(true);
+        } catch (final ClassNotFoundException | NoSuchFieldException e) {
+            LoggerFactory.getLogger(TokenAuthCollectorAdaptor.class)
+                    .warn("Mate Interpolator.ToBeInterpolated not reachable; "
+                            + "token-auth substitution will only see top-level Map values "
+                            + "(not their JAXB-wrapped innards).", e);
+        }
+        TO_BE_INTERPOLATED_CLASS = klass;
+        TO_BE_INTERPOLATED_VALUE_FIELD = field;
+    }
```
cgorantla: avoid changes in Metadata common code doesn't mean, we avoid using it completely.

Avoid reflection, you can add these two methods in 

public static boolean isPleaseInterpolate(Object value) {
      return value instanceof ToBeInterpolated;
  }

  public static Object unwrap(Object value) {
      return value instanceof ToBeInterpolated
              ? ((ToBeInterpolated) value).value
              : value;
  }

features/collection/core/src/main/java/org/opennms/netmgt/collection/core/CollectionSpecification.java
```diff
@@ -81,8 +81,9 @@ public class CollectionSpecification {
     private final ReadablePollOutagesDao m_pollOutagesDao;
     private final String collectorImplClassName;
     private final ScopeProvider scopeProvider;
+    private final java.util.List<org.opennms.netmgt.collection.api.CollectorAdaptor> m_collectorAdaptors;
```
cgorantla:
```suggestion
    private final List<CollectorAdaptor> m_collectorAdaptors;
```

don't expand package naming

cgorantla (commented):
Suggested few comments, I still need to go through whole code one more time

It is difficult to go through the code as there is too much commenting for every method.

Only add comments that are strictly necessary.

Also lot of places, package naming is expanded like `java.util.List<org.opennms.netmgt.collection.api.CollectorAdaptor>` which should be avoided.

opennms-config/src/main/java/org/opennms/netmgt/config/tokenauth/TokenAuthCollectorAdaptor.java
```diff
@@ -0,0 +1,278 @@
...
+        final Object raw = parameters.get(AUTH_NAMES_USED_PARAM);
+        if (!(raw instanceof String) || ((String) raw).isEmpty()) {
+            return result;
+        }
+        for (final String name : ((String) raw).split(",")) {
+            final String trimmed = name.trim();
+            if (trimmed.isEmpty()) {
+                continue;
+            }
+            tokenProvider.invalidate(trimmed);
```
cgorantla, line 137: Do we need to do this invalidation of tokens for every failure case or should we limit this to auth failures ?

features/collection/client-rpc/src/main/java/org/opennms/netmgt/collection/client/rpc/CollectorRequestBuilderImpl.java
```diff
@@ -121,7 +137,16 @@ public CompletableFuture<CollectionSet> execute() {
...
+        // Adaptors run before Mate's Interpolator so they can resolve
+        // their own ${prefix:...} placeholders before unknown prefixes
+        // get stripped.
+        Map<String, Object> preInterpolated = attributes;
+        for (final CollectorAdaptor adaptor : adaptors) {
+            final Map<String, Object> next = adaptor.beforeCollect(agent, preInterpolated);
+            preInterpolated = next != null ? next : preInterpolated;
+        }
+
+        final Map<String, Object> interpolatedAttributes = Interpolator.interpolateObjects(preInterpolated, scope);
```
cgorantla: may be we can remove `beforeCollect` and it's implementation as we are not carrying any auth related config in parameters.

features/collection/shell-commands/src/test/java/org/opennms/netmgt/collection/commands/CollectCommandAdaptorTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * Verifies that {@link CollectCommand#registerAdaptors(CollectorRequestBuilder)}
+ * pulls every {@link CollectorAdaptor} OSGi service and registers it on the
+ * request builder. Without this, ad-hoc {@code opennms:collect} bypasses
+ * adaptors like {@code TokenAuthCollectorAdaptor} and {@code ${token:...}}
+ * placeholders go out unresolved.
+ */
+@RunWith(MockitoJUnitRunner.class)
+public class CollectCommandAdaptorTest {
```
cgorantla: No need for this test

opennms-config/src/main/java/org/opennms/netmgt/config/tokenauth/TokenAuthConfigReloadListener.java
```diff
@@ -0,0 +1,172 @@
...
+ * <p>Late-binding registration: this bean lives in the daoContext, which
+ * is loaded as a parent of the eventDaemonContext while that context is
+ * still being constructed -- so the {@link EventIpcManager} OSGi service
+ * isn't published yet when this bean is wired. We listen for
+ * {@link ContextRefreshedEvent}s, and on each one we try
+ * {@link EventIpcManagerFactory#getIpcManager()}; once it succeeds we
+ * register as a listener (one-shot, guarded by {@code registered}).</p>
+ */
+public class TokenAuthConfigReloadListener
+        implements EventListener, ApplicationListener<ContextRefreshedEvent>, DisposableBean {
```
cgorantla, line 65: `ApplicationListener<ContextRefreshedEvent>` is not the pattern we use for this. But I leave it at that as there is no easy alternative here.

cgorantla (commented):
Looks like we are adding tests for every source file.
 Not sure if all of them are worth it. May be worth looking at reducing them.

opennms-config/src/main/java/org/opennms/netmgt/config/tokenauth/TokenAuthConfigReloadListener.java
marshallmassengill: Yeah... there are some interactions at play here.  I'm willing to refactor but I'm not sure how.  As it is, this is needed becauseEventIpcManager isn't published when the daoContext constructs this bean... at least as I understand it.

protocols/xml/src/main/java/org/opennms/protocols/http/HttpUrlConnection.java
```diff
@@ -185,6 +186,13 @@ public InputStream getInputStream() throws IOException {
...
             // Get Response
             CloseableHttpResponse response = m_clientWrapper.execute(request);
+            final int status = response.getStatusLine().getStatusCode();
+            if (status >= 400) {
+                EntityUtils.consumeQuietly(response.getEntity());
+                // "auth failure:" prefix on 401/403 lets the adaptor gate cache invalidation.
+                final String prefix = (status == 401 || status == 403) ? "auth failure: " : "";
+                throw new IOException(prefix + "HTTP " + status + " from " + m_url);
+            }
             return response.getEntity().getContent();
```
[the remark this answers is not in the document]
marshallmassengill: Will resolve.

opennms-config/src/main/java/org/opennms/netmgt/config/tokenauth/TokenAuthCollectorAdaptor.java
```diff
@@ -0,0 +1,232 @@
...
+        if (v instanceof String) {
+            return ((String) v).contains("${token:");
+        }
+        if (Interpolator.isPleaseInterpolate(v)) {
+            return mayContainTokenPlaceholder(Interpolator.unwrap(v));
+        }
+        if (v.getClass().isAnnotationPresent(XmlRootElement.class)) {
+            final String marshaled = JaxbUtils.marshal(v);
+            return marshaled != null && marshaled.contains("${token:");
+        }
```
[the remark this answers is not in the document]
marshallmassengill: I don't think this is an actual problem worth addressing.

features/collection/client-rpc/src/main/java/org/opennms/netmgt/collection/client/rpc/CollectorRequestBuilderImpl.java
```diff
@@ -148,28 +164,83 @@ public CompletableFuture<CollectionSet> execute() {
...
-        final Map<String, Object> allAttributes = new HashMap<>();
-        allAttributes.putAll(interpolatedAttributes);
-        allAttributes.putAll(runtimeAttributes);
+        // which should be included in the request.
+        Map<String, Object> rawRuntimeAttributes = serviceCollector.getRuntimeAttributes(agent, interpolatedAttributes);
+        // Same as above: adaptors substitute before Mate sees the tree.
+        for (final CollectorAdaptor adaptor : adaptors) {
+            final Map<String, Object> next = adaptor.beforeRuntimeInterpolation(agent, rawRuntimeAttributes);
+            rawRuntimeAttributes = next != null ? next : rawRuntimeAttributes;
+        }
```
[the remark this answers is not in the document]
marshallmassengill: This is just wrong.  Not going to address.

features/collection/shell-commands/src/main/java/org/opennms/netmgt/collection/commands/CollectCommand.java
```diff
@@ -191,6 +206,28 @@ public Void execute() {
...
+        } catch (InvalidSyntaxException e) {
+            return builder;
+        }
+        if (refs == null) {
+            return builder;
+        }
+        for (final ServiceReference<CollectorAdaptor> ref : refs) {
+            final CollectorAdaptor adaptor = bundleContext.getService(ref);
+            if (adaptor != null) {
+                builder = builder.withAdaptor(adaptor);
```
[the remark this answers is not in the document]
marshallmassengill: Will resolve.

features/collection/shell-commands/src/main/java/org/opennms/netmgt/collection/commands/CollectCommand.java
```diff
@@ -117,6 +123,9 @@ public class CollectCommand implements Action {
     @Reference
     private PersisterFactory persisterFactory;
 
+    @Reference
```
cgorantla, line 122: you can remove `bundleContext` reference and use  List<CollectorAdaptor> adaptors;   I think that should work

marshallmassengill: Good find.  Implemented.  Much cleaner.

opennms-config/src/main/java/org/opennms/netmgt/config/tokenauth/TokenAuthCollectorAdaptor.java
```diff
@@ -0,0 +1,232 @@
...
+    public void setEntityScopeProvider(final EntityScopeProvider entityScopeProvider) {
+        this.entityScopeProvider = entityScopeProvider;
+    }
+
+    @Override
+    public Map<String, Object> beforeRuntimeInterpolation(final CollectionAgent agent,
+                                                          final Map<String, Object> runtimeAttributes) {
+        if (runtimeAttributes == null || runtimeAttributes.isEmpty()) {
+            return runtimeAttributes;
+        }
```
cgorantla, line 86: May be we can exit early when there is no auth configured at all 

```
if (tokenAuthConfigFactory.getAuths().isEmpty()) {
          return runtimeAttributes;          
    }
```

cgorantla (commented):
One improvement we can do otherwise LGTM!

opennms-config/src/main/java/org/opennms/netmgt/config/tokenauth/TokenAuthCollectorAdaptor.java
marshallmassengill: Latest has a fix for this.  Good suggestion.

features/collection/collectors/src/main/java/org/opennms/netmgt/collectd/HttpCollector.java
```diff
@@ -259,9 +259,22 @@ private static void doCollection(final HttpCollectorAgent collectorAgent, final
...
+            try (final CloseableHttpResponse response = clientWrapper.execute(method)) {
+                final int status = response.getStatusLine().getStatusCode();
+                if (status == 401 || status == 403) {
+                    // Auth-related refusal. Drain the entity (the
+                    // try-with-resources closes the response itself)
+                    // and surface as a collection failure so the
+                    // controller-side TokenAuthCollectorAdaptor can
+                    // invalidate any cached token used in this request.
+                    EntityUtils.consumeQuietly(response.getEntity());
+                    throw new HttpCollectorException("auth failure: "
```
cgorantla, line 280: I think all HttpCollectorExceptions gets caught here and resolves to just returning FAILED status
https://github.com/OpenNMS/opennms/blob/1a726bf5b878efd20d973b9b16702e3c6aeb8524/features/collection/collectors/src/main/java/org/opennms/netmgt/collectd/HttpCollector.java#L177

cgorantla (commented):
There is one more bug that might still be there.

features/collection/collectors/src/main/java/org/opennms/netmgt/collectd/HttpCollector.java
marshallmassengill: Hopefully addressed.

cgorantla (approved):
LGTM!

PR #8497, opened 2026-05-01
NMS-19753: Core: Investigate if its possible to move away from Confd

mershad-manesh:
### All Contributors

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19753


marshallmassengill:
I looked through everything I could and there wasn't anything that felt truly off to me.  I also used CC and did a review on this.  These were the only two things that I think were worth mentioning here that got flagged and both seem mostly minor:

  - prom-jmx-default-config.yaml lost env-var-driven knobs. The old confd template let operators tune startDelaySeconds, lowercaseOutputName, lowercaseOutputLabelNames, autoExcludeObjectNameAttributes, includeObjectNames, and excludeObjectNames via the /java/agent/prom-jmx-exporter/* keys (which mapped to env vars under confd's env backend). The new file is fully static; the only escape hatch is "mount your own YAML and point PROM_JMX_EXPORTER_CONFIG at it." For users who were tweaking just includeObjectNames, that's a meaningful UX downgrade — they now have to maintain the entire YAML. Either keep an envsubst path for this file too, or call out the regression explicitly in the migration section.
  - No XML-escaping on envsubst substitutions. service-configuration.xml.tmpl and trapd-configuration.xml.tmpl are XML files where ${OPENNMS_TRAPD_ADDRESS} / ${CORE_SERVICE_*_ENABLED} get raw-interpolated. If an operator passes a value containing <, >, &, or ", the resulting XML breaks. The expected values are true/false and an IP address, so this is operator-controlled and unlikely to be exploited, but a minute of input validation in processEnvConfig (e.g. assert booleans match ^(true|false)$) would prevent a confusing parse failure later in configTester

opennms-container/core/container-fs/etc/templates/service-configuration.xml.tmpl
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0"?>
```
cgorantla: This is not needed after https://github.com/OpenNMS/opennms/pull/8510

opennms-container/core/container-fs/etc/templates/trapd-configuration.xml.tmpl
```diff
@@ -0,0 +1 @@
+<trapd-configuration xmlns="http://xmlns.opennms.org/xsd/config/trapd" snmp-trap-address="${OPENNMS_TRAPD_ADDRESS}" snmp-trap-port="${OPENNMS_TRAPD_PORT}" new-suspect-on-trap="${OPENNMS_TRAPD_NEW_SUSPECT_ON_TRAP}" include-raw-message="${OPENNMS_TRAPD_INCLUDE_RAW_MESSAGE}" threads="${OPENNMS_TRAPD_THREADS}" queue-size="${OPENNMS_TRAPD_QUEUE_SIZE}" batch-size="${OPENNMS_TRAPD_BATCH_SIZE}" batch-interval="${OPENNMS_TRAPD_BATCH_INTERVAL}"/>
```
cgorantla, line 1: trapd configuration is completely in DB, we don't need this anymore

opennms-container/core/container-fs/entrypoint.sh
```diff
@@ -100,9 +100,145 @@ configTester() {
...
+  # Copy static config files; Karaf resolves ${env:VAR:-default} at load time
+  mkdir -p "${OPENNMS_HOME}/etc/opennms.properties.d"
+  rsync -r "${CONTAINER_CONFIG_ETC}/opennms.properties.d/" "${OPENNMS_HOME}/etc/opennms.properties.d/"
+  cp "${CONTAINER_CONFIG_ETC}/org.apache.karaf.shell.cfg" "${OPENNMS_HOME}/etc/"
+
+  # Remove legacy confd-generated property files to prevent stale/duplicate settings
+  rm -f "${OPENNMS_HOME}/etc/opennms.properties.d/"_confd.*.properties
+
+  # Process service-configuration.xml from template with defaults for unset variables
+  (
```
cgorantla: can be deleted after https://github.com/OpenNMS/opennms/pull/8510

opennms-container/core/container-fs/entrypoint.sh
```diff
@@ -100,9 +100,145 @@ configTester() {
...
+    validateBool CORE_SERVICE_TELEMETRYD_ENABLED          "$CORE_SERVICE_TELEMETRYD_ENABLED"
+    validateBool CORE_SERVICE_TRAPD_ENABLED               "$CORE_SERVICE_TRAPD_ENABLED"
+    validateBool CORE_SERVICE_PERSPECTIVEPOLLER_ENABLED   "$CORE_SERVICE_PERSPECTIVEPOLLER_ENABLED"
+
+    envsubst < "${CONTAINER_CONFIG_ETC}/templates/service-configuration.xml.tmpl" \
+              > "${OPENNMS_HOME}/etc/service-configuration.xml"
+  )
+
+  # Process trapd-configuration.xml from template with defaults for unset variables
+  (
```
cgorantla: same here. None of these will work with Trapd config in DB

cgorantla (changes requested):

PR #8499, opened 2026-05-01
NMS-19640: Sentinel: Investigate if its possible to move away from Confd

mershad-manesh:
### All Contributors

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19640


docs/modules/deployment/pages/prometheus-monitoring.adoc
```diff
@@ -231,11 +231,10 @@ In the container images the JMX Prometheus Exporter is bundled with the containe
 The metric exporter is disabled by default and you can enable it by setting the environment variable `PROM_JMX_EXPORTER_ENABLED=true`.
 Configure your Prometheus to scrape the metrics from the HTTP endpoint http://your-app:9299/metrics.
 Replace `your-app` with the host or domain name where your Core, Minion or Sentinel is running in your network.
-The environment variables can be used for Core, Minion and Sentinel in the same way and are documented in the xref:reference:configuration/core-docker.adoc#core-oci-confd[docker environment] section.
+The environment variables can be used for Core, Minion, and Sentinel in the same way and are documented in the xref:reference:configuration/core-docker.adoc#core-oci-confd[Core docker environment] section.
```
christianpape, line 234: The anchor is named *-confd. This should be changed, right?

docs/modules/deployment/pages/prometheus-monitoring.adoc
```diff
@@ -231,11 +231,10 @@ In the container images the JMX Prometheus Exporter is bundled with the containe
...
+To customize the JMX exporter configuration, inject your own `config.yaml` file to the path configured with `PROM_JMX_EXPORTER_CONFIG`.
 
-IMPORTANT: In confd you have access to some basic JMX Prometheus Exporter configurations. If you need a more sophisticated configuration with JMX rules, you need to inject your own `config.yaml` file.
+IMPORTANT: If you need a more sophisticated configuration with JMX rules, you need to inject your own `config.yaml` file.
 
-For configuring container deployments with Confd refer to xref:reference:configuration/core-docker.adoc#core-oci-confd[Core Confd], xref:reference:configuration/minion-confd/minion-confd.adoc#minion-confd[Minion Confd], and xref:reference:configuration/sentinel-features.adoc#sentinel-oci-confd[Sentinel Confd].
+For configuring container deployments, refer to xref:reference:configuration/core-docker.adoc#core-oci-confd[Core], xref:reference:configuration/minion-docker.adoc#minion-docker[Minion], and xref:reference:configuration/sentinel-features.adoc#sentinel-configuration[Sentinel].
```
christianpape, line 240: Same here.

christianpape (changes requested):

opennms-container/sentinel/container-fs/entrypoint.sh
```diff
@@ -94,6 +94,83 @@ setCredentials() {
...
+    # If config exists in file, replace it. Otherwise, append to file.
+    if grep -E -q "^#?\s*$key\s*=" "$file"; then
+        sed -r -i "s@^#?\s*$key\s*=.*@$key=$value@g" "$file" #note that no config values may contain an '@' char
+    else
+        echo "$key=$value" >> "$file"
+    fi
+}
+
+function parseEnvironment() {
+    IFS=$'\n'
```
christianpape: parseEnvironment() sets IFS=$'\n' without local, and now runs unconditionally on every start. The leak hits applyKarafDebugLogging() at line 241, where for log in $(sed "s/,/ /g" ...) no longer splits on spaces. KARAF_DEBUG_LOGGING=org.opennms.a,org.opennms.b writes the malformed key log4j2.logger.org_opennms_a org_opennms_b.level = DEBUG into org.ops4j.pax.logging.cfg — an invalid log4j2.* root key there breaks logging config for the whole container. Fix: local IFS=$'\n'.

opennms-container/sentinel/container-fs/entrypoint.sh
```diff
@@ -94,6 +94,83 @@ setCredentials() {
...
+      "sentinel-ipc.boot"
+    )
+    local boot_file
+    for boot_file in "${managed_boot_files[@]}"; do
+      rm -f "${FEATURES_BOOT_DIR}/${boot_file}"
+    done
+
+    apply_template() {
+        local name="$1"
+        cp "${FEATURES_BOOT_TEMPLATES_DIR}/${name}" "${FEATURES_BOOT_DIR}/${name}"
```
christianpape: Config moved from runtime generation into COPY-ed files under container-fs/etc/, but /opt/sentinel/etc is a declared VOLUME (Dockerfile:104). Upgrading a deployment with an existing named/bind-mounted etc volume leaves it without the new files. With SENTINEL_IPC=kafka, the cp "${FEATURES_BOOT_TEMPLATES_DIR}/sentinel-ipc.boot" fails, and the new set -eE + ERR trap turns that into a crash-loop. With JMS it's quieter but worse: missing org.opennms.netmgt.distributed.datasource.cfg / org.opennms.sentinel.controller.cfg means datasource and broker-url silently fall back to blueprint localhost. Generate placeholders from the entrypoint when absent rather than relying on COPY alone.

opennms-container/sentinel/container-fs/etc/org.opennms.features.flows.persistence.elastic.cfg
```diff
@@ -0,0 +1 @@
+elasticUrl=${env:ELASTICSEARCH_URL:-http://elasticsearch:9200}
```
christianpape: The cfg now always exists with elasticUrl=${env:ELASTICSEARCH_URL:-http://elasticsearch:9200}. Previously confd created the file only when /elasticsearch/url was set, and the bundle default is http://localhost:9200. So every sentinel container — including ones not using flows — now points at hostname elasticsearch, which only resolves in the docker-compose examples. Same shape for sink.kafka{,.consumer}.cfg defaulting to kafka:29092: forgetting KAFKA_IPC_BOOTSTRAP_SERVERS yields a wrong-host connection instead of a clear failure.

opennms-container/sentinel/container-fs/entrypoint.sh
```diff
@@ -94,6 +94,83 @@ setCredentials() {
...
+    do
+        env_var=$(echo "$VAR" | cut -d= -f1)
+
+        if [[ $env_var =~ ^KAFKA_IPC_ ]]; then
+            ipc_name=$(echo "$env_var" | cut -d_ -f3- | tr '[:upper:]' '[:lower:]' | tr _ .)
+            updateConfig "$ipc_name" "${!env_var}" "${SENTINEL_HOME}/etc/org.opennms.core.ipc.sink.kafka.cfg"
+            updateConfig "$ipc_name" "${!env_var}" "${SENTINEL_HOME}/etc/org.opennms.core.ipc.sink.kafka.consumer.cfg"
+        fi
+
+        if [[ $env_var =~ ^ELASTICSEARCH_ ]]; then
```
christianpape: [[ $env_var =~ ^ELASTICSEARCH_ ]] matches anything with that prefix, and the *) fallback writes it verbatim. In Kubernetes a Service named elasticsearch in the same namespace auto-injects ELASTICSEARCH_SERVICE_HOST, ELASTICSEARCH_PORT_9200_TCP_ADDR, … into every pod (legacy Docker links too), so the cfg accumulates junk keys like service.host=… on every restart. Conversely, because the fallback lowercases and dot-joins, nothing beyond the five aliased keys can ever match this PID's camelCase properties (globalUser, globalPassword) — those env vars silently do nothing. Use an explicit allowlist, as the confd template did.

.circleci/epoch
```diff
@@ -1,2 +1,2 @@
 # increment this to force a full build
-34
+29
```
christianpape: Head branch has 29; merge base and base branch have 34. Merging decrements it. Looks like a merge-resolution artifact from 85e672bf1af/1a72ad47cb9, and it silently reverts cache-invalidation state other branches depend on.

opennms-container/sentinel/container-fs/entrypoint.sh
```diff
@@ -94,6 +94,83 @@ setCredentials() {
...
+    key=$1
+    value=$2
+    file=$3
+
+    # Omit $value here, in case there is sensitive information
+    echo "[Configuring] '$key' in '$file'"
+
+    # If config exists in file, replace it. Otherwise, append to file.
+    if grep -E -q "^#?\s*$key\s*=" "$file"; then
+        sed -r -i "s@^#?\s*$key\s*=.*@$key=$value@g" "$file" #note that no config values may contain an '@' char
```
christianpape: substitutes values into a sed replacement, so &, @ (the delimiter) or \ corrupt the file. The append branch masks it on first boot; every restart takes the sed branch. Real case: KAFKA_IPC_SASL_JAAS_CONFIG='…password="p&q";' works once, then mangles.

docs/modules/reference/pages/configuration/sentinel-features.adoc
```diff
@@ -328,101 +328,114 @@ admin@sentinel> config:update
...
+| `KAFKA_IPC_GROUP_ID`          | Kafka consumer group ID. | `OpenNMS`
+| `KAFKA_IPC_*`                 | Any additional Kafka client property. | `KAFKA_IPC_AUTO_OFFSET_RESET=latest`
+|===
+
+=== Datasource configuration
+
+[options="header, autowidth"]
+|===
+| Variable | Description | Default value
+| `POSTGRES_HOST`     | PostgreSQL host. | –
```
christianpape: This lists "–" as the default for OPENNMS_BROKER_URL and all POSTGRES_*, but Dockerfile:141-153 actually defaults them to tcp://127.0.0.1:61616, localhost, 5432, opennms, postgres. Readers will expect fail-fast where it quietly hits localhost.

PR #8500, opened 2026-05-01
NMS-19754: DCB: Add inline SSH configuration capture

dino2gnt:
Adds a `capture: ` DCB DSL statement that can be used to scrape the config out of the SSH session output without pushing it via TFTP.

Tested locally using a dcb script of:
```
await: ]$
send: cat /etc/hosts
await: ]$
send: cat /etc/snmp/snmpd.conf
capture: cat /etc/snmp/snmpd.conf
await: ]$
```
Only the output of `cat /etc/snmp/snmpd.conf` is captured as the configuration in this case.  

Bare `capture: ` with no args will capture everything from that point to the end of the output.  I think both cases are useful.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19754

Assisted-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/features/device-config/retrieval/src/main/java/org/opennms/features/deviceconfig/retrieval/impl/RetrieverImpl.java
+++ b/features/device-config/retrieval/src/main/java/org/opennms/features/deviceconfig/retrieval/impl/RetrieverImpl.java
@@ -38,6 +38,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
+import java.util.stream.Stream;
 
 import org.opennms.core.concurrent.FutureUtils;
 import org.opennms.features.deviceconfig.retrieval.api.Retriever;
@@ -103,6 +104,37 @@ public class RetrieverImpl implements Retriever, AutoCloseable {
         // set the ip address and port of the tftp server
         vs.put(SCRIPT_VAR_TFTP_SERVER_PORT, String.valueOf(tftpServer.getPort()));
         vs.put(SCRIPT_VAR_CONFIG_TYPE, configType);
+
+        // If the script contains a 'capture:' statement, capture device output inline.
+        // No file transfer (TFTP/SCP/FTP) is used — the captured bytes come back in scriptResult.capturedConfig.
+        if (scriptContainsCapture(script)) {
+            return CompletableFuture.supplyAsync(() -> {
+                SshScriptingService.Result scriptResult;
+                try {
+                    scriptResult = sshScriptingService.execute(
+                            script, user, password, authKey, target, hostKeyFingerprint, shell, vs, timeout);
+                } catch (Exception e) {
+                    var msg = scriptingFailureMsg(target, e.getMessage());
+                    LOG.error(msg, e);
+                    return Either.<Failure, Success>left(new Failure(msg));
+                }
+
+                if (scriptResult.isFailed()) {
+                    return Either.<Failure, Success>left(new Failure(
+                            scriptingFailureMsg(target, scriptResult.message),
+                            scriptResult.stdout, scriptResult.stderr, scriptResult.scriptOutput));
+                }
+
+                if (scriptResult.capturedConfig.isEmpty()) {
+                    return Either.<Failure, Success>left(new Failure(
+                            "no config captured - script must include a 'capture:' statement - target: " + target,
+                            scriptResult.stdout, scriptResult.stderr, scriptResult.scriptOutput));
+                }
+
+                return Either.<Failure, Success>right(
+                        new Success(scriptResult.capturedConfig.get(), configType, scriptResult.scriptOutput));
+            }, executor);
+        }
 
         if (protocol == Protocol.TFTP) {
             // Keep timeout common between TFTP server and scripting service
@@ -228,6 +260,12 @@ public class RetrieverImpl implements Retriever, AutoCloseable {
                 }
             }
         }
+    }
+
+    static boolean scriptContainsCapture(String script) {
+        return Stream.of(script.split("\\\\n|\\n"))
+                .map(String::trim)
+                .anyMatch(line -> line.equals("capture:") || line.startsWith("capture:"));
     }
 
     static String scriptingFailureMsg(final SocketAddress target, String msg) {
--- a/features/device-config/retrieval/src/main/java/org/opennms/features/deviceconfig/retrieval/impl/RetrieverImpl.java
+++ b/features/device-config/retrieval/src/main/java/org/opennms/features/deviceconfig/retrieval/impl/RetrieverImpl.java
@@ -125,9 +125,9 @@ public class RetrieverImpl implements Retriever, AutoCloseable {
                             scriptResult.stdout, scriptResult.stderr, scriptResult.scriptOutput));
                 }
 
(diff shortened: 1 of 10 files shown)


marshallmassengill:
Add:
```
  "attribution": {
    "commit": ""
  },
```
to your ~./claude/settings.json file so you don't pick up the "co-authored" by claude on the commits.  Just put a comment in the PR: "Assisted-by: Claude Code: Sonnet 4.6".  Keeps the commit history clean for authors.

features/device-config/retrieval/src/main/java/org/opennms/features/deviceconfig/retrieval/impl/RetrieverImpl.java
```diff
@@ -104,6 +105,37 @@ public CompletionStage<Either<Failure, Success>> retrieveConfig(
...
+
+                if (scriptResult.isFailed()) {
+                    return Either.<Failure, Success>left(new Failure(
+                            scriptingFailureMsg(target, scriptResult.message),
+                            scriptResult.stdout, scriptResult.stderr, scriptResult.scriptOutput));
+                }
+
+                if (scriptResult.capturedConfig.isEmpty()) {
+                    return Either.<Failure, Success>left(new Failure(
+                            "no config captured - script must include a 'capture:' statement - target: " + target,
```
marshallmassengill: The wording here seems off to me.  It's basically an empty response, right?

Maybe something like: 
"capture: statement parsed but produced no output - target: " + target
 or
"internal error: capture: statement did not yield any captured bytes - target: " + target

features/device-config/retrieval/src/main/java/org/opennms/features/deviceconfig/retrieval/impl/RetrieverImpl.java
```diff
@@ -230,6 +262,12 @@ public void onFileReceived(InetAddress address, String fileName, byte[] content)
         }
     }
 
+    static boolean scriptContainsCapture(String script) {
+        return Stream.of(script.split("\\\\n|\\n"))
+                .map(String::trim)
+                .anyMatch(line -> line.equals("capture:") || line.startsWith("capture:"));
```
marshallmassengill: You can drop the "line.equals("capture:")" and just use the startsWith, I think?  Seems like one would get the other.

features/device-config/ssh-scripting/src/main/java/org/opennms/features/deviceconfig/sshscripting/impl/Statement.java
```diff
@@ -89,6 +89,34 @@ public void execute(SshInteraction interaction, String string) throws Exception
...
+            /**
+             * {@code capture:} may optionally take a string argument.
+             * With no argument ({@code capture:}), capture starts immediately at the current stdout position.
+             * With an argument ({@code capture: show running-config}), the argument is first awaited
+             * (consuming the device's echo of the preceding send: command), then capture starts —
+             * so the echoed command line is excluded from the captured bytes.
+             */
+            @Override
+            public Statement parse(String line) {
+                if (line.equals(name() + ":")) {
```
marshallmassengill, line 102: I asked this above but I don't think you need both line.equals and line.startsWith.  I think you can remove the line.equals path here.

features/device-config/retrieval/src/test/java/org/opennms/features/deviceconfig/retrieval/impl/RetrieverImplTest.java
```diff
@@ -28,8 +28,10 @@
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyBoolean;
```
marshallmassengill: This doesn't seem to ever be called.

marshallmassengill (changes requested):

marshallmassengill:
One more thing that showed up when I asked CC about test coverage:

- No test exercises capture: <string> (i.e. awaitAndCapture) in either StatementTest (parse-only there) or RetrieverImplTest — only bare capture: is in the unit tests.
- The unit tests mock the SshScriptingService and pre-bake capturedConfig bytes, so the byte-offset bookkeeping in SshInteractionImpl (captureStart/captureEnd/trim) is not exercised by any automated test. An end-to-end test in SshScriptingServiceImplIT against the existing in-process Apache MINA SSHD echo server would catch the trim/concurrency edges and the prompt-exclusion behavior the docs mention.
- Add at least one IT in SshScriptingServiceImplIT covering both capture: forms end-to-end against the embedded SSHD.

features/device-config/ssh-scripting/src/main/java/org/opennms/features/deviceconfig/sshscripting/impl/Statement.java
dino2gnt: We can do as you describe in the instance you pointed put above, but you cannot here.  `capture:` and `capture: <string>` actually behave differently.

marshallmassengill (dismissed):
Changes look good.  There is still the attribution to claude in the first commit.  It may be worth fixing that with a force push but that may also mean you need squash the commits.  I'll let @cgorantla comment on that.

dino2gnt:
> Changes look good. There is still the attribution to claude in the first commit. It may be worth fixing that with a force push but that may also mean you need squash the commits. I'll let @cgorantla comment on that.

Ah.  I never had any intention of fixing that.  Why do we care?

marshallmassengill:
clean commiter history

dino2gnt:
> clean commiter history

If Claude wrote it, all or in part, and Claude committed it, I don't see any reason that Claude shouldn't have a `co-authored-by` tag. It's not dirtying up the commit history, it's keeping it _honest_.

features/device-config/retrieval/src/main/java/org/opennms/features/deviceconfig/retrieval/impl/RetrieverImpl.java
```diff
@@ -104,6 +105,37 @@ public CompletionStage<Either<Failure, Success>> retrieveConfig(
...
+                    return Either.<Failure, Success>left(new Failure(msg));
+                }
+
+                if (scriptResult.isFailed()) {
+                    return Either.<Failure, Success>left(new Failure(
+                            scriptingFailureMsg(target, scriptResult.message),
+                            scriptResult.stdout, scriptResult.stderr, scriptResult.scriptOutput));
+                }
+
+                if (scriptResult.capturedConfig.isEmpty()) {
```
cgorantla: May be there are other conditions like whitespace, single character returns like `\n` or \r`.  These cases can be real failures.

cgorantla (commented):
One minor thing to fix

It would be good to test this with some real devices and address any corner cases.

features/device-config/retrieval/src/main/java/org/opennms/features/deviceconfig/retrieval/impl/RetrieverImpl.java
dino2gnt: This is a very good point.

I added a check for printable characters in the captured output, so if all we get back is non-printable chars or whitespace, it should fail.

cgorantla (approved):
LGTM!

PR #8506, opened 2026-05-05
NMS-19772: Fix blank standalone geomap page

marshallmassengill:
Fixes the blank `geomap/standalone.jsp` page reported by customers.

## What this PR contains

This is one small required fix bundled with two separate improvements that the same files made convenient. Calling them out explicitly because only the first is needed to resolve the reported bug.

### 1. The actual fix — `standalone.jsp` height calculation

`refresh()` sized `#map-container` as:

```js
var height = $(window).height() - $("#footer").outerHeight() - $("#header").outerHeight();
```

`#header` was removed when the top navbar moved to the side menu (`#opennms-sidemenu-container`). `$("#header").outerHeight()` returns `undefined` on an empty selector, the arithmetic produces `NaN`, `$("#map-container").height(NaN)` is a no-op, the `.geomap` and `#map` divs cascade at `height: 100%` to a zero-height parent, and leaflet mounts onto a 0×0 element. The map is functionally there; it just has no pixels.

The fix defaults a missing selector's height to 0 so the math stays valid. **This single change is sufficient to make the page render.**

### 2. Additionally deprecation hygiene in `map.jsp`

The inline script used `$('<%= mapId %>').ready(handler)`, which expands to `$('map').ready(handler)` — a tag selector matching the HTML `<map>` element (none on the page) combined with the legacy `$(selector).ready()` API that was deprecated in jQuery 1.8.

This still works today only because `jquery-migrate` restores it (with a deprecation warning). It will silently break when migrate is eventually dropped. Replaced with the canonical `$(function() {...})` shortcut. **Not a fix for the customer-reported bug** — empirically `geomap.render()` was running fine via the migrate path before this change. Treating it as latent regression cleanup so a future jQuery bump doesn't reintroduce a different blank-page failure.

### 3. Additionally separate latent bug in parameter rendering

When the page is hit without `?strategy=...&severity=...`, the JSP previously emitted the literal four-character JavaScript string `"null"` (`<%= null %>` writes the word) into the inline script. The geomap-js client's `isUndefinedOrNull(...)` check treats the string `"null"` as a real value and forwards it to `/api/v2/geolocation/config?strategy=null`, which breaks **marker loading** (not the basic page render). Parameterless embeds appeared to "work" but never showed any node markers.

`GeomapPageHelper` renders missing parameters as the JS `null` literal (no quotes), so the client-side default substitution fires correctly. **Independent of the blank-page bug**, but shipping in the same PR because the same JSP and same iframe consumer are involved.

### Tests
- `GeomapPageHelperTest` — pins the param-default rendering rules (null → bare `null`, value → sanitized + quoted).
- `GeomapStandalonePageIT` — Selenium smoke test asserting (a) leaflet container mounts, (b) `#map-container` has non-zero rendered height — the assertion that catches the height-calc regression even when the DOM-presence check would still pass on a 0×0 container, and (c) the inline script renders `strategy: null` (not `strategy: "null"`) without query params and `strategy: "Outages"` with them.

## Branches affected

`standalone.jsp` and `map.jsp` are byte-identical across `foundation-2024`, `foundation-2025`, and `develop` (one comment-block difference in `map.jsp` aside, unrelated to this fix). Author against `foundation-2024` and forward-merge.

Assisted-by: Claude Code:Opus 4.7

### DIFF
--- a/opennms-webapp/src/main/java/org/opennms/web/geomap/GeomapPageHelper.java
+++ b/opennms-webapp/src/main/java/org/opennms/web/geomap/GeomapPageHelper.java
@@ -1,0 +1,73 @@
+/*
+ * Licensed to The OpenNMS Group, Inc (TOG) under one or more
+ * contributor license agreements.  See the LICENSE.md file
+ * distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * TOG licenses this file to You under the GNU Affero General
+ * Public License Version 3 (the "License") or (at your option)
+ * any later version.  You may not use this file except in
+ * compliance with the License.  You may obtain a copy of the
+ * License at:
+ *
+ *      https://www.gnu.org/licenses/agpl-3.0.txt
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied.  See the License for the specific
+ * language governing permissions and limitations under the
+ * License.
+ */
+package org.opennms.web.geomap;
+
+import org.opennms.core.utils.WebSecurityUtils;
+
+/**
+ * Helpers used by {@code geomap/includes/map.jsp} to render request
+ * parameters as JavaScript literals embedded directly into a
+ * {@code <script>} block. Centralized here so the conversion is unit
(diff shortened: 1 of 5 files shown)


synqotik (dismissed):
LGTM!

dino2gnt (dismissed):
The changes in standalone.jsp are enough to get the map back, for me.

christianpape (approved):
LGTM.

PR #8507, opened 2026-05-05
NMS-19773: Emit an event when config changes between DCB backup intervals

dino2gnt:
Check the incoming device config against the previous config and emit a `configChanged` event if the two are different.  As some devices [citation needed] like to insert nonsense like timestamps or other rotating content into commented headers in their configs, also offer the capability to optionally ignore commented content in the device configs when comparing the two versions.

Targeted this to `develop`, not sure if this should use `release-36.x` instead, or if we're still merging one into the other, or what the deal is.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19773

Assisted-By: Claude Sonnet 4.6

### DIFF
--- a/features/device-config/api/src/main/java/org/opennms/features/deviceconfig/service/DeviceConfigUtil.java
+++ b/features/device-config/api/src/main/java/org/opennms/features/deviceconfig/service/DeviceConfigUtil.java
@@ -27,10 +27,55 @@ import com.google.common.io.ByteStreams;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.nio.charset.Charset;
+import java.nio.charset.IllegalCharsetNameException;
+import java.nio.charset.UnsupportedCharsetException;
+import java.util.Arrays;
 import java.util.zip.GZIPInputStream;
 
 public class DeviceConfigUtil {
-    
+
+    /**
+     * Returns true if two config byte arrays are considered equal.
+     * Comparison is done as text using the given encoding.
+     * When {@code ignoreComments} is true, C-style comments ({@code //}, {@code #},
+     * {@code /* ... *\/}) are stripped from both sides before comparing.
+     */
+    public static boolean configsAreEqual(byte[] a, byte[] b, String encoding, boolean ignoreComments) {
+        if (a == null && b == null) return true;
+        if (a == null || b == null) return false;
+        if (!ignoreComments) {
+            return Arrays.equals(a, b);
+        }
+        try {
+            Charset charset = Charset.forName(encoding);
+            return stripComments(new String(a, charset)).equals(stripComments(new String(b, charset)));
+        } catch (IllegalCharsetNameException | UnsupportedCharsetException e) {
+            return Arrays.equals(a, b);
+        }
+    }
+
+    /**
+     * Removes C-style comments from a configuration string:
+     * <ul>
+     *   <li>{@code /* ... *\/} block comments (including multi-line)</li>
+     *   <li>{@code //} line comments</li>
+     *   <li>{@code #} line comments</li>
+     * </ul>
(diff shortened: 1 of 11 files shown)


features/device-config/api/src/main/java/org/opennms/features/deviceconfig/service/DeviceConfigUtil.java
```diff
@@ -27,10 +27,55 @@
...
+     *   <li>{@code #} line comments</li>
+     * </ul>
+     * Lines that are empty or whitespace-only after removal are also dropped,
+     * so a diff that only adds or removes comments is not treated as a change.
+     */
+    public static String stripComments(String content) {
+        // block comments /* ... */ — (?s) makes . match newlines
+        content = content.replaceAll("(?s)/\\*.*?\\*/", "");
+        // // line comments
+        content = content.replaceAll("//[^\r\n]*", "");
```
christianpape, line 72: This might be a problem, since URLs like tftp://... or https://... are often used in configurations. Even encrypted passwords may have these characters.

features/device-config/api/src/main/java/org/opennms/features/deviceconfig/service/DeviceConfigUtil.java
```diff
@@ -27,10 +27,55 @@
...
+        // # line comments
+        content = content.replaceAll("#[^\r\n]*", "");
```
christianpape, line 74: Even this character might be used in configurations. I saw it on Cisco, also inside encrypted passwords on other devices.

features/device-config/api/src/main/java/org/opennms/features/deviceconfig/service/DeviceConfigUtil.java
christianpape: If we want to ignore that, we must at least mention it in the documentation.

features/device-config/api/src/main/java/org/opennms/features/deviceconfig/service/DeviceConfigUtil.java
christianpape: If we want to ignore that, we must at least mention it in the documentation.

christianpape (changes requested):
Please add notes to docs, that when enabling `ignore-comments` that /**/, // or # might filter out relevant pieces of the configuration while comparing.

features/device-config/api/src/main/java/org/opennms/features/deviceconfig/service/DeviceConfigUtil.java
marshallmassengill: Good finds.

dino2gnt: That's true, and why it defaults to `false`.  It's pretty liberal in what it will strip out and will even miss some commenting styles used in network device configs that I don't see often, like semicolons or something.

I figured this was better than trying to kludge something that ignores anything that looks like a datestamp before compare.  If there's nothing dynamic inserted into the config that changes every cycle, then this option isn't necessary for that case.

If you have a recommendation or suggestion on how else to handle this I'm all ears.

I'll update the docs to make it clearer that `ignore-comments`  can have undesired side effects on comparisons when enabled.

features/device-config/api/src/main/java/org/opennms/features/deviceconfig/service/DeviceConfigUtil.java
dino2gnt: Added a warning to the docs.

christianpape (approved):
LGTM.

PR #7597, opened 2025-01-17
NMS-17066: Fix Measurement TrendLine

rssntn67:
TrendLine was nnot able to parse Long Term measument data because was not able to manage Double.NaN.
Added unit test with real data fetched

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-17066

PR #8515, opened 2026-05-08
NMS-19750:  resolve failed tests for forecast feature

marshallmassengill:
This should resolve the failed tests @mershad-manesh and it should pass through to F2025 now.


cgorantla (commented):
These seems to be new tests. Which tests are we fixing ?

Also can we update description/header with issue number ? 
Don't need to create new issue but can mention original issue number.

marshallmassengill:
Updated the issue.  It's because I put my fix into foundation-2024 and apparently there was something that got done at some point and added to 2025: https://github.com/OpenNMS/opennms/pull/7597 to try and fix something similar and I didn't initially pick those up.  This should fix that... hopefully.

PR #8517, opened 2026-05-12
NMS-19795: Password Gate: Complexity checking on client side not working correctly

synqotik:
The client side checking for password complexity on the password gate page had a bug where certain passwords that did not contain the required one or more special characters were getting through, resulting in a generic error message.

The error was caused by HTML-encoding the Java regex when included in the Javscript on the JSP page. `&` was being encoded as `&amp;`, and those literal characters were being matched for the password having a special character, so any passwords containing `[a|m|p|;]` would pass.

Note that we should handle the service side error more gracefully - it's thrown when the server detects an invalid password, but currently it looks like the server had an error, and gives no useful information. We can tackle that separately, but with this client side fix, users should not really encounter that any more.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19795


cgorantla (approved):
LGTM!

PR #8521, opened 2026-05-15
Update backshift

cgorantla:
Looks like backshift didn't actually move to 1.3.5 in package-lock.json.
Removing `"node_modules/backshift": ` section and doing clean npm install updated the version.


synqotik:
@cgorantla  We should fix this in a different way, that will be retained.

1. locally, update `package.json`. Under `overrides`, add: `"backshift": "github:OpenNMS/backshift#v1.3.5",`

2. delete `package-lock.json` locally

3. run `npm install`

4. that should regenerate `package-lock.json` with the correct version

5. commit both files

synqotik (approved):
LGTM!

PR #8522, opened 2026-05-15
NMS-19719: Bring SNMP Data Collection Config from plugins to DB

cgorantla:
This update brings SNMP data collection config from plugins to DB.
If there is a match, it will update with config from plugins.
Also disables deleting config from Plugins. Only enable/disable is possible.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19719


synqotik (approved):
LGTM!

PR #8526, opened 2026-05-19
Docs: Upate opennms-full-stack diagram on the docs landing page

marshallmassengill:
The previous diagram_implementation.png references Casandra instead of a more generic Time Series Database. opennms-full-stack.png is the refreshed asset that ships alongside it in the same images directory; switch the image:: macro to reference it so the rendered landing page matches the current artwork.  The AI file for this one exists and are in the same directory.


indigo423 (dismissed):
go for it, you can delete the old one. If we need it, that's why we have git.

marshallmassengill:
Fair point and one I'd have thought of if I'd slept last night.  Will update to remove it.

indigo423 (approved):
yeah :)

PR #8527, opened 2026-05-19
NMS-19800 : 36.0.0: Wrong HOME directories set in Minion

mershad-manesh:
### All Contributors

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19800

### DIFF
--- a/opennms-base-assembly/src/main/filtered/bin/configure-snmp.sh
+++ b/opennms-base-assembly/src/main/filtered/bin/configure-snmp.sh
@@ -38,7 +38,15 @@
 #############################################################
 
 
-OPENNMS_HOME="${install.dir}"
+if [ -z "${OPENNMS_HOME}" ]; then
+  _script="$0"
+  case "$_script" in
+    */*) ;;
+    *) _script="$(command -v "$_script" 2>/dev/null || echo "$_script")" ;;
+  esac
+  OPENNMS_HOME="$(cd "$(dirname "$_script")/.." && pwd)"
+  unset _script
+fi
 SEND_EVENT="${OPENNMS_HOME}/bin/send-event.pl"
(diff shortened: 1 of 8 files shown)


cgorantla (commented):
Do we know why did it break ?

We should probably fix at the rpm/debian layer to set install.dir or OPENNMS_HOME properly and reuse it across scripts ?

mershad-manesh:
> Do we know why did it break ?
> 
> We should probably fix at the rpm/debian layer to set install.dir or OPENNMS_HOME properly and reuse it across scripts ?

These scripts are using  placeholder variables ( install.dir and install.bin.dir) which is resolved at build time instead of runtime.
The change proposed here is to make it detect the `OPENNMS_HOME` at runtime instead of it being hardcoded; this solution works across Debian/RPM and Containers.

cgorantla:
> > Do we know why did it break ?
> > We should probably fix at the rpm/debian layer to set install.dir or OPENNMS_HOME properly and reuse it across scripts ?
> 
> These scripts are using placeholder variables ( install.dir and install.bin.dir) which is resolved at build time instead of runtime. The change proposed here is to make it detect the `OPENNMS_HOME` at runtime instead of it being hardcoded; this solution works across Debian/RPM and Containers.

I understand that.  All our etc and scripts use this pattern. I want to check resolving OPENNMS_HOME was broken somewhere.

Also we are missing changes to two files mentioned in the issue.  

```
fix-permissions:  MINION_HOME='/home/circleci/project/target/minion-36.0.0'
update-package-permissions:  MINION_HOME='/home/circleci/project/target/minion-36.0.0'
```

cgorantla (dismissed):
LGTM!

christianpape (approved):
LGTM.

PR #8535, opened 2026-05-26
NMS-16118: Structured Node List Page: handle query params

synqotik:
We are replacing the legacy JSP Node List (`element/nodeList.htm`) page with a new Vue UI page, currently known as the `Structured Node List` page.

One final thing we've needed to do is to make sure links from other pages back to the node list page will work correctly with the new page. Those links can contains query string filter parameters to filter the node list.

We previously did work to handle some of these. This PR does the rest of the work to handle all of them. Some work in the Nodes v2 Rest API was also needed to handle some cases which were not fully handled by the FIQL -> Criteria processing.

Once this PR is merged, a 2nd PR will be done which actually relinks all existing links to the legacy node list page to the new Vue page, and will also remove the legacy page.

We handle the `category1/category2` filtering from the Surveillance View. The filter will be union within category sets, then intersection between category sets. I.e.:

`(category1Item || category1Item || category1Item) && (category2Item || category2Item || category2Item)`

The UI has been updated to allow an optional 2nd category. If needed we could add more category sets.

Added full support for monitored service filtering, including accepting different query params and adding to the UI filters.

Fixed various issues re case-insensitive filtering and various other things.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16118


christianpape (dismissed):
LGTM, but my Javascript/Typescript skills are pretty limited.

synqotik:
Converting to draft, fixing a few things and testing.

cgorantla (approved):
LGTM!

PR #8536, opened 2026-05-28
Docs: Document show-import as a vCenter connectivity check

marshallmassengill:
Add a "Test vCenter connectivity from the Karaf shell" section to the VMware handler page covering the opennms:show-import command.  Also a slight tweak to the wording to make it clear that they should CLONE the read-only role.


dino2gnt (approved):
looks fine to me.

PR #8539, opened 2026-06-01
NMS-19849: Add VmwareDatastoreCollector for vCenter datastore capacity

marshallmassengill:
Adding datastore monitoring: [NMS-19849](https://opennms.atlassian.net/browse/NMS-19849)

This includes:
* New vmwareDatastoreCapacity resource type (version-agnostic)
* Default vmware-datastore-capacity collection block with the attribute list the collector understands
* Service definition and collector mapping in the vmware6 and vmware7 packages of collectd-configuration.xml
* Combined datastore-capacity graph definition under snmp-graph.properties.d
* Compliance test parallel to VmwareCollectorComplianceTest
* Extractor unit tests covering null/zero edge cases
* Collect-loop unit tests exercising happy path with lab-modeled datastores, per-datastore summary error isolation, search failure, empty enumeration, and the no-usable-groups short circuit
* Reference documentation page and nav entry

Assisted-by: Claude Code:Opus 4.7/4.8

[NMS-19849]: https://opennms.atlassian.net/browse/NMS-19849?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opennms-base-assembly/src/main/filtered/etc/datacollection-config.xml
+++ b/opennms-base-assembly/src/main/filtered/etc/datacollection-config.xml
@@ -83,6 +83,7 @@
       <include-collection dataCollectionGroup="VMware-Cim"/>
       <include-collection dataCollectionGroup="VMware6"/>
       <include-collection dataCollectionGroup="VMware7"/>
+      <include-collection dataCollectionGroup="VMware-Datastore"/>
       <include-collection dataCollectionGroup="Zertico"/>
       <include-collection dataCollectionGroup="Zeus"/>
    </snmp-collection>
--- a/opennms-base-assembly/src/main/filtered/etc/datacollection-config.xml
+++ b/opennms-base-assembly/src/main/filtered/etc/datacollection-config.xml
@@ -83,7 +83,6 @@
       <include-collection dataCollectionGroup="VMware-Cim"/>
       <include-collection dataCollectionGroup="VMware6"/>
       <include-collection dataCollectionGroup="VMware7"/>
-      <include-collection dataCollectionGroup="VMware-Datastore"/>
       <include-collection dataCollectionGroup="Zertico"/>
       <include-collection dataCollectionGroup="Zeus"/>
    </snmp-collection>
--- a/docs/modules/reference/nav.adoc
+++ b/docs/modules/reference/nav.adoc
@@ -32,6 +32,7 @@
 ** xref:performance-data-collection/collectors/tca.adoc[]
 ** xref:performance-data-collection/collectors/vmware.adoc[]
 ** xref:performance-data-collection/collectors/vmware-cim.adoc[]
+** xref:performance-data-collection/collectors/vmware-datastore.adoc[]
 ** xref:performance-data-collection/collectors/wsman.adoc[]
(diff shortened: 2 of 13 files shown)


opennms-base-assembly/src/main/filtered/etc/datacollection-config.xml
```diff
@@ -83,6 +83,7 @@
       <include-collection dataCollectionGroup="VMware-Cim"/>
       <include-collection dataCollectionGroup="VMware6"/>
       <include-collection dataCollectionGroup="VMware7"/>
+      <include-collection dataCollectionGroup="VMware-Datastore"/>
```
dino2gnt: Why is this here?

dino2gnt (commented):
I only have one specific critique, but aside from that, this seems like a very specific use case for an entirely new detector.  

Not sure I like this on a conceptual level.  Collectors should be generally useful. I think I would like this better if it were an extension of the existing vmware collector.

 I'm not going to block it, though.

opennms-base-assembly/src/main/filtered/etc/datacollection-config.xml
marshallmassengill: So it's independent of the other two - I guess I could simplify and remove the vSphere 6 stuff since it's long since end of life and then combine this with the vsphere7 bits.  Going to ponder this while I respond to the other thing you raised.

marshallmassengill:
>Not sure I like this on a conceptual level. Collectors should be generally useful. I think I would like this better if it were an extension of the existing vmware collector.

So we actually get a bunch of performance data back for hosts/VMs today but one of the big missing piece is datastore information - it's kind of odd that we don't get it but it's a pretty big omission.  

One of the reasons to make this a separate collector is that it lets someone choose which objects they want to attach it to instead of duplicating the collection for multiple hosts (since they share the datastore).  Though maybe it makes sense to just let it duplicate for each host object... it's pretty minimal in terms of collection.

Personally what I think I'd love to do is refactor the entire VMware collectors/inventory into a plugin... that may actually resolve all of this.

I'm ok if this doesn't get merged but it's been missing for a while and driven me crazy.  One of the few things that it's very useful to threshold off of for VMware.

opennms-base-assembly/src/main/filtered/etc/datacollection-config.xml
dino2gnt: No, why is it in `datacollection-config.xml` at all?  it's not an SNMP collection.

If it's just packaging a `resourceType` this should go into `resource-types.d/`, not here.  If the collector doesn't know to look `resource-types.d/`, please fix that.

marshallmassengill: Should be good now.

marshallmassengill:
Did math... going to refactor this to use the existing VMware collector.  Not at all worth breaking it out like this.  I thought the API calls were worse than they are... they really are pretty small, even for a large VMware environment.

PR #8541, opened 2026-06-01
NMS-19849: Add Vmware Datastore Collections for vCenter datastore capacity

marshallmassengill:
Attempt number two to add VMware Datastore stats for collection for [NMS-19849](https://opennms.atlassian.net/browse/NMS-19849).  Refactored to keep this as part of the VMware collector instead of adding another collector for it.  It does mean that data is duplicated for each host but honestly, I think that's a worthy tradeoff for this to be simpler.

This includes:

* New vmwareDatastoreCapacity resource type (version-agnostic)
* Works with default Vmware collections (no new service needed)
* Combined datastore-capacity graph definition under snmp-graph.properties.d
* Extractor unit tests covering null/zero edge cases
* Reference documentation page and nav entry

Assisted-by: Claude Code:Opus 4.7/4.8

[NMS-19849]: https://opennms.atlassian.net/browse/NMS-19849?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/modules/reference/nav.adoc
+++ b/docs/modules/reference/nav.adoc
@@ -32,6 +32,7 @@
 ** xref:performance-data-collection/collectors/tca.adoc[]
 ** xref:performance-data-collection/collectors/vmware.adoc[]
 ** xref:performance-data-collection/collectors/vmware-cim.adoc[]
+** xref:performance-data-collection/collectors/vmware-datastore.adoc[]
 ** xref:performance-data-collection/collectors/wsman.adoc[]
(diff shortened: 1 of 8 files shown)


dino2gnt (commented):
I like this better.

marshallmassengill:
I think I do too... I think I had convinced myself that this was going to be this huge over-duplication of collected data but it really isn't.

cgorantla (approved):
LGTM!

PR #8547, opened 2026-06-04
NMS-18291: (retarget) add 64 gauges for net-snmp memory

marshallmassengill:
cherry picked from https://github.com/OpenNMS/opennms/pull/7924

Wanted to try and get this in.  Been out here for a while.  Seems like good changes.


opennms-base-assembly/src/main/filtered/etc/datacollection/netsnmp.xml
```diff
@@ -69,6 +69,17 @@
       <!-- Error flag.  1 indicates very little swap space left -->
       <mibObj oid=".1.3.6.1.4.1.2021.4.100" instance="0" alias="memSwapError" type="integer"/>
    </group>
+   <group name="ucd-memory-X" ifType="ignore">
```
cgorantla, line 72: As we shipped these into DB on `release-36.x`, existing installations won't see this.

I think we may need code changes to handle this type of upgrades.

marshallmassengill:
That's a fair point. I'll see about updating it.

features/config/upgrade/src/main/java/org/opennms/config/upgrade/datacollection/SnmpDataCollectionDefaultsUpdate.java
```diff
@@ -0,0 +1,274 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(SnmpDataCollectionDefaultsUpdate.class);
+
+    static final String UPDATE_ID_UCD_MEMORY_X = "NMS-18291-ucd-memory-X";
+
+    static final String NET_SNMP_SOURCE_NAME = "Net-SNMP";
+    static final String NET_SNMP_SYSTEMDEF_NAME = "Net-SNMP";
+    static final String UCD_MEMORY_GROUP_NAME = "ucd-memory";
+    static final String UCD_MEMORY_X_GROUP_NAME = "ucd-memory-X";
+    static final String UCD_MEMORY_X_RESOURCE = "/datacollection-defaults/nms-18291-ucd-memory-x.xml";
+
```
cgorantla: May be we should avoid hardcoding of these here. 
Just use one method to take care of one file, applyUcdMemoryX. Everything else should be common.

cgorantla (changes requested):
This is good. 

There is one more thing we need to do.  For us to support git based etc we need to know the changes we applied in DB. 

`features/config/upgrade/src/main/resources/datacollection-defaults/nms-18291-ucd-memory-x.xml` should be copied into archive `etc_archive/datacollection/updates` once we migrated them to DB.

If needed, may be add one more field to `snmp_collection_defaults_log`  as description to understand what got updated.

cgorantla (commented):
LGTM!

Need to run CI before merging as this  PR seems to be coming from fork.

PR #8550, opened 2026-06-05
NMS-18291: (retarget) add 64 gauges for net-snmp memory

marshallmassengill:
Once more with smoke testing...

cherry picked from https://github.com/OpenNMS/opennms/pull/7924

Wanted to try and get this in. Been out here for a while. Seems like good changes.


dino2gnt (approved):
cool

marshallmassengill:
@cgorantla - looking at this again today and I think this is a decent solution but want your input on it.  The CI failure is related to something Scott was working on and I'm going to pull in the latest and it should pass.

SnmpDataCollectionDefaultsUpdate handles post-migration deltas (like NMS-18291's ucd-memory-X gauges) to existing databases additively, once each. It runs after the XML→DB migration and tracks applied deltas in a dedicated snmp_collection_defaults_log ledger (keyed by id, not content).  This gets around the fact that Liquibase's DATABASECHANGELOG would run too early. Updates only insert missing groups or append missing systemDef references; they never overwrite admin-owned rows. 

To add a future delta: drop a declarative fragment XML under datacollection-defaults/ and register one DefaultConfigUpdate(id, resource, description) entry.

cgorantla (approved):
LGTM!

PR #8555, opened 2026-06-10
NMS-19836: Move thresh-configuration.xml and thresholds.xml to DB

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-19836

### DIFF
--- a/opennms-config-dao/thresholding/impl/src/main/java/org/opennms/netmgt/config/dao/thresholding/impl/OnmsThreshdDao.java
+++ b/opennms-config-dao/thresholding/impl/src/main/java/org/opennms/netmgt/config/dao/thresholding/impl/OnmsThreshdDao.java
@@ -21,93 +21,134 @@
  */
 package org.opennms.netmgt.config.dao.thresholding.impl;
 
-import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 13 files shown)


opennms-config-dao/thresholding/impl/src/main/java/org/opennms/netmgt/config/dao/thresholding/impl/OnmsThreshdDao.java
```diff
@@ -21,116 +21,150 @@
...
+            synchronized (this) {
+                if (cms == null) {
+                    cms = BeanUtils.getBean("daoContext", "configurationManagerService", ConfigurationManagerService.class);
+                    cms.registerEventHandler(EventType.UPDATE,
+                            new ConfigUpdateInfo(CONFIG_NAME, ConfigDefinition.DEFAULT_CONFIG_ID),
+                            info -> onConfigChanged());
+                }
+            }
+        }
+        return cms;
```
cgorantla, line 73: Instead of using CM directly, should we explore creating class that exposes AbstractCmJaxbConfigDao and using that here for both `thresholds.xml` and `thresh-configuration.xml`

cgorantla (approved):
LGTM. 
May be an end to end test can be added after we incorporate Rest API.

PR #8561, opened 2026-06-11
NMS-19786: Exclude outdated jackson version from hawtio

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-19786


cgorantla (approved):
LGTM!

PR #8562, opened 2026-06-11
NMS-19874: Topology views: persistence + `/api/v2/topology/views` REST API

marshallmassengill:
This adds the persistence and CRUD API for the new topology UI's custom
(user-composed) views: a shared catalog of named view documents.

- New `features/topology-views` module: `TopologyView` document model and a DAO
  backed by the JSON key-value store (`kvstore_jsonb`, context
  `topology-views`). The canvas itself is an opaque JSON `definition` document
  to the backend, so front-end model evolution (new element kinds) needs no
  backend change. Catalog metadata (unique name, owner) is stored alongside.
- New `/api/v2/topology/views[/{id}]` resource in `opennms-webapp-rest`
  (`TopologyViewRestService` + `TopologyViewDTO`): lightweight JSON CRUD, the
  `definition` exposed as nested JSON (no double-escaping), owner stamped from
  the security principal, 409 on duplicate name, 400/404 on bad input/missing
  id. A default view is seeded on first start (keyed by id, so a renamed
  default is not recreated).
- Access control is the standard `/api/v2/**` RBAC: any authenticated user
  reads, ROLE_REST/ROLE_ADMIN writes. No new security rules.
- The module reaches the webapp classpath via `opennms-base-assembly`, the same
  path as other feature DAOs; no webapp XML changes.

Testing: `TopologyViewDaoIT` (4) against a temporary database and
`TopologyViewRestServiceIT` (4) covering the CRUD lifecycle, JSON round-trip,
owner stamping, error statuses, and RBAC. Also verified live (CRUD, anonymous
→ 401, ROLE_REST-only user can write).

Consumed by the topology UI PR, but useful standalone to any REST client.

Jira: https://opennms.atlassian.net/browse/NMS-19874


marshallmassengill:
Folded the other one into here @cgorantla.  Just FYI.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/TopologyViewRestService.java
```diff
@@ -0,0 +1,219 @@
...
+        return toDto(require(id));
+    }
+
+    @POST
+    public Response create(@Context final UriInfo uriInfo, @Context final SecurityContext securityContext, final TopologyViewDTO dto) {
+        if (dto == null || dto.getName() == null || dto.getName().trim().isEmpty()) {
+            throw webException(Response.Status.BAD_REQUEST, "A view name is required");
+        }
+        // Normalize once; validation, collision check, persistence and error
+        // messages all use the trimmed name so "Core" and "Core " can't coexist.
```
cgorantla: nit: can remove this comment. Mostly self explanatory

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/TopologyViewRestService.java
```diff
@@ -0,0 +1,219 @@
...
+        }
+
+        final TopologyView view = new TopologyView();
+        view.setName(name);
+        view.setDefinition(writeDefinition(dto.getDefinition()));
+        view.setOwner(ownerOf(securityContext));
+        view.setCreated(new Date());
+
+        final String id = getDao().save(view);
+        return Response.created(uriInfo.getAbsolutePathBuilder().path(String.valueOf(id)).build()).build();
```
cgorantla: Although it is rare, it is possible to create duplicate view with the same name as there is no protection on DB side with unique key.   We might want to wrap get/create  into synchronized block with custom object key

features/topology-views/src/main/java/org/opennms/netmgt/topology/views/impl/TopologyViewJsonStore.java
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    /** Wire as the bean {@code init-method}: seed the Default view if absent. */
+    public void init() {
+        try {
+            // Key off the stable id, not the name: a user can rename the Default
+            // view (its key stays DEFAULT_VIEW_ID), and re-seeding by name would
+            // then overwrite that renamed view with a fresh empty Default. We
+            // only re-seed if the baseline entry was actually deleted.
+            if (get(DEFAULT_VIEW_ID) == null) {
```
cgorantla:
```suggestion
            if (get(DEFAULT_VIEW_ID) == null && findByName(DEFAULT_VIEW_NAME) == null)) {
```
This can occur when the user deletes default view and creates a new view with name `default` and restarts OpenNMS.  Very rare but possible rare scenario.

cgorantla (approved):
LGTM!

PR #8563, opened 2026-06-11
NMS-19875: Topology assets: image storage for view backgrounds and custom node icons

marshallmassengill:
Custom topology views need server-side image storage for two planned
features: background images (floor plans, rack diagrams) behind the canvas,
and operator-uploaded node icons. Both reduce to the same problem — binary
image bytes referenced by id from a view's definition document — so one
shared asset catalog serves both, distinguished by a `kind`
(`background` | `icon`).

- New `features/topology-assets` module: image bytes in the generic binary
  key-value store (`PostgresBlobStore`, `kvstore_bytea`) and catalog metadata
  (name, kind, MIME type, size, owner, timestamps) as a JSON document in the
  JSON key-value store, keyed by the same generated UUID — listings never
  drag image payloads along. No new schema. The blob store is constructed
  internally from the DataSource: the only BlobStore exposed as a shared
  service in the core context is thresholding's no-op, and registering a
  second one would make that lookup ambiguous.
- `/api/v2/topology/assets` (`TopologyAssetRestService`): POST raw image
  bytes (no multipart; name/kind as query parameters), GET list with
  optional `?kind=` filter, GET `{id}` serving the bytes under the asset's
  own content type with an ETag + `max-age` (icons reused across many nodes
  revalidate as 304s), GET `{id}/meta`, DELETE. Raster types only
  (png/jpeg/gif/webp); SVG is excluded for now since it can carry active
  content and assets are served from the application origin. Per-kind size
  caps: icons 512 KiB, backgrounds 10 MiB.
- Access is the standard `/api/v2` RBAC: any authenticated user reads,
  ROLE_REST/ROLE_ADMIN writes.
- The base assembly now ships the module and the Postgres BlobStore
  implementation (only the no-op blob store was in `lib` before).

Testing: `TopologyAssetDaoIT` (3) against a temporary database and
`TopologyAssetRestServiceIT` (3) covering the upload/serve/delete lifecycle,
byte integrity, ETag revalidation, kind filtering, and the validation
statuses (400/413/415/404). Also verified live (upload → byte-identical
serve → 304 → delete; anonymous → 401).

The consuming UI (background rendering and the node icon picker) ships in
the topology UI PR; this PR is the storage/API half and is independently
mergeable (the UI degrades gracefully without it).

Jira: https://opennms.atlassian.net/browse/NMS-19875


cgorantla (commented):
I still need to review all PRs thoroughly but I think this can be combined with https://github.com/OpenNMS/opennms/pull/8562.  No need to expose separate topology-assets maven module. It can be combined with topology-views.

marshallmassengill:
I've updated https://github.com/OpenNMS/opennms/pull/8562 to handle this.

PR #8564, opened 2026-06-11
NMS-19876: Topology (Preview) REST API for operator info-panel templates (node and edge scoped)

marshallmassengill:
Carries the etc/infopanel/ Jinjava template investment over to the new Vue topology map without the Vaadin dependency. A new read-only resource renders the operator's templates server-side and returns [{title, order, html}] for the UI's Inspector:

  - GET /api/v2/topology/infopanel?nodeId={id} — node-scoped panels. Context matches the legacy GenericInfoPanelItemProvider (node, nodeResource, read-only nodeDao/resourceDao wrappers, measurements), so existing node-scoped templates render unchanged.
  - GET /api/v2/topology/infopanel/edge?sourceNodeId=&targetNodeId=[&sourcePort=&targetPort=&protocol=] — edge-scoped panels. Templates receive an edge variable mirroring the legacy LinkdEdge surface (edge.discoveredBy, edge.sourcePort/targetPort with the node, port name, and the port's resolved OnmsSnmpInterface/ifIndex, matched by ifName then ifDescr). Legacy edge templates need one rename: edge.sourcePort.vertex →edge.sourcePort.node. The interface resolution is deliberate groundwork for planned link status/metrics work, which will key on the same identity.

Implementation: a Vaadin-free InfoPanelRenderer (jinjava + transitive deps packaged into the webapp); templates re-read per request so operators iterate without restarts; template errors skip that template rather than fail the request; missing template directory returns an empty list. Covered by renderer unit tests (node and edge contexts, visibility/order contract) and a REST integration test.

Jira: https://opennms.atlassian.net/browse/NMS-19876


opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/infopanel/MeasurementsWrapper.java
```diff
@@ -0,0 +1,171 @@
...
+        final QueryResponse.WrappedPrimitive[] columns = measurementsService.query(request).getColumns();
+        final double[] valuesIn = columns[0].getList();
+        final double[] valuesOut = columns[1].getList();
+
+        for (int i = valuesIn.length - 1; i >= 0; i--) {
+            if (!Double.isNaN(valuesIn[i]) && !Double.isNaN(valuesOut[i])) {
+                return Arrays.asList(valuesIn[i], valuesOut[i]);
+            }
+        }
+        return Arrays.asList(Double.NaN, Double.NaN);
```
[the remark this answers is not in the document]
marshallmassengill: Oddly enough, this is almost a copy of the original and I wonder if it's broken now...

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/infopanel/MeasurementsWrapper.java
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    /** Last value for a resource/attribute with the given aggregation. */
+    public double getLastValue(final String resource, final String attribute, final String aggregation) throws MeasurementException {
+        return getLastValue(resource, attribute, aggregation, true);
+    }
+
+    public double getLastValue(final String resource, final String attribute, final String aggregation, final boolean relaxed) throws MeasurementException {
+        final long end = System.currentTimeMillis();
+        final long start = end - (15 * 60 * 1000);
```
synqotik: Make this a constant, and have a note that it represents a 15 minute window. Maybe `WINDOW_AMOUNT_MS`, or a better name.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/infopanel/MeasurementsWrapper.java
```diff
@@ -0,0 +1,215 @@
...
+
+        final QueryResponse.WrappedPrimitive[] columns = queryInt(resource, attribute, start, end, 300000, aggregation, relaxed).getColumns();
```
synqotik: Make 300000 a constant, explain what time it represents.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/infopanel/MeasurementsWrapper.java
```diff
@@ -0,0 +1,215 @@
...
+        return Collections.emptyList();
+    }
+
+    /**
+     * Compute the in/out percentage utilization of a node interface (by ifName),
+     * using HC octet attributes with non-HC fallbacks.
+     */
+    public List<Double> computeUtilization(final OnmsNode node, final String ifName) throws MeasurementException {
+        final long end = System.currentTimeMillis();
+        final long start = end - (15 * 60 * 1000);
```
synqotik: Same as above.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/infopanel/MeasurementsWrapper.java
```diff
@@ -0,0 +1,215 @@
...
+
+        for (final OnmsSnmpInterface snmpInterface : node.getSnmpInterfaces()) {
+            if (ifName.equals(snmpInterface.getIfName())) {
+                final String resourceId = "node[" + node.getId() + "].interfaceSnmp[" + snmpInterface.computeLabelForRRD() + "]";
```
synqotik, line 106: Same as above (constant for 300000(.

synqotik (changes requested):
Overall looks good, just a few small things to address.

synqotik (approved):
LGTM!

cgorantla (commented):
As it is , this seems to be fine.

But we need to really think if we need to carry jinjava templates with our new UI.  Are operators really investing in these templates ?

marshallmassengill:
> As it is , this seems to be fine.
> 
> But we need to really think if we need to carry jinjava templates with our new UI. Are operators really investing in these templates ?

It's being used by a few customers.  We can discuss.  I'm open to a replacement in the long term but would suggest we keep this for now until we put our heads together about a suitable replacement.  I'm not opposed to saying jinja/jinjava should be replaced with some other DSL.

PR #8565, opened 2026-06-11
NMS-19877: Topology (Preview) graph-service provider for the path outage hierarchy

marshallmassengill:
Exposes the node-parent / critical-path hierarchy (nodeParentID) as a native graph-service provider — GET /api/v2/graphs/pathoutage/pathoutage — making the legacy Path Outage topology available to the new Vue map and any other Graph Service consumer, without bridging the Vaadin-era provider.

The provider (new features/graph/provider/pathoutage module) builds the parent→child forest from NodeDao inside a read-only transaction, since the parent association is lazy and the REST call path supplies no session.

Vertices carry label and nodeID so standard node-info/status enrichment applies; nodes with neither parent nor children are omitted to keep the graph focused on the configured hierarchy. Registers with cacheInvalidateInterval=300 so parent changes appear within five minutes (without it, the graph service caches the first load indefinitely). Covered by unit tests and an integration test exercising the lazy parent under real Hibernate with no caller transaction — the exact REST-path condition.

Jira: https://opennms.atlassian.net/browse/NMS-19877


features/graph/provider/pathoutage/src/main/java/org/opennms/netmgt/graph/provider/pathoutage/PathOutageGraphProvider.java
```diff
@@ -0,0 +1,146 @@
...
+        return sessionUtils.withReadOnlyTransaction(() -> {
+            final GenericGraph.GenericGraphBuilder builder = GenericGraph.builder()
+                    .graphInfo(getGraphInfo())
+                    .id(NAMESPACE)
+                    // Resolve node + alarm-based status for the vertices, like the other
+                    // node-backed providers; the SPA also colors by nodeID independently.
+                    .property(GenericProperties.Enrichment.RESOLVE_NODES, true)
+                    .property(GenericProperties.Enrichment.DEFAULT_STATUS, true);
+
+            final List<OnmsNode> nodes = nodeDao.findAll();
```
synqotik: I think we could limit this to get just nodes with parents, could greatly reduce the amount of data in memory and in the loops below.

synqotik (changes requested):

synqotik (dismissed):
LGTM!

features/graph/provider/pathoutage/src/main/java/org/opennms/netmgt/graph/provider/pathoutage/PathOutageGraphProvider.java
```diff
@@ -0,0 +1,151 @@
...
+            final List<OnmsNode> nodes = nodeDao.findMatching(
+                    new CriteriaBuilder(OnmsNode.class).isNotNull("parent").toCriteria());
+
+            // First pass: collect the participating node ids (the children above
+            // plus their parents) and their labels, and remember the parent->child pairs.
+            final Map<Integer, String> labelById = Maps.newHashMap();
+            final Set<Integer> participating = new LinkedHashSet<>();
+            // child -> parent: value-based (unlike a Set of arrays), and a child
+            // has exactly one parent, so duplicates are structurally impossible.
+            final Map<Integer, Integer> parentByChild = new LinkedHashMap<>();
```
cgorantla, line 105:
```suggestion
            final Map<Integer, Integer> parentByChild = new LinkedHashMap<>();
            final Map<Integer, String> labels = nodeDao.getAllLabelsById();
```

features/graph/provider/pathoutage/src/main/java/org/opennms/netmgt/graph/provider/pathoutage/PathOutageGraphProvider.java
```diff
@@ -0,0 +1,151 @@
...
+            for (final OnmsNode node : nodes) {
+                final OnmsNode parent = node.getParent();
+                if (parent == null) {
+                    continue;
+                }
```
cgorantla:
```suggestion
    final Integer parentId = node.getNodeParentId();
    final String parentLabel = labels.get(parentId);
    if (parentLabel == null) {
        // dangling nodeParentID (parent row deleted) — skip this relationship
        continue;
    }
```

node.getParent() will return proxy and parent.getLabel may fail in scenarios where parent gets deleted see https://opennms.atlassian.net/browse/NMS-19971

cgorantla (commented):
Couple of suggestions that may be needed before we fix https://opennms.atlassian.net/browse/NMS-19971

features/graph/provider/pathoutage/src/main/java/org/opennms/netmgt/graph/provider/pathoutage/PathOutageGraphProvider.java
marshallmassengill: Should be fixed now.  I'll see if I can look at NMS-19971.

cgorantla (approved):
LGTM!

PR #8566, opened 2026-06-11
NMS-19879: Remove discontinued activemq-camel dependency

marshallmassengill:
Apache discontinued the activemq-camel component after the 5.16.x line, so it was pinned at 5.16.8 while the rest of ActiveMQ moved to 5.19.x. Although NMS-19698 pinned the broker/client/pool transitives to  ${activemqVersion}, the activemq-camel-5.16.8.jar and activemq-spring-5.16.8.jar artifacts still shipped in $OPENNMS_HOME/lib, and customer security scanners flag them (e.g. CVE-2026-34197 against ActiveMQ < 5.19.4).

 ActiveMQComponent was only ever used as a plain Camel JmsComponent: every instantiation injects a camel-jms JmsConfiguration with an externally defined connection factory, so none of its ActiveMQ-specific behavior was in play. This change swaps the bean class in the four wiring XMLs, declares camel-jms and activemq-spring (which parses etc/opennms-activemq.xml for the embedded broker and previously arrived transitively via activemq-camel) as direct, version-managed dependencies, and drops the 5.16.8 bundle lines from the Karaf features. Bean ids and the opennms.broker/onms.broker service aliases are unchanged.

With this change the assembled lib/ contains no 5.16.8 jars — all ActiveMQ artifacts follow ${activemqVersion}.

I ran a smoke test with a local minion and it all seems to check out.  It's probably worth more testing if anyone has a cycle for it.

Assisted-by: Claude Code: Opus 4.8/Fable 5

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19879


christianpape (approved):
LGTM.

cgorantla (approved):
LGTM!

PR #8570, opened 2026-06-12
NMS-19883: Speed up the Vue UI: cache and compress static assets, fix OpenAPI page load

marshallmassengill:
Vue UI bundles (/opennms/ui/assets/, /opennms/ui-components/assets/) were covered by the NMS-14947 no-store rule and served uncompressed, so browsers re-downloaded several MB on every page view — including legacy JSP pages, which load the menu bundle. The OpenAPI page also fetched its two specs serially and re-rendered everything on each visit.

Changes:
  - jetty.xml: give the two static asset paths Cache-Control: no-cache (cache + revalidate; names aren't content-hashed, so no max-age) and gzip them via a path-scoped GzipHandler (already in jetty-server, no new dependency). Everything else keeps no-store. 
  - OpenAPI.vue: fetch specs in parallel, cache them for the SPA session, defer the below-fold v1 doc, and strip the "example": null entries that crashed RapiDoc (~1,100 uncaught TypeErrors per view, now 0).
  - WebappIT: smoke test asserting the new cache headers.
  
Result: OpenAPI page ~7 s → ~2 s first visit, ~0.3 s revisits; repeat visits to any Vue UI page are 304s instead of full downloads.

Jira: https://opennms.atlassian.net/browse/NMS-19883

### DIFF
--- a/opennms-jetty/src/main/resources/org/opennms/netmgt/jetty/jetty.xml
+++ b/opennms-jetty/src/main/resources/org/opennms/netmgt/jetty/jetty.xml
@@ -129,18 +129,30 @@
           </New>
         </Item>
         <!-- Jira NMS-14947 Cacheable HTTPS Responses - Cache Control Directive Missing or Misconfigured -->
-        <!-- no caching for everything in "/opennms/*" excluding static assets "/opennms/assets/*"  -->
+        <!-- no caching for everything in "/opennms/*" excluding the static assets of the legacy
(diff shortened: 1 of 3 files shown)


synqotik (approved):
LGTM!

OpenAPI is so much faster now.

It looks like `Cache-control: no-cache` should handle what we are doing now in Vue, where we are currently not content-hashing etc. We can adjust in the future as needed.

PR #8572, opened 2026-06-12
NMS-19882: guard HeartbeatConsumer against misconfigured Minions

dino2gnt:
HeartbeatConsumer will attempt to provision minions with missing IDs, fouling up the Minions requisition and causing it to be re-created empty, which then forces all the minions to be re-provisioned until the invalid minion(s) send its next heartbeat.  Repeat.

Check for and bail out on null / empty foreignId from the heartbeat message, and validate the node object before adding it to the req.

This one was _not_ assisted by Claude, believe it or not.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19882
* SUPPORT-3239

### DIFF
--- a/features/minion/heartbeat/consumer/src/main/java/org/opennms/minion/heartbeat/consumer/HeartbeatConsumer.java
+++ b/features/minion/heartbeat/consumer/src/main/java/org/opennms/minion/heartbeat/consumer/HeartbeatConsumer.java
@@ -129,6 +129,11 @@ public class HeartbeatConsumer implements MessageConsumer<MinionIdentityDTO, Min
         LOG.info("Received heartbeat for Minion with id: {} at location: {}",
                 minionHandle.getId(), minionHandle.getLocation());
 
+        if (minionHandle.getId() == null || minionHandle.getId().isEmpty()) {
+            LOG.warn("Refusing to process Minion heartbeat with null 'id' for location: '{}'", minionHandle.getLocation());
+            return;
+        }
+
         OnmsMinion minion = minionDao.findById(minionHandle.getId());
         if (minion == null) {
             minion = new OnmsMinion();
@@ -235,10 +240,6 @@ public class HeartbeatConsumer implements MessageConsumer<MinionIdentityDTO, Min
 
         // Return if minion with this foreignId and location already exists.
         String foreignId = minion.getLabel() != null ? minion.getLabel() : minion.getId();
-        if (foreignId.isEmpty()) {
-            LOG.warn("Heartbeat received with null ID and label for location '{}', aborting this heartbeat!", minion.getLocation());
-            return;
-        }
         List<OnmsNode> nodes = nodeDao.findByForeignIdForLocation(foreignId, nextLocation);
         if (!nodes.isEmpty()) {
             //check for existing requisitions the policy and detectors are in place
--- a/features/minion/heartbeat/consumer/src/main/java/org/opennms/minion/heartbeat/consumer/HeartbeatConsumer.java
+++ b/features/minion/heartbeat/consumer/src/main/java/org/opennms/minion/heartbeat/consumer/HeartbeatConsumer.java
@@ -129,7 +129,7 @@ public class HeartbeatConsumer implements MessageConsumer<MinionIdentityDTO, Min
         LOG.info("Received heartbeat for Minion with id: {} at location: {}",
                 minionHandle.getId(), minionHandle.getLocation());
 
-        if (minionHandle.getId() == null || minionHandle.getId().isEmpty()) {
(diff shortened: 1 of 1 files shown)


marshallmassengill:
Wild.  Will review on Monday.  I could see how this combined with the other issue we were potentially seeing could even be a double whammy... assuming the other one is still legit.

features/minion/heartbeat/consumer/src/main/java/org/opennms/minion/heartbeat/consumer/HeartbeatConsumer.java
```diff
@@ -232,6 +235,10 @@ private void provision(final OnmsMinion minion,
 
         // Return if minion with this foreignId and location already exists.
         String foreignId = minion.getLabel() != null ? minion.getLabel() : minion.getId();
+        if (foreignId.isEmpty()) {
```
cgorantla:
```suggestion
        if (foreignId == null || foreignId.isEmpty()) {
```

features/minion/heartbeat/consumer/src/main/java/org/opennms/minion/heartbeat/consumer/HeartbeatConsumer.java
cgorantla, line 131: Should probably evaluate minionId here for null or empty otherwise a Minion with empty Id would be persisted.

dino2gnt: Good idea.

features/minion/heartbeat/consumer/src/main/java/org/opennms/minion/heartbeat/consumer/HeartbeatConsumer.java
```diff
@@ -127,6 +129,11 @@ public void handleMessage(MinionIdentityDTO minionHandle) {
         LOG.info("Received heartbeat for Minion with id: {} at location: {}",
                 minionHandle.getId(), minionHandle.getLocation());
 
+        if (minionHandle.getId() == null || minionHandle.getId().isEmpty()) {
+            LOG.warn("Refusing to process Minion heartbeat with null 'id' for location: '{}'", minionHandle.getLocation());
```
cgorantla:
```suggestion
            LOG.warn("Refusing to process Minion heartbeat with invalid 'id' for location: '{}'", minionHandle.getLocation());
```

features/minion/heartbeat/consumer/src/main/java/org/opennms/minion/heartbeat/consumer/HeartbeatConsumer.java
```diff
@@ -127,6 +129,11 @@ public void handleMessage(MinionIdentityDTO minionHandle) {
         LOG.info("Received heartbeat for Minion with id: {} at location: {}",
                 minionHandle.getId(), minionHandle.getLocation());
 
+        if (minionHandle.getId() == null || minionHandle.getId().isEmpty()) {
```
cgorantla: As this is directly coming from user input, we should also handle blank here.  
```suggestion
        if (minionHandle.getId() == null || minionHandle.getId().isBlank()) {
```

features/minion/heartbeat/consumer/src/main/java/org/opennms/minion/heartbeat/consumer/HeartbeatConsumer.java
```diff
@@ -172,9 +179,13 @@ public void handleMessage(MinionIdentityDTO minionHandle) {
...
 
-            this.provision(onmsMinion,
-                    prevLocation,
-                    nextLocation);
+            try {
+                this.provision(onmsMinion,
+                        prevLocation,
+                        nextLocation) ;
+            } catch (ValidationException e) {
+                throw new RuntimeException("Heartbeat provisioner failed to validate: ", e);
```
cgorantla: LOG an error and return instead of throwing RuntimeException.

features/minion/heartbeat/consumer/src/main/java/org/opennms/minion/heartbeat/consumer/HeartbeatConsumer.java
dino2gnt: Do we need both?  Can it be empty without being null, or blank without being null?

cgorantla: blank should cover empty and black.

cgorantla (dismissed):
LGTM!

cgorantla (approved):
LGTM!

PR #8574, opened 2026-06-15
Docs: Expand provision policy docs with examples and updates

marshallmassengill:
Added some examples (thanks @JSimo!) and attempted to bring all of these up to date and get wording more reasonable.

I did a lot of tests to validate the examples including digging into the code where I changed the specific wording that was there.  If I missed something or something doesn't look right then let me know and I'll get it corrected.  

There were a few changes in here that surprised me a bit and I think were either in there for historical reasons or they just have never been updated after some code changed.  Hopefully correct now but I'd love feedback.

Assisted-by: Claude Code: Opus 4.7/4.8


indigo423 (approved):
great improvement

PR #8575, opened 2026-06-15
NMS-19901: Fix startup deadlock in SnmpPeerFactory scope init

JSimo:
We were experiencing deadlocks during startup, sometimes. As we needed to restart a few times to get OpenNMS to launch. This commit targets fix to this deadlock.

Replace class-level synchronization in secure credentials scope lazy init with a dedicated lock so SNMP interface poller threads do not hold SnmpPeerFactory.class while resolving Spring beans, preventing lock-order inversion with ContextRegistry during startup. Likely due our usage of snmp interface poller, being less common across OpenNMS versions.

### All Contributors

### Contribution Checklist

### What's Next?

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19901

### DIFF
--- a/opennms-config/src/main/java/org/opennms/netmgt/config/SnmpPeerFactory.java
+++ b/opennms-config/src/main/java/org/opennms/netmgt/config/SnmpPeerFactory.java
@@ -189,17 +189,27 @@ public class SnmpPeerFactory implements SnmpAgentConfigFactory {
     }
 
     private static void encryptSingletonIfEnabled() {
-        final SnmpPeerFactory singleton = s_singleton;
-        if (singleton != null && singleton.encryptionEnabled) {
-            singleton.encryptSnmpConfig();
+        final SnmpPeerFactory singleton;
+        synchronized (SnmpPeerFactory.class) {
+            singleton = s_singleton;
+            if (singleton == null || !singleton.encryptionEnabled) {
+                return;
+            }
         }
+        singleton.encryptSnmpConfig();
     }
 
     private void encryptSnmpConfig() {
         initializeTextEncryptor();
-        if (textEncryptor != null) {
+        if (textEncryptor == null) {
+            return;
+        }
+        synchronized (SnmpPeerFactory.class) {
+            if (this != s_singleton) {
+                return;
+            }
             try {
-                s_singleton.saveCurrent();
+                saveCurrent();
             } catch (IOException e) {
                 LOG.debug("Exception while saving encrypted credentials");
             }
--- a/opennms-config/src/main/java/org/opennms/netmgt/config/SnmpPeerFactory.java
+++ b/opennms-config/src/main/java/org/opennms/netmgt/config/SnmpPeerFactory.java
@@ -99,7 +99,7 @@ public class SnmpPeerFactory implements SnmpAgentConfigFactory {
     /**
      * The singleton instance of this factory
      */
-    private static SnmpPeerFactory s_singleton = null;
+    private static volatile SnmpPeerFactory s_singleton = null;
 
     /**
      * This member is set to true if the configuration file has been loaded.
@@ -128,8 +128,9 @@ public class SnmpPeerFactory implements SnmpAgentConfigFactory {
      * {@code synchronized (SnmpPeerFactory.class)} when calling {@link BeanUtils#getBean}:
      * bean lookup can take {@code ContextRegistry}'s lock while Spring still holds that lock
      * and invokes {@link #init()}, which needs the class monitor (deadlock with SNMP
-     * interface poller). The same applies to {@code textEncryptor} resolution in
-     * {@link #encryptSingletonIfEnabled()}, which must run outside the class monitor.
+     * interface poller). The same applies to {@link #initializeTextEncryptor()} in
+     * {@link #encryptSnmpConfig()}; encrypt save is re-synchronized on the class monitor
+     * after the bean is resolved.
      */
     private static final Object secureCredentialsScopeInitLock = new Object();
 
--- a/opennms-config/src/main/java/org/opennms/netmgt/config/SnmpPeerFactory.java
+++ b/opennms-config/src/main/java/org/opennms/netmgt/config/SnmpPeerFactory.java
@@ -121,19 +121,17 @@ public class SnmpPeerFactory implements SnmpAgentConfigFactory {
 
     private final Boolean encryptionEnabled = Boolean.getBoolean(ENCRYPTION_ENABLED);
 
-    private static Scope secureCredentialsVaultScope;
+    private static volatile Scope secureCredentialsVaultScope;
 
     /**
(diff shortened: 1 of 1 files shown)


marshallmassengill (changes requested):
It looks like getSnmpConfigDao() may have the same issue so I think we would want to fix it as well.

There may be other issues in here as well.  I'll let others comment on approval.  The idea is sound and we appreciate the work to diagnose this.

cgorantla (changes requested):
I fixed one scenario here in this PR https://github.com/OpenNMS/opennms/pull/7885/changes 

Probably that exposed this one.

We need to re-target this to `foundation-2024`.

cgorantla:
> It looks like getSnmpConfigDao() may have the same issue so I think we would want to fix it as well.
> 
> There may be other issues in here as well. I'll let others comment on approval. The idea is sound and we appreciate the work to diagnose this.

This one is only relevant on `release-36.x`.  So if @JSimo re-target this to `foundation-2024` we are good.  

I can open another issue for  fixing `snmpConfigDao` stuff on `release-36.x`

cgorantla:
> I fixed one scenario here in this PR https://github.com/OpenNMS/opennms/pull/7885/changes
> 
> Probably that exposed this one.
> 
> We need to re-target this to `foundation-2024`.

Thanks for your contribution @JSimo.  Let me know if you can re-target, otherwise I can take care of it after merging this.

JSimo:
> > I fixed one scenario here in this PR https://github.com/OpenNMS/opennms/pull/7885/changes
> > Probably that exposed this one.
> > We need to re-target this to `foundation-2024`.
> 
> Thanks for your contribution @JSimo. Let me know if you can re-target, otherwise I can take care of it after merging this.

Re-targeted to `foundation-2024`.

opennms-config/src/main/java/org/opennms/netmgt/config/SnmpPeerFactory.java
```diff
@@ -124,6 +124,18 @@ public class SnmpPeerFactory implements SnmpAgentConfigFactory {
     private static Scope secureCredentialsVaultScope;
```
indigo423: **P1 (must-fix) — broken double-checked locking: this field needs `volatile`.** `High`

The change converts `getSecureCredentialsScope()` from `static synchronized` to double-checked locking with a lock-free fast-path read (`Scope scope = secureCredentialsVaultScope; if (scope != null) return scope;`). But `secureCredentialsVaultScope` is a plain (non-`volatile`) static, which makes this textbook *broken* DCL: the write inside the `synchronized` block has no happens-before edge to the unsynchronized fast-path read, so another thread can observe a non-null reference to an unsafely-published `Scope` (or never see the write at all). The old `static synchronized` getter was correct precisely because every read held the monitor — this rewrite removes that guarantee. The unsynchronized `setSecureCredentialsVaultScope(...)` setter races the same way. Declaring the field `volatile` closes both:

```suggestion
    private static volatile Scope secureCredentialsVaultScope;
```

indigo423 (commented):
The deadlock diagnosis and the move off the `SnmpPeerFactory.class` monitor are sound — these are correctness gaps on top of that. Apply or dismiss per your call.

**Inline suggestion (1):**
- **P1** `High` — broken double-checked locking; `secureCredentialsVaultScope` must be `volatile`. One-line fix posted inline.

**Important — the deadlock is only partially fixed (no inline suggestion; lands outside this diff):**
- **D1** `High` — `init()` (`SnmpPeerFactory.java:175`) is `static synchronized` and, when `encryptionEnabled=true`, calls `BeanUtils.getBean("daoContext", "textEncryptor", ...)` via `encryptSnmpConfig()` → `initializeTextEncryptor()` (line ~795) **while holding the `SnmpPeerFactory.class` monitor**. That is the identical `class-monitor → Spring ContextRegistry` lock ordering this PR's own new Javadoc identifies as the deadlock cause. This PR removed the class monitor only from the `getSecureCredentialsScope()` path; the `init() → getBean(textEncryptor)` path retains the same inversion. With SNMP config encryption enabled, the startup deadlock can therefore still occur. Suggested direction: resolve the `textEncryptor` bean outside the `init()` monitor, mirroring what was done for the scope. (Left as a comment rather than a suggestion since it restructures code this PR didn't touch — your call on scope.)

**Minor (your call):**
- **D2** `Low` (pre-existing) — on `BeanUtils.getBean` returning null / `FatalBeanException`, the field stays null and every subsequent call re-runs bean resolution + re-logs the WARN. Identical to prior behavior; noted because this is a hot poller path.
- **D3** `Low` (question) — the new public no-arg `SnmpPeerFactory()` constructor is unrelated to the locking fix. All Spring beans wire via `factory-method="getInstance"`, so it changes no instantiation semantics and isn't needed for the fix — was it intentional, or leftover? Consider dropping it to keep the change focused.

**Cleared during review (not findings):** removing `static synchronized` from this one method does not weaken any invariant other synchronized methods relied on — no other synchronized member touches `secureCredentialsVaultScope`, and `m_config` is guarded by the separate `m_lock` ReadWriteLock.

opennms-config/src/main/java/org/opennms/netmgt/config/SnmpPeerFactory.java
JSimo: Updated according to PR comments. Will take this back and use the updated commit internally in our test instance to validate.

cgorantla: volatile change seems fine.  Although it is not strictly required here.  Another thread may enter   `synchronized (secureCredentialsScopeInitLock) {` block if it didn't read updated value but that shouldn't cause any issues here.

opennms-config/src/main/java/org/opennms/netmgt/config/SnmpPeerFactory.java
```diff
@@ -160,17 +170,28 @@ private Lock getWriteLock() {
...
         }
-        if (s_singleton.encryptionEnabled) {
-            s_singleton.encryptSnmpConfig();
+    }
+
+    private static void encryptSingletonIfEnabled() {
+        final SnmpPeerFactory singleton = s_singleton;
+        if (singleton != null && singleton.encryptionEnabled) {
+            singleton.encryptSnmpConfig();
         }
```
cgorantla, line 199: Note that `setInstance` and `setFile` also can update singleton. `init`, `setInstance`, `setFile`, they are all synchronized.  `encryptSnmpConfig` was part of init.

Now `encryptSingletonIfEnabled` is out of the synchronized block, this may cause issues.

JSimo: Added an fix for it.

cgorantla (approved):
LGTM!

Thanks for your contribution @JSimo!

Will merge once once CI passes.

marshallmassengill (approved):
LGTM!

PR #8577, opened 2026-06-15
NMS-19902: bound the VMware CIM connection and probe each host address

marshallmassengill:
The VMware integration connects directly to a host system's CIM service (port 5989) through the SBLIM client, which opens its socket with no connection timeout. An unreachable CIM port therefore blocked the importer's single thread for the operating system connect timeout, stalling requisition imports (and likewise the VmwareCim monitor and collector).

- Add a bounded reachability check (checkCimReachable) in VmwareViJavaAccess.queryCimObjects, using a new cimTimeout that is wired from the cimTimeout/timeout parameters in the importer, monitor, and collector; setHttpTimeOut now uses it as well.
- Log a clear WARN when a host's CIM service is unreachable during import (previously DEBUG only, so the cause was invisible).
- Probe each interface address that is passed to queryCimObjects rather than reusing the first one cached per host, so a host with a reachable secondary address is no longer wrongly marked unreachable.
- Add VmwareCimReachabilityTest and document the cimTimeout parameter.

I validated this in my homelab.  Works as described above.

Assisted-by: Claude Code: Opus 4.7/4.8

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19902

### DIFF
--- a/integrations/opennms-vmware/src/main/java/org/opennms/protocols/vmware/VmwareViJavaAccess.java
+++ b/integrations/opennms-vmware/src/main/java/org/opennms/protocols/vmware/VmwareViJavaAccess.java
@@ -423,7 +423,7 @@ public class VmwareViJavaAccess implements AutoCloseable {
             // An explicit address was supplied (for example, the importer probing each of a
             // host's interface addresses in turn). Use it directly rather than a value cached
             // from an earlier call, so every candidate address is actually tried.
-            cimAgentAddress = "https://" + primaryIpAddress + ":5989";
+            cimAgentAddress = cimUrlFor(primaryIpAddress);
         } else {
             if (!m_hostSystemCimUrls.containsKey(hostSystem)) {
                 final String ipAddress = getPrimaryHostSystemIpAddress(hostSystem);
@@ -433,7 +433,7 @@ public class VmwareViJavaAccess implements AutoCloseable {
                     return cimObjects;
                 }
 
-                m_hostSystemCimUrls.put(hostSystem, "https://" + ipAddress + ":5989");
+                m_hostSystemCimUrls.put(hostSystem, cimUrlFor(ipAddress));
             }
 
             cimAgentAddress = m_hostSystemCimUrls.get(hostSystem);
@@ -452,7 +452,7 @@ public class VmwareViJavaAccess implements AutoCloseable {
         CIMNameSpace ns = new CIMNameSpace(cimAgentAddress, namespace);
         CIMClient cimClient = new CIMClient(ns, userPr, pwCred);
 
-        cimClient.getSessionProperties().setHttpTimeOut(m_cimTimeout);
+        cimClient.getSessionProperties().setHttpTimeOut(boundedTimeout(m_cimTimeout));
 
         // very important to query esx5 hosts
         cimClient.useMPost(false);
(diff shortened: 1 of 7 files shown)


dino2gnt (dismissed):
I'm fine with it.

christianpape (dismissed):
LGTM.

integrations/opennms-vmware/src/main/java/org/opennms/protocols/vmware/VmwareViJavaAccess.java
```diff
@@ -447,6 +479,39 @@ public List<CIMObject> queryCimObjects(HostSystem hostSystem, String cimClass) t
...
+
+        try (Socket socket = new Socket()) {
+            socket.connect(new InetSocketAddress(host, port), timeout);
+        } catch (IOException e) {
+            final ConnectException connectException = new ConnectException("CIM service at " + host + ":" + port
+                    + " is not reachable within " + timeout + " ms (" + e.getMessage() + ")");
+            connectException.initCause(e);
+            throw connectException;
+        }
+    }
```
indigo423, line 522: **P2/P3/P4/P5 + P1 helper.** A few robustness gaps in `checkCimReachable`, plus the helper that the IPv6 fix (call-site suggestions below) needs:

- **P2 (`cimTimeout = 0`):** `Socket.connect(addr, 0)` is an *infinite* timeout per the JDK contract, re-introducing the unbounded block this PR removes. `cimTimeout` comes straight from operator params with no range check (`VmwareImportRequest` only validates parseability).
- **P3 (negative):** `Socket.connect` throws an unchecked `IllegalArgumentException` on a negative timeout; it isn't caught here nor by the importer's `catch (ConnectException | RemoteException | CIMException)`, so it aborts the import path.
- **P4:** `URI.create` throws `IllegalArgumentException` on malformed authority chars *before* the `host == null` guard runs.
- **P5:** every `IOException` (connection-refused, no-route) is reported as "not reachable within N ms", but those aren't timeouts.

This suggestion clamps non-positive timeouts to `DEFAULT_TIMEOUT`, wraps `URI.create`, rewords the message, and adds the `cimUrlFor` helper used by the two call-site suggestions.

```suggestion
    protected void checkCimReachable(final String cimAgentAddress, final int timeout) throws ConnectException {
        final String host;
        final int port;
        try {
            final URI uri = URI.create(cimAgentAddress);
            host = uri.getHost();
            port = uri.getPort();
        } catch (IllegalArgumentException e) {
            // Unparseable address; let the CIM client surface the problem.
            return;
        }

        if (host == null || port < 0) {
            // Malformed address; let the CIM client surface the problem.
            return;
        }

        // Socket.connect treats 0 as an infinite timeout and rejects negative values with an
        // IllegalArgumentException, so clamp non-positive values to the default bound.
        final int connectTimeout = timeout > 0 ? timeout : DEFAULT_TIMEOUT;

        try (Socket socket = new Socket()) {
            socket.connect(new InetSocketAddress(host, port), connectTimeout);
        } catch (IOException e) {
            final ConnectException connectException = new ConnectException("CIM service at " + host + ":" + port
                    + " could not be reached (timeout " + connectTimeout + " ms): " + e.getMessage());
            connectException.initCause(e);
            throw connectException;
        }
    }

    /**
     * Builds the CIM agent URL for an address, bracketing IPv6 literals so the result
     * is a valid URI authority (for example {@code https://[fe80::1]:5989}).
     *
     * @param ipAddress the host address
     * @return the CIM agent URL
     */
    private static String cimUrlFor(final String ipAddress) {
        final boolean ipv6 = ipAddress.indexOf(':') >= 0 && !ipAddress.startsWith("[");
        return "https://" + (ipv6 ? "[" + ipAddress + "]" : ipAddress) + ":5989";
    }
```

integrations/opennms-vmware/src/main/java/org/opennms/protocols/vmware/VmwareViJavaAccess.java
```diff
@@ -394,23 +417,32 @@ public List<CIMObject> queryCimObjects(HostSystem hostSystem, String cimClass, S
...
+        final String cimAgentAddress;
 
-            if (ipAddress == null) {
-                ipAddress = getPrimaryHostSystemIpAddress(hostSystem);
-            }
+        if (primaryIpAddress != null) {
+            // An explicit address was supplied (for example, the importer probing each of a
+            // host's interface addresses in turn). Use it directly rather than a value cached
+            // from an earlier call, so every candidate address is actually tried.
+            cimAgentAddress = "https://" + primaryIpAddress + ":5989";
```
indigo423: **P1 (IPv6 host silently disables the probe).** For an IPv6 host, `InetAddressUtils.str()` returns a bare literal (`fe80::1`), so `"https://" + ipAddress + ":5989"` is an invalid URI authority — `URI.create(...).getHost()` returns `null`, `checkCimReachable` hits its `host == null` early-out and does nothing, and the unbounded connect this PR fixes still hangs (and the same malformed URL reaches `CIMClient`). Route through the `cimUrlFor` helper from the `checkCimReachable` suggestion so IPv6 literals get bracketed.

```suggestion
            cimAgentAddress = cimUrlFor(primaryIpAddress);
```

indigo423 (commented):
The core fix is sound; these are robustness suggestions on the new connect-bounding path. Each inline comment is a one-click GitHub suggestion — apply or dismiss per your call.

**Patches suggested inline (5):**
- **P1** `High` — IPv6 host address is unbracketed, so the probe silently no-ops and the unbounded connect still hangs on IPv6 ESXi hosts.
- **P2** `High` — `cimTimeout = 0` → `Socket.connect(addr, 0)` is an infinite timeout.
- **P3** `Medium` — negative `cimTimeout` throws an uncaught `IllegalArgumentException`.
- **P4** `Low` — `URI.create` can throw before the null-host guard.
- **P5** `Low` — non-timeout `IOException`s are mislabeled as timeouts.

**Deferred (not posted as suggestions — design/intent calls for you):**
- Collector & monitor wire the generic `timeout` into `setCimTimeout` rather than a dedicated `cimTimeout`; the documented `cimTimeout` (default 3000) knob therefore doesn't exist on those two paths. Behavior is preserved at the default, but the contract diverges from the importer + docs.
- The probe is TOCTOU + double-connect and only bounds TCP, not the TLS handshake (`setHttpTimeOut` is a read timeout) — narrows but doesn't fully close the unbounded-connect window.
- `new InetSocketAddress(host, port)` resolves DNS eagerly, outside the `cimTimeout` bound — harmless for IP literals, relevant for hostname inputs.
- `ConnectException` is now logged at WARN per unreachable host — intentional, but watch for log-flood on large CIM-less vCenters.

**Test follow-up:** consider adding IPv6 (P1) and `cimTimeout = 0`/negative (P2/P3) cases to `VmwareCimReachabilityTest` so these can't silently regress.

integrations/opennms-vmware/src/main/java/org/opennms/protocols/vmware/VmwareViJavaAccess.java
```diff
@@ -394,23 +417,32 @@ public List<CIMObject> queryCimObjects(HostSystem hostSystem, String cimClass, S
...
 
+                if (ipAddress == null) {
+                    logger.warn("Cannot determine ip address for host system '{}'", hostSystem.getMOR().getVal());
+                    return cimObjects;
+                }
 
-            if (ipAddress == null) {
-                logger.warn("Cannot determine ip address for host system '{}'", hostSystem.getMOR().getVal());
-                return cimObjects;
+                m_hostSystemCimUrls.put(hostSystem, "https://" + ipAddress + ":5989");
```
indigo423: **P1 (IPv6), cache path.** Same fix on the cached branch.

```suggestion
                m_hostSystemCimUrls.put(hostSystem, cimUrlFor(ipAddress));
```

marshallmassengill:
I think I've addressed the comments @indigo423.  The IPv6 stuff ended up not being too bad.

indigo423 (approved):
no objections on my end

cgorantla (approved):
LGTM!

PR #8581, opened 2026-06-16
NMS-19843: Update PrimeVue presets to match Feather themes throughout app

synqotik:
Part of converting Feather -> PrimeVue. Update app-wide themes to have more-or-less FeatherDS themes applied to PrimeVue components going forward, including both light/dark modes, without needing component-specific `:deep()` or other overrides.

This also gives us a central place to globally update any PrimeVue styles as desired.

Note that `PrimeVueTest.vue` is a temporary test file to view Feather and PrimeVue side-by-side to visually confirm themes are correct. Will eventually delete it once the conversion is complete.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19843

### DIFF
--- a/ui/src/components/Nodes/AssetFilterPanel.vue
+++ b/ui/src/components/Nodes/AssetFilterPanel.vue
@@ -162,37 +162,10 @@ onMounted(() => {
     :deep(.p-datatable-tbody > tr > td) {
       padding: 0.25rem 0.5rem;
       vertical-align: middle;
-      border-color: var(--feather-border-on-surface);
-      color: var(--feather-primary-text-on-surface);
     }
 
     :deep(.p-datatable-thead > tr > th) {
       padding: 0.4rem 0.5rem;
-      background-color: var(--feather-background);
-      border-bottom: 1px solid var(--feather-border-on-surface);
-      color: var(--feather-secondary-text-on-surface);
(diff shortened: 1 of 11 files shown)


marshallmassengill (approved):
Looks good.  The one thing that got flagged that may be worth a look is this one:

SnmpDataCollectionSourceProfilesDrawer.vue deletes an entire <style lang="scss"> global block that handled the teleported .p-autocomplete-overlay. The preset's overlay.select tokens cover background/border/color, but the deleted block also set:
  - font-family: var(--feather-font-family) on the overlay and input — not a color token, so it isn't replaced by the preset. The dropdown text may revert to PrimeVue's default font.
  - .p-autocomplete-option hover/selected background (rgba(255,255,255,0.06)) — now falls back to Aura's option-focus defaults.

I don't think that's actually a huge concern mind you.

synqotik:
> Looks good. The one thing that got flagged that may be worth a look is this one:
> 

Good catch! I will have that fixed in the next PR

synqotik:
> > Looks good. The one thing that got flagged that may be worth a look is this one:
> 
> Good catch! I will have that fixed in the next PR

Note, this has been fixed in what will be the next PR.

PR #8582, opened 2026-06-16
NMS-19905: Fix reflected XSS in application base-URL helper

marshallmassengill:
Util.calculateUrlBase/substituteUrl build the app base URL from request-derived values (scheme, Host/X-Forwarded-Host header, context path), and callers emit it unescaped into HTML/JS — most notably __baseHref in bootstrap.jsp. Because the default template resolves the host from X-Forwarded-Host (no HTTP host-char validation), a poisoned header reflects into the response: reflected XSS. SonarQube flags 142 instances of this (javasecurity:S5131, High).

Fix: HTML-encode the assembled URL once in substituteUrl — the single chokepoint both calculateUrlBase overloads pass through — via the existing WebSecurityUtils.sanitizeString (OWASP Encode.forHtml). Legitimate URL characters pass through unchanged; HTML metacharacters are escaped. This clears every calculateUrlBase-rooted alert at the source, with no per-sink JSP edits and no new dependency.

Testing: Added two UtilTest regression cases (legitimate URL unchanged; malicious X-Forwarded-Host neutralized). opennms-web-api builds; UtilTest passes 8/8. Branch is -smoke to run the smoke suite.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19905

### DIFF
--- a/opennms-web-api/src/main/java/org/opennms/web/api/Util.java
+++ b/opennms-web-api/src/main/java/org/opennms/web/api/Util.java
@@ -141,7 +141,11 @@ public abstract class Util extends Object {
                 out.append(c);
             }
         }
-        return out.toString();
+        // The scheme, host, and context path are derived from the request (e.g. the Host
+        // or X-Forwarded-Host header) and are emitted unescaped into HTML/JS by callers,
+        // so HTML-encode here to neutralize reflected XSS. Legitimate URL characters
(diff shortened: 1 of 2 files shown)


cgorantla (approved):
Looks good. 

Thanks for taking care of it.

PR #8583, opened 2026-06-17
NMS-17999: Replace outdated json and commons-io library in hawtio-osgi

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-17999
* JIRA: https://opennms.atlassian.net/browse/NMS-17997


cgorantla (commented):
I think we should add a smoke test on OpenNMS/Minion/Sentinel to check if hawtio feature loads cleanly

christianpape:
> I think we should add a smoke test on OpenNMS/Minion/Sentinel to check if hawtio feature loads cleanly

Done.

smoke-test/src/test/java/org/opennms/smoketest/hawtio/AbstractHawtioIT.java
```diff
@@ -0,0 +1,69 @@
...
+                        final String command ="bundle:list";
+                        pipe.println(command);
+                        pipe.println("logout");
+
+                        // Wait for karaf to process the commands
+                        await().atMost(10, SECONDS).until(sshClient.isShellClosedCallable());
+
+                        // Read stdout and verify
+                        final String shellOutput = sshClient.getStdout();
+                        final boolean bundleActive = Arrays.stream(shellOutput.split("\n"))
```
cgorantla: I think it would be good to check if `hawtio` feature itself comes up properly. May need to enable this feature explicitly in this test.

christianpape: done.

smoke-test/src/test/java/org/opennms/smoketest/hawtio/HawtioMinionIT.java
```diff
@@ -0,0 +1,41 @@
...
+package org.opennms.smoketest.hawtio;
+
+import org.junit.ClassRule;
+import org.junit.experimental.categories.Category;
+import org.opennms.smoketest.junit.MinionTests;
+import org.opennms.smoketest.stacks.OpenNMSStack;
+import org.opennms.smoketest.stacks.StackModel;
+
+@Category(MinionTests.class)
+public class HawtioMinionIT extends AbstractHawtioIT {
```
cgorantla: If `hawtio` was installed already on Minion/Sentinel, we can skip this test as health checks already validates this feature.  Same observation for Sentinel.

smoke-test/src/test/java/org/opennms/smoketest/hawtio/HawtioCoreIT.java
```diff
@@ -0,0 +1,42 @@
...
+public class HawtioCoreIT extends OpenNMSSeleniumIT {
+
+    @Before
+    public void setUp() throws Exception {
+        driver.get(getBaseUrlInternal() + "hawtio/jvm/connect");
+    }
+
+    @Test
+    public void hasContent() {
+        assertNotNull(driver.findElement(By.xpath("/html/body/hawtio-app/hawtio-loading/div/page/div[2]/page-header/div[1]/div[2]/img")));
```
cgorantla: Only keep this test and remove the test in Abstract.  Also may be update this test to validate hawtio Rest API instead of depending upon an html element that may not be stable across releases.

christianpape: done.

cgorantla (approved):
LGTM!

PR #8590, opened 2026-06-18
NMS-18185 : Upgrade Spring and Hibernate to 5.x

cgorantla:
This PR upgrades Spring and Hibernate to 5.x versions.
Removes Eclipse Gemini blueprint used in sentinel with a custom Spring 5.3 extender.
Several updates to Hibernate queries to support 5.x
camel route/builder component migration
Spring security updates as well.

Also this removes custom opennms patched versions for Spring/Hibernate.

### Dependency version bumps 

  | Component | Before | After |
  |---|---|---|
  | Spring Framework | 4.2.9 (custom `.ONMS.1` patched) | 5.3.39 (stock ServiceMix bundles) |
  | Spring Security | 4.2.21 (`_ONMS_3` patched) | 5.8.16 |
  | Hibernate ORM | 3.6.11 (custom ONMS patch) | 5.6.15.Final (stock) |
  | Hibernate Validator | 4.3.2 | 6.2.5.Final |
  | Apache Camel | 2.21.5 | 3.22.4 |
  | Eclipse Gemini Blueprint | 2.1.0 | removed (replaced by custom Spring 5.3 OSGi extender) |

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-18185


cgorantla:
Just to note that we were trying to move to Spring 6.x and Hibernate 6.x, but we hit a roadblock w.r.to Karaf. 

Hibernate 6 forces Spring 6, Spring 6 is one version across OpenNMS code base, and its webapp needs jakarta.servlet (Jetty 11/12) which Karaf 4.4.x can't provide. 

We will do this migration once Karaf 4.5 gets released.

cgorantla:
Re-targeted this to `release-36.x`

christianpape (dismissed):
LGTM.

marshallmassengill (dismissed):
LGTM!  Very exciting stuff!

marshallmassengill (dismissed):
I'm going to put the claude review output below.  Some of this does seem like it is worth addressing but I'm not sure if any of it should hold this up.

Should fix before merge

  1. The EOL Spring 4.2.9(.ONMS.1) line is still defined and shipped — contradicting the PR's stated goal. container/features/pom.xml:19-20 keeps legacySpring42PatchedVersion=4.2.9.RELEASE_1.ONMS.1, spring-legacy.xml still defines the feature, and opennms-full-assembly/pom.xml:323 still explicitly pulls spring/4.2.9.RELEASE_1 into the shipped Karaf system repo (I verified all three at the PR head). The Spring4Shell-era jars remain downloadable at build time (keeping the private-Maven dependency the PR claims to remove) and runtime-installable via feature:install.

  2. custom.properties exports ~19 phantom Spring packages at 5.3.39 that no longer exist in Spring 5 — org.springframework.orm.hibernate3/.hibernate4/.jdo, oxm.castor, oxm.xmlbeans, web.servlet.mvc.multiaction, beans.factory.access, ui.velocity, etc. (verified against ServiceMix spring-orm 5.3.39_2, which contains only orm.hibernate5/orm.jpa). Any bundle importing one resolves cleanly against the system bundle, then dies with NoClassDefFoundError at first use — phantom exports turn resolve-time errors into runtime ones. Also javax.persistence is still exported at version=2.0 while the platform moved to JPA 2.2, leaving two providers of the entity-annotation packages.

  3. The antlr 2.7.7 wrap in the hibernate5 feature dropped its Import-Package=org.hibernate.hql.ast override without adding the Hibernate 5 equivalent 
  (org.hibernate.hql.internal.ast). Hibernate's classic HQL translator hands antlr the token class by name, and antlr must load it itself. Plausible failure: ClassNotFoundException: HqlToken on HQL queries inside OSGi — exactly the Sentinel DAO path the PR's own comment in features/distributed/dao/impl/pom.xml calls out. Medium confidence (TCCL fallback may mask it) — worth one targeted Sentinel HQL smoke test before merge.

  4. Log-assertion checks were deleted from three IT suites with no explanation. EventdIT.java:128 has MockLogAppender.assertNoWarningsOrGreater() commented out (also inherited by the NMS-16978 EventdIT); NotificationsITCase.java:187 deletes the same check for every notifd IT subclass; LatencyStoringServiceMonitorAdaptorIT likewise. Question for the author: what now logs at WARN+ during an eventd round-trip under Hibernate 5? These deletions can mask exactly the regressions this upgrade could introduce.

  5. Five shipped spring-security.d/*.disabled example configs (ldap, activeDirectory, radius, both SSO/Kerberos variants) still pin spring-security-4.2.xsd. As shipped they're inert, but the moment a user enables one and adds any security-namespace element, Spring Security 5.8's namespace version check hard-fails context startup — breaking all webapp logins with a confusing error. This is precisely the failure mode the PR's own release notes warn about; bumping five schemaLocation lines closes it.

  6. spring-security-opennms — the single riskiest rewritten feature — is missing from the karaf-maven-plugin verify gate in container/features/pom.xml, so the 4.2→5.8 feature rewrite is the one major upgrade with no build-time resolution check. Relatedly, the retained spring-security-kerberos 1.0.1.RELEASE_ONMS_3 bundle was compiled against Security 4.x APIs and is acknowledged-but-unverified against 5.8.

  7. Hand-built ProducerTemplates leak on blueprint reload in all four IPC entry points — CamelRpcClientFactory, CamelRemoteMessageDispatcherFactory, JmsTwinPublisher, JmsTwinSubscriber now call camelContext.createProducerTemplate() in init but never stop the template in stop/destroy (the removed @EndpointInject versions were context-managed). These bundles use cm:property-placeholder update-strategy="reload", so every config-admin update of org.opennms.core.ipc on a Minion leaks another template's producer cache and executor threads.

  Worth fixing or answering (minor)

  - Transaction-timeout propagation silently dropped from all DAO queries. The old HibernateTemplate.find() path applied the remaining Spring transaction timeout as a JDBC query timeout; the hand-rolled session.createQuery() replacements in AbstractDaoHibernate (find/findUnique/queryInt/bulkDelete/pagination) never do, and EventCountDaoHibernate explicitly deleted its applyTransactionTimeout call. Slow queries in timed transactions now run to completion.
  - AbstractDaoHibernate.lock() is now a native INSERT ... ON CONFLICT, which auto-flushes the whole session (no synchronized query spaces) and throws TransactionRequiredException outside a transaction — both new preconditions relative to the old get(..., PESSIMISTIC_WRITE). Its one caller (UpsertTemplate, mid-provisiond-scan) mostly masks this, but it's a semantic change worth a comment.
  - Operator-facing break: IpInterfaceDaoHibernate.findByServiceType's HQL is overridable via the org.opennms.dao.ipinterface.findByServiceType system property; any install overriding it with legacy ? parameters gets a runtime QueryException after upgrade. The release note covers code-level callers but not this property — add a line.
  - UpperCaseMd5PasswordEncoder.matches() uses String.equalsIgnoreCase() where the removed Spring Security encoder used constant-time comparison — a reintroduced timing side channel in a security-labeled rewrite (low practical exploitability; only optional user-maintained configs wire it).
  - Duplicate spec-package exporters: two javax.validation features coexist (1.0.0.GA in features-core.xml, 2.0.1.Final in features.xml — all refs unversioned,
  resolver-order-dependent), and opennms-distributed-core-jms installs both javax.jms-api 2.0 and geronimo-jms_1.1_spec, risking split class spaces / ClassCastException on JMS objects depending on wiring order. The team clearly engineered around JMS 1.1/2.0 elsewhere; this double install is the residue.
  - Spring extender edges (the core lifecycle design survived adversarial reading — generation tokens, never holding the lifecycle lock across refresh(), and the closing
  fail-fast are all sound): (a) the osgi: compat parser silently ignores every Gemini attribute except interface/filter — cardinality="0..1" silently becomes mandatory, bean-name binds to an arbitrary service; it should reject what it doesn't implement (in-tree usage is only three id+interface references, so this is latent third-party/plugin risk); (b) shutdown is synchronous, unordered, and unbounded — a hanging bean destroy hangs karaf stop where Gemini bounded it at ~10s and ordered by dependency; (c) permanently broken contexts (bad XML, CNFE) retry every ≤5 minutes forever, indistinguishable from waiting on a service; (d) BundleWiring.listResources() can return null on a stale wiring → avoidable NPE in findBundleResources; (e) NamespaceProviderRegistry and the wildcard resource-index arithmetic have zero test coverage — the piece most likely to regress silently (empty packagesToScan).
  - CamelMessageConsumerManager logs "Route started and consuming from..." unconditionally after addRoutes() — before the context has started, and from a second JmsQueueNameFactory rather than the endpoint the route uses. Sentinel smoke tests keying on this line can pass while the consumer is down.
  - BMP stats aggregator now writes each interval as one all-or-nothing transaction (one bad row rolls back the whole interval, vs. one lost row before), and its new
  guarded() wrapper swallows Throwable including Errors with only a log line.
  - Test-quality nits: NginxHttpsProxyIT.redirectsStayOnTheProxy is effectively vacuous (assertion nested in an if that a non-redirect skips; the matcher accepts leaked backend URLs); FilterFavoriteServiceIT now wraps calls in a test-side transaction so the service's own new @Transactional proxying is no longer exercised; @Transactional on @Before methods in MinionDaoIT/IpInterfaceDaoHibernateIT is a no-op that will trap the next non-transactional test method; CustomHttpClientConfigurer's HttpClient 3→4 port drops preemptive auth (currently no in-tree users — latent).
  - Cosmetic drift: blueprint/spring XMLs still pin camel-*-2.21.5.xsd / spring-beans-4.2.xsd schemaLocations (verified harmless at runtime — Camel 3's spring.schemas remaps them — but misleading); stale undefined feature refs (spring/4.0.9, 4.1.9, 4.3.30) in container/shared/pom.xml and full-assembly; hibernate5 feature hardcodes third-party versions including byte-buddy 1.12.18 vs the root pom's 1.14.9.

cgorantla:
> I'm going to put the claude review output below. Some of this does seem like it is worth addressing but I'm not sure if any of it should hold this up.
> 
> Should fix before merge
> 
> 1. The EOL Spring 4.2.9(.ONMS.1) line is still defined and shipped — contradicting the PR's stated goal. container/features/pom.xml:19-20 keeps legacySpring42PatchedVersion=4.2.9.RELEASE_1.ONMS.1, spring-legacy.xml still defines the feature, and opennms-full-assembly/pom.xml:323 still explicitly pulls spring/4.2.9.RELEASE_1 into the shipped Karaf system repo (I verified all three at the PR head). The Spring4Shell-era jars remain downloadable at build time (keeping the private-Maven dependency the PR claims to remove) and runtime-installable via feature:install.
> 2. custom.properties exports ~19 phantom Spring packages at 5.3.39 that no longer exist in Spring 5 — org.springframework.orm.hibernate3/.hibernate4/.jdo, oxm.castor, oxm.xmlbeans, web.servlet.mvc.multiaction, beans.factory.access, ui.velocity, etc. (verified against ServiceMix spring-orm 5.3.39_2, which contains only orm.hibernate5/orm.jpa). Any bundle importing one resolves cleanly against the system bundle, then dies with NoClassDefFoundError at first use — phantom exports turn resolve-time errors into runtime ones. Also javax.persistence is still exported at version=2.0 while the platform moved to JPA 2.2, leaving two providers of the entity-annotation packages.
> 3. The antlr 2.7.7 wrap in the hibernate5 feature dropped its Import-Package=org.hibernate.hql.ast override without adding the Hibernate 5 equivalent
>    (org.hibernate.hql.internal.ast). Hibernate's classic HQL translator hands antlr the token class by name, and antlr must load it itself. Plausible failure: ClassNotFoundException: HqlToken on HQL queries inside OSGi — exactly the Sentinel DAO path the PR's own comment in features/distributed/dao/impl/pom.xml calls out. Medium confidence (TCCL fallback may mask it) — worth one targeted Sentinel HQL smoke test before merge.
> 4. Log-assertion checks were deleted from three IT suites with no explanation. EventdIT.java:128 has MockLogAppender.assertNoWarningsOrGreater() commented out (also inherited by the NMS-16978 EventdIT); NotificationsITCase.java:187 deletes the same check for every notifd IT subclass; LatencyStoringServiceMonitorAdaptorIT likewise. Question for the author: what now logs at WARN+ during an eventd round-trip under Hibernate 5? These deletions can mask exactly the regressions this upgrade could introduce.
> 5. Five shipped spring-security.d/*.disabled example configs (ldap, activeDirectory, radius, both SSO/Kerberos variants) still pin spring-security-4.2.xsd. As shipped they're inert, but the moment a user enables one and adds any security-namespace element, Spring Security 5.8's namespace version check hard-fails context startup — breaking all webapp logins with a confusing error. This is precisely the failure mode the PR's own release notes warn about; bumping five schemaLocation lines closes it.
> 6. spring-security-opennms — the single riskiest rewritten feature — is missing from the karaf-maven-plugin verify gate in container/features/pom.xml, so the 4.2→5.8 feature rewrite is the one major upgrade with no build-time resolution check. Relatedly, the retained spring-security-kerberos 1.0.1.RELEASE_ONMS_3 bundle was compiled against Security 4.x APIs and is acknowledged-but-unverified against 5.8.
> 7. Hand-built ProducerTemplates leak on blueprint reload in all four IPC entry points — CamelRpcClientFactory, CamelRemoteMessageDispatcherFactory, JmsTwinPublisher, JmsTwinSubscriber now call camelContext.createProducerTemplate() in init but never stop the template in stop/destroy (the removed @EndpointInject versions were context-managed). These bundles use cm:property-placeholder update-strategy="reload", so every config-admin update of org.opennms.core.ipc on a Minion leaks another template's producer cache and executor threads.
> 
> Worth fixing or answering (minor)
> 
> * Transaction-timeout propagation silently dropped from all DAO queries. The old HibernateTemplate.find() path applied the remaining Spring transaction timeout as a JDBC query timeout; the hand-rolled session.createQuery() replacements in AbstractDaoHibernate (find/findUnique/queryInt/bulkDelete/pagination) never do, and EventCountDaoHibernate explicitly deleted its applyTransactionTimeout call. Slow queries in timed transactions now run to completion.
> * AbstractDaoHibernate.lock() is now a native INSERT ... ON CONFLICT, which auto-flushes the whole session (no synchronized query spaces) and throws TransactionRequiredException outside a transaction — both new preconditions relative to the old get(..., PESSIMISTIC_WRITE). Its one caller (UpsertTemplate, mid-provisiond-scan) mostly masks this, but it's a semantic change worth a comment.
> * Operator-facing break: IpInterfaceDaoHibernate.findByServiceType's HQL is overridable via the org.opennms.dao.ipinterface.findByServiceType system property; any install overriding it with legacy ? parameters gets a runtime QueryException after upgrade. The release note covers code-level callers but not this property — add a line.
> * UpperCaseMd5PasswordEncoder.matches() uses String.equalsIgnoreCase() where the removed Spring Security encoder used constant-time comparison — a reintroduced timing side channel in a security-labeled rewrite (low practical exploitability; only optional user-maintained configs wire it).
> * Duplicate spec-package exporters: two javax.validation features coexist (1.0.0.GA in features-core.xml, 2.0.1.Final in features.xml — all refs unversioned,
>   resolver-order-dependent), and opennms-distributed-core-jms installs both javax.jms-api 2.0 and geronimo-jms_1.1_spec, risking split class spaces / ClassCastException on JMS objects depending on wiring order. The team clearly engineered around JMS 1.1/2.0 elsewhere; this double install is the residue.
> * Spring extender edges (the core lifecycle design survived adversarial reading — generation tokens, never holding the lifecycle lock across refresh(), and the closing
>   fail-fast are all sound): (a) the osgi: compat parser silently ignores every Gemini attribute except interface/filter — cardinality="0..1" silently becomes mandatory, bean-name binds to an arbitrary service; it should reject what it doesn't implement (in-tree usage is only three id+interface references, so this is latent third-party/plugin risk); (b) shutdown is synchronous, unordered, and unbounded — a hanging bean destroy hangs karaf stop where Gemini bounded it at ~10s and ordered by dependency; (c) permanently broken contexts (bad XML, CNFE) retry every ≤5 minutes forever, indistinguishable from waiting on a service; (d) BundleWiring.listResources() can return null on a stale wiring → avoidable NPE in findBundleResources; (e) NamespaceProviderRegistry and the wildcard resource-index arithmetic have zero test coverage — the piece most likely to regress silently (empty packagesToScan).
> * CamelMessageConsumerManager logs "Route started and consuming from..." unconditionally after addRoutes() — before the context has started, and from a second JmsQueueNameFactory rather than the endpoint the route uses. Sentinel smoke tests keying on this line can pass while the consumer is down.
> * BMP stats aggregator now writes each interval as one all-or-nothing transaction (one bad row rolls back the whole interval, vs. one lost row before), and its new
>   guarded() wrapper swallows Throwable including Errors with only a log line.
> * Test-quality nits: NginxHttpsProxyIT.redirectsStayOnTheProxy is effectively vacuous (assertion nested in an if that a non-redirect skips; the matcher accepts leaked backend URLs); FilterFavoriteServiceIT now wraps calls in a test-side transaction so the service's own new @transactional proxying is no longer exercised; @transactional on @before methods in MinionDaoIT/IpInterfaceDaoHibernateIT is a no-op that will trap the next non-transactional test method; CustomHttpClientConfigurer's HttpClient 3→4 port drops preemptive auth (currently no in-tree users — latent).
> * Cosmetic drift: blueprint/spring XMLs still pin camel-*-2.21.5.xsd / spring-beans-4.2.xsd schemaLocations (verified harmless at runtime — Camel 3's spring.schemas remaps them — but misleading); stale undefined feature refs (spring/4.0.9, 4.1.9, 4.3.30) in container/shared/pom.xml and full-assembly; hibernate5 feature hardcodes third-party versions including byte-buddy 1.12.18 vs the root pom's 1.14.9.

Most of these should be resolved.

marshallmassengill (approved):
LGTM!

PR #8594, opened 2026-06-24
NMS-19899: "OID Value format is invalid" for Mask

dino2gnt:
oid mask validation was using OID_PATTERN instead of OID_MASK_PATTERN when validating oid masks in the systemdef drawer.  If it's a mask, use the OID_MASK_PATTERN instead.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19899


cgorantla (approved):
LGTM! 

Good to see UI is getting used :-)

PR #8596, opened 2026-06-25
NMS-19928: Fix filter rule precedence bug by parenthesizing the parsed rule

marshallmassengill:
I've run a few different tests against this and it really does seem to make a performance difference at larger scales so I think it's worth putting in the PR for.

Filter rules are translated to SQL WHERE clauses in JdbcFilterDao, and callers append further constraints to that clause: the deleted/isManaged filter in getIPAddressList, and the nodeID/ipAddr/serviceName constraints in getSQLStatement(rule, nodeId, ipaddr, service). parseRule emitted the rule unparenthesized, so for a rule with a top-level OR the appended "AND ..." bound only to the last OR branch:

```
WHERE <a> OR <b> AND ipInterface.ipaddr = ?
parses as
WHERE <a> OR (<b> AND ipInterface.ipaddr = ?)
```

This caused both a correctness bug (an address matching neither branch was still reported valid whenever the other branch matched some interface) and full-table scans on large instances, since the address no longer restricted the query.

Wrap the parsed rule in parentheses so every appended constraint applies to the whole rule: 

`WHERE (<a> OR <b>) AND ipInterface.ipaddr = ?`

Adds a unit test for the generated SQL and an IT regression case for an OR-rule resolved by address.

Assisted by Claude Opus 4.8

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19928

### DIFF
--- a/opennms-config/src/main/java/org/opennms/netmgt/filter/JdbcFilterDao.java
+++ b/opennms-config/src/main/java/org/opennms/netmgt/filter/JdbcFilterDao.java
@@ -708,7 +708,11 @@ public class JdbcFilterDao implements FilterDao, InitializingBean {
             }
             regex.appendTail(tempStringBuff);
             sqlRule = tempStringBuff.toString();
-            return "WHERE " + sqlRule;
+            // Wrap the parsed rule in parentheses so that any clauses appended by callers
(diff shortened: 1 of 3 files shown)


cgorantla (approved):
Awesome. LGTM!

PR #8597, opened 2026-06-26
NMS-19547: Stop black-hole hosts from exhausting the provisiond scan pool

marshallmassengill:
I do not actually know if this will resolve NMS-19547 at this point but I am about 90% confident this is the core issue based on the logs we got from the customer.  Cisco Catalyst switches with web interface enabled are genuinely notorious for having small client pools.  I can confirm that testing with a webserver resulted in the exact behavior that we've seen.  I haven't run this through all the tests so it's possible CircleCI catches something.

Assisted by Claude Opus 4.8

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19547


cgorantla (approved):
LGTM!

Don't see any issues with it.

PR #8599, opened 2026-06-26
NMS-16980: de-flake NodeDetailPageIT.verifyTimeline

marshallmassengill:
Got annoyed with this failing.  Trying a fix for it.

- 5s budget for JS-injected s + server-rendered PNGs  -> atMost(Duration.ofMillis(LOAD_TIMEOUT)) — the suite's standard 120s page-load budget (still returns in ~3-5s on success)
- Transient StaleElementReferenceException/NPE while the DOM churns aborted the whole await - > .ignoreExceptions() — retries on the next poll instead
- getAttribute("src") could be null → NPE in the filter  -> guarded with src != null && src.contains("timeline") 

Assisted by Claude Opus 4.8

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-16980


cgorantla (approved):
LGTM!

PR #8601, opened 2026-06-26
Replace XML-backed user management with PostgreSQL opennms db user table as authoritative user source

makikorp:
Update the Helm repo where users.xml is removed and opennms database table user becomes user authoritative source. The users.xml file is ephemeral and whenever a kubectl delete pod opennms, kubectl rollout restart, or helm upgrade is run all the user data is lost. The fix is that we move the user data into the opennms database as a user table. The data will be persisted over pod deletion, restarts, and upgrades.


dino2gnt:
I wouldn't do this in a release branch.  This should target `develop`.

core/schema/src/main/liquibase/36.0.2/changelog.xml
```diff
@@ -0,0 +1,148 @@
...
+            <column name="user_comments" type="text"/>
+            <column name="password" type="varchar(512)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="password_salt" type="boolean" defaultValueBoolean="true">
+                <constraints nullable="false"/>
+            </column>
+            <column name="tui_pin" type="varchar(32)"/>
+            <column name="time_zone_id" type="varchar(64)"/>
+            <column name="email" type="varchar(256)"/>
```
dino2gnt, line 30: Can we not?

https://wiki.postgresql.org/wiki/Don%27t_Do_This#Don't_use_varchar(n)_by_default

cgorantla:
@makikorp  Thanks for your contribution.  This needs more work and deserves a feature branch. 

1. The feature needs to be targeted to develop 
2. We need to handle any existing users.xml as part of upgrade and move them to DB.
3. All tests need to pass.

Also better to ship smaller PRs one by one into feature branch.  First of them can be adding liquibase tables for users.

PR #8602, opened 2026-06-26
feat(ui): global plugin overlay slot via Teleport (companion to OIA PR #181)

joseanesONMS:
ALEC-302

## Summary

Plugins that declare `UIExtension.getGlobalModuleFileName()` (new OIA default method, [OIA PR #181](https://github.com/OpenNMS/opennms-integration-api/pull/181)) now get their module loaded and rendered in a `<Teleport to="body">` at the root App level — visible on every page, not just under Plugins > menu.

## Changes

| File | What changed |
|---|---|
| `ui/src/main/App.vue` | Import `GlobalPlugin`; compute `globalPlugins`; add `<Teleport to="body">` |
| `ui/src/components/Plugin/GlobalPlugin.vue` | **New** — loads a plugin's global module via the existing `Container` component |
| `ui/src/types/index.ts` | Add optional `globalModuleFileName?: string` to the `Plugin` type |

## Backward compatibility

- Plugins without `getGlobalModuleFileName()` (all existing plugins) return `null`; the REST layer omits or nulls the field; the `v-for` filter skips them — **zero behaviour change**.
- Old OpenNMS that does not serve this field: `globalPlugins` is always empty, `<Teleport>` renders nothing — **no visual or functional regression**.
- No new REST endpoints, no schema migration, no OSGi changes.

## Dependencies

Requires OIA [PR #181](https://github.com/OpenNMS/opennms-integration-api/pull/181) to be merged and a new OIA snapshot published before `globalModuleFileName` appears in `/rest/plugins` responses for plugins that implement the method.

## First consumer

ALEC-302 — ALEC context-aware chatbot overlay.

### DIFF
--- a/ui/src/components/Plugin/GlobalPlugin.vue
+++ b/ui/src/components/Plugin/GlobalPlugin.vue
@@ -1,0 +1,48 @@
+<script setup lang="ts">
+///
+/// Licensed to The OpenNMS Group, Inc (TOG) under one or more
+/// contributor license agreements.  See the LICENSE.md file
+/// distributed with this work for additional information
+/// regarding copyright ownership.
+///
+/// TOG licenses this file to You under the GNU Affero General
+/// Public License Version 3 (the "License") or (at your option)
+/// any later version.  You may not use this file except in
+/// compliance with the License.  You may obtain a copy of the
+/// License at:
+///
+///      https://www.gnu.org/licenses/agpl-3.0.txt
+///
+/// Unless required by applicable law or agreed to in writing,
+/// software distributed under the License is distributed on an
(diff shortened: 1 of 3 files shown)


joseanesONMS:
@marshallmassengill @cgorantla — updated the PR description. Please take a look when you get a chance.

synqotik:
@joseanesONMS 

This seems reasonable for a single use case, but we may need to think more about it.

For example, if there are multiple global plugins, this could get messy very quickly.

If the purpose is to have a chat window, it may make sense to build that into the product (with placement, styling and chat functionality, but not connected to any other functionality in particular), then have a way to connect it with plugins that actually provide the feature / data / etc.

PR #8604, opened 2026-07-01
NMS-19951: Resolve potential wedge in CamelRpcClientFactory for local

marshallmassengill:
Bound local (same-location) RPC execution to the request TTL, or the default RPC timeout when none is set, mirroring the budget the remote path already enforces. A module whose future never completes no longer hangs the caller, which was the root cause of the provisiond scan wedge.

Adds LocalRpcTimeoutTest: a local request with a TTL whose module never completes now completes exceptionally with a TimeoutException at ~TTL; control case still returns the module result.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19951


dino2gnt (approved):
Good.

cgorantla (approved):
LGTM!

PR #8614, opened 2026-07-02
DOCS: fixes for situation and classifications REST API payload examples

dino2gnt:
DOCS DOCS DOCS

Fixups for the example API payloads for situations and classifications.


marshallmassengill (approved):
Looks good

PR #8615, opened 2026-07-02
NMS-19953: Support SSH public-key auth for the Karaf shell on core

marshallmassengill:
Wire Karaf's PublickeyLoginModule into the OpenNMS JAAS realm so the Karaf shell (SSH, port 8101) accepts public-key authentication in addition to password authentication. The key module is "sufficient" and runs before the "required" OpenNMSLoginModule, so a matching key logs in and a missing or non-matching key falls back to password auth. Default behavior is unchanged for installs with no configured keys.

- blueprint.xml: add PublickeyLoginModule (sufficient) reading ${karaf.etc}/keys.properties, plus an Aries property-placeholder so ${karaf.etc} resolves inside the jaas:module option.
- pom.xml: add org.apache.karaf.jaas.modules.publickey to Import-Package; the blueprint references the class only by string, so bnd would not otherwise import it.
- docs: document enabling public-key auth on the Karaf shell page.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19953


cgorantla (approved):
Awesome. LGTM!

dino2gnt (approved):
LGTM!

PR #8616, opened 2026-07-02
NMS-19954: Add clear-based alarm reconciliation (alarmSyncClear) to t…

marshallmassengill:
Adds an alarmSyncClear option that changes how alarm synchronization reconciles reduction keys present on the topic but no longer in the database. When enabled, such orphans are republished with severity CLEARED instead of a null tombstone, with a note appended to the log message. This avoids writing null values that some downstream consumers (e.g. Kafka Connect sinks) mishandle.

The clear is idempotent: once the topic shows a key as CLEARED, later reconciliation cycles leave it alone. alarmSyncClear only takes effect when alarmSync is enabled; otherwise it is warned about and ignored.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19954


cgorantla:
Looks like this is subset of https://github.com/OpenNMS/opennms/pull/8617, we can close this.

marshallmassengill:
> Looks like this is subset of #8617, we can close this.

You're right... I didn't catch this.

PR #8617, opened 2026-07-02
NMS-19955: Alarm Stream mode for Alarm Kafka Producer

marshallmassengill:
Adds an alarmStream option (default false) that switches the alarms topic from changelog/KTable semantics to an append-only event stream. Updates stay keyed by reduction key, but deletes become silent (no null tombstone is emitted).

Today the topic represents a delete as a null-value tombstone, which suits consumers that read it as a compacted KTable but trips up consumers that just want a record per alarm change. In stream mode, a consumer sees the rising edge (an alarm is raised or escalates) and the falling edge (the alarm's CLEARED update) as ordinary records, and simply stops seeing a key once it's resolved.

What's included
  - alarmStream config flag + blueprint wiring (streamMode on the producer).
  - Silent-delete path in OpennmsKafkaProducer.updateAlarm.
  - A startup guard: because silent deletes rely on the preceding CLEARED, stream mode can't use the default tombstone reconciliation. If alarmStream and alarmSync are both on without alarmSyncClear, synchronization is disabled at startup with a warning rather than run incompletely.
  - KafkaAlarmDataSync.isReady() made null-safe so a disabled datasync doesn't fail on an incoming snapshot.
  - Docs (new "Alarm stream mode" section) and integration tests.

Stacks on NMS-19954 (alarmSyncClear) — the reconciliation strategy that clears orphaned alarms instead of tombstoning them, which is what makes stream mode safe. This PR targets that branch until NMS-19954 merges to release-36.x.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19955


dino2gnt:
I don't understand why we would want this.

marshallmassengill:
> I don't understand why we would want this.

I get the sentiment but it's mostly because it's a paradigm that seems to be adopted thanks to the likes of LogicMonitor and Splunk.  We've had several folks who expect it to work this way and there isn't a great reason we can't support it... we already do this with the Alarm History ES feature - this is just doing it for Kafka too.

cgorantla (approved):
LGTM!

PR #8620, opened 2026-07-06
NMS-19970: Tolerate unknown properties when reading CM config JSON

synqotik:
Tolerate unknown properties when reading CM config JSON.

This issue surfaced during NMS-19723 where we needed to add a field to `Snmpv3User` for Trap Configuration masked credentials. If the DB contains a newer configuration, then even starting up an older version of OpenNMS will fail when deserializing the config JSON via CM (Content Management), because it strictly fails on unknown properties.

This PR make a change to ignore unknown properties, so that a persisted config written by a newer version (with fields not yet known to the running code) can still be read. Without this, adding a field to a config class (e.g. `Snmpv3User.id`) breaks reading that config on any older version during a downgrade, rollback, or rolling upgrade.
    
Adds a regression test that reproduces the failure (`ConfigConversionException` caused by Jackson's `UnrecognizedPropertyException`).

*NOTE* that using a newer config with an older already-built build will still fail. You'll need to rebuild the older build with this fix first, but now the fix is available back to 2024.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19970


marshallmassengill (approved):
Good find.

PR #8623, opened 2026-07-09
NMS-19971: Foreign Key with ON DELETE SET NULL for node.nodeparentid

marshallmassengill:
node.nodeparentid (the path-outage/critical-path parent, OnmsNode.getParent()) had no foreign-key constraint, so deleting a parent node left its children pointing at a nonexistent row. Dereferencing the lazy parent proxy for such a dangling reference throws Hibernate ObjectNotFoundException, and no deletion path cleaned it up: DefaultProvisionService.deleteNode() removes only the node row, and vacuumd deletes type-'D' nodes via raw SQL that bypasses application logic entirely.

Fix at the database level so every deletion path is covered. New 36.0.3 changeset: null out already-dangling references, index node(nodeparentid), then add a self-referential FK node.nodeparentid -> node.nodeid with onDelete=SET NULL -- the pattern ipInterface already uses for snmpinterfaceid (snmpinterface_fkey2). Registered in the master changelog after 36.0.1.

Assisted by Anthropic Claude Fable 5

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19971

### DIFF
--- a/core/schema/src/main/liquibase/36.0.3/changelog.xml
+++ b/core/schema/src/main/liquibase/36.0.3/changelog.xml
@@ -1,0 +1,72 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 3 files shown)


cgorantla (dismissed):
LGTM!

marshallmassengill:
Closing due to a weird circleCI issue

PR #8627, opened 2026-07-13
NMS-20004: tolerate unrepresentable ipaddr values in filter results

marshallmassengill:
A single ipinterface row whose ipaddr value can't be converted to an InetAddress aborted the entire filter evaluation with an opaque UndeclaredThrowableException — and the default *.*.*.* rule matches every row, so one bad value could break every package filter evaluation for pollerd, collectd, threshd, and notifd. Trigger values include IPv6 zone IDs naming an interface that doesn't resolve locally (fe80::…%eth0), hostname-like strings (which also caused a DNS lookup per row), and empty strings (silently converted to loopback).

Fix: convert result rows tolerantly — accept only IP literals (no DNS resolution; numeric zones unchanged), fall back to the zone-stripped address for interface-name zones, and log-and-skip anything still unrepresentable instead of failing the evaluation. Unit tests cover each failing shape; JdbcFilterDaoIT passes unchanged.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20004

### DIFF
--- a/opennms-config/src/main/java/org/opennms/netmgt/filter/JdbcFilterDao.java
+++ b/opennms-config/src/main/java/org/opennms/netmgt/filter/JdbcFilterDao.java
@@ -243,7 +243,7 @@ public class JdbcFilterDao implements FilterDao, InitializingBean {
                 // Iterate through the result and build the array list
                 while (rset.next()) {
                     final Integer nodeId = rset.getInt(1);
-                    final InetAddress ipaddr = addr(rset.getString(2));
+                    final InetAddress ipaddr = toInetAddressOrNull(rset.getString(2));
                     final String serviceName = rset.getString(3);
                     if (ipaddr == null || serviceName == null) {
                         continue;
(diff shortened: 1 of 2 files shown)


dino2gnt:
Looks green to me.

dino2gnt (dismissed):

opennms-config/src/main/java/org/opennms/netmgt/filter/JdbcFilterDao.java
```diff
@@ -373,6 +376,55 @@ private List<InetAddress> getIPAddressList(final String rule, final boolean filt
...
+    static InetAddress toInetAddressOrNull(final String rawAddress) {
+        if (rawAddress == null) {
+            return null;
+        }
+        final String value = rawAddress.trim();
+        final int percent = value.indexOf('%');
+        final String addressPart = percent < 0 ? value : value.substring(0, percent);
+        final String zone = percent < 0 ? null : value.substring(percent + 1);
+        final boolean v6 = addressPart.indexOf(':') >= 0;
+        final boolean literal = v6
```
cgorantla: May be we can use `InetAddresses.isInetAddress()` from guava to validate inet address instead of our own regex.

cgorantla (dismissed):
LGTM!

opennms-config/src/main/java/org/opennms/netmgt/filter/JdbcFilterDao.java
marshallmassengill: Implemented this.  Good suggestion.

cgorantla (approved):
LGTM!

PR #8632, opened 2026-07-15
NMS-20008: thresholding-enabled parameter is ignored by collectd

marshallmassengill:
Collectd creates a thresholding session and runs threshold processing on every collection regardless, so setting thresholding-enabled="false" in collectd-configuration.xml silently does nothing.

Possible Fix: CollectableService skips thresholding session creation when the parameter is explicitly false. Default remains enabled, so configurations that don't set the parameter are unaffected.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20008


cgorantla (approved):
LGTM!

PR #8633, opened 2026-07-15
NMS-20009: Poller builds the metadata scope twice per poll

marshallmassengill:
PollerRequestBuilderImpl.execute() builds the node/interface/service metadata scope twice per poll. Once when interpolating the configured attributes and again when interpolating the monitor's runtime attributes.

Possible Fix: memoize the scope on the request builder so it is built at most once per poll request

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20009


cgorantla (approved):
LGTM!

PR #8635, opened 2026-07-15
NMS-20006: revert NMS-19951 local RPC TTL bound (PR #8604)

marshallmassengill:
Reverts the change from PR #8604 (NMS-19951) that bounded local (same-location) RPC execution.

This restores the code in core/ipc/rpc/camel/CamelRpcClientFactory.java byte-for-byte to its pre-PR state.

We still need to tackle how we want to actually address this but for now, we can revert the change.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20006


dino2gnt (approved):
When it goes green, go for it.

cgorantla (approved):
LGTM!

PR #8636, opened 2026-07-15
NMS-19981: PrimeVue migration: Replace Feather icons, completely remove Feather

synqotik:
Final major step in FeatherDS -> PrimeVue migration. This completely replaces Feather with either PrimeVue or just our own vendored components/styles.

See the Jira ticket for a few screenshots. For the most part, should not notice any difference from before.

After this PR, this epic should be complete, other than any remaining bugs/issues found during actual testing.

- Continue to use the Feather icons, since PrimeVue doesn't have full replacements for all of them, plus we'd like to remain vendor-neutral. Since we can freely use Feather, they have been imported into this project under `ui/src/components/icons`.

- Replace all uses of any Feather components, in some cases created `Onms` common components, in other cases just needed inline markup.

- Replaced all Feather styles, colors, CSS, mixins, types, etc. Replaced Feather grid with `onms-grid`. Replaced Feather `table` with `onms-table`.

- Add `onms-` SCSS token layer, rewire any `--feather-*` into `--onms-*`, convert Feather severity colors to `onms`

- Completely remove all `@featherds` packages

- Lots of fixes to get correct or improved behavior from PrimeVue or else vendored components/styles

- Various GeoMap fixes to ensure everything works properly, popups always display fully, things are aligned, Alarms/Nodes tables scroll properly, etc.

- Remove any dead code

- Updated docs as much as possible to remove Feather references. Some are kept in to note behaviors which we kept

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19981


synqotik:
Note, smoke tests passed in a previous build, latest build did not change any code.

synqotik:
Updates:
- added `ui/THIRD-PARTY-LICENSE.md` to add attribution for FeatherDS functionality that we "vendored".
- Map keyboard navigation restored, pan/scroll and zoom work with arrow and `+/-` keys
- Map: clicking `Escape` closes popup
- Map: popups now pop up *below* the marker. This fixes issue where map was being auto-scrolled to fix popup *above* the marker. Popup behavior is now a lot more natural. If popup is opened near the bottom of the screen, the map will scroll up to display it

marshallmassengill (dismissed):
This was the only real finding claude had.  I'm not really sure it's legit so I'm saying this is a LPTM! (Looks Plucked To Me).  Well done.

"Every legacy JSP page loses link styling, and a saved dark theme is never applied on JSP load. onms-base.scss:79-92 ships a global a { color: var(--onms-clickable-normal) } rule into the menu bundle CSS that bootstrap.jsp:203 loads on every JSP page. But that token (and the other 11 theme-varying tokens, including all severity colors) is defined only under .open-light / .open-dark in onms-theme.scss:49-77 — and src/menu/main.ts never calls initTheme(), and nothing in opennms-webapp sets that class (zero grep hits). The old menu entry imported Feather's open-light.css, which defined everything on :root, so tokens always resolved. Result: on node pages, alarm lists, admin pages — every anchor's color declaration is invalid-at-computed-value and falls back to inherited body text color instead of link blue; and PrimeVue's darkModeSelector: '.open-dark' never activates on JSP pages regardless of the user's saved preference (the menu's own pills survive only because Menubar.vue re-declares those tokens locally with literals). Fix: call initTheme() in the menu entry, or give the theme-varying tokens :root light-mode defaults."

synqotik:
> This was the only real finding claude had. I'm not really sure it's legit so I'm saying this is a LPTM! (Looks Plucked To Me). Well done.

Thanks!

> 
> "Every legacy JSP page loses link styling, and a saved dark theme is never applied on JSP load

Believe I just fixed most of this.

synqotik:
I would prefer to merge this and have a follow-up ticket for some of the issues found in reviews, along with other things I already know about.

marshallmassengill (approved):
Why did the chicken cross the road?
To get to the de-feather side.

PR #8638, opened 2026-07-16
NMS-20027: Compute flow proportional sums with Painless scripts

marshallmassengill:
Replace the proportional_sum aggregation provided by the elasticsearch-drift-plugin with an equivalent scripted_metric aggregation built from inline Painless scripts. This removes the need to install a plugin build that exactly matches the Elasticsearch version, and makes flow queries work against stock Elasticsearch and OpenSearch clusters.

Putting this in as a draft for now.  Would love feedback. 

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20027


features/flows/itests/src/test/java/org/opennms/netmgt/flows/elastic/FlowQueryIT.java
```diff
@@ -115,32 +114,18 @@ public class FlowQueryIT {
 
     // Elasticsearch version used for testing
     private static final String ES_VERSION = "8.18.2";
-    private static final String DRIFT_PLUGIN_VERSION = "2.0.7";
 
+    // Stock Elasticsearch, on purpose: the proportional sums used by the series/totals
+    // queries are computed with inline Painless scripts and must work without the
+    // elasticsearch-drift-plugin being installed.
```
dino2gnt, line 120: Drift is the reference implementation. Until these Painless scripts have been battle-tested at scale, I'd want to see the tests run both side by side for the newest supported version, and verify that we produce the same results in both cases.

indigo423:
I did an A/B test with the drift plugin vs. the painless script, persisting 1M flows in Elasticsearch. The performance gain is measurable and revealed also the existence of [NMS-20001] (see "Data Quality" section). Here is the [A/B painless vs. drift report](https://github.com/user-attachments/files/30274626/report.html).


[NMS-20001]: https://opennms.atlassian.net/browse/NMS-20001

features/flows/itests/src/test/java/org/opennms/netmgt/flows/elastic/FlowQueryIT.java
marshallmassengill: I've got this setup for tests now.  At least in theory.

marshallmassengill:
Latest commit came from some feedback from a particularly heavy flows customer.  Including a refinement to one of the scripts that should be useful for them.  Overall, I'm thinking this is looking pretty good but definitely provide more feedback if you have it or latest changes didn't address things that we still want to address.

marshallmassengill:
Should be an actual comparison IT now @dino2gnt.  Hopefully it even passes.

PR #8640, opened 2026-07-20
NMS-20045: Backend for new UI for Surveillance Views

marshallmassengill:
This issue covers the backend stuff:

- View definitions move to the database (the generic JSON key-value store, same approach as the custom topology views). A default surveillance view is created rather than doing an import (I did explore both and I think this is the cleaner option). 
- A new /api/v2/surveillance/views API provides CRUD over view definitions (writes require ROLE_ADMIN), per-user default-view resolution using the existing naming convention (username, then group name, then the configured default), and the computed data the UI needs: cell status ({id}/status) and the alarm/notification/RTC drill-downs ({id}/alarms, {id}/notifications, {id}/rtc).
- The Vaadin surveillance views remain in place and functional; the replacement Vue UI and the removal of the Vaadin feature are handled separately.

Trying to get us setup for Vaadin removal down the road.

Assisted by Anthropic Claude Fable 5

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20045

### DIFF
--- a/features/surveillance-views/src/main/java/org/opennms/netmgt/surveillance/views/impl/DefaultSurveillanceViewDataService.java
+++ b/features/surveillance-views/src/main/java/org/opennms/netmgt/surveillance/views/impl/DefaultSurveillanceViewDataService.java
@@ -1,0 +1,383 @@
+/*
+ * Licensed to The OpenNMS Group, Inc (TOG) under one or more
+ * contributor license agreements.  See the LICENSE.md file
+ * distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * TOG licenses this file to You under the GNU Affero General
+ * Public License Version 3 (the "License") or (at your option)
+ * any later version.  You may not use this file except in
+ * compliance with the License.  You may obtain a copy of the
+ * License at:
+ *
+ *      https://www.gnu.org/licenses/agpl-3.0.txt
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied.  See the License for the specific
+ * language governing permissions and limitations under the
+ * License.
+ */
+package org.opennms.netmgt.surveillance.views.impl;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.opennms.core.criteria.Alias;
+import org.opennms.core.criteria.Criteria;
+import org.opennms.core.criteria.CriteriaBuilder;
+import org.opennms.core.criteria.Order;
+import org.opennms.core.criteria.restrictions.Restriction;
+import org.opennms.core.criteria.restrictions.Restrictions;
+import org.opennms.core.criteria.restrictions.SqlRestriction.Type;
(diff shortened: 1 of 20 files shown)


features/surveillance-views/src/main/java/org/opennms/netmgt/surveillance/views/impl/DefaultSurveillanceViewDataService.java
```diff
@@ -0,0 +1,383 @@
...
+
+        stringBuffer.append(nodeIdProperty + " in (select distinct cn.nodeId from category_node cn join categories c on cn.categoryId = c.categoryId where c.categoryName in (");
+
+        String[] questionMarks = new String[rowCategoryCount];
+        Arrays.fill(questionMarks, "?");
+        stringBuffer.append(String.join(",", questionMarks));
+
+        stringBuffer.append("))");
+
+        stringBuffer.append("and " + nodeIdProperty + " in (select distinct cn.nodeId from category_node cn join categories c on cn.categoryId = c.categoryId where c.categoryName in (");
```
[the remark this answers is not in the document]
marshallmassengill: This is just outright wrong.  Go home copilot and think about your choices.

features/surveillance-views/src/main/java/org/opennms/netmgt/surveillance/views/impl/DefaultSurveillanceViewDataService.java
```diff
@@ -0,0 +1,383 @@
...
+/**
+ * Port of the aggregation logic from the retired Vaadin
+ * {@code DefaultSurveillanceViewService}: the category-intersection SQL, the
+ * notification severity bucketing, and the 24-hour RTC math are carried over
+ * unchanged. Everything runs inside {@link TransactionOperations} and returns
+ * detached value objects.
+ */
+public class DefaultSurveillanceViewDataService implements SurveillanceViewDataService {
+
+    private NodeDao m_nodeDao;
```
synqotik, line 68: Remove the `m_` prefixes.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/SurveillanceViewRestService.java
```diff
@@ -0,0 +1,419 @@
...
+
+    private static SurveillanceViewDTO.Def toDefDto(final SurveillanceViewDef def) {
+        final SurveillanceViewDTO.Def dto = new SurveillanceViewDTO.Def();
+        dto.setLabel(def.getLabel());
+        dto.setCategories(new ArrayList<>(def.getCategories()));
+        dto.setReportCategory(def.getReportCategory());
+        return dto;
+    }
+
+    private static void requireAdmin(final SecurityContext securityContext) {
```
synqotik, line 385: Just a note. `requireAdmin`, `usernameOf` and `webException` are good patterns. Probably shouldn't do this in this PR, but worth considering making a base class or helper class of these at some point.

synqotik:
LGTM! Probably need @cgorantla to also review.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/SurveillanceViewRestService.java
marshallmassengill: That does make sense.  I'll see if we can get something else open for that.

marshallmassengill:
@cgorantla - can you review this one as well?  I can push this to develop if needed.  Would like this in place so we can retire the vaadin surveillance views in H37.  Will also be useful for bringing back an "ops board" that is vue based.

PR #8641, opened 2026-07-20
NMS-20006: bound local RPC execution with a timeout decoupled from the TTL

marshallmassengill:
Attempting to fix the actual thing I was trying to fix in the first place with this one.  I've kept this tied to the same NMS as I do think we want to address this and just leaving it as the reversion isn't enough.  I'll open it as a draft for now to get comments on the approach.

This restores protection against the failure mode of a module future that never completes wedging its caller without re-introducing the regression: local execution in CamelRpcClientFactory is bounded at 30 minutes by default, configurable via the org.opennms.core.ipc.rpc.local.timeout system property (0 disables). 

The request TTL no longer affects local execution, so no TTL configuration can re-introduce the regression. RpcExceptionUtils now maps java.util.concurrent.TimeoutException to onTimedOut, so an expired local execution is reported as a timeout rather than an unexpected exception.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20006

### DIFF
--- a/core/ipc/rpc/api/src/main/java/org/opennms/core/rpc/api/RpcExceptionUtils.java
+++ b/core/ipc/rpc/api/src/main/java/org/opennms/core/rpc/api/RpcExceptionUtils.java
@@ -23,6 +23,7 @@ package org.opennms.core.rpc.api;
 
 import java.util.Objects;
 import java.util.Optional;
+import java.util.concurrent.TimeoutException;
 
 public class RpcExceptionUtils {
 
@@ -57,6 +58,8 @@ public class RpcExceptionUtils {
         } else if (t instanceof java.util.concurrent.RejectedExecutionException) {
             return Optional.ofNullable(visitor.onRejected(t));
         } else if (t instanceof RequestTimedOutException) {
+            return Optional.ofNullable(visitor.onTimedOut(t));
+        } else if (t instanceof TimeoutException) {
             return Optional.ofNullable(visitor.onTimedOut(t));
         } else {
             return null;
--- a/core/ipc/rpc/camel/src/test/java/org/opennms/core/rpc/camel/LocalRpcTimeoutTest.java
+++ b/core/ipc/rpc/camel/src/test/java/org/opennms/core/rpc/camel/LocalRpcTimeoutTest.java
@@ -1,0 +1,140 @@
+/*
+ * Licensed to The OpenNMS Group, Inc (TOG) under one or more
+ * contributor license agreements.  See the LICENSE.md file
+ * distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * TOG licenses this file to You under the GNU Affero General
+ * Public License Version 3 (the "License") or (at your option)
+ * any later version.  You may not use this file except in
+ * compliance with the License.  You may obtain a copy of the
+ * License at:
+ *
+ *      https://www.gnu.org/licenses/agpl-3.0.txt
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied.  See the License for the specific
+ * language governing permissions and limitations under the
+ * License.
+ */
+package org.opennms.core.rpc.camel;
+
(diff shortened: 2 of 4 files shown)


cgorantla (commented):
I think we are complicating this too much.

Our default mode is ActiveMQ(camel) and local RPCs will only run in Camel. This doesn't need to be part of kafka/grpc as this would never happen.  

It just need timeout bound to 30mins ( by default and configurable in CamelRpcClientFactory.java)

marshallmassengill:
Revised accordingly.  Thanks for the feedback @cgorantla

dino2gnt (commented):
As I said in chat, I don't think this fixes any real problem. 🤷

core/ipc/rpc/api/src/test/java/org/opennms/core/rpc/api/RpcExceptionUtilsTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+import org.junit.Test;
+
+/**
+ * Exception-to-callback mapping of {@link RpcExceptionUtils}. Timeouts must reach
+ * {@code onTimedOut} — a bare {@link TimeoutException} falling through to
+ * {@code onUnknown} is how the NMS-20006 outage was mislabeled as
+ * "Unexpected exception".
+ */
+public class RpcExceptionUtilsTest {
```
cgorantla: you don't probably need these tets

core/ipc/rpc/camel/src/test/java/org/opennms/core/rpc/camel/LocalRpcTimeoutTest.java
```diff
@@ -0,0 +1,157 @@
...
+/**
+ * NMS-20006 — local (same-location) RPC execution is bounded by a fixed,
+ * configurable timeout instead of the request TTL.
+ *
+ * <p>Slow-but-completing operations return their result even when the TTL (from
+ * snmp-config ttl, the collectd ttl service parameter, or the interval default)
+ * has long expired; a module future that never completes fails at the local
+ * timeout instead of wedging the caller (cf. NMS-19951).</p>
+ */
+public class LocalRpcTimeoutTest {
```
cgorantla, line 49: I think we only need one test that validates custom timeout like 5 ms ( and see that test execute returns properly, may be with awaitility)

cgorantla (commented):
Looks fine to me. It just needs one test that tests custom timeout.

core/ipc/rpc/camel/src/test/java/org/opennms/core/rpc/camel/LocalRpcTimeoutTest.java
marshallmassengill: Got this one.

core/ipc/rpc/api/src/test/java/org/opennms/core/rpc/api/RpcExceptionUtilsTest.java
marshallmassengill: I want to keep one of these so we don't end up with the regression again...

core/ipc/rpc/api/src/main/java/org/opennms/core/rpc/api/RpcExceptionUtils.java
```diff
@@ -58,6 +58,8 @@ private static <T> Optional<T> handleExceptionInternal(Throwable t, RpcException
             return Optional.ofNullable(visitor.onRejected(t));
         } else if (t instanceof RequestTimedOutException) {
             return Optional.ofNullable(visitor.onTimedOut(t));
+        } else if (t instanceof java.util.concurrent.TimeoutException) {
```
cgorantla: nit: import java.util.concurrent.

cgorantla (dismissed):
LGTM!

core/ipc/rpc/api/src/main/java/org/opennms/core/rpc/api/RpcExceptionUtils.java
marshallmassengill: Resolved.

cgorantla (approved):
LGTM!

PR #8646, opened 2026-07-21
NMS-20046: Handle Poller/Collect RPC executor shutdown

cgorantla:
Daemon shutdown can take several minutes because orderly executor shutdown waits for active RPC tasks to complete. Since these RPCs may have long timeouts, they should be interrupted during shutdown.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20046


christianpape (approved):
LGTM.

marshallmassengill (approved):
LGTM!

PR #8649, opened 2026-07-21
NMS-20049: harden smoke test login against CSRF/session race

marshallmassengill:
Tired of errors... trying to fix.

Fix in AbstractOpenNMSSeleniumHelper: park on about:blank after cookie deletion, retry the login if it lands on accessDenied.jsp, and make skipPasswordGate() a no-op when not on the gate page.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20049


synqotik (approved):
LGTM! Great catch!

PR #8657, opened 2026-07-23
NMS-20033: UI styling, behavior and pr review issues (post-PrimeVue migration)

synqotik:
# NMS-20033: UI styling & behavior fixes (post-PrimeVue migration)

Follow-up fixes and polish from PR review of the FeatherDS → PrimeVue migration. This branch introduces a small reusable seam component, migrates all icon-only buttons onto it, and resolves a batch of styling and behavior issues across the Vue SPA (icons, buttons, inputs, spacing, the geographical map, resource graphs, and dark mode).

## Highlights

- **New `OnmsIconButton` seam component** and migration of every icon-only button to it.
- Consistent **icon sizing** (icon buttons and search-field glyphs).
- **Form-control consistency**: outlined Cancel buttons, a single global input height, `FormField` adoption.
- **Geographical map** fixes (popup clipping, control positioning, layers dropdown, Esc handling, keep-alive cleanup, resize reflow).
- **Resource Graphs** reliability fixes (Chart.js lifecycle, "Graph All" race, reactive seeding).
- **Dark-mode footer** background.

---

### `OnmsIconButton` seam component + migration
- Added `components/Common/OnmsIconButton.vue` — a thin wrapper around a PrimeVue `Button` + the vendored `OnmsIcon`, part of the `onms-` seam layer that insulates the app from PrimeVue. It bakes in consistent icon sizing (default `1.5rem`, overridable via `iconSize`), forwards `title` as the icon's accessible name **and** as the button's native tooltip, and lets all other attributes/`Button` props fall through. Unit tests added.
- Migrated **~65 icon-only buttons across ~33 files** (Configuration, EventConfiguration, Nodes, SnmpDataCollection, Trapd, Logs, Device, SCV, ZenithConnect, FileEditor) from the raw `<Button><OnmsIcon/></Button>` pattern to `OnmsIconButton`.
  - Normalized icon size to `1.5rem` (fixing under- and over-sized glyphs) and removed the scattered per-file `font-size` overrides; `FileSidebar` keeps its intentionally larger `2em` toolbar icons via the `iconSize` prop.
  - Added **accessible names** to 8 previously-unlabeled icon buttons.
  - Refactored the `ScvInputIcon` wrapper to build on `OnmsIconButton` while preserving its public API and circular-badge look.
  - Removed now-dead CSS classes left behind after the migration.

### Buttons
- **Cancel buttons**: added the `outlined` prop to every text-variant Cancel button so they read as bordered.
- **Outlined + text border fix**: a `Button` with both `outlined` and `text` rendered no visible border (PrimeVue's `p-button-text` sets a transparent border-color that won over the outlined border). Added a global rule so the outline shows for that combination.
- **DCB action buttons**: normalized the mixed-source action-button icons (`View History` / `Download` / `Backup` / `Compare`) to a single `1.5em` size and vertically centered icon + label (the `Compare` asset had hardcoded `24px` + a `-10px` margin hack that broke alignment).

### Inputs & form controls
- **Search-field icons**: the IconField magnifying-glass glyph (an `OnmsIcon`) was inconsistently sized (too small on most screens, `1.75rem` on a few). Normalized to `1.5rem` via one global `.p-inputicon` rule and removed the per-screen overrides.
- **Global input height**: promoted `height: 3rem` to the global `.p-inputtext` / `.p-select` rule (was `3.75rem`, with `FormField` overriding to `3rem`). ~92% of controls were already `3rem` via `FormField`; this brings the remaining non-`FormField` controls into line and retires the vestigial `3.75rem` default. `FormField`'s pilot height override was removed.
- **`FormField` adoption**: wrapped the "Add Profile" and "Add Source" autocompletes in `FormField` for label/structure consistency.

### Spacing
- Restored spacing lost to **orphaned FeatherDS spacing utility classes** (`mb-m`, `mr-m`, `pb-xl`) — the last unresolved Feather utilities — by defining local scoped equivalents backed by the `--onms-spacing-*` tokens (verified equal to the original FeatherDS values). Fixes the External Requisitions form rendering its fields flush.

### Geographical map (Leaflet)
- **Popup clipping**: extended the `popupopen` overflow handling to all edges (left/right/top/bottom), so popups on edge-most markers no longer render half outside the map.
- **Control positioning**: gave `.geo-map` a positioning context so the Search / Show Severity overlays anchor to the map area (not the viewport, where they slid under the pinned side-menu rail); reduced the Search / Severity / nav-control top inset from `80px` to `2em`.
- **Layers control**: bridged the hover dead-gap that let the layers dropdown collapse before you reached it, and made the dropdown theme-aware (was hardcoded white in dark mode).
- **Esc handling**: the map-search AutoComplete's Esc no longer also closes an open marker popup.
- **keep-alive cleanup**: the document Esc listener is now removed on `deactivated` (the map is `<keep-alive>`'d and never unmounts).
- **Resize reflow**: a `ResizeObserver` calls `invalidateSize()` so the map reflows when the side menu is pinned (or on any container resize).

### Resource Graphs
- **Chart.js lifecycle**: destroy chart instances before reusing a canvas and on unmount — fixes the console error `Canvas is already in use…` and graphs silently failing to render.
- **"Graph All" race**: await the graph-definitions fetch before navigating, so the graphs view no longer mounts before its data is ready (graphs failing to appear on the first try).
- **Reactive seeding**: the graphs view now seeds its initial graphs reactively from the store rather than a setup-time snapshot, so the first render depends on data availability rather than mount timing.
- **Tabs**: uppercased the `Graph` / `Data` tab labels.

### Dark mode
- **Footer**: the footer hardcoded a light background and stayed light in dark mode on Vue SPA pages. It now uses `var(--p-content-background)` (and a theme-aware top border).

---

Most changes are CSS/behavior on screens without unit-test coverage; these were verified visually in the running app (light and dark themes).

## Notes for reviewers

- `OnmsIconButton` is the recommended pattern for future icon-only buttons.
- `text` + `outlined` on a Button is a slightly contradictory combination we chose to keep for now (the global border rule makes it render correctly); revisit if/when Cancel buttons move to a dedicated seam/wrapper component.
- `SnmpDataCollectionSourceDetail`'s "Source" field was intentionally left as an inline detail row (not wrapped in `FormField`) so it stays consistent with its sibling read-only fields; the global 3rem height already applies to it.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20033

### DIFF
--- a/opennms-webapp/src/main/webapp/WEB-INF/jsp/admin/userGroupView/groups/list.jsp
+++ b/opennms-webapp/src/main/webapp/WEB-INF/jsp/admin/userGroupView/groups/list.jsp
@@ -119,7 +119,7 @@
                 <a id="${group.name}.doDelete" href="javascript:deleteGroup('${fn:escapeXml(group.name)}')" onclick="return confirm('Are you sure you want to delete the group ${fn:escapeXml(group.name)}?')"><i class="fas fa-trash-can fa-2x"></i></a>
               </c:when>
               <c:otherwise>
-                <i class="fas fa-trash-can fa-2x" style="color: var(--feather-clickable-normal);" onclick="alert('Sorry, the ${fn:escapeXml(group.name)} group cannot be deleted.')"></i>
+                <i class="fas fa-trash-can fa-2x" onclick="alert('Sorry, the ${fn:escapeXml(group.name)} group cannot be deleted.')"></i>
               </c:otherwise>
             </c:choose>
           </td>
--- a/opennms-webapp/src/main/webapp/admin/userGroupView/users/list.jsp
+++ b/opennms-webapp/src/main/webapp/admin/userGroupView/users/list.jsp
@@ -159,7 +159,7 @@
             <a id="<%= "users("+sanitizedUserId+").doDelete" %>" href="javascript:deleteUser('<%=sanitizedUserId%>')" onclick="return confirm('Are you sure you want to delete the user <%=sanitizedUserId%>?')"><i class="fas fa-trash-can fa-2x"></i></a>
           </td>
           <% } else { %>
-          <td rowspan="2" class="text-center" style="color: var(--feather-clickable-normal);">
+          <td rowspan="2" class="text-center">
             <i class="fas fa-trash-can fa-2x" onclick="alert('Sorry, the admin user cannot be deleted.')"></i>
           </td>
           <% } %>
--- a/ui/src/components/Common/FormField.vue
+++ b/ui/src/components/Common/FormField.vue
@@ -86,17 +86,11 @@ watch(errorId, () => nextTick(syncAriaDescribedby))
   display: flex;
   flex-direction: column;
 
-  // Pilot-only: shorter controls inside FormField. Once every screen is
-  // converted off IftaLabel/FloatLabel, promote 3rem to the global
-  // .p-inputtext / .p-select rule in primevue-overrides.scss and delete
-  // this block.
-  :deep(.p-inputtext),
-  :deep(.p-select) {
-    height: 3rem;
-  }
-
-  // MultiSelect grows with its chip display, so normalize via min-height to
-  // match the other controls when empty while still allowing it to grow.
+  // Input/Select height (3rem) is now the global default (see
+  // primevue-overrides.scss .p-inputtext / .p-select), so no per-field height
+  // override is needed here. MultiSelect grows with its chip display, so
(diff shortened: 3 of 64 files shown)


marshallmassengill (approved):
LGTM!  I think there is some minor stuff around hover/active and outlined + text borders but it can't wait for you to get to the next stuff.  Looks like Graphs.vue may also need seeding.

PR #8659, opened 2026-07-24
NMS-20055: Fix rest endpoints that expose protected fields

cgorantla:
Protected fields (primary keys, category authorizedGroups, node foreignSource/foreignId/type) could be set via a blind BeanWrapper copy. Adds an allow-list guard so those are ignored on update (still settable on create)

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20055

### DIFF
--- a/opennms-web-api/src/main/java/org/opennms/web/api/RestUtils.java
+++ b/opennms-web-api/src/main/java/org/opennms/web/api/RestUtils.java
@@ -30,7 +30,11 @@ package org.opennms.web.api;
 
 import java.beans.PropertyEditor;
 import java.net.InetAddress;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.Date;
+import java.util.HashSet;
+import java.util.Set;
 
 import javax.ws.rs.core.MultivaluedMap;
 import javax.xml.datatype.XMLGregorianCalendar;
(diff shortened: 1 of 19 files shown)


marshallmassengill (changes requested):
Couple findings:

doUpdateProperties does params.remove("foreignSource"/"foreignId"/"type") on literal keys, but setBeanProperties resolves each key through convertNameToPropertyName (normalizes _/-, lower-cases). So foreign_source, foreign-source, foreign_id, and Type survive the remove, get normalized to the real property, and are written. Those three aren't in IMMUTABLE_PROPERTIES. A ROLE_REST PUT /api/v2/nodes/1 with foreign_source=X still does the requisition takeover the PR claims to block. 

Possible Fix: filter on convertNameToPropertyName(key), or move the check inside setBeanProperties (create doesn't use that path).

Medium — updateMany aborts the batch on a non-error 304. IfServiceRestService/NodeMonitoredServiceRestService return notModified() (304, family REDIRECTION) when nothing changed. That fails the != SUCCESSFUL test, so the first unchanged item stops the loop, leaves the rest un-updated, and returns 304 instead of 204. Also: services with the default doUpdateProperties now return 501 instead of the old silent 204 — a deliberate-or-not behavior change worth noting. 

Possible Fix: only stop on CLIENT_ERROR/SERVER_ERROR.

There also doesn't appear to be a test for the V2 path above at all so that is probably worth adding.

christianpape (dismissed):
LGTM.

marshallmassengill (approved):
I think this is looking good now.

christianpape (approved):
LGTM.

PR #8660, opened 2026-07-24
NMS-19096: Refresh NodeScan scan stamp on each scheduled run

marshallmassengill:
A NodeScan scheduled with scheduleWithFixedDelay is reused for every periodic run, but m_scanStamp was set once in the constructor and never advanced.  This fixes that behavior and brings the behavior in line with the force rescan behavior for removing interfaces.  This has actually been latent since NMS-112 (~2010).

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19096
* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-112


dino2gnt:
Wow.

PR #8664, opened 2026-07-24
Docs: Updates for H36 for RHEL 10 directions and Java 21

marshallmassengill:
Doc updates for 36 addressing install on RHEL 10 and Java 21.


mershad-manesh (dismissed):

mershad-manesh:
The changes looks good; lets get feedback from @indigo423  and @dino2gnt  about these changes

dino2gnt (dismissed):
No real feedback except that I would have been fine with leaving the Adoptium bits in there as the repo for JDK 21, just like we always point to PGDG for PostgreSQL.

But it's whatever.

marshallmassengill:
> No real feedback except that I would have been fine with leaving the Adoptium bits in there as the repo for JDK 21, just like we always point to PGDG for PostgreSQL.
> 
> But it's whatever.

Any reason why not to use the RHEL packages?

dino2gnt:
No, except in that case we rely on the ephemeral whims of IBM / Redhat / CentOS / Rocky / Alma maintainers keeping JDK21 in their repos for the accuracy of our documentation.

marshallmassengill:
Ephemeral whims is the name of my post-new wave classic punk jazz band.

docs/modules/deployment/pages/core/centos-rhel10/install-core.adoc
```diff
@@ -1,7 +1,3 @@
-ifeval::["{page-component-title}" == "Horizon"]
-include::../centos/adoptium.adoc[]
```
indigo423, line 2: This here is something to think about a bit more. The "always adoptium" for JDK 21 is here because, it isn't available from all default packages in various distros we support.

If you take that away you need now to think about all other distros where you have to use it or not, this will come back particularly with JDK 25 which is the next LTS released soon.

If you just keep adoptium you don't care anymore, if you remove it from some distros, you have to deal now wit distro specific variations.

I'm not against removing it but I want to make sure this is considered in your removal.

indigo423 (dismissed):
[-0](https://www.apache.org/foundation/voting.html#expressing-votes-1-0-1-and-fractions) on the adoptium one.

marshallmassengill:
If we do that then I think we should make the adoptium thing sit above the specific version install directions and just point people at it as a place to get the JDK/JRE.  I'm good with that I think.

marshallmassengill:
I've added a section for Java install and linked to adoptium.  I think this is cleaner and removes us from having to maintain separate steps for each OS version.  Feedback welcomed.

dino2gnt (approved):
👍

indigo423 (approved):
go for it.

PR #8669, opened 2026-07-27
NMS-20079: Remove the checked-in GWT sandbox from topology-map plugins

marshallmassengill:
The sandbox directory was an Eclipse-only GWT terminal-widget experiment added in 2012 (c87937f682b) ~58 MB of tracked blobs. Its only reference was the license-check exclusion in the root pom.

Cleaning assisted by Anthropic Claude Opus 5.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20079


synqotik (approved):
LGTM! Great find, nice to delete all these dead files.

dino2gnt (approved):
👍

PR #8670, opened 2026-07-27
NMS-19975: Fix flicker when navigating to JSP pages

synqotik:
# NMS-19975: Fix screen flicker when navigating to JSP pages

## Problem

Navigating **to** any JSP page (from a Vue SPA page or from another JSP page) caused a visible flicker: a light/white background painted first — including where the dark top menu bar and side rail belong — and the dark menu chrome only appeared a moment later.

## Root cause

JSP navigations are full document loads (unlike Vue SPA routing, which is client-side). The server-rendered HTML contains only an empty `<div id="opennms-sidemenu-container">` plus a `<script type="module">` tag for the menu bundle. On first paint nothing dark exists where the fixed menubar and side rail belong, so the light page background flashes through on every JSP navigation. The dark chrome appears only after the menu bundle is fetched, parsed, and executed and Vue mounts.

A related latent bug: the standalone menu bundle never called `initTheme()` (only the SPA did), so the persisted `open-light`/`open-dark` theme class was never applied to `<html>`/`<body>` on JSP page loads — only the in-page theme toggle ever applied it. As a result, the `.open-dark #opennms-sidemenu-container` token overrides silently didn't apply on freshly loaded JSP pages in dark mode.

## Changes

**`ui/src/styles/opennms-styles.scss`** — Among other things, ensures dark background/chrome before re-render.

**`opennms-webapp/src/main/webapp/includes/bootstrap.jsp`** —
- Inline `<head>` script that applies the persisted `open-light`/`open-dark` class to `<html>` before first paint (storage key and class names mirror `ui/src/services/themeService.ts`; falls back to light if `localStorage` is unavailable).
- `<link rel="modulepreload">` for the menu bundle, so fetch/compile starts during head parsing instead of when the parser reaches the script tag near the end of the body.

**`ui/src/menu/main.ts`** — the standalone menu app now calls `useAppStore().initTheme()` before mount, mirroring the SPA (`ui/src/main/main.ts`). This fixes the latent theme-class bug above and re-asserts the class on `<body>` after the JSP inline script has set it on `<html>`.

## Out of scope (deliberately deferred)

These were deferred as they are bigger changes with minimal improvements. Main issue to fix in future would be some minor redraw differences between Vue SPA top menu and the embedded Vue menu on JSP pages, e.g. search box placement is slightly different.

- Cache-header changes for `/ui-components/assets/*` (currently `Cache-Control: no-cache`).
- Server-side inlining of the menu JSON to avoid the post-mount `/rest/menu` pop-in.
- Cross-document View Transitions polish.
- Residual minor redraw differences as the real menu replaces the skeleton.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19975

### DIFF
--- a/opennms-webapp/src/main/webapp/includes/bootstrap.jsp
+++ b/opennms-webapp/src/main/webapp/includes/bootstrap.jsp
@@ -200,7 +200,32 @@
   </c:if>
 
   <%-- Vue side menu --%>
+  <%--
+    Apply the persisted UI theme class to <html> before first paint so
+    theme-dependent styles (PrimeVue's .open-dark darkModeSelector, the menu
+    token overrides in the Vue menu's stylesheet) are correct from the very
+    first frame instead of flipping when the menu bundle mounts (NMS-19975).
+    Storage key and class names mirror ui/src/services/themeService.ts; the
+    menu app re-asserts the class (and adds it to <body>) via initTheme().
+  --%>
+  <script type="text/javascript">
+    (function () {
+      var theme = 'open-light';
+      try {
+        var stored = window.localStorage.getItem('theme');
+        if (stored === 'open-light' || stored === 'open-dark') {
+          theme = stored;
+        }
+      } catch (e) {
+        // localStorage unavailable; fall through to the default theme
+      }
+      document.documentElement.classList.add(theme);
+    })();
+  </script>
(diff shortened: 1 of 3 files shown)


marshallmassengill (approved):
LGTM!

PR #8671, opened 2026-07-27
NMS-20029: UI seam layer, Phase 1

synqotik:
# NMS-20029: Introduce `@opennms/onms-ui` seam layer (tranche 1)

First tranche of implementing a "seam" UI component wrapper layer.

The seam insulates OpenNMS (and, in a later phase, plugin UIs) from third-party UI framework churn: app code depends on a stable, OpenNMS-owned API instead of PrimeVue (or another library) directly, so the underlying framework can be replaced without rewriting consumers. See `ui/packages/onms-ui/README.md` for the seam rules.

This is also part of a strategy to expose this to plugins. The way this is implemented, plugins can use the `OnmsUI` layer as if it were a 3rd party library, but all the code and interface is in the OpenNMS code repo instead of a separate repo like FeatherDS or `opennms-js`, which makes it much easier for OpenNMS development.

Note that the seam layer does not contain any actual implementation, it's basically an interface containing all the things we would need in a component, then delegating to the underlying component library for implementation.

## What

Adds a new pnpm workspace package, `ui/packages/onms-ui` (`@opennms/onms-ui`), containing owned `Onms-` wrapper components around PrimeVue, and converts all core UI usage of the wrapped modules to go through them. Direct imports of the wrapped PrimeVue modules are now banned in `ui/src` via ESLint `no-restricted-imports`.

Components in this tranche: `OnmsButton`, `OnmsIconButton`, `OnmsIcon` (moved from `ui/src`), `OnmsInputText`, `OnmsTextarea`, `OnmsPassword`, `OnmsCheckbox`, `OnmsSelect`, `OnmsAutoComplete`, `OnmsTag`, `OnmsDialog`, `OnmsSpinner`, and `OnmsToastHost` + `useOnmsToast` (with `useSnackbar` kept as a thin adapter — its ~180 call sites are unchanged).

## Design notes

- **Strict API + escape hatch**: each wrapper declares explicit props/slots/emits; DOM attrs (`class`, `data-*`, `aria-*`, native events) fall through; `unsafePt` ("unsafe passthrough") exists as a documented, unsupported escape hatch to pass framework-specific props.
- **Workspace source, no build step yet**: core compiles the package sources directly (`workspace:*`, peer-deps on `vue`/`primevue`). Lib build / npm publishing arrives with the plugin-sharing phase.
- New `variant="ghost"` on the button wrappers = PrimeVue `text`+`outlined` (the bordered Cancel style our `primevue-overrides.scss` already targets).
- Each wrapper ships a contract test in `ui/tests/onms-ui/` (prop mapping, emit forwarding, slot forwarding). Note that these tests do *not* test the actual UI functionality/implementation.

## Intentional behavior changes (everything else is 1:1)

- 18 `v-model.trim` inputs now actually trim — PrimeVue silently ignores the modifier; the wrapper implements it.
- Removed a duplicate toast outlet in `ProvisionDConfig.vue` that double-rendered every toast on that page.

## Out of scope (follow-ups)

DataTable/Column, Tabs, Menu, Drawer, Chip, ToggleSwitch, InputNumber (tranche 2); theme-token relocation, lib build, npm publish, and runtime exposure for plugins.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20029


marshallmassengill (approved):
LGTM!

Some minor things claude caught that are worth a look:

  - Trim never re-syncs the visible input. OnmsInputText emits the trimmed value, but when the trimmed value equals the current model (e.g. user types a trailing space), no re-render occurs and the DOM keeps "foo " while the model holds "foo". Native
  v-model.trim force-resets the element on change; the wrapper doesn't. Cosmetic, but the model and screen can disagree at submit time.
  - Duplicate-toast suppression outlives dismissal: activeKeys clears on the life timeout, not on close, so re-firing a message the user just dismissed is silently swallowed for up to 4s. (Pre-existing behavior faithfully moved, but this was the moment
  to fix it.)
  - ONMS_UI_VERSION = '0.1.0' hardcoded in index.ts duplicates package.json — guaranteed drift, and a test asserts it, so it will drift loudly at the least useful moment.
  - "license": "AGPL v3" isn't a valid SPDX expression (AGPL-3.0-or-later). Irrelevant while private, wrong the day this hits npm — which is the stated plan.
  - The as OnmsToastSeverity cast in useSnackbar compiles only by luck of string equality between MessageSeverity and the seam type; a mapping (or deriving one from the other) would fail at compile time if either side changes.

synqotik:
> LGTM!
> 
> Some minor things claude caught that are worth a look:

Thanks, I will address these as part of Phase 2.

PR #8672, opened 2026-07-27
NMS-20080: Remove vendored legacy OpenLayers and Leaflet from web-assets

marshallmassengill:
OpenLayers 2.10 was compiled into the shipped leaflet-js bundle via a require() but no code calls its API.

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20080


dino2gnt (approved):
Yank it alllll out.

synqotik (approved):
Nice!!

PR #8674, opened 2026-07-27
NMS-20082: Remove sample REST dumps served as webapp static content

marshallmassengill:
opennms-webapp/src/main/webapp/xml/ held two checked-in sample REST responses (all-events.xml, nodes-limit-20.xml) referenced by nothing and served publicly by the webapp.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20082


dino2gnt (approved):
👍

cgorantla (approved):
LGTM!

PR #8675, opened 2026-07-27
NMS-20083: Remove dead build and release scripts

marshallmassengill:
None of these are referenced by anything in the repository; current CI runs from .circleci. The nightly-*.sh scripts could not run at all: they abort looking for buildtool.pl, which does not exist in the repo. bin/bamboo.pl dates to 2013 (the -Pbuild-bamboo Maven profile and bamboo_* env fallbacks in makedeb.sh/makerpm.sh are unrelated and untouched). src/site/site.xml (2009) was referenced only by its own license-check exclusion, removed with it. debian/opennms-contrib.dirs belonged to a package debian/control no longer declares.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20083


dino2gnt (dismissed):
👍

marshallmassengill:
Changing to Develop from a discussion with @mershad-manesh

runtests.sh
mershad-manesh, line 1: I'm ok with removing this.. but just for reference this script is generally used to simulate what CircleCI does for running test cases

marshallmassengill: I'll update and put it back and put a dated comment for that.  Suspect that will cut down on anything flagging it in the future.

marshallmassengill: Done.

mershad-manesh (approved):
I believe the changes look ok.

PR #8683, opened 2026-07-28
NMS-20092: Remove EOL BouncyCastle jdk15on and SAML deps

marshallmassengill:
Removes the EOL BouncyCastle jdk15on provider and orphaned SAML-era dependencies from the spring-security aggregator.

- bcprov-ext/bcprov/bcpkix-jdk15on 1.70 shipped a second, EOL crypto provider alongside the jdk18on 1.84 the rest of the build uses.
- esapi and xmlsec lost their last consumer when spring-security-saml2-core was dropped; their orphaned root-pom property/dependencyManagement entries are removed too.
- Dropping esapi also removes its transitive baggage (antisamy, bcutil-jdk15on, neko-htmlunit, xom, httpclient5/httpcore5).
- Login and authenticated REST verified on a local build; the LDAP/Kerberos artifacts are untouched.

We probably do want to actually add proper SAML auth at some point but I suspect removing these won't impact that.  Claude put a comment/hint here for that future case.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20092

### DIFF
--- a/dependencies/spring-security/pom.xml
+++ b/dependencies/spring-security/pom.xml
@@ -39,27 +39,9 @@
     </dependency>
     <!-- spring-security-openid + openid4java dropped: the OpenID 1.0 module was removed in Spring
          Security 5.x and was unwired in OpenNMS. spring-security-saml2-core (EOL 1.0.x extension,
-         incompatible with 5.x) likewise dropped; reintroduce via spring-security-saml2-service-provider
-         if SAML SSO is needed. -->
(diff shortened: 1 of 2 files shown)


marshallmassengill:
Need to get https://github.com/OpenNMS/opennms/pull/8684 in first and rebase before this one is going to pass.

cgorantla (approved):
LGTM!

PR #8684, opened 2026-07-28
NMS-20093: Restore xstream and kotlin-stdlib dependencyManagement pins

marshallmassengill:
NMS-20085 removed these as unused, but they were pinning the versions of transitive dependencies that we ship.

- xstream 1.4.21 -> 1.4.20 (via drools-xml-support 1.4.21 was a CVE-motivated bump)
- kotlin-stdlib/-common 1.9.10 -> 1.4.10 (via okhttp). 

The restored entries carry comments naming the transitive consumer they govern.

This regression fix brought to you by the letter R and assisted by Anthropic Claude Opus 5. 

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20093

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4028,6 +4028,13 @@
         <artifactId>caffeine</artifactId>
         <version>${caffeineVersion}</version>
       </dependency>
+      <!-- no direct users, but pins the xstream shipped in lib/ (transitive of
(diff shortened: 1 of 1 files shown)


cgorantla (approved):
LGTM!

PR #8639, opened 2026-07-17
NMS-20032: implement a postgresql-based Flows feature

dino2gnt:
This implements a PostgreSQL-based `flows` feature, mimicking the functionality of and implementing the same features as `features/flows/elastic`, including a `PostgresFlowRepository` and `PostgresFlowQueryService`.  The `PostgresFlowQueryService` bakes in the proportional sum aggregation currently packaged for Elasticseach as Drift.

It supports both "internal" and "external" databases and the feature is loadable on Sentinel, targeting an external database for offload.

The table schema is fairly simple and uses PostgreSQL native partitioning:
```
                        Partitioned table "public.flow"
      Column       |           Type           | Collation | Nullable | Default
-------------------+--------------------------+-----------+----------+---------
 flow_ts           | timestamp with time zone |           | not null |
 received_at       | timestamp with time zone |           | not null | now()
 delta_switched    | bigint                   |           | not null |
 last_switched     | bigint                   |           | not null |
 first_switched    | bigint                   |           |          |
 bytes             | bigint                   |           |          |
 packets           | bigint                   |           |          |
 sampling_interval | double precision         |           |          |
 direction         | text                     |           |          |
 application       | text                     |           |          |
 convo_key         | text                     |           |          |
 src_addr          | inet                     |           |          |
 dst_addr          | inet                     |           |          |
 protocol          | smallint                 |           |          |
 dscp              | smallint                 |           |          |
 exporter_node_id  | integer                  |           |          |
 input_snmp        | integer                  |           |          |
 output_snmp       | integer                  |           |          |
 location          | text                     |           |          |
 document          | jsonb                    |           | not null |
Partition key: RANGE (flow_ts)
Indexes:
    "flow_application_idx" btree (application)
    "flow_convo_key_idx" btree (convo_key)
    "flow_document_gin_idx" gin (document)
    "flow_exporter_idx" btree (exporter_node_id)
    "flow_ts_idx" btree (flow_ts)
```

On a Core instance, partition creation and expiration is managed by `vacuumd`.  On an external database, `pg_partman` and `pg_cron` are recommended.

In my testing, I was able to scale this to ~15,000 flow documents a second with only the tunables suggested by pgtune.leopard.in.ua.

Testing environment consisted of:
10c / 16G PostgreSQL 18:
```
# from pgtune for OLTP
max_connections = 100
shared_buffers = 4GB
effective_cache_size = 12GB
maintenance_work_mem = 1GB
checkpoint_completion_target = 0.9
[...]

PR #8685, opened 2026-07-28
NMS-20094: bring flow aggregation feature into OpenNMS

dino2gnt:
Since there is talk of deprecating Nephron and archiving the repository, bring the Java core of the flow aggregation into OpenNMS itself so that flow processing features can leverage it directly.

I didn't write any of this.  It is a copy of the Nephron code massaged into its own package by Claude.  I have verified it compiles, loads on Core, loads on Sentinel.  I have tested this with the PostgreSQL flow backend and verified the aggregated totals match the expected values, and had Claude write tests that do the same.  It is verified functional to the maximum degree I am able to do so; I have a test environment up running this code processing flows from a modified `udpgen`.

This was originally part of one big, ugly PR for the PostgreSQL flows. (see references)

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20094
* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20032
* PostgreSQL Flow PR: https://github.com/OpenNMS/opennms/pull/8639


marshallmassengill (dismissed):
I'm good with this.  I think it opens up the door for us to do this on sentinels (and the core), which makes sense to me.

marshallmassengill:
@dino2gnt - some comments from a claude review of this that are probably worth addressing:

Correctness

  - Sink failure loses data, log claims retry. Windows are evicted before sink.accept(); if the sink throws, the batch is gone but flushLoop logs "will retry on the next tick."
  - One bad timestamp poisons everything. The global watermark takes unvalidated lastSwitched; a far-future timestamp makes all subsequent flows "late" and silently dropped, forever. Also deltaSwitched=0 makes add() iterate millions of window numbers. No timestamp sanity clamps anywhere.
  - add() races the flusher. The lateness check and computeIfAbsent aren't atomic with window eviction — bytes can land in an already-emitted map and vanish, uncounted.
  - Re-emission is undefined. A late flow after an idle flush recreates and re-emits the window; whether duplicate (window, key, writer) rows double-count or blow a constraint is left to the backend, and the SPI doesn't say.
  - FlowInput.from doesn't null-check getBytes() (unboxing NPE), and UNKNOWN direction is silently treated as egress.

  Hygiene

  - The "backend-agnostic" module isn't: flusher thread named postgres-flow-aggregator, javadoc @links to FlowAggWriter/PostgresFlowQueryService which live in PR #8639.
  - AggregatedFlowSink is exported but never used — everything traffics in plain Consumer<List<AggregatedFlow>>.
  - perNodeShift: Math.abs(Integer.MIN_VALUE) is negative for exactly one nodeId. Unused in the wired mode, but a landmine.
  - Leftover thinking-out-loud test comment (// ... = 1 >= 1000? no... push further); missing trailing newlines; five ~200k-iteration fuzz loops slow every build.

dino2gnt:
My Claude agreed with your Claude.

PR #8688, opened 2026-07-28
NMS-20081: UI seam layer, Phase 2

synqotik:
Second phase of the "seam" UI component wrapper layer started in #8671 (NMS-20029). App code depends on a stable, OpenNMS-owned API instead of PrimeVue directly, so the underlying framework can be swapped without rewriting consumers. See `ui/packages/onms-ui/README.md` for the seam rules.

Large diff, but low risk: the overwhelming majority is mechanical, byte-compatible tag/import renames onto wrappers that forward 1:1. All intentional behavior changes are enumerated below.

## What

Takes the package from 13 to **35 components** and bans direct imports of every wrapped PrimeVue module in `ui/src` (ESLint `no-restricted-imports`). Two tranches in one branch:

**Tranche 2 — the wide middle of the component surface.** New wrappers: `OnmsTabs`/`OnmsTabList`/`OnmsTab`/`OnmsTabPanels`/`OnmsTabPanel`, `OnmsMenu` (+ `OnmsMenuItem` type), `OnmsDrawer`, `OnmsChip`, `OnmsToggleSwitch`, `OnmsRadioButton`, `OnmsInputNumber`, `OnmsSearchInput` (composite: field + baked search icon, unifying 13 hand-rolled search fields), `OnmsPopover`, `OnmsPanel`, `OnmsCard`, `OnmsListbox`, `OnmsMultiSelect`, `OnmsDatePicker`. `ConfirmationDialog`/`MessageDialog` moved into the package as `OnmsConfirmationDialog`/`OnmsMessageDialog` (30 consumers repointed, APIs unchanged).

**Tranche 3 — tables.** `OnmsTable` wraps DataTable (data/pagination/lazy/sort/row-expansion/row-editing/virtual-scroll surface — exactly what the app uses; selection mode, filters, CSV export etc. are deliberately not exposed). `OnmsColumn` is a **compile-time typed re-export** of PrimeVue Column rather than a component wrapper: DataTable discovers columns by walking its slot vnode tree for `type.name === 'Column'`, and a wrapper vnode is only discovered when it carries an explicit `key` — too fragile to require at ~120 usages. Runtime behavior is therefore identical by construction; the seam narrowing is enforced by `vue-tsc`. All 29 consumer files migrated; `primevue/datatable`/`primevue/column` banned with zero exceptions.

Also addresses all five review comments from #8671 (see below).

## Design notes

- Wrapper style unchanged from tranche 1: explicit props/slots/emits, DOM attrs fall through, `unsafePt` as the documented escape hatch. Exception: `OnmsColumn`'s escape hatch keeps PrimeVue's `pt` name — a type cast can't rename a runtime prop (documented in the README).
- `OnmsTable` bakes `<th scope="col">` onto every column header via table-level passthrough, deep-merged under `unsafePt` so consumer keys win. Three tables previously hand-rolled this per column; the rest gain it.
- Deliberate seam defaults (each README-documented): `OnmsDrawer` opens right (every usage), `OnmsMenu` is a popup by default (every usage), `OnmsInputNumber` disables digit grouping (all numeric fields here are ports/counts/intervals), `OnmsTab` bakes the uppercase tab styling.
- `OnmsInputNumber.inputProps`: PrimeVue 4.5.5's InputNumber has no such prop, so plain DOM attrs are routed through `pt.pcInputText.root` to reach the real `<input>` — this *fixed* two sites whose `data-test`/`aria-label` were silently inert, and two broken label-for associations.
- Event payloads keep PrimeVue's field shapes under seam-owned types (`OnmsTablePageEvent`, `OnmsTableSortEvent`, …) so existing handler bodies compile unchanged.
- Contract tests for every wrapper in `ui/tests/onms-ui/` (prop mapping, emit forwarding, slot forwarding, pt merge paths); suite is 2,429 tests green.

## Intentional behavior changes (everything else is 1:1)

**Device Config Backup screen (the one real UX change, requested):**
- Custom infinite scroll replaced with standard server-side pagination: 20 rows/page, options [20, 50, 100, 200]. The table no longer body-scrolls inside a fixed viewport, and the DOM-coupled scroll watcher into PrimeVue internals is gone.
- Search/group-filter/status/sort refetches preserve the selected page size (previously hardcoded 20), and the paginator restores its page size on remount.
- **Select-all now does what it advertises**: Download/Backup act on *every* config matching the current filters, resolved at action time. Previously they silently acted on only the loaded page's id snapshot — after paging, Backup errored and Download shipped the wrong set. Downloads over 500 configs arrive as multiple zip files (URL-length limit on the ids GET).

**Cosmetic/markup:**
- Every migrated table's header cells gain `scope="col"` (a11y, markup-only).
- Node detail "Interfaces" tabs are now uppercase like every other tab set.
- The three RRA editor fields (XFF/Steps/Rows) no longer show digit-grouping separators while editing.
- 13 search inputs now share one look via `OnmsSearchInput`.

**From the #8671 review comments (thanks @marshallmassengill):**
- `v-model.trim` inputs re-sync the visible value on change, matching native Vue semantics (screen can no longer disagree with the model at submit time).
- Dismissing a toast immediately lifts duplicate-suppression for that message (previously suppressed until the original would have expired).
- `ONMS_UI_VERSION` is single-sourced from `package.json`; the package license is a valid SPDX expression (`AGPL-3.0-or-later`); `useSnackbar`'s severity cast replaced with a compile-checked total map.

## Out of scope (follow-ups)

Theme-token relocation into the package, lib build + `.d.ts` emission, npm publish, `window.OnmsUI` runtime exposure for plugins, ServiceNow plugin migration. Ticketed minors: contract tests for the `#expansion` slot scope and object-form `expandedRows`; surfacing per-chunk download failures on the DCB screen (pre-existing silent-failure pattern).

There are probably a number of small styling tweaks that are still needed. I'd like to get this merged first, since other work depends on it. I'll open additional issues for any styling or other issues.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20081
* Previous phase: https://github.com/OpenNMS/opennms/pull/8671 (NMS-20029)


marshallmassengill (approved):
Approving.  Couple items from Claude though:

- Select-all is sticky with no visual truth. allSelected is only cleared by an explicit row selection — nothing resets it (or the header checkbox) on search, filter, status, or page changes. So: select all under vendor=Cisco, clear the filter, hit
  Download → you get everything in the system, while page-2 row checkboxes render unchecked next to a checked header. "Resolved at action time" is defensible, but the selection UI and the action set can silently disagree. Clear allSelected on query-param
  changes or show an "N selected" count.
- Backup-all is one unbounded POST. Download chunks at 500 ids for URL length, but the test pins backup sending all 1200 configs in a single request — on a 10k-device estate that's a multi-MB POST kicking off 10k backups with no confirmation. Needs
  chunking or at least a confirm dialog.
- The fetch-all limit trusts a stale deviceConfigTotal from the last list fetch — configs added since are silently excluded, and the backend may cap limit anyway.
- Multiple programmatic downloads in one gesture are popup-blocked in most browsers past the first; jsdom tests can't see that. Worth one manual check with >500 configs.

synqotik:
> Approving. Couple items from Claude though:

@marshallmassengill thanks, I'll open a ticket for the DCB issues. There are a bunch of decisions and trade offs re how this should work.

PR #8637, opened 2026-07-16
NMS-20026: Fix TicketNotificationStrategy wiring

cgorantla:
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20026

PR #8691, opened 2026-07-29
NMS-20101: Make TicketNotificationStrategy usable and document it

marshallmassengill:
Builds on #8637 (so that needs to merge first). The strategy shipped with no reference in any notificationCommands.xml and no documentation, so it was unreachable without hand-written configuration and a naïve wiring hits the "no alarm-id" error because the delivery-time alarm lookup races alarmd.

- Adds a createTicket example command and a Create-Ticket example destination path whose non-zero initial-delay avoids the alarmd race
- Adds a strategy docs page under Bonus Notification Methods covering prerequisites and caveats
- Replaces the hardcoded admin ticket user with an optional ticketUser argument, settable per notification; absent or blank values fall back to admin

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20101

### DIFF
--- a/docs/modules/operation/pages/deep-dive/notifications/strategies/ticket.adoc
+++ b/docs/modules/operation/pages/deep-dive/notifications/strategies/ticket.adoc
@@ -19,7 +19,8 @@ Use a destination path with a non-zero `initial-delay` to avoid this; 30 seconds
 The strategy reads only event parameters, so the target serves to satisfy the destination path schema.
 * Target a single user rather than a group.
 Notifd runs each command once per member of a group target, and each run sends its own create-ticket event.
-The strategy skips the create event when the alarm already has a ticket, but near-simultaneous deliveries can race the asynchronous ticket creation, so a group target can still produce duplicate tickets.
+The strategy skips the create event when the alarm already has an active ticket; tickets that are closed, resolved, or cancelled, or whose creation failed, do not block a new one.
+Near-simultaneous deliveries can race the asynchronous ticket creation, so a group target can still produce duplicate tickets.
 * The create-ticket event carries a `user` parameter that defaults to `admin`.
 You can override it with the optional `ticketUser` argument described below.
 Whether the value appears on the created ticket depends on the ticketer plugin.
@@ -60,6 +61,9 @@ Add a destination path with an initial delay to `destinationPaths.xml`:
     </target>
 </path>
 ----
+
+NOTE: Use a single user as the target, not a group.
+Notifd runs the command once per member of a group target, and near-simultaneous runs can each create a ticket for the same alarm.
 
 Reference the `Create-Ticket` destination path from each notification in `notifications.xml` that should open a ticket.
 
--- a/docs/modules/operation/pages/deep-dive/notifications/strategies/ticket.adoc
+++ b/docs/modules/operation/pages/deep-dive/notifications/strategies/ticket.adoc
@@ -1,0 +1,70 @@
+
+= Trouble Ticket Notifications
+:description: How to configure {page-component-title} to create trouble tickets from notifications.
+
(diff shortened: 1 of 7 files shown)


opennms-base-assembly/src/main/filtered/etc/examples/destinationPaths.xml
```diff
@@ -216,4 +216,15 @@
                 <command>growlMessage</command>
         </target>
     </path>
+    <!-- Creates a trouble ticket for the alarm associated with the notification's event.
```
cgorantla, line 219: we should avoid adding config changes in foundation 2024, That's one of the reason I kept other PR limited to fixing.

Everything else can stay. Just need to avoid config updates.

marshallmassengill: Even for an example?

cgorantla: didn't realize this one is example.

marshallmassengill: All good.  @dino2gnt now has me investigating if we actually create rpmnew files for things in the examples folder (we probably shouldn't).

marshallmassengill: no rpmnew if the file hasn't been modified from stock so I think this should be ok.  Most folks aren't modifying the examples files regularly or ever.

cgorantla (commented):
LGTM, need to resole conflicts

docs/modules/operation/pages/deep-dive/notifications/strategies/ticket.adoc
```diff
@@ -0,0 +1,70 @@
...
+</command>
+----
+
+Add a destination path with an initial delay to `destinationPaths.xml`:
+
+[source, xml]
+----
+<path name="Create-Ticket" initial-delay="30s">
+    <target>
+        <name>Admin</name>
```
cgorantla: This may create multiple tickets for each user if  Admin group has multiple users.

marshallmassengill: Investigating.

marshallmassengill: Updated.

features/notifications/ticket-strategy/src/main/java/org/opennms/netmgt/notifd/TicketNotificationStrategy.java
```diff
@@ -150,13 +157,21 @@ public int send(List<Argument> arguments) {
...
-        /* Log everything we know so far.
-         * The tticketid and tticketstate are only informational.
+
+        /* Guard against duplicate tickets: a previous notification, an escalation,
+         * or another member of a group target may already have created one.
+         * Near-simultaneous deliveries can still race the asynchronous ticket
+         * creation, so single-user targets remain the recommended configuration.
          */
-        LOG.info("Got event-uei='{}' with event-id='{}', notice-id='{}', alarm-type='{}', alarm-id='{}', tticket-id='{}'and tticket-state='{}'", eventUEI, eventID, noticeID, alarmType, alarmState.getAlarmID(), alarmState.getTticketID(), alarmState.getTticketState());
+        if( StringUtils.isNotBlank(alarmState.getTticketID()) ) {
```
cgorantla: I think this will prevent tickets being created for tickets that are already closed. So we should at least check if the ticket is open or create_pending state

marshallmassengill: I think latest update should address this and make all of this more clear.  Let me know what you think.

cgorantla (approved):
LGTM!

I have created a Jira issue to implement `close ticket` workflow to complement this https://opennms.atlassian.net/browse/NMS-20105

PR #8698, opened 2026-07-30
NMS-20106: versioned user management API and PrimeVue Manage Users page

joseanesONMS:
[NMS-20106](https://opennms.atlassian.net/browse/NMS-20106): a versioned user management API (`/api/v2/users`) so users can be provisioned by external tooling, and a PrimeVue Manage Users page that is a straight visualization of it. `users.xml` stays the system of record via the existing `UserManager`; fields the API does not expose (XMPP among them, deliberately hidden in the UI) and the password survive updates untouched.

- Endpoints: list/get/create/update/set-password/rename/delete plus available-roles, admin-only via new Spring Security rules and in-code checks.
- Responses never contain the password hash in any form (the v1 API returns it to admins).
- The admin/rtc delete and rename protections and the admin ROLE_ADMIN-retention guard are enforced server-side; the legacy page only hid the buttons.
- Requests are validated up front (ids, roles, time zones, duty schedule grammar, comments markup) and applied to a detached copy, so a rejected request leaves no partial state, while hand-edited values are grandfathered so files never become uneditable.
- The page validates fields inline against the same rules and shows API rejections inside the dialogs.
- 22 integration tests (mock managers, `users.xml` never touched) plus Vitest store/dialog tests; verified end to end against a local instance including byte-level `users.xml` restoration.

[NMS-20106]: https://opennms.atlassian.net/browse/NMS-20106?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/UsersRestService.java
+++ b/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/UsersRestService.java
@@ -1,0 +1,323 @@
+/*
+ * Licensed to The OpenNMS Group, Inc (TOG) under one or more
+ * contributor license agreements.  See the LICENSE.md file
+ * distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * TOG licenses this file to You under the GNU Affero General
+ * Public License Version 3 (the "License") or (at your option)
+ * any later version.  You may not use this file except in
+ * compliance with the License.  You may obtain a copy of the
+ * License at:
+ *
+ *      https://www.gnu.org/licenses/agpl-3.0.txt
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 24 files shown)


joseanesONMS:
Follow-up commit from the group-management self-review, applying the shared findings here: duty schedules with begin after end are now rejected for new entries (DutySchedule.isInSchedule compares within a single calendar day, so an overnight range never matches at runtime — accepting it silently disabled the schedule), while strings already stored on a record keep round-tripping so hand-edited users.xml entries stay editable; dot-segment ids rejected; emptied form fields now clear values instead of silently preserving them; and a collapsed in-context help panel was added to the page. IT grows to 17 tests.

joseanesONMS:

joseanesONMS:
Replaced by #8713 (in-repo branch so CI runs).

PR #8699, opened 2026-07-30
NMS-20107: versioned group management API and PrimeVue Manage Groups page

joseanesONMS:
[NMS-20107](https://opennms.atlassian.net/browse/NMS-20107): a versioned group management API (`/api/v2/groups`) and a PrimeVue Manage Groups page that is a straight visualization of it; independent companion to #8698. `groups.xml` stays the system of record via the existing `GroupManager`, with member order (the notification escalation order) and `default-map` preserved end to end.

- Endpoints: list/get/create/update/rename/delete, admin-only via new Spring Security rules and in-code checks.
- The Admin group cannot be deleted or renamed, enforced server-side (the legacy page only hid the buttons).
- Renames follow the on-call roles' membership-group references in the same file write, deletes are rejected while a role still references the group, and both go through `GroupService` so the DB category authorizations are migrated or cleared as the legacy page did.
- Overnight duty schedules are rejected for new entries because the runtime never matches them; stored strings, hand-edited comments and stale members are grandfathered so files never become uneditable.
- The page validates fields inline, shows API rejections inside the dialogs, and includes a collapsed in-context help panel.
- 19 integration tests (mock managers, `groups.xml` never touched) plus Vitest store/dialog tests; verified end to end including hand-editing `groups.xml` between API calls.

[NMS-20107]: https://opennms.atlassian.net/browse/NMS-20107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/GroupsRestService.java
+++ b/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/GroupsRestService.java
@@ -74,9 +74,9 @@ public class GroupsRestService implements GroupsRestApi {
     private static final Set<String> PROTECTED_GROUPS = Set.of("Admin");
 
     /** Markup per the legacy controller, plus URL-path-segment safety. */
-    private static final Pattern INVALID_NAME = Pattern.compile(".*[&<>\"`':/\\\\%?#\\s]+.*");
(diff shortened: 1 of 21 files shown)


joseanesONMS:

joseanesONMS:
Replaced by #8714 (in-repo branch so CI runs).

PR #8700, opened 2026-07-30
NMS-20109: Active/passive HA for OpenNMS core

cgorantla:
Add built-in active/passive high availability for a pair of OpenNMS core instances, coordinated through the shared PostgreSQL database. 

**Architecture**

(**Core idea and initial implementation credits to @dino2gnt** !)

Two instances share one coordination table (ha_instance_status). All timing decisions use the database clock. Each node is configured in etc/ha-configuration.xml with a static role (PRIMARY/SECONDARY) and one of two supervision modes:

coordinator (default): OpenNMS runs the HA state machine itself — startup gating on the standby, heartbeat monitoring, automatic promotion with anti-flap verification, DEGRADED/manual-failback handling, split-brain resolution via active_since arbitration, and config sync.

heartbeat-only: an external supervisor (HA agent or third-party tooling) owns promotion/fencing; OpenNMS only publishes its liveness heartbeat and never gates startup

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20109


cgorantla:
Lot more tests done. Should be ready for review.  Need docs.

cgorantla:
> Lot more tests done. Should be ready for review. Need docs.

Updated docs. Tested locally on Mac with two instances with one Postgres. Seems to work fine.

docs/modules/operation/pages/deep-dive/admin/high-availability.adoc
```diff
@@ -0,0 +1,111 @@
+
+= High Availability
+:description: Configure an active/passive {page-component-title} pair coordinated through the shared PostgreSQL database.
+
+ifndef::opennms-prime[]
```
dino2gnt: Is this the right gate?

I think the most common pattern is:
```
ifeval::["{page-component-title}" == "..."]
```

Should also add the Karaf shell commands to the shell reference page.

dino2gnt (changes requested):

dino2gnt (dismissed):
👍

marshallmassengill (changes requested):
These are minor but I think we want to get them documented/sorted:

The enable steps don't have running the install script.... A staged standby has no etc/configured and bin/opennms start refuses outright without it.... I kind of think we may want to consider an option for the install script that does this "install -standby" or something that preps a standby host.

We don't document a procedure for creating the hasync account.  Not a big deal but something we should probably do in these docs to be explicit with customers.

"DEGRADED" means two things in one ha-status table. STATE=DEGRADED (restarted primary awaiting failback) versus a DEGRADED=YES column (stale heartbeat, or SECONDARY holding ACTIVE). After a normal successful failover both rows read DEGRADED=YES, which reads as total failure.

We need to document an approximate failover time length and latency requirements.  I'm going to do further testing on this.

Some other nits that claude caught:
- nav.adoc:257 still gates on ifdef::opennms-prime[]
- Page :description: (line 2) sits outside the conditional, so the Horizon build emits metadata advertising HA even though the body is a "not included" stub.
- <activation><property> only sees CLI properties, so if the Meridian pipeline sets it as a pom property neither profile activates and HA silently doesn't ship.
- Defaults disagree in a couple of spots: failover-threshold-seconds: class 30, shipped XML 60, javadoc 30. sync-interval-seconds: class 300, shipped XML 60. The page states no defaults, so omitting an element gives 300s sync, not the 60 commented in the file.
- Silent clamping undocumented. clampConfig forces heartbeat ≥5, threshold ≥20 and ≥2× heartbeat, sync ≥5. The 2× floor silently overrides a configured threshold.

christianpape:
> "DEGRADED" means two things in one ha-status table. STATE=DEGRADED (restarted primary awaiting failback) versus a DEGRADED=YES column (stale heartbeat, or SECONDARY holding ACTIVE). After a normal successful failover both rows read DEGRADED=YES, which reads as total failure.

I agree, I was also a bit confused about this.

opennms-full-assembly/pom.xml
```diff
@@ -55,6 +55,41 @@
         <build.assembly.format>dir</build.assembly.format>
       </properties>
     </profile>
+    <profile>
```
mershad-manesh, line 58: We shouldn't create extra profiles, for testing it's ok

cgorantla:
> These are minor but I think we want to get them documented/sorted:
> 
> The enable steps don't have running the install script.... A staged standby has no etc/configured and bin/opennms start refuses outright without it.... I kind of think we may want to consider an option for the install script that does this "install -standby" or something that preps a standby host.
> 
> We don't document a procedure for creating the hasync account. Not a big deal but something we should probably do in these docs to be explicit with customers.
> 
> "DEGRADED" means two things in one ha-status table. STATE=DEGRADED (restarted primary awaiting failback) versus a DEGRADED=YES column (stale heartbeat, or SECONDARY holding ACTIVE). After a normal successful failover both rows read DEGRADED=YES, which reads as total failure.
> 
> We need to document an approximate failover time length and latency requirements. I'm going to do further testing on this.
> 
> Some other nits that claude caught:
> 
> * nav.adoc:257 still gates on ifdef::opennms-prime[]
> * Page :description: (line 2) sits outside the conditional, so the Horizon build emits metadata advertising HA even though the body is a "not included" stub.
> * only sees CLI properties, so if the Meridian pipeline sets it as a pom property neither profile activates and HA silently doesn't ship.
> * Defaults disagree in a couple of spots: failover-threshold-seconds: class 30, shipped XML 60, javadoc 30. sync-interval-seconds: class 300, shipped XML 60. The page states no defaults, so omitting an element gives 300s sync, not the 60 commented in the file.
> * Silent clamping undocumented. clampConfig forces heartbeat ≥5, threshold ≥20 and ≥2× heartbeat, sync ≥5. The 2× floor silently overrides a configured threshold.

Most of these should be addressed in the latest changes and some other stuff that I found in testing.  We still need to add docs in install script for secondary node.  I will do it in next iteration,

dino2gnt:
Secondary node can just `touch ${OPENNMS_HOME}/etc/configured`...  or, for that matter, run the installer normally. It's idempotent.

cgorantla:
> Secondary node can just `touch ${OPENNMS_HOME}/etc/configured`... or, for that matter, run the installer normally. It's idempotent.

Updated doc.

cgorantla:
Some follow-ups captured here https://opennms.atlassian.net/browse/NMS-20110

dino2gnt (approved):

christianpape (approved):
LGTM.

opennms-full-assembly/pom.xml
mershad-manesh: We can use something like
```
 <property>
     <name>env.CIRCLE_PROJECT_REPONAME</name>
     <value>opennms-prime</value>
 </property>
```

The only downside with this that the user who wants to run the build locally needs to set that environment variable

cgorantla: I added this to follow up list on HA https://opennms.atlassian.net/browse/NMS-20110

marshallmassengill (approved):
LGTM

PR #8701, opened 2026-07-30
NMS-20102: Migrate all mail sending/reading from JavaMail 1.4.7 to Jakarta Mail 2.1 (Part of rework JavaMailNotificationStrategy)

marshallmassengill:
This was part of the rework on the JavaMailNotificationStrategy work to support modernizing notifications.  In thoery, this keeps all existing config and shouldn't break anything.

Migrates all mail sending/reading from JavaMail 1.4.7 to Jakarta Mail 2.1 (Eclipse Angus) and adds OAuth2/XOAUTH2 SMTP authentication, which Microsoft 365 and Gmail now require.

- Existing javamail-configuration.{xml,properties} files load unchanged; compatibility is guarded by WillItUnmarshalIT/ConfigTesterTest and new GreenMail wire tests.
- XOAUTH2 needs only configuration: a token-auth definition plus ${token:<name>} as the mail password (docs include an M365 client-credentials example).
- com.sun.mail:mailapi:1.6.7 stays on the classpath for EclipseLink MOXy 2.5.1, which hard-references javax.mail.internet.MimeMultipart; the flat-classpath Angus jar is shaded to remove javax-era resource names that break the javax-side mailcap lookup.
- Behavior changes (released behavior): the jmta transport is removed (use-jmta accepted, warns, sends via SMTP); <javamail-property> entries now reach the mail session (previously discarded); Angus defaults mail.smtp.ssl.checkserveridentity=true, so mismatched certificates on SSL/TLS connections now fail; the javax.mail API no longer ships SMTP/IMAP/POP3 providers (affects third-party plugins only).

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20102


dino2gnt (dismissed):
This touches a surprisingly large number of things.

If it goes green, ship it.  When this is all moved into a UI + REST, we can revisit the configuration parsing and some of those 20+-year-old TODOs and FIXMEs and rip it all out.

marshallmassengill:
> This touches a surprisingly large number of things.
> 
> If it goes green, ship it. When this is all moved into a UI + REST, we can revisit the configuration parsing and some of those 20+-year-old TODOs and FIXMEs and rip it all out.

Yep.  I was genuinely shocked at how rooted this is in there... it's not ideal but my thought process is that if I'm touching this at all, this is the way to do it for now.  I think once this is in, it opens up the door for a lot of future work.

dino2gnt:
> I think once this is in, it opens up the door for a lot of future work.

Absolutely.  And besides, even if this is completely broken, it's only ~10% more broken than things are _today_. :laughing:

marshallmassengill:
@indigo423 - I can confirm this should work for everything in OpenNMS that touches mail today but if you find something that doesn't work, let me know. I tested locally with mailpit, the following:

- JavaMailNotificationStrategy
- MailerServlet
- DefaultReportWrapperService
- MailTransportMonitor
- MailAckProcessor (ackd)
- EmailNorthbounder

I can also confirm that token auth works with gmail.  I haven't been able to directly test O365 but I don't think we should avoid shipping this.

christianpape (dismissed):
LGTM.

docs/modules/releasenotes/pages/whatsnew.adoc
```diff
@@ -83,6 +83,25 @@ See xref:operation:deep-dive/events/event-configuration.adoc#migrating-event-fil
...
+The `use-jmta` attribute and the `org.opennms.core.utils.useJMTA` property are still accepted so existing files keep loading.
+Setting `useJMTA=true` in the properties file logs a warning and sends over SMTP to the configured mail host (default `127.0.0.1:25`) instead; the `javamail-configuration.xml` attribute never selected the transport on that path, so its behavior is unchanged.
+MailTransportMonitor `sendmail-test` definitions that relied on `use-jmta` (its default is `true`) now also deliver through the configured sendmail-host; if no SMTP listener runs there, the test - correctly - reports the service down.
+* The `javax.mail` API remains on the classpath for internal use, but without SMTP/IMAP/POP3 providers or `com.sun.mail.*` classes.
+Third-party plugins that send mail through `javax.mail.Session` need to move to `jakarta.mail` or bundle their own provider.
+* `<javamail-property>` entries in `javamail-configuration.xml` now take effect on the sendmail path; previously they were silently ignored.
+If your configuration contains such entries that were added but never worked, they will start applying after the upgrade.
```
cgorantla: I would restrict whatsnew.doc to 2 or 3 lines per item that actually changed. Anything else can be moved to different page if it needs more details.

docs/modules/operation/pages/deep-dive/meta-data.adoc
```diff
@@ -284,6 +284,10 @@ Furthermore, some {page-component-title} configuration files allow the use of Me
 * xref:reference:service-assurance/monitors/JCifsMonitor.adoc[JCIFS monitor]: jcifs.properties
 * xref:reference:service-assurance/monitors/JmxMonitor.adoc[JMX monitor], xref:reference:performance-data-collection/collectors/jmx.adoc[JMX collector]: jmx-config.xml, jmx-datacollection-config.xml, jmx-datacollection-config/*.xml
 * xref:operation:deep-dive/notifications/introduction.adoc[Notification configuration]: javamail-configuration.xml, javamail-configuration.properties
++
+The outbound mail credential fields (`user-auth` on a `sendmail-config` in the XML, `authenticateUser`/`authenticatePassword` in the properties file) additionally resolve the `token` context (`${token:<name>}`), which supplies an access token from a xref:operation:deep-dive/admin/configuration/token-authentication.adoc[token authentication] definition.
+This is mainly useful for OAuth2 (XOAUTH2) SMTP authentication.
+Note that `readmail-config` credentials (the ackd mail reader) do not currently support metadata expressions.
```
cgorantla: Do we need to mention what is not supported here ?

cgorantla (dismissed):
There is lot of churn in this PR for keeping conflicting javax.mail and jakarta.mail related packages.  Might be worth looking at this again after migrating to newer Spring/Hibernate versions.

LGTM if it works in the field.

docs/modules/releasenotes/pages/whatsnew.adoc
marshallmassengill: greatly reduced.

cgorantla (approved):
LGTM!

PR #8702, opened 2026-07-30
NMS-20108: On-Call Roles management via versioned REST API and PrimeVue UI

joseanesONMS:
Rewrites Admin → On-Call Roles as a PrimeVue page backed by a new versioned REST API, keeping groups.xml as the system of record.

- Adds admin-only `/api/v2/on-call-roles`: list/get/create/update/rename/delete plus a computed month calendar that reuses the exact interval resolution notifd uses; `time-zone` is reported so clients render times in server time.
- Hand-edited roles round-trip untouched and stay editable: file changes are picked up on read, schedule entries already stored skip new-entry validation, the optional time `id` attribute is preserved, and a role whose schedule the runtime cannot evaluate reports `schedule-error` instead of failing the list.
- New entries are validated and canonicalized to the exact forms `BasicScheduleUtils` dispatches on (zero-padded widths, English month names, lowercase weekday, unpadded monthly day); overnight `begins>ends` entries are rejected because the runtime never matches them.
- The page shows the legacy list columns plus an explicit supervisor-fallback indicator, a month calendar with one-off coverage add/remove (content-matched removal, refetch-before-mutate, confirmation before removing recurring entries), rename/delete with referential warnings, and inline field validation with API errors shown in the dialogs.
- The admin menu entry now points at the new page. 16 integration tests cover the API; store, validation and dialog behavior are covered by vitest.

### DIFF
--- a/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/OnCallRolesRestService.java
+++ b/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/OnCallRolesRestService.java
@@ -1,0 +1,555 @@
+/*
+ * Licensed to The OpenNMS Group, Inc (TOG) under one or more
+ * contributor license agreements.  See the LICENSE.md file
+ * distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * TOG licenses this file to You under the GNU Affero General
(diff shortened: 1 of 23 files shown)


joseanesONMS:

joseanesONMS:
Replaced by #8715 (in-repo branch so CI runs).

PR #8703, opened 2026-07-30
NMS-20117: Add docker compose developer stack

cgorantla:
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20117


mershad-manesh:
Neat idea, should we put it in https://github.com/opennms-forge/stack-play instead? 
It would be good to get feedback from @indigo423 as well

PR #8693, opened 2026-07-29
NMS-20100: PrimeVue Notifications page with a General configuration tab (base)

joseanesONMS:
Rewrites the Notifications page as PrimeVue backed by REST, split from the original single PR for ease of review. This base PR carries the page itself and a Configure Notifications dialog with only the General tab; the Event Notifications, Destination Paths and Path Outages tabs follow in dependent PRs.

- Notifications page: notice queries (per-user, outstanding, acknowledged), the outstanding-notices list with acknowledge, CSV export and printing, and collapsible explanations.
- Configure Notifications dialog (gear icon) with the General tab: the system-wide notification on/off switch.
- Adds admin-only /rest/notification-config endpoints for the notifd status plus the read side of destination paths (shared by the dependent tab PRs), delegating to the same file-backed factories the legacy wizards use — the XML files stay the system of record.
- Browser delivery of notices for /ui pages.
- The admin menu entry points at the new page.

Dependent PRs: Event Notifications tab, Destination Paths tab, Path Outages tab (each contains this PR's commit until it merges).

PR #8706, opened 2026-07-30
NMS-20100: Event Notifications tab

joseanesONMS:
Adds the Event Notifications tab to the Configure Notifications dialog. Depends on #8693 (the base Notifications page PR); review the last commit only until that merges.

- Tab with the event-notification list and per-notification on/off toggles.
- Add/edit dialog with a searchable UEI picker fed by the event configuration; delete with confirmation.
- /rest/notification-config/event-notifications endpoints wrap the notifications.xml factory the legacy wizard uses; payloads are validated up front because the factory removes the old entry before a failing save.


joseanesONMS:
Replaced by #8710 (NMS-20118, in-repo stacked on #8709).

PR #8707, opened 2026-07-30
NMS-20100: Destination Paths tab

joseanesONMS:
Adds the Destination Paths tab to the Configure Notifications dialog. Depends on #8693 (the base Notifications page PR); review the last commit only until that merges.

- Tab with the path list and a test-trigger action.
- Editor covering initial delay, targets (users, groups, on-call roles, email addresses) with per-target commands, intervals and auto-notify, and escalations; the Browser notification method is selectable end to end.
- /rest/notification-config destination-path mutations, commands and on-call-roles endpoints wrap the same destinationPaths.xml, notificationCommands.xml and groups.xml factories the legacy wizard uses; a path rename follows into the event notifications that reference it.
- GroupManager rebuilds its role map on every parse so removing the last role in groups.xml is not masked by a stale cache.

### DIFF
--- a/opennms-config/src/main/java/org/opennms/netmgt/config/GroupManager.java
+++ b/opennms-config/src/main/java/org/opennms/netmgt/config/GroupManager.java
@@ -144,11 +144,11 @@ public abstract class GroupManager implements GroupConfig {
         }
         buildDutySchedules(m_groups);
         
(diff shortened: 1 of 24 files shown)


joseanesONMS:
Replaced by #8711 (NMS-20119, in-repo stacked on #8709).

PR #8708, opened 2026-07-30
NMS-20100: Path Outages tab

joseanesONMS:
Adds the Path Outages tab to the Configure Notifications dialog. Depends on #8693 (the base Notifications page PR); review the last commit only until that merges.

- Tab with the list of configured critical paths and an apply/clear flow driven by a filter rule with a node-count preview.
- Storage is and stays the pathoutage table; the endpoints mirror the legacy NotificationWizardServlet semantics, including clearing the path when the critical IP is blank.


joseanesONMS:
Replaced by #8712 (NMS-20120, in-repo stacked on #8709).

PR #8709, opened 2026-07-30
NMS-20100: PrimeVue Notifications page with a General configuration tab

joseanesONMS:
First slice of the NMS-20100 epic (Notifications admin rework), targeting the epic feature branch `jira/NMS-20100`. Carries the PrimeVue Notifications page and a Configure Notifications dialog with only the General tab; the Event Notifications (NMS-20118), Destination Paths (NMS-20119) and Path Outages (NMS-20120) tabs follow in stacked PRs.

- Notifications page: notice queries (per-user, outstanding, acknowledged), the outstanding-notices list with acknowledge, CSV export and printing, and collapsible explanations.
- Configure Notifications dialog (gear icon) with the General tab: the system-wide notification on/off switch.
- Adds admin-only /rest/notification-config endpoints for the notifd status plus the read side of destination paths (shared by the tab PRs), delegating to the same file-backed factories the legacy wizards use — the XML files stay the system of record.
- Browser delivery of notices for /ui pages; the admin menu entry points at the new page.


joseanesONMS:
@marshallmassengill could you take a review pass when you get a chance? This is the base of the NMS-20100 stack — #8710, #8711 and #8712 each add one tab on top of it.

marshallmassengill (changes requested):
Some stuff that I think is worth adjusting here:

Bugs worth fixing before merge

1. The notices list can hang forever when /whoami fails. AdminNotifications.vue gates noticesStore.load() on authStore.loaded, via whenever(authLoaded, ...). But authStore.getWhoAmI() sets loaded.value = true only inside if (resp), and whoAmIService returns a falsy value on error. So if the whoami call fails for any reason, loaded stays false permanently, the whenever watcher never fires, and the user gets a page that renders "No outstanding notices found." with no spinner and no error — indistinguishable from genuinely having no notices. The legacy JSP had no such dependency. Setting loaded in a finally block (in authStore, which fixes the same latent issue for rolesAreLoaded) is the cleanest fix.

2. One transient failure permanently bricks the config toggle. In ConfigureNotificationsDialog.vue, populated latches to true after the first populate() regardless of whether it succeeded. getNotificationConfigStatus() returns null on error, and the switch is :disabled="store.notifdStatus === null". A single failed GET — a server restart, a 500 — leaves the toggle disabled and the label reading "Notifications are unknown" for the lifetime of the page, and reopening the dialog won't retry because populated is already true. Either latch only on success, or re-populate on every open.

3. Acknowledging the last row of the last page leaves you on a dead page. noticesStore.acknowledge() reloads without rewinding first. When the ack drops totalCount below the current offset, the reload returns zero rows and the component renders EmptyList even though notices exist. applyPreset resets first correctly; acknowledge should clamp it too.

Export path

4. CSV and print silently truncate at 1000 notices. EXPORT_LIMIT = 1000 caps fetchAllForExport, and the print header then states ${rows.length} notices as though that were the whole set. A user exporting 5000 acknowledged notices gets 1000 with no indication anything was dropped. At minimum, compare against store.totalCount and surface a snackbar when the cap bites.

5. CSV formula injection. escapeCell wraps in quotes and doubles embedded quotes, but doesn't neutralise a leading =, +, -, or @. Notice subject and text-message content derives from event data, which on a trap-fed system is externally influenced — so a crafted varbind can land a formula in the exported file that Excel will execute on open. Prefixing such cells with a single quote is the usual mitigation.

marshallmassengill:
So I think because of how you split this, we're going to need to do a final pass across everything once it's all in place.  I do think this is a better approach though and makes things a bit easier to review.

joseanesONMS:
All five addressed: the whoami fix went into authStore so it also covers `rolesAreLoaded`, the dialog now retries until the status loads, acking the last row rewinds a page, exports warn when the cap truncates (print header says "first 1000 of N"), and CSV cells get the quote-prefix guard. Agreed on a final pass across the stack once all four are in.

marshallmassengill (changes requested):
Some more changes here:

1. Whoami fallback shows all users' notices under "Your Outstanding Notices." loaded now always flips, but whoAmI.id stays undefined, so effectiveUser → null and browseNotices drops the user filter. Make the yourOutstanding/userSearch presets refuse to run without a user id and surface an error instead of silently widening.
2. Dead code in NotificationConfigRestService. Unused javax.ws.rs.DELETE and POST imports, the never-called isBlank, three stray blank lines before the closing brace, and a class javadoc claiming coverage of notifications.xml and notificationCommands.xml that this PR doesn't touch.
3. Ack rewind steps back only one page. Clamp first to the last valid offset from totalCount rather than subtracting a single rows.
4. sendStatusEvent sends only remoteUser; UpdateNotifdStatusServlet also sends remoteHost and remoteAddr. Match it or note the omission as deliberate.
5. OnmsIconButton is driven by bare text/rounded fallthrough attrs instead of the seam's declared variant prop. Still two occurrences in NoticesTable.vue.
6. IT hygiene. Shared target/test-work-dir/etc with opennms.home set globally and never restored; the "cold-start ordering regression guard" comment claims more than the test proves; testDestinationPathsReadable still carries a stale comment about sibling tests renaming paths.
7. useBrowserNotifications has no test despite running on every /ui page — reconnect loop, array unwrapping, snackbar fallback all unverified.
9. Rename AdminNotifications.vue → Notifications (the page isn't admin-gated; only the gear is).
10. Declare all four tabs (final order + default) in ConfigureNotificationsDialog.vue with placeholder panels, and replace the populate() Promise.all with per-tab loaders. This removes most of the 7–8 file conflicts each sibling pair currently has, forces the tab-order decision to be deliberate rather than an artifact of merge sequence, and fixes #8710's retry-gap finding as a side effect.

Resolving these should get this into a state for merging at least.

joseanesONMS:
All ten addressed. The dialog now declares all four tabs with placeholders and per-tab success-latching loaders, so the siblings each replace one panel — their conflicts collapsed to a loader entry and an import. User-scoped queries refuse to run without a user id, the ack clamp derives from totalCount, the status event carries remoteHost/remoteAddr, the container is Notifications.vue, the icon buttons use the variant prop, the IT restores opennms.home, and the composable has a test suite covering the reconnect loop, unwrapping and the snackbar fallback.

marshallmassengill (changes requested):
A few more but getting close I think... I'm honestly 50/50 on most of these because they are really minor and some of them are going to get addressed with the other PRs:

- Dialog defaults to event-notifications, a placeholder here. Default to general until #8710 lands.
- populate() and getDestinationPaths() now unreferenced, but notificationConfigStore.test.ts:70 still tests them.
- sendStatusEvent reads request.getRemoteHost() outside the try — 500 after the status already persisted.
 - 127.0.0.1.tm0.epoch and tmlog0.log committed under opennms-webapp-rest/; all siblings inherit them.
 - gear button still text rounded
 - setRelativeHomeDirectory not restored

marshallmassengill (changes requested):
Couple more changes needed:

The menu entry has roles: null, so every user lands on this page, and Acknowledge issues PUT /rest/notifications/{id}, which falls through to the /rest/** catch-all restricted to ROLE_REST,ROLE_ADMIN. Legacy /notification/acknowledge was under /** → ROLE_USER. The fix has precedent ten lines up in the same security file: the alarm block carves out /rest/alarms/*\?ack=true for ROLE_USER with !hasRole('ROLE_READONLY').

useRole.ts has no ROLE_READONLY at all, so the button renders for read-only users where list.jsp explicitly hid it

There is also a regression here: fetchAllForExport rebuilds the preset→user mapping itself instead of going through the guarded load(), so it still fails open on a failed whoami.

opennms-webapp/src/main/webapp/WEB-INF/applicationContext-spring-security.xml
```diff
@@ -37,6 +37,38 @@
...
+  -->
+  <http pattern="/rest/notifications/**" realm="OpenNMS Realm" create-session="never" use-expressions="true">
+    <csrf disabled="true" />
+    <headers defaults-disabled="true" />
+
+    <!-- OPTIONS pre-flight requests should always be accepted -->
+    <intercept-url pattern="/rest/notifications/**" method="OPTIONS" access="hasAnyRole('ROLE_ANONYMOUS','ROLE_REST','ROLE_ADMIN','ROLE_USER','ROLE_MOBILE','ROLE_FILESYSTEM_EDITOR')"/>
+
+    <!-- acknowledge / unacknowledge a notification -->
+    <intercept-url pattern="/rest/notifications/**" method="PUT" access="hasAnyRole('ROLE_USER','ROLE_MOBILE','ROLE_REST','ROLE_ADMIN') and !hasRole('ROLE_READONLY')" />
```
marshallmassengill: `/rest/notifications/**` with `method="PUT"` also matches the collection root, so this covers `NotificationRestService.updateNotifications` (NotificationRestService.java:200), which acks or unacks by arbitrary criteria. An empty form body matches every notification, so a `ROLE_USER` can now mass-**un**acknowledge the whole system.

The legacy `AcknowledgeNotificationController` cited for parity requires an explicit `notices` id list and has no unack path, and the alarms rules this mirrors are pinned to `/rest/alarms/*\?ack=true`. Suggest pinning the user-level grant to the single-resource path and leaving the collection at REST/ADMIN. Order matters, first match wins, and the second rule is required rather than redundant: a PUT matching no rule inside this block is permitted with no authorization at all.

```suggestion
    <!-- one notification: a normal user may ack, a read-only user may not -->
    <intercept-url pattern="/rest/notifications/*" method="PUT" access="hasAnyRole('ROLE_USER','ROLE_MOBILE','ROLE_REST','ROLE_ADMIN') and !hasRole('ROLE_READONLY')" />
    <!-- collection PUT acks by arbitrary criteria; also required so no PUT falls through unauthorized -->
    <intercept-url pattern="/rest/notifications/**" method="PUT" access="hasAnyRole('ROLE_REST','ROLE_ADMIN')" />
```

Worth noting this rule can't be covered by a module IT. `AbstractSpringJerseyRestTestCase` doesn't install the security filter chain, which is why `NotificationConfigRestServiceIT.testForbiddenForNonAdmin` passes on the in-code `assertAdmin` instead. Verification is a smoke test or a manual two-account check.

ui/src/services/notificationConfigService.ts
```diff
@@ -0,0 +1,77 @@
...
+}
+
+const getDestinationPaths = async (): Promise<DestinationPath[]> => {
+  try {
+    startSpinner()
+    const resp = await rest.get(`${endpoint}/destination-paths`)
+    return resp.data?.path ?? []
+  } catch (_err) {
+    showSnackBar({ msg: 'Failed to load destination paths.' })
+    return []
```
marshallmassengill: Returning `[]` on failure makes a failed load indistinguishable from an empty result, which is what lets the tab loaders latch (see the comment on `ConfigureNotificationsDialog.vue`). `null` on failure lets the store report success.

```suggestion
const getDestinationPaths = async (): Promise<DestinationPath[] | null> => {
  try {
    startSpinner()
    const resp = await rest.get(`${endpoint}/destination-paths`)
    return resp.data?.path ?? []
  } catch (_err) {
    showSnackBar({ msg: 'Failed to load destination paths.' })
    return null
```

ui/src/stores/notificationConfigStore.ts
```diff
@@ -0,0 +1,48 @@
...
+import { NotifdStatus } from '@/types/notificationConfig'
+import { defineStore } from 'pinia'
+import { ref } from 'vue'
+
+export const useNotificationConfigStore = defineStore('notificationConfigStore', () => {
+  const notifdStatus = ref<NotifdStatus | null>(null)
+
+  const getStatus = async () => {
+    notifdStatus.value = await API.getNotificationConfigStatus()
+  }
```
marshallmassengill: Report whether the fetch actually worked, so the tab loader doesn't have to infer it.

```suggestion
  const getStatus = async (): Promise<boolean> => {
    notifdStatus.value = await API.getNotificationConfigStatus()
    return notifdStatus.value !== null
  }
```

ui/src/components/AdminNotifications/ConfigureNotificationsDialog.vue
```diff
@@ -0,0 +1,153 @@
...
+// Every tab loads its own data on first activation and latches only on
+// success, so a transient failure (restart, 500) retries on the next visit
+// instead of leaving the tab dead for the lifetime of the page.
+const loadedTabs = ref(new Set<string>())
+
+const TAB_LOADERS: Record<string, () => Promise<boolean>> = {
+  general: async () => {
+    await store.getStatus()
+    return store.notifdStatus !== null
+  }
```
marshallmassengill: The comment above is the right contract, but it's easy to opt out of, and all three sibling tab PRs currently do: #8710, #8711 and #8712 each add a loader whose body ends in `return true`. Because their service getters swallow errors and return `[]`, one failed fetch latches the tab as loaded and the retry never happens. For #8710 that's the default tab reading "No event notifications configured." for the life of the page.

Fixing it here sets the pattern they copy on rebase, paired with the service returning `null` on failure. Multi-fetch tabs then become `async () => (await Promise.all([...])).every(Boolean)`.

```suggestion
  general: () => store.getStatus()
```

opennms-config/src/main/java/org/opennms/netmgt/config/DestinationPathManager.java
```diff
@@ -173,9 +173,31 @@ public boolean pathHasTarget(Path path, String target) throws IOException {
...
+    private interface ConfigChange {
+        void apply() throws IOException;
+    }
+
+    /**
+     * Applies an in-memory change and persists it, restoring the previous paths if
+     * persistence fails. saveCurrent() marshals the model only after the caller has
+     * already changed the map, so a change that leaves the config unmarshallable —
+     * e.g. removing the last path, which violates destinationPaths.xsd — would
+     * otherwise leave memory diverged from the on-disk file until a restart.
```
marshallmassengill, line 189: Worth recording that this snapshot is shallow while `NotificationManager`'s is deep (marshal and reparse). It's correct today only because every mutator here replaces whole `Path` objects. I'd keep it shallow rather than deepen it, since a deep copy would add a schema-validated marshal to every happy-path write to guard a hazard that doesn't exist yet, but the constraint should be written down.

```suggestion
     * otherwise leave memory diverged from the on-disk file until a restart.
     *
     * Shallow snapshot: safe only while every mutator replaces whole Path objects. An
     * in-place edit like NotificationManager.replaceNotification would need a deep copy.
```

opennms-config/src/main/java/org/opennms/netmgt/config/DestinationPathManager.java
```diff
@@ -173,9 +173,31 @@ public boolean pathHasTarget(Path path, String target) throws IOException {
...
+     */
+    private synchronized void saveWithRollback(final ConfigChange change) throws IOException {
+        final Map<String, Path> snapshot = new TreeMap<>(m_destinationPaths);
+        try {
+            change.apply();
+            saveCurrent();
+        } catch (final RuntimeException | IOException e) {
+            m_destinationPaths.clear();
+            m_destinationPaths.putAll(snapshot);
+            throw e;
```
marshallmassengill: If the restore itself throws, the original cause is lost and `throw e` never runs.

```suggestion
            try {
                m_destinationPaths.clear();
                m_destinationPaths.putAll(snapshot);
            } catch (final RuntimeException restoreFailure) {
                e.addSuppressed(restoreFailure);
            }
            throw e;
```

opennms-config/src/main/java/org/opennms/netmgt/config/NotificationManager.java
```diff
@@ -1030,8 +1031,31 @@ public List<String> getNotificationNames() throws IOException {
...
+     * file until a restart (update() will not reload a file whose mtime is unchanged).
+     */
+    private synchronized void saveWithRollback(final ConfigChange change) throws IOException, ClassNotFoundException {
+        final String snapshot = JaxbUtils.marshal(m_notifications);
+        try {
+            change.apply();
+            saveCurrent();
+        } catch (final RuntimeException | IOException | ClassNotFoundException e) {
+            parseXML(new StringReader(snapshot));
+            throw e;
```
marshallmassengill: Same as in `DestinationPathManager`, and more likely here: `parseXML` does a validating unmarshal, so it can throw and replace the real cause.

```suggestion
            try {
                parseXML(new StringReader(snapshot));
            } catch (final RuntimeException restoreFailure) {
                e.addSuppressed(restoreFailure);
            }
            throw e;
```

ui/tests/stores/notificationConfigStore.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  })
+
+  describe('Initial State', () => {
+    it('should start empty with unknown notifd status', () => {
+      expect(store.notifdStatus).toBeNull()
+    })
+  })
+
+  describe('notifd status', () => {
+    it('should load the status', async () => {
```
marshallmassengill, line 33: One assertion here would pin the contract the tab loaders depend on, so a sibling PR can't quietly regress it:

```ts
it('reports failure so the tab loader does not latch', async () => {
  vi.mocked(API.getNotificationConfigStatus).mockResolvedValue(null)
  expect(await store.getStatus()).toBe(false)
})
```

marshallmassengill (commented):
Fourth-pass review. The `saveWithRollback` work is the right fix and it resolves the in-memory/on-disk divergence that was blocking #8710 (`varbind`) and #8711 (deleting the last destination path) by inheritance, so both of those downgrade once the siblings rebase. `aFailedReplaceLeavesNoHalfUpdatedEntry` covers the hardest case. The ack authz fix, the read-only hide and the export whoami guard all look right, and `canAcknowledgeNotifications` fails closed on a failed whoami.

One blocker inline: the new PUT rule is broader than its stated intent and lets a normal user mass-unacknowledge everything. The rest are smaller.

Two follow-ups that have no line to attach to here:

1. A shared `serverDetail(err, fallback)` helper in `ui/src/services/serviceHelpers.ts`. Several server-side guards carry a message that is the only explanation the user gets, and the sibling PRs' `delete` and `status` calls discard `err.response.data`. I'll raise it on #8710 and #8711 where it's consumed.
2. `addPath` and `addNotification` rollback cases in the two new test classes. Both suites cover removal, and `NotificationManager` also covers replace, but neither covers an add.

Also worth knowing: all three siblings are still based on `e0bcb101c65`, one commit behind this.

marshallmassengill (changes requested):

joseanesONMS:
All addressed in the latest push:

- **Ack authz scope:** split the notification PUT rule — only the single-resource path (`/rest/notifications/*`) is granted to ROLE_USER; the collection PUT (acks/unacks by arbitrary criteria, so a mass-unack vector) stays REST/ADMIN, with both rules present so no PUT falls through unauthorized.
- **Rollback cause preservation:** both config managers now restore inside their own try/catch and `addSuppressed` a failed restore rather than replacing the original cause; the shallow-snapshot constraint in `DestinationPathManager` is documented.
- **Loader contract:** `getDestinationPaths` returns `null` on failure and `getStatus` returns a boolean, so the general tab loader reports real success (`() => store.getStatus()`); added a store test pinning it.

Noted your point that the authz rule can't be covered by a module IT (the filter chain isn't installed in `AbstractSpringJerseyRestTestCase`) — verifying via smoke / a two-account manual check.

opennms-webapp/src/main/webapp/WEB-INF/applicationContext-spring-security.xml
```diff
@@ -37,6 +37,41 @@
...
+  -->
+  <http pattern="/rest/notifications/**" realm="OpenNMS Realm" create-session="never" use-expressions="true">
+    <csrf disabled="true" />
+    <headers defaults-disabled="true" />
+
+    <!-- OPTIONS pre-flight requests should always be accepted -->
+    <intercept-url pattern="/rest/notifications/**" method="OPTIONS" access="hasAnyRole('ROLE_ANONYMOUS','ROLE_REST','ROLE_ADMIN','ROLE_USER','ROLE_MOBILE','ROLE_FILESYSTEM_EDITOR')"/>
+
+    <!-- one notification: a normal user may ack, a read-only user may not -->
+    <intercept-url pattern="/rest/notifications/*" method="PUT" access="hasAnyRole('ROLE_USER','ROLE_MOBILE','ROLE_REST','ROLE_ADMIN') and !hasRole('ROLE_READONLY')" />
```
marshallmassengill, line 59: `/rest/notifications/*` also matches `/rest/notifications/`, and JAX-RS routes that trailing-slash form to the collection method rather than `@Path("{notifId}")`, since the default `{param}` regex can't match an empty segment. So the criteria-based unack is still reachable by a `ROLE_USER`. Pinning the trailing-slash form ahead of the narrow rule closes it without touching the other two.

```suggestion
    <!-- the trailing-slash form of the collection PUT, which /rest/notifications/* also matches -->
    <intercept-url pattern="/rest/notifications/" method="PUT" access="hasAnyRole('ROLE_REST','ROLE_ADMIN')" />
    <!-- one notification: a normal user may ack, a read-only user may not -->
    <intercept-url pattern="/rest/notifications/*" method="PUT" access="hasAnyRole('ROLE_USER','ROLE_MOBILE','ROLE_REST','ROLE_ADMIN') and !hasRole('ROLE_READONLY')" />
```

The matcher behaviour is measured against spring-core 5.3.39's `AntPathMatcher`; the CXF routing half is from the spec, so worth one `curl -X PUT` to confirm.

marshallmassengill:
With that last one, if it goes green then I think this is ready to merge pending approval from @synqotik.  This is going to require the other PRs go in after so it may be that we want to get them all lined up at once or we may want to get this one merged and then rebase the others.  Apologies for the churn but this is in a much better state now.

joseanesONMS:
Good catch on the trailing slash — fixed in `5bf81b2`. Pinned `/rest/notifications/` (trailing-slash form) to REST/ADMIN ahead of the narrow `/rest/notifications/*` rule, so the criteria-based collection ack/unack is no longer reachable by ROLE_USER; the single-resource and `/**` collection rules are unchanged. Order is trailing-slash → single-resource → collection, first match wins.

Agreed this one isn't module-IT-coverable (the filter chain isn't installed in `AbstractSpringJerseyRestTestCase`) — will confirm with a two-account `curl -X PUT` against the deployed rules.

joseanesONMS:
@marshallmassengill — the base contract the sibling reviews pointed here is in place: `ensureTabLoaded` latches a tab only when its loader resolves truthy, so each tab's fetch (now returning real success in #8710/#8711/#8712) retries on a transient failure instead of latching dead. The three siblings have been updated and re-requested.

marshallmassengill (changes requested):
OnmsButton icon="pi pi-search" at NotificationQueriesCard.vue:34,52. OnmsButton has no icon prop, so it rides $attrs into PrimeVue Button. Both are icon-only buttons with aria-label and no label, which is what OnmsIconButton exists for, and NoticesTable in the same commit already does it correctly with :icon="DownloadFileIcon".

maximizable and :style on OnmsDialog, :loading on OnmsTable. The dialog seam routes sizing through a width prop and does not expose maximizable; loading is not declared on OnmsTable at all. Same class as the OnmsIconButton text rounded finding the author already fixed in pass 2.

Verify visually with a human eyeball.

marshallmassengill (approved):
Looks good to merge into jira/NMS-20100.  We can sort out the merge into release-36.x once we get this done.

Make sure you squash and merge for this one.

joseanesONMS:
@marshallmassengill @synqotik — widened the Configure Notifications dialog to 1400px (still capped at 95vw) so the Event Notifications table fits without the horizontal scrollbar. The seam-API review fixes are already in this PR.

PR #8710, opened 2026-07-30
NMS-20118: Event Notifications tab

joseanesONMS:
**Tickets:** NMS-20118 (epic NMS-20100).

Adds the Event Notifications tab to Configure Notifications, rebased onto the epic with the review feedback applied.

- Table shows a read-only status tag, icon Edit/Delete actions, and an overflow menu offering only the opposite enable/disable action.
- Long UEIs are truncated so the table no longer scrolls horizontally.
- Editor dialog covers the UEI, targets, rule and advanced parameters; every field uses the FormField seam with inline info-icon help (no direct-PrimeVue IftaLabel).
- Backed by new NotificationConfig REST endpoints (with IT) plus store/service/type support.

### DIFF
--- a/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/NotificationConfigRestService.java
+++ b/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/NotificationConfigRestService.java
@@ -53,6 +53,7 @@ import org.opennms.netmgt.config.NotifdConfigFactory;
 import org.opennms.netmgt.config.NotificationFactory;
 import org.opennms.netmgt.filter.FilterDaoFactory;
 import org.opennms.netmgt.filter.api.FilterDao;
+import org.opennms.netmgt.filter.api.FilterParseException;
 import org.opennms.netmgt.config.destinationPaths.DestinationPaths;
 import org.opennms.netmgt.config.notifications.Notification;
 import org.opennms.netmgt.config.notifications.Notifications;
@@ -399,12 +400,20 @@ public class NotificationConfigRestService extends OnmsRestService {
         final FilterDao filterDao = FilterDaoFactory.getInstance();
         try {
             filterDao.validateRule(rule);
-        } catch (final Exception e) {
+        } catch (final FilterParseException e) {
+            // A malformed rule is user input; report it as an invalid result.
+            // Infrastructure failures propagate to a 500 instead.
             result.setValid(false);
             result.setError(e.getMessage());
             return result;
         }
         result.setValid(true);
+        // The match preview materializes a Map over every interface/service the
+        // rule selects, so only build it when the caller explicitly asks. The
+        // save path leaves preview false and pays only for validateRule above.
+        if (request == null || !request.isPreview()) {
+            return result;
+        }
         try {
             final Map<InetAddress, Set<String>> map = filterDao.getIPAddressServiceMap(rule);
             result.setMatchCount(map.size());
@@ -431,6 +440,7 @@ public class NotificationConfigRestService extends OnmsRestService {
     @XmlRootElement(name = "rule-request")
     public static class RuleRequestDTO {
         private String m_rule;
+        private boolean m_preview;
 
         @XmlElement(name = "rule")
         public String getRule() {
@@ -439,6 +449,18 @@ public class NotificationConfigRestService extends OnmsRestService {
 
         public void setRule(final String rule) {
             m_rule = rule;
+        }
+
+        // When false (the default, used by the save path), the endpoint only
+        // validates the rule; the caller must opt in to the match preview, which
+        // materializes the full interface/service map.
+        @XmlElement(name = "preview")
+        public boolean isPreview() {
+            return m_preview;
+        }
+
+        public void setPreview(final boolean preview) {
+            m_preview = preview;
         }
     }
 
@@ -533,9 +555,11 @@ public class NotificationConfigRestService extends OnmsRestService {
         }
         try {
             FilterDaoFactory.getInstance().validateRule(notification.getRule().getContent());
-        } catch (final Exception e) {
+        } catch (final FilterParseException e) {
             throw getException(Status.BAD_REQUEST, "The rule is not a valid filter: {}", e.getMessage());
         }
+        // Infrastructure failures (uninitialized factory, DB outage) are not the
+        // caller's fault; let them propagate to a 500 rather than a 400.
         if (isBlank(notification.getDestinationPath())) {
             throw getException(Status.BAD_REQUEST, "The event notification requires a destinationPath");
         }
--- a/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/NotificationConfigRestService.java
+++ b/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/NotificationConfigRestService.java
@@ -21,14 +21,19 @@
  */
 package org.opennms.web.rest.v1;
 
+import java.net.InetAddress;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 import javax.servlet.http.HttpServletRequest;
 
 import javax.ws.rs.Consumes;
+import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
+import javax.ws.rs.POST;
 import javax.ws.rs.PUT;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
@@ -39,11 +44,18 @@ import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
 import javax.ws.rs.core.SecurityContext;
 import javax.xml.bind.annotation.XmlAttribute;
+import javax.xml.bind.annotation.XmlElement;
 import javax.xml.bind.annotation.XmlRootElement;
 
+import org.opennms.core.utils.InetAddressUtils;
 import org.opennms.netmgt.config.DestinationPathFactory;
 import org.opennms.netmgt.config.NotifdConfigFactory;
+import org.opennms.netmgt.config.NotificationFactory;
+import org.opennms.netmgt.filter.FilterDaoFactory;
+import org.opennms.netmgt.filter.api.FilterDao;
 import org.opennms.netmgt.config.destinationPaths.DestinationPaths;
+import org.opennms.netmgt.config.notifications.Notification;
+import org.opennms.netmgt.config.notifications.Notifications;
 import org.opennms.netmgt.events.api.EventProxy;
 import org.opennms.netmgt.model.events.EventBuilder;
 import org.opennms.web.api.Authentication;
@@ -65,6 +77,8 @@ import io.swagger.v3.oas.annotations.tags.Tag;
  * <ul>
  * <li><b>GET /notification-config/status</b> global notifd on/off status</li>
  * <li><b>PUT /notification-config/status</b> turn notifd on or off</li>
+ * <li><b>GET/POST/PUT/DELETE /notification-config/event-notifications[/{name}]</b> event notification CRUD (notifications.xml)</li>
+ * <li><b>PUT /notification-config/event-notifications/{name}/status</b> toggle one event notification</li>
  * <li><b>GET /notification-config/destination-paths</b> all destination paths</li>
  * <li><b>GET /notification-config/destination-paths/{name}</b> one destination path</li>
  * </ul>
@@ -75,6 +89,8 @@ import io.swagger.v3.oas.annotations.tags.Tag;
 public class NotificationConfigRestService extends OnmsRestService {
 
     private static final Logger LOG = LoggerFactory.getLogger(NotificationConfigRestService.class);
+
+    private static final int RULE_PREVIEW_LIMIT = 250;
 
     @Autowired
     @Qualifier("eventProxy")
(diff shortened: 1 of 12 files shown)


joseanesONMS:

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/NotificationConfigRestService.java
```diff
@@ -200,6 +338,37 @@ private static NotifdConfigFactory getNotifdConfigFactory() throws Exception {
...
+    private static NotificationFactory getNotificationFactory() throws Exception {
+        // NotificationFactory's constructor requires an initialized NotifdConfigFactory
+        NotifdConfigFactory.init();
+        NotificationFactory.init();
+        return NotificationFactory.getInstance();
+    }
+
+    /**
+     * Rejects payloads that would fail the schema-validated marshal inside the
+     * factories: the managers remove the old entry from the in-memory config
```
marshallmassengill: replaceNotification is an update path that mutates in place. Mischaracterization at a minimum so either adjust the comment here or adjust code.  I suspect its worth addressing the code.

marshallmassengill (changes requested):
Mostly ok but a couple things:

Deleting the last event notification corrupts the running config. notifications.xsd requires at least one notification element. removeNotification drops the entry from memory and then calls saveCurrent(), which marshals with a schema attached. Memory ends up with zero notifications, notifications.xml still has one, the client gets a 500, and NotificationFactory.update() only reloads when the file mtime changes, which it didn't. notifd serves an empty notification config until OpenNMS restarts. The PR's validation can't catch this because DELETE has no body.

status is missing from validateEventNotification. replaceNotification overwrites the live object field by field through setters that assert non-empty, in this order: writeable, name, description, uei, rule, destinationPath, noticeQueue, textMessage, subject, numericMessage, status, varbind, parameters, then save. A PUT omitting status therefore passes validation, then throws inside setStatus after five fields have already been written onto the live object. Nothing saves, nothing rolls back, and the next successful save of any other notification persists the half-updated entry. POST has the same exposure via addNotification.

joseanesONMS:
All addressed: deleting the last notification now 400s with a pointer to turn it off instead; validation covers status (and parameter names/values, which the in-place chain can also die on after clearing the live entry's parameters); and the comment now describes what addNotification and replaceNotification actually do. Tests cover the delete-last guard and the missing-status atomicity on both POST and PUT.

marshallmassengill (changes requested):
The delete guard's message still never reaches the user. notificationConfigService.ts:141 discards err.response.data and Delete is not disabled at length === 1, so the guard works but reads as an unexplained failure.

TAB_LOADERS['event-notifications'] returns literal true, and this PR makes it the default tab. All three siblings have this now, so it belongs in the base rather than fixed three times.

joseanesONMS:
@marshallmassengill — the review items are addressed on the current tip:

- **Deleting the last event notification corrupts the config.** `deleteEventNotification` now guards `getNotifications().size() <= 1` and returns 400 ("the last event notification cannot be deleted; turn it off instead") rather than emptying memory and then 500ing on the schema-validated save.
- **`status` missing from `validateEventNotification`.** Added — a PUT/POST without an `on`/`off` status is rejected before any asserting setter runs, so there's no half-written live object.
- **Delete-guard message never reaches the user.** `notificationConfigService` now surfaces `err.response.data`, so the 400 reason is shown instead of an unexplained failure.
- **`TAB_LOADERS['event-notifications']` returned literal `true`.** The loader now returns `Promise.all([...]).every(Boolean)`; the base `ensureTabLoaded` latches only on success, so a transient GET failure retries on the next visit.

marshallmassengill (changes requested):
Two blockers on this one still:

1. The dark-mode white dialog panel is still unfixed. The two commits cancel exactly (git diff 4842a005999^ bbe2ce6652c is empty), and no open-dark or .p-dialog rule exists in Notifications.vue, the components, or ui/src/styles. 46a0c0562e5 fixes a different thing, the generated-rule preview contrast. Second pass carrying this.
  2. Both rule-builder service pickers are unlabelled. OnmsMultiSelect declares neither labelId nor inputId, and PrimeVue's MultiSelect has no labelId either, so the value lands as a dead labelid attribute and <label for="rule-services"> associates with nothing. Unlike the OnmsSelect cases there is no correct route through the seam, so the wrapper needs inputId added.

One minor finding worth maybe addressing:

Two real bugs in the new rule builder. buildRule's paren guard only checks the first character, so an OR'd IP filter loses its grouping: (A) | (B) plus a service becomes (A) | (B) & (isX). And parseRule's /is[A-Za-z]/ "too advanced" guard is a substring test, so a filter containing dist, list or history gets refused by the Builder with a misleading note.

joseanesONMS:
@marshallmassengill — updated on the release-36.x-synced epic (`jira/NMS-20100` now merges `release-36.x`, and this branch merges the synced epic; CI green).

Review items addressed:
- Rule-builder pickers are labelled via a new `OnmsMultiSelect` `inputId` seam; the `OnmsSelect` controls use `inputId`.
- `buildRule` now groups the IP filter with a real single-enclosing-paren check (an OR'd filter keeps its grouping); `parseRule`'s service-clause guard is a word boundary, so filters containing `dist`/`list`/`history` are no longer refused.
- The two no-op dark-mode commits are dropped.

Also added interactive `?` help across the form (event-replacement token list, Advanced options, destination path) via a click-anchored popover, since the native title tooltip never fired on click/touch.

synqotik:
A number of initial comments re UI.

First, is this targeting `release-36.x` or `develop`? It looks like `jira/NMS-20100` has `pom.xml` etc. with version 37 instead of 36.0.3. You may need to rebase that branch onto `release-36.x`, probably will be a bunch of merge conflicts. Then rebase this onto `jira/NMS-20100`.

Remove any direct PrimeVue references, this should actually have failed the ui build. In `ui/src/components/AdminNotifications/EventNotificationEditorDialog.vue` remove any `IftaLabel`. You can use `FormField` with `label` and `hint` instead. See `src/components/SnmpConfiguration/SnmpConfigPairedFieldInputs.vue` for an example.

Any info icon buttons should be using Info instead of Question Mark icons. Can use `OnmsIconButton` with `InfoIcon`. See `src/components/Nodes/NodesTable.vue`, "Node Search Help" for an example.

I would change that "About Notices and Escalation" expansion panel to be a MessageDialog, like on other pages. Use the info icon button next to main `Notifications` title, again see other pages for the pattern.

Instead of the gear icon button for going to configuration, should be a button, `+ Configure Notifications` or `+ Configure`. In the future maybe we use a gear icon, but should do it everywhere.

Then, that should navigate to a new route / new set of pages like `admin/notifications-config`. Everything in `src/components/AdminNotifications/EventNotificationEditorDialog.vue` should be a new, separate page rather than a dialog.

Actually, not sure why the route is `admin/notifications`, can just be `notifications` and `notifications-config`.

The search inputs in `src/components/AdminNotifications/NotificationQueriesCard.vue` should be changed to `OnmsSearchInput` components.

Components should be moved to `components/Notifications` and `components/NotificationsConfig` instead of `components/AdminNotifications`.

The query-links in `src/components/AdminNotifications/NotificationQueriesCard.vue` should be changed to use `OnmsTabPanels`. Then put the "Check notices for user" and other search box in a row above that.

In "Configure Notification", "Event Notifications", table, Actions, please follow same pattern as elsewhere like in `NodesTable.vue`. In this case, can have Edit and Delete icon buttons (no text). I would say, Status column should be read only and have an `OnmsTag` (with severity color) for the Status (i.e. enabled/disabled). Then in Action column also have an icon button `...` with a dropdown containing "Enable" and "Disable". Even better, only have the one that's opposite of current status.

I need to create some notifications locally and test the table, but the above can be fixed first.

joseanesONMS:
@synqotik @marshallmassengill Ready for another look. Rebased onto the epic (now 36.0.3-SNAPSHOT on release-36.x) and applied the review feedback — replaced direct-PrimeVue `IftaLabel` with the `FormField` seam and swapped the `?` badges for info-icon buttons. The table is rebuilt to the read-only status-tag + icon-action + overflow-menu pattern with truncated UEIs. Build, lint and store tests are green locally; CI is running.

marshallmassengill (changes requested):
Still one blocker:

Every save materializes the entire inventory's IP/service map. validateRule (NotificationConfigRestService.java:390) always calls filterDao.getIPAddressServiceMap(rule) at :409 and only truncates the response at :413; RULE_PREVIEW_LIMIT caps what is returned, not the work or the heap. validateForm calls that endpoint on every save (EventNotificationEditorDialog.vue:546), and the default rule for a new notification is IPADDR IPLIKE *.*.*.*, so the common case builds a Map<InetAddress, Set<String>> over every interface and service on the box before discarding all but 250 entries. On a large install that turns each save into a full filter evaluation plus a large transient allocation in the webapp heap. Split validation from preview (a preview flag on the request, or a separate endpoint) so the save path only does filterDao.validateRule, which the server already runs again at :535 anyway.

marshallmassengill:
Some minor findings too:

2. Opening an unchanged notification rewrites its rule. :602 sets builder mode via parseRule(form.rule), which writes ipFilter; the watcher at :502 then fires and reassigns form.rule = buildRule(), wrapping the IP filter in parentheses. Open the IT's own fixture (IPADDR != '0.0.0.0'), change only the description, save, and the stored rule becomes (IPADDR != '0.0.0.0'). It is semantically identical but it is an unrequested config mutation and it makes notifications.xml diffs noisy. It is also inconsistent: a rule that already equals the default leaves ipFilter unchanged, the watcher never fires, and no parentheses are added.

  3. Client validation misses the parameter rule the server enforces. :623 keeps any row with a non-blank name regardless of value, but Parameter.setValue asserts non-empty (Parameter.java:66) and validateEventNotification rejects it with 400. A name with a blank value produces a server error plus a generic errors.general banner and no inline error on the offending row. Validate both cells, or drop rows where either is blank.

  4. Three props the new UI depends on are outside the seam's declared API. fluid on OnmsMultiSelect (:148, :162) is undeclared even though this PR edits that very wrapper to declare inputId, and dropdownMode="current" (:28) is undeclared on OnmsAutoComplete. They work only because the wrapper roots are the PrimeVue components with default inheritAttrs, which the onms-ui README explicitly puts outside the contract ("anything else that happens to fall through is not part of the contract"). completeOnFocus at least has precedent in two existing files; dropdownMode is new here.

  5. icon="pi pi-plus" puts a raw PrimeIcons class in app code (EventNotificationsTable.vue:8). OnmsButton declares no icon prop and the repo has OnmsIcon/OnmsIconButton for this. There is one precedent in containers/Notifications.vue, so this compounds an existing leak rather than starting one, but the PR body claims no direct-PrimeVue usage.

  6. Neither new read endpoint has IT coverage, and both fail silently in the UI. /services (:375) and /rule/validate (:390) are untested; the 403 test gained only GET event-notifications, leaving POST/PUT/DELETE//services//rule/validate without authz coverage. getNotificationServices returns null on error behind a snackbar, so a broken /services shows an empty Services picker with no other signal.

  7. Infrastructure failures in rule validation are reported as user input errors. :535 wraps FilterDaoFactory.getInstance().validateRule(...) so that a DB outage or an uninitialized factory returns 400 "The rule is not a valid filter: …". At :399 the same getInstance() sits outside the try and escapes the class's error mapping entirely.

  8. The rule builder emits is<ServiceName> without checking the name is a legal filter identifier. getServiceNames returns every row of the service table verbatim; a name containing a space produces isMy Service, which will not parse. Low likelihood, but the builder is the default mode.

  9. writeable?: string is the wrong type (types/notificationConfig.ts); Notification.setWriteable takes a Boolean (Notification.java:112). No runtime effect since save() only spreads it through, but it will mislead the next editor of that form.

joseanesONMS:
@synqotik @marshallmassengill Updated and green on the latest epic. Addressed the review:
- Blocker: `/rule/validate` now takes a `preview` flag — the save path only parses the rule; the interface/service map is built solely for the explicit Validate action.
- Opening an unchanged notification no longer rewrites its rule (build guarded during load, so the stored rule stays byte-identical).
- Blank parameter values are dropped client-side before save.
- `fluid`/`dropdownMode` are now declared on the seam wrappers (OnmsMultiSelect/OnmsAutoComplete) on the epic; `pi pi-plus` replaced with OnmsIcon.
- Rule builder skips service names that aren't legal filter identifiers; `writeable` typed as boolean.
- Rule validation maps infra failures to 500 (FilterParseException → 400) and gained IT + 403 coverage for `/services` and `/rule/validate`.
combined-builds is passing.

marshallmassengill (approved):
LGTM!

synqotik (changes requested):
The `NotificationConfigRestService` and test should be moved to `v2` instead of `v1`.

Also remove the `m_` prefixes in the DTOs, and remove any of the `XmlAttribute`, etc., we should just be using JSON for everything.

joseanesONMS:
@synqotik Addressed all three:

- **v1 → v2** — moved `NotificationConfigRestService` (and its IT) to `org.opennms.web.rest.v2`; it now serves under `/api/v2/notification-config` (auto-registered by the v2 component-scan). Decoupled it from the v1 `OnmsRestService` base — the read/write/update lock and the `getException` helpers it relied on now live on the service itself, so the concurrency behavior is unchanged. The UI service switched to the `v2` axios instance.
- **`m_` prefixes** — removed from all DTO fields (`NotificationStatus`, `RuleRequestDTO`, `RuleValidationDTO`, `RuleMatchDTO`).
- **JAXB** — dropped `@XmlRootElement/@XmlElement/@XmlAttribute`; the DTOs are plain POJOs now. Wire names are unchanged (already camelCase: `rule`, `valid`, `matchCount`, `ipAddress`, `services`, `status`), so the JSON contract is identical.

Verified the refactor compiles and `NotificationConfigRestServiceIT` passes (8/8) against a 37.x-equivalent local build; leaving the 36.x build to CI. This service is shared with the Destination Paths / Path Outages tabs, so I'll carry the same move into those next.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/NotificationConfigRestService.java
```diff
@@ -0,0 +1,611 @@
...
+@Tag(name = "Notification-config", description = "Notification Configuration API")
+public class NotificationConfigRestService {
+
+    private static final Logger LOG = LoggerFactory.getLogger(NotificationConfigRestService.class);
+
+    private static final int RULE_PREVIEW_LIMIT = 250;
+
+    @Autowired
+    @Qualifier("eventProxy")
+    protected EventProxy m_eventProxy;
```
synqotik: Remove `m_` prefix. Can probably get rid of the `@Qualifier` once that is done.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/NotificationConfigRestService.java
```diff
@@ -0,0 +1,611 @@
...
+
+    // Per-instance read/write/update lock serializing config mutations, carried
+    // over from the base REST service this was split out of when it moved to v2.
+    private final ReadWriteUpdateLock m_globalLock = new ReentrantReadWriteUpdateLock();
+    private final Lock m_readLock = m_globalLock.updateLock();
```
synqotik: Remove `m_` prefixes from these 3.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/NotificationConfigRestService.java
```diff
@@ -0,0 +1,611 @@
...
+                throw getException(Status.BAD_REQUEST, "Every parameter requires a non-empty name and value");
+            }
+        }
+        final org.opennms.netmgt.config.notifications.Varbind varbind = notification.getVarbind();
+        if (varbind != null && (isBlank(varbind.getVbname()) || isBlank(varbind.getVbvalue()))) {
+            throw getException(Status.BAD_REQUEST, "A varbind requires both a name and a value");
+        }
+    }
+
+    private static boolean isBlank(final String value) {
```
synqotik: Use `org.apache.commons.lang.StringUtils.isBlank()` instead.

synqotik (changes requested):
A couple more things to fix, otherwise looks good.

joseanesONMS:
@synqotik Done:
- Removed the `m_` prefix from the service fields (`eventProxy` and the lock fields).
- Dropped `@Qualifier("eventProxy")` — the field name now matches the bean, so `@Autowired` resolves it by name.
- Replaced the hand-rolled `isBlank` helper with `org.apache.commons.lang.StringUtils.isBlank`.

Carried the same `m_`/`isBlank` cleanup into #8711 and #8712 since they share the service.

synqotik (approved):
LGTM!

synqotik:
@joseanesONMS looks good to merge!

PR #8711, opened 2026-07-30
NMS-20119: Destination Paths tab

joseanesONMS:
**Tickets:** NMS-20119 (epic NMS-20100).

Adds the Destination Paths tab to Configure Notifications, rebased onto the epic with the review feedback applied.

- Table lists destination paths with add/edit/delete.
- Editor covers the initial delay, initial targets and escalations, with per-target notification methods and interval.
- Fields use the FormField seam with inline info-icon help (no direct-PrimeVue IftaLabel), and the editor's field alignment is fixed.
- Backed by new NotificationConfig REST endpoints (with IT), a GroupManager role-map fix, and store/service/type support.

### DIFF
--- a/opennms-config/src/main/java/org/opennms/netmgt/config/DestinationPathManager.java
+++ b/opennms-config/src/main/java/org/opennms/netmgt/config/DestinationPathManager.java
@@ -60,7 +60,7 @@ public abstract class DestinationPathManager {
      * @param stream a {@link java.io.InputStream} object.
      * @throws IOException 
      */
-    protected void parseXML(final InputStream stream) throws IOException {
+    protected synchronized void parseXML(final InputStream stream) throws IOException {
         try (final InputStreamReader isr = new InputStreamReader(stream)) {
             allPaths = JaxbUtils.unmarshal(DestinationPaths.class, isr);
             oldHeader = allPaths.getHeader();
@@ -69,9 +69,16 @@ public abstract class DestinationPathManager {
     }
 
     private void initializeDestinationPaths() {
+        // Build the replacement map and swap the field reference in one
+        // assignment rather than clearing in place: a reload replaces memory
+        // with the file (instead of merging over it), while a reader holding an
+        // earlier unmodifiableMap keeps a consistent snapshot and saveCurrent()'s
+        // iteration is never emptied mid-flight.
+        final Map<String, Path> paths = new TreeMap<>();
         for (Path curPath : allPaths.getPaths()) {
-            m_destinationPaths.put(curPath.getName(), curPath);
+            paths.put(curPath.getName(), curPath);
         }
+        m_destinationPaths = paths;
     }
 
     /**
--- a/opennms-config/src/main/java/org/opennms/netmgt/config/GroupManager.java
+++ b/opennms-config/src/main/java/org/opennms/netmgt/config/GroupManager.java
@@ -144,11 +144,11 @@ public abstract class GroupManager implements GroupConfig {
         }
         buildDutySchedules(m_groups);
         
-        if (groupinfo.getRoles().size() > 0) {
-            m_roles = new LinkedHashMap<String, Role>();
-            for (final Role role : groupinfo.getRoles()) {
-                m_roles.put(role.getName(), role);
-            }
+        // always rebuild, so removing the last role from groups.xml doesn't
+        // leave the previous roles cached until restart
+        m_roles = new LinkedHashMap<String, Role>();
+        for (final Role role : groupinfo.getRoles()) {
+            m_roles.put(role.getName(), role);
         }
     }
 
--- a/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/NotificationConfigRestService.java
+++ b/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/NotificationConfigRestService.java
@@ -228,18 +228,35 @@ public class NotificationConfigRestService extends OnmsRestService {
             if (renamed && getDestinationPathFactory().getPath(path.getName()) != null) {
                 throw getException(Status.BAD_REQUEST, "Destination path {} already exists.", path.getName());
             }
+            final org.opennms.netmgt.config.destinationPaths.Path oldPath = getDestinationPathFactory().getPath(name);
             getDestinationPathFactory().replacePath(name, path);
             if (renamed) {
                 // keep event notifications pointing at the renamed path
-                boolean changed = false;
+                final List<Notification> touched = new ArrayList<>();
                 for (final Notification n : getNotificationFactory().getNotifications().values()) {
                     if (name.equals(n.getDestinationPath())) {
                         n.setDestinationPath(path.getName());
-                        changed = true;
+                        touched.add(n);
                     }
                 }
-                if (changed) {
-                    getNotificationFactory().saveCurrent();
+                if (!touched.isEmpty()) {
+                    try {
+                        getNotificationFactory().saveCurrent();
+                    } catch (final Exception notifSaveError) {
+                        // The path rename already persisted; undo it and revert the
+                        // in-memory notifications so destinationPaths.xml and
+                        // notifications.xml stay consistent instead of leaving
+                        // notifications pointing at a name that no longer exists.
+                        try {
+                            getDestinationPathFactory().replacePath(path.getName(), oldPath);
+                        } catch (final Exception rollbackError) {
+                            notifSaveError.addSuppressed(rollbackError);
+                        }
+                        for (final Notification n : touched) {
+                            n.setDestinationPath(name);
+                        }
+                        throw notifSaveError;
+                    }
                 }
             }
             return Response.noContent().build();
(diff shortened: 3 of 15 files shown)


joseanesONMS:

marshallmassengill (changes requested):
Deleting the last destination path still corrupts the config. NotificationConfigRestService.java:256 has a 404 check and no size guard. Reloading on save failure instead would also cover addPath and replacePath, which diverge the same way on any I/O failure.
TAB_LOADERS['destination-paths'] returns literal true. ConfigureNotificationsDialog.vue:95. All three fetches swallow errors, so one failure latches the tab and leaves the pickers empty for the page's life. **Third of three siblings with this bug, which argues for fixing the contract in the base rather than in each tab.**

Commands outside ENABLED_METHODS can be removed but not re-added, so editing a legacy textPage target is one-way. And getOnCallRoles actually depends on the GroupManager change, so the merge plan's "split it out, it runs in parallel" is wrong; #8711 needs it first.

joseanesONMS:
@marshallmassengill — addressed the review:

- **Deleting the last destination path / save-failure divergence.** `addPath`/`replacePath`/`removePath` mutate memory then `saveCurrent()`, so a failed save left memory ahead of the file until restart. All three now reload the factory from disk in their failure path (12904c3f163). Verified by `NotificationConfigRestServiceIT` (lifecycle, invalid-update-keeps-existing, rename).
- **`TAB_LOADERS['destination-paths']` returned literal `true`.** It now returns `Promise.all([...]).every(Boolean)`; the base `ensureTabLoaded` latches only on success, so a failed GET retries.
- **Legacy commands outside `ENABLED_METHODS` were one-way.** `originalLegacyCommands` keeps a target's pre-existing legacy commands selectable for the edit, so removing one can be undone.
- **`getOnCallRoles` depends on the GroupManager change.** This PR now carries it (77610ca1af8): `initializeGroupsAndRoles` always rebuilds `m_roles`, so removing the last role isn't left cached. `testOnCallRolesFollowFileChanges` covers exactly this; the full IT is 10/10 green locally.

marshallmassengill (changes requested):
This has the same major issue as #8710 and one other thing:

The Notification Methods control is still unlabelled. TargetRowEditor.vue:55 passes :labelId to OnmsMultiSelect, which declares neither labelId nor inputId, and PrimeVue's MultiSelect has no labelId either. #8710 has the identical defect on its two rule-builder pickers, so one inputId addition to the wrapper fixes both.

joseanesONMS:
@marshallmassengill — updated on the release-36.x-synced epic (`jira/NMS-20100` now merges `release-36.x`; CI green).

Review items:
- The delay, target-type and target-name `OnmsSelect` controls now use `inputId`.
- The two no-op dark-mode commits are dropped.
- The Notification Methods `OnmsMultiSelect` label will follow once #8710's `OnmsMultiSelect` `inputId` seam lands in the epic and this rebases — per your sequencing.

New UX: a new path defaults Initial Delay to 30s; the methods with no supported config path (IRC, the Microblog variants, XMPP message/group) are hidden from the picker (a legacy path already using one keeps it); and interactive `?` help on Initial Delay, Escalate After and the per-target Interval.

joseanesONMS:
@synqotik @marshallmassengill Ready for another look. Rebased onto the epic (now 36.0.3-SNAPSHOT on release-36.x) and applied the review feedback — replaced direct-PrimeVue `IftaLabel` with the `FormField` seam and swapped the `?` badges for info-icon buttons. The destination-path editor's field alignment is fixed. Build, lint and store tests are green locally; CI is running.

joseanesONMS:
@synqotik @marshallmassengill Updated and green on the latest epic. The OnmsMultiSelect labelling fix lands once on the epic (the wrapper now declares/forwards `inputId`), and the Notification Methods control here uses `:inputId` so its label associates. UI CI is passing.

marshallmassengill (changes requested):
. reloadDestinationPathsQuietly() doesn't do what its comment claims. DestinationPathManager.initializeDestinationPaths() (DestinationPathManager.java:71) put()s into m_destinationPaths without clearing it, so reload() merges the file over memory instead of replacing it. A failed add leaves the un-persisted path in memory after the reload, and the next successful saveCurrent() writes it to disk. Only the delete case is actually repaired.
 
Deleting the last path still 500s with a raw JAXB message. destinationPaths.xsd:57 is minOccurs="1" and JaxbUtils.getMarshallerFor sets the schema, so the marshal throws. The reload now keeps memory consistent, but the size guard from the last round never landed, and deleteDestinationPath in notificationConfigService.ts is the one mutation that discards the server's text/plain detail, so the user gets "Failed to delete destination path 'X'." with no reason.
 
A rename that fails at the notifications step reports failure after committing. replacePath has already persisted by the time NotificationFactory.saveCurrent() runs; if that throws, the endpoint returns 500, store.updateDestinationPath returns false and skips the table refresh, so the UI keeps showing the old name while disk has the new one — and the in-memory Notification objects now carry a destination path that isn't in notifications.xml.

joseanesONMS:
@marshallmassengill Addressed all three:
1. **Reload merges instead of replaces** — `DestinationPathManager.initializeDestinationPaths()` now clears `m_destinationPaths` before repopulating, so `reload()` replaces memory with the file; a failed mutation's entry no longer survives a reload to be re-persisted.
2. **Deleting the last path** — the endpoint now rejects it with a 400 ("…the only one configured; at least one must remain") instead of 500ing on schema validation, and `deleteDestinationPath` in the service surfaces the server's message. Added an IT (`testDeleteLastDestinationPathRejected`).
3. **Rename partial-commit** — if the follow-up `NotificationFactory.saveCurrent()` throws, the persisted path rename is now rolled back and the in-memory notifications reverted, so `destinationPaths.xml` and `notifications.xml` can't diverge.

Verified the destination-path ITs (lifecycle, invalid-update, last-path) pass locally on 37.x; pushed the same to the 36.x branch.

marshallmassengill (changes requested):
The clear() in initializeDestinationPaths is the wrong shape of fix. Every mutator on DestinationPathManager is synchronized (addPath, replacePath, removePath, saveCurrent), but reload(), parseXML and initializeDestinationPaths take no lock — so the one code path that now structurally removes entries is the only one that runs unsynchronized. Two consequences:

  - saveCurrent() iterates m_destinationPaths.values() under its own monitor; a concurrent reload can clear the map mid-iteration, which either throws ConcurrentModificationException or, worse, marshals and persists a truncated destinationPaths.xml. The pre-fix merge behaviour could only ever add an entry, so this failure mode is new.
  - There is now a window where the map is empty. NotificationConfigProviderImpl.getDestinationPathNames() and the GET /destination-paths handler both iterate the live view returned by getPaths(), and BroadcastEventProcessor resolves paths per event.

That window is on the common path, not a rare one: saveCurrent() never refreshes m_lastModified, so the next read after any successful mutation always triggers a full reload. Build a local TreeMap and assign it to the field instead of mutating in place — readers holding an earlier unmodifiableMap keep a consistent snapshot — and mark parseXML synchronized, matching GroupManager.parseXml.

joseanesONMS:
@marshallmassengill Good catch on the concurrency — reworked per your suggestion. `initializeDestinationPaths()` now builds a fresh `TreeMap` and swaps the field reference in one assignment instead of clearing in place, so a reader holding an earlier `unmodifiableMap` keeps a consistent snapshot and `saveCurrent()`'s iteration is never emptied mid-flight (no truncated `destinationPaths.xml`, no empty-map window). `parseXML` is now `synchronized` to match `GroupManager.parseXml`. The last-path-delete guard and the rename rollback are unchanged. Destination-path ITs (lifecycle, invalid-update, last-path) pass against the reworked manager.

marshallmassengill (approved):
No blockers at this point. LGTM!

synqotik (changes requested):
Same comments as 20118: Move Rest service to `v2`, remove `m_` prefixes, remove any `XmlAttribute` or similar.

joseanesONMS:
@synqotik Applied the same v2 move as #8710 here:

- **v1 → v2** — moved `NotificationConfigRestService` (and its IT) to `org.opennms.web.rest.v2` (served at `/api/v2/notification-config`, auto-registered by the v2 component-scan). Decoupled from the v1 `OnmsRestService` base — its read/write/update lock and `getException` helpers now live on the service itself, so the concurrency is unchanged. The UI service switched to the `v2` axios instance.
- **`m_` prefixes** removed from the DTO fields.
- **JAXB** (`@XmlRootElement/@XmlAttribute`) removed — plain POJOs; wire names unchanged.

Same 36.x local-build constraint as #8710, so leaving the build/IT to CI (the identical refactor compiled + passed its IT on the 37.x-equivalent).

joseanesONMS:
@synqotik Applied the same follow-up as #8710: removed the `m_` prefix from all service fields, dropped the redundant `@Qualifier`, and switched the null-or-blank checks to `org.apache.commons.lang.StringUtils.isBlank`.

joseanesONMS:
CI caught a compile error in my last push: a leftover `NotificationConfigRestService::isBlank` method reference pointed at the local `isBlank` helper I'd removed. Switched it to `StringUtils::isBlank`. (#8710 and #8712 don't have that reference; #8710's build was already green.)

PR #8712, opened 2026-07-30
NMS-20120: Path Outages tab

joseanesONMS:
**Tickets:** NMS-20120 (epic NMS-20100).

Adds the Path Outages tab to Configure Notifications, rebased onto the epic with the review feedback applied.

- Critical-path editor with a filter-rule preview and an apply/clear flow (a blank IP clears the path for matching nodes).
- Fields use the FormField seam with inline info-icon help; the interactive area is width-capped so it no longer overflows horizontally.
- Backed by new NotificationConfig REST endpoints (with IT) plus store/service/type support.

### DIFF
--- a/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/NotificationConfigRestService.java
+++ b/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/NotificationConfigRestService.java
@@ -21,14 +21,22 @@
  */
 package org.opennms.web.rest.v1;
 
+import java.net.InetAddress;
 import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.SortedMap;
 
 import javax.servlet.http.HttpServletRequest;
 
 import javax.ws.rs.Consumes;
+import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
+import javax.ws.rs.POST;
 import javax.ws.rs.PUT;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
@@ -42,10 +50,20 @@ import javax.xml.bind.annotation.XmlAttribute;
 import javax.xml.bind.annotation.XmlRootElement;
 
 import org.opennms.netmgt.config.DestinationPathFactory;
+import org.opennms.core.criteria.CriteriaBuilder;
+import org.opennms.core.utils.InetAddressUtils;
 import org.opennms.netmgt.config.NotifdConfigFactory;
 import org.opennms.netmgt.config.destinationPaths.DestinationPaths;
+import org.opennms.netmgt.dao.api.NodeDao;
+import org.opennms.netmgt.dao.api.PathOutageDao;
+import org.opennms.netmgt.dao.api.SessionUtils;
 import org.opennms.netmgt.events.api.EventProxy;
+import org.opennms.netmgt.filter.FilterDaoFactory;
+import org.opennms.netmgt.filter.api.FilterParseException;
+import org.opennms.netmgt.model.OnmsNode;
+import org.opennms.netmgt.model.OnmsPathOutage;
 import org.opennms.netmgt.model.events.EventBuilder;
+import org.springframework.dao.DataAccessException;
 import org.opennms.web.api.Authentication;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -67,6 +85,7 @@ import io.swagger.v3.oas.annotations.tags.Tag;
  * <li><b>PUT /notification-config/status</b> turn notifd on or off</li>
  * <li><b>GET /notification-config/destination-paths</b> all destination paths</li>
  * <li><b>GET /notification-config/destination-paths/{name}</b> one destination path</li>
+ * <li><b>GET/POST/DELETE /notification-config/path-outages</b> critical paths (pathoutage table)</li>
  * </ul>
  */
 @Component("notificationConfigRestService")
@@ -79,6 +98,20 @@ public class NotificationConfigRestService extends OnmsRestService {
     @Autowired
     @Qualifier("eventProxy")
     protected EventProxy m_eventProxy;
+
+    @Autowired
+    private PathOutageDao m_pathOutageDao;
+
+    @Autowired
+    private NodeDao m_nodeDao;
+
+    @Autowired
+    private SessionUtils m_sessionUtils;
+
+    // Flush/clear the Hibernate session every N rows during a bulk path-outage
+    // apply so a rule that matches many nodes doesn't accumulate the whole batch
+    // in the persistence context.
+    private static final int PATH_OUTAGE_BATCH = 200;
 
     @XmlRootElement(name = "notification-status")
     public static class NotificationStatus {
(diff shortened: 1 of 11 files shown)


joseanesONMS:

marshallmassengill (changes requested):
Minor stuff on this one... it's actually shockingly self-contained.

- **Tab opens pre-armed to clear every node.** PathOutagesTab.vue:228-229. rule defaults to IPADDR IPLIKE *.*.*.*, blank criticalIp makes the button "Clear Critical Path". Start rule empty (as was the case with legacy behavior).
- **Confirmation count is nodes matching the rule, not nodes with a path.** PathOutagesTab.vue:250. Says 4823 when 3 rows delete. Count existing paths in the preview.
- **Tab loader returns literal true.** ConfigureNotificationsDialog.vue:95. Failed GET latches loaded, never retries. Return null from getPathOutages on error.
- Only raw JDBC in the REST module; PathOutageDao has full CRUD plus getAllCriticalPaths().
-  isBlank dead again. NotificationConfigRestService.java:230.

joseanesONMS:
@marshallmassengill — addressed:

- **Tab opened pre-armed to clear every node.** `rule` now defaults to empty, so a freshly opened tab isn't a "Clear Critical Path" over `*.*.*.*` with a blank critical IP.
- **Confirmation count was rule-matches, not affected paths.** When clearing (blank critical IP) the count is now the existing critical paths among the matched nodes (`pathOutages` filtered by the matched nodeIds), not every node matching the rule.
- **Tab loader returned literal `true`.** It now returns `await store.getPathOutages()` (null on error), so a failed GET isn't latched and retries.
- **Dead `isBlank`.** Removed; the path-outage validation uses `String.isBlank()` directly.

marshallmassengill (changes requested):
I had to root cause this one a bit after the other two because it wasn't fully clear to me what was going on so this feedback is a bit deeper and points to where to possibly address the issues and it may impact the other PRs that are still open.

1. OnmsMultiSelect cannot be labelled at all

The issue. OnmsMultiSelect declares neither labelId nor inputId, and PrimeVue's MultiSelect only has inputId. So :labelId on it lands as a dead labelid DOM attribute and the paired <label for=...> associates with nothing. Two places hit this: #8710's rule-builder pickers (EventNotificationEditorDialog.vue:133,147) and #8711's Notification Methods picker (TargetRowEditor.vue:55).

Where to fix it. ui/packages/onms-ui/src/components/OnmsMultiSelect.vue needs inputId added and forwarded. That is a shared-package change, so it should not land twice. My preference is a small standalone seam PR under NMS-20081, because #8710 already carries one seam change (OnmsAutoComplete's #option slot) and I flagged that as something a feature PR should not be doing. If the ceremony is not worth it, put it in #8710 since that lands first, and have #8711 consume it on rebase.

2. labelId on OnmsSelect is a fallthrough, not a defect

The issue. OnmsSelect declares inputId. Passing labelId works only because it falls through to PrimeVue's Select, which does declare it. The label does get associated, so nothing is broken for users. It is the same $attrs reliance the base removed in 5d2d97719b5.

Where to fix it. Consumer side only, no seam change: rename to inputId in each PR that has it. #8710 has two, #8711 has three, #8712 has one. Cheap, independent, no coordination needed.

------------------ The below is probably for @synqotik awareness (if nothing else):

3. The dark dialog panel is not a missing rule, and I would not fix it in any of these PRs

The issue. This is the one I stated badly. #8710 and #8711 each added a hard-coded #15182b rule and then withdrew it, netting to zero, so the epic carries four commits that do nothing. But when I traced the theming, the infrastructure is already correct: primevue-setup.ts sets darkModeSelector: '.open-dark', themeService.applyThemeClass puts the class on both <html> and <body> (with a comment explaining why <html> matters for PrimeVue's :root variable references), and opennms-preset.ts already defines overlay.modal.background: '#15182B' in the dark colour scheme. That is the exact value both band-aids hard-coded, and a body-teleported dialog sits inside .open-dark, so the token should reach it.

So either the symptom has a narrower cause nobody has identified, or it was observed under conditions where the theme class was not yet applied. Two authors independently wrote and withdrew the same rule, which suggests neither root-caused it.

Where to address it. Not in these three PRs. Drop all four commits. If the white panel is real, it affects every OnmsDialog in the app rather than these tabs, so it belongs against the theme or the seam (opennms-preset.ts or OnmsDialog.vue) with a reproduction attached. I cannot confirm the symptom from code, and checking it visually is yours rather than mine.

joseanesONMS:
@marshallmassengill — updated on the release-36.x-synced epic (`jira/NMS-20100` now merges `release-36.x`; CI green).

Earlier review items remain addressed (empty-rule default, tab-loader retry, affected-path count, dead `isBlank`), and the `OnmsSelect` uses `inputId`.

New: the Critical Path IP validates client-side (marked invalid/red and Apply blocked unless it is a valid IPv4/IPv6 or blank), and the Node Filter Rule is marked invalid when the backend cannot evaluate it on Preview/Apply, plus a `?` popover explaining the filter syntax.

joseanesONMS:
@synqotik @marshallmassengill Ready for another look. Rebased onto the epic (now 36.0.3-SNAPSHOT on release-36.x) and applied the review feedback — replaced direct-PrimeVue `IftaLabel` with the `FormField` seam and swapped the `?` badges for info-icon buttons. The interactive area is width-capped so it no longer overflows horizontally. Build, lint and store tests are green locally; CI is running.

joseanesONMS:
@synqotik @marshallmassengill Updated and green on the latest epic. This tab's controls already associate via `inputId`, and it now sits on the epic that carries the shared seam fix; the withdrawn dark-mode commits are gone. combined-builds is passing.

marshallmassengill (changes requested):
A couple issues for this one:

The clear-path confirmation count is computed from the 200-node-capped preview list, so clearing a rule that matches thousands of nodes confirms "200 nodes" and then clears all of them.

raw JDBC instead of PathOutageDao (carried over from your first round); FilterDaoFactory.setInstance(mock(…)) never restored, leaking a mock into later ITs in the fork; ~3 full rule evaluations per Apply; uncapped delete+insert batch under writeLock(); unmapped 500 from previewPathOutageRule on a DataAccessException; zero-match apply reports success; deletePathOutage discards the server's detail and returns 204 for a missing row; dead scoped .field-error rule; isValidIp rejects ::ffff:192.168.1.1 and accepts :::::; criticalSvc unvalidated server-side.

joseanesONMS:
@marshallmassengill Addressed the batch (full `NotificationConfigRestServiceIT` — 21 tests — passes locally):
- **Raw JDBC → PathOutageDao/NodeDao** in `SessionUtils` transactions (no more `DataSourceFactory`/`writeLock` for path outages); read via `findAll`, delete via `get`+`delete`, apply via transactional `save` with periodic flush/clear (bounds the batch).
- **Clear count**: the confirmation no longer shows the preview-capped number — it reads "every node matching the rule that currently has one".
- **~3 rule evals → 1** per apply (dropped the redundant `validateRule`; `getNodeMap` parses).
- **DataAccessException** from rule evaluation now maps to 500, not an unmapped stack.
- **Zero-match apply → 400**; **non-ICMP service → 400**; **delete of an unconfigured path → 404** (and the UI surfaces the server message).
- **isValidIp** fixed (`::ffff:192.168.1.1` valid, `:::::` rejected); dead `.field-error` CSS removed; the IT restores `FilterDaoFactory` in @After and seeds its node through Hibernate so the DAO path is actually exercised.

New ITs cover the zero-match/non-ICMP/404 cases.

marshallmassengill (changes requested):
1. criticalSvc is validated case-insensitively but stored verbatim. !"ICMP".equalsIgnoreCase(criticalSvc) lets "icmp" through, and the trimmed original is what reaches new OnmsPathOutage(node, criticalIp, criticalSvc). Pollerd tolerates that — DefaultPollContext.testCriticalPath also compares with equalsIgnoreCase and pings regardless — but PathOutageDaoHibernate joins with pathOutage.criticalPathServiceName = serviceType.name, which is case-sensitive. So a lowercase row silently drops out of getNodesForPathOutage, getNodesInPath and the dependency views built on them, while suppression keeps working. Store the literal "ICMP" rather than the user's input.
 
2. getPathOutages looks like an N+1. m_pathOutageDao.findAll() followed by po.getNode().getLabel() per row, against an @OneToOne @PrimaryKeyJoinColumn association — Hibernate typically  issues a separate select per row for that mapping, where the SQL this replaced was one LEFT JOIN. Worth confirming with SQL logging on an install with a few hundred critical paths. Related and more consequential at scale: the apply loop is now m_pathOutageDao.get(nodeId) plus m_nodeDao.get(nodeId) plus a save per node, where the previous version issued two batched statements for the whole set. The flush()/clear() every 200 keeps the persistence context bounded, but IPADDR IPLIKE *.*.*.* on a large estate is now three round trips per node inside a single transaction. That's the cost of the DAO conversion you asked for, so it may well be the right trade — it's just worth being deliberate about rather than discovering later.

Minor: InetAddressUtils.addr(InetAddressUtils.normalize(raw)) is addr(str(addr(raw))), since normalize already round-trips through addr. addr(raw) alone does the job and throws the same IllegalArgumentException.

joseanesONMS:
Addressed all three:

1. **Store canonical `ICMP`** — `applyPathOutage` now persists the literal `"ICMP"` regardless of request casing (validated case-insensitively, stored canonical). The path-outage joins compare `criticalPathServiceName` to `serviceType.name` case-sensitively, so a verbatim `icmp` would silently drop out of `getNodesForPathOutage`.
2. **N+1 / redundant round-trips** — `getPathOutages` now batch-fetches labels via a single `CriteriaBuilder(OnmsNode.class).in("id", nodeIds)` `findMatching` into a `Map<Integer,String>` instead of a per-row `getLabel()`. The apply loop updates in place (`update`) when a row already exists rather than delete+insert, and uses `nodeDao.load()` (proxy, no query) for the insert path.
3. **`addr(raw)`** — dropped the redundant `InetAddressUtils.normalize` wrap; `addr` already parses and canonicalizes.

marshallmassengill (changes requested):
One blocker: java.util.Set is not imported. NotificationConfigRestService.java:358 declares final Set<Integer> nodeIds = new HashSet<>();, and the import block adds HashMap and HashSet but not Set. Nothing else in the file imports it and there are no wildcard imports, so the module doesn't compile. That matches ci/circleci: build failing on this head while build-ui passes.

joseanesONMS:
Good catch — fixed. Added `import java.util.Set;` (the batch-label rewrite declared `Set<Integer> nodeIds` but only `HashSet` was imported). Re-verified every symbol referenced by the path-outage code resolves to an import.

marshallmassengill (approved):
LGTM!

synqotik (changes requested):
Same comments as 20118: Move Rest service to `v2`, remove `m_` prefixes, remove any `XmlAttribute` or similar.

joseanesONMS:
@synqotik Applied the same v2 move as #8710 here:

- **v1 → v2** — moved `NotificationConfigRestService` (and its IT) to `org.opennms.web.rest.v2` (served at `/api/v2/notification-config`, auto-registered by the v2 component-scan). Decoupled from the v1 `OnmsRestService` base — its read/write/update lock and `getException` helpers now live on the service itself, so the concurrency is unchanged. The UI service switched to the `v2` axios instance.
- **`m_` prefixes** removed from the DTO fields.
- **JAXB** (`@XmlRootElement/@XmlAttribute`) removed — plain POJOs; wire names unchanged.

Same 36.x local-build constraint as #8710, so leaving the build/IT to CI (the identical refactor compiled + passed its IT on the 37.x-equivalent).

joseanesONMS:
@synqotik Applied the same follow-up as #8710: removed the `m_` prefix from all service fields, dropped the redundant `@Qualifier`, and switched the null-or-blank checks to `org.apache.commons.lang.StringUtils.isBlank`.

PR #8713, opened 2026-07-30
NMS-20106: versioned user management API and PrimeVue Manage Users page

joseanesONMS:
[NMS-20106](https://opennms.atlassian.net/browse/NMS-20106): a versioned user management API (`/api/v2/users`) so users can be provisioned by external tooling, and a PrimeVue Manage Users page that is a straight visualization of it. `users.xml` stays the system of record via the existing `UserManager`; fields the API does not expose (XMPP among them, deliberately hidden in the UI) and the password survive updates untouched.

- Endpoints: list/get/create/update/set-password/rename/delete plus available-roles, admin-only via new Spring Security rules and in-code checks.
- Responses never contain the password hash in any form (the v1 API returns it to admins).
- The admin/rtc delete and rename protections and the admin ROLE_ADMIN-retention guard are enforced server-side; the legacy page only hid the buttons.
- Requests are validated up front (ids, roles, time zones, duty schedule grammar, comments markup) and applied to a detached copy, so a rejected request leaves no partial state, while hand-edited values are grandfathered so files never become uneditable.
- The page validates fields inline against the same rules and shows API rejections inside the dialogs.
- 22 integration tests (mock managers, `users.xml` never touched) plus Vitest store/dialog tests; verified end to end against a local instance including byte-level `users.xml` restoration.

[NMS-20106]: https://opennms.atlassian.net/browse/NMS-20106?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
- Replaces fork PR #8698 with an in-repo branch so CI runs.


joseanesONMS:

joseanesONMS:
@synqotik @marshallmassengill — ready for review. This page now uses the `@opennms/onms-ui` Onms-XXX wrappers instead of direct PrimeVue (Button, Dialog, InputText, Password, MultiSelect, DataTable, Column, Tag). `Message` and `IftaLabel` have no wrapper yet, so they stay on PrimeVue. `pnpm lint` is clean (ran `lint:fix`), and the `vue-tsc` build and the UI test suite pass. No behaviour change.

marshallmassengill (changes requested):
One blocker: Rename leaves on-call role supervisors dangling, which is the exact case delete guards against. deleteUser walks m_groupManager.getRoles() and returns 400 if the user supervises any role, with a comment noting that GroupManager.deleteUser "leaves role supervisors dangling". GroupManager.renameUser has the same hole: it rewrites group memberships and Schedule names but never touches role.getSupervisor() (GroupManager.java:498-528). So POST /api/v2/users/{id}/rename on a supervisor silently points the rota at a user id that no longer exists, killing the supervisor fallback with no error. Either apply the same 400 guard or fix up the supervisor as part of the rename.

joseanesONMS:
@marshallmassengill Addressed the blocker:

**Rename leaves on-call role supervisors dangling.** Fixed at the root in `GroupManager.renameUser` — it now rewrites `role.getSupervisor()` alongside the group memberships and schedule names it already rewrote, so a renamed supervisor follows the rota to the new id instead of pointing at a user that no longer exists. I went with the fix-up rather than the 400 guard: a rename preserves the user (unlike delete, where blocking is the only option), so the reference should follow it. Fixing it in the manager also covers the legacy JSP rename path. New IT `testRenameCarriesOnCallRoleSupervisor` asserts the supervisor carries over (full `UsersRestServiceIT` green locally).

Also brought the dialogs onto the current `@opennms/onms-ui` seam while here: dropped the nested `IftaLabel` for `FormField`'s own label/for/error, and replaced the `primevue/message` banners with a themed `role=alert` element — the ManageUsers dialogs now import no un-wrapped PrimeVue primitive. `eslint` clean, vitest + `vue-tsc` build pass.

ui/src/types/userAdmin.ts
```diff
@@ -0,0 +1,47 @@
...
+/// License.
+///
+
+// Wire shapes of the v2 user management API (/api/v2/users). Field names
+// follow users.xml; the password hash is never part of any response, and
+// contact types the API does not expose (XMPP among them) are preserved
+// server-side on update.
+
+export interface ManagedUser {
+  'user-id': string
```
synqotik: These should all be camelCase. REST API should also be camelCase JSON.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/model/UserDto.java
```diff
@@ -0,0 +1,152 @@
...
+import javax.xml.bind.annotation.XmlElement;
+import javax.xml.bind.annotation.XmlRootElement;
+
+/**
+ * A user as exposed by the v2 user management API. Field names mirror
+ * users.xml where they overlap; the password hash is deliberately never part
+ * of this representation, and contact types the API does not expose (XMPP,
+ * microblog, phones, pager PINs) are preserved server-side on update.
+ */
+@XmlRootElement(name = "user")
```
synqotik: This should all be pure JSON, camelCase, and no XML. If we get/save from/to XML, that mapping can be done inside the REST service.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/model/UserPasswordRequest.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package org.opennms.web.rest.v2.model;
+
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlElement;
+import javax.xml.bind.annotation.XmlRootElement;
+
+@XmlRootElement(name = "user-password-request")
+@XmlAccessorType(XmlAccessType.FIELD)
```
synqotik: Same as other comments, anything in `org.opennms.web.rest.v2.model` should be JSON and not XML.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/UsersRestService.java
```diff
@@ -0,0 +1,527 @@
...
+
+    private Response serverError(final String format, final Exception e) {
+        if (e instanceof IllegalArgumentException) {
+            return Response.status(Status.BAD_REQUEST).entity(e.getMessage()).build();
+        }
+        LOG.error(String.format(format, e.getMessage()), e);
+        return Response.status(Status.INTERNAL_SERVER_ERROR).entity(String.format(format, e.getMessage())).build();
+    }
+
+    private static boolean isBlank(final String value) {
```
synqotik: Should use Java `String.isBlank()` instead.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/UsersRestService.java
```diff
@@ -0,0 +1,527 @@
...
+        return value == null || value.isBlank();
+    }
+
+    private static String trimToNull(final String value) {
```
synqotik: Use `org.apache.commons.lang3.StringUtils.trimToEmpty` or similar.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/UsersRestService.java
```diff
@@ -0,0 +1,527 @@
...
+
+    @Override
+    public Response updateUser(final SecurityContext securityContext, final String userId, final UserDto dto) {
+        assertAdmin(securityContext);
+        if (dto == null) {
+            return Response.status(Status.BAD_REQUEST).entity("A user body is required.").build();
+        }
+        if (dto.getUserId() != null && !userId.equals(dto.getUserId())) {
+            return Response.status(Status.BAD_REQUEST)
+                    .entity("The user-id in the body does not match the request path; use the rename endpoint to change ids.").build();
```
synqotik: Here and elsewhere should probably be `userId` not `user-id` (after fixing everything to camelCase Json).

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/UsersRestService.java
```diff
@@ -0,0 +1,527 @@
...
+        if (dto.getUserComments() != null && INVALID_COMMENTS.matcher(dto.getUserComments()).find()
+                && (existing == null || !dto.getUserComments().equals(existing.getUserComments().orElse(null)))) {
+            throw new IllegalArgumentException("The comments must not contain any HTML markup.");
+        }
+        final String timeZoneId = trimToNull(dto.getTimeZoneId());
+        if (timeZoneId != null) {
+            try {
+                java.time.ZoneId.of(timeZoneId);
+            } catch (final RuntimeException e) {
+                throw new IllegalArgumentException("Invalid time-zone-id: " + timeZoneId);
```
synqotik: -> camelCase

joseanesONMS:
@synqotik Addressed the API-shape feedback:

- **camelCase JSON, no XML** — dropped the JAXB annotations from `UserDto`, `UserWriteRequest`, `UserPasswordRequest`, and `UserRenameRequest`, so they serialize as plain camelCase JSON (`userId`, `fullName`, `pagerEmail`, `userComments`, `dutySchedules`, `roles`, `readOnly`, `newUserId`) — matching the existing plain-POJO DTOs like `Snmpv3UserDto`. The users.xml mapping already lives in the REST service.
- **UI camelCase** — `ui/src/types/userAdmin.ts` and everything downstream (service, store, dialogs, table, tests) now use the camelCase fields.
- **`userId` not `user-id`** in the REST messages, and the two Java nits: `isBlank`/`trimToNull` now go through `org.apache.commons.lang3.StringUtils`.

Verified: full `UsersRestServiceIT` (23) green with the new wire format, UI eslint clean, 24 vitest, `vue-tsc` build passes.

I'm applying the same camelCase-JSON convention to #8714 (groups) and auditing the notification PRs' v2 DTOs for consistency.

ui/src/containers/ManageUsers.vue
```diff
@@ -0,0 +1,54 @@
+<template>
+  <div class="onms-row">
+    <div class="onms-col-12">
+      <BreadCrumbs :items="breadcrumbs" />
+    </div>
+  </div>
+  <div class="manage-users-container">
+    <h1 class="page-title">Manage Users</h1>
+    <UsersHelpPanel />
```
synqotik: Instead of this being an expandable panel, consider using established pattern of an `OnmsIconButton` with an Info icon which pops up an `OnmsDialog` with the help info.

We should probably discuss best UX pattern going forward, but I'd like to be consistent in how contextual help shows up.

ui/src/components/ManageUsers/UsersTable.vue
```diff
@@ -0,0 +1,244 @@
...
+      :value="store.users"
+      paginator
+      dataKey="user-id"
+      :rows="10"
+      :rowsPerPageOptions="[10, 20, 50, 100]"
+      class="data-table"
+      data-test="users-table"
+    >
+      <OnmsColumn
+        field="user-id"
```
synqotik: Note, the `field` props will all need to change to `camelCase` once this is changed in `types/userAdmin`

ui/src/components/ManageUsers/UsersTable.vue
```diff
@@ -0,0 +1,244 @@
...
+        header="Email"
+        sortable
+      />
+      <OnmsColumn
+        field="pager-email"
+        header="Pager Email"
+      />
+      <OnmsColumn header="Roles">
+        <template #body="{ data }">
+          <span class="roles">{{ (data.role ?? []).join(', ') || '-' }}</span>
```
synqotik: Maybe `'--'` instead of just `'-'` ?

ui/src/components/ManageUsers/UsersTable.vue
```diff
@@ -0,0 +1,244 @@
...
+        </template>
+      </OnmsColumn>
+      <OnmsColumn header="Actions">
+        <template #body="{ data }">
+          <span
+            v-if="!isPathAddressable(data['user-id'])"
+            class="unaddressable"
+            v-tooltip.top="'This user-id contains / \\ or %, which the API cannot address; manage it by editing users.xml.'"
```
synqotik, line 69: May need to be `v-onms-tooltip`

ui/src/components/ManageUsers/UsersTable.vue
```diff
@@ -0,0 +1,244 @@
...
+            data-test="unaddressable-note"
+          >file-managed</span>
+          <div v-else class="action-container">
+            <OnmsButton
```
synqotik: Since there are multiple operations, I think we want something similar to other patterns we have. Can have `OnmsIconButton` for Edit and Delete, then an ellipse button with a dropdown for other operations like Change password, rename, etc.

ui/src/components/ManageUsers/UsersTable.vue
```diff
@@ -0,0 +1,244 @@
...
+
+    <div v-if="!store.users.length">
+      <EmptyList
+        :content="emptyListContent"
+        data-test="empty-list"
+      />
+    </div>
+  </TableCard>
+
+  <UserEditorDialog
```
synqotik, line 114: I need to take a look at this, i.e. popup dialogs vs. just having a separate "page".

joseanesONMS:
Thanks @synqotik — went through all six:

**Fixed now (pushed):**
- **Column `field` props** (UsersTable:25/41/51) — good catch. `fullName` and `pagerEmail` still bound to the old hyphenated names after the camelCase pass, so those columns rendered blank; fixed.
- **`--` for empty roles** (:56) — done.

**Needs the seam, not on this branch:**
- **`v-onms-tooltip`** (:64) — the OnmsTooltip directive (NMS-20054) isn't present on `develop` yet; `v-tooltip` here resolves to PrimeVue's `Tooltip`, registered globally in `theme/primevue-setup.ts`. Switching to `v-onms-tooltip` now would be an unknown directive. Happy to flip it the moment that seam lands or this branch picks it up.

**Design points — agree, and they're cross-cutting, so let's align before I change them:**
- **Help panel → Info-icon + OnmsDialog** (ManageUsers:9) — the same `About` TogglePanel pattern is on Groups and the notification config pages too, so I'd rather change it once, consistently. Happy to.
- **Row actions → Edit/Delete + ellipsis dropdown** (:68) — same story; this is a shared admin-table pattern (users, groups). Let's settle the shape and I'll apply it across all of them.
- **Dialogs vs. a separate page** (:112) — your call on direction; I can go either way.

I left these three as-is pending your steer rather than guessing. The camelCase-JSON convention is also applied to #8714 (groups) now.

joseanesONMS:
@synqotik Implemented the three UI patterns we discussed, on Manage Users as the template:

- **Help** — the inline About panel is gone; there's now an Info-icon button in the table header that opens the help copy in an `OnmsDialog`. Extracted a shared `Common/AboutDialogButton.vue` (title + slotted content) so Groups and the notification About panels can reuse it.
- **Row actions** — collapsed the four buttons to inline **Edit** and **Delete** plus a **overflow `OnmsMenu`** carrying **Change Password** and **Rename**. Protected-account disabling (admin/rtc) and the file-managed note for unaddressable ids are preserved.
- **Editor** — kept the modal dialogs, per your read.

eslint clean, full vitest suite (2191) green, `vue-tsc` build passes.

If this shape looks right to you, I'll mirror the same `AboutDialogButton` + overflow-menu pattern onto #8714 (Groups) and reuse the About button for the notification config pages. Holding those until you confirm so I build it once.

synqotik:
@joseanesONMS thanks changes look good!

One more thing, make same change to `opennms-webapp-rest/src/main/webapp/WEB-INF/menu/menu-template-default.json` as `opennms-webapp-rest/src/main/webapp/WEB-INF/menu/menu-template.json`, files should be identical.

ui/src/components/ManageUsers/UsersTable.vue
synqotik: These are ok for now. We may want a bigger discussion on overall UX re Drawers (which I don't really like...) vs. Dialogs vs. a separate page.

joseanesONMS:
Done — `menu-template-default.json` now carries the same `ui/index.html#/admin/users` URL, and the two files are byte-identical again. Pushed.

Noted on the Drawers/Dialogs/page UX — happy to have that broader discussion separately; leaving the dialogs as-is for now. I'll mirror the Info-dialog + overflow-menu patterns onto #8714 (Groups) next, now that the shape's confirmed.

synqotik (dismissed):
LGTM! Let's try this out and we can make adjustments as needed.

marshallmassengill (changes requested):
1. Every write re-HTML-escapes full name and all contact addresses, cumulatively. copyOf (UsersRestService.java:325) rebuilds the user through setFullName / Contact.setInfo, and both
run WebSecurityUtils.sanitizeString  Encode.forHtml, which is not idempotent. Verified against the real config-model classes:

stored:  O&#39;Brien &amp; Sons | bill&amp;ted@example.com save 1:  O&amp;#39;Brien &amp;amp; Sons | bill&amp;amp;ted@example.com save 2:  O&amp;amp;#39;Brien ...

So PUT /users/{id} and PUT /users/{id}/password mangle fullName, email, pagerEmail, and every unexposed contact (XMPP, phones, numeric/text pager PINs and service providers) one layer deeper each time, for any value containing & < > " '. O'Brien becomes O&#39;Brien in the table after a single unrelated role edit. This is new: the legacy JSP is stable at one layer because the browser decodes the entities out of the value= attribute before POSTing them back; a JSON round-trip doesn't. It also contradicts "the password survives updates untouched"-style claims in the PR body and in UsersAbout.vue ("preserved untouched"). Fix by cloning without the sanitizing setters (JAXB marshal/unmarshal uses field access, @XmlAccessorType(NONE) + field annotations) and only writing fullName/contacts when the DTO value actually differs from stored, the way userComments is already grandfathered. testUpdatePreservesUnexposedContacts misses it only because xmpp@jabber.example.com has nothing escapable.

2. Repointing the menu drops the only admin path to half the user record. modifyUser.jsp edits duty schedules, numeric/text paging service + PIN, work/mobile/home phone, telephone PIN,
and time zone; the new page edits none of them and neither does the API surface them for editing. After this change, "User Management  Manage Users" reaches the lesser editor, and the
full one survives only under Admin  "Configure Users, Groups and On-Call Roles"  "Configure Users". Duty schedules and pager contacts are load-bearing for notifd. Either keep the menu
on the JSP until parity, or add a link from the new page to the legacy editor for the uncovered fields.

joseanesONMS:
@marshallmassengill — both blockers addressed, plus the red build-ui check.

**Cumulative HTML-escaping of full name and contacts.** Root-caused to `copyOf` rebuilding the user through the config-model setters (`setFullName`, `Contact.setInfo`), which run a non-idempotent `Encode.forHtml` — so every write escaped the stored value one layer deeper, exactly as you traced. `copyOf` now clones via a JAXB round-trip (`User` is `@XmlAccessorType(NONE)` with field annotations, `Contact` is `FIELD`, so marshal/unmarshal uses field access and bypasses the sanitizing setters), and `applyDto` only writes full name / email / pager-email / the new contacts through the setter when the value actually changed. New IT stores `O'Brien & Sons` / `bill&ted@example.com` and asserts a password change and an unrelated edit leave them byte-for-byte unchanged.

**Menu routed to a lesser editor.** Rather than repoint the menu, I brought the new page to parity so it's no longer lesser: `UserDto` + the service now expose work/mobile/home phone and numeric/text pager service + PIN (mapped to the `workPhone`/`mobilePhone`/`homePhone` and `numericPage`/`textPage` contacts, info + serviceProvider), and the editor adds those fields plus telephone PIN, a time-zone picker, and an add/remove duty-schedule list. tui-pin, time-zone-id and duty schedules were already on the DTO; they just weren't surfaced. ITs cover create + partial-update round-trips of the new contacts; editor tests cover load/save and the duty-schedule list.

**build-ui** was failing on lint (arrow-parens/quote style) — fixed, `pnpm lint` clean.

`UsersRestServiceIT` green for the new coverage; UI eslint + vitest + `vue-tsc` build pass.

joseanesONMS:
@marshallmassengill friendly ping — the cumulative HTML-escaping fix (JAXB copy + change-only writes), the field-parity work (phones, pager service/PIN, telephone PIN, time zone, and a day/time duty-schedule editor), and the build-ui lint failure are all pushed and green. Ready for another pass when you have a moment.

marshallmassengill (dismissed):
No blockers.

joseanesONMS:
Merged current `develop` in to clear conflicts after #8714 landed — **mechanical, no logic changes**. Resolutions were all shared-wiring unions: both endpoints' rules in `applicationContext-spring-security.xml`, both routes in the Vue router, both service groups in `services/index.ts`, and the shared `AboutDialogButton` taking develop's copy. Also repointed `UsersTable`/`DutyScheduleEditor` icon imports to the `@opennms/onms-ui/icons/*` seam that develop migrated to. Lint, `vue-tsc` build, and the full user/group suite are green. The push dismissed the prior approval — @marshallmassengill could you re-approve when you have a moment? Conflicts are cleared and it's mergeable.

marshallmassengill (changes requested):
Couple items:

  1. UsersRestService.java:403 (high) — tuiPin is the only field written with no validation, but users.xsd constrains tui-pin to [0-9]+. A PUT with 1234x gets into m_users via _writeUser, then _saveCurrent's schema-validating marshal throws: 500, users.xml unwritten, and since mtime/size are unchanged UserFactory.isUpdateNeeded() stays false, so the bad user is pinned in memory and every later user create/update/delete fails until restart.
  2. UsersRestService.java:213 (medium) — updateUser/setPassword have no rollbackPhantomUser equivalent (which createUser does have at :173). A failed setPassword returns 500 with the new hash live in memory, so the user logs in with it until a restart silently reverts. Same at :235.
  3. DutyScheduleEditor.vue:134 (medium) — serialize() canonicalizes day order and strips zero-padding, and the mount-time watch emits, so merely opening the dialog rewrites MoTu2000-0800 to MoTu2000-800. The server grandfathers only byte-identical strings, so validateDutySchedule then rejects it and the admin can't save even an email-only edit. The existing test passes only because its fixture is already canonical.

joseanesONMS:
@marshallmassengill all three addressed in `39a756ddf7a`:

1. **tuiPin** — validated against the `users.xsd` `[0-9]+` pattern with a 400 before it can reach the in-memory map. The new IT covers both halves of your scenario: `1234x` is rejected AND a valid write still succeeds afterward, proving the map isn't poisoned.
2. **Rollback** — `updateUser` and `setPassword` now mirror `createUser`'s discipline: a failed save re-saves the previous state (`restorePreviousUser`), so memory and `users.xml` re-align and a "failed" password can no longer be used to log in.
3. **Duty-schedule rewrite-on-open** — parsed rows keep the incoming string and serialize it byte-identically until that row is actually edited (day toggle or time input), and the watcher no longer emits when serialization equals the incoming value — opening the dialog provably changes nothing. Tests use your exact shape (`TuMo2000-0800`): no emit on mount, untouched rows survive a neighbor's edit verbatim, canonicalization only on a real edit. You were right that the old test passed only because its fixture was canonical — the new ones aren't.

`UsersRestServiceIT` 25/26 locally — the one failure is `testRenameCarriesOnCallRoleSupervisor`, the known local-env quirk that's green in CI. Push carries `!smoke`.

marshallmassengill (approved):
No blockers for me at this point but hold off on merging until @synqotik has had a chance to review.

joseanesONMS:
Screenshot 2026-09-01 at 14 20 14
Screenshot 2026-09-01 at 14 20 07
Screenshot 2026-09-01 at 14 19 59
@synqotik @marshallmassengill

synqotik:
## Manage Users

This has been approved, and I'm ok with it being merged, but needs some follow up. Could be done before merging, or afterwards.

Instead of opening a dialog for Add User and Edit User, it should be a separate page. URL can route to: `ui/index.html#/admin/users/:username`.

There's just too much info for a dialog, it gets messy.

I'd have 3 tabs for the Add/Edit users page: General Information (everything up to and including Time Zone); Duty Schedule; Roles.

The Time Zone picker needs some work. I'd consider splitting it into 2 autocomplete dropdowns, the first one is the item to the left of the `/`, i.e. `Africa`, `America`, `Pacific` etc. The second one would be the city or area. But replace `_` with spaces and allow case-insensitive search.

The Duty Schedules are kind of clunky. After I add Time support to `OnmsDatePicker` you can use that with the `timeOnly` setting. Also once this is in a tab instead of a narrow dialog, could have the Day section in an `OnmsSelect` and expanded with full name of the day of week. You can have Day of Week, Start Time, End Time on a row, with an Add button at end, then a table of the items added with a Delete `OnmsIconButton` to remove.

## Manage Groups

I guess OK to have a dialog to add a new group on this one since there isn't too much info.

For Rename Group, I'd pre-populate it with the current group name.

I'd add a little space between Comments and Members. Might rename `Members` -> `Group Members`. I'd put the `Order matters...` text on a separate line.

PR #8714, opened 2026-07-30
NMS-20107: versioned group management API and PrimeVue Manage Groups page

joseanesONMS:
[NMS-20107](https://opennms.atlassian.net/browse/NMS-20107): a versioned group management API (`/api/v2/groups`) and a PrimeVue Manage Groups page that is a straight visualization of it; independent companion to #8698. `groups.xml` stays the system of record via the existing `GroupManager`, with member order (the notification escalation order) and `default-map` preserved end to end.

- Endpoints: list/get/create/update/rename/delete, admin-only via new Spring Security rules and in-code checks.
- The Admin group cannot be deleted or renamed, enforced server-side (the legacy page only hid the buttons).
- Renames follow the on-call roles' membership-group references in the same file write, deletes are rejected while a role still references the group, and both go through `GroupService` so the DB category authorizations are migrated or cleared as the legacy page did.
- Overnight duty schedules are rejected for new entries because the runtime never matches them; stored strings, hand-edited comments and stale members are grandfathered so files never become uneditable.
- The page validates fields inline, shows API rejections inside the dialogs, and includes a collapsed in-context help panel.
- 19 integration tests (mock managers, `groups.xml` never touched) plus Vitest store/dialog tests; verified end to end including hand-editing `groups.xml` between API calls.

[NMS-20107]: https://opennms.atlassian.net/browse/NMS-20107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
- Replaces fork PR #8699 with an in-repo branch so CI runs.


joseanesONMS:

joseanesONMS:
@synqotik @marshallmassengill — ready for review. This page now uses the `@opennms/onms-ui` Onms-XXX wrappers instead of direct PrimeVue (Button, Dialog, InputText, Select, DataTable, Column, Tag). `Message` and `IftaLabel` have no wrapper yet, so they stay on PrimeVue. `pnpm lint` is clean (ran `lint:fix`), and the `vue-tsc` build and the UI test suite pass. No behaviour change.

marshallmassengill (changes requested):
One blocker: A failed rename reports the wrong failure and leaves a phantom rename in memory. GroupManager.renameGroup removes the old key, calls grp.setName(newName), puts the new key, and only then calls saveGroups() (GroupManager.java:476-488). If that save throws, memory holds the new name while groups.xml still holds the old one, and update() will not re-read because the file's mtime never changed. The catch block then evaluates m_groupManager.hasGroup(newName) && !m_groupManager.hasGroup(name) against that memory, takes it as proof the file rename succeeded, and raises "The group was renamed, but migrating its category authorizations failed; review the group's authorized categories" while skipping the role rollback. Nothing was persisted, the operator is pointed at the wrong thing, and every read serves the phantom rename plus the repointed roles until restart. The condition needs to distinguish "the file write succeeded" from "the in-memory map was mutated", and the failure path needs a reload from disk.

joseanesONMS:
@marshallmassengill Addressed the blocker:

**A failed rename reports the wrong failure and leaves a phantom rename in memory.** Fixed at the root in `GroupManager.renameGroup`. It was mutating the in-memory map before `saveGroups()`, so a failed save left memory holding the rename while `groups.xml` still held the old name — and the catch's `hasGroup(newName) && !hasGroup(name)` check read that phantom as proof the file rename had landed, raised the category-authorization error, and skipped the role rollback.

The fix makes the in-memory rename atomic: if `saveGroups()` throws, the map is rolled back to the old name before rethrowing. `hasGroup()` now only ever reports what `groups.xml` actually received, so the REST condition genuinely distinguishes "the file write succeeded" from "the map was mutated," and the role rollback runs on the real failure. I went with atomic rollback rather than a disk reload because `update()` is mtime-gated (the very reason the phantom survived) — rolling back at the source keeps memory equal to disk without depending on it. New `GroupManagerRenameTest` covers both the rolled-back and committed paths; `GroupsRestServiceIT` rename tests stay green.

Also brought the dialogs onto the current `@opennms/onms-ui` seam while here: dropped the nested `IftaLabel` for `FormField`'s own label/for/error, and replaced the `primevue/message` banners with a themed `role=alert` element — the ManageGroups dialogs now import no un-wrapped PrimeVue primitive. `eslint` clean, vitest + `vue-tsc` build pass.

joseanesONMS:
@synqotik Applied the same camelCase-JSON convention you asked for on #8713 here too, for consistency across the admin pages:

- **camelCase JSON, no XML** — dropped the JAXB annotations from `GroupDto` and `GroupRenameRequest`, so they serialize as plain camelCase JSON (`users`, `dutySchedules`, `newName`) instead of the hyphenated/singular XML element names. The groups.xml mapping already lives in the REST service.
- **UI camelCase** — `ui/src/types/groupAdmin.ts` and everything downstream (service, store, dialogs, table, tests) follow.
- Swapped the hand-rolled `isBlank`/`trimToNull` for `org.apache.commons.lang3.StringUtils`, matching the users service.

Verified: full `GroupsRestServiceIT` (19) green with the new wire format, UI eslint clean, 24 vitest, `vue-tsc` build passes.

joseanesONMS:
Mirrored the confirmed Manage Users patterns here, and applied the menu fix:

- **Help** — inline About panel replaced with an Info-icon button in the table header that opens the help in an `OnmsDialog`, reusing the shared `Common/AboutDialogButton.vue`.
- **Row actions** — inline **Edit** and **Delete** plus a **overflow `OnmsMenu`** carrying **Rename**. Protected-group disabling (Admin) and the file-managed note are preserved.
- **menu-template-default.json** — pointed the Manage Groups entry at `ui/index.html#/admin/groups` so it stays identical to `menu-template.json` (same fix as #8713).

eslint clean, full vitest (2191) green, `vue-tsc` build passes. (`AboutDialogButton.vue` is shared with #8713; it'll dedupe cleanly when both land on develop.)

joseanesONMS:
@marshallmassengill @synqotik — ready for another pass when you have a moment; all feedback is addressed:

- **@marshallmassengill** — the rename blocker is fixed: `GroupManager.renameGroup` is now atomic, rolling back the in-memory rename if the save throws so `hasGroup()` can never report a phantom rename (and the REST catch's file-persisted-vs-memory-mutated check is honest again). New `GroupManagerRenameTest` covers both the rolled-back and committed paths; `GroupsRestServiceIT` stays green.
- **@synqotik** — applied the camelCase-JSON convention (matching #8713: `users`, `dutySchedules`, `newName`, StringUtils helpers) and mirrored the confirmed Manage Users UI patterns here — Info-dialog help via the shared `AboutDialogButton`, and inline Edit/Delete plus a overflow `OnmsMenu` (Rename). Also fixed `menu-template-default.json` to match `menu-template.json`.

Thanks!

marshallmassengill (changes requested):
Two blockers:

1. The category authorizations never reach the database. GroupsRestService calls m_groupService.renameGroup(...) and m_groupService.deleteGroup(...) directly, with no transaction. DefaultGroupService.setAuthorizedCategories only mutates detached OnmsCategory entities (category.getAuthorizedGroups().add/remove) and never calls a save, and OpenSessionInViewFilter is mapped /* in web.xml (covering the /api/v2/* mapping at line 309) with Hibernate's flush mode set to MANUAL. The legacy page works because it goes through DefaultWebGroupRepository, every method of which is @Transactional. So through the new API: delete removes the group from groups.xml but leaves its authorized_group rows behind, which is precisely the failure the
code comment says it prevents ("a future group reusing the name would silently inherit the old authorizations"), and rename silently drops the group's authorized categories instead of migrating them. <tx:annotation-driven/> is already in all four REST contexts and a dozen sibling v2 services use @Transactional, so this is a missing annotation on the two methods (or a route through WebGroupRepository). No IT asserts on categories, so nothing catches it.

2. The member picker is always empty, so no member can be added to any group. getGroupMemberCandidates reads resp.data?.users, but v1 /rest/users serializes the list under user: OnmsUserList.getObjects() is annotated @JsonProperty("user"), and the checked-in expectation fixture opennms-webapp-rest/src/test/resources/v1/users.json is {"count":2,"offset":0,"totalCount":2,"user":[...]}. ?? [] swallows the undefined, and OnmsSelect in GroupEditorDialog.vue is a plain options dropdown with no free-text entry. The store and dialog tests both mock the store's memberCandidates, so the service function is never exercised.

joseanesONMS:
@marshallmassengill — both blockers addressed:

**Category authorizations never reached the DB.** Root-caused to the MANUAL-flush OSIV session: `GroupService.renameGroup`/`deleteGroup` mutate managed `OnmsCategory` entities without an explicit save, so with no transaction the changes were discarded — delete left orphaned `authorized_group` rows and rename dropped the group's categories. Annotated both REST methods `@Transactional` (matching the legacy `@Transactional` `WebGroupRepository` path); Spring flips the session to AUTO-flush and commits. `GroupsRestServiceIT` now seeds a category authorization in a transaction, calls the endpoint, and asserts the rows migrate on rename and clear on delete.

**Member picker always empty.** `getGroupMemberCandidates` read `resp.data.users`, but v1 `/rest/users` serializes the list under `user` (`@JsonProperty("user")`). Reads the correct key now, with a new service test that exercises the wrapper shape (array and single-object) so the function is no longer untested.

`GroupsRestServiceIT` green (incl. the two new category tests + the rename/delete regression); UI `eslint` clean, vitest + `vue-tsc` build pass.

marshallmassengill (approved):
LGTM.

joseanesONMS:
@synqotik requested improvements under https://opennms.atlassian.net/browse/NMS-20282

PR #8716, opened 2026-07-30
NMS-20121: remove actiond

dino2gnt:
Never properly documented, unknown legacy daemon, functionally superseded by multiple alternative services. It will not be missed.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20121


marshallmassengill (changes requested):
Couple things that are worth looking at:

The whatsnew.adoc should be more generic to address debian where the conf file is orphaned (I think) and there isn't an rpmsave.

ServiceConfigMigratorOffline (order 6) unmarshals core/upgrade/src/main/resources/default/service-configuration-14.0.0.xml which still contains OpenNMS:Name=Actiond / org.opennms.netmgt.actiond.jmx.Actiond, and re-injects missing baseline services into the user's file on pre-14.0.0 upgrades.

docs/.../events/event-tokens.adoc:166-168 still lists autoaction, operaction, tticket as token-expansion targets. EventExpander expands them, but it now expands text nothing consumes so a one-line "retained as event data only" note there would be good.

dino2gnt:
> ServiceConfigMigratorOffline (order 6) unmarshals core/upgrade/src/main/resources/default/service-configuration-14.0.0.xml which still contains OpenNMS:Name=Actiond / org.opennms.netmgt.actiond.jmx.Actiond, and re-injects missing baseline services into the user's file on pre-14.0.0 upgrades.

...which is what it's supposed to do.

marshallmassengill (dismissed):
LGTM!

docs/modules/releasenotes/pages/whatsnew.adoc
```diff
@@ -131,6 +131,21 @@ These provided auto-rotating, dashlet-composed displays intended for wall-mounte
...
+
+The `<autoaction>`, `<tticket>`, and `<operaction>` elements remain valid in event configuration, so existing customized event definitions continue to validate and require no changes.
+However, they are no longer executed — they are retained as event data only.
+(`<operaction>` has not been executed by any daemon for several major releases.)
+
+If you relied on autoactions to run shell commands in response to events, migrate that logic to Scriptd (`etc/scriptd-configuration.xml`), which executes BSF/Groovy scripts on event receipt, or to a notification command via `etc/notificationCommands.xml`.
+
+The `CORE_SERVICE_ACTIOND_ENABLED` environment variable no longer has any effect.
+No action is required in `service-configuration.xml`: a leftover `<service><name>OpenNMS:Name=Actiond</name></service>` entry is ignored on startup and logs a warning, because that file is an enable/disable overlay on the shipped service catalog rather than the catalog itself.
+An unmodified `etc/actiond-configuration.xml` is removed on upgrade; if you customized it, your copy is left behind by the package manager and can be deleted.
```
cgorantla: I prefer whatsnew to be quick read.  Details can be a new page. See docs/modules/releasenotes/pages/spring-hibernate-upgrade.adoc as an example.

cgorantla (dismissed):
LGTM, one comment on release notes if it makes sense.

docs/modules/releasenotes/pages/whatsnew.adoc
dino2gnt: Happy to prune it down;  there's no level of detail needed because there are no users of this service past or present.

marshallmassengill (approved):
Image

PR #8717, opened 2026-07-30
NMS-20054: Plugin shared UI, Phase 1

synqotik:
# NMS-20054: Plugin shared UI, Phase 1 — expose `@opennms/onms-ui` as `window.OnmsUI`

Follow-on to the seam-layer work in #8671 (NMS-20029) and #8688 (NMS-20081). Those phases built the OpenNMS-owned wrapper API over PrimeVue and migrated all core usage onto it; this phase makes the seam **runtime-shared with plugins**, which was one of the points of putting it behind a stable API in the first place: plugins import `@opennms/onms-ui` normally, externalize it at build time, and resolve every component at runtime from the host's single copy — so plugin UIs always match the host's theme, dark mode, and z-index stack, and a future framework swap inside the seam can't strand them.

This is the same convention plugins already use for `vue`/`pinia`/`vue-router` (the `window.*` globals in `main.ts`), extended to the component library.

## Benefits / Impact

- Completely removes/replaces Feather requirement for plugin authoring
- Removes need for plugins to specifically reference PrimeVue or any other specific framework
- Plugin authors can use a rich, defined set of UI components (`onms-ui`)
- Components using this defined set will automatically work in light/dark mode, look native to OpenNMS
- As OpenNMS updates implementation, or even swaps underlying UI frameworks, plugins using the components will automatically see the updates at runtime
- Plugin code doesn't need any of the UI component implementation (since it shares it with the OpenNMS code mounted at `window.OnmsUI`), so the "compiled" plugin JS files should be far smaller, e.g. 10s of kB vs. 100s or more kB
- We can transition if desired to publish the `onms-ui` library, or even just types, to NPM, with small impact on the plugin code

## What

- **`window.OnmsUI`** — core assigns the package's entire module namespace to the global, next to the existing `window.Vue`/`Pinia`/`VueRouter`. The namespace object *is* the plugin contract: a new export-shape test (`ui/tests/onms-ui/exports.test.ts`) pins the exact 41-name runtime export list so an accidental barrel change fails CI instead of breaking externalized plugin imports at load time. Version marker: `window.OnmsUI.ONMS_UI_VERSION`.
- **Committed example plugin** (`ui/packages/onms-ui-example-plugin`) — a real, working externalized plugin: `vite-plugin-externals` maps `vue`/`pinia`/`vue-router`/`@opennms/onms-ui` to the window globals, the built module is 5.4 KB with zero framework code, and it exercises the risk areas end-to-end (host toast outlets, dialog z-index, table, tabs, theme tokens/dark mode). Its README doubles as the first plugin-developer document (externals contract, `window[extensionId]` module contract, version handshake). It has its own `typecheck` gate and is excluded from the core bundle/build.
- **Dev harness** — a dev-only vite middleware (`apply: 'serve'`, structurally absent from production builds) serves built plugin modules at `/plugin-modules/<extensionId>/…`, and dev-only routes mount them through the *same* `externalComponent()`/`Container.vue` path production plugins use. `VITE_EXAMPLE_PLUGIN=true pnpm dev` → `/#/example-plugin`. The middleware is parameterized (`pluginDevServer(extensionId, distDir)`), and a second harness behind `VITE_SERVICENOW_PLUGIN_DIST` points at an `opennms-servicenow-plugin` checkout's built UI — that was used to develop MPLUG-100 (below).
- **Tooltip seam (ride-along)** — the package re-exports PrimeVue's Tooltip directive as `OnmsTooltip`, registered as `v-onms-tooltip`; 14 templates swept, `primevue/tooltip` banned. `SideMenu.vue`'s tieredmenu is now the **only** sanctioned direct PrimeVue import left in `ui/src`. The rename is behavior-neutral: PrimeVue keys the directive's internals (pt name, `data-pc-name`, z-index bucket) off `BaseTooltip.extend('tooltip', …)`, not the registration name — verified against primevue@4.5.5 source.
- **Contract-test gaps (ride-along)** — OnmsTable's `#expansion` slot scope and the object-form (dataKey-keyed) `expandedRows` channel are now pinned by tests.
- Package version → **0.4.0**; README gains "Runtime exposure for plugins" and `OnmsTooltip` sections.

## Design notes

- The global exposes the **whole namespace**, not a curated subset: it must be exactly what `import { X } from '@opennms/onms-ui'` resolves to under externalization, or plugin imports break in confusing ways. The export-shape test keeps additions conscious.
- The dev middleware validates path containment (`resolve` + prefix check + `isFile`) — a raw client sending unencoded `../` can't escape the plugin dist directory (connect matches its mount prefix without normalizing `..`).
- One dev-tooling repair along the way: `pnpm dev` had been failing on the root page since the Menubar migration — the `ProductLogo` `resolve.alias` used a relative replacement, which dev-mode import analysis resolves against the *importer's* directory (production builds resolve it differently, which is why `pnpm build` never showed it). Verified pre-existing against the branch base; fixed by making the alias absolute.
- The example plugin deliberately has no stylesheet of its own (inline `var(--p-*)` tokens only): it must follow host theme/dark mode with zero CSS, and real plugins ship CSS via the existing `/rest/plugins/ui-extension/css/{id}` endpoint instead.

## Intentional behavior changes

None user-visible. `window.OnmsUI` is additive; `v-tooltip → v-onms-tooltip` is an internal rename of the same directive (pixel-identical rendering); the example plugin and dev harnesses are dev-only and structurally excluded from production builds.

## Validation

- Full suite (2,435 tests), lint, `vue-tsc`, production build, and the example package's typecheck all green; the final review verified the 41-export namespace survives intact in the minified production bundle and that no example-plugin/harness artifacts appear in it.
- Smoke-tested in dev: example plugin renders from `window.OnmsUI` inside host chrome, toasts fire through the **host's** outlets, dialogs stack correctly, dark mode follows.
- **Real-world proof:** the ServiceNow plugin's UI was migrated onto this contract (MPLUG-100, `opennms-servicenow-plugin`) using the dev harness in this PR — that PR is complete and its review is queued behind this merge.

## Out of scope (follow-ups)

Theme-token relocation into the package, lib build + `.d.ts` emission, npm publish + version-line scheme, plugin archetype/docs beyond the example README. Ticketed minors: `realpathSync` hardening of the dev middleware containment (symlink-in-dist is build-artifact/supply-chain scope), a pre-existing happy-dom XHR-teardown flake in two Nodes test files, a stale comment in `OnmsChip.vue`.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20054
* Consumes this contract: MPLUG-100 (opennms-servicenow-plugin UI migration — PR review pending this merge)
* Previous phases: #8671 (NMS-20029), #8688 (NMS-20081)


marshallmassengill (approved):
LGTM.  I would suggest reducing the verbosity of the PR comment itself though.

Claude did highlight these suggestions and you may want to pull them into Phase 2:

1. .env.local activates the ServiceNow harness in the app but not in the middleware — ui/vite.config.ts:99 says the opposite. Verified: with VITE_SERVICENOW_PLUGIN_DIST in .env.local, import.meta.env.VITE_SERVICENOW_PLUGIN_DIST is populated (Vite loads .env.local), so main.ts:121 registers the route, adds the stylesheet <link>, and eagerly await externalComponent(script) — while process.env.VITE_SERVICENOW_PLUGIN_DIST is undefined, because dotenv.config() reads only .env. I confirmed both halves: process.env came back undefined, and /plugin-modules/serviceNowUiExtension/…es.js returned 200 text/html (SPA fallback) instead of the module. The developer gets a MIME-type module error and a plugin whose child routes never register. Either drop the .env.local claim in the comment, or have config() read .env.local too / switch the middleware to Vite's loadEnv.

2. The example plugin is not wired into any CI gate. .circleci/main/jobs/build/build-ui.yml runs only pnpm install && pnpm build:all && pnpm test; root tsconfig.json excludes packages/onms-ui-example-plugin; nothing invokes its typecheck or build. exports.test.ts pins export names, not prop/emit/slot shapes — so the one artifact that would fail on OnmsButton prop drift never runs. That undercuts the PR's central claim ("plugins always match the host"), and it's the cheapest fix here: the build is 122 ms and the typecheck a couple of seconds. Adding both to build-ui would turn the reference consumer into a real contract test. Its pnpm install cost is already being paid in CI regardless.

3. The dev middleware sits in front of Vite's host-check and CORS middlewares. Direct server.middlewares.use() inside configureServer (ui/vite.config.ts:50) installs pre-internal. Confirmed live: Host: evil.example gets 403 on / but 200 + the module body on /plugin-modules/exampleUiExtension/…es.js, and the response lacks the Vary: Origin header Vite's own responses carry. Low impact — dev-only, read-only, confined to the plugin dist dir — but it's the same DNS-rebinding class Vite's host check exists for, and the fix is to return () => { server.middlewares.use(…) } so it lands after the internal stack. Worth folding into the already-ticketed realpathSync hardening.

4. pnpm lint is not in CI, so the new primevue/tooltip ban is advisory. Pre-existing across all three seam phases, but this PR leans on ESLint as the enforcement mechanism in both the description and the README. Separately: the no-restricted-imports block is scoped to files: ['src/**/*.ts', 'src/**/*.vue'], so packages/onms-ui-example-plugin/src — the file authors are pointed at as the template — can import PrimeVue directly with no complaint. The onms-ui README's "bans importing primevue/tooltip directly outside this package" overstates the actual scope (tests and other packages are exempt).

5. The plugin-developer README asks for something not yet possible. README.md:96 tells authors to pin a @opennms/onms-ui dependency, but the package is private: true, unpublished, and emits no .d.ts. Publishing is explicitly out of scope, which is fine — but the doc reads as actionable today. One sentence saying the package isn't consumable outside this repo yet, and what to do in the interim, would close the gap.

6. ui/vite.config.menu.ts:42 still has the relative ProductLogo alias. Menubar.vue is shared by both apps (src/main/App.vue and src/menu/App.vue), so the menu config carries exactly the bug just fixed in vite.config.ts. pnpm build:menu passes, so it's latent — but the two configs now disagree for no stated reason.

7. .pathname instead of fileURLToPath at ui/vite.config.ts:82. URL.pathname is percent-encoded, so a checkout path containing a space resolves to %20 and fails. fileURLToPath is imported three lines above. Same flaw exists on the pre-existing '@/' alias at line 76, so this is consistency-with-a-known-wart rather than a new break — but it's a new instance.

8. README polish. README.md:126 says open /example-plugin; the router is createWebHashHistory('/opennms/ui'), so it's /#/example-plugin (the PR description gets this right). And README.md:79 and :126 both reference "(Task 5 …)" — internal task numbering in a document explicitly framed as the first external plugin-developer doc.

9. Dependency choice worth a conscious decision. vite-plugin-externals@0.6.2 last shipped in 2023 and drags magic-string@0.25.9, deprecated sourcemap-codec@1.4.8, es-module-lexer@0.4.1, and fs-extra@10 into the workspace lockfile. Rollup's native external + output.globals produces the same output with no new dependency. Since the README makes this the recommended third-party toolchain, the choice propagates outward.

10. Trivia. exports.test.ts, OnmsTooltip.test.ts, and the example package's sources lack the AGPL header that sibling OnmsTooltip.ts carries (ui/ isn't covered by the mycila license gate, so cosmetic — but inconsistent within one PR). The example-plugin dist is ignored three times over (.gitignore + an explicit eslint ignores entry, when includeIgnoreFile already reads .gitignore). And the exampleUiExtension middleware mounts unconditionally in dev, unlike its VITE_EXAMPLE_PLUGIN-gated route.

synqotik:
> LGTM. I would suggest reducing the verbosity of the PR comment itself though.
> 
Thanks I'll open an issue to work through all of these issues and also retroactively reduce the PR description (may move it to the Jira ticket as it provides a lot of context).

PR #8719, opened 2026-07-31
NMS-20124: point the notification bell at the new Notifications page

joseanesONMS:
**Tickets:** NMS-20124 (epic NMS-20100).

Points the top-bar notification bell at the new Notifications page instead of the legacy JSPs, rebased onto the epic.

- Bell entries deep-link to the notices page with a query preset (outstanding / team-outstanding views).
- The preset re-applies when it changes without remounting the page.
- Carries the supporting notices store and service changes.

### DIFF
--- a/ui/src/components/AdminNotifications/NotificationQueriesCard.vue
+++ b/ui/src/components/AdminNotifications/NotificationQueriesCard.vue
@@ -9,6 +9,12 @@
         >Your outstanding notices</a>
         <span class="separator">|</span>
         <a
+          :class="{ active: store.preset === 'teamOutstanding' }"
+          data-test="query-team-outstanding"
+          @click.prevent="store.applyPreset('teamOutstanding')"
+        >Outstanding for anyone but you</a>
+        <span class="separator">|</span>
+        <a
           :class="{ active: store.preset === 'allOutstanding' }"
           data-test="query-all-outstanding"
           @click.prevent="store.applyPreset('allOutstanding')"
--- a/ui/src/containers/Notifications.vue
+++ b/ui/src/containers/Notifications.vue
@@ -40,7 +40,8 @@
 </template>
 
 <script setup lang="ts">
-import { computed, onMounted, ref } from 'vue'
+import { computed, onMounted, ref, watch } from 'vue'
+import { useRoute } from 'vue-router'
 
 import { OnmsButton, OnmsIconButton } from '@opennms/onms-ui'
 
@@ -57,8 +58,10 @@ import { useAuthStore } from '@/stores/authStore'
 import { useMenuStore } from '@/stores/menuStore'
 import { useNoticesStore } from '@/stores/noticesStore'
 import { BreadCrumb } from '@/types'
+import { NoticeQueryPreset } from '@/types/notices'
 
 const authStore = useAuthStore()
+const route = useRoute()
 const menuStore = useMenuStore()
 const noticesStore = useNoticesStore()
 const { adminRole } = useRole()
@@ -80,11 +83,30 @@ const breadcrumbs = computed<BreadCrumb[]>(() => {
 // silently drop the user filter and show every user's notices.
 const authLoaded = computed<boolean>(() => authStore.loaded)
 
+// the top-bar bell deep-links here with a preset (NMS-20124)
+const PRESETS = ['yourOutstanding', 'teamOutstanding', 'allOutstanding', 'allAcknowledged'] as const
+const initialLoad = () => {
+  const preset = route.query.preset as string
+  if ((PRESETS as readonly string[]).includes(preset)) {
+    noticesStore.applyPreset(preset as NoticeQueryPreset)
+  } else {
+    noticesStore.load()
+  }
+}
+
 onMounted(() => {
   if (authLoaded.value) {
-    noticesStore.load()
+    initialLoad()
   } else {
-    whenever(authLoaded, () => noticesStore.load(), { once: true })
+    whenever(authLoaded, () => initialLoad(), { once: true })
+  }
+})
+
+// Clicking a bell link while already on this page only changes the hash query,
+// so the component is not remounted — re-apply the preset when it changes.
+watch(() => route.query.preset, () => {
+  if (authLoaded.value) {
+    initialLoad()
   }
 })
 </script>
--- a/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/NotificationRestService.java
+++ b/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/NotificationRestService.java
@@ -267,6 +267,23 @@ public class NotificationRestService extends OnmsRestService {
         builder.alias("event", "event", JoinType.LEFT_JOIN);
         builder.alias("usersNotified", "usersNotified", JoinType.LEFT_JOIN);
 
+        // "assigned to anyone but this user": same restriction the summary
+        // endpoint uses for teamUnacknowledgedCount, including notices that
+        // notified nobody; not expressible through the generic filters
+        final String excludeNotifiedUser = params.getFirst("excludeNotifiedUser");
+        if (excludeNotifiedUser != null) {
+            // always consume the param so a blank value isn't treated as an
+            // (unknown) entity property by applyQueryFilters below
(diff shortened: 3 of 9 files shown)


joseanesONMS:

joseanesONMS:
@marshallmassengill @synqotik — ready for review. Since the PR was opened:

- Individual bell entries now open the new page (your-outstanding preset) instead of `detail.jsp`.
- A bell click while already on the page re-applies the preset via a `route.query.preset` watch (a hash-query change doesn't remount the route).
- `excludeNotifiedUser` now marks the criteria `distinct` (the `usersNotified` LEFT_JOIN otherwise duplicated notices and inflated the count) and ignores a blank value instead of 500ing.
- Dark-mode text fix for the page and its dialogs.

Lint clean, UI suite green (2451 passed), backend module builds.

Integration note: the CSV export (`fetchForExport`) landed on #8709 after this branch cut, so it isn't here yet. When this rebases onto the base that has it, the export call must carry `excludeUser` for the team preset — otherwise a team-preset export silently widens to all users.

joseanesONMS:
@marshallmassengill @synqotik — dark-mode fix for the notification dialogs. PrimeVue's dialog *chrome* (panel/header/title) doesn't follow the dark tokens in this build, so the panel stayed white while the content and fields went dark — the earlier text-only fix then left light text on a white panel (no contrast). The notification dialog classes never land on the teleported `.p-dialog` root, so this targets `.p-dialog` directly to pin the surface dark and force the title, field labels and header icons light. Verified in a running instance across all four config dialogs.

Heads up: it's a deliberately broad `.p-dialog` override (with `!important` to beat PrimeVue's runtime-injected styles), scoped to load with the Notifications page. Happy to relocate it to a global theme file if you'd rather it live app-wide.

opennms-webapp-rest/127.0.0.1.tm0.epoch
synqotik, line 1: Do *not* check in this file! These are temporary files.

opennms-webapp-rest/tmlog0.log
synqotik, line 1: Do *not* check in this file! These are temporary files.

ui/src/components/AdminNotifications/ConfigureNotificationsDialog.vue
```diff
@@ -1,42 +1,43 @@
 <template>
```
synqotik, line 1: Use `OnmsDialog`, `OnmsTab`, etc. here.

ui/src/components/AdminNotifications/ConfigureNotificationsDialog.vue
```diff
@@ -1,42 +1,43 @@
...
             Path outage management arrives with NMS-20120.
           </p>
-        </OnmsTabPanel>
-        <OnmsTabPanel value="general">
+        </TabPanel>
+        <TabPanel value="general">
           <div class="general-tab">
             <div class="status-toggle">
-              <OnmsToggleSwitch
+              <PToggleSwitch
```
synqotik: Use `OnmsToggleSwitch`

ui/src/components/AdminNotifications/NoticesTable.vue
```diff
@@ -27,11 +27,12 @@
       </div>
     </div>
 
-    <OnmsTable
+    <DataTable
```
synqotik: Use `OnmsTable`

ui/src/components/AdminNotifications/NoticesTable.vue
```diff
@@ -231,9 +241,9 @@ const downloadCsv = async () => {
   }
   const csv = [
```
synqotik, line 232: Use `composables/useDownload` here as much as possible.

ui/src/components/AdminNotifications/NoticesTable.vue
synqotik, line 253: Is there a way to make this a separate Vue component vs. inline `html`? We should avoid big HTML strings.

ui/src/components/AdminNotifications/NotificationExplanationsCard.vue
synqotik: Use `OnmsIcon`

ui/src/components/AdminNotifications/NotificationQueriesCard.vue
synqotik, line 2: Use `OnmsCard`

ui/src/components/AdminNotifications/NotificationQueriesCard.vue
```diff
@@ -23,35 +29,35 @@
       <div class="query-forms">
         <div class="search-row">
           <IftaLabel>
```
synqotik, line 31: Use `OnmsInputText`

ui/src/components/AdminNotifications/NotificationQueriesCard.vue
```diff
@@ -23,35 +29,35 @@
...
               v-model="noticeSearch"
               data-test="notice-search-input"
               @keyup.enter="goToNotice"
             />
             <label for="notification-notice-search">Get details for notice</label>
           </IftaLabel>
-          <OnmsIconButton
-            variant="outlined"
-            :icon="Search"
+          <Button
```
synqotik: Use `OnmsButton` or `OnmsIconButton`

ui/src/containers/Notifications.vue
```diff
@@ -99,3 +122,59 @@ onMounted(() => {
...
 
 </style>
+
+<!--
+  Dark-mode text fix (NON-scoped): the migrated PrimeVue/Onms components' plain
+  and custom text does not pick up dark-theme tokens in dark mode (a systemic
+  issue shared with the dashboard). The `.open-dark` theme class on <html> IS
+  reliable, and teleported dialogs live under <html> too — so key explicit
+  readable colours off it for the notification page and its dialogs/tabs.
+-->
```
synqotik: This should all be removed, don't want a bunch of custom CSS in individual files.

See if the dark mode issue goes away when you update to latest `releaset-36.x` or `develop` and use `OnmsDialog`, etc. If it doesn't then we should have a separate PR to fix that globally.

synqotik:
> @marshallmassengill @synqotik — dark-mode fix for the notification dialogs

@joseanesONMS 
Try getting latest `release-36.x` or `develop` (may need to first update the target branch), then use all `OnmsUI` components. If there's still an issue, we should fix it in a separate PR just for that. I think the latest updates I made should fix any dark mode issues, but there still may be some. I don't want a bunch of file-specific custom CSS.

joseanesONMS:
@synqotik addressed — rebased onto the current base so the whole diff is now just the bell's own work (the deep-link + `teamOutstanding` preset):

- The two temp files are gone, and the migrated `OnmsDialog`/tabs/table/card come from the base now (your wrapper comments on ConfigureNotificationsDialog / NoticesTable / NotificationQueriesCard resolve against the base's versions — the only bell change left in that card is the "anyone but you" link).
- **Removed** the inline `html.open-dark` / `.p-dialog` CSS from `Notifications.vue`. If any dark-mode gap remains after your latest theme updates, I'll do it as a separate global PR rather than file-specific CSS, per your note.
- Also fixed a real bug I'd flagged: `fetchForExport` now carries `excludeUser`, so a `teamOutstanding` export stays scoped instead of silently widening to everyone.

`TableCard → OnmsCard` (comment on line 2) is app-wide — 14 consumers — so I'd rather do that as its own cleanup than in the bell PR. Let me know if you'd prefer otherwise.

joseanesONMS:
@synqotik @marshallmassengill Ready for another look. Rebased onto the epic (now 36.0.3-SNAPSHOT on release-36.x); the notices container now also hosts the About-Notices dialog and Configure button that arrived with the shared epic changes. Build, lint and store tests are green locally; CI is running.

joseanesONMS:
@synqotik @marshallmassengill Updated and green on the latest epic. Temp files are gone and the container carries the About-Notices dialog + Configure button from the epic. The broader onms-ui wrapper suggestions (OnmsSearchInput/OnmsTabs/OnmsCard, inline-HTML extraction) are the routed-pages restructure, which lands as its own follow-up PR. combined-builds is passing.

synqotik (approved):
LGTM!

PR #8720, opened 2026-07-31
NMS-20105: Close the trouble ticket when the alarm clears

cgorantla:
The notification command now sends a close for a cleared alarm that still has an active ticket, instead of creating another one, so the resolution delivery completes the workflow. Renames the example command to troubleTicket.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20105

### DIFF
--- a/docs/modules/operation/pages/deep-dive/notifications/strategies/ticket.adoc
+++ b/docs/modules/operation/pages/deep-dive/notifications/strategies/ticket.adoc
@@ -1,15 +1,40 @@
 
 = Trouble Ticket Notifications
-:description: How to configure {page-component-title} to create trouble tickets from notifications.
+:description: How to configure {page-component-title} to create and close trouble tickets from notifications.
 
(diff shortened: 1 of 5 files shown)


marshallmassengill (changes requested):
- Close races alarmd, fails silently. sendResolvedNotificationsToUser queues at now.  initial-delay applies only to the original notice. If notifd's SELECT a.severity beats alarmd's clear commit, isCleared() is false, control falls to the create branch, hasActiveTicket() returns 0, and nothing happens. answeredby LIKE '%auto-acknowledged%' is written synchronously before the re-run (NotificationManager.java:713-729).
- Manually acked notice never closes its ticket (not sure this matters but putting the observation in here)
- CLOSE_PENDING isn't in INACTIVE_TICKET_STATES, so an in-flight UI close gets a duplicate close; ditto every escalation stage / group member. Create side has a guard but close side doesn't.
- isTicketCreationPending can't fire for this strategy's own creates. createTicketForAlarm sets only OPEN/CREATE_FAILED
- Tests leave m_severity == 0, not a valid OnmsSeverity id

marshallmassengill (approved):
LGTM!

PR #8546, opened 2026-06-03
NMS-19851: Configurable system-wide dashboard (PrimeVue)

joseanesONMS:
Implements the configurable system-wide homepage dashboard (NMS-19851) on the Vue 3 + PrimeVue /ui stack, reaching feature parity with the legacy JSP homepage and adding configurability on top. Reachable at `ui/#/dashboard`; a beta button on the legacy homepage links to it.

- Edit mode: drag/resize/add/remove/rename/collapse panels on a 12-column grid, per-panel options, global timeframe and auto-refresh with pause, full-screen NOC mode, and reset-to-default.
- One system-wide layout persisted via the new `/api/v2/dashboard` endpoints (JsonStore kv-store); per-user layouts are a future step.
- Panels cover the legacy front page (pending situations/alarms/outages, status donuts, availability, regional map, business services, notifications, news, resource graphs, KSC reports, quick search) plus new Notes, HTML Content, Top N and Metric Chart panels.
- Panel titles, availability category rows and donut slices deep-link to the corresponding legacy pages.
- Verified end to end against a local 37.x instance: layout persistence round-trip, live panel data, dark mode, full screen, and the deep links.

PR #8721, opened 2026-07-31
NMS-19851: configurable dashboard framework (base)

joseanesONMS:
First slice of the dashboard rework (NMS-19851): the configurable-dashboard framework with an empty default layout. Parity panels (NMS-20126, #8722) and new panels stack on this.

- Layout engine: grid-layout-plus with drag/resize/collapse/rename/remove on a 12-column grid, edit mode, and an add-panel picker driven by a panel registry.
- Panels squeeze up/down to pack against their neighbours (auto-fit), toggleable per dashboard in edit mode and persisted; auto-height panels measure their content and keep that height across layout reloads.
- One system-wide layout persisted as JSON via new admin-only `/api/v2/dashboard` endpoints (reads open, writes admin-only); a corrupt/partial document is normalized rather than crashing the page.
- Global auto-refresh with pause, a timeframe control (presets + custom range), and a surveillance-category/IP filter, all in the toolbar; full-screen with a CSS fallback.
- Built on the `@opennms/onms-ui` `Onms-*` wrapper components (no direct PrimeVue); ships the content-agnostic panels that exercise the framework — Sample (hidden), Notes, HTML Content — plus a beta link on the legacy homepage.
- Reachable at `ui/#/dashboard`; frontend + backend tests included. Replaces #8546.


marshallmassengill (approved):
I think this is likely ok.  I would hold off on a merge until @synqotik can review though.

PR #8722, opened 2026-07-31
NMS-20126: dashboard parity panels and default layout

joseanesONMS:
NMS-20126 (dashboard rework): the panels that reproduce the legacy homepage, and the default layout that arranges them like the front page. Stacks on #8721 (the framework base), so review its top commit.

- Panels: Pending Situations, Nodes with Pending Alarms, Nodes with Service Outages, Status Overview (severity donuts), Availability, Regional Status map, Business Services, Applications, Notifications, News Feed, Resource Graphs, Graph Collections, Quick Search.
- Default-layout heights match the measured legacy panel heights (center column and News Feed), so the new page occupies about the same space as the old one.
- Status Overview donut centre text (total + Alarms/Outages) is drawn on-canvas, proportional and vertically centred, with a slightly thinner ring.
- The global filter constrains the node-scoped panels; alarm parity matches the legacy boxes (unacknowledged, warning-and-above); the news feed honours `opennms.newsFeedPanel.show=false` with no outbound request; the "Try the new Dashboard (beta)" link sits below the Quick Search box.


marshallmassengill (changes requested):
1. AvailabilityPanel.vue:56 — filters RTC categories (etc/categories.xml: "Network Interfaces", "Web Servers") against surveillance categories from /rest/categories ("Routers", "Servers"). Two namespaces, no stock overlap, so any global filter selection blanks the panel.
  2. statusService.ts:62 — no limit sent, so QueryParameters.DEFAULT_LIMIT = 10 applies with no severity ordering, and the problem-severity filter then runs on those 10 rows. Shows "no pending alarms" while problems sit past row 10; the endpoint accepts severityFilter server-side.
  3. KscReportsPanel.vue:27 — GETs KSC/index.jsp?report=<text>, but that page's filter is a client-side ng-model and reads no report param, so the search silently does nothing. Legacy used a rest/ksc typeahead → KSC/customView.htm?type=custom&report=<id>.

**I would also suggest that if we are creating "new" stuff for KSC, we should use the new name for them and not reference them as KSC.**

  4. situationService.ts:31 — situationAlarmCount isn't a field on AlarmDTO (it has relatedAlarms + affectedNodeCount), so rows show "3 nodes" where legacy showed "affecting 3 nodes having 7 alarms".
  5. RegionalMapPanel.vue:92 — 2000 nodes + 2000 alarms per refresh tick, unordered and unfiltered; silent truncation both ways, and props.filter is ignored while the panel ships in the default layout.
  6. RegionalMapPanel.vue:53 — hardcoded OSM tile URL; geolocationStore.fetchUserDefinedTileProvider() already resolves the configured gwt.openlayers.url.
  7. statusService.ts:30 (+ situation/outage/newsfeed) — errors caught into [] and rendered as all-clear, inconsistent with availabilityService's deliberate null and notificationService's false in the same PR.
  8. defaultLayout.ts:28 — default heights converted to pixels but addPanel still writes rows; DashboardGrid.toPx's h <= 40 magnitude heuristic is the only thing keeping both working, and 56 is now duplicated.
  9. defaultLayout.ts:71 — newsfeed panel keeps its 9 rows when opennms.newsFeedPanel.show=false; legacy omitted the box.
  10. NotificationsPanel.vue:44 — count is teamUnacknowledgedCount (excludes you), link lists all unacknowledged.
  11. ResourceGraphsPanel.vue:38 — title promises graphs, form goes to the node list; legacy went to graph/chooseresource.jsp.
  12. NewsFeedPanel.vue:31 — safeLink resolves a missing link against the app origin instead of returning #.
  13. NodesWithAlarmsPanel.vue:106 — per-node counts only reflect the first 500-alarm page.
  14. RegionalMapPanel.vue:47 — saved pan/zoom key isn't per-panel.

Not sure how doable that KSC one is but if we're doing this then we might as well address it here.

joseanesONMS:
@marshallmassengill all 14 addressed in `53a0dc7f4c5`:

1. **Availability** — no longer filters RTC categories by surveillance-category names (disjoint namespaces); the panel shows the full rollup regardless of the global filter.
2. **Status lists** — `severityFilter` ×4 + `orderBy=severity&order=desc` + explicit limit, so the server fetches everything, filters, sorts, and only then caps (AbstractStatusService clears the DAO limit for severity queries). No more all-clear-past-row-10.
3. **Graph Collections** (renamed from KSC per your suggestion — type, component, and label) — real typeahead over `rest/ksc`, resolves the name to an id, opens `KSC/customView.htm?type=custom&report=<id>`; unknown names get a hint instead of a silent no-op.
4. **Situations** — count derived from `relatedAlarms`; wording restored to "affecting N nodes having M alarms".
5+6+14. **Regional map** — honors the dashboard filter (node ids + ip clause), severity-orders the capped alarm fetch so truncation keeps the worst, uses the configured tile provider via `geolocationStore` (OSM only as fallback), and keys pan/zoom per panel id.
7. **Error contract** — status/situation/outage/newsfeed services return `null` on failure and every panel renders an "Unable to load …" line instead of an all-clear.
8. **Heights** — `ROW_PX` is the single source of truth; `addPanel` and the default layout both write pixels; the grid heuristic remains only for pre-existing saved layouts.
9. **News feed** — a 204 (`opennms.newsFeedPanel.show=false`) omits the panel from the default layout, as the legacy homepage did.
10. **Notifications** — the team line now counts `totalUnacknowledgedCount`, matching what its link lists.
11. **Resource graphs** — honest launcher into the Vue graphs browser; `graph/chooseresource.jsp` no longer exists on develop, so this deviates from the legacy target deliberately.
12. **News-feed links** — a missing link renders as `#`, never the app origin.
13. **Nodes with alarms** — reads the v1 `alarms/summaries` endpoint (the legacy `AlarmDao.getNodeAlarmSummaries` source): exact server-side per-node counts, narrowed by the filter resolved to node ids.

64 tests green (10 files, incl. new coverage for the query contracts, the disabled-feed default, and the Graph Collections navigation); full eslint + `vue-tsc` pass. Push carries `!smoke`.

marshallmassengill (changes requested):
Two additional blockers to look into:

  - ui/src/services/statusService.ts:93 — reads resp.data['businessservice']/['application'], but the DTO lists are @JsonProperty("businessservices")/("applications") (singular = XML element name only); both panels always render "There are no pending alarms." Verified live.
  - ui/src/components/Dashboard/panels/RegionalMapPanel.vue:190 — buildFilterClauses() emits node.id==N, which only works on alarm/outage entities; /api/v2/nodes?_s=(node.id==1) 500s ("String cannot be cast to Integer"), so any active category filter blanks the map with no error.

joseanesONMS:
@marshallmassengill both addressed in `fd524f6e78a`:

- **Status lists** — `statusService` now reads the plural `businessservices` / `applications` keys the DTO lists emit; the test mocks the real shape (the old mock used the singular key, which is why it passed).
- **Regional map filter** — `filterFiqlClauses` takes the node-id property; the map queries `/nodes` with `id==N` and alarms with `node.id==N`. Alarms/situations/outages callers are unchanged.

Tests, eslint and `vue-tsc` green. Push carries `!smoke`.

marshallmassengill (changes requested):
One more blocker:

- QuickSearchPanel.vue:33,61 — all four forms submit to /opennms/element/nodeList.htm, which no longer exists on develop. No file, no dispatcher-servlet.xml mapping, no @RequestMapping; the only references in the tree are the panel itself. Every Quick Search submit 404s. Current legacy targets: element/node.jsp?node=<id> for node ID, and ui/index.html#/nodes?nodename=… / ?iplike=… / ?monitoredService=<name> for the rest (element/index.jsp:submitNodeSearch). The service select also sends service=<id>; the live param is monitoredService keyed by service name.

joseanesONMS:
@marshallmassengill fixed in `720492a3879`: the Quick Search forms no longer post to `element/nodeList.htm`. Each one pushes the `/nodes` route in-app with the query keys that page tracks — `nodename` (+`listInterfaces=true`), `iplike`, and `monitoredService` keyed by service **name** (the select's value was the numeric id, which the nodes page doesn't resolve). A node id is sent as `nodeId`, which the nodes page already redirects to `element/node.jsp?node=<id>`. Empty submits are ignored. New component test pins all four routes; eslint + `vue-tsc` green, push carries `!smoke`.

marshallmassengill (approved):
This looks good to me.  Wait for @synqotik's approval to merge.

PR #8723, opened 2026-07-31
NMS-20127: dashboard Top-N and Metric Chart panels

joseanesONMS:
NMS-20127 (dashboard rework): two new panels with no legacy homepage equivalent. Stacks on #8721 (the framework base), so review its top commit.

- Top-N ranks entities by a chosen KPI over the selected timeframe (attributing each measurement to the source its response label encodes, not hash-ordered position).
- Metric Chart plots one metric on one entity over time.
- Both honor the dashboard timeframe (including custom ranges), guard against stale concurrent responses, and add their KPI/entity option editors to the panel options dialog. Add-only — not placed in the default layout.

## External References

- Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20127


marshallmassengill (dismissed):
I don't see any blockers.  Especially since this is going into a feature branch.

joseanesONMS:
@marshallmassengill heads-up on the approval state: merging the base PR #8721 auto-closed this stacked PR (its base branch went away), and reopening/retargeting it to `develop` made GitHub dismiss your approval. The diff is exactly what you approved, plus a mechanical develop sync (the base content now on develop is byte-identical, so the shared files resolve to this branch's versions; lint/tests/build green). Re-approve when you get a moment and I'll merge.

marshallmassengill (changes requested):
Can you post some screenshots of how this ends up appearing in the UI?

joseanesONMS:
Screenshot 2026-08-31 at 14 14 43
Screenshot 2026-08-31 at 14 11 51

joseanesONMS:
Screenshot 2026-08-31 at 14 22 28
Screenshot 2026-08-31 at 14 22 21

marshallmassengill:
Ok... so this PR adds the chart and Top-N panel types?

joseanesONMS:
> Ok... so this PR adds the chart and Top-N panel types?

Yes.  There is a PR to create a Parity Dashboard behind a New Dashboard button.
Then there is another PR to add the ability to edit.
Then there is another PR to add some new panels.

I need to do other new PRs (not done yet), for:
PDF Export.
Remove old Dashboard, and replace with new.

joseanesONMS:
@marshallmassengill screenshots are up above (two comments on Aug 31) and CI is green on `4c42767200a`. Re-requesting review.

marshallmassengill (changes requested):
A potential blocker: A node with more than one ICMP-polled IP interface produces a duplicate Top-N row per address, and the chart silently uses whichever address comes first. Both stem from stamping the parent node's label onto every child resource.

At the very least, that needs to be noted but I think there are ways to resolve it.  Here are some other things we need to look into as well:
 
- The chart spans gaps instead of breaking on them, contradicting the service comment and hiding outages.
- Top-N silently ranks only the first 250 candidates, truncated in resource-tree order.
- Every failure renders as "No data in this timeframe", with nothing logged.
- Each panel refetches the whole resource tree per refresh, one depth level deeper than it uses.
- Top-N downloads up to a quarter million values to compute five numbers, averaged in the browser. This one seems particularly wasteful.
- Both the KPI and Metric selectors ship with exactly one option (fine if we are going to expand things later).
- The chart defaults to an entity named localhost, so it may open empty on some installs.
- N is clamped on save but not on read, so a bad stored value shows as no data.
- The theme observer watches the body element as well as the root, I don't know if that's right.

joseanesONMS:
@marshallmassengill all addressed in `80852389b00`:

- Multi-interface nodes: each source label now carries the interface address when a node polls more than one, so Top-N shows one row per address and the chart entity list is unambiguous.
- Gaps: the chart breaks the line on null points instead of bridging them.
- The 250 cap is now a batch size; every candidate is ranked (test with 253 sources).
- Each source is asked for ~24 coarse points instead of ~1000 (verified the RRD backend returns sane averages at that step); both panels share a cached resource tree fetched at depth 1, which already carries the attributes.
- Failures propagate: the panels show "Unable to load" and log the cause, distinct from "no data".
- No more `localhost` default: with nothing configured the chart shows the first entity carrying the metric and names it in the legend.
- N is clamped on read as well as on save, with one shared rule.
- The theme observer watches the root element only.
- The single-option selectors stay as they are, per your note.

eslint, `vue-tsc` build and the dashboard/service tests are green; push carries `!smoke`.

marshallmassengill (changes requested):
Some more things to address:

1. The entity label is unstable, and the chart stores it as identity. The address qualifier is added only when a node has more than one matching source, so an existing source's label changes when a second interface starts being polled. A saved panel then resolves to nothing and shows "No data". Store the resource identifier and keep the label for display.
2. Batch fan-out is unbounded. Every batch request is dispatched inside the loop, so concurrency equals the number of batches, once per refresh tick per panel. Each batch is cheap now, so running them in sequence removes the spike.
3. The cache-invalidation helper is exported but only tests call it. Nothing in the app can clear the five-minute window, so a newly provisioned node stays out of the rankings and the entity dropdown until it lapses. Wire it to the dashboard refresh or drop the export.
4. Both selectors offer exactly one choice, since the metric registry holds a single entry.

joseanesONMS:
@marshallmassengill addressed in `10c13132a33` (develop merged in first):

1. The chart stores the entity's resource id (`entity`) and keeps the label (`entityLabel`) for display only; a label stored by an earlier save still resolves. Test covers a node gaining a second interface.
2. Top-N batches now run sequentially; a test asserts at most one `/measurements` request in flight.
3. The resource-tree cache is cleared from the dashboard store's `tickRefresh`, so every refresh (manual or interval) sees newly provisioned nodes; the TTL only bounds reuse within a tick.
4. The registry now has response-time KPIs for ICMP, SNMP, HTTP, HTTPS, DNS and SSH, and both selectors list only the KPIs that some resource on the system actually carries (falling back to the full list when none does).

marshallmassengill (changes requested):
I'm torn on approving this one at the moment. I can see it being useful but it needs more thought into how to make these re-usable.

Specially we need to do a few things still probably: 

Filter or mark stale entities on metric change. In PanelOptionsDialog.vue, loadEntities re-adds the current entity even when it does not carry the newly selected metric, so saving yields a "No data" panel with no hint. Either drop it from the list when the metric changes, or append "(no data for this metric)" to its label.

Drop or defer the SNMP KPI. The stock poller-configuration.xml has no ds-name="snmp", so SNMP Response Time is hidden by the availability filter on every default install. Remove it from TOPN_KPIS until SNMP response time is collected by default, or leave it with a comment saying why it rarely appears.

But those two are really just symptoms here.  I think this needs to be thought about a bit more holistically.

joseanesONMS:
@marshallmassengill both items are in, on top of a fresh develop merge:

- A saved entity without data for the selected metric is listed as "(no data for this metric)" when the options open, and dropped with the selection cleared once the metric is changed.
- The SNMP KPI is removed, with a comment on why (no `snmp` ds-name in the stock poller configuration).

A second pass on the same files also fixed: opening the options of a chart with a non-default metric no longer clears its entity, entity loads carry a sequence so an older result cannot win, Apply waits for the load, a panel saved with a label is moved onto the resource id, the resource-tree cache keeps an in-flight fetch across a refresh tick, and two nodes sharing a label are told apart by address. Component tests for the options dialog pin these.

Understood that the holistic pass on reusability is still to come.

marshallmassengill (approved):
I think the code for this is ok at this point so I'm approving.  I'm still torn on how this can be consumed in a reusable way though and I think final approval should come from @synqotik.  I know the intent is to get it onto the main page to start with.

PR #8727, opened 2026-08-03
NMS-20128: Vue UI Manage Minions page

joseanesONMS:
Migrates the Manage Minions admin page from the legacy AngularJS list to a Vue /ui screen. Reuses the existing `/api/v2/minions` REST unchanged — no backend, model, or schema change.

- The menu entry is repointed to the new page and gated to `ROLE_ADMIN`; `MenuHeaderIT` is updated to assert the new page.
- Edit is read-before-write (GET then PUT) so server-maintained status, version, and last-seen survive an edit, and unexposed minion properties round-trip.
- The ID column links to the minion's auto-created requisition node via the same `foreignId` lookup the legacy page used.
- The table adds column sort, a client-side global search, loading/empty/error states, and a refresh button.
- The list fetch is bounded (2,000) with a "showing first N" note rather than an unbounded fetch; search and paging are client-side, which suits the small minion counts.
- Delete treats an already-removed minion (404) as success.
- Edits read the current row before the PUT so server-maintained fields survive, but two admins saving the same minion still resolve last-writer-wins; the minion has no locking token (its `version` is the software version).

## External References

- Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20128


joseanesONMS:
@synqotik @marshallmassengill — ready for review. This page now uses the `@opennms/onms-ui` Onms-XXX wrappers instead of direct PrimeVue (Button, Dialog, InputText, DataTable, Column, Tag; the IconField/InputIcon search box is now OnmsSearchInput). `Message` and `IftaLabel` have no wrapper yet, so they stay on PrimeVue. `pnpm lint` is clean (ran `lint:fix`), and the `vue-tsc` build and the UI test suite pass. No behaviour change.

joseanesONMS:
@marshallmassengill @synqotik — rebased on current develop (shared `services/index.ts` union + repointing the minion dialog/table icons to the `@opennms/onms-ui/icons/*` seam develop migrated to). No conflicts; eslint, `vue-tsc` build, and the minion tests are green. Ready for review.

marshallmassengill (changes requested):
Couple blockers:

  1. ui/src/services/minionAdminService.ts contains a raw NUL byte (offset 2965, inside the minionNodeKey template literal). Git classifies the file as binary — the PR diff literally reads Bin 0 -> 5444 bytes, so that file is unreviewable on GitHub. Use \u0000 (which is what minionAdminService.test.ts and the legacy index.js both write) or a printable separator.
  2. menu-template-default.json still points manageMinions at minion/index.jsp with roles: null. The sibling Manage Groups migration needed a dedicated follow-up commit (6a2c064cb74) for exactly this omission.

joseanesONMS:
@marshallmassengill both fixed:
1. The `minionNodeKey` separator is now a unicode escape instead of a raw NUL byte, so git treats `minionAdminService.ts` as text again and the diff is reviewable. Runtime key is unchanged; the service test already asserts the escaped form (13 tests green).
2. Repointed `manageMinions` at the Vue page with `ROLE_ADMIN` in both `menu-template-default.json` and `menu-template-alt.json`, matching `menu-template.json`.

marshallmassengill (dismissed):
No blockers for me but attaching some screenshots would be useful.  @synqotik should still review this before merging.

joseanesONMS:
Screenshot 2026-09-01 at 14 18 08
Screenshot 2026-09-01 at 14 17 53
Screenshot 2026-09-01 at 14 17 45
@marshallmassengill @synqotik

ui/src/services/minionAdminService.ts
```diff
@@ -0,0 +1,153 @@
...
+/// language governing permissions and limitations under the
+/// License.
+///
+
+import useSnackbar from '@/composables/useSnackbar'
+import useSpinner from '@/composables/useSpinner'
+import { Minion } from '@/types/minionAdmin'
+import { v2 } from './axiosInstances'
+
+// PrimeVue Manage Minions (NMS-20128). Reuses the existing v2
```
synqotik: Remove any mention of PrimeVue, here and anywhere else.

Also this comment should be something like:

```
Uses the existing v2 AbstractDaoRestService CRUD at /api/v2/minions.
```

ui/src/services/minionAdminService.ts
```diff
@@ -0,0 +1,153 @@
...
+  return fallback
+}
+
+// null on failure (not []) so callers can keep showing the previous list
+// Bound the fetch rather than limit=0 (unbounded); minions are few in practice.
+const LIST_CAP = 2000
+
+const listMinions = async (): Promise<{ minions: Minion[]; totalCount: number } | null> => {
+  try {
+    startSpinner()
```
synqotik: Remove any Spinners or Snackbars from the service (which reminds me, we need to go back and do this for older services). Services should not have any UI functionality.

Services can just return `false` if there's an error and UI (which is actually calling the store) can create the snackbar and error message (so the store needs to bubble up the error result too).

If there is specific info from the server that's needed for the error message, like in `updateMinion` and `deleteMinion` where you are using `errorMessage()` to format error info from the server, you can use `createSuccessResponse` and `createFailureResult` from `@/types/validation`.

See `snmpConfigService.ts`, `saveSnmpProfile()` etc. for an example of this.

For the other cases here, the caller already has the info needed to populate an error messages (caller has minion id, etc.), so can just return `false`.

ui/src/stores/minionAdminStore.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  const deleteMinion = async (id: string) => {
+    const error = await API.deleteMinion(id)
+    if (error === null) {
+      await getMinions()
+    }
+    return error
+  }
+
+  return { minions, loadError, isLoading, totalCount, truncated, nodeIdFor, getMinions, updateMinion, deleteMinion }
```
synqotik: Items should be on separate lines.

ui/src/services/minionAdminService.ts
```diff
@@ -0,0 +1,153 @@
...
+        map[minionNodeKey(String(n.foreignId), n.location ?? null)] = Number(n.id)
+      }
+    }
+    return map
+  } catch (_err) {
+    return {}
+  }
+}
+
+export interface MinionEdit {
```
synqotik: I'd put this in `types/minionAdmin.ts`.

ui/src/containers/ManageMinions.vue
```diff
@@ -0,0 +1,54 @@
+<template>
+  <div class="onms-row">
+    <div class="onms-col-12">
+      <BreadCrumbs :items="breadcrumbs" />
+    </div>
+  </div>
+  <div class="manage-minions-container">
+    <h1 class="page-title">Manage Minions</h1>
+    <MinionsHelpPanel />
```
synqotik: Rather than being an expandable panel, follow the pattern we are using elsewhere with an Info `OnmsIconButton` that pops up a dialog with that help info.

ui/src/components/ManageMinions/MinionsTable.vue
```diff
@@ -0,0 +1,237 @@
...
+      </OnmsColumn>
+      <OnmsColumn header="Properties">
+        <template #body="{ data }">
+          <span class="props-count">{{ Object.keys(data.properties ?? {}).length }} propert{{ Object.keys(data.properties ?? {}).length === 1 ? 'y' : 'ies' }}</span>
+        </template>
+      </OnmsColumn>
+      <OnmsColumn header="Actions">
+        <template #body="{ data }">
+          <div class="action-container">
+            <OnmsButton
```
synqotik: Here, I'd use the `OnmsIconButton` with Edit and Delete icons, like we do on other similar tables.

ui/src/components/ManageMinions/MinionsTable.vue
```diff
@@ -0,0 +1,237 @@
...
+        </template>
+      </OnmsColumn>
+      <OnmsColumn field="label" header="Label" sortable />
+      <OnmsColumn field="location" header="Location" sortable />
+      <OnmsColumn field="type" header="Type" sortable />
+      <OnmsColumn field="status" header="Status" sortable>
+        <template #body="{ data }">
+          <OnmsTag
+            :value="data.status ?? 'unknown'"
+            :severity="statusSeverity(data.status)"
```
synqotik: Note that `minion.status` is something like `Started`, `Stopped`, etc., *not* a severity. Don't have to do it here, but either here or a follow to map the minion actually statuses to an `OnmsTag` "severity" to change the color.

ui/src/components/ManageMinions/MinionEditorDialog.vue
```diff
@@ -0,0 +1,226 @@
...
+            />
+          </li>
+        </ul>
+        <p v-else class="no-props">No properties.</p>
+        <small v-if="dupKeyProblem" class="field-error" data-test="prop-error">{{ dupKeyProblem }}</small>
+      </div>
+    </div>
+
+    <template #footer>
+      <OnmsButton variant="text" label="Cancel" data-test="cancel-button" @click="emit('update:visible', false)" />
```
synqotik: Cancel button should be `variant="ghost"`, or I think you can just leave `variant` out and it defaults to `ghost` (i.e. it has a border).

synqotik (changes requested):
Overall looks good, but I have some comments.

May also want to merge in the latest `develop`, I noticed it doesn't have the most recent menubar, etc.

I don't have actual minions set up, but I faked the response to at least see the edit screen.

Normally I'd say to put the editing on a separate page, but it seems small enough to keep it as a dialog. We can always revisit.

ui/src/services/minionAdminService.ts
joseanesONMS: Done: comment now reads "Uses the existing v2 AbstractDaoRestService CRUD at /api/v2/minions." No other PrimeVue mentions remain in the minion files.

ui/src/services/minionAdminService.ts
joseanesONMS: Done. The service no longer touches the spinner or snackbar; updateMinion/deleteMinion return ValidationResult via createSuccessResponse/createFailureResult (server detail kept where it is a short plain message), listMinions still returns null on failure. The store bubbles the result and the dialog/table/container show the toast or inline error.

ui/src/stores/minionAdminStore.ts
joseanesONMS: Done.

ui/src/services/minionAdminService.ts
joseanesONMS: Done, MinionEdit now lives in types/minionAdmin.ts.

ui/src/containers/ManageMinions.vue
joseanesONMS: Done: replaced with AboutDialogButton (Info icon in the card header) wrapping a MinionsAbout component, same as Users/Groups/Scheduled Outages.

ui/src/components/ManageMinions/MinionsTable.vue
joseanesONMS: Done, Edit and Delete are OnmsIconButton with the action/Edit and action/Delete icons.

ui/src/components/ManageMinions/MinionsTable.vue
joseanesONMS: Agreed it is not a severity. The values MinionStatusTracker writes are up / down / unknown, so the mapping is up -> success, down -> danger, unknown -> warn, anything else -> secondary.

ui/src/components/ManageMinions/MinionEditorDialog.vue
joseanesONMS: Done, variant="ghost".

joseanesONMS:
@synqotik all eight comments addressed in d840f8ea71c and develop merged in (7d2211a14db): service has no UI side effects and returns ValidationResult, MinionEdit moved to types, About dialog replaces the panel, icon Edit/Delete buttons, ghost Cancel, store return list one per line. Minion status values are up/down/unknown from MinionStatusTracker and map to success/danger/warn.

marshallmassengill (changes requested):
Couple things:

1. The node lookup builds a query clause from unescaped minion ids. A comma reads as OR and a semicolon as AND in the filter grammar, so an id containing either changes the query's meaning and a parenthesis breaks it. Ids come from a controller config property with no validation on the write path, though in practice they are generated as UUIDs. Robustness rather than security, and the worst case is wrong or missing node links.
2. That same query grows with the minion count and fails well below the cap. One clause per minion in a single URL, about 54 bytes each once encoded, so a typical 8 KB request limit is reached near 150 minions while the list cap is 2000. The failure is silent: every node link disappears with nothing in the interface to explain it. Chunk the lookup, or skip the links past a threshold and say so.
3. Read-before-write narrows the lost-update window without closing it. Two admins editing the same minion still resolve last-writer-wins. The version field is the minion's software version, not a locking token, so nothing better is available and this belongs in the description rather than the code.

Additionally:
- Rename the helper for that severity issue to something that names the output rather than the input, such as statusTagSeverity or statusColor. The current name reads as "the severity of the status", which is the misreading.
- Replace the existing comment, which lists the values, with one that says what the function is doing, that minion status is a state string and this maps it onto the tag's color vocabulary.

Hopefully that helps.

joseanesONMS:
@marshallmassengill addressed in `4464d36f152`:

1. Ids containing FIQL grammar characters (, ; ( ) = ! ~ < > whitespace) are left out of the node lookup and simply get no link; in practice ids are UUIDs.
2. The lookup runs in chunks of 50 ids (roughly 2.7 KB of URL per request), so the links survive at any minion count; a failed chunk only loses its own links, the rest still resolve. Tests cover both.
3. Lost-update: noted in the description that read-before-write narrows but does not close the window, since the minion carries no locking token.

Also renamed the helper to `statusTagSeverity` with a comment saying it maps the state string onto the tag colour vocabulary.

marshallmassengill (approved):
I am good with this but @synqotik should review before we merge this in.

cgorantla (commented):
Configure OpenNMS admin page (/opennms/admin/index.jsp)
and opennms-base-assembly/src/main/filtered/etc/search-actions.xml were not updated with new pages.

joseanesONMS:
@cgorantla done in e3e7d244709: the Configure OpenNMS page and search-actions.xml now point at the Vue page. The same fix is pushed to the sibling PRs #8728, #8729 and #8715, which had the same omission.

PR #8728, opened 2026-08-03
NMS-20129: Vue UI Manage Monitoring Locations page

joseanesONMS:
Migrates the Manage Monitoring Locations admin page to a PrimeVue /ui screen. Reuses the existing `/api/v2/monitoringLocations` CRUD unchanged — no backend or schema change.

- The menu entry is repointed; `MenuHeaderIT` is updated to assert the new page.
- Update is read-before-write, preserving geolocation, tags, and other fields the form does not expose.
- The table adds a client-side global search, full-column sort with a default order, empty/error states, and a bounded fetch (2,000) with a "showing first N" note.
- Priority is validated client-side to 1–2,147,483,647 (the DB column is a 32-bit integer) with Save blocked on any invalid value; the location name blocks URL- and FIQL-unsafe characters.
- These field validations are client-side only — the REST API still accepts out-of-range values.
- Delete treats a 404 as success, and the Default location cannot be deleted.

### DIFF
--- a/opennms-webapp-rest/src/main/webapp/WEB-INF/menu/menu-template.json
+++ b/opennms-webapp-rest/src/main/webapp/WEB-INF/menu/menu-template.json
@@ -335,7 +335,7 @@
         {
           "id": "manageMonitoringLocations",
           "name": "Manage Monitoring Locations",
-          "url": "locations/index.jsp",
+          "url": "ui/index.html#/admin/monitoring-locations",
           "locationMatch": "",
           "roles": null
         }
--- a/smoke-test/src/test/java/org/opennms/smoketest/MenuHeaderIT.java
+++ b/smoke-test/src/test/java/org/opennms/smoketest/MenuHeaderIT.java
@@ -137,7 +137,8 @@ public class MenuHeaderIT extends OpenNMSSeleniumIT {
         wait.until(ExpectedConditions.presenceOfElementLocated(By.xpath("//div[@class='card-header']/span[text()='Applications']")));
 
         clickMenuItem("Distributed Monitoring", "Manage Monitoring Locations");
-        wait.until(ExpectedConditions.visibilityOfElementLocated(By.xpath("//ol[@class='breadcrumb']/li[contains(text()[normalize-space()], 'Monitoring Locations')]")));
+        // now a /ui (Vue) page rather than the legacy JSP breadcrumb
+        wait.until(ExpectedConditions.visibilityOfElementLocated(By.xpath("//div[@id='app']//h1[@class='page-title' and normalize-space(text())='Manage Monitoring Locations']")));
 
         // Manage Inventory Menu
         clickMenuItem("Manage Inventory", "Provisioning Requisitions");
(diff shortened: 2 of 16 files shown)


joseanesONMS:
@synqotik @marshallmassengill — ready for review. This page now uses the `@opennms/onms-ui` Onms-XXX wrappers instead of direct PrimeVue (Button, Dialog, InputText, InputNumber, DataTable, Column; the IconField/InputIcon search box is now OnmsSearchInput). `Message` and `IftaLabel` have no wrapper yet, so they stay on PrimeVue. `pnpm lint` is clean (ran `lint:fix`), and the `vue-tsc` build and the UI test suite pass. No behaviour change.

marshallmassengill (changes requested):
Couple blockers to address:

  - Only menu-template.json is repointed. menu-template-default.json and menu-template-alt.json still send users to locations/index.jsp, the same omission as 8727's second blocker. All three also keep roles: null while the route guard demands admin, so non-admins see the entry and get bounced. Set ROLE_ADMIN as the minions entry does.
  - The branch is 435 commits behind develop. Develop has since gained the patterns 8727 was asked to adopt (ValidationResult in types/validation, AboutDialogButton, OnmsIconButton), so merge before the items below.

Some other things worth looking at too:
  - The delete confirmation and the help panel say nodes in a deleted location "are left without a valid location". fk_node_location was re-added in 19.1.0 without ON DELETE CASCADE, so the delete fails on a constraint violation, and the 500 body is HTML so the user sees only the generic failure text. Say the location must be empty of nodes first. Minions have no FK and do keep a dangling location string.
  - The service calls the spinner and snackbar directly. The other reviewer asked 8727 to move that to the components and return ValidationResult. Apply the same here so both pages match. monitoringLocationAdminService.ts:11.
  - Pattern alignment with 8727 after its review: help content as AboutDialogButton rather than a TogglePanel, Edit and Delete as OnmsIconButton, Cancel as variant="ghost".
  - The create form rejects whitespace in a location name. The legacy form and the database accept it, so admins with names like "Data Center 1" can no longer create such a location while existing ones still edit. Keep the FIQL and path characters blocked, allow spaces.
  - The priority hint says the default location is 100. The Default row is inserted with no priority at all. Reword or drop.
  - Creating a name that already exists is not checked client-side, and the server error is another HTML 500. The loaded list is right there for a friendlier message. Nit.

joseanesONMS:
@marshallmassengill addressed in 3f04520eed1 and the commit after it, on top of a fresh develop merge:

- All three menu templates point at the Vue page with `ROLE_ADMIN`.
- Service returns `ValidationResult` with no spinner or snackbar; the store bubbles it and the components toast or show the error inline, same shape as the other admin pages.
- Help is an `AboutDialogButton`, Edit and Delete are `OnmsIconButton`, Cancel is ghost.
- Spaces are allowed in names again; only path and markup characters are blocked, and a duplicate name is caught client-side.
- Priority hint reads "Lower numbers sort first (default 100)".
- Delete text says the location must have no nodes assigned (the node foreign key has no cascade) and that Minions keep the old name until re-registered.

A second pass also found that a location named `.` resolved its item URL to the collection, so its Delete would have hit the bulk endpoint; `isPathAddressable` now rejects `.` and `..`, and 5xx bodies are never shown as messages.

marshallmassengill (dismissed):
Good with me now.  I would wait for @synqotik to review before merging.

cgorantla (commented):
Configure OpenNMS admin page (/opennms/admin/index.jsp)
and opennms-base-assembly/src/main/filtered/etc/search-actions.xml were not updated with new pages.

PR #8729, opened 2026-08-03
NMS-20131: Vue UI Manage Surveillance Categories page

joseanesONMS:
Migrates the Manage Surveillance Categories admin page to a Vue /ui screen. Reuses the existing `/rest/categories` CRUD and the granular node-membership endpoints unchanged — no backend change.

- The menu entry is repointed; `MenuHeaderIT` is updated to assert the new page.
- Node assignment is a server-paged, searchable node table (reusing the Node List's `nodeService`, so it scales to large inventories) with a per-row membership toggle applied immediately.
- Requisitioned nodes are locked by default behind an opt-in override, with a warning that a provisioning sync overwrites their categories.
- Category create/edit and description update are supported; delete shows a generic "may break references" warning (no surveillance-views.xml cross-check — that would require a backend endpoint that does not exist today).
- Category names block URL- and FIQL-unsafe characters (client-side).
- Empty, error, and loading states are handled throughout.

## External References

- Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20131

### DIFF
--- a/ui/src/components/ManageCategories/CategoriesHelpPanel.vue
+++ b/ui/src/components/ManageCategories/CategoriesHelpPanel.vue
@@ -7,7 +7,7 @@
   >
     <template #header>
       <span class="panel-header">
-        <i class="pi pi-question-circle" aria-hidden="true" />
+        <OnmsIcon :icon="Help" class="help-icon" />
         About Surveillance Categories
       </span>
     </template>
@@ -38,8 +38,10 @@
 </template>
 
 <script setup lang="ts">
+import { OnmsIcon } from '@opennms/onms-ui'
 import { ref } from 'vue'
 
+import Help from '@/components/icons/action/Help.vue'
 import TogglePanel from '@/components/Common/TogglePanel.vue'
 
 const collapsed = ref(true)
@@ -52,7 +54,7 @@ const collapsed = ref(true)
   gap: 0.5rem;
   font-weight: 600;
 
-  .pi-question-circle {
+  .help-icon {
     color: var(--p-primary-color);
   }
 }
--- a/ui/src/components/ManageCategories/CategoriesHelpPanel.vue
+++ b/ui/src/components/ManageCategories/CategoriesHelpPanel.vue
@@ -1,0 +1,82 @@
+<template>
+  <TogglePanel
+    :collapsed="collapsed"
+    class="categories-help-panel"
+    data-test="categories-help-panel"
+    @update:collapsed="(value: boolean) => (collapsed = value)"
+  >
+    <template #header>
+      <span class="panel-header">
+        <i class="pi pi-question-circle" aria-hidden="true" />
+        About Surveillance Categories
+      </span>
+    </template>
+    <div class="help-columns">
+      <div class="help-section">
+        <div class="section-title">What surveillance categories are for</div>
+        <p>
+          A surveillance category is a named grouping of nodes used across
+          OpenNMS for filtering, dashboards, notifications, and the surveillance
+          view. A node can belong to any number of categories, and group
+          authorizations can restrict which categories a user group sees.
+        </p>
+      </div>
+      <div class="help-section">
+        <div class="section-title">How to use this page</div>
+        <p>
+          <strong>Add New Category</strong> creates a category with an optional
+          description; <strong>Edit</strong> changes the description (the name is
+          the identifier and cannot be changed). <strong>Manage Nodes</strong>
+          assigns and unassigns nodes. <strong>Delete</strong> removes the
+          category — filters or authorizations that reference it by name stop
+          matching, so review them afterward. The same operations are available
+          to tooling through the <code>/rest/categories</code> REST API.
+        </p>
+      </div>
+    </div>
+  </TogglePanel>
+</template>
+
+<script setup lang="ts">
+import { ref } from 'vue'
+
+import TogglePanel from '@/components/Common/TogglePanel.vue'
+
+const collapsed = ref(true)
+</script>
+
+<style lang="scss" scoped>
+.panel-header {
+  display: inline-flex;
+  align-items: center;
+  gap: 0.5rem;
+  font-weight: 600;
+
+  .pi-question-circle {
+    color: var(--p-primary-color);
+  }
+}
+
+.help-columns {
+  display: flex;
+  gap: 2.5rem;
+  flex-wrap: wrap;
+
+  .help-section {
+    flex: 1;
+    min-width: 320px;
+  }
+}
+
+.section-title {
+  font-size: 1rem;
+  font-weight: 600;
+  margin-bottom: 0.5rem;
+}
+
+p {
+  margin: 0 0 0.75rem 0;
+  font-size: 0.9rem;
+  line-height: 1.5;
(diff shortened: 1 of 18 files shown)


ui/src/components/ManageCategories/CategoriesHelpPanel.vue
```diff
@@ -0,0 +1,82 @@
+<template>
+  <TogglePanel
+    :collapsed="collapsed"
+    class="categories-help-panel"
+    data-test="categories-help-panel"
+    @update:collapsed="(value: boolean) => (collapsed = value)"
+  >
+    <template #header>
+      <span class="panel-header">
+        <i class="pi pi-question-circle" aria-hidden="true" />
```
synqotik: Make sure to merge/rebase from/onto the latest `develop`.

Then, use `OnmsIcon` or `OnmsIconButton` here instead of `pi-question-circle`.

Basically, rewrite everything to use the `Onms-` components instead of any direct PrimeVue components.

ui/src/components/ManageCategories/CategoriesTable.vue
```diff
@@ -0,0 +1,167 @@
+<template>
+  <TableCard class="categories-table">
+    <div class="header">
+      <div class="card-title">Surveillance Categories</div>
+      <Button
```
synqotik: Use `OnmsIconButton`, `OnmsTable`

ui/src/components/ManageCategories/CategoryEditorDialog.vue
```diff
@@ -0,0 +1,145 @@
+<template>
+  <Dialog
```
synqotik: Use `OnmsDialog`.

ui/src/components/ManageCategories/CategoryEditorDialog.vue
```diff
@@ -0,0 +1,145 @@
...
+    class="category-editor-dialog"
+    :style="{ width: '520px', maxWidth: '95vw' }"
+    data-test="category-editor-dialog"
+    @update:visible="(value: boolean) => emit('update:visible', value)"
+  >
+    <div class="form-column">
+      <Message v-if="errorText" severity="error" :closable="false" data-test="dialog-error">{{ errorText }}</Message>
+
+      <FormField v-if="!isEditing">
+        <IftaLabel>
```
synqotik, line 14: Use `OnmsInputText`

joseanesONMS:
@synqotik — done. Rebased onto develop and migrated all four Manage Categories components off direct PrimeVue onto the Onms-* seam wrappers:

- **Buttons** → `OnmsButton` (Add/Manage Nodes/Edit/Delete/Cancel/Save/Close), with the `+` glyph now an `OnmsIcon`.
- **Tables** → `OnmsTable` / `OnmsColumn`; the nodes table keeps its lazy-load spinner.
- **Dialogs** → `OnmsDialog` (sizing via the `width` prop).
- **Inputs** → `OnmsInputText`; the node search is now `OnmsSearchInput`.
- **Toggle/Checkbox** → `OnmsToggleSwitch` / `OnmsCheckbox`.
- Help-panel `pi-question-circle` → `OnmsIcon`.

`Message` and `IftaLabel` stay on PrimeVue as no seam wrapper exists for them yet. Lint clean and the ManageCategories tests pass.

ui/src/components/ManageCategories/CategoriesHelpPanel.vue
joseanesONMS: Done in f6fd3c4149f: develop merged in, and the collapsible panel is replaced by the shared AboutDialogButton (Info OnmsIconButton opening the help in a dialog) like Users/Groups. No direct PrimeVue imports remain in the page.

ui/src/components/ManageCategories/CategoriesTable.vue
joseanesONMS: Done: OnmsTable (already) and OnmsIconButton for Manage Nodes / Edit / Delete with the network/Nodes, action/Edit and action/Delete seam icons.

ui/src/components/ManageCategories/CategoryEditorDialog.vue
joseanesONMS: Done, OnmsDialog.

ui/src/components/ManageCategories/CategoryEditorDialog.vue
joseanesONMS: Done: OnmsInputText inside FormField (label/required/error props); the PrimeVue IftaLabel and Message imports are gone.

joseanesONMS:
@synqotik @marshallmassengill develop merged in and the page brought onto the current conventions in `f6fd3c4149f`: no direct PrimeVue imports, OnmsIconButton row actions, About dialog instead of the panel, ghost Cancel, and the service returns ValidationResult with the store bubbling it up so the components show toasts and inline errors (same shape as the Minions page). Tests, eslint and the typed build are green.

marshallmassengill (changes requested):
Couple more items to dig into:

  - Only menu-template.json is repointed. menu-template-default.json and menu-template-alt.json still send users to admin/categories.htm (their entry id is surveillanceCategories, not manageSurveillanceCategories). All three keep roles: null. The legacy URL was admin-only by virtue of /admin/**; the new one is not, so set ROLE_ADMIN or every user sees an entry that bounces them.

We should probably restrict this to admins?  I think?

Other findings
  - Membership add is a body-less PUT to an endpoint declared @Consumes(APPLICATION_XML). CXF treats a missing Content-Type as a wildcard so it should match, but the only in-tree precedent (the REST IT) sends an explicit XML content type, and I could not verify live. Send Content-Type: application/xml on that PUT so matching does not depend on CXF defaults. categoryAdminService.ts:74.
  - Category names are varchar(64) and the editor has no length limit. Overflow surfaces as an HTML 500 the error filter discards, so the user sees only the generic failure. Add maxlength and a client check.
  - The delete 404-as-success branch never fires: deleteCategory on the server returns 400, not 404, for a missing category. Harmless, but either match on 400 with the "does not exist" text or drop the branch.
  - The name search allows _, which is a SQL LIKE single-character wildcard behind FIQL ==*term*, so a_b also matches axb. Cosmetic over-match, not a defect.

joseanesONMS:
@marshallmassengill addressed in f7f19461dc9 and the commit after it, on top of a fresh develop merge:

- All three menu templates now point at the Vue page and carry `ROLE_ADMIN`; writes were already admin-only on `/rest/**`.
- The membership PUT sends `Content-Type: application/xml`.
- Name capped at 64 and description at 256 (column widths), with inline errors.
- The 404-as-success branch is gone; the server's 400 text surfaces instead.
- `_` is dropped from the node search term so it no longer acts as a LIKE wildcard.

A second pass also gave the node dialog a members-only default (via `category.id` in the v2 search), relaxed the name rules to what the legacy page accepted, and made a failed toggle re-read the page.

marshallmassengill (dismissed):
I think this is ok... we should probably talk about the inclusion of "_" characters though.  Not sure how that should be addressed in the long term.  I would wait for @synqotik to review first.

PR #8730, opened 2026-08-03
NMS-20152: Forecasting and descriptive legend states on Resource Graphs

joseanesONMS:
Reintroduce forecasting on the new Resource Graphs page and replace the "NaN" legend stats with descriptive states.

The forecast runs entirely through the existing server-side measurement filters — no R dependency, and no client-side statistics to keep in sync.

- The overlay (Holt-Winters fit, confidence bounds, polynomial trend) comes from a single `POST /rest/measurements` running the Outlier → HoltWinters → Trend → Chomp chain; the page only renders the returned columns.
- Holt-Winters is multiplicative with bounds `delta = z·σ·√min(h, period)`, matching the shipped `HWForecast` filter — the bounds widen and then plateau at one seasonal period rather than fanning open.
- `Graph Start` drives the trailing `Chomp` cutoff, so training uses the full window while only the visible range is returned.
- 1-day / 7-day / 31-day templates plus a Custom panel with per-field validation (confidence and outlier thresholds bounded, season × 2 < training window, integers ≥ 1).
- A banner surfaces forecast failure conditions — no fit, an all-NaN fit (including the multiplicative zero-in-season case), or zero-width bounds — matching the legacy page's `checkForecastWarning`.
- Legend: a stat of exactly `0` printed "NaN" (d3 `formatPrefix`) — now renders `0.00`; non-finite stats render **No Data** or **Invalid Data**.
- Covered by `ForecastGraph.test.ts` (data wiring and the warning conditions) and the extended legend-state tests; full `ui` suite green.


synqotik:
@joseanesONMS fix merge issues and update re seam. Also run `pnpm lint:fix` to fix formatting, etc. Other than that looks good (assuming all the algorithms are correct...).

joseanesONMS:
@marshallmassengill @synqotik ready for review.

Adds a Forecasting screen for Resource Graphs (Holt-Winters + polynomial trend, client-side) and descriptive legend states. Since opening:
- Forecast controls are on the `@opennms/onms-ui` seam — `OnmsSelect`/`OnmsInputNumber` inside `FormField` (no direct PrimeVue, no hand-rolled labels).
- Added component tests for `ForecastGraph.vue` alongside the existing forecast-math unit tests.
- Rebased onto the latest `develop`.

`pnpm lint` clean; UI build + tests green.

marshallmassengill (changes requested):
Couple blockers:

Really validate #1 because it's pulling in R info from somewhere and it's been removed for a while.  We need to make sure that is cleared up.

1. The stated premise is false. The server-side forecast has no R dependency. HWForecast.java is @FilterInfo(name="HoltWinters"), pure commons-math3, and HWForecast.checkForecastSupport() is // noop, forecasting always supported now. HoltWintersR is a separate, unused filter. The legacy index.js builds Chomp/Outlier/HoltWinters/Trend — every one of them R-free. The only stale artifact is forecast.jsp's error string, which can never fire. The real blocker was that GraphMetricsPayload has no filters field while QueryRequest does (@XmlElement(name="filter")); that's a ~10-line type addition, not 300 lines of hand-rolled statistics and a second HW implementation to keep in sync.

2. graphStart silently does nothing on the forecast plot. runForecast only reads trainingStart; the legacy chain ended with Chomp{cutoffDate: graphStartInMillis}. The field is offered, validated, and shipped in all three templates while having no effect, and forecasting.adoc says "Graph Start sets what's actually visible on screen."

3. removeOutliers is two-sided; the filter it replaces is not. OutlierFilter.java only NaNs values greater than the quantile. The new version also clips below 1-q, which deletes the seasonal troughs Holt-Winters exists to learn. Verified on a capacity-planning shape (500 + 0.4·i, small daily ripple, 14d/1000 samples): 25 leading and 25 trailing samples dropped — the newest 8.4 hours discarded, forecast grid shifted 8.4h into the past, no warning. Same 2.5% on the 31-day template is ~9 days.

4. The change contradicts the docs it ships alongside. forecasting.adoc states multiplicative seasonality (new code is additive), delta = z·σ·√min(h, period) (new code is uncapped √h), Trend #00ffff (new code #00b4d8). HWForecast.java caps the horizon deliberately: "so bounds widen briefly and then plateau rather than fanning open without bound." The PR lists removing that cap as a feature. Docs aren't touched.

joseanesONMS:
@marshallmassengill You were right on every count — validated, then pivoted the whole feature to server-side forecasting:

**#1 (false premise / use the server's R-free HWForecast).** Confirmed: `HWForecast` is `@FilterInfo("HoltWinters")`, pure commons-math3, `checkForecastSupport()` is a noop, and `QueryRequest` already takes a `filter` chain while `GraphMetricsPayload` didn't. `ForecastGraph` now POSTs an **Outlier → HoltWinters → Trend → Chomp** chain to `/measurements` and renders the returned `HWFit`/`HWLwr`/`HWUpr`/`Trend` columns. **Deleted `forecasting.ts`** (the ~300 lines of hand-rolled statistics + the second HW impl) and its unit tests; added the `filter` field to `GraphMetricsPayload`.

**#2 (graphStart did nothing).** Now effective — it's the trailing `Chomp` `cutoffDate`, so training uses the full window and only Graph Start → horizon is drawn.

**#3 (two-sided outlier).** Gone — we use the server's one-sided `Outlier` filter.

**#4 (docs contradictions).** Resolved by using the server: the delta is now `HWForecast`'s capped `z·σ·√min(h, period)`, seasonality/behavior is whatever `HWForecast` does (single source of truth), Trend is back to `#00ffff`, and I updated the warnings section in `forecasting.adoc` to match the shipped behavior.

@synqotik seam + lint addressed too: the rewritten `ForecastGraph` stays on `OnmsSelect`/`OnmsInputNumber`/`FormField`, `pnpm lint` is clean, and it's rebased on the latest develop.

UI: lint clean, `vue-tsc` build passes, resource-graph tests green (the 8 unrelated failures are pre-existing `localStorage`/`theme` service tests from the develop merge). Deploying to my local instance next to eyeball the server-backed overlays.

marshallmassengill (changes requested):
Couple issues:

Forecast failures are silent. runForecast only warns when resp.timestamps is empty.

HWForecast hardcodes MULTIPLICATIVE seasonality, so any metric that reaches 0 inside its season divides by zero. The user gets a green line, a cyan trend, three empty legend entries, and no explanation. The same commit deletes the three documented warnings for exactly these conditions from forecasting.adoc, and core/web-assets/.../forecast/checkForecastWarning.js, which implements them, is still in-tree.

joseanesONMS:
@marshallmassengill — both addressed.

**Silent forecast failures.** `runForecast` now inspects the returned columns instead of only checking for empty timestamps, mirroring the conditions the legacy page reports via `checkForecastWarning.js`. The historical data line still renders, and a banner explains the missing/degenerate overlay when: no `HWFit` column comes back, the fit is all-NaN, or the confidence bounds have zero width.

**Multiplicative-seasonality zero.** The all-NaN-fit case is reported specifically when the source series touches zero within the window — "the metric reaches zero within its season, which the multiplicative Holt-Winters model cannot forecast" — so that green-line-plus-empty-legend case is no longer unexplained. I left `checkForecastWarning.js` in place: it's still live for the legacy `forecast.jsp` (`apps/forecast/index.js` requires it), so rather than remove or duplicate it I ported the same checks into the Vue component. The per-condition warnings table is restored in `forecasting.adoc`, with the zero-in-season cause added.

Component tests cover all three banner conditions; UI eslint + vitest + `vue-tsc` build pass.

joseanesONMS:
@marshallmassengill friendly ping — the forecast failure banners are pushed (no fit / all-NaN fit incl. the multiplicative zero-in-season case / zero-width bounds), the warnings table is restored in the docs, and I've corrected the stale PR description that still described the old client-side approach. Ready for another pass.

marshallmassengill (changes requested):
Bit a blocker here:

ForecastGraph.vue:190,379 — isFetchable requires attribute && resourceId && !expression, so only DEF-backed metrics survive the series filter. But RrdGraphConverter._onLine/_onArea set series.metric to the drawn source, which in most stock definitions is a CDEF ({name, expression}, no attribute or resourceId). Those series are dropped; when a report draws only CDEFs (we have quite a few graphs that do this) the page falls through to "This graph has no forecastable metrics" and no forecast is possible.

The legacy page posted the whole model — DEFs as source, CDEFs as expression, the shape Graph.vue:216-243 already builds — and passed the series name as inputColumn. The single-source label: 'data' shortcut is what rules CDEFs out.

a3f2x0:
Run [RS] string and log AKA claims to point in time. Request statements of disclosure. Specifically, elaborate on the status of relationship between yourself and the disregarded entity. Include a descriptive report on the specific engagement, particularly if unwanted or unwarranted physical contact was applied, the amount of times it happened and details of group participants. Finally, what was the reason that caused the members to deviate
from pursuing initial consent and what was the reason for withholding information and revenues that should have included the indemnified person. 
Until these reports are reviewed a request is being made to block any and all changes that may alter course of subjection to this entire respiratory. 
08-24-2026::8:07 A.M

joseanesONMS:
Reworked the forecast fetch per your last review: the tab now posts the whole graph model — DEFs as `source`, CDEFs as `expression`, the same payload `Graph.vue` builds — and passes the selected series' metric name to the filter chain as `inputColumn`, matching the legacy page. The `isFetchable` DEF-only gate and the single-source `label: 'data'` shortcut are gone, so expression-drawn series (most stock graphs) are forecastable again.

The selected column is located by response label instead of assumed at index 0; probed against a live instance that an expression-backed `inputColumn` returns HW fit/bounds and trend, with the data column ordered last in the response. Tests cover the CDEF path end-to-end (model payload shape + `inputColumn`), and the branch is refreshed with current develop.

marshallmassengill (dismissed):
No blockers for me on this now but wait for @synqotik to review before merging.

ui/src/components/Resources/utils/LegendFormatter.ts
synqotik, line 1: Rename this file `legendFormatter.ts`, e.g. camelCase.

ui/src/components/Resources/ForecastGraph.vue
```diff
@@ -0,0 +1,448 @@
...
+
+const onTemplateChange = () => {
+  const t = templates.find(x => x.id === selectedTemplateId.value)
+  if (t) {
+    options.value = { ...t.options }
+  }
+}
+
+// carry over (and complete) the legacy form validation the JSP enforced
+const optionProblems = computed<Record<string, string>>(() => {
```
synqotik, line 153: Move this into a separate file, `components/Resources/utils/validation.ts`.

ui/src/components/Resources/ForecastGraph.vue
```diff
@@ -0,0 +1,448 @@
...
+  if (!(o.confidenceLevel > 0 && o.confidenceLevel < 1)) {
+    p.confidenceLevel = 'Must be between 0 and 1.'
+  }
+  if (!intGe1(o.trendOrder)) {
+    p.trendOrder = 'Must be a whole number ≥ 1.'
+  }
+  return p
+})
+
+const canForecast = computed(() => !!selectedMetric.value && Object.keys(optionProblems.value).length === 0)
```
synqotik, line 183: Use curly braces around all `computed` bodies, even if they are just one liners.

ui/src/components/Resources/ForecastGraph.vue
```diff
@@ -0,0 +1,448 @@
...
+const props = defineProps<{
+  label: string
+  forecastDefinition: string
+  forecastResourceId: string
+}>()
+
+const canvasRef = ref<HTMLCanvasElement | null>(null)
+let chart: Chart | null = null
+
+const model = ref<{ title: string; verticalLabel: string; metrics: Metric[]; series: Series[] } | null>(null)
```
synqotik, line 110: Make an `interface` for this type, put into `src/components/Resources/types.ts`. Make sure to use the type anywhere `model` is used.

ui/src/components/Resources/ForecastGraph.vue
```diff
@@ -0,0 +1,448 @@
...
+})
+
+const canForecast = computed(() => !!selectedMetric.value && Object.keys(optionProblems.value).length === 0)
+
+// Every drawn series is forecastable: the whole model is posted — DEFs as
+// sources, CDEFs as expressions, the same shape Graph.vue builds — and the
+// selected series' metric name is the filter chain's inputColumn. The old
+// single-source label:'data' shortcut ruled out CDEF-drawn series, which most
+// stock graph definitions use.
+const metricFor = (seriesMetricName: string): Metric | undefined =>
```
synqotik, line 190: Make a new file, `src/components/Resources/utils/forecastUtils.ts` or similar name.

Put the following into it:

- metricFor
- buildModelPayload
- selectedColumn
- fetchColumn
- line
- forecastFIlters
- columnByLabel
- forecastWarningFor
- toPoints

Plus some others, noted in separate comments.

ui/src/components/Resources/ForecastGraph.vue
```diff
@@ -0,0 +1,448 @@
...
+    const start = end - options.value.trainingStart * DAY_MS
+    const resp = await API.getGraphMetrics(
+      buildModelPayload(selectedMetric.value, start, end, forecastFilters(selectedMetric.value, options.value)) as any)
+    if (!resp || !(resp.timestamps ?? []).length) {
+      warning.value = 'Could not load data for the selected metric.'
+      drawChart([])
+      return
+    }
+    const ts = resp.timestamps as number[]
+    const datasets: any[] = [line(seriesName(), '#7EE600', toPoints(ts, selectedColumn(resp, selectedMetric.value)))]
```
synqotik, line 377: Refactor this creation of `datasets` into a function in the new `forecastUtils.ts` file.

Also, use constants for all colors.

ui/src/components/Resources/ForecastGraph.vue
```diff
@@ -0,0 +1,448 @@
...
+  forecasting.value = true
+  try {
+    if (!metricFor(selectedMetric.value)) {
+      warning.value = 'Could not load data for the selected metric.'
+      drawChart([])
+      return
+    }
+    const end = Date.now()
+    const start = end - options.value.trainingStart * DAY_MS
+    const resp = await API.getGraphMetrics(
```
synqotik, line 369: Make this `resp` object strongly typed, and use that type everywhere we pass `resp`, e.g. `selectedColumn`, `columnByLabel`, `forecastWarningFor` and a few others. Need to fix `getGraphMetrics` and put the type into `src/types/index.ts` or similar.

ui/src/components/Resources/ForecastGraph.vue
```diff
@@ -0,0 +1,448 @@
...
+
+const model = ref<{ title: string; verticalLabel: string; metrics: Metric[]; series: Series[] } | null>(null)
+const series = ref<Series[]>([])
+const selectedMetric = ref<string>('')
+const loadError = ref<string | null>(null)
+const warning = ref<string | null>(null)
+const forecasting = ref(false)
+
+const defaultOptions: ForecastOptions = {
+  trainingStart: 14, graphStart: 7, season: 1, forecasts: 1,
```
synqotik, line 118: put each on separate line.

ui/src/components/Resources/ForecastGraph.vue
```diff
@@ -0,0 +1,448 @@
...
+  }
+  const timestamps = (resp.timestamps ?? []) as number[]
+  const values = selectedColumn(resp, seriesMetricName)
+  // the API's step and returned timestamp spacing are already in milliseconds
+  const stepMs = timestamps.length > 1 ? timestamps[1] - timestamps[0] : Math.max(1, Math.floor((endMs - startMs) / 1000))
+  return { timestamps, values, stepMs }
+}
+
+const line = (label: string, color: string, points: { x: number; y: number }[], dash = false, fill = false) => ({
+  label, data: points, borderColor: color, backgroundColor: color,
```
synqotik, line 247: put each `name: value` on a separate line. Ideally create an `interface` for this data shape.

synqotik (changes requested):
Added some comments, mostly code organization.

Also, may want to wait until `OnmsDatePicker` PR (which includes `TimeControls`) gets into `develop` and merge/rebase that in.

ui/src/components/Resources/utils/LegendFormatter.ts
joseanesONMS: Done, renamed to legendFormatter.ts (importers and the test updated).

ui/src/components/Resources/ForecastGraph.vue
joseanesONMS: Done: forecastOptionProblems() in utils/validation.ts; the component only wraps it in a computed.

ui/src/components/Resources/ForecastGraph.vue
joseanesONMS: Done, every computed body uses braces.

ui/src/components/Resources/ForecastGraph.vue
joseanesONMS: Done: ForecastModel in components/Resources/types.ts, used for the ref, the converter output and every helper that takes the model.

ui/src/components/Resources/ForecastGraph.vue
joseanesONMS: Done: utils/forecastUtils.ts now holds metricFor, buildModelPayload, selectedColumn, fetchColumn, line, forecastFilters, columnByLabel, forecastWarningFor, toPoints plus buildForecastDatasets and the colour constants; the component keeps state, drawChart, reset and runForecast.

ui/src/components/Resources/ForecastGraph.vue
joseanesONMS: Done: buildForecastDatasets() in forecastUtils.ts, colours in FORECAST_COLORS.

ui/src/components/Resources/ForecastGraph.vue
joseanesONMS: Done: resp is GraphMetricsResponse everywhere (getGraphMetrics already returned it); its columns field in src/types/index.ts is now { values: number[] }[] instead of a one-element tuple. buildModelPayload returns GraphMetricsPayload so the `as any` casts are gone.

ui/src/components/Resources/ForecastGraph.vue
joseanesONMS: Done.

ui/src/components/Resources/ForecastGraph.vue
joseanesONMS: Done: one property per line, typed as ForecastLineDataset (and ForecastPoint) in components/Resources/types.ts.

joseanesONMS:
@synqotik all nine comments addressed in `ead4e256c6f`, and develop (with OnmsDatePicker / TimeControls) merged in. ForecastGraph.vue now keeps only state and chart drawing; helpers live in utils/forecastUtils.ts and utils/validation.ts with types in components/Resources/types.ts, responses are typed GraphMetricsResponse, colours are constants, and LegendFormatter.ts is renamed legendFormatter.ts. New unit tests cover the validation, payload, datasets and warning helpers.

PR #8734, opened 2026-08-03
NMS-20113: Replace the Slack and Mattermost notification channels with a generic webhook

marshallmassengill:
Replaces the Slack and Mattermost notification strategies with a single WebhookNotificationStrategy that takes the URL, method, content type, headers, and body from the notification command, so it can also target Microsoft Teams, Discord (for all Gen Z users I guess), and non-chat endpoints.  

Highlights:

  - Body templates substitute ${subject}, ${textMessage}, and any declared command argument. Values are escaped when the payload is JSON, with a |raw modifier for fragments that are themselves JSON.
  - Template tokens exclude colons so they cannot collide with the metadata DSL, which notifd resolves against notification parameters before the command runs.
  - The rendered JSON body is parsed before it is sent, so a template mistake is reported locally rather than as a 400 from the receiver.
  - Any 2xx counts as success, narrowed by an optional -success-match for receivers that report failure behind a 200.
  - Adds jackson-databind as an explicit dependency of opennms-services; it was already on the compile classpath transitively.
  - Replaces slack.adoc and mattermost.adoc with one webhook.adoc covering Slack, Mattermost, Teams, Discord, a non-chat example, and a migration table.
  -useSystemProxy now defaults to false, matching HttpNotificationStrategy; the removed strategies defaulted it to true.
  
I'm going to need to resolve the conflicts with the whatsnew and such based on the other notifd update PRs.

Improvements assisted by Anthropic Claude Opus 5.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20113


dino2gnt (dismissed):
Looks reasonable to me.  This has been on your list for a long time, excellent work getting it done.

dino2gnt (approved):

PR #8741, opened 2026-08-05
NMS-20164: Lock OSGi resolved dependencies to the root pom versions

dino2gnt:
Lock netty and jackson-databind to the versions specified in the root pom, so OSGi resolution doesn't pull in older, vulnerable versions.  Make a similar fix with Newts 1.5.7, so it won't pull in older transient dependencies either. 

Updating the actual package versions looked hugely invasive.  This should get everything on consistent versions of these dependencies, matching what's defined in pom.xml.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20164
* https://opennms.atlassian.net/browse/SUPPORT-3278

### DIFF
--- a/container/features/pom.xml
+++ b/container/features/pom.xml
@@ -104,6 +104,46 @@
               </artifactItems>
             </configuration>
           </execution>
+          <execution>
+            <id>copy-jaxrs-feature</id>
+            <phase>process-resources</phase>
+            <goals>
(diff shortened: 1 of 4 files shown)


christianpape (approved):
LGTM.

dino2gnt:
All green! :tada:

PR #8742, opened 2026-08-05
NMS-20029: forward the #option slot in OnmsAutoComplete

joseanesONMS:
Fixes a seam-wrapper gap: OnmsAutoComplete forwarded only `#empty` and `#footer`, so migrating a type-ahead from PrimeVue AutoComplete to the wrapper silently dropped any custom `#option` template and fell back to the bare `optionLabel`.

- Forward PrimeVue AutoComplete's `#option` slot (with its slot props) alongside `#empty`/`#footer`.
- Add a unit test asserting the `#option` slot reaches the inner AutoComplete.
- No consumer on develop currently relies on `#option`, so this is additive with no behavior change to existing usages; it unblocks rich suggestion rendering (e.g. a UEI plus its event label).

### DIFF
--- a/ui/packages/onms-ui/src/components/OnmsAutoComplete.vue
+++ b/ui/packages/onms-ui/src/components/OnmsAutoComplete.vue
@@ -28,6 +28,15 @@
     >
       <slot name="footer" />
     </template>
+    <template
+      v-if="$slots.option"
+      #option="slotProps"
+    >
+      <slot
+        name="option"
+        v-bind="slotProps"
+      />
+    </template>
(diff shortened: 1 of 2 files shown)


synqotik:
@joseanesONMS looks good - this could be targeted to `release-36.x`. Once that's done, I'll approve.

joseanesONMS:
@synqotik retargeted to `release-36.x` and rebased so it's just the single seam commit. Ready for your approval — thanks!

synqotik (approved):
LGTM!

PR #8743, opened 2026-08-05
NMS-20163: Generate the OpenAPI documents at build time

marshallmassengill:
New `opennms-openapi-docs` module generates the v1 and v2 documents during the build; a small resource serves them from the classpath. Nothing is committed.

| On an instance | Serves | Comes from |
|---|---|---|
| `/opennms/rest/openapi.json`, `.yaml` | v1 document | generated `openapi/openapi-v1.json`, served by `.../web/rest/v1/OpenApiResource.java` |
| `/opennms/api/v2/openapi.json`, `.yaml` | v2 document | generated `openapi/openapi-v2.json`, served by `.../web/rest/v2/OpenApiResource.java` |
| `/opennms/rest/api-docs/` | Swagger UI, v1 | `.../web/rest/v1/SwaggerUiResource.java` |
| `/opennms/api/v2/api-docs/` | Swagger UI, v2 | `.../web/rest/v2/SwaggerUiResource.java` |
| `/opennms/ui/index.html#/open-api` | Vue page, both documents | `ui/src/containers/OpenAPI.vue`, unchanged |

The behavior is in `org.opennms.web.rest.support.openapi.AbstractStaticOpenApiResource` and `AbstractSwaggerUiResource`, both under `opennms-webapp-rest`. The generator lives in src/main/java and its dependencies are provided, so nothing swagger-related ships in that artifact. `opennms-assemblies/webapp-full` ships the documents jar, because they describe every ReST module merged into that war and several of those depend on `opennms-webapp-rest`.

 - `OpenApiDocWriter` runs at `process-classes` and writes into `target/classes/openapi`, so there is no regenerate step, nothing to leave stale, and no generated file to conflict on a forward merge. To read a document without an instance, unpack it from the module jar.
- The generated documents are content-identical to what the runtime produced. 
- First request to `openapi.json` drops from ~1.3s to under 0.05s.
- `OpenApiResourceCoverageTest` fails the build when a module contributing ReST resources is not a dependency of `opennms-openapi-docs`, which would otherwise drop its endpoints silently.
 - Generation is forked rather than run in the Maven JVM, where the plugin realm shadows part of the project classpath and changes the output; `OpenApiDocWriter` checks and refuses.
 - Swagger UI moves from 3.23.11 to 5.25.3. 3.23.11... this fixes a CVE in the process but it broke the UI... different PR coming to resolve that.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20163


opennms-openapi-docs/src/main/java/org/opennms/openapi/OpenApiDocGenerator.java
```diff
@@ -0,0 +1,166 @@
...
+import io.swagger.v3.oas.models.servers.Server;
+
+/**
+ * Produces the OpenAPI documents shipped in src/main/resources/openapi.
+ *
+ * The metadata and base packages here must match
+ * {@code opennms-webapp-rest/src/main/webapp/WEB-INF/applicationContext-cxf-rest-v*.xml},
+ * or the documents stop describing what is deployed.
+ */
+public final class OpenApiDocGenerator {
```
synqotik, line 57: This file should be in `src/main/java`, not `src/test/java`.

synqotik:
This is a nice change.

It does seem that it should be part of the build process, or else part of CI build only. Right now it looks like a dev has to remember to run `mvn -pl opennms-openapi-docs test -Dopenapi.regenerate=true` any time they think they have updated something.

Of course, if part of CI build, CI would also need to see if there is any diff, then if there is, commit it automatically.

Alternatively, it could *not* be run, and be part of some release documentation to make sure someone runs it before a release. Developers can run it or not if they remember to while changing relevant code.

marshallmassengill:
> This is a nice change.
> 
> It does seem that it should be part of the build process, or else part of CI build only. Right now it looks like a dev has to remember to run `mvn -pl opennms-openapi-docs test -Dopenapi.regenerate=true` any time they think they have updated something.
> 
> Of course, if part of CI build, CI would also need to see if there is any diff, then if there is, commit it automatically.
> 
> Alternatively, it could _not_ be run, and be part of some release documentation to make sure someone runs it before a release. Developers can run it or not if they remember to while changing relevant code.

I can't say for certain this is the right approach but the general thought is that penApiDocsUpToDateTest regenerates and diffs against the committed copies so drift fails the build and the message includes the command to fix it. 

This also runs in the integration-test job rather than build, and the find-tests walk will pick up a change to opennms-webapp-rest, any of the contributing *.rest modules, or opennms-model/opennms-dao.

opennms-openapi-docs/src/main/java/org/opennms/openapi/OpenApiDocGenerator.java
marshallmassengill: Should be fixed with the latest push but I'm going to wait before I run CI again.

synqotik:
> This also runs in the integration-test job rather than build, and the find-tests walk will pick up a change to opennms-webapp-rest, any of the contributing *.rest modules, or opennms-model/opennms-dao.

Clever!

synqotik (dismissed):
LGTM!

opennms-openapi-docs/README.adoc
```diff
@@ -0,0 +1,57 @@
...
+After changing anything the documents describe, which means JAX-RS annotations,
+`io.swagger.v3.oas.annotations`, or the shape of a model class that appears in a
+signature:
+
+  mvn -pl opennms-openapi-docs test -Dopenapi.regenerate=true
+
+then commit the result. `OpenApiDocsUpToDateTest` fails the build if the
+committed copies have drifted, so this is not optional.
+
+== Adding a ReST module
```
cgorantla: This section doesn't read well for me.  It looks like none of this needed for the intended user.

opennms-openapi-docs/README.adoc
```diff
@@ -0,0 +1,57 @@
...
+The documents are committed so that a change to the ReST surface shows up in
+review.
+
+== Regenerating
+
+After changing anything the documents describe, which means JAX-RS annotations,
+`io.swagger.v3.oas.annotations`, or the shape of a model class that appears in a
+signature:
+
+  mvn -pl opennms-openapi-docs test -Dopenapi.regenerate=true
```
cgorantla: may need to compile and install it's dependencies with `-am`

```suggestion
 mvn -pl opennms-openapi-docs -am test -Dopenapi.regenerate=true
```

cgorantla: Probably skip `-am` as it can pull lot of dependencies but make sure to build other dependent modules first and then do `mvn -pl opennms-openapi-docs test -Dopenapi.regenerate=true`

opennms-openapi-docs/README.adoc
marshallmassengill: removed entirely

opennms-openapi-docs/README.adoc
marshallmassengill: marking as resolved but let me know if this isn't right

mershad-manesh:
Just some thoughts on this approach -
I think this could turn into a failure mode that keeps coming back, and it gets worse the further the branch travels.

A few ways this bites us:
* **Forget to regenerate**: an API change lands without an updated JSON file, the test fails the build, and now someone has to run the regen command, commit the result, and re-push before it can merge.
* **Merge forward**: the same test can fail again, on its own, once the change merges forward — legitimately, since the target branch may have added or changed endpoints in the meantime. Either way it's another manual fix, and it's not just this repo either.
* **Release**: this adds a "JSON files must be up to date" requirement right before a release, so a last-minute change can turn into a blocker.

One thing worth considering: generate the docs as a build step inside `opennms-assemblies/webapp-full` itself, where the full classpath already lives, instead of a standalone module — and don't commit the JSON at all.

marshallmassengill:
> Just some thoughts on this approach - I think this could turn into a failure mode that keeps coming back, and it gets worse the further the branch travels.
> 
> A few ways this bites us:
> 
>     * **Forget to regenerate**: an API change lands without an updated JSON file, the test fails the build, and now someone has to run the regen command, commit the result, and re-push before it can merge.
> 
>     * **Merge forward**: the same test can fail again, on its own, once the change merges forward — legitimately, since the target branch may have added or changed endpoints in the meantime. Either way it's another manual fix, and it's not just this repo either.
> 
>     * **Release**: this adds a "JSON files must be up to date" requirement right before a release, so a last-minute change can turn into a blocker.
> 
> 
> One thing worth considering: generate the docs as a build step inside `opennms-assemblies/webapp-full` itself, where the full classpath already lives, instead of a standalone module — and don't commit the JSON at all.

Good call outs.

So just to be aware, webapp-full: opennms-assemblies isn't in the root reactor, it's built by assemble.pl, which only runs in the tarball-assembly-only job... which I'm sure you're aware but I wanted to call that out.

I've pushed an update that keeps the module, generates into target/classes at process-classes, drops the committed JSON and the drift test. Output should be identical.

This does change things to where the json isn't being committed in the repo...

opennms-openapi-docs/pom.xml
```diff
@@ -0,0 +1,390 @@
...
+    <dependency>
+      <groupId>org.opennms.features.config.dao</groupId>
+      <artifactId>org.opennms.features.config.dao.api</artifactId>
+      <version>${project.version}</version>
+      <scope>provided</scope>
+    </dependency>
+
+    <!--
+      A module contributing to org.opennms.web.rest.v1 or .v2 and missing here loses
+      its endpoints silently; OpenApiDocsContentTest guards the ones known today.
```
mershad-manesh: We should add a test that walks the source tree for `@Path` classes and diffs against this module's classpath to catch a missing module.

marshallmassengill: Just pushed something for this.  Take a look.

opennms-openapi-docs/src/test/java/org/opennms/openapi/OpenApiResourceCoverageTest.java
```diff
@@ -0,0 +1,185 @@
...
+ * Walks the source tree for JAX-RS root resources and checks each one is on this
+ * module's classpath.
+ *
+ * A module that contributes to org.opennms.web.rest.v1 or .v2 and is not a
+ * dependency here still generates a valid document, just one that is missing
+ * those endpoints. Nothing else notices.
+ */
+public class OpenApiResourceCoverageTest {
+
+    private static final Path REPO_ROOT = Paths.get("..").toAbsolutePath().normalize();
```
mershad-manesh: This could be flaky.. For example what happens if we move the folder/file..

You can try adding something like this to the pom file
```
<plugin>
  <groupId>org.apache.maven.plugins</groupId>
  <artifactId>maven-surefire-plugin</artifactId>
  <configuration>
    <systemPropertyVariables>
      <opennms.repoRoot>${maven.multiModuleProjectDirectory}</opennms.repoRoot>
    </systemPropertyVariables>
  </configuration>
</plugin>
```


Then use 
```
private static final Path REPO_ROOT = Paths.get(System.getProperty("opennms.repoRoot"));
```

opennms-openapi-docs/src/test/java/org/opennms/openapi/OpenApiResourceCoverageTest.java
```diff
@@ -0,0 +1,185 @@
...
+
+    private static final Set<String> PRUNED = Set.of("target", ".git", "node_modules", ".claude");
+
+    /** The generator only registers resources in these packages and below. */
+    private static final Pattern REST_PACKAGE =
+            Pattern.compile("^\\s*package\\s+(org\\.opennms\\.web\\.rest\\.v[12](?:\\.[\\w.]+)?)\\s*;", Pattern.MULTILINE);
+
+    /** Guards against a parser change quietly reducing this to a no-op. */
+    private static final int EXPECTED_AT_LEAST = 50;
```
mershad-manesh: just a suggestion (thinking out loud here :) )...

Avoid hardcoded value, the ClassPath should give you a dynamic count. 

replace 
```
        assertTrue("found only " + rootResources + " root resources under " + REPO_ROOT
                        + "; the source scan is broken, not the classpath",
                rootResources >= EXPECTED_AT_LEAST);
```

with something like this..
```
import org.apache.cxf.common.util.ClasspathScanner;

    private static int classpathResourceCount() {
        int count = 0;
        for (final OpenApiDocGenerator.Api api : OpenApiDocGenerator.Api.values()) {
            count += ClasspathScanner
                    .findClasses(ClasspathScanner.parsePackages(api.getBasePackages()), javax.ws.rs.Path.class)
                    .getOrDefault(javax.ws.rs.Path.class, java.util.List.of())
                    .size();
        }
        return count;
    }


        final int classpathCount = classpathResourceCount();
        assertTrue("found only " + rootResources + " root resources under " + REPO_ROOT
                        + ", but " + classpathCount + " are on the classpath; the source scan is broken, not the classpath",
                rootResources >= classpathCount);
```

marshallmassengill: Check the latest.  I think I've got a fix for this now.

mershad-manesh (dismissed):
Overall the changes looks good. There are few issues which we should either resolve in this PR or in another PR.
* Double scrollbar on Endpoints page
* The page layout lags when we resize the window

marshallmassengill:
> Overall the changes looks good. There are few issues which we should either resolve in this PR or in another PR.
> 
> * Double scrollbar on Endpoints page
> * The page layout lags when we resize the window

I think I fixed everything you raised here.  Pushing it to this one.

synqotik:
@marshallmassengill here are 3 issues to consider, the first one being by far most important (may not case re Windows one):

1. /api-docs without a trailing slash serves a broken page (AbstractSwaggerUiResource.java). index.html is served at both api-docs and
  api-docs/, but from the slash-less URL every relative reference — the assets and the ../openapi.json spec URL — resolves one directory too high.
  A redirect from api-docs to api-docs/ is the standard fix. The shipped links all include the slash, so only hand-typed URLs hit it.
  2. The build will fail on Windows (opennms-openapi-docs/pom.xml). The forked generator passes the entire project classpath inline via
  -classpath; on this project that exceeds Windows' 32K CreateProcess limit. Ironic given the pom carries a Windows profile for java.exe.
  exec-maven-plugin's <longClasspath>true</longClasspath> fixes it.
  3. Deleting OpenApiIT leaves the shipping path untested. The serving resource deliberately degrades to 404 when the docs jar is absent, so a
  packaging regression (e.g., the webapp-full dependency being dropped) now ships green. Worth a smoke test that GET /opennms/rest/openapi.json
  returns 200 on a running instance.

synqotik (dismissed):
LGTM!

Just see my comment re `/api-docs without a trailing slash serves a broken page`, which should probably be fixed.

synqotik (approved):
LGTM! Thanks for the fixes.

PR #8749, opened 2026-08-05
NMS-20167: Sidemenu flicker, expand/collapse hotkey

synqotik:
The side menu was triggering menu item flyouts on mouseover. In particular, when the menu was collapsed, mouseover would expand the menu, then open the flyout menus. This made the menu "flicker" and seem unstable and unwieldy.

This fix removes that. The behavior is now that you need to click on an item (when either expanded or collapsed) in order to trigger menu flyout on mouseover. Note, this is now actually the native PrimeVue behavior. We were overriding it to mimic some behavior in Feather.

Added tooltips to the collapsed icons and expand/collapse button.

Added `Ctrl-\` (`Ctrl-Backslash`) hotkey to toggle menu expansion. Note that this is disabled if user is in an input/edit field.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20167

### DIFF
--- a/ui/src/components/Menu/SideMenu.vue
+++ b/ui/src/components/Menu/SideMenu.vue
@@ -17,6 +17,7 @@
     </button>
 
     <TieredMenu
+      ref="tieredMenuRef"
       :model="topPanels as never"
       class="onms-side-menu__menu"
       breakpoint="0px"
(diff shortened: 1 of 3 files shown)


marshallmassengill (approved):
LGTM!

Minor findings if you think any are worth addressing:

  1. SideMenu.vue:119 — togglePinned calls hide() unconditionally, and TieredMenu's hide() also resets focusedItemInfo to {index: -1}; a keyboard user mid-arrow-navigation who presses Ctrl+\ loses their position, and the next ArrowDown falls into findFirstFocusedItemIndex() and jumps back to the top of the menu. Guard on tieredMenuRef.value?.activeItemPath?.length.
  2. SideMenu.vue:29 — the tooltip is the only affordance for a collapsed root item's name, but PrimeVue binds focus/blur only with the .focus modifier, and the visible <span class="onms-side-menu__label"> is display:none when collapsed, which drops it from the accessibility tree. A tab-focused root icon therefore has no accessible name and no tooltip; v-onms-tooltip.focus plus an :aria-label on the <a> would close both.
  3. SideMenu.vue:11,29 — primevue-setup.ts sets zIndex.tooltip: 1110 while the rail is z-index: 2000 and .p-tieredmenu-submenu is 2001, so any tooltip overlapping the rail edge or an open flyout paints behind it. Most reachable via the toggle button's tooltip while a top flyout is open.
  4. SideMenu.vue:12 — the tooltip advertises Ctrl+\ but aria-label doesn't; aria-keyshortcuts="Control+\" on the button would expose it to AT.
  5. No test covers onGlobalKeydown (modifier exclusions, repeat, editable-target bail), which is the most regression-prone logic added; there is no SideMenu spec file at all, and utils.test.ts only asserts the topLevel flag.
  6. Dropping @mouseleave/onRailLeave means an open flyout now only closes on an outside click, so a position: fixed flyout visually detached from the rail can sit over page content indefinitely. Worth confirming it's the accepted trade-off for a nav rail.

synqotik:
> LGTM!
> 
> Minor findings if you think any are worth addressing:
> 

Thanks, will address some of these in a follow up.

PR #8756, opened 2026-08-06
NMS-20175: Node List UI Fixes - less default columns, indentation

synqotik:
## NMS-20175: Node List page UI fixes

Follow-up UI fixes for the Vue Node List page (NMS-20125):

- Trimmed default columns to Node Label, IP Address, Monitoring Location, Foreign Source, Foreign ID — Sys Contact/Location/Description and Flows are still available via Customize Columns, just no longer default, so headers aren't cramped. 

- Existing saved column preferences are unaffected.

- Expanded interface rows now align with the first data column instead of the far left of the row

-  Renamed the legacy "Node Structure" internals to "Node List" (useNodeListStore, buildNodeListQuery, etc.) — pure mechanical rename, no behavior change.

- Reodered the outages toggles in Advanced Node Filters, put the 2 outage-related ones together, with "Node with asset info only" afterwards

### Deferred until later

- Sorting by IP interface remains intentionally disabled: analysis (attached to the Jira) showed the v2 join-based ordering breaks pagination (row fan-out) and text-column IP ordering is lexicographic; a proper management-IP sort needs a derived property + index and is deferred.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20175


marshallmassengill (approved):
LGTM!

PR #8758, opened 2026-08-06
NMS-20174: Fix Menubar/SideMenu issues in Firefox and Safari on JSP pages

synqotik:
NMS-20174: Fix Vue Menubar/SideMenu issues in Firefox and Safari on JSP pages

The Vue menu (Menubar + SideMenu) failed to render or misbehaved in Safari and Firefox on legacy JSP pages, while appearing to work in Chrome.

Issues included:

- menus not displaying at all
- menus overlapping actual content
- logout not working (mostly in Firefox)
- other issues where assets were being aggressively cached (especially in Safari), so users were more likely to have stale CSS, etc.

Investigation with Playwright (WebKit/Firefox/Chromium against a live instance) traced the symptoms to six distinct defects, each fixed separately:

1. Menu bundle assets required authentication (applicationContext-spring-security.xml)
  
bootstrap.jsp references /ui-components/assets/index.{js,css} (the Vue menu code for JSP pages) from every page, including the unauthenticated login page. The unauthenticated request 302'd to login.jsp, and Safari cached that text/html response under the asset URL. After login, the module script was refused (since 'text/html' is not a valid JavaScript MIME type), so the menu never mounted and only the `:empty` CSS skeleton background showed. Chrome refetches instead of trusting the cached redirect, which masked the bug. Fixed by adding `/ui-components/**` to the anonymous static-asset rules,
  alongside the existing `/assets/**`, `/css/**`, etc.

  2. Login page unintentionally ran the menu app (bootstrap.jsp)

`login.jsp` requests quiet mode via `Bootstrap.flags("quiet")`, but the Vue-menu block only checked the param.quiet include parameter — so the login
  page always emitted the menu <script>. Fix #1 unmasked this: the menu then executed unauthenticated, its REST calls triggered the browser's native basic-auth prompt on the login page, and Spring Security captured a REST request as the post-login saved request — the password gate's Skip button then redirected to `/rest`, which browsers download as a zero-byte file. The menu block now honors both quiet mechanisms.

3. Saved REST/API requests as redirect targets (LoginModuleUtils)

Defense-in-depth for #2: `isInvalidSavedRequestUrl()` now rejects `/rest` and `/api servlet` paths (in addition to `.css/.js/.ttf`), so a captured REST XHR can never become a top-level post-login navigation, regardless of how it entered the request cache.

  4. Blocked localStorage aborted the menu mount (themeService.ts, localStorageService.ts)

Safari with "Block all cookies" (and Firefox with cookies disabled) throws `SecurityError` on any `localStorage` access. The menu app reads `localStorage` before `app.mount()` (to get the current theme) and during `SideMenu` setup (to check menu pin state), so the throw prevented the menu from ever mounting. All `localStorage` reads/writes are now guarded with sane fallbacks (default theme/preferences), matching the already-guarded inline script in
  bootstrap.jsp. Also tolerates corrupted JSON.

  5. Logout from JSP pages silently failed in Firefox (UserSelfServiceMenuItem.vue)

The Logout item is an anchor with `href="#"`, and JSP pages set `<base href>` — so `#` resolves against the base URL and an uncancelled click starts a full-page navigation racing the async logout POST. Firefox tears the XHR down before it is sent (the promise never settles), leaving the user logged in; Chrome/Safari usually let the POST win, masking the race. The click handler now cancels the anchor's default navigation.

  6. Stale CSS/JS skew from unversioned bundle URLs (add cache-busting for Vue/JSP in bootstrap.jsp)
  
The menu bundle uses fixed filenames, so Safari could hold a stale index.css alongside a fresh index.js; the stale CSS's scoped data-v hashes no longer match the new markup and the menu renders unstyled (tiny icons/text, no padding). The bundle links now carry a `?v=<file mtime>` cache-buster (same scheme used by `load-assets.jsp`), identical on the `modulepreload` and `script` tag so the URL-keyed module map fetches once.

**Note for users upgrading with previously affected Safari profiles:** one hard reload may be needed to evict cache entries poisoned before fix #1; from then on the versioned URLs prevent recurrence.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20174


marshallmassengill (approved):
LGTM.

Minor findings that may be worth looking at:

1. /ui-components/** is broader than the assets it justifies: the unpacked dist-menu artifact also ships /ui-components/index.html, a stray copy of the SPA bootstrap page (<script src="/opennms/ui/assets/index.js">), now anonymously reachable. Narrow to /ui-components/assets/**, or add a denyAll above it mirroring the /assets/load-assets.jsp* precedent two lines up.
2. Fix #2 changes 9 more JSPs than the description says: graph/forecast.jsp, graph/nrtg.jsp, admin/discovery/add-{er,ir,url,ex-url,specific}.jsp, admin/userGroupView/users/newPassword.jsp, includes/legendInfo-box.jsp all set Bootstrap.flags("quiet") and lose the Vue menu. Every one is a window.open popup, so the outcome looks right (it's likely the "menus overlapping content" symptom), but it should be stated and spot-checked rather than discovered by QA.
3. The head still emits the stylesheet and modulepreload unconditionally, so the login page still fetches a bundle it will never execute. If that preload isn't deliberate cache-warming for the post-login page, gating those two links on the same quiet condition would remove the need for fix #1's anonymous rule entirely.
4. UserSelfServiceMenuItem.test.ts "does not cancel navigation for regular items" does not test that. Mutation-checked: moving event?.preventDefault() above the logout branch leaves all 3 tests green. Needs expect(event.defaultPrevented).toBe(false).
5. The cache-buster does two getRealPath + File.lastModified() stats on every legacy page render; the load-assets.jsp precedent it cites reads a cached AssetLocator.lastModified(). Memoize in a ServletContext attribute.
6. ?v= from deployed-file mtime differs per node, so two servers behind a load balancer (or rolling container deploys) hand out different versions for identical bytes. Content-hashed filenames in vite.config.menu.ts plus a manifest would be the durable fix.
7. Nothing pins fixes #1 or #2, the two most silently-regressible. Two WebappIT assertions would cover it: unauthenticated GET of ui-components/assets/index.js returns 200 with a JS content type, and login.jsp's body contains no ui-components/assets/index.js script tag.
8. Fix #3 adds a second band-aid directly under the existing // TODO: Determine why assets are getting saved in the requestCache. A HttpSessionRequestCache.setRequestMatcher(...) excluding XHR / non-document requests would close the class instead of enumerating path shapes.
9. url.toLowerCase() has no Locale, so /API misses under -Duser.language=tr. Pre-existing for the suffixes, now inherited by the prefixes.

synqotik:
> LGTM.
> 
> Minor findings that may be worth looking at:

Agree, some of these should be fixed. Going to merge this so that things are usable in Safari, etc. and have a follow up ticket to fix some of these.

PR #8762, opened 2026-08-07
NMS-20161: Apply criteria fetch modes after the distinct rewrite for v1 /rest/alarms 

marshallmassengill:
I'm a bit torn on this one.  A customer reported it and this does seem to resolve the issue but I'm not sure if it's actually needed with the v2 API.  I'd like feedback on this at the very least.  It's probably worth fixing but this isn't urgent.

I also want to see if CI goes green for the full ITs.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20161

### DIFF
--- a/opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/HibernateCriteriaConverter.java
+++ b/opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/HibernateCriteriaConverter.java
@@ -23,8 +23,10 @@ package org.opennms.netmgt.dao.hibernate;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -127,6 +129,8 @@ public class HibernateCriteriaConverter implements CriteriaConverter<DetachedCri
 
         private Set<org.hibernate.criterion.Criterion> m_criterions = new LinkedHashSet<>();
 
+        private Map<String, FetchMode> m_fetchModes = new LinkedHashMap<>();
+
         private boolean m_distinct = false;
 
         private Integer m_limit;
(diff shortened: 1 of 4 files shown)


dino2gnt:
Seems low risk, and since you already did the work to fix it...

christianpape (dismissed):
LGTM.

dino2gnt (dismissed):
All green, shipit.

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/HibernateCriteriaConverter.java
```diff
@@ -229,18 +239,20 @@ public void visitAlias(final Alias alias) {
...
+            // held rather than applied here, because the distinct rewrite in
+            // getCriteria() replaces the criteria these would be set on
             switch (fetch.getFetchType()) {
             case DEFAULT:
-                m_criteria.setFetchMode(fetch.getAttribute(), FetchMode.DEFAULT);
+                m_fetchModes.put(fetch.getAttribute(), FetchMode.DEFAULT);
                 break;
             case EAGER:
-                m_criteria.setFetchMode(fetch.getAttribute(), FetchMode.JOIN);
+                m_fetchModes.put(fetch.getAttribute(), FetchMode.JOIN);
```
cgorantla, line 304: Found a possible bug that might happen : This change applies all eager fetches to the outer query.  `fetch("ipInterfaces", EAGER)` would multiply outer rows and break distinct() results and pagination. There are currently no such callers in the repository, but it is a valid API use.

Possible to somehow just fix it for lastEvent or may be explicit handling for Collection fetches ?

christianpape: Good catch!

marshallmassengill: Fixed by asking the session factory whether the fetch path is collection-valued and skipping the join for those, so distinct() and row-based paging both stay intact and the association loads the way it did before. The guard is limited to the distinct() path and to FetchMode.JOIN; a to-many join fetch without distinct() still multiplies rows as it always has, with a test to pin that down. Three ITs cover it, and two of them fail without the guard.

christianpape (approved):
LGTM.

marshallmassengill:
@cgorantla - can you make sure you're good with changes?

PR #8763, opened 2026-08-07
NMS-19725: Dependabot updates for Vue UI (August 2026)

synqotik:
# NMS-19725: Update Vue UI dependencies from open Dependabot PRs (August 2026)

Applies the open Dependabot updates for the Vue UI (`/ui`), remediates all osv-scanner findings with CVSS ≥ 4.0, and cleans up test-suite noise uncovered along the way.

## Dependency updates

| Library | From | To | Type | Notes |
|---------|------|----|------|-------|
| axios | 1.15.0 | 1.19.0 | minor | PR #8647. Newer header typings include `null`; added `as string \| undefined` on `content-type` reads in `useDownload.ts` and `snmpDataCollectionService.ts` |
| ace-builds | 1.43.6 | 1.44.0 | minor | PR #8557 |
| fast-xml-parser | 5.6.0 | 5.10.1 | minor | PR #8469 |
| fast-xml-builder | 1.1.4 | 1.3.0 | minor | PR #8514 — transient; pnpm override |
| typescript | ~5.4.5 | ~5.9.3 | minor | Prerequisite for the majors below; no code changes needed |
| vue-tsc | 2.2.12 | 3.3.9 | major | PR #8654. Stricter checking: one `value!` assertion in `OnmsTabs.vue` |
| @vitejs/plugin-vue | 5.2.4 | 6.0.8 | major | PR #8652. v6 ships `.d.mts`-only types; `tsconfig.json` switched to `"moduleResolution": "bundler"` (enabled by TS 5.9 + vue-tsc 3) |
| diff | 8.0.4 | 9.0.0 | major | PR #8653. Removed stale `@types/diff` (diff ships its own types since v6) |
| uuid | 10.0.0 | *removed* | — | PR #8480 targeted v14, but uuid is unused (last usage removed in NMS-18195); removed `uuid` + `@types/uuid` instead |
| happy-dom | 9.20.3 | 20.11.2 | major | Security (2× CVSS 9.3, 1× 7.5). Former v20 blocker was FeatherDS, now removed (NMS-19981) |

## Security remediations (osv-scanner, CVSS ≥ 4.0)

Transient packages fixed via `pnpm.overrides`: `shell-quote` → 1.10.0, `immutable` → 5.1.9, `svgo` → 3.3.4, `brace-expansion` → 2.1.4 / 5.0.9, `js-cookie` → 3.0.8, `js-yaml` → 4.3.1, `postcss` → 8.5.26. Remaining: only `vue@2.7.16` (CVSS 3.7, below threshold, unfixable transient).

## Test cleanup

- **`GeneralConfiguration.test.ts`**: pass numbers instead of strings to numeric refs bound to `OnmsInputNumber`, silencing `[Vue warn]` prop type-check output.
- **Nodes table tests** (`EventsTable`, `OutagesTable`, `IpInterfacesTable`, `SnmpInterfacesTable`): the store fetch action is now mocked *before* mounting. Previously the `onMounted` fetch ran the real action (`stubActions: false`) and fired an actual network request, surfacing as unattributed unhandled `ECONNREFUSED AggregateError` noise in test output. Also simplified the `OutagesTable` double-mount workaround this made unnecessary.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19725


marshallmassengill (changes requested):
There's a merge conflict.

Couple findings via claude.  Most of these are minor I would think but be aware of them:

2. Five override floors sit below the version the description credits with the fix. shell-quote ^1.9.0 against a claimed 1.10.0, immutable ^5.1.8 against 5.1.9, js-cookie ^3.0.7 against 3.0.8, postcss ^8.5.23 against 8.5.26, fast-xml-builder ^1.2.0 against 1.3.0. I checked the lockfile and every one currently resolves to the fixed version, so nothing is vulnerable today. But the range still permits a vulnerable resolution on any future lock regeneration, which is the single thing a security override exists to prevent. Set the floor to the fixed version.

3. The two axios casts silence the new null rather than handling it. useDownload.ts:74 and both sites in snmpDataCollectionService.ts feed the value straight into new Blob(..., { type: contentType }). BlobPropertyBag.type is a WebIDL DOMString, so a runtime null stringifies to "null" instead of being treated as absent, and the downloaded file carries Content-Type: null. ?? undefined is the same size of edit and actually handles the case the new typings added.

4. diff 8 to 9 is a major with no code change, and its only consumer is untested. DCBModalConfigDiffContent.vue:82 uses diffLines, and there is no test anywhere under ui/tests for that component. Worth checking the v9 diffLines return shape against it by hand.

  5. OnmsTabs.vue :value="value!" asserts non-null on a prop the seam declares optional with no default. Compile-time only, so behaviour is unchanged, but it answers a real question by silencing it. Either give value a default or make it required.

  6. happy-dom 9 to 20 is eleven majors in the test environment and required no test changes. That is either good news or a sign the suite is insensitive to what happy-dom emulates. Deliberately breaking one assertion to confirm the suite still fails would settle it cheaply.

synqotik:
> There's a merge conflict.
> 
> Couple findings via claude. Most of these are minor I would think but be aware of them:

Fixed all of these, plus targeted `release-36.x` instead of `develop`.

2. Lockfile conflict — should now resolve cleanly. I think it may have been due to PR incorrectly targeting `develop`

3. Override floors — all five bumped to the fixed versions (shell-quote ^1.10.0, immutable ^5.1.9, js-cookie ^3.0.8, postcss ^8.5.26, fast-xml-builder ^1.3.0). The lockfile diff only re-records the ranges; resolved versions were already the fixed ones.

4. Axios casts — the reviewer's diagnosis was right but the literal ?? undefined doesn't compile: axios types the header as string | string[] | number | boolean | AxiosHeaders | undefined, which is wider than BlobPropertyBag.type accepts. I used a typeof x === 'string' ? x : undefined narrow instead at all three sites — no cast at all, and it handles the runtime null (and any other non-string) the cast was silencing.

5. diff 8→9 — verified by hand against diff's release notes: every v9 breaking change is in parsePatch/formatPatch/ES5 target; the diffLines change-object shape (added/removed flags) is untouched. Also added the missing coverage: tests/components/Device/dcbModalConfigDiffContent.test.ts drives the component through checkbox selection → compare and asserts the rendered -N +N counts for both a single changed line and two non-contiguous changed regions, so a future diff major that alters the shape will fail visibly.

6. OnmsTabs — made value required rather than defaulted, since PrimeVue Tabs itself requires it and all 11 consumers already pass it. The ! and the withDefaults wrapper are gone; the existing contract test still passes.

7. happy-dom sensitivity — ran the mutation check: flipped one disabled-attribute assertion in deviceConfigBackup.test.ts, the suite failed on exactly that line, reverted. The assertions do read real happy-dom-rendered DOM state, so the clean 9→20 upgrade is genuinely good news, not an insensitive suite.

marshallmassengill (approved):
Ship it!

PR #8769, opened 2026-08-10
NMS-20200: Fix browser notifications in Firefox and on Chrome

marshallmassengill:
This should fix browser notifications.  I have not backported this since no one is complaining about it but I suspect we could.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20200

### DIFF
--- a/core/web-assets/src/main/assets/js/apps/notifications/index.js
+++ b/core/web-assets/src/main/assets/js/apps/notifications/index.js
@@ -24,40 +24,218 @@
 import Util from 'lib/util';
 import $ from 'vendor/jquery-js';
 
-$(function() {
-    if ('Notification' in window) {
-        let notificationSocket = null;
+const SERVICE_WORKER_PATH = 'notification-sw.js';
+const STREAM_PATH = 'notification/stream';
+const NOTICE_LIST_PATH = 'notification/browse?acktype=unack';
+const RECONNECT_MIN_MS = 1000;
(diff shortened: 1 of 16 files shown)


synqotik:
@marshallmassengill  Overall looks good! A couple small things worth fixing:

 1. connect() guard (core/web-assets/src/main/assets/js/apps/notifications/index.js:91): add if (socket !== null) { return; } at the top to close the narrow disconnect-then-restart race where a pending reconnect timer and a fresh start() could each open a socket.
 
  2. Title fallback (optional, show() in the same file): message.head || 'OpenNMS' so a misconfigured notificationCommands.xml (missing -subject) can't produce a notification titled "undefined" — since org.json's put() drops null-valued keys server-side.

marshallmassengill:
> @marshallmassengill Overall looks good! A couple small things worth fixing:
> 
>     1. connect() guard (core/web-assets/src/main/assets/js/apps/notifications/index.js:91): add if (socket !== null) { return; } at the top to close the narrow disconnect-then-restart race where a pending reconnect timer and a fresh start() could each open a socket.
> 
>     2. Title fallback (optional, show() in the same file): message.head || 'OpenNMS' so a misconfigured notificationCommands.xml (missing -subject) can't produce a notification titled "undefined" — since org.json's put() drops null-valued keys server-side.

Should be addressed now.

synqotik (approved):
LGTM!

PR #8770, opened 2026-08-10
Docs: Fix argument element order in notification command examples

marshallmassengill:
The XML examples put `<switch> before <substitution> inside <argument>`, but notificationCommands.xsd declares the sequence as substitution then switch.

scripting.adoc carried the same inversion before the webhook page existed (not my fault!!!) so I fixed it too.

This doc update brought to you by the letters X, S, and D... and assisted by Anthropic Claude Opus 5.


dino2gnt (approved):
lol

PR #8771, opened 2026-08-10
NMS-20201: Make top menubar more responsive at narrow window widths

synqotik:
Make the top menubar more responsive when width narrows.

This does the following:

- Keeps OpenNMS logo from being painted over by the Search box
- Makes Search box smaller for smaller screen widths
- For smaller widths, changes the Date to a Calendar icon with a tooltip

Specifically:

- 1440px: search at its 24em cap, full two-line date display
- 1100px: search shrinks fluidly, date text intact on single lines
- 1024px: date swaps to the calendar icon; hover shows a tooltip with the date text
- For lower widths, the menu bar (notification / self-service icons) will overflow off the right edge instead of overlapping 

Note: the search box width is now `clamp(10em, 28vw, 24em)`, so this intentionally changes wide/desktop viewports too — the previous fixed `30em` (480px) minimum is capped at `24em` (384px) at every width, not just narrow ones.

### Narrow width, Search box shrinks, Calendar icon for date, icons overflow

menu-at-670px

### Narrow width showing date tooltip

menubar-690-tooltip

### Narrow width showing narrower search box, but search results still wider

menubar-690-search

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20201

### DIFF
--- a/ui/src/components/Menu/Menubar.vue
+++ b/ui/src/components/Menu/Menubar.vue
@@ -29,11 +29,12 @@
 
       <span
         class="date-icon-wrapper"
-        v-onms-tooltip="`${formattedTime} ${formattedDate}`"
-        tabindex="0"
-        :aria-label="`${formattedTime} ${formattedDate}`"
+        v-onms-tooltip="dateTimeLabel"
       >
-        <OnmsIcon :icon="CalendarIcon" />
+        <OnmsIcon
+          :icon="CalendarIcon"
+          :title="dateTimeLabel"
+        />
       </span>
 
        <span title="Toggle Light/Dark Mode">
@@ -93,6 +94,11 @@ const displayAddNodeButton = computed(() => (mainMenu?.value.displayAddNodeButto
 
 const formattedDate = computed<string>(() => mainMenu.value?.formattedDate ?? '')
 const formattedTime = computed<string>(() => mainMenu.value?.formattedTime ?? '')
+
+// Empty (falsy) until the menu data loads, which keeps the tooltip unbound —
+// PrimeVue's string-form tooltip does not trim whitespace-only values and
+// would otherwise show an empty box.
+const dateTimeLabel = computed<string>(() => [formattedTime.value, formattedDate.value].filter(Boolean).join(' '))
 
 useOutsideClick(outsideClick.value, () => {
   resetMenuItems()
@@ -232,11 +238,14 @@ onUnmounted(() => {
 }
 
 // The logo is a wide wordmark SVG (viewBox 624x69). Size by height and let
-// width follow; override the SVG's own max-width so it isn't clamped/distorted.
+// width follow. The max-width is a guard for rebranded logos substituted via
+// VITE_APP_LOGO_NAME: since the left column refuses to shrink below the logo,
+// an unbounded ultra-wide asset would crowd out the rest of the bar. An SVG
+// wider than the cap letterboxes (scales down) rather than distorts.
 .onms-menubar__logo {
   height: 1.75rem;
   width: auto;
-  max-width: none;
+  max-width: 18rem;
 }
 
 // Notification-status colors stay on their light-mode values so the indicator
@@ -289,6 +298,7 @@ onUnmounted(() => {
   flex-direction: column;
   font-family: var(--onms-header-font-family);
   font-size: 0.875rem;
+  margin-left: 1em;
   margin-right: 1em;
   // Keep time/date each on one line; without this the text word-wraps and
   // spills out of the fixed-height bar when the right column is squeezed.
@@ -321,14 +331,8 @@ onUnmounted(() => {
     display: inline-flex;
     align-items: center;
     font-size: 24px;
+    margin-left: 1em;
     margin-right: 1em;
-    outline: none;
-
-    &:focus-visible {
-      outline: 2px solid var(--onms-primary);
-      outline-offset: 2px;
-      border-radius: 4px;
-    }
   }
 }
 </style>
--- a/ui/src/components/Menu/Menubar.vue
+++ b/ui/src/components/Menu/Menubar.vue
@@ -28,8 +28,9 @@
       </div>
 
       <span
+        v-if="dateTimeLabel"
         class="date-icon-wrapper"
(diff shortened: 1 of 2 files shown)


marshallmassengill (changes requested):
I think this first one is actually minor but leaving it in here:

1. Below 1024px the date is unreachable on touch and by keyboard. v-onms-tooltip with no modifier binds only mouseenter/mouseleave/click, and its click handler calls hide() (primevue/tooltip/index.mjs:112-120,186-190) — a tap never shows the tooltip. Keyboard focus binds nothing at all: the focus/blur listeners are added only under modifiers.focus (:104-110), so the tabindex="0" at Menubar.vue:33 is a focus stop that reveals nothing. <1024px is predominantly touch, i.e. the change removes the date exactly where it's meant to preserve it. Needs at minimum v-onms-tooltip.focus, and for touch either a tap-toggle or keeping the text visible at a smaller font.

2. aria-label on a bare <span> is not exposed. Menubar.vue:31-36: the span has no role, so it maps to generic, where aria-label is ARIA-prohibited (axe rule aria-prohibited-attr) and not reliably surfaced; the icon inside is aria-hidden="true" because no title prop is passed to OnmsIcon (OnmsIcon.vue:40-46). Combined with display: none on .date-wrapper, screen-reader users lose the date
entirely below 1024px and gain a nameless tab stop. The .light-dark block 6 lines below is the existing pattern: pass title to OnmsIcon, or add role="img" to the span.

Minor Findings

3. min-width: 24em on the results dropdown overflows off-viewport. Search.vue:292. The dropdown is left: 0; width: 100% inside the search wrapper, whose left offset is ~300px once the logo pins the left column; 24em (384px) then ends at ~685px. Inside a position: fixed header, that overflow generates no scrollbar, so the right edge of results is unreachable precisely at the widths where
the floor engages. Anchor it with right: 0; left: auto, or clamp against the viewport.

4. The right column now overflows instead of overlapping, taking the user menu with it. Both __left (after removing min-width: 0) and __right sit at min-width: auto while __center is flex: 0 0 auto, so all three are inflexible and the squeeze lands on __right. Below ~700px the notifications and self-service menus scroll off-screen and can't be clicked — overlapping icons were at least still hittable. The PR body accepts this, but collapsing the search to an icon at the same breakpoint would avoid it.

5. The left column is now pinned to the logo's full intrinsic width. Menubar.vue:206. With LogoHorizon (viewBox 624x69 at 1.75rem) that's ~253px; a wider rebranded logo via VITE_APP_LOGO_NAME moves finding 4's overflow point earlier. Worth a max-width on .onms-menubar__logo rather than relying on whatever asset the build substitutes.

6. Blank tooltip before mainMenu loads. formattedDate/formattedTime default to '', so the value is " " — truthy, and PrimeVue only trims empty values for the object form, not the string form (index.mjs:61-71). Transient empty tooltip box on hover during load.

7. Wide-viewport search shrinks too. Search.vue:282: 30em  24em cap applies at every width, not just narrow ones. Intended per the PR body, but it's a desktop-visible change and the description frames the work as narrow-width only.

mershad-manesh (dismissed):
Ran a quick test, overall the change looks good;
Minor issue that was observed has to do with the spacing between the Calendar, search box  and the notification Icon. 
On a small screen the Calendar Icon is really close to the search box.

Image

synqotik:
Thanks @marshallmassengill and @mershad-manesh for the reviews. All seven findings (Marshall) were checked against the sources (PrimeVue's tooltip directive and OnmsIcon), and most are addressed in the latest commits. I also found same issue as @mershad-manesh, so that is fixed as well.

Summary of what changed and what we deliberately left alone:
  
## Fixed

- Item 2 (aria-label not exposed) + Item 6 (blank tooltip during load): The date string is now a computed that filters out empty parts and joins, so it's empty (falsy) until the menu data loads — PrimeVue's string-form tooltip unbinds on a falsy value, which removes the transient empty tooltip box. The same computed is passed as title to OnmsIcon, so screen readers now get the date below 1024px. The unexposed aria-label and the tabindex on the wrapping span are removed — see Item 1 below for why the tab stop went away rather than getting wired up.

- Item 3 (dropdown off-viewport): The results dropdown is now anchored `right: 0; left: auto`. At widths where the input is wider than the 24em floor this renders identically; when the floor engages, the dropdown grows leftward and stays fully on-screen.

- Item 5 (unbounded logo width): `.onms-menubar__logo` now has max-width: 18rem — comfortably above both bundled wordmarks (~253px), so nothing changes today, but an ultra-wide rebranded logo letterboxes instead of pushing the overflow point earlier. (Consumers should be adhering to proper logo dimensions anyway.)

- Spacing (from @mershad-manesh's test, and I found same issue): Added 1em left margin to both the date text and the calendar icon so they no longer sit flush against the search box when squeezed.

- Item 4 (partial): Lowered the search floor from 12em to 10em, which offsets the added spacing and moves the right-column overflow point closer to ~600px.

## Deliberately not changed
  
- Item 1 (touch/keyboard tooltip): Touch is descoped — we don't currently support touchscreens, and doing that properly is a design effort of its own. For keyboard: the suggested `v-tooltip.focus` wouldn't work as a minimum fix, because PrimeVue's bindEvents is either/or — with the `.focus` modifier it binds only focus/blur and drops mouseenter/mouseleave, so we'd trade hover (the common case) for focus.
Instead the icon is now non-focusable and the date is exposed via the svg's accessible name, so screen readers get it; sighted keyboard-only users lose the date below 1024px, which we're accepting for now given it's the server's page-load timestamp rather than live data.

- Item 4 (search collapsing to an icon): Agreed this is the better long-term shape, but it introduces a new expand/collapse interaction that deserves real design. This PR is intentionally a tactical improvement; a more comprehensive responsive treatment of the menubar is planned as follow-up work, and the collapse idea belongs there. Below ~600px the bar still degrades by overflowing on the right, per the PR description.

- Item 7 (wide-viewport cap): Intentional — the 24em cap at desktop widths is part of the design (the old 30em box was oversized). The PR description has been updated to state this explicitly rather than framing the change as narrow-width-only.

marshallmassengill (approved):
LGTM!

PR #8772, opened 2026-08-11
NMS-20150: Fix plugin UI issues

synqotik:
Fix some issues found after the https://opennms.atlassian.net/browse/NMS-20054 work (PR: #8717).

### Changes

- **Plugin build contract**: replaced the unmaintained `vite-plugin-externals` with `rollup-plugin-external-globals` in the example plugin's build. Framework imports (`vue`, `pinia`, `vue-router`, `@opennms/onms-ui`) still resolve from the host page's window globals rather than being bundled, matching how production plugins are built. A new `verify-dist.mjs` script asserts the built artifact actually honors this contract (no residual framework imports, `window.*` rewrites present), since `vite build` exits 0 even when externalization silently fails.
- **Dev harness fixes** (`ui/vite.config.ts`): the plugin-module dev middleware is now installed after Vite's internal middlewares and matches on `req.originalUrl`, so it works with Vite's SPA fallback and host-check/CORS protections; it's gated behind `VITE_EXAMPLE_PLUGIN=true` to match the route gating in `main.ts`.
- **Vite config correctness**: both configs now use `defineConfig(({ mode }) => …)` with `loadEnv` instead of dotenv, so config-side env gates read exactly the same file set as `import.meta.env` in app code (including `.env.development[.local]`); aliases use `fileURLToPath` (fixes checkouts with spaces in the path) and the `ProductLogo` alias is absolute so dev and build resolve identically. `dotenv` dependency removed; `ui/*.local` env overrides are now gitignored.
- **CI**: the example plugin now typechecks, builds, and artifact-verifies in the UI CI job (`pnpm check:example-plugin`), acting as a contract test for the plugin build setup; it runs after the unit tests so test failures surface first.
- **Lint**: blanket ESLint ban on PrimeVue imports in plugin code (`primevue/**` — PrimeVue is not part of the host plugin contract); tidied the existing seam-layer messages.
- **Docs**: example plugin and onms-ui READMEs reworked for external plugin developers; the example plugin declares its `rollup` peer dependency explicitly since its config is meant to be copied.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20150


marshallmassengill (changes requested):
One blocker and a suggestion for a fix:

The new VITE_EXAMPLE_PLUGIN gate reads a different env source than the route it claims to match. main.ts:105 uses import.meta.env (Vite reads .env, .env.local, .env.development, .env.development.local); the config gate uses process.env after dotenv.config({ path: ['.env.local', '.env'] }), which skips the mode files. ui/.env.development is committed and already load-bearing here, so putting the flag in .env.development.local registers the route without the middleware: the module request falls through to the SPA fallback and returns index.html. Use defineConfig(({ mode }) => …) with loadEnv(mode, __dirname, '') instead of dotenv, which also fixes the same split for VITE_APP_LOGO_NAME, VITE_SERVICENOW_PLUGIN_DIST, and vite.config.menu.ts.

The rest of these are minor:

- The new CI contract test cannot catch the failure mode the externals swap introduces. externalGlobals' transform bails silently twice: the Object.keys(globals).every(id => !code.includes(id)) guard, and a this.parse() failure that only calls context.warn/debug and returns undefined (index.js:66-79). Paired with the explicit external: [...], a bail leaves a bare import { OnmsButton } from "@opennms/onms-ui" in the ES output; vite build still exits 0 and the break appears only in a browser. Under vite-plugin-externals the equivalent bail bundled Vue: fat but working. Have check:example-plugin assert the artifact (no from"vue" / from"@opennms/onms-ui" in dist, window.OnmsUI present), otherwise the gate only proves the build ran.
- external: ['vue', 'pinia', 'vue-router', '@opennms/onms-ui'] is redundant. The plugin's options hook unshifts its own resolver, whose resolveId returns false for every key in the globals map (index.js:31-35,55-65). Keeping it is what converts a silent transform bail into a runtime break rather than a bundled-but-working module, so it is a real choice: state it in the comment or drop it.
- group: ['primevue', 'primevue/*'] is not the blanket ban the comment describes. no-restricted-imports patterns use gitignore-style globs where * does not cross /, so primevue/a/b escapes. Use primevue/**. Low reachability against PrimeVue 4, which has no commonly used deep subpaths, but the stated intent is "ANY primevue import".
- rollup is an undeclared peer of the new dev dependency. rollup-plugin-external-globals@0.13.0 declares peerDependencies: { rollup: ^4.59.0 }; the example package lists only vite: ^6.4.3 and the lockfile satisfies the peer from the hoisted rollup@4.62.2. Since this README targets external plugin developers copying the config, add rollup to the package's devDependencies or call it out.
- .env.local is not gitignored. git check-ignore matches neither .env.local nor .env.development.local, and both configs now document .env.local as the place for dev-harness overrides. Add *.local to ui/.gitignore.
- CI ordering hides test results behind the new build. pnpm check:example-plugin && pnpm test puts a full extra vite build ahead of the unit tests; running it after pnpm test costs nothing.

synqotik:
@marshallmassengill all issues addressed. Fixed all except the `external` one, we are leaving that in as it signals intent.

1. env-source split fixed with loadEnv, dotenv removed
2. verify-dist.mjs asserts the built artifact honors the externals contract
3. kept external, corrected the comment (pushed back on the reviewer's tradeoff claim)
4. primevue/** glob fix, probe-verified against deep subpaths
5. rollup declared in the example package's devDependencies
6. ui/*.local gitignored
7. CI runs unit tests before the plugin contract check

mershad-manesh (approved):
The changes looks good to me.

marshallmassengill (approved):
LGTM!

PR #8773, opened 2026-08-11
NMS-19667: Use plexus-utils 3.6.1

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-19667


marshallmassengill (approved):
LGTM!  Once it goes green anyway... seems to be failing a test still.

PR #8774, opened 2026-08-11
NMS-19979: clear data collection alarms after a restart or collectd reload

marshallmassengill:
My attempt at a fix for the collectd reload bug. A dataCollectionFailed alarm never clears if collection recovers across an OpenNMS restart or a collectd configuration reload.

CollectableService keeps the last collection status in memory and updateStatus() only emits an event on a transition. That status was seeded to SUCCEEDED, and both a restart and rebuildScheduler() discard and rebuild every CollectableService, so the recovery is no longer a transition, no dataCollectionSucceeded is sent, and the alarm is orphaned. Seed it to UNKNOWN instead.
 
 - The first collection after a restart or reload always transitions: a success emits dataCollectionSucceeded and clears the orphaned alarm, a failure emits dataCollectionFailed as before.
 - Costs one extra event per collected service per restart or reload, not one per collection cycle.
 - An unmatched dataCollectionSucceeded creates a Normal severity alarm of its own, which the default alarmd cleanUp rule deletes after five minutes.
 - The default reduction key is node scoped (%uei%:%dpname%:%nodeid%), so any one service recovering clears the node's alarm.

Tests: two cases in CollectableServiceTest, one for the transition on first success and one confirming later successful cycles stay quiet.
  
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19979

### DIFF
--- a/opennms-services/src/main/java/org/opennms/netmgt/collectd/CollectableService.java
+++ b/opennms-services/src/main/java/org/opennms/netmgt/collectd/CollectableService.java
@@ -156,7 +156,10 @@ public class CollectableService implements ReadyRunnable {
         m_persisterFactory = persisterFactory;
 
         m_nodeId = iface.getNode().getId().intValue();
-        m_status = CollectionStatus.SUCCEEDED;
+        // Start out with no opinion rather than assuming success, so the first successful collection is a
+        // transition and emits the event that clears an alarm left over from a previous scheduling
+        // generation. A restart and a collectd reload both discard and rebuild every CollectableService.
+        m_status = CollectionStatus.UNKNOWN;
 
         m_updates = new CollectorUpdates();
 
--- a/opennms-services/src/test/java/org/opennms/netmgt/collectd/CollectableServiceTest.java
+++ b/opennms-services/src/test/java/org/opennms/netmgt/collectd/CollectableServiceTest.java
@@ -61,6 +61,8 @@ import org.opennms.netmgt.collection.support.builder.NodeLevelResource;
 import org.opennms.netmgt.dao.api.IpInterfaceDao;
 import org.opennms.netmgt.dao.api.ResourceStorageDao;
 import org.opennms.netmgt.dao.mock.MockEventIpcManager;
+import org.opennms.netmgt.events.api.EventConstants;
+import org.opennms.netmgt.events.api.EventIpcManager;
 import org.opennms.netmgt.events.api.EventIpcManagerFactory;
 import org.opennms.netmgt.model.OnmsIpInterface;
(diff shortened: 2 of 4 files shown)


dino2gnt:
Why not simplify this to clear all `dataCollectionFailed` alarms on startup instead?  Avoids the database hit trying to populate initial status and if it's actually failed, it will fail again on the first collection cycle and create a new alarm.

marshallmassengill:
> Why not simplify this to clear all `dataCollectionFailed` alarms on startup instead? Avoids the database hit trying to populate initial status and if it's actually failed, it will fail again on the first collection cycle and create a new alarm.

And this is exactly why I opened this as a draft.  Honestly... I'm not sure I can come up with a reason not to do that other than keeping alarms through reboots somehow seems important to me and feels like a breaking change... which is also why I'm not sure if this is a bug or an enhancement since it's changing behavior that's been like this for so long.

dino2gnt:
If that is the concern, you could unconditionally emit a `dataCollectionSucceeded` on every successful collection?

marshallmassengill:
> If that is the concern, you could unconditionally emit a `dataCollectionSucceeded` on every successful collection?

That's not bad... I suspect we would want to add something to vacuumd to clear those after some amount of time if they aren't picked up by the defaults.  I'd have to look at it.  I'll test this one.  That's definitely simple enough.

marshallmassengill:
> If that is the concern, you could unconditionally emit a `dataCollectionSucceeded` on every successful collection?

Ok... I don't think this one makes sense when you're doing something like Chance and have collections going at scale.  That could bog things down quite a bit with a lot of extra events.

dino2gnt:
If you set `m_status = CollectionStatus.UNKNOWN` instead of `SUCCEEDED`, only the first successful collect will emit a `dataCollectionSucceeded` which will clear any existing `dataCollectionFailed` from a previous cycle.

marshallmassengill:
Pushed an update to see if we can get this simplified a bit based on @dino2gnt's suggestion.

cgorantla (dismissed):
This looks clean.

On develop, we should add packageName to `dataCollectionFailed` and `dataCollectionSucceeded` events so that we could differentiate between failures on a given package ( may be because of specific parameters)

marshallmassengill:
> This looks clean.
> 
> On develop, we should add packageName to `dataCollectionFailed` and `dataCollectionSucceeded` events so that we could differentiate between failures on a given package ( may be because of specific parameters)

I'll see about getting an NMS open for that once this gets merged.  It does seem like a useful addition.

christianpape (approved):
LGTM.

cgorantla (approved):
LGTM!

PR #8778, opened 2026-08-11
NMS-20203: Kafka sink consumer manager tracks unstarted consumers, leaking live ones on reload

marshallmassengill:
Found what I think is a pretty minor bug.

KafkaMessageConsumerManager.startConsumingForModule() added a second, never-started KafkaConsumerRunner to its tracking list instead of the one it started, so stopConsumingForModule() shut down inert objects while the live consumers kept running.

- Orphaned consumers keep their Kafka group partitions but dispatch against a SinkModule key that is no longer registered, so AbstractMessageConsumerManager.dispatch() drops the record while auto-commit advances the offset. Telemetryd and Sentinel adapter reloads lose telemetry in proportion to the partitions those consumers hold, compounding per reload.
- The unstarted duplicates never subscribe, so throughput was unaffected, but each held 2 file descriptors and 6 JMX MBeans that the MBean registration kept alive through GC.
- shutdown() used executor.shutdown(), which never ended the poll loops; it now stops the runners first, then bounds the wait with a shutdownNow() fallback.
- New IT asserts the consumer group drains to zero members after unregistering. It times out against the unfixed code.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20203

### DIFF
--- a/core/ipc/sink/kafka/itests/src/test/java/org/opennms/core/ipc/sink/kafka/itests/HeartbeatSinkIT.java
+++ b/core/ipc/sink/kafka/itests/src/test/java/org/opennms/core/ipc/sink/kafka/itests/HeartbeatSinkIT.java
@@ -40,6 +40,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.kafka.clients.admin.AdminClient;
 import org.apache.kafka.clients.admin.AdminClientConfig;
+import org.awaitility.core.ConditionFactory;
 import org.junit.Before;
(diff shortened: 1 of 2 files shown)


dino2gnt (commented):
Line R273.  I'm guessing this is the actual important bit for this issue?

dino2gnt (dismissed):
Sure.

christianpape (approved):
LGTM.

cgorantla (approved):
LGTM!

PR #8779, opened 2026-08-11
NMS-20204: Reduce per-event overhead in the event translator

marshallmassengill:
EventTranslatorConfigFactory repeated most of its work on every event.

- sql values query once per event instead of twice
- Regexes compile at load, not per event (and not per parm)
- Events clone only after a mapping matches, not before
- cloneEvent() copies fields instead of serializing: ~52 µs → ~0.74 µs
- m_translationSpecs made volatile for the reload race

Per trap on the shipped link-down config: 6 DB round trips → 3, regex compilations → 0.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20204


opennms-config/src/main/java/org/opennms/netmgt/config/EventTranslatorConfigFactory.java
```diff
@@ -790,10 +757,8 @@ public String getAttributeValue(Event e) {
                     return (parm.getValue() == null ? "" : parm.getValue().getContent());
                 }
 
-                String trimmedAttrName = StringUtils.removeStart(attrName, "~");
-
-                if (attrName.startsWith("~") && (parm.getParmName().matches(trimmedAttrName))) {
-                    LOG.debug("getAttributeValue: eventParm name: '{} matches translation parameter name expression: ' {}", trimmedAttrName, parm.getParmName());
+                if (m_namePattern != null && m_namePattern.matcher(parm.getParmName()).matches()) {
+                    LOG.debug("getAttributeValue: eventParm name: '{} matches translation parameter name expression: ' {}", parm.getParmName(), m_namePattern);
```
christianpape: I agree, the quotes of these two LOG.debug() statements in ParameterValueSpec.getAttributeValue() are weird. Please fix.

```suggestion
LOG.debug("getAttributeValue: eventParm name: '{}' equals translation parameter name: '{}'", attrName, parm.getParmName());
return (parm.getValue() == null ? "" : parm.getValue().getContent());
}

                if (m_namePattern != null && m_namePattern.matcher(parm.getParmName()).matches()) {
                    LOG.debug("getAttributeValue: eventParm name: '{}' matches translation parameter name expression: '{}'", parm.getParmName(), m_namePattern);
```

christianpape (changes requested):

opennms-config/src/main/java/org/opennms/netmgt/config/EventTranslatorConfigFactory.java
marshallmassengill: :+1:

christianpape (approved):
LGTM.

cgorantla (approved):
LGTM!

PR #8784, opened 2026-08-12
NMS-20208: Replace adhoc graphs workflow

dino2gnt:
Replace the jsp adhoc graphs workflow with a modern, feature-rich Vue SPA.

This also contains fixes for issues encountered while developing it:
* The time-picker had broken durations for 7 of 12 choices
 * The Data tab in the Vue graphs was backwards, and lacked a real timestamp (was HH:MM)
 * OnmsTabs.test.ts was flakey and failed about 1 time in 3

image

image

image

image

[Ad-Hoc_Graph.pdf](https://github.com/user-attachments/files/30990615/Ad-Hoc_Graph.pdf)

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20208

### DIFF
--- a/ui/src/components/AdhocGraphs/AdhocChartToolbar.vue
+++ b/ui/src/components/AdhocGraphs/AdhocChartToolbar.vue
@@ -1,0 +1,226 @@
+<template>
+  <div class="chart-toolbar">
+    <div class="toolbar-row">
+      <TimeControls @updateTime="(value: StartEndTime) => emit('updateTime', value)" />
+
+      <div class="toolbar-actions">
+        <OnmsButton
+          variant="outlined"
+          :disabled="!canQuery"
+          :loading="loading"
+          data-test="toolbar-refresh"
+          @click="emit('refresh')"
+        >Refresh</OnmsButton>
+        <OnmsIconButton
+          variant="outlined"
+          :icon="DownloadFile"
+          title="Download this graph's data as CSV"
+          :disabled="!hasData"
+          data-test="toolbar-csv"
+          @click="emit('exportCsv')"
+        />
+        <OnmsIconButton
+          variant="outlined"
+          :icon="PdfIcon"
+          title="Download this graph as a PDF"
+          :disabled="!hasData"
+          data-test="toolbar-pdf"
+          @click="emit('exportPdf')"
+        />
+        <OnmsIconButton
+          variant="outlined"
+          :icon="LinkIcon"
+          title="Copy a link to this graph"
+          :disabled="!canQuery"
+          data-test="toolbar-share"
+          @click="emit('share')"
+        />
+        <OnmsIconButton
+          v-if="!viewOnly"
+          variant="outlined"
+          :icon="CodeIcon"
+          title="Show this graph as an RRDtool graph definition"
+          :disabled="!canQuery"
+          data-test="toolbar-definition"
+          @click="emit('showDefinition')"
+        />
+        <OnmsIconButton
+          v-if="!viewOnly"
+          variant="outlined"
+          :icon="expanded ? FullscreenExitIcon : FullscreenIcon"
+          :title="expanded ? 'Show the selectors again' : 'Expand the graph, hiding the selectors'"
+          data-test="toolbar-expand"
+          @click="emit('toggleExpand')"
+        />
+        <OnmsIconButton
+          v-if="!viewOnly"
+          variant="outlined"
+          :icon="PopOutIcon"
+          title="Open this graph on its own, in a new tab"
+          :disabled="!canQuery"
+          data-test="toolbar-popout"
+          @click="emit('popOut')"
+        />
+        <OnmsButton
+          v-if="!viewOnly"
+          variant="ghost"
+          data-test="toolbar-clear"
+          @click="emit('clear')"
+        >Clear all</OnmsButton>
+      </div>
+    </div>
+
+    <div
+      v-if="!viewOnly && !expanded"
+      class="toolbar-row toolbar-fields"
+    >
+      <FormField
+        label="Graph title"
+        for="adhoc-title"
+        class="field-title"
+      >
+        <OnmsInputText
+          id="adhoc-title"
+          :modelValue="config.title"
+          placeholder="Untitled ad-hoc graph"
+          data-test="toolbar-title"
+          @update:modelValue="value => emit('update', { title: (value ?? '') as string })"
+        />
+      </FormField>
+
+      <FormField
+        label="Vertical label"
+        for="adhoc-vlabel"
+        class="field-vlabel"
+      >
+        <OnmsInputText
+          id="adhoc-vlabel"
+          :modelValue="config.verticalLabel"
+          placeholder="e.g. bits per second"
+          data-test="toolbar-vlabel"
+          @update:modelValue="value => emit('update', { verticalLabel: (value ?? '') as string })"
+        />
+      </FormField>
+
+      <FormField
+        label="Data points"
+        for="adhoc-resolution"
+        class="field-resolution"
+        hint="Samples across the range."
+      >
+        <OnmsInputNumber
+          inputId="adhoc-resolution"
+          :modelValue="config.resolution"
+          :min="10"
+          :max="MAX_RESOLUTION"
+          data-test="toolbar-resolution"
+          @update:modelValue="value => emit('update', { resolution: Number(value) || DEFAULT_RESOLUTION })"
+        />
+      </FormField>
+
+      <FormField
+        label="Stack series"
+        for="adhoc-stacked"
+        class="field-stacked"
+      >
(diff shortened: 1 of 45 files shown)


marshallmassengill (dismissed):
One blocker:

~ as the URL field separator can silently corrupts expressions.  Percent-encode the
  fields, or reject ~ in the expression editor.

Minor stuff too but not a real blocker:

  1. The copied .properties block has no reports= line, so pasting it into a new file under snmp-graph.properties.d/ yields nothing. The dialog's intro text says to add the report name to that list, but the copy button is the affordance and the copied text is what lands in the file — a # add <name> to reports= comment in the block would close it.
  2. report.<name>.name=${escapeForProperties(title)} doesn't strip newlines the way escapeQuoted does for the command, so a multi-line title breaks the properties block.
  3. Report ids collide by construction: two different ad-hoc graphs over the same resource type and attributes both emit adhoc.interfaceSnmp.ifHCInOctets_ifHCOutOctets, and the second silently replaces the first.
  4. applyCustomTime still doesn't check start < end — an end date before the start emits a reversed window labelled minutes, and disableCustomTimeBtn only tests for presence.
  5. searchNodes interpolates the raw term into FIQL (label==*${term}*), so a , or ; in the search box produces a malformed filter, a failed request, and an empty list with no explanation.
  6. mapWithConcurrency has no per-item error isolation; it depends entirely on getResourceById/getResourceForNode swallowing their own errors. Any worker that rejects takes down Promise.all,
  propagates out of loadResources, and leaves resourcesLoading stuck at true.
  7. syncUrl silently blanks the address bar past MAX_QUERY_LENGTH. buildShareUrl's callers tell the user; the passive path doesn't, so a large graph loses bookmarkability with no signal until someone tries to copy.
  8. expressionIssues errors only when every identifier is unknown, so in * 8 + typo passes local validation and fails at the server — deliberate per the comment, but it lets the common single-typo case through.
  9. displayTime hardcodes EEE MMM d HH:mm:ss yyyy while the comment justifies it as matching d3's %c "in the default locale"; %c is locale-adaptive and this is not.

ui/src/components/AdhocGraphs/RrdDefinitionDialog.vue
```diff
@@ -0,0 +1,134 @@
...
+      >{{ definition.properties }}</pre>
+      <p class="definition-note">
+        The time range is not part of a definition &mdash; the viewer chooses it.
+        Consolidation, colors, styles and expressions are carried over as-is.
+      </p>
+    </div>
+
+    <template #footer>
+      <OnmsButton
+        v-if="definition"
```
synqotik, line 46: Can make this `variant="filled"`

ui/src/components/AdhocGraphs/AdhocChartToolbar.vue
```diff
@@ -0,0 +1,226 @@
+<template>
+  <div class="chart-toolbar">
+    <div class="toolbar-row">
+      <TimeControls @updateTime="(value: StartEndTime) => emit('updateTime', value)" />
+
+      <div class="toolbar-actions">
+        <OnmsButton
```
synqotik: Maybe make this also an `OnmsIconButton` with the following icon.

```
import Refresh from '@/components/icons/navigation/Refresh.vue'
```

ui/src/components/AdhocGraphs/AdhocChartToolbar.vue
```diff
@@ -0,0 +1,226 @@
...
+        <OnmsIconButton
+          v-if="!viewOnly"
+          variant="outlined"
+          :icon="PopOutIcon"
+          title="Open this graph on its own, in a new tab"
+          :disabled="!canQuery"
+          data-test="toolbar-popout"
+          @click="emit('popOut')"
+        />
+        <OnmsButton
```
synqotik: Maybe `OnmsIconButton` with the `actions/Cancel` icon?

Just so this looks consistent and items are also all the same height.

Also, you can add `v-onms-tooltip` property to all these `OnmsIconButton` to get a nice, larger/more readable tooltip.

ui/src/components/AdhocGraphs/AdhocChartToolbar.vue
```diff
@@ -0,0 +1,226 @@
+<template>
+  <div class="chart-toolbar">
+    <div class="toolbar-row">
+      <TimeControls @updateTime="(value: StartEndTime) => emit('updateTime', value)" />
+
+      <div class="toolbar-actions">
+        <OnmsButton
+          variant="outlined"
+          :disabled="!canQuery"
```
synqotik: Change to `OnmsIconButton` with `v-onms-tooltip` (see comment below) and use following icon:

```
import Refresh from '@/components/icons/navigation/Refresh.vue'
```

ui/src/components/AdhocGraphs/AdhocChartToolbar.vue
```diff
@@ -0,0 +1,226 @@
+<template>
+  <div class="chart-toolbar">
+    <div class="toolbar-row">
+      <TimeControls @updateTime="(value: StartEndTime) => emit('updateTime', value)" />
```
synqotik: There should probably be some kind of label for this like `Time Range:`. Otherwise unless you open it, kind of hard to know what it's doing (we need to fix this on Resource Graphs as well...).

Could try wrapping it in a `FormField`, or else just similar-looking text to the left or above it.

ui/src/components/AdhocGraphs/RrdDefinitionDialog.vue
```diff
@@ -0,0 +1,134 @@
...
+        variant="outlined"
```
synqotik: Think this should be `variant="filled"`.

ui/src/components/AdhocGraphs/SelectionColumn.vue
```diff
@@ -0,0 +1,234 @@
...
+      :modelValue="filterTerm"
+      :placeholder="filterPlaceholder"
+      :ariaLabel="`Filter ${title}`"
+      :dataTest="`${dataTest}-filter`"
+      @update:modelValue="onFilter"
+    />
+
+    <div class="column-actions">
+      <OnmsButton
+        variant="text"
```
synqotik: I think should be `variant="ghost"`.

ui/src/components/AdhocGraphs/SelectionColumn.vue
```diff
@@ -0,0 +1,234 @@
...
+        :disabled="!visibleOptions.length"
+        :data-test="`${dataTest}-select-all`"
+        @click="selectAllVisible"
+      >Select all</OnmsButton>
+      <OnmsButton
+        variant="text"
```
synqotik: I think should be `variant="ghost"`.

synqotik (changes requested):
Overall looks great. I added some picky comments.

I do think there may be better ways to organize some things, but will take some time to determine exactly how. Seems pretty usable as-is, so would just be refinements.

My suggestion would be to change menu item to `Adhoc Graphs (Preview)`. Let's get this into `develop` and people can use it and we can iterate on the UX, plus other suggestions.

dino2gnt:
>     2. report..name=${escapeForProperties(title)} doesn't strip newlines the way escapeQuoted does for the command, so a multi-line title breaks the properties block.
>     3. Report ids collide by construction: two different ad-hoc graphs over the same resource type and attributes both emit adhoc.interfaceSnmp.ifHCInOctets_ifHCOutOctets, and the second silently replaces the first.
>     4. applyCustomTime still doesn't check start < end — an end date before the start emits a reversed window labelled minutes, and disableCustomTimeBtn only tests for presence.
>     8. expressionIssues errors only when every identifier is unknown, so in * 8 + typo passes local validation and fails at the server — deliberate per the comment, but it lets the common single-typo case through.

I'm choosing to ignore these for now.

synqotik:
Fixes look good. Just remembered two more things.

- Need a title. See `src/containers/SnmpConfiguration.vue` for an example, something like this:

```
    <div class="header">
      <div class="heading">
        <h2>Custom Peformance Graphs</h2>
      </div>
    </div>
```

- Need a `smoke-test` in `MenuHeaderIT.java` for this. There's already ones for `Resource Graphs` there. It can just check for the above header/h2 title.

synqotik (dismissed):
LGTM!

May want to wait for @marshallmassengill to concur.

marshallmassengill (dismissed):
LGTM!  I think this should make it into release-36.x instead of waiting for 37. What do you think @synqotik?

synqotik:
> LGTM! I think this should make it into release-36.x instead of waiting for 37. What do you think @synqotik?

@marshallmassengill @dino2gnt 
OK with me.

marshallmassengill (approved):
Ship it!

synqotik (approved):
LGTM!

PR #8793, opened 2026-08-17
NMS-20243: Move icons to onms-ui

synqotik:
## Overview

Move the Vue icons into `packages/onms-ui`, along all the other "seam" UI components. This is cleaner organizationally, will be good if/when we eventually publish this package, allows plugin authors to more easily consume the icons (and thus use "approved" icons).

## Summary

The icon set (262 SVG single-file components, originally vendored from FeatherDS) moves from `ui/src/components/icons/` into the seam package, alongside the `OnmsIcon` wrapper that renders it:

```
ui/src/components/icons/<category>/<Name>.vue
  → ui/packages/onms-ui/src/icons/<category>/<Name>.vue
```

Icons are exposed through a new `"./icons/*"` **subpath export** — deliberately *not* the package barrel. Call sites change specifier only:

```diff
-import DeleteIcon from '@/components/icons/action/Delete.vue'
+import DeleteIcon from '@opennms/onms-ui/icons/action/Delete.vue'
```

## Why

Under `@/components/icons/`, the icons were reachable **only** through the core app's own Vite alias (`ui/vite.config.ts`). Two consequences:

- **`OnmsIcon` was a wrapper around nothing it owned.** The package shipped the renderer; the app shipped every payload. The useful unit was split across a package boundary, and `OnmsIcon`'s own doc comment described the icons as belonging to "the consuming app."
- **No plugin could import an icon at all.** An external plugin repo has no `@/` alias into the OpenNMS `ui/src` tree and no way to install it, so its only option was copying SVG files into its own repo.

The icons now sit behind a real package specifier, which is a stable, documentable public path.

## Additional

Also fixed some unit test noise (unrelated but small fix).

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20243


marshallmassengill (approved):
No notes.  Ship it.

PR #8794, opened 2026-08-17
NMS-6400: OpenNMS syslogd allow receiving messages on TCP (and via TLS)

marshallmassengill:
Adds syslog ingestion over TCP, with optional TLS, for both the core and Minions. Off unless configured, so existing installs keep listening on UDP alone.

  - syslog-tcp-port in syslogd-configuration.xml starts a TCP listener alongside the UDP one. Unset by default.
  - Both RFC 6587 framings are supported. The default auto reads the framing from the first message of a connection and latches it.
  - TLS follows RFC 5425, with none, optional or require client authentication. Unusable certificate material stops the listener rather than falling back to plaintext.
  - A Minion configures the same thing as syslog.tcp.* keys on the syslog listener feature it already installs.
  - Messages reach the existing sink, one connection's messages in order, with reads paused while a dispatch is outstanding so a slow sink becomes TCP backpressure.
  - Verified end to end against real rsyslog and syslog-ng: both framings against plaintext, TLS and mutual TLS, on both ingestion paths. New unit tests, integration tests and a Minion smoke test.

This pull request brought to you by the letter N for necromancy and assisted by Anthropic Claude Opus 5.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-6400


dino2gnt:
:crown: 

You trying to steal my "King Shit of Developing Features No One Asked For" crown ?

Unless there's a _really good reason_, this needs to target `develop`, IMO.

marshallmassengill:
> :crown: 
> 
> 
> 
> You trying to steal my "King Shit of Developing Features No One Asked For" crown ?
> 
> 
> 
> Unless there's a _really good reason_, this needs to target `develop`, IMO.

LOL

I can target it at develop. We do have a semi-important request for this but I'm good to put it on develop and backport later.  It does need testing.

cgorantla (changes requested):
Would prefer to model this as

```
<configuration
      listen-address="0.0.0.0"
      syslog-port="514"
      parser="...">

      <tcp port="601"
           listen-address="0.0.0.0"
           framing="auto"
           max-message-size="65536"
           max-connections="1024"
           idle-timeout="0">

          <tls enabled="true"
               cert-filepath="..."
               private-key-filepath="..."
               trust-cert-filepath="..."
               client-auth="require"/>
      </tcp>
  </configuration>
```

This way `tcp` remains completely optional and `tls` is embedded within `tcp` and can be extended if we need to add more settings.

.cfg can retain flat but should be able to populate the tcp config object.

marshallmassengill:
> Would prefer to model this as
> 
> ```
> <configuration
>       listen-address="0.0.0.0"
>       syslog-port="514"
>       parser="...">
> 
>       <tcp port="601"
>            listen-address="0.0.0.0"
>            framing="auto"
>            max-message-size="65536"
>            max-connections="1024"
>            idle-timeout="0">
> 
>           <tls enabled="true"
>                cert-filepath="..."
>                private-key-filepath="..."
>                trust-cert-filepath="..."
>                client-auth="require"/>
>       </tcp>
>   </configuration>
> ```
> 
> This way `tcp` remains completely optional and `tls` is embedded within `tcp` and can be extended if we need to add more settings.
> 
> .cfg can retain flat but should be able to populate the tcp config object.

I attempted to address in latest.  Check and let me know if that is what you had in mind.

opennms-config/src/main/java/org/opennms/netmgt/config/SyslogdConfig.java
```diff
@@ -152,6 +153,15 @@ public interface SyslogdConfig {
...
     boolean shouldIncludeRawSyslogmessage();
 
+    /**
+     * The settings for the TCP listener, which is disabled unless a TCP port has been
+     * configured. Never null, so callers check {@link SyslogTcpConfig#isEnabled()}
+     * rather than the return value.
+     *
+     * @return the TCP listener settings
+     */
+    SyslogTcpConfig getTcpConfig();
```
indigo423: When we add this new abstract method breaks implementer at compile time. Making it a default returning new SyslogTcpConfig() costs not much and keeps those implementers building and resolves the "Never null" contradiction in the javadoc above.

features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/SyslogTcpFrameDecoder.java
```diff
@@ -0,0 +1,251 @@
...
+            // non-transparent framing before the first real message.
+            skipEmptyFrameTrailers(in);
+            if (!in.isReadable()) {
+                return;
+            }
+            final byte first = in.getByte(in.readerIndex());
+            activeFraming = (first >= '0' && first <= '9')
+                    ? SyslogTcpFraming.OCTET_COUNTING
+                    : SyslogTcpFraming.NON_TRANSPARENT;
+            LOG.info("Detected {} framing for syslog TCP connection from {}", activeFraming, source);
```
indigo423, line 101: Not really a blocker, RFC 3164 / 5424 start with `<`. It misfires on a sender whose lines start with a digit and carry no PRI. I sent `2026-08-25 device reported something` and got octet-counting latched, then `CorruptedFrameException: unexpected byte 0x2D in the length prefix`. The `0x2D` being the hyphen in the date. The sender reconnects, relatches, and dies again, so the operator sees "syslog over TCP doesn't work" with one WARN per connection, no metric and no event. Latching is still the right design over per-frame guessing. What's missing is a counter for connections dropped on framing errors, which would make this self-diagnosing.

features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/SyslogTcpFrameDecoder.java
```diff
@@ -0,0 +1,251 @@
...
+                }
+                return;
+            }
+
+            final int frameLength = trailerIndex - start;
+            if (frameLength > maxMessageSize) {
+                throw new TooLongFrameException(String.format(
+                        "Syslog frame from %s is %d bytes, which exceeds the %d byte maximum",
+                        source, frameLength, maxMessageSize));
+            }
```
indigo423, line 211: When we throw this exception here, it reaches the `SyslogTcpExceptionHandler`, which closes the channel. For the octet-counting path at :154 that's correct.

The position of the next message is genuinely unknown once you distrust the length prefix. Under non-transparent framing it isn't: `trailerIndex` is already the delimiter, so skipping to it and continuing would drop the one oversized message instead of the connection.

As written, a sender that periodically emits one long line reconnects in a loop and loses whatever was in flight each time. Confirmed with a 70053-byte line: WARN, connection closed, 0 events, next connection accepted normally.

I would treat this as a follow-up, not a blocker.

features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/SyslogTcpListener.java
```diff
@@ -0,0 +1,402 @@
...
+                    Throwable failure = null;
+                    try {
+                        final CompletableFuture<?> dispatched = m_dispatcher.send(next);
+                        if (waitForDispatch) {
+                            try {
+                                dispatched.get(m_dispatchTimeoutMs, TimeUnit.MILLISECONDS);
+                            } catch (TimeoutException e) {
+                                // Paying this per message is worse than losing the ordering
+                                // guarantee, so stop waiting for the rest of the connection.
+                                waitForDispatch = false;
```
indigo423, line 334:
```java
 76        private static final long DEFAULT_DISPATCH_TIMEOUT_MS = 30_000;
...
103        /** Overridden only by tests, which cannot afford to wait out the real timeout. */
104        private long m_dispatchTimeoutMs = DEFAULT_DISPATCH_TIMEOUT_MS;
```

That's one we should take a closer look and has some smell to me. The commit message frames this timeout as a workaround for the situation a Minion config reload leaving the sing completing the wrong futures. It seems like broader than that.

During my test it fired 9 times on the **Core**, with no reload anywhere. I had 9 distinct connections in one 30-second window. So the actual rule is that any sink slow enough to leave a future uncompleted for 30 s permanently downgrades that connection for its remaining life, losing in-order delivery and read-side backpressure together, with only the sink queue's `blockWhenFull` left underneath. 

The trade-off you describe in the comment is the right one to make. My concern is that a busy install will hit it routinely rather than exceptionally. 

Two asks: re-scope the commit message so the next person reading it doesn't assume reload-only, and lift the timeout into `SyslogTcpConfig` so it isn't a constant only tests can reach. Also worth deciding whether the one-shot WARN is enough. The connection stays degraded silently afterwards.

features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/SyslogTcpListener.java
```diff
@@ -0,0 +1,402 @@
...
+                            // Counts channels that are still closing: they hold resources until
+                            // their close completes, so a connection-per-message sender can be
+                            // refused below its apparent concurrency.
+                            if (m_channels.size() >= m_config.getMaxConnections()) {
+                                // Debug: a client retrying in a loop would fill the log.
+                                LOG.debug("Refusing syslog TCP connection from {}: already at the {} connection limit",
+                                        ch.remoteAddress(), m_config.getMaxConnections());
+                                ch.close();
+                                return;
+                            }
```
indigo423, line 203: Just a hint: Since the refusal is at DEBUG by design, an operator hitting the cap has no visible signal at default log levels. If you add the framing-error counter from comment above, a refused-connection counter alongside it would come nearly free.

features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/SyslogTcpListener.java
```diff
@@ -0,0 +1,402 @@
...
+        boolean isBusy() {
+            return dispatchInFlight || !pending.isEmpty();
+        }
+
+        @Override
+        protected void channelRead0(final ChannelHandlerContext ctx, final SyslogConnection connection) {
+            pending.add(connection);
+            ctx.channel().config().setAutoRead(false);
+            dispatchNext(ctx);
+        }
```
indigo423, line 285: One outstanding dispatch per channel with `autoRead(false)` in between means every message pays a full round trip through the dispatch pool, the sink queue and back onto the event loop before the next byte is read.

When I tested with rsyslog and syslog-ng both default to one connection per target, so an entire site can funnel through that single-outstanding path where UDP had no such constraint. I tried to measure it and threw the numbers away: the persisted counts turned out to be lag rather than loss (a controlled 5000-message burst reached 4939 and the backlog kept draining ~15 min after traffic stopped), the socket-accepted counts measured the kernel buffer rather than this handler, and Eventd/Postgres saturated well before `SyslogDispatchHandler` did. 

`SyslogTcpListenerDispatchIT` is already most of the harness needed. An idea would bet use this listener plus a no-op dispatch and turning this into a benchmark would give us a real per-connection ceiling. Not a blocker to me just something I've noticed and while we are here having a number would be nice.

features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/SyslogTcpListener.java
```diff
@@ -0,0 +1,402 @@
...
+                    .sync();
+
+            LOG.info("Listening for syslog messages over TCP on {}", describeAddress());
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            LOG.warn("Interrupted while binding the syslog TCP listener on {}", describeAddress(), e);
+            stop();
+        } catch (Throwable e) {
+            // Throwable, so an Error cannot leave nothing in the log to say why the port is dead.
+            LOG.error("Failed to bind the syslog TCP listener on {}", describeAddress(), e);
```
indigo423, line 218: Seems to me the same here as above in 157 - 160.  The gap is downstream: after either branch, and after the `catch (Throwable)` in `SinkDispatchingSyslogReceiver:81-83`, OpenNMS is running with syslog TCP silently absent and nothing outside `karaf.log` says so. `isStarted()` at :125 exists but nothing polls it. An event when a configured listener fails to start would turn "we stopped receiving syslog last Tuesday" into an alarm.

Hint untested and analysis only, it's more like a follow-up and not a blocker to me.

features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/SyslogTcpListener.java
```diff
@@ -0,0 +1,402 @@
...
+            m_framing = m_config.resolveFraming();
+            if (m_config.isTlsEnabled()) {
+                m_sslContext = SyslogTcpSslContextFactory.create(m_config);
+                LOG.info("TLS enabled for the syslog TCP listener on {}, client authentication is {}",
+                        describeAddress(), m_config.getTls().getClientAuth());
+            }
+        } catch (Throwable e) {
+            LOG.error("Not starting the syslog TCP listener on {}: {}", describeAddress(), e.getMessage(), e);
+            return;
+        }
```
indigo423, line 165: A bad framing or certificate path must stop the listener rather than leave a port that accepts and drops, and catching `Throwable` so an `Error` still explains a dead port is a nice touch.

The gap is downstream: after either branch, and after the `catch (Throwable)` in `SinkDispatchingSyslogReceiver:81-83`, OpenNMS is running with syslog TCP silently absent and nothing outside `karaf.log` says so. `isStarted()` at :125 exists but nothing polls it. An event when a configured listener fails to start would turn "we stopped receiving syslog last Tuesday" into an alarm. Hint untested and analysis only, it's more like a follow-up and not a blocker to me.

features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/SyslogTcpListener.java
```diff
@@ -0,0 +1,402 @@
...
+
+    private void shutdownGracefully(final EventLoopGroup group, final String description) {
+        if (group == null) {
+            return;
+        }
+        if (!group.shutdownGracefully(0, SHUTDOWN_TIMEOUT_SECONDS, TimeUnit.SECONDS)
+                .awaitUninterruptibly(SHUTDOWN_TIMEOUT_SECONDS, TimeUnit.SECONDS)) {
+            LOG.warn("Syslog TCP {} group did not shut down within {} seconds", description, SHUTDOWN_TIMEOUT_SECONDS);
+        }
+    }
```
indigo423, line 430: Three sequential 15-second waits, all `awaitUninterruptibly`, plus an unbounded `m_channels.close()` wait, sit on the shared `stop()` path. The early return at :346-348 should mean a UDP-only install skips all of it, since none of the three fields is ever set. Can we double check if that is correct? If it isn't every UDP-only install pays up to 45 s extra shutdown for a feature it doesn't use.

indigo423 (commented):
Approach LGTM here a few topics I would take a closer look and this one is especially one I would investigate a bit deeper:

A bad TCP value takes UDP down with it on a Minion.** The Core is protected: `SinkDispatchingSyslogReceiver.run()` wraps TCP setup in `catch (Throwable)` so UDP starts regardless. The Minion builds `SyslogTcpConfig` as a Blueprint bean whose setters throw, and a throwing Blueprint bean fails the whole container which is the one also owns the UDP listener.

With `framing` set to a value outside the accepted set I got `ComponentDefinitionException` on `setFraming` and **both** listeners bound zero sockets. 
When restored it to `auto` it brought both back. This is the only asymmetry between the two ingestion paths. Seems like it fails in the unsafe direction, and the property placeholder reloads, so it can strike long after a healthy start.

features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/SyslogTcpListener.java
marshallmassengill: Now a WARN rate-limited to once a minute, naming max-connections, with the per-connection detail still at DEBUG.

features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/SyslogTcpListener.java
marshallmassengill: Planning on updating the comment and commit message for this one.

features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/SyslogTcpListener.java
marshallmassengill: Agreed.  This is a good idea.  I'll capture it as an enhancement NMS.

opennms-config/src/main/java/org/opennms/netmgt/config/SyslogdConfig.java
marshallmassengill: Will put this as a default returning an empty config.

features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/SyslogTcpListener.java
marshallmassengill: I take it you evaluated this one again.  I am going to update this a bit so this doesn't get flagged again later.  I think adding aUdpOnlyInstallBuildsNoTcpListener will do it.

features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/SyslogTcpFrameDecoder.java
marshallmassengill: Keeping the latch as is for the reason you said.  I think the counter will need AbstractMessageDispatcherFactory.getMetrics() out of core.ipc.sink.common... and then some other wiring.  I think it would be better to capture that in an enhancement along with the refused-connection counter.

features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/SyslogTcpFrameDecoder.java
marshallmassengill: Updating this to discard just the oversized message and keep going. Octet-counting still closes and I've validated that.

features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/SyslogTcpListener.java
marshallmassengill: Will add this to enhancement NMS as well.

features/events/syslog/src/main/java/org/opennms/netmgt/syslogd/SyslogTcpListener.java
marshallmassengill: Ok... I did a bunch of testing on this to understand what was happening with ordered/unordered.  Where I landed is that I think I've got this exactly backwards at the moment and I think what we want is for unordered to become the default (this gives it parity with UDP effectively) and then add a ordered="true" so any installs that want the stronger guarantee and can accept the tradeoff and have the option of it.  At least that's the next commit I'm working on because I think it makes sense.

marshallmassengill: See: https://github.com/OpenNMS/opennms/pull/8794/commits/da1ad43f0f7d79ac3317a5ed183ac9964122362b

PR #8797, opened 2026-08-18
NMS-20160 / NMS-20162 — Search field clear button, map search icons, first-class tooltips

synqotik:
# NMS-20160 / NMS-20162 — Search field clear button, map search icons, first-class tooltips

Two related seam-layer improvements plus three bugs found on the way.

## NMS-20160 — search fields

**`OnmsSearchInput`** is now a three-part composite: leading search icon, input, and a
trailing clear button that appears once there is something to clear.

- The clear affordance is a real `<button>`, not an `InputIcon`.
- The component now owns "input fills the field". PrimeVue anchors `IconField`'s icons to
  the *field's* edges and leaves filling it to the caller, so six call sites had each
  hand-patched the same two rules with a comment about the icon "floating far out in the
  container". All seven copies are gone, along with the `padding-right` overrides
  `IconField` supplies itself.

**Menubar search** drops its hand-rolled input, absolute search icon and results-toolbar
close button for the shared component. Click-outside dismissal (which the close button
existed to work around) is wired up properly via the vendored `useOutsideClick`. The shift-shift focus shortcut no longer
reaches into the search markup with a `querySelector` — `Search` exposes `focus()`.

**Map search** (`MapSearch.vue`) swaps its PrimeIcons font glyph for the seam's search
icon and gains a clear control. Clearing resets the selected chips, the map's node filter,
the pending debounce, and `mapStore.nodeSearchTerm`.  Made the map search clear control a real button.

## NMS-20162 — `tooltip` prop on `OnmsIconButton`

```vue
<OnmsIconButton :icon="Delete" tooltip="Delete" />
```

All 25 icon-button call sites use it (Adhoc toolbar, Configuration screens, DCB modals and
table, file-editor sidebar, SCV input, event-config upload tab). The directive stays for
every other host — plain elements, `OnmsIcon`, `OnmsChip` — and for positioning modifiers,
which have no prop equivalent (yet). When `tooltip` is set the native `title` attribute is
dropped so the browser tooltip doesn't duplicate the rich one; `title` still names the
button for assistive tech.

## Bugs fixed along the way

1. **Every tooltip in the app painted at the wrong z-index.** All `v-onms-tooltip` directive
   call sites move from ~1000 to 2100.

2. **Menubar search text and icon were invisible in light mode**, fixed.

3. **`Menubar`'s outside-click handler had never run.** It was called with
   `outsideClick.value` (undefined at setup — the composable reads `.value` when the click
   arrives) and its returned `active` ref was never set.

4. **Side menu overlap when on Topology screen.** The Side menu was overlapping the Vaadin-based Topology screen, now fixed.

5. **Search box issues when on Topology screen.** Updates to fix the main top Menubar's Search box when user is on the Topology screen.

#### OnmsSearchInput with text and close icon

onmssearchinput-with-text

#### OnmsSearchInput with no text and no close icon (just placeholder text)

onmssearchinput-no-text

#### Main search box with results

main-search-box-with-results

#### GeoMap search

geomap-search

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20160
* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20162


marshallmassengill (changes requested):
A couple of blockers:

Blockers

  1. The z-index fix only covers tooltips whose value is non-empty at mount, so the headline claim overstates it. PrimeVue's beforeMount does if (!options.value) return before the $_ptooltipZIndex assignment (tooltip/index.mjs:33,:51), and updated never sets that property. Probed with the wrapper installed: value '' at mount then set to 'arrived later' gives value: "arrived later", zIndex: undefined; value present at mount gives 2100. So "All v-onms-tooltip directive call sites move from ~1000 to 2100" holds for literals, not for values that arrive with data. Three workarounds for this one root cause remain in the tree: Menubar.vue v-if="dateTimeLabel", NodeTooltipCell.vue v-if="text", and the new OnmsIconButton :key. The Menubar comment now says the seam "handles that centrally now", but that v-if is still load-bearing.

  2. The :key remount in OnmsIconButton drops keyboard focus to <body>. Probed: focus the button, set tooltip from undefined to a string, and the DOM node is replaced with document.activeElement === document.body. It is also avoidable. ZIndex.set('tooltip', tooltipElement, el.$_ptooltipZIndex) reads the property at show time (tooltip/index.mjs:222), so re-applying the resolved z-index in the wrapper's updated covers late values without unmounting anything. Verified against a <script setup>-style host: same DOM node, focus kept, zIndex: 2100. That one change would also retire both v-if workarounds above.

synqotik:
> A couple of blockers:
> 
> Blockers
> 
>     1. The z-index fix only covers tooltips whose value is non-empty at mount,
>     2. The :key remount in OnmsIconButton drops keyboard focus to . Probed

@marshallmassengill  Fixed.

Also fixed side menu overlap when expanded over Topology. And fixed an issue in `OnmsSearchInput` where it wasn't displaying correctly on the Topology page.

marshallmassengill (approved):
Not blockers but worth noting:

- OnmsSearchInput's always-rendered clear <span> and its comment are superseded by the later padding commit — the span is position: absolute, and padding-inline-end: var(--onms-search-input-inset) !important now reserves the space unconditionally, so the .p-inputtext:not(:last-child) mechanism the comment cites (real, in @primeuix/styles/iconfield) no longer decides anything. The identical comment on MapSearch's __clear-slot is still load-bearing, which makes the stale one easy to mistake for live reasoning.
 - primevue-overrides.scss:118 is now dead — .p-inputicon { font-size: 1.5rem; margin-top: -0.75rem } duplicates exactly what OnmsSearchInput's scoped style declares, and OnmsSearchInput.vue is the only IconField/InputIcon consumer in the tree (every other hit is ScvInputIcon, an OnmsIconButton).
 - tooltip now feeds OnmsIcon's aria-label, so the ~20 call sites that already set aria-label on the button gain a named role="img" child inside an already-named button, where they previously got aria-hidden="true". Harmless for naming, but it puts a labelled image into the a11y tree at every one of them.
 - Menu/Search.vue:347 reintroduces :deep(.p-inputtext) { width: 100% } — the seam already sets it, so one copy comes back as seven go.

PR #8802, opened 2026-08-19
NMS-20246: Fail the build on dangling static image references in the webapp

synqotik:
# NMS-20246: Fail the build on dangling static image references in the webapp

## Summary

Adds a single unit test, `opennms-webapp/src/test/java/org/opennms/web/StaticAssetReferenceTest.java`, that scans the webapp sources for statically written `images/...` references and fails the build if any of them does not resolve to a file that is actually present.

No runtime code changes, no new dependencies, no configuration changes. Runs in about 0.13s.

## Why

A missing image referenced from a JSP is invisible to the build. There is no compile error, no packaging error and no server-side error — the page renders with a broken image or a blank background, and the only trace is a 404 in the browser. Nothing in CI looks at it.

That is what made the `images/` cleanup in NMS-20091 risky. It removed 61 unreferenced files on the strength of a reference audit, which was the right call and correctly scoped for this repository. But where such an audit is wrong, the mistake is silent: one of the removed wallpapers was still referenced by customized JSPs in a downstream distribution built from this code, and the breakage only surfaced when someone happened to open a login page.

The audit itself was not the problem — the absence of any mechanical check was. This test supplies one, so the next `images/` cleanup gets a build failure instead of a bug report.

## What the test does

Walks `src/main/webapp` and reads every `.jsp`, `.jspf`, `.tag`, `.tagx`, `.ftl`, `.html`, `.htm`, `.css` and `.js` file. Matching happens in two steps, which is what keeps it from guessing:

1. **Match the whole token.** A regex picks up complete URL or path tokens ending in an image extension. Its character class excludes quotes, whitespace and CSS delimiters, so a token is naturally bounded by the attribute quoting or `url(...)` parentheses around it.
2. **Classify the token by its path segments.** A token counts as a local webapp image only when `images` is an *entire* path segment followed by a filename, the token carries no URL scheme, and it is not protocol-relative.

That covers the reference forms the webapp actually uses — bare `images/foo.png`, server-absolute `/opennms/images/foo.png`, relative `../images/foo.png`, and the JSTL `<c:url value="/images/foo.png"/>` form — while excluding anything that cannot be checked against the local filesystem or is simply a different directory.

Each surviving reference must resolve to a regular file under `src/main/webapp`. References are additionally compared against the case the file has on disk, so a mismatch that works on a case-insensitive developer filesystem cannot reach a case-sensitive deployment as a 404.

On failure the assertion names the unresolved reference and every file that refers to it:

```
Image references in the webapp that do not resolve to a file under src/main/webapp.
Restore the file, or update the reference if the image is genuinely gone -- do not
leave the reference dangling, because nothing else in the build will catch it:
  images/wallpapers/background_dark.png -- no such file
      referenced by: account/selfService/passwordGate.jsp, login.jsp
```

## Review feedback addressed

The first version of the matcher anchored on the literal text `images/` with nothing to its left, which produced false positives on two shapes, both confirmed to fail the build on an otherwise unrelated commit:

```
<img src="https://cdn.example.com/images/x.png">
<img src="/opennms/custom-images/logo.png">
```

The first is served by another host and cannot be checked locally. The second is a different directory, and was reported as a missing `images/logo.png` — a file nothing had referenced.

The fix is the match-then-classify split described above, rather than a boundary assertion added to the regex. Splitting on `/` and requiring a whole segment also excludes the `myimages/` and `theme_images/` shapes, and turns the `assets/` exclusion into an ordinary segment comparison instead of a lookbehind, so both exclusions are now expressed the same way.

Each of these cases is pinned by a named unit test over the classifier, none of which touch the filesystem. Note that the defect only ever produced false *positives* — a failing build on an innocent commit, never a silently missed dangling reference — but it would have made the test a nuisance the first time someone added a CDN image, which is how a check like this ends up disabled.

## Scope and deliberate omissions

**It never asserts that an image *is* referenced.** Flagging unused files is exactly the reasoning that produced the original incident, so it is out of scope by design. Dead files remain a judgment call.

**`assets/images/...` is excluded.** Those belong to `core/web-assets`, are served from a different root, and are already checked at build time — `webpack.config.js` uses `file-loader`, which fails the bundle on an unresolvable `url()`. The plain JSPs have no bundler, which is precisely why they need this test.

**Runtime-constructed paths are skipped by construction.** The pattern's character class excludes quotes and whitespace, so `"images/" + name + ".png"` does not match rather than matching incorrectly. Exactly one image-like reference in the webapp falls outside the pattern, `graph/graph.png` in `graph/adhoc4.jsp`, which is a servlet URL rather than a file.

**Guards against a vacuous pass.** The test asserts that the webapp root exists and that the scan found at least 10 references, because a test that silently finds nothing is worse than no test. The relative-path convention follows the existing `NavBarControllerTest`, which resolves `src/main/webapp` the same way.

Two limits worth stating: the test checks source, not the assembled WAR, so it will not catch a file that exists but is excluded from packaging; and it does not cover images referenced from SCSS or Vue in the newer UI. Both would be separate checks.

## Notes for reviewers

* **Test-only change, no UI impact**, so the branch does not carry `-smoke`. It adds no runtime code and cannot alter rendered output.
* No files under `$OPENNMS_HOME/etc/` are touched, as required for a `foundation-*` target.
* New code, and it is itself the test.
* The behavior is documented in the test's class-level Javadoc, including why the unused-file direction is intentionally not checked. No user-facing documentation change seemed warranted; happy to add some if a reviewer disagrees.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20246

### DIFF
--- a/opennms-webapp/src/test/java/org/opennms/web/StaticAssetReferenceTest.java
+++ b/opennms-webapp/src/test/java/org/opennms/web/StaticAssetReferenceTest.java
@@ -1,0 +1,189 @@
+/*
+ * Licensed to The OpenNMS Group, Inc (TOG) under one or more
+ * contributor license agreements.  See the LICENSE.md file
+ * distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * TOG licenses this file to You under the GNU Affero General
+ * Public License Version 3 (the "License") or (at your option)
+ * any later version.  You may not use this file except in
+ * compliance with the License.  You may obtain a copy of the
+ * License at:
+ *
+ *      https://www.gnu.org/licenses/agpl-3.0.txt
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied.  See the License for the specific
+ * language governing permissions and limitations under the
+ * License.
+ */
+package org.opennms.web;
+
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
+import java.util.TreeSet;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
+
+import org.junit.Test;
+
+/**
+ * Verifies that every statically written <code>images/...</code> reference in the webapp
+ * resolves to a file that is actually present under <code>src/main/webapp</code>.
+ *
+ * A dangling image reference in a JSP produces no compile error, no build failure and no
+ * server-side error -- the page renders with a broken image or a blank background, and the
+ * only symptom is a 404 in the browser. That makes it easy to delete an image that is still
+ * in use: NMS-20091 removed 61 unreferenced files from <code>images/</code>, and the one
+ * mistake it made (a wallpaper whose only consumers were downstream PoweredBy JSPs) went
+ * unnoticed until someone looked at a login page. This test turns that class of mistake
+ * into a build failure.
+ *
+ * Note what this deliberately does <em>not</em> do: it never asserts that an image is
+ * referenced. Unused files are a cleanup question for a human, not a test failure.
+ */
+public class StaticAssetReferenceTest {
+
+    /**
+     * Surefire runs with the module directory as the working directory, so the webapp
+     * sources are addressable as a relative path. NavBarControllerTest relies on the same
+     * convention via {@code new MockServletContext("file:src/main/webapp")}.
+     */
+    private static final Path WEBAPP_ROOT = Path.of("src", "main", "webapp");
+
+    /** File types that can carry a static reference to an image. */
+    private static final Set<String> SCANNED_SUFFIXES =
+            Set.of(".jsp", ".jspf", ".tag", ".tagx", ".ftl", ".html", ".htm", ".css", ".js");
+
+    /**
+     * Matches an image path anchored on its {@code images/} segment, which covers every form
+     * these references take in practice: bare ({@code images/foo.png}), server-absolute
+     * ({@code /opennms/images/foo.png}) and JSTL-wrapped ({@code <c:url value="/images/foo.png"/>}).
+     *
+     * <p>The character class excludes quotes and whitespace, so a path assembled at runtime
+     * ({@code "images/" + name + ".png"}) simply does not match rather than matching wrongly.
+     *
+     * <p>The lookbehind skips {@code assets/images/...}, which lives in the core/web-assets
+     * module, is served from a different root, and is already checked at build time -- webpack's
+     * file-loader fails the bundle on an unresolvable {@code url()}.
+     */
+    private static final Pattern IMAGE_REFERENCE = Pattern.compile(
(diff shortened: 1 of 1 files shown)


marshallmassengill (changes requested):
One blocker:

No boundary or origin check before images/ — verified:  and  both fail the build, the latter reported as the nonexistent images/logo.png. Any future external image URL or any directory whose name ends in images breaks an unrelated commit. Needs (?<![A-Za-z0-9_-]) plus a scheme guard.

synqotik:
> One blocker:
> 
> No boundary or origin check before images/

@marshallmassengill Good catch, should be fixed in latest commit. See the "Review feedback addressed" section in the PR description for details.

marshallmassengill (approved):
I'm good with this but there are some additional findings (none are real blockers):

Cross-context absolute URLs are still classified as local — verified end-to-end:  fails the build as a missing images/logo.png. "No scheme" is not the same as "this webapp"; the segments before images need to be empty, ./.., or the context path.
  - The assets guard only inspects the immediately preceding segment — /opennms/assets/theme/images/x.png is accepted as local, and core/web-assets/src/main/assets/js/apps/geomap-js/images is a real nested images directory under that tree, so the shape is not hypothetical.
  - No trailing boundary after the extension — proven live in the current tree: ${trendDefinition.icon} in WEB-INF/jsp/trend/trend.jsp:36 yields the token trendDefinition.ico; by the same truncation images/logo.svgz becomes images/logo.svg and fails as missing.
  - Dot segments aren't normalized before the case comparison — verified: the valid images/wallpapers/../wallpapers/background_dark.png fails with case does not match the file on disk, which is  images/wallpapers/background_dark.png, a message about case for a reference whose case is correct. Path.of(reference).normalize() in describeIfUnresolvable closes it.
  - /images/foo.png and ../images/foo.png are now asserted as valid forms, which pins the false negative rather than fixing it: a raw src="/images/x.png" 404s under the /opennms context and a ../images/ written at the wrong depth 404s too, and both pass.
  - assertNull(localWebappImageReference("images/")) is unreachable — IMAGE_TOKEN requires an extension, so the scan can never produce that token.
  - recognizesTheReferenceFormsUsedInTheWebapp asserts four forms the webapp doesn't use — /opennms/images/…, /images/…, ../images/…, Images/…; the tree contains only the bare form and ${baseHref}images/….

- Coverage is 19 of the 54 files under webapp/images/; the other 35 are referenced only from vaadin-theme.scss, dashboard-theme/styles.css, DefaultRrdGraphService.java:67, opennms.properties:230, notifications/index.js:56, and eXtremeComponents imagePath.
  - The assets/images bullet's webpack rationale still doesn't cover the actual gap: those SCSS files reference /opennms/images/..., the webapp root, which webpack leaves alone.
  - The motivating file, images/wallpapers/IMG_9269-X2.jpg, had no in-repo reference, so the test would still have passed at the moment of the NMS-20091 regression.
  - The "exactly one image-like reference falls outside the pattern" bullet is now measurably wrong — the scan rejects 8 tokens, and 5 of them (favicon.ico, favicon-16x16.png, favicon-32x32.png, apple-touch-icon.png, safari-pinned-tab.svg) are real webapp-root files referenced from includes/bootstrap.jsp and unchecked; the classifier already holds these tokens and could check them.
  - Commented-out references still fail the build; actualCaseOf still reports a symlink's target as a case mismatch; the case branch still cannot execute on Linux CI; readString(UTF_8) still throws on a non-UTF-8 byte; the floor of 10 still tolerates losing half of 19.

synqotik:
> I'm good with this but there are some additional findings (none are real blockers):


Thanks. We can add some of these in the future if they are needed. This at least sets a precedent of checking. As noted in some cases, like in `core/web-assets`, many of those will be caught by the bundler, so we are purposefully not testing them here.

PR #8803, opened 2026-08-19
NMS-19878: Topology Preview UI

marshallmassengill:
This is the much worked on Topology Preview UI for H36/M2026.  Almost all of this is UI work.  I think this likely still needs docs but those can wait until we can get screenshots.  There is a high chance that the footer is going to cause issues but I've validated it as much as I can in different browsers and resolutions.

I've made some choices for rendering around number of nodes in the view that I think make sense so this is scalable in the long term.  If you test with >3000 connected nodes then you'll see them.... only time will tell if those are good choices.

I would love feedback.  If you need test data then let me know.

Assisted by Anthropic Claude Opus 4.8, 5, and Fable 5.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19878


dino2gnt (dismissed):
I don't care if it eats babies and kills your pets, let's get it out there for feedback and fixes until it's good enough

mershad-manesh:
ran a quick test and found a couple of minor issues,

- Double-clicking an empty area of the viewport to create a label places it offset from the cursor instead of directly under it.
- Creating a new view, using "Save as", or renaming a view opens a browser `prompt()` dialog. This should be replaced with a proper modal/dialog component.
- Drawing a box (using "Draw Box") in an empty area of the viewport can cause it to snap outside the viewport bounds.

marshallmassengill:
> ran a quick test and found a couple of minor issues,
> 
>     * Double-clicking an empty area of the viewport to create a label places it offset from the cursor instead of directly under it.
> 
>     * Creating a new view, using "Save as", or renaming a view opens a browser `prompt()` dialog. This should be replaced with a proper modal/dialog component.
> 
>     * Drawing a box (using "Draw Box") in an empty area of the viewport can cause it to snap outside the viewport bounds.

Thank you!  What browser?

mershad-manesh:
> > ran a quick test and found a couple of minor issues,
> > ```
> > * Double-clicking an empty area of the viewport to create a label places it offset from the cursor instead of directly under it.
> > 
> > * Creating a new view, using "Save as", or renaming a view opens a browser `prompt()` dialog. This should be replaced with a proper modal/dialog component.
> > 
> > * Drawing a box (using "Draw Box") in an empty area of the viewport can cause it to snap outside the viewport bounds.
> > ```
> 
> Thank you! What browser?

Edge

synqotik:
@marshallmassengill Might make sense to get the latest UI changes in `foundation-2026`.

(removed my comment re targeting `develop` - I think because this is a preview and not yet a replacement it's ok for `foundation-2026`)

synqotik (dismissed):
Seems reasonable to me. Since it's a preview we can always fix/iterate as needed.

marshallmassengill:
Updates in for you to review @mershad-manesh

synqotik (changes requested):
Overall view: I feel this needs some UX iterations before H37 release, but OK for H36 as long as it's understood it's a Preview.

One note, going to Node Details is going to the Vue Node Details page which as of H36 is *not* complete. I'd suggest going to the JSP Node Details page for now.

I will open a separate ticket to go over UX improvements, but I can approve after the following are fixed.

I'd say normally this many `onms-ui` changes should go under a separate PR, but they mostly seem ok. I'd review and fix the following before approval/merge.

---

1. OnmsSearchInput's new fluid prop is a no-op — and the rationale documented for it is wrong

OnmsSearchInput.vue#L11-L13, fluid=true does exactly one thing: it adds p-inputtext-fluid to the <input>, whose only rule is width: 100% — declared inside @layer primevue.
But this component's own untouched scoped style block already applies, unconditionally:

OnmsSearchInput.vue#L136-L142 — .p-iconfield 
.p-inputtext { width: 100% }, unlayered and scoped, which per this file's own comment at L116-L118
("any unlayered author rule beats a layer regardless of specificity") outranks the layered one.
We know that selector reaches the <input> because the !important paddings on the adjacent lines are what fixed NMS-20160.

And the container premise is also already false: .p-iconfield is display: block; width: 100% at L131-L132.

So README L152-154 — "the input inside
keeps its intrinsic width unless fluid is set, so a field that should fill its column needs both" — describes behavior that doesn't exist. The input was already
filling the container before this prop.

Because both rules declare the same width: 100%, nothing renders wrong; the defect is a dead prop on the seam's public surface plus a misleading rule in the README
that will send the next person down the wrong path. TopologyPalette.vue passes fluid alongside class="palette-search" in exactly the pattern the README prescribes,
so whatever sizing problem motivated this was solved by something else. The new test only asserts the class is present, never a width, so it passes either way.
Suggest either dropping the prop and the README bullet, or keeping the prop and correcting the bullet to say the input is already fluid in this composite.

Reviewer's Preference: drop the prop and the README bullet.

2. OnmsColorPicker normalizes the model value but not the swatches it compares against

normalized canonicalizes to lowercase #rrggbb (L127-L139, but the three selection tests compare it against raw swatches entries:

OnmsColorPicker.vue#L27-L33 (color === normalized, twice) and L138 (props.swatches.includes(normalized.value)).

With :swatches="['#FF0000']" and a value of #FF0000, the matching swatch never gets is-selected or aria-pressed="true", and the footer reads "(custom)" for a color
that is literally in the palette. Nothing in the README constrains swatches to lowercase.

Relatedly, pick() emits the raw swatch string with no toCssHex (L140-L144), unlike
onSpectrum right below it, so the swatch path can emit a value that violates the component's stated "#rrggbb in both directions" contract.

One-line fix: compare and emit through toCssHex, e.g. a normalizedSwatches computed plus emit('update:modelValue', toCssHex(color)). Worth noting the test named
"marks the swatch matching the current value, case-insensitively" only exercises the model-value side (uppercase modelValue, lowercase swatch), so it passes while
the swatch side is unhandled.

3. The default palette doesn't contain the topology default it claims to

OnmsColorPicker.vue#L73-L85: "The neutrals row
carries the shape stroke/fill and label defaults the topology map already ships, so a stock view reads as on-palette," echoed at README L174-176.

#64748b, #cbd5e1, #000000 and #1d2939 are all present. #9aa7b8 — the link colour in TopologyCanvas.vue and the fallback for the "Link label color" field in the
style panel — is not; the palette has the adjacent #94a3b8 instead. So that one field opens showing "(custom)" with nothing highlighted on a stock view.
OnmsColorPicker.test.ts uses #9aa7b8 as its example of an off-palette value, which is a good tell. Either add #9aa7b8 to the neutrals row or change the topology
default to #94a3b8.

synqotik (approved):
LGTM! We can iterate on this after the merge.

PR #8807, opened 2026-08-20
NMS-19841: Move java.io.tmpdir and activemq.data out of Karaf's data directory

mershad-manesh:
### All Contributors

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19841


cgorantla (commented):
We need a way to migrate existing activemq store.
Might be worth pointing this in docs and provide an option to keep the old path.

Looks like drools is using `java.io.tmpdir `  and might need migration for existing kei session state files.

dino2gnt:
> We need a way to migrate existing activemq store.

Do we, though? We run fix-karaf-setup with every invocation of the installer, which blows it away. it's already lossy...

cgorantla:
> > We need a way to migrate existing activemq store.
> 
> Do we, though? We run fix-karaf-setup with every invocation of the installer, which blows it away. it's already lossy...

corner case. Only needed if there is huge backlog and user don't want any lose any traps/flows.
This migration should help them in future. So I'm good with moving this directory out of karaf data.  Just a note in release notes is sufficient.

mershad-manesh:
@dino2gnt @cgorantla how about something like this ? https://github.com/OpenNMS/opennms/pull/8807/changes/6ee35ef832bd38216537229b5b7ba37cb8708d87

cgorantla:
> @dino2gnt @cgorantla how about something like this ? [6ee35ef](https://github.com/OpenNMS/opennms/commit/6ee35ef832bd38216537229b5b7ba37cb8708d87)

Probably not needed. As @dino2gnt said, we were already deleting everything under data. So we may not require this change.

docs/modules/releasenotes/pages/whatsnew.adoc
```diff
@@ -91,6 +91,23 @@ After upgrade, manage event configurations through the *Administration -> Manage
...
+Any ActiveMQ store contents and persisted Drools correlation-engine or northbounder session state under the old
+`data/tmp` location are *not* migrated to the new paths -- they are cleared during the upgrade, the same as with any
+other upgrade that clears the Karaf cache. The broker and any persisted rule sessions simply start fresh at the new
+locations afterward. Going forward, however, routine restarts (including container restarts) no longer clear this
+data, since it now lives outside `$\{OPENNMS_HOME}/data`.
+
+To keep the previous paths instead, set `-Djava.io.tmpdir=$\{OPENNMS_HOME}/data/tmp` and
+`-Dactivemq.data=$\{OPENNMS_HOME}/data/tmp/activemq` in `ADDITIONAL_MANAGER_OPTIONS`
+(see xref:operation:deep-dive/admin/configuration/startup.adoc[Startup Configuration]), or in `JAVA_OPTS` for
+container deployments.
```
cgorantla: Ideally this should be couple of lines and limited to one paragraph.

mershad-manesh: Good point, I have reduced it

cgorantla (approved):
LGTM!

PR #8808, opened 2026-08-20
NMS-20182: Various UI fixes after the PrimeVue / OnmsUI migration

synqotik:
# NMS-20182: Various UI fixes after the PrimeVue / OnmsUI migration

A batch of independent UI fixes left over from the move to PrimeVue and the `onms-ui` seam
layer. Each is small and self-contained; they are grouped here to avoid seven near-empty PRs.
Every change was verified against a running instance rather than by inspection alone.

Summarized most of the changes, but left some detail in here for future reference (e.g. re how footers were fixed).

---

## 1. Alarm Details Sticky/Journal memo textareas unreadable in dark mode

**Fix.** Added `textarea` to the existing rule in `opennms-styles.scss` that already pins native
selects, checkboxes and radios to `color-scheme: light`. Legacy textareas are now dark-on-light in
both themes. Only the menu bundle loads this stylesheet, so the SPA's PrimeVue `Textarea` is
unaffected.

---

## 2. `a:visited` overriding button label colors

**Problem.** On the Alarm and Event list pages, menu buttons rendered blue-on-grey instead of
white once their URL had been visited — "View all alarms", "Short/Long Listing", "View all events"
and the `Next` pagination button. Items rendered as anchors *without* an `href` stayed white.

**Cause.** `a:visited` includes a pseudo-class, so it outranks Bootstrap's single-class
`.btn-secondary`.

**Fix.** Removed the rule from both places it was declared, and dropped some previous workarounds.

---

## 3. Drawers and modal dialogs let the page scroll behind them

**Problem.** Opening a drawer (e.g. Customize Columns on the Node List) and scrolling scrolled the
node table behind it — the page's scrollbar sat at the viewport edge and read as the drawer's. The
Advanced Filters drawer was worse: it showed two adjacent scrollbars, only one of which was its own.

**Fix.** Three parts:

- Turned on PrimeVue's `blockScroll` in the seam wrappers.
- `blockScroll` alone was not enough: `html { overflow-x: hidden }` makes `<html>` the SPA's scroll
  container, so the body class PrimeVue toggles never reaches the element that actually scrolls. The
  lock is now applied to `<html>`; added `scrollbar-gutter: stable` so hiding the page scrollbar does not shift the fixed masthead and rail.
- Removed the two node drawers' inner `height: 100%; overflow: auto`, so PrimeVue's
  `.p-drawer-content` is each drawer's single scroll container.

Trade-off worth calling out: `scrollbar-gutter: stable` reserves the gutter permanently, so every SPA
page gives up ~15px of width whether it scrolls or not. The alternative is the fixed masthead and
rail jumping 15px every time an overlay opens.

---

## 4. SNMP "Add" button not aligned with the inputs beside it

**Problem.** On the SNMP configuration details panel, the Add button's top edge sat a label-height
above the tops of the IP address inputs in the same row.

**Fix.** Used `FormField` and gave it a `reserveLabelSpace` prop

---

## 5. JSP footer stopped at the end of content instead of the bottom of the window

**Problem.** On JSP pages the footer sat wherever the content ended — hundreds of pixels above the
bottom of the window on short pages (157px into a 969px viewport on the node list). The SPA already
behaves correctly: footer at the window bottom when content is short, after the content when it
scrolls.

**Fix.** `#content` becomes a flex column with `min-height: 100vh`, the same shape the SPA gets from
its layout grid's `1fr` content row. Because page markup must not become flex items — that stretches
inline-level children and collapses ones sized by auto margins — `bootstrap.jsp` now wraps everything
but the footer in a `#content-body` div. `#content` therefore has exactly two flex items, all page
markup lays out inside a plain block exactly as before, `#content-body` takes the slack and the
footer follows it.

The `custom-footer` includes an install can contribute are emitted after the footer, so they get a
plain `#custom-footer` block for the same reason, and the two footer-less pages (`login.jsp`,
`passwordGate.jsp`) close the wrapper themselves since they never include the footer.

Also dropped the SPA's default `body` margin, which sat outside `.app-layout`'s `min-height: 100vh`
and made every page overflow the viewport by ~16px. Both stacks now put the footer exactly at the
bottom edge.

Vaadin pages emit no `#content` and are deliberately untouched.

---

## 6. SPA footer's spacer div caused scrollbars with nothing to scroll

**Problem.** Pages whose content wanted exactly the height the shell leaves them — SNMP Config,
OpenAPI, Resource Graphs — grew a scrollbar for 16px of nothing.

**Cause.** A `<div style="height: 1rem">` inside the layout's footer row made that row 57px tall
instead of 41px, taking 16px from the `1fr` content row.

**Fix.** Removed the spacer. The footer looks the same without it — it keeps its own padding and top
border — and those pages now fit exactly.

`ui/src/components/Layout/Footer.vue`

---

## 7. Map, Logs and SCV pages clipped the footer

**Problem.** These three pages overflowed the viewport and pushed the footer off the bottom of the
window — most visibly on the map, where it was cut in half.

**Cause.** Each sizes its tall element with a hand-picked `calc(100vh - x)`, and each `x` was too
small for the chrome actually above and below it.

**Fix.** Derive the subtraction instead of guessing it.

---

## 8. Searchable Vue pages (unrelated, folded in)

Added Vue Resource Graphs, Geographical Map and Topological Map to `search-actions.xml` so they turn
up in search.

---

## Review round

Four items from review, fixed in `NMS-20182: Address PR review on the footer and scroll-lock
changes`:

- **`custom-footer` includes were outside the wrapper.** They are emitted after the footer, so an
  install's markup landed as a direct flex item of `#content` — the exact case `#content-body`
  exists to prevent. Now wrapped in a
  `#custom-footer` block, emitted only when the directory exists. Wrapping rather than moving the
  `#content-body` close, because these includes belong *after* the footer.
- **The scroll lock matched `.p-overlay-mask` too broadly.** DataTable, Tree, TreeTable, Image,
  Galleria and SpeedDial all hard-code that class on their masks, while Dialog and Drawer add it only
  when modal — so the day anyone wires up `OnmsTable`'s loading prop, every table fetch would have
  frozen page scroll. Narrowed to `.p-dialog-mask.p-overlay-mask` / `.p-drawer-mask.p-overlay-mask`,
  which keeps the nesting fix and cannot over-match. On the leave transition: PrimeVue's own
  `disableDocumentSettings()` also runs in `onAfterLeave`, so both mechanisms expire together and
  this adds no lag.
- **The 41px footer band was a literal in four files**, derived from a `Footer.vue` height this PR
  changes. Now a real `--onms-footer-height` token defined in `Footer.vue` next to the element it
  describes, mirroring how `Menubar.vue` owns `--onms-header-height`; `.footer` takes its
  `min-height` from the token so the two cannot drift. Note it is a *min*: at very narrow widths the
  copyright line wraps and the band exceeds the token, the one case those page calculations still
  under-count.
- **`login.jsp` and `passwordGate.jsp` left the divs unclosed** — they include `bootstrap.jsp` but
  deliberately not `bootstrap-footer.jsp`. Both now close `#content-body` and `#content` themselves.
  They stay footer-less by design.

Three items reviewed and consciously left alone:

- The repeated `51px` / `30px` / `4px` chrome literals. Same drift exposure as the footer band, but
  breadcrumbs are slated for removal in a near-future release.
- `search-actions.xml`: "Topology Map" does point at the legacy Vaadin page, so the commit message's
  "Vue pages" framing is loose, and the `(Preview)` suffix is inconsistent with "Geographical Map".
  Cosmetic; both entries work as intended.
- `scrollbar-gutter: stable`, accepted knowingly — see the trade-off note above.

---

## Known follow-ups (not in this PR)

- Deployed SPA and menu assets have unhashed filenames (`assets/index.js`), so browsers serve stale
  bundles after a deploy. Worth a separate issue — it also makes local verification error-prone.
- Form rows align labels and controls per-field, so a label that wraps to two lines pushes its own
  input down and out of line with its neighbours. Fixing that properly means putting label/control/
  hint on shared grid tracks (CSS subgrid).
- Map, Logs and SCV still use pixel arithmetic for their heights rather than the flex-fill pattern
  in `AdhocChart.vue`. Converting them would need the app shell to hand its height down, which
  touches every page.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20182

### DIFF
--- a/core/web-assets/src/main/assets/style/opennms-theme.scss
+++ b/core/web-assets/src/main/assets/style/opennms-theme.scss
@@ -370,6 +370,32 @@ iframe.vaadin-fullscreen {
   padding-top: var(--onms-legacy-content-offset);
   padding-left: var(--onms-legacy-content-offset);
 
+  // Stack the page in a column at least as tall as the window, so that #footer
+  // ends up at the bottom of the window on a short page instead of stopping
+  // wherever the content happens to end — the same result the Vue SPA gets from
+  // its layout grid's 1fr content row. On a page taller than the window there is
+  // no slack to take and the footer simply follows the content, as before.
+  // `box-sizing: border-box` is global, so 100vh already covers the padding-top
+  // above, and body carries no margin or padding of its own: the column starts at
+  // the top of the viewport and ends exactly at its bottom.
+  //
+  // bootstrap.jsp wraps everything but the footer in #content-body to keep this
+  // container down to two flex items. Page markup must not become flex items
+  // itself: that stretches inline-level children (a .btn-group grew to the full
+  // content width on the outage list), collapses elements whose width comes from
+  // auto margins (<hr> shrank to nothing on the reports page), and stops margin
+  // collapsing. Inside the plain block that is #content-body, all of it lays out
+  // exactly as it did before.
+  display: flex;
+  flex-direction: column;
+  min-height: 100vh;
+
+  // Takes every bit of leftover height (and never shrinks below its content), so
+  // the footer after it is left sitting at the bottom of the column.
+  > #content-body {
+    flex: 1 0 auto;
+  }
+
   // Revert padding as growl added this, which was required in bootstrap 3 but is
   // no longer required, so without this the padding to the right is too much
   .growl-container > .growl-item {
--- a/opennms-base-assembly/src/main/filtered/etc/search-actions.xml
+++ b/opennms-base-assembly/src/main/filtered/etc/search-actions.xml
@@ -209,6 +209,10 @@
             <role>ROLE_ADMIN</role>
         </roles>
     </action>
+    <action>
+        <label>Resource Graphs (Preview)</label>
+        <url>ui/index.html#/resource-graphs</url>
+    </action>
 
     <!-- Distributed Monitoring Card -->
     <action>
@@ -303,5 +307,15 @@
     <action>
         <label>Show Nodes without Flows</label>
         <url>ui/index.html#/nodes?flows=false</url>
+    </action>
+
+    <!-- Maps -->
+    <action>
+        <label>Geographical Map</label>
+        <url>ui/index.html#/map</url>
+    </action>
+    <action>
+        <label>Topology Map</label>
+        <url>topology</url>
     </action>
 </actions>
--- a/opennms-webapp/src/main/webapp/WEB-INF/jsp/alarm/list.jsp
+++ b/opennms-webapp/src/main/webapp/WEB-INF/jsp/alarm/list.jsp
@@ -319,8 +319,8 @@
     <a class="btn btn-secondary" href="<%=this.makeLink(callback, parms, "long", favorite)%>" title="Detailed List of Alarms">Long Listing</a>
       </c:otherwise>
     </c:choose>
-    <a class="btn btn-secondary text-white" onclick="$('#severityLegendModal').modal()">Severity Legend</a>
-    <a class="btn btn-secondary text-white" onclick="$('#helpModal').modal()">Help</a>
+    <a class="btn btn-secondary" onclick="$('#severityLegendModal').modal()">Severity Legend</a>
+    <a class="btn btn-secondary" onclick="$('#helpModal').modal()">Help</a>
 
     <% if (req.isUserInRole( Authentication.ROLE_ADMIN ) || !req.isUserInRole( Authentication.ROLE_READONLY)) { %>
       <% if ( alarmCount > 0 ) { %>
--- a/opennms-webapp/src/main/webapp/account/selfService/passwordGate.jsp
+++ b/opennms-webapp/src/main/webapp/account/selfService/passwordGate.jsp
@@ -236,3 +236,16 @@
         </div>
     </div>
 </div>
+
+<%--
+  These pages include bootstrap.jsp but not bootstrap-footer.jsp — they are
+  deliberately footer-less — so nothing else closes the two <div> tags that
+  include opens. Close them here: #content is a flex column at least as tall as
+  the window (see #content in opennms-theme.scss), and leaving the wrapper open
+  would put this page's markup and any later content in the wrong box. The
+  matching tags are in bootstrap.jsp, so they are emitted from JSP fragments to
+  keep the Eclipse HTML validator quiet, as that file does. <body> and <html>
+  stay unclosed here exactly as before.
+--%>
+<%= "</div>" %><!-- id="content-body" -->
+<%= "</div>" %><!-- id="content" class="container-fluid" -->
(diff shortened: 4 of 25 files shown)


marshallmassengill (dismissed):
Good but a few findings (not blockers):

- bootstrap-footer.jsp emits the custom-footer includes after the #content-body close, so any visible markup an install contributes there becomes a direct flex item of #content — exactly the stretched-inline-children / collapsed-auto-margin failure the PR's own comment documents as the reason for the wrapper. Closing #content-body after that loop, or wrapping it, keeps the container at two items.
 - html:has(.p-overlay-mask) ties the scroll lock to a class six PrimeVue components share rather than to modal overlays. DataTable, Tree and TreeTable put p-datatable-mask p-overlay-mask on their loading mask unconditionally (Image, Galleria and SpeedDial too), while Dialog and Drawer add p-overlay-mask only when modal. It's latent today only because OnmsTable deliberately doesn't expose loading — the seam README invites adding it, and the day someone does, every table fetch freezes page scroll. html:has(.p-dialog-mask.p-overlay-mask), html:has(.p-drawer-mask.p-overlay-mask) keeps the nesting fix and can't over-match. The mask also lives inside <Transition  name="p-overlay-mask">, so the lock outlasts the close by the transition duration.
 - The 41px footer band is hardcoded in four files (Logs/Editor.vue, Logs/Logs.vue, SCV/SCVForm.vue, containers/Map.vue) and derived from a Footer.vue height this same PR changes. --onms-header-height is a real global token (:root in an unscoped block in Menubar.vue), so each calc is half derived and half magic number; a sibling --onms-footer-height, consumed by Footer.vue itself, would stop those four pages drifting the next time the footer moves.
 - login.jsp and passwordGate.jsp include bootstrap.jsp without bootstrap-footer.jsp, so they now emit an unclosed #content-body on top of the already-unclosed #content. Browsers recover at </body>, but those two are also the only pages that pick up the new min-height: 100vh flex column with no footer for it to position.
 - The 51px breadcrumb row and the 30px/4px card paddings are repeated as literals across three files with the reasoning living only in comments. The PR body lists this as a known follow-up, so it's a scope call rather than an oversight, but it's the same drift exposure as the footer band.
 - search-actions.xml: "Topology Map" points at topology, the legacy Vaadin page, so it doesn't fit the commit's "Searchable Vue pages" framing; and "Resource Graphs (Preview)" carries a (Preview) suffix that "Geographical Map" — equally a Vue page — does not.
 - scrollbar-gutter: stable on html reserves the gutter permanently, so every SPA page gives up ~15px of width whether it scrolls or not. Reasonable in exchange for the fixed chrome not jumping, but worth saying out loud since it applies to every page, not just the ones with overlays.

synqotik:
> Good but a few findings (not blockers):
> 


@marshallmassengill fixed 4, left 3 alone, see reasons in the Summary under **Review round**.

marshallmassengill (dismissed):
LGTM!

marshallmassengill (approved):
Everything looks good.  This did come up as a finding but I'm not sure it's breaking anything... just worth noting if you end up making other changes:

login.jsp:246 / passwordGate.jsp:250 close the divs unguarded. bootstrap.jsp opens #content/#content-body only in the param.superQuiet != 'true' branch, and bootstrap-footer.jsp mirrors that guard on both closes; these two emit the closes unconditionally, so login.jsp?superQuiet=true produces two stray </div>.

PR #8809, opened 2026-08-21
NMS-20110: HA config sync observability, version guard, and gated status

cgorantla:
- skip sync while the two nodes run different releases
- record sync success/failure and surface it in ha-status and REST
- flag RESTART REQUIRED when sync replaces startup-only config
- make sync roots configurable; add deploy for plugins
- report a gated standby as running (exit 161) instead of stopped
- drop the unused --partner-rest-url shell option

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20110

### DIFF
--- a/core/daemon/src/main/java/org/opennms/netmgt/vmmgr/Invoker.java
+++ b/core/daemon/src/main/java/org/opennms/netmgt/vmmgr/Invoker.java
@@ -92,6 +92,11 @@ public class Invoker {
     private List<InvokerService> m_services;
 
     private final Path m_statusPath;
+    /** Written instead of a service list when the process is up but its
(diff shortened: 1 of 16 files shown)


marshallmassengill (changes requested):
I know this is still in draft but figured I would give feedback now.  Two things that I think are worth looking at as blockers:

- HaConfigSyncer.sync() fetches into any root the partner's manifest names, so #root lib writes into lib/ on the standby. Reject entries whose root is not in the local getSyncRoots(), as getSyncFile already does on the serving side.
- propagateDeletions() walks local roots but deletes against the partner's manifest. A standby with deploy configured and an active without it deletes every local plugin. Advertise the root set in the manifest and only propagate deletions for roots it contains.

PR #8810, opened 2026-08-21
NMS-20158: Implement Mib compiler with Rest API and Vue

cgorantla:
Implement Mib compiler with Rest API and Vue
Should save event confs/data collection config directly to DB
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20158


synqotik:
Generally looks good, however these findings all seems worth fixing. I did notice the confirmation dialog issues while testing.

**High — both confirmation dialogs render with an empty body**

`ui/src/containers/MibCompiler.vue:64` and `:77`. The `<p>` is in the default slot;
`OnmsConfirmationDialog.vue:17` only renders `<slot name="content">`. Result: "Delete MIB File"
shows Delete/Cancel with no indication of *which* file or directory, and after a 409 the overwrite
dialog shows only a title and an "Overwrite" button — a destructive action with its explanation
stripped. Other consumers in the repo do this correctly (e.g. `DeleteEventConfigSourceDialog.vue:9`
wraps in `<template #content>`). Fix is a one-line wrap on each.

**Medium — `write()` silently no-ops on server failure**

`GenerateGraphsDialog.vue:104`: `if (result.written) { … }` with no `else`. If the second call
returns `success: false` (e.g. a dependency MIB was removed from `compiled/` between preview and
click), nothing happens — no snackbar, no `failed` emit, no step change. The preview path in the
same file does emit `failed` on `!result.success`.

**Medium — rejected files permanently block re-upload of the same name**

`UploadMibFiles.vue:164`: `existingNames` is seeded from `mibFiles.value`, which retains files
rejected client-side for size/extension and never sent to the server. Pick an oversized
`ACME-MIB.txt` → rejected but recorded → trim it and re-pick → flagged duplicate, never uploaded,
despite being in neither `pending/` nor `compiled/`. Only "Clear Logs" escapes, which also wipes the
logs the user is reading. Seed from the accepted set (`toUpload`) instead.

**Low-medium — `sourceName` derived from the wrong source**

`GenerateDataCollectionDialog.vue:144` assumes the upload endpoint derives the source name from the
file name, but `DataCollectionConfRestService` uses `dcg.getName() != null ? dcg.getName() : fileName`.
Since the dialog invites editing the XML, changing `<datacollection-group name="…">` desyncs
`isNewSource` from the server's check — the "at least one profile required" gate can be bypassed
(400 from server) or wrongly demanded. Also mismatches without any edit for MIB names containing
spaces, since `suggestedFileName` underscores them (`MibFileService.java:220`).

**Low — client size cap is half the server's**

`mibFilesValidator.ts:26` caps at 5 MB; `MibFileService.java:73` allows 10 MB. Large vendor MIBs in
the 6–10 MB range are rejected in the browser with no way to upload them.

marshallmassengill (changes requested):
Couple things to note.  I don't think any of these are blockers really... more convenience for some of them.  Others are just for sanity sake.

- MibFileService.readMibFile decodes strictly as UTF-8, so anything with a © in a copyright symbol will throw MalformedInputException and the viewer gets a 500. I think throw a more explicit warning or read as ISO-8859-1 or use a lenient decoder.
- UI mibFilesValidator.ts:28 and the accept filter reject extensionless files.  We should note this somewhere so anyone uploading a vendor mib without an extension is aware.
-UI size cap is 5 MB while the server cap is 10 MB.
-GenerateDataCollectionDialog.vue:142 derives isNewSource from the suggested file name, but the server keys on the datacollection-group name attribute, so an edited name in the textarea misreports new vs existing.
- GenerateGraphsDialog.vue:100 ignores success:false or written:false on the write call, so a failed write shows nothing.
- UploadMibFiles.vue:164 duplicate check is case-insensitive and includes previously rejected files, so a corrected re-upload of the same name is refused until logs are cleared.
- GenerateEventsDialog.test.ts mocks isWellFormedXml, so well-formedness is never exercised (May not be an issue?  Not sure how far down the rabbit hole is appropriate for that test?).

synqotik:
@cgorantla Overall looks good! Couple things I noticed.

I'd put the ellipsis menu after the view/delete menu icons.

Screenshot 2026-09-22 at 16 01 43

Also noticed if you upload e.g. `BCN-TC-MIB.mib` to `Pending`, and try to compile it, it will complain about missing dependencies like `SNMPv2-SMI`. However in my case, `SNMPv2-SMI.txt` is already in the `Compiled` section, so it's a bit confusing. I'm not sure if this is existing behavior or not, not a blocker if so.

Could be that we should check the `Compiled` folder if there are already-existing compiled MIBs that are needed for a new item in `Pending`, maybe we temporarily copy them over or something.

marshallmassengill:
> Also noticed if you upload e.g. `BCN-TC-MIB.mib` to `Pending`, and try to compile it, it will complain about missing dependencies like `SNMPv2-SMI`. However in my case, `SNMPv2-SMI.txt` is already in the `Compiled` section, so it's a bit confusing. I'm not sure if this is existing behavior or not, not a blocker if so.

I think this is somewhat existing behavior... it doesn't need to be addressed here but we don't do any sort of intelligent mapping/checking for MIB dependencies and it is an opportunity to improve ease of use if we can do some basics.

It's been a hot minute but I recall talking to @dino2gnt about this at one point and I think it's some of the more common ones that we typically miss like SNMPv2-SMI and sometimes the .txt or .mib and other times not having those extensions.

synqotik:
> I think this is somewhat existing behavior... 

@marshallmassengill agree, I think it's existing behavior. Should not be a blocker, perhaps add a ticket to look into this as a nice enhancement.

PR #8811, opened 2026-08-24
NMS-20254: Expose a discovered link's interfaces on its graph API edge

marshallmassengill:
Targeting this at foundation-2026 so it's available for additional preview work later this year.

Edges served by /api/v2/graphs now carry sourceIfIndex, targetIfIndex and discoveryProtocol (this ties a rendered link to the two interfaces it runs between).  I think this was pretty much always meant to be there based on some older NMS issues but it looks like we just never passed it through.

The values are strings. The graph REST converters have no case for a boxed Integer and fall through to toString(), so an Integer would serialize identically while claiming to be a number. PathOutageGraphProvider already writes nodeID this way.

The keys live on LegacyEdge rather than GenericProperties, which is offered for any element and where an ifIndex means nothing on a business-service or application edge. BusinessServiceEdge nests its keys the same way.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20254

### DIFF
--- a/features/graph/provider/legacy/pom.xml
+++ b/features/graph/provider/legacy/pom.xml
@@ -48,5 +48,16 @@
             <artifactId>org.opennms.features.topology.api</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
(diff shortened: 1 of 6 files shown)


dino2gnt (approved):
Seems reasonable.

christianpape (approved):
LGTM.

PR #8814, opened 2026-08-25
NMS-20258: Resolve SQL events with Translator after reload

marshallmassengill:
Found a bug while testing event translator configs.

EventTranslatorConfigFactory.update() calls unmarshall(stream) → unmarshall(stream, null), which sets m_dbConnFactory = null. After any Translator config reload, every sql value spec throws NullPointerException: ... "this.m_db" is null and those mappings stop translating until OpenNMS restarts.  

This, fixes that.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20258


dino2gnt (approved):
Sure.

cgorantla (approved):
LGTM!

PR #8816, opened 2026-08-26
NMS-20262: JSON snmp-config wipes inherited timeout and retry values

marshallmassengill:
GET /rest/v2/snmp-config/download?format=json can’t do a proper restore. It serializes SnmpConfig through its getters, and those getters substitute a default when the backing field is null, so values a definition inherited from the top-level config are written out as if they had been set on the definition.

I validated this by setting a global timeout="1800" and a definition with a read-community, the effective timeout for 127.0.0.1 drops from 1800 to 0 after a JSON download/upload cycle.  

I don’t think this is intended behavior so this is a fix for it and a test.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20262


synqotik (approved):
Good find, makes sense. This is certainly one way around the "return default if value is null" behavior that the model classes have.

PR #8817, opened 2026-08-26
NMS-20263: Improve shipped OpenAPI/Swagger Docs

marshallmassengill:
After some user feedback in mattermost it's pretty apparent we need to add response/request examples and schemas to pretty much all of our API docs.  This gets that done in bulk so it's a hefty update but it does not touch any runtime logic.

In theory, every operation in both v1 and v2 APIs now carries a summary, a unique operationId, a request schema and example where it reads a body, and a schema and example for each status code it can actually return. **Each was exercised against a running instance, so the annotations describe observed behavior rather than intent.**

I leveraged multiple claude agents to do this in one go so I would treat this as a bulk operation on the premise that some documentation that could be wrong is better than a complete void of documentation.  

It definitely will need further refinement and review.  I've spot checked a handful of these and they do seem accurate enough that someone can more easily sort out how the API works and it's a meaningful improvement over what we have today.

This one is a "I'd claim _assisted_ but that feels like cheating" by Anthropic Claude Opus 5 (x10).

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20263

### DIFF
--- a/features/bsm/rest/api/pom.xml
+++ b/features/bsm/rest/api/pom.xml
@@ -20,6 +20,8 @@
             <Bundle-RequiredExecutionEnvironment>JavaSE-1.8</Bundle-RequiredExecutionEnvironment>
             <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
             <Bundle-Version>${project.version}</Bundle-Version>
+            <!-- Swagger annotations are compile-time only; Sentinel and Minion do not ship the bundle. -->
+            <Import-Package>io.swagger.v3.oas.annotations;resolution:=optional,io.swagger.v3.oas.annotations.*;resolution:=optional,*</Import-Package>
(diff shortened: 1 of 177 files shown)


synqotik (approved):
LGTM! This will be a major help, even if you're just looking at the code itself.

One thought, in some cases "documentation-only types" were created, i.e. DTOs. Example, `opennms-webapp-rest/src/main/java/org/opennms/web/rest/v1/config/DataCollectionLookupResponse.java`. May want to have a separate follow up to see if we could use these strongly-typed classes instead of `LinkedHashMap`, etc.

marshallmassengill:
Thanks @synqotik!  Based on what @cgorantla said this morning, I'm going to merge this.  I'm going to open an NMS for follow-up.

PR #8818, opened 2026-08-26
NMS-20265: Add PostgreSQL health metrics to the default JDBC collection

dino2gnt:
Collect dead tuples, backend states by pg_stat_activity state, xid horizon ages (backend snapshots, replication slots, standby feedback, prepared xacts), and autovacuum worker usage vs autovacuum_max_workers via the OpenNMS-DB service, so the queries run against the opennms database and datasource. Add matching graph definitions and a "postgresql" threshold group covering xid horizon age, idle-in- transaction backends, and autovacuum worker saturation.

The pg_stat_activity queries need the pg_monitor role, so the installer now grants it to the OpenNMS user during database preparation, and a `pg_sa_masked` canary datasource counts sessions hidden from the monitoring role (nonzero means the grant is missing / user cannot read the views).

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20265

### DIFF
--- a/core/schema/src/main/java/org/opennms/core/schema/Migrator.java
+++ b/core/schema/src/main/java/org/opennms/core/schema/Migrator.java
@@ -461,6 +461,38 @@ public class Migrator {
     }
 
     /**
+     * <p>grantPgMonitorToUser</p>
+     *
+     * Grants the built-in pg_monitor role to the OpenNMS database user so the
+     * PostgreSQL data collection queries in the default JDBC collection can see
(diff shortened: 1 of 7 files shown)


marshallmassengill (approved):
No notes from me.  All of the things claude found I checked out and it's full of crap.  This looks good as far as I can tell and it should help us with diagnosing customer problems.

cgorantla (approved):
LGTM!

PR #8820, opened 2026-08-26
NMS-20268: Migrate the Scheduled Outages page to Vue UI

joseanesONMS:
Migrates the legacy `admin/sched-outages` JSP (list + editor) to a Vue UI backed by the existing `/rest/sched-outages` API.

Behavior is preserved: outages, their node/interface selection, type and time spans, and the notifications/polling/thresholds/data-collection references are edited exactly as before — this is a UI migration, not a change to what an outage does.

- List page shows each outage's type, node labels/interface addresses, time-span count, and which subsystems reference it, with create and delete; all memberships come from a single name-less applies-to call.
- Editor manages node/interface pickers, the outage type, its time spans, and the Applies-To matrix; "Select all nodes and interfaces" maps to the `match-any` interface, and a load failure on an existing outage disables editing rather than presenting an empty form.
- Interface autocomplete searches `ipHostName==*term*` (the v2 API rejects wildcards on `ipAddress`); a complete IP matches `ipAddress` exactly and is always offerable itself.
- Time spans are written in the exact wire formats the parser keys on by length — `dd-MMM-yyyy HH:mm:ss` for `specific`, `HH:mm:ss` for daily/weekly/monthly — matching the legacy form.
- Adds one read-only endpoint, `GET /sched-outages[/{name}]/applies-to` (admin-gated), returning the notifd flag plus every poller/collectd/threshd package with its outage calendars and an applied flag for the named variant.
- The sidebar menu, Admin page, and global-search action all point at the Vue page.
- Uses Onms-XXX components; full `eslint` and `vue-tsc` pass, 38 UI tests and `ScheduledOutagesRestServiceIT` (14) pass, `WillItUnmarshalIT` covers the search-actions change, and `MenuHeaderIT` targets the Vue page.
- Not included: the legacy "Add with path outage dependency" option (no REST support today); tracked as a follow-up.

### DIFF
--- a/ui/src/components/ScheduledOutages/NodeInterfacePicker.vue
+++ b/ui/src/components/ScheduledOutages/NodeInterfacePicker.vue
@@ -1,0 +1,156 @@
+<template>
+  <div class="picker" :data-test="`picker-${mode}`">
+    <div class="picker-title">{{ mode === 'node' ? 'Nodes' : 'Interfaces' }}</div>
+    <FormField :label="`Search (max 200 results)`" :for="`picker-${mode}-input`">
+      <div class="search-row">
+        <OnmsAutoComplete
+          v-model="selection"
+          :inputId="`picker-${mode}-input`"
+          :suggestions="suggestions"
+          optionLabel="label"
+          :placeholder="mode === 'node' ? 'Node label' : 'IP address'"
+          forceSelection
+          fluid
+          :data-test="`picker-${mode}-search`"
+          @complete="onComplete"
+        />
+        <OnmsButton
+          label="Add"
+          icon="pi pi-plus"
+          :disabled="!selection"
+          :data-test="`picker-${mode}-add`"
+          @click="addSelection"
+        />
+      </div>
+    </FormField>
+
+    <div class="current">
+      <div class="current-title">Current selection:</div>
+      <p v-if="!items.length" class="none" :data-test="`picker-${mode}-empty`">
+        {{ mode === 'node' ? 'No specific nodes selected' : 'No specific interfaces selected' }}
+      </p>
+      <ul v-else class="chips">
+        <li v-for="(item, index) in items" :key="index" class="chip">
+          <span>{{ labelFor(item) }}</span>
+          <OnmsIconButton
+            :icon="Delete"
+            severity="danger"
+            :title="`Remove ${labelFor(item)}`"
+            :aria-label="`Remove ${labelFor(item)}`"
+            :data-test="`picker-${mode}-remove`"
+            @click="emit('remove', index)"
+          />
+        </li>
+      </ul>
+    </div>
+  </div>
+</template>
+
+<script setup lang="ts">
+import { ref } from 'vue'
+import { OnmsAutoComplete, OnmsButton, OnmsIconButton } from '@opennms/onms-ui'
+import Delete from '@opennms/onms-ui/icons/action/Delete.vue'
+import FormField from '@/components/Common/FormField.vue'
+import { OutageInterface, OutageNode } from '@/types/scheduledOutage'
+import { searchOutageInterfaces, searchOutageNodes } from '@/services/scheduledOutagesService'
+
+interface NodeSuggestion { label: string, id: number }
+interface InterfaceSuggestion { label: string, address: string }
+type Suggestion = NodeSuggestion | InterfaceSuggestion
+
+const props = defineProps<{
+  mode: 'node' | 'interface'
+  items: (OutageNode | OutageInterface)[]
+}>()
+
+const emit = defineEmits<{
+  add: [value: OutageNode | OutageInterface]
+  remove: [index: number]
+}>()
+
+const selection = ref<Suggestion | null>(null)
+const suggestions = ref<Suggestion[]>([])
+
+const onComplete = async (query: string) => {
+  if (props.mode === 'node') {
+    const nodes = await searchOutageNodes(query)
+    suggestions.value = nodes.map(n => ({ label: `${n.label} (id ${n.id})`, id: n.id }))
+  } else {
+    const interfaces = await searchOutageInterfaces(query)
+    suggestions.value = interfaces.map(i => ({
+      label: i.nodeLabel ? `${i.address} — ${i.nodeLabel}` : i.address,
+      address: i.address
+    }))
+  }
+}
+
+const addSelection = () => {
+  const sel = selection.value
+  if (!sel) {
+    return
+  }
+  if (props.mode === 'node' && 'id' in sel) {
+    emit('add', { id: sel.id })
+  } else if (props.mode === 'interface' && 'address' in sel) {
+    emit('add', { address: sel.address })
+  }
+  selection.value = null
+}
+
+const labelFor = (item: OutageNode | OutageInterface): string =>
+  'id' in item ? `Node id ${item.id}` : item.address
+</script>
+
+<style scoped lang="scss">
+.picker {
+  .picker-title {
+    font-weight: 600;
+    margin-bottom: 0.5rem;
+  }
+
+  .search-row {
+    display: flex;
+    align-items: flex-start;
+    gap: 0.5rem;
+
+    :deep(.p-autocomplete) {
+      flex: 1 1 auto;
+      min-width: 0;
+    }
+  }
+
+  .current {
+    margin-top: 0.75rem;
+  }
+
+  .current-title {
+    font-weight: 600;
+    margin-bottom: 0.25rem;
+  }
+
+  .none {
+    margin: 0;
+    font-style: italic;
+    color: var(--p-text-muted-color);
+  }
+
+  .chips {
+    list-style: none;
+    margin: 0;
+    padding: 0;
+    display: flex;
+    flex-direction: column;
+    gap: 0.25rem;
+  }
+
+  .chip {
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    gap: 0.5rem;
+    padding: 0.15rem 0.25rem;
+    border-radius: 4px;
+    background: var(--p-content-hover-background, rgba(127, 127, 127, 0.08));
+  }
+}
+</style>
--- a/ui/src/components/ScheduledOutages/NodeInterfacePicker.vue
+++ b/ui/src/components/ScheduledOutages/NodeInterfacePicker.vue
@@ -54,17 +54,20 @@ import FormField from '@/components/Common/FormField.vue'
 import { OutageInterface, OutageNode } from '@/types/scheduledOutage'
 import { searchOutageInterfaces, searchOutageNodes } from '@/services/scheduledOutagesService'
 
-interface NodeSuggestion { label: string, id: number }
+interface NodeSuggestion { label: string, id: number, nodeLabel: string }
 interface InterfaceSuggestion { label: string, address: string }
 type Suggestion = NodeSuggestion | InterfaceSuggestion
 
 const props = defineProps<{
   mode: 'node' | 'interface'
   items: (OutageNode | OutageInterface)[]
+  // node id -> label, resolved by the editor; ids without an entry are shown
+  // as not-found (deleted nodes still referenced by the outage config)
+  nodeLabels?: Record<number, string>
 }>()
 
 const emit = defineEmits<{
-  add: [value: OutageNode | OutageInterface]
+  add: [value: OutageNode | OutageInterface, label?: string]
   remove: [index: number]
 }>()
 
@@ -74,7 +77,7 @@ const suggestions = ref<Suggestion[]>([])
 const onComplete = async (query: string) => {
   if (props.mode === 'node') {
     const nodes = await searchOutageNodes(query)
-    suggestions.value = nodes.map(n => ({ label: `${n.label} (id ${n.id})`, id: n.id }))
+    suggestions.value = nodes.map(n => ({ label: `${n.label} (id ${n.id})`, id: n.id, nodeLabel: n.label }))
   } else {
     const interfaces = await searchOutageInterfaces(query)
     suggestions.value = interfaces.map(i => ({
@@ -90,15 +93,20 @@ const addSelection = () => {
     return
   }
   if (props.mode === 'node' && 'id' in sel) {
-    emit('add', { id: sel.id })
+    emit('add', { id: sel.id }, sel.nodeLabel)
   } else if (props.mode === 'interface' && 'address' in sel) {
     emit('add', { address: sel.address })
   }
   selection.value = null
 }
 
-const labelFor = (item: OutageNode | OutageInterface): string =>
-  'id' in item ? `Node id ${item.id}` : item.address
+const labelFor = (item: OutageNode | OutageInterface): string => {
+  if (!('id' in item)) {
+    return item.address
+  }
+  const label = props.nodeLabels?.[item.id]
+  return label ? `${label} (id ${item.id})` : `Node id ${item.id} (not found)`
+}
 </script>
 
 <style scoped lang="scss">
--- a/ui/src/components/ScheduledOutages/NodeInterfacePicker.vue
+++ b/ui/src/components/ScheduledOutages/NodeInterfacePicker.vue
@@ -26,30 +26,31 @@
 
     <div class="current">
       <div class="current-title">Current selection:</div>
-      <p v-if="!items.length" class="none" :data-test="`picker-${mode}-empty`">
+      <!-- match-any lives in the interface list; the node picker mirrors it as
+           a display-only chip so both read as "everything selected" -->
+      <div v-if="mode === 'node' && matchAny" class="chips">
+        <OnmsChip label="All Nodes" :data-test="`picker-${mode}-all`" />
+      </div>
+      <p v-else-if="!items.length" class="none" :data-test="`picker-${mode}-empty`">
         {{ mode === 'node' ? 'No specific nodes selected' : 'No specific interfaces selected' }}
       </p>
-      <ul v-else class="chips">
-        <li v-for="(item, index) in items" :key="index" class="chip">
-          <span>{{ labelFor(item) }}</span>
-          <OnmsIconButton
-            :icon="Delete"
-            severity="danger"
-            :title="`Remove ${labelFor(item)}`"
-            :aria-label="`Remove ${labelFor(item)}`"
-            :data-test="`picker-${mode}-remove`"
-            @click="emit('remove', index)"
-          />
-        </li>
-      </ul>
+      <div v-else class="chips">
+        <OnmsChip
+          v-for="(item, index) in items"
+          :key="index"
+          :label="labelFor(item)"
+          removable
+          :data-test="`picker-${mode}-chip`"
+          @remove="emit('remove', index)"
+        />
+      </div>
     </div>
   </div>
 </template>
 
 <script setup lang="ts">
 import { ref } from 'vue'
-import { OnmsAutoComplete, OnmsButton, OnmsIconButton } from '@opennms/onms-ui'
-import Delete from '@opennms/onms-ui/icons/action/Delete.vue'
+import { OnmsAutoComplete, OnmsButton, OnmsChip } from '@opennms/onms-ui'
 import FormField from '@/components/Common/FormField.vue'
 import { OutageInterface, OutageNode } from '@/types/scheduledOutage'
 import { searchOutageInterfaces, searchOutageNodes } from '@/services/scheduledOutagesService'
@@ -64,6 +65,8 @@ const props = defineProps<{
   // node id -> label, resolved by the editor; ids without an entry are shown
   // as not-found (deleted nodes still referenced by the outage config)
   nodeLabels?: Record<number, string>
+  // the outage applies to everything; the node picker shows an All Nodes chip
+  matchAny?: boolean
 }>()
 
 const emit = defineEmits<{
@@ -102,7 +105,7 @@ const addSelection = () => {
 
 const labelFor = (item: OutageNode | OutageInterface): string => {
   if (!('id' in item)) {
-    return item.address
+    return item.address === 'match-any' ? 'All Interfaces' : item.address
   }
   const label = props.nodeLabels?.[item.id]
   return label ? `${label} (id ${item.id})` : `Node id ${item.id} (not found)`
@@ -118,7 +121,7 @@ const labelFor = (item: OutageNode | OutageInterface): string => {
 
   .search-row {
     display: flex;
-    align-items: flex-start;
+    align-items: center;
     gap: 0.5rem;
 
     :deep(.p-autocomplete) {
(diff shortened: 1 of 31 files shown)


joseanesONMS:
Migrates the Scheduled Outages list + editor to Vue over the existing `/rest/sched-outages` API, plus one read-only `applies-to` endpoint so the subsystem matrix loads in a single call. Uses Onms-XXX components; eslint/`vue-tsc` clean, 12 UI tests and `ScheduledOutagesRestServiceIT` (14) green, and I verified the new endpoint live on a running instance. The legacy "Add with path outage dependency" option has no REST support today and is deferred as a follow-up.

marshallmassengill (changes requested):
Blockers

  1. build-ui is red. 8 @stylistic/object-curly-spacing errors in ui/tests/services/scheduledOutagesService.test.ts (lines 45, 51×2, 58, 67, 75, 97, 104), all auto-fixable with pnpm lint:fix. Reproduced  locally. The PR body says "eslint and vue-tsc pass"; vue-tsc does, eslint does not.
  2. The interface picker cannot add anything. I ran the exact URL searchOutageInterfaces builds against a live 36.0.3 instance: /api/v2/ipinterfaces?limit=200&_s=ipInterface.ipAddress==*192* returns 500,  "could not resolve property: ipInterface of: org.opennms.netmgt.model.OnmsIpInterface" (the root entity's properties are registered bare, SearchProperties.java:515). Correcting it to ipAddress does not rescue it: that property is typed IP_ADDRESS, so the wildcard becomes a LIKE pattern and the server answers 500 "Invalid IPAddress %192%". Only a fully exact ipAddress==127.0.0.4 succeeds, and the code wraps every term in *…* unconditionally. Since OnmsAutoComplete has forceSelection and Add is gated on a chosen suggestion, there is no path to adding an interface. Legacy did this with getAllManagedIpInterfacesLike(term) behind jsonIpInterfaces.jsp; v2 has no equivalent (I checked ipLike=, it's silently ignored). Workable substitutes: exact-IP ipAddress==, ipHostName==*term* (verified working), or fetching a page and filtering client-side.
  3. A transient read failure in the editor silently becomes data loss. onMounted treats getScheduledOutage(name) === null identically to "new": the form renders empty under the real outage name, and Save POSTs a whole-object replace, wiping nodes, interfaces, and times. getScheduledOutage swallows 404, 500, and network errors alike, so a blip on load plus a Save destroys the outage. Load failure needs to  be distinguished from isNew and Save blocked.

Additional  Findings

  4. ScheduledOutages.load() fires one applies-to request per outage, each of which re-reads all four subsystem configs and enumerates every poller/collectd/threshd package, with Promise.all blocking the table until all finish. The legacy JSP read each config once and did contains(outageName) per row (index.jsp:141-158). The new endpoint's per-outage shape is what forces the N+1 the PR body says it avoids.
  5. selectAll and onTypeChange use window.confirm; the list page in this same PR uses OnmsConfirmationDialog.
  6. Selection parity loss: the legacy list showed each node's label and each interface address (flagging deleted/not-found nodes); the new list shows "3 node(s), 1 interface(s)" and the editor's chips read "Node id 5" rather than the label.
  7. Test coverage is thin where it mattered: 16 tests (body says 12) cover only outageTime and the service. Nothing covers either container, the pickers,  TimeSpanEditor, or AppliesToMatrix, and searchOutageInterfaces has no test at all. The node-search test asserts a URL string against a mock, so that shape of test could not have caught #2 either.
  8. On a first-load failure the page shows "Showing the last known list" above an empty table.
  9. opennms-webapp/src/main/webapp/admin/index.jsp:203 and opennms-base-assembly/src/main/filtered/etc/search-actions.xml:160 still route to the legacy JSP, so the sidebar and the Admin page / global search land on different pages. Same gap as 8727 and 8735.
  10. Smaller: breadcrumb { label: name, to: '' }; no guard for an empty ?name=; yearOptions spans only currentYear..+10, so a past-year specific span can be displayed but not re-entered; AppliedMark renders a bare ✓/— with no accessible text.

joseanesONMS:
@marshallmassengill all three blockers and the findings are addressed in `4b307f8c2b1`:

1. **Lint** — `pnpm lint` now runs clean over the whole tree (the 8 test-file errors are fixed); the body's claim is re-verified.
2. **Interface picker** — partial terms search `ipHostName==*term*` and a complete IP matches `ipAddress` exactly (verified against a running instance, same results you saw: bare-property + no wildcards on `ipAddress`); a complete typed IP is always offerable since poll-outages accepts any address. Service tests pin both URL shapes.
3. **Load-failure data loss** — a failed read of an existing outage now disables editing with an explicit message instead of rendering an empty form; covered by a container test.
4. The list derives every outage's memberships from one name-less `applies-to` call — the endpoint now exposes each package's outage calendars (IT asserts the contract).
5. `window.confirm` replaced with `OnmsConfirmationDialog` for both confirms.
6. The list shows node labels/interface addresses (single OR query for labels) and the editor chips read `label (id N)` / flag not-found nodes.
7. 38 UI tests now, including both containers, the picker, TimeSpanEditor, AppliesToMatrix, and `searchOutageInterfaces`. The container test also caught (and the commit fixes) the editor card rendering blank under `OnmsCard`'s content slot.
8–10. First-load failure wording fixed, `admin/index.jsp` + `search-actions.xml` repointed (`WillItUnmarshalIT` 888/888), breadcrumb/empty-name/year-range/aria fixed.

The push carries `!smoke` so the smoke tests run on this build.

marshallmassengill:
Check CI and make sure this passes and then I'll do another review @joseanesONMS

joseanesONMS:
@marshallmassengill the `smoke-test-core` failure was `AdminPageIT.testAllLinks`: repointing the Admin page's Scheduled Outages link at the Vue page left that entry asserting the legacy JSP's `New Name` input. Fixed in `5b3a4081edd` — the entry now asserts the Vue list page's card title (same locator family as the `MenuHeaderIT` change already in the PR). `ScheduledOutageIT` drives the legacy JSP by direct URL and still passes, so it's untouched. Everything else in the run was green (build, build-ui, integration-test, smoke-minion/sentinel); the push carries `!smoke` so the full set re-runs.

joseanesONMS:
@marshallmassengill CI is fully green on `ec4d75a2ea4` — build, build-ui, integration-test, and all three smoke suites (the earlier `smoke-test-minion` red was a zero-failure timeout; it passed on the re-run, and `smoke-test-core` is green with the `AdminPageIT` fix). Ready for your re-review.

joseanesONMS:
@marshallmassengill friendly ping — CI has been fully green on `ec4d75a2ea4` since Friday (build, build-ui, integration-test, and all three smoke suites) and the branch is conflict-free against develop. Ready for that second review whenever you have a moment.

ui/src/containers/ScheduledOutages.vue
```diff
@@ -0,0 +1,305 @@
...
+    <div class="onms-col-12">
+      <BreadCrumbs :items="breadcrumbs" />
+    </div>
+  </div>
+
+  <div class="scheduled-outages">
+    <TableCard>
+      <div class="header">
+        <div class="card-title">Scheduled Outages</div>
+        <div class="header-actions">
```
synqotik: I'd remove the `FormField` / input from the `.header` section and just have a `Create New Outage` button, that can pop up a dialog where the user can type in the name and create. Also the `Create` button is clipped. Under the title `div.card-title`, I'd put some short text, like "View and create Scheduled Outages", then have the info icon button next to it.

ui/src/services/scheduledOutagesService.ts
```diff
@@ -0,0 +1,194 @@
...
+    await rest.delete(path)
+  }
+}
+
+// Node label autocomplete for the node picker (id + label).
+export const searchOutageNodes = async (query: string): Promise<{ id: number, label: string }[]> => {
+  try {
+    const term = sanitizeSearchTerm(query)
+    const filter = term ? `&_s=node.label==*${term}*` : ''
+    const resp = await v2.get(`/nodes?limit=200${filter}`)
```
synqotik: Ideally this should be a case-insensitive search.

ui/src/components/ScheduledOutages/NodeInterfacePicker.vue
```diff
@@ -0,0 +1,164 @@
...
+  add: [value: OutageNode | OutageInterface, label?: string]
+  remove: [index: number]
+}>()
+
+const selection = ref<Suggestion | null>(null)
+const suggestions = ref<Suggestion[]>([])
+
+const onComplete = async (query: string) => {
+  if (props.mode === 'node') {
+    const nodes = await searchOutageNodes(query)
```
synqotik, line 82: Should be case-insensitive search.

ui/src/components/ScheduledOutages/NodeInterfacePicker.vue
```diff
@@ -0,0 +1,164 @@
...
+          :inputId="`picker-${mode}-input`"
+          :suggestions="suggestions"
+          optionLabel="label"
+          :placeholder="mode === 'node' ? 'Node label' : 'IP address'"
+          forceSelection
+          fluid
+          :data-test="`picker-${mode}-search`"
+          @complete="onComplete"
+        />
+        <OnmsButton
```
synqotik, line 17: The `Add` buttons aren't fully vertically-aligned, they should be vertically centered with the input box itself, not the entire `FormField`.

ui/src/containers/ScheduledOutageEditor.vue
```diff
@@ -0,0 +1,525 @@
...
+      <BreadCrumbs :items="breadcrumbs" />
+    </div>
+  </div>
+
+  <div class="outage-editor">
+    <OnmsCard>
+      <template #content>
+      <div class="page-header">
+        <h2 class="headline3" data-test="editor-title">Editing Outage: {{ name }}</h2>
+        <OnmsButton variant="text" label="Cancel" data-test="cancel" @click="goBack" />
```
synqotik: Instead of "Cancel", I'd either have an X icon, or else use the `<- Go Back` before the title, like on SNMP Configuration etc.

ui/src/containers/ScheduledOutageEditor.vue
```diff
@@ -0,0 +1,525 @@
...
+              :collectors="applicability.collectors"
+              @update:notifications="applicability.notifications = $event"
+              @togglePackage="togglePackage"
+              @setAll="setAll"
+            />
+          </section>
+        </div>
+
+        <div v-if="!loadFailed" class="actions">
+          <OnmsButton label="Save Outage" data-test="save" :disabled="saving" @click="save" />
```
synqotik: Maybe instead of "Save Outage", could be just "Save" if you are editing, or "Create" if it's a new one. (Also, "Outage" is confusing since your saving a Scheduled Outage, not an Outage).

ui/src/components/ScheduledOutages/NodeInterfacePicker.vue
```diff
@@ -0,0 +1,164 @@
...
+      </div>
+    </FormField>
+
+    <div class="current">
+      <div class="current-title">Current selection:</div>
+      <p v-if="!items.length" class="none" :data-test="`picker-${mode}-empty`">
+        {{ mode === 'node' ? 'No specific nodes selected' : 'No specific interfaces selected' }}
+      </p>
+      <ul v-else class="chips">
+        <li v-for="(item, index) in items" :key="index" class="chip">
```
synqotik: I'd use `OnmsChip` here.

ui/src/containers/ScheduledOutageEditor.vue
```diff
@@ -0,0 +1,525 @@
...
+
+        <div v-else class="editor-grid">
+          <section class="selection">
+            <h3 class="section-title">Nodes and Interfaces</h3>
+            <div class="pickers">
+              <NodeInterfacePicker mode="node" :items="outage.node ?? []" :nodeLabels="nodeLabels" @add="addNode" @remove="removeNode" />
+              <NodeInterfacePicker mode="interface" :items="outage.interface ?? []" @add="addInterface" @remove="removeInterface" />
+            </div>
+            <div class="match-any-row">
+              <OnmsButton
```
synqotik, line 36: Image

A few things to consider changing here. In `NodeInterfacePicker`, use `OnmsChip`. 

If user selects "Select All", instead of the `match-any` chip, it should be a more "friendly" name like `All Nodes` or `All Interfaces`

ui/src/containers/ScheduledOutageEditor.vue
```diff
@@ -0,0 +1,525 @@
...
+                variant="outlined"
+                label="Select all nodes and interfaces"
+                data-test="match-any"
+                @click="selectAll"
+              />
+              <span v-if="isMatchAny" class="match-any-note" data-test="match-any-note">
```
synqotik: Not sure about this text, maybe fix the Chips inside `Nodes` and `Interfaces` above to say `All Nodes` and `All Interfaces`.

ui/src/containers/ScheduledOutageEditor.vue
```diff
@@ -0,0 +1,525 @@
...
+                Applies to all nodes and interfaces.
+              </span>
+            </div>
+            <p v-if="showSelectionError" class="field-error" data-test="selection-error">
+              You must select at least one node or interface for this scheduled outage.
+            </p>
+          </section>
+
+          <section class="schedule">
```
synqotik, line 48: There's a bunch of vertical whitespace here above the Outage Type section.

ui/src/containers/ScheduledOutageEditor.vue
```diff
@@ -0,0 +1,525 @@
...
+            <FormField label="Outage Type" for="outage-type">
+              <OnmsSelect
+                v-model="outage.type"
```
synqotik, line 51: This should probably have a pre-selected default.

ui/src/components/ScheduledOutages/TimeSpanEditor.vue
```diff
@@ -0,0 +1,154 @@
...
+        </div>
+        <div class="span-row">
+          <span class="span-label">End</span>
+          <OnmsSelect v-model="fields.endHour" :options="HOURS" optionLabel="label" optionValue="value" data-test="time-end-hour" />
+          <OnmsSelect v-model="fields.endMinute" :options="MINUTES" optionLabel="label" optionValue="value" />
+          <OnmsSelect v-model="fields.endSecond" :options="SECONDS" optionLabel="label" optionValue="value" />
+        </div>
+      </template>
+
+      <OnmsButton label="Add Outage" icon="pi pi-plus" data-test="add-time" @click="addSpan" />
```
synqotik: This should probably say "Add Timespan", and also currently it's the entire width of the panel, should just be wide enough for it's own content.

ui/src/components/ScheduledOutages/TimeSpanEditor.vue
```diff
@@ -0,0 +1,154 @@
+<template>
+  <div class="time-spans" data-test="time-spans">
+    <div class="section-title">Time spans</div>
+
+    <div class="new-span">
+      <!-- specific: full start/end dates; others: time-of-day (+ optional day) -->
+      <template v-if="type === 'specific'">
+        <div class="span-row">
+          <span class="span-label">Start</span>
```
synqotik, line 9: I'd say most of these (day of month, start time, etc. etc.) should have some pre-selected value that the user could change. I think some of them are, but some aren't.

I'd also change `Day` to either `Day of Week` or `Day of Month`.

We should also see if we can use a Date/Time picker control. I will open a separate ticket and look into that, don't want to do it as part of this work. Leave as-is with all the `OnmsSelect` for now, then once I make the new control you can fix.

synqotik (changes requested):
Added some UI/UX comments.

joseanesONMS:
@synqotik all 13 addressed in `ef757202ff8` — thanks for the thorough pass:

1. **List header** — the inline name field is gone; a `Create New Outage` button opens a small dialog (name + duplicate check, Enter submits). The title now has "View and create scheduled outages." underneath with the info button next to it.
2. **Node search** — truly case-insensitive now via the v1 `comparator=ilike` endpoint (v2 FIQL wildcards turn into a case-sensitive LIKE; verified against a live instance).
3. **Interface search** — v1 has no ipinterfaces endpoint, so it ORs the raw and lowercased term; DNS names are stored lowercase, which makes it case-insensitive in practice. A true ilike there needs a server-side change.
4. **Add button alignment** — the search row centers the button on the input now.
5. **Top Cancel** — replaced with the `← Go Back` before the title, per the SNMP Data Collection pattern.
6. **Save Outage** — reads `Create` for a new outage, `Save` when editing; the title says New/Edit Scheduled Outage accordingly.
7/8/9. **Chips** — selections are `OnmsChip`s (removable); with Select All the pickers show **All Nodes** / **All Interfaces** instead of the `match-any` token, and the redundant note text is gone.
10. **Whitespace** — the grid rows are content-sized now, so the blank band above Outage Type is gone (it was the applies column's spare height being distributed).
11. **Outage Type** — a new outage starts pre-selected on Specific.
12. **Add Timespan** — renamed and fit-content width.
13. **Day rows** — labeled Day of Week / Day of Month, and every select has a valid pre-selected value (the shared day field now resets when the type switches — it used to keep 'sunday' into monthly). Left the `OnmsSelect`s in place for the future date/time control as you suggested.

marshallmassengill (changes requested):
Couple blockers with that latest update:

  - applicationContext-spring-security.xml:138 — the new admin gates use exact paths, so GET /rest/sched-outages/applies-to/ (trailing slash) misses AntPathMatcher, falls to the /rest/** GET rule, and serves every poller/collectd/threshd package and calendar to any ROLE_USER; use …/applies-to/**.
  - ui/src/containers/ScheduledOutageEditor.vue:261 — addNode doesn't strip the match-any interface (legacy editoutage.jsp:69 did), so with "Select all nodes and interfaces" active the added node is invisible and Save writes match-any plus <node id>.
  - ui/src/containers/ScheduledOutages.vue:278 — confirmDelete's catch sets loadError, then finally { await load() } clears it; a failed DELETE shows nothing at all.

joseanesONMS:
@marshallmassengill all three addressed in `84b3a127171`:

- **applies-to gate** — both intercept-url patterns now end in `/**`, so `GET /rest/sched-outages/applies-to/` (trailing slash) and the per-outage form stay admin-only instead of falling through to the `/rest/**` ROLE_USER rule. Checked the patterns against `AntPathMatcher` for both spellings and confirmed they don't widen to the other `sched-outages` GETs.
- **addNode** — adding a specific node strips the `match-any` pseudo-interface, matching legacy `editoutage.jsp`; container test covers it.
- **Delete failure** — reported through a separate `actionError` that the post-delete reload doesn't reset; container test covers a rejected DELETE followed by a successful reload.

Tests, eslint and `vue-tsc` green; push carries `!smoke`.

marshallmassengill (changes requested):
Two minor things but should be addressed:

- Partial-IP interface search regresses. searchOutageInterfaces sends _s=ipHostName==*term* for anything that isn't a complete IP, so typing 192.168.1. finds nothing wherever reverse DNS resolved (ipHostName holds the hostname, not the address). Legacy jsonIpInterfaces.jsp called NetworkElementFactory.getAllManagedIpInterfacesLike, which is ilike ipHostName ANYWHERE OR ilike ipAddress ANYWHERE (NetworkElementFactory.java:620). The fix follows an existing pattern: register new IpLikeCriteriaBehavior("ipAddress") in IpInterfaceRestService.getCriteriaBehaviors() the way AlarmRestService:161 and EventRestService:134 do, then query ipAddress==192.168.1.*. NodeRestService:291 states that behavior works directly against the ipinterface table, which is this endpoint's root alias.

- An applies-to read failure renders as authoritative "not applied". getOutageApplicability returns null on error; the editor then shows "No packages configured." for all three subsystems with notifications unchecked and no error line, and the list page's four columns all show "—". Every other null-vs-empty case in the PR is handled deliberately; this one is not, and no test covers it.

joseanesONMS:
@marshallmassengill both addressed in `39edccb7e01`:

- **Partial-IP search** — `/api/v2/ipinterfaces` now accepts iplike patterns on `ipAddress`, following the Alarm/Event pattern. One wrinkle: `IplikeRestriction` interpolates the attribute name as the SQL column, so the behavior is registered under the FIQL key `ipAddress` but names the `ipaddr` column (`new IpLikeCriteriaBehavior("ipAddr")`); `ipAddress` as the column name would have hit a nonexistent column. The picker widens a dotted IPv4 prefix to four octets (`192.168.1.` → `192.168.1.*`) and ORs it with the hostname wildcard, so it matches whether or not reverse DNS resolved. `IpInterfaceRestServiceIT` asserts `10.10.10.*` / `10.10.*.*` hit, `10.10.11.*` is 204, and the OR with a non-matching hostname still returns the row; service tests pin the URL shapes.
- **Applies-to read failure** — the editor shows an error in place of the matrix and skips membership PUT/DELETEs on save (the outage itself stays editable); the list renders the four columns as `?` / `aria-label="unknown"` with a message instead of `—`. Container tests cover both.

`IpInterfaceRestServiceIT` + `ScheduledOutagesRestServiceIT` 16/16 locally, 31 UI tests, eslint and `vue-tsc` green; push carries `!smoke`.

marshallmassengill (changes requested):
Regression to resolve:

- IpInterfaceRestService.java:115 — map.put("ipAddress", new IpLikeCriteriaBehavior("ipAddr")) lands after putAll(IP_INTERFACE_BEHAVIORS), so it replaces the INET_ADDRESS_CONVERTER behavior for all ipAddress queries, not just wildcard ones. Exact lookups now render as ipLike({alias}.ipAddr, ?) against the database function, and iplike.sql's IPv6 branch requires both the address and the rule to match a regex of exactly eight colon groups, falling through to else return 'f'. So ipAddress==fe80::1 silently returns nothing where the old eq matched it via the InetAddress user type, and this PR's own picker sends exactly that shape: isCompleteIpAddress accepts colon forms and the exact branch emits _s=ipAddress==${term}. IPv6 exact search in the interface picker regressed while IPv4 partial search was fixed. isWildcard is already passed to beforeVisit, so gating iplike on it (keeping InetAddress equality for literal values) fixes both without touching the exact path. An IPv6 case in IpInterfaceRestServiceIT alongside the four new IPv4 asserts would have caught it, and JUnitTemporaryDatabase can force the pl/pgsql iplike to make the test meaningful.

Possible fix: beforeVisit already receives isWildcard, so route to iplike only when the value actually carries a wildcard and let the existing INET_ADDRESS_CONVERTER plus eq handle literals. That needs a converter that yields an InetAddress for the non-wildcard case, since IpLikeCriteriaBehavior uses Function.identity(), so it is a small subclass rather than a one-liner. Overriding a distinct search key instead of ipAddress would also work and would leave the released contract alone.

joseanesONMS:
@marshallmassengill fixed in `9fc640ae34a` along the lines you suggested: `ipAddress` is now an `IpAddressCriteriaBehavior` whose converter yields an `InetAddress` for literals (the visitor's `eq`/`ne` path is untouched) and only a wildcard value is applied as `iplike` on the `ipaddr` column with the property skipped — `shouldSkipProperty` returns `isWildcard`.

`IpInterfaceRestServiceIT.testFiqlSearch` now runs with `plpgsqlIplike = true`, adds an `fe80::1` interface, and asserts the literal IPv4 and IPv6 matches, `!=`, and a literal miss (204) next to the four wildcard cases. I confirmed the IPv6 assertion fails (200 expected, 204 returned) with the previous plain-iplike registration before restoring the fix. Push carries `!smoke`.

marshallmassengill (approved):
I'm good with this.  Wait for approval from @synqotik to merge.

synqotik (approved):
LGTM!

A few things to fix, can be done *after* this merge.

In the "Create New Outage" dialog, need more vertical spacing between rows. `Cancel` button should be `variant=ghost`.

`Cancel` button on main page should also be `variant=ghost`.

Need `1rem` spacing above `Time Spans`.

Consider using the updated `OnmsDatePicker` to pick dates/times, or at least times.

PR #8821, opened 2026-08-27
NMS-20100: Migrate the Notifications admin page to Vue UI

joseanesONMS:
Migrates the Notifications admin area from the legacy JSPs to a single PrimeVue/Vue page, landing the NMS-20100 epic to develop.

This is the umbrella PR for the five slices already reviewed and merged into the epic branch; it carries no new behavior beyond them and is refreshed on current develop.

- Notifications page with tabs: General configuration, Event Notifications, Destination Paths, and Path Outages (#8709, #8710, #8711, #8712).
- The notification bell links to the new page (#8719).
- Config is served by `NotificationConfigRestService` under `/api/v2/notification-config`, per synqotik's review (moved to v2, no `m_` prefixes, no `@XmlAttribute`).
- Uses Onms-XXX components; the merge from develop adopted the onms-ui icon seam and the `FormField` `reserveLabelSpace` addition.
- Verified after the develop merge: eslint and `vue-tsc` build pass, `opennms-webapp-rest` compiles, and the FormField/notification unit tests are green.
- Reviewers: the slices were approved/addressed individually; this PR is for landing the epic as a whole.


marshallmassengill (changes requested):
Blockers:

1. The "Test" button on Destination Paths can never work. ui/src/services/notificationConfigService.ts:testDestinationPath posts to v2 (/opennms/api/v2/notifications/destination-paths/{name}/trigger), but triggerDestinationPath is defined only on the v1 resource (opennms-webapp-rest/.../v1/NotificationRestService.java:223). The v2 notifications resource is a plain AbstractDaoRestService whose only sub-paths are {id}, count, properties[/{id}], so a three-segment POST 404s and the user gets the generic "Failed to trigger test notification" snackbar. Use the rest axios instance. No IT covers this.
  2. Event Severity is editable but silently discarded on every edit. The new dialog exposes an event-severity select and sends it (EventNotificationEditorDialog.vue:276,663), but NotificationManager.replaceNotification copies twelve fields and never calls setEventSeverity (NotificationManager.java:1090-1116). POST persists it, PUT ignores it, so changing or clearing severity on an existing notification is a no-op that reports success. The legacy JSP wizard never exposed the field, so this PR is what makes the gap reachable.

joseanesONMS:
@marshallmassengill both blockers fixed in `4af103ab555`:

1. **Destination-path Test button** — `testDestinationPath` now posts via the v1 `rest` instance (the trigger endpoint lives on `v1/NotificationRestService`); a service test pins the instance and path so a v2 regression can't slip back in.
2. **Event severity on update** — `NotificationManager.replaceNotification` now copies `eventSeverity`; `NotificationConfigRestServiceIT` covers set, change, and clear on update (22/22 green locally against Postgres).

The push carries `!smoke` so the smoke tests run on this build.

marshallmassengill (changes requested):
Some minor issues now:

1. Deleting a destination path has no referential check. NotificationConfigRestService.deleteDestinationPath (opennms-webapp-rest/.../v2/NotificationConfigRestService.java:481) removes a path even when event notifications reference it, leaving them pointing at a name that no longer exists. updateDestinationPath goes to real trouble to cascade a rename and has an IT for it; delete is the asymmetric case.
  2. Config references are never validated against what exists. validateEventNotification doesn't check that destinationPath names a real path, and validateDestinationPath doesn't check target names against users/groups/roles or commands against notificationCommands.xml. event-severity is type="string" in notifications.xsd with no enumeration, so any value is accepted too. The legacy wizard constrained all of these by dropdown; the REST API now accepts configs that fail silently at delivery time.
  3. DestinationPathManager.saveWithRollback restores by mutating the live map. m_destinationPaths.clear() + putAll(snapshot) exposes the transient empty state that the new initializeDestinationPaths() comment says it swaps references to avoid, and readers hold the Collections.unmodifiableMap view without synchronization. Fix is m_destinationPaths = snapshot;.
  4. The critical-IP check resolves hostnames over DNS. applyPathOutage (NotificationConfigRestService.java:998) validates with InetAddressUtils.addr, which is InetAddress.getByName (IPAddress.java:315). The only guard ahead of it is validateRule("IPADDR IPLIKE " + raw), a filter-syntax check rather than an IP-literal check, so a hostname that parses is silently resolved through a blocking lookup with no timeout and stored.
  5. useBrowserNotifications reconnects forever and can throw on startup. socket.onclose schedules a retry every 5s with no cap and no stop on auth failure or repeated handshake rejection, and new  WebSocket(...) is unguarded, so a site setting opennms.web.base-url without a scheme throws a SyntaxError out of the whenever callback in App.vue.
  6. No security rule for /api/v2/notification-config. GET falls to the /api/v2/** catch-all (ROLE_USER, ROLE_MOBILE), with only the in-code assertAdmin stopping it; /api/v2/groups/** twelve lines away is pinned to ROLE_ADMIN for the same "drives notification routing" reason. The block comment also names /rest/notification-config, which is not where the service lives.
  7. FormField's new label row is app-wide. .form-field__label-row (moved margin plus min-height: 1.25rem) applies to all ~60 FormField consumers, not just the new page.
  8. Lock usage is inconsistent. getServiceNames (:580) is the only endpoint that skips readLock(), and previewPathOutageRule (:970) takes it while only touching the database.

joseanesONMS:
@marshallmassengill all eight addressed in `0defc068002`:

1. **Delete referential check** — deleting a path that notifications reference returns 409 naming them; unreferenced paths delete as before (IT covers both).
2. **Reference validation** — `destinationPath` must exist; targets must be a user, group, on-call role, or email address; commands must exist in `notificationCommands.xml`; `event-severity` must be one of the seven canonical names. ITs cover each rejection and the valid-target trio (user/role/email). `NotificationConfigRestServiceIT` 25/25 locally.
3. **Rollback** — `saveWithRollback` restores by swapping the map reference; no transient empty view.
4. **Critical IP** — parsed with `InetAddresses.forString` (IP literals only, no DNS); hostnames 400.
5. **WebSocket** — bounded backoff (reset on successful open, max 10 attempts, 30s cap) and the constructor is guarded, so a scheme-less base-url can't throw out of the watcher.
6. **Security** — `/api/v2/notification-config/**` pinned to ROLE_ADMIN for every method; the stale `/rest/notification-config` comment corrected.
7. **FormField** — the fixed row height now applies only when a help icon shares the row (template-driven class); plain-label consumers keep their exact prior spacing.
8. **Locks** — `getServiceNames` takes the read lock; `previewPathOutageRule` (database-only) drops it.

Push carries `!smoke`.

joseanesONMS:
@marshallmassengill CI is fully green on `0defc068002` — build, build-ui, integration-test, and all three smoke suites. Ready for your re-review.

marshallmassengill (changes requested):
Couple items to resolve still:

- useBrowserNotifications.ts:107 — Safari/Firefox permission request kills the feature. .catch on the callback-only form throws after started = true and before connect(); Firefox ignores a request with no user gesture. Fix already in the base from #8769 (index.js:236 shim + click-triggered opt-in).
 - useBrowserNotifications.ts:54 — Chrome for Android drops every notice silently. Page-scoped constructor throws, ws.onmessage's catch swallows it, snackbar fallback never runs. Fix also in the base (showNotification path, notification-sw.js, worker-src 'self' all present on develop).
- Delay fields accept free text and blow up in notifd. Three fields unvalidated on both ends: Initial Delay (DestinationPathEditorDialog.vue:31), escalation delay (:101), target Interval (TargetRowEditor.vue:82). TimeConverter.convertToMillis rejects 5min and 1h30m. Worst path is BroadcastEventProcessor:645, which fires after insertNotice, so the notice row persists, no task is scheduled, and the exception escapes into eventd's dispatch.

I'm choosing not to take personal offense that you didn't pull in the fixes I worked so hard on in #8769

joseanesONMS:
@marshallmassengill fair point on #8769, and thank you for it — that groundwork is exactly what made these fixes small. All three are in `9b3c1ddcf2f` (which also merges current develop):

1. **Safari/Firefox** — the composable no longer calls `requestPermission` at all. Your click-triggered opt-in from NMS-20200 owns granting (permission is origin-wide); until granted, every message rides the snackbar. No callback-form `.catch`, no gesture-less request.
2. **Chrome for Android** — delivery now goes through your `notification-sw.js` via `registration.showNotification` (registered from the composable, two-arg `then` like core/web-assets); the page-scoped constructor is a guarded fallback, and every failure path lands in the snackbar instead of being swallowed. Tests cover the worker path, the constructor throw, and the no-request contract.
3. **Delay fields** — initial delay, escalation delay, and target interval are validated against `TimeConverter`'s grammar on both ends: inline errors + save gating in the dialogs, 400 with the accepted grammar from the API. IT covers `5min`/`1h30m`/free text rejected and the full grammar accepted (`NotificationConfigRestServiceIT` 26/26 locally).

Push carries `!smoke`.

marshallmassengill (changes requested):
Looking better but still a few things to address:

  - NotificationConfigRestService.validateDelay accepts values its own error message excludes, and that the UI rejects. It delegates to TimeConverter.convertToMillis, which has no sign or range guard.  Apply the same regex server-side, or reject a non-positive result.
  - swRegistration is captured when register() resolves, not when the worker activates. showNotification() rejects with InvalidStateError while active is null, so notices in the first moments after a cold load are snackbar-only. It degrades safely via the two-arg then, but core/web-assets' withServiceWorker deliberately waits for activated with a 5s timeout and documents why navigator.serviceWorker.ready is the wrong tool.
  - The worker is registered only if permission is already granted at page load. startBrowserNotifications is { once: true } behind a started guard, so a mid-session grant (the opt-in lives on other pages) leaves swRegistration null until reload — on Chrome for Android that means snackbar only.
  - Blank escalation delay passes the client gate but the server rejects it. isValidNotifdDuration('') returns true by design and isValid never checks blankness, while validateDestinationPath does. Clearing the editable select enables Save and yields a 400. The helper's own comment says callers requiring a value should check blank separately; this one doesn't.
  - NoticesTable.vue:241 — detached anchor plus same-tick revokeObjectURL; eventConfigService.saveBlobAsFile is the better in-tree pattern.
  - DestinationPathManager.java:53 — m_destinationPaths is non-volatile while getPath/getPaths are still unsynchronized, so the snapshot swap is unsafely published.
  - NotificationRestService.java:129 — teamUnacknowledgedCount has no distinct() while the list path does, so the bell badge and the page disagree on multi-recipient notices (this pre-dates this PR but it's worth addressing here).

joseanesONMS:
@marshallmassengill all seven addressed in `4a23a71b5a2`:

1. **validateDelay** — the UI's grammar regex now runs server-side before `convertToMillis` (whose `Float.parseFloat` takes signs, NaN and Infinity); IT adds `-5s`, `+5s`, `NaN`, `Infinity`, `1e3s` → 400.
2. **Activation** — the composable now mirrors `withServiceWorker`: the cached registration promise resolves only when the worker is active, with the 5s bounded wait and null on redundancy, then falls back constructor → snackbar. Test covers a worker that activates late.
3. **Mid-session grant** — registration happens lazily per message instead of only at page load, so a grant from the NMS-20200 opt-in on another page starts worker delivery without a reload (the already-granted case still warms up at start). Test flips permission after start and asserts the worker gets used.
4. **Blank escalation delay** — checked client-side now (`escalationDelayError` handles blank, `isValid` uses it), matching the destination-path validator.
5. **CSV export** — `saveBlobAsFile` (exported from eventConfigService) replaces the detached anchor + same-tick revoke.
6. **DestinationPathManager** — `m_destinationPaths` is `volatile` so the reload swap publishes safely to the unsynchronized readers.
7. **teamUnacknowledgedCount** — gains the `distinct()` the list query already had, so the bell badge and page agree on multi-recipient notices.

`NotificationConfigRestServiceIT` 26/26, both `NotificationRestServiceIT`s green vs local Postgres; UI 10/10 on the composable, lint and vue-tsc clean.

marshallmassengill (changes requested):
Some additional blockers:

- DestinationPathManager mutators still mutate the TreeMap in place while readers read it unlocked. addPath/replacePath/removePath do m_destinationPaths.put/remove under synchronized, but getPath/getPaths (line 94/109) read the field with no lock — a data race on a non-thread-safe TreeMap. The new volatile and the comment claiming "a reader holding an earlier unmodifiableMap keeps a consistent snapshot" only hold for the reload swap and the rollback restore, not for the normal mutation path. Applying the change to a copy and publishing the reference on success fixes the race and deletes the snapshot/restore code.
- validateDestinationPathReferences resolves targets from stale cached config. It calls UserFactory.getInstance().getUserNames(), getGroupNames() and getRoleNames() with no update() — while getOnCallRoles, twenty lines above, explicitly calls manager.update() and even comments that "getRoleNames() reads cached state without checking the file" (and testOnCallRolesFollowFileChanges covers it). A group or on-call role added by hand or through the new /api/v2/on-call-roles gets a spurious 400 "Target X is not a user, group, on-call role, or email address" until an unrelated read triggers a reload.
- Notification and destination-path names are only checked for blankness. validateEventNotification, validateDestinationPath, and both editor dialogs (EventNotificationEditorDialog.validateForm, DestinationPathEditorDialog.isValid) accept a name containing /, \ or %; PUT/DELETE .../{name} then encodes it and DefaultHttpFirewall rejects the request, so the entry can never be edited or deleted through the API. adminValidation.ts already exports validateAdminName and isPathAddressable, and the sibling admin services use them for exactly this.

joseanesONMS:
@marshallmassengill all three addressed in `9e5eedf9e17`:

- **Data race** — `DestinationPathManager` is copy-on-write now: each mutator applies its change to a copy, `saveCurrent(copy)` persists it, and the volatile reference is published only on success. The snapshot/restore code is gone; `DestinationPathManagerRollbackTest` still passes since a rejected save simply never publishes.
- **Stale references** — one correction to the diagnosis: `getUserNames()` and `getGroupNames()` already call `update()` internally; only `getRoleNames()` reads cached state. The validator now calls `update()` before it. New IT adds a role to `groups.xml` behind the manager and asserts a path targeting it is accepted straight away.
- **Names** — `validateEventNotification` and `validateDestinationPath` reject `/`, `\`, `%` and control characters (`validateAddressableName`), which also covers renames since those go through the PUT body. Both dialogs use `isPathAddressable` for an inline field error. IT covers all four characters on create for both types, plus a rename to a bad name (400, entry keeps its name). Existing default names contain none of these.

`NotificationConfigRestServiceIT` 28/28 locally, config tests, UI tests, eslint and `vue-tsc` green; push carries `!smoke`.

marshallmassengill (dismissed):
This looks good for me now.  I'm a bit worried about when we introduce these and keeping the homepage static vs allowing customization but we can discuss.  Please wait for @synqotik to look at this before merging.  Also, CI needs to pass though I suspect it's just flaky test.

joseanesONMS:
Merged current develop (38 commits, clean) and added the menu split requested: Administration keeps **Notifications** on the legacy page and gains **Notifications (Preview)** right below it for the Vue page; the bell menu keeps pointing at the preview page. MenuHeaderIT covers both entries. The earlier smoke-test-minion failure was a container step timeout with no failing test.

synqotik (approved):
LGTM! There are a number of UX changes to be made before this is fully released, but let's get this into `develop` so we can test it out and refine the workflow. I will open a separate ticket for the initial changes I think we should make first.

PR #8822, opened 2026-08-27
NMS-20251: bumped netty version

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-20251


marshallmassengill:
@christianpape - can you target this one at f2023?  I suspect it's going to get flagged by a customer and if we can do it on that for PB then we should.

christianpape:
> @christianpape - can you target this one at f2023? I suspect it's going to get flagged by a customer and if we can do it on that for PB then we should.

done.

synqotik (approved):

PR #8823, opened 2026-08-27
NMS-19927: Metadata-Driven Kafka Metric Topic Routing

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-19927

### DIFF
--- a/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/collection/KafkaPersisterActivator.java
+++ b/features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/collection/KafkaPersisterActivator.java
@@ -21,18 +21,26 @@
  */
 package org.opennms.features.kafka.producer.collection;
 
+import java.util.Dictionary;
+import java.util.Hashtable;
+
+import org.opennms.core.mate.api.ContextKey;
+import org.opennms.core.mate.api.EntityScopeProvider;
 import org.opennms.netmgt.collection.api.PersisterFactory;
 import org.opennms.netmgt.dao.api.NodeDao;
 import org.opennms.netmgt.dao.api.ResourceDao;
 import org.opennms.netmgt.dao.api.SessionUtils;
 import org.osgi.framework.BundleActivator;
 import org.osgi.framework.BundleContext;
+import org.osgi.framework.ServiceReference;
+import org.osgi.framework.ServiceRegistration;
 import org.osgi.service.cm.ConfigurationAdmin;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.Dictionary;
-import java.util.Hashtable;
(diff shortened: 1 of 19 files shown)


features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/shell/MetricRoutingTest.java
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+ * Previews the Kafka topic that collected metrics for a given node, interface or service would be
+ * routed to. The counterpart of {@code opennms:metadata-test} for metric routing.
+ */
+@Command(scope = "opennms", name = "kafka-metric-routing-test",
+        description = "Show the Kafka topic that metrics for the given node, interface or service are routed to.")
+@Service
+@SuppressWarnings("java:S106")
+public class MetricRoutingTest implements Action {
```
cgorantla, line 47: Check if this still works if metric routing is disabled.  Users might want to run this command without enabling routing.

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/collection/KafkaPersisterActivator.java
```diff
@@ -44,13 +52,32 @@ public class KafkaPersisterActivator implements BundleActivator {
...
+    private KafkaPersisterFactory kafkaPersisterFactory;
+
     @Override
     public void start(BundleContext context) throws Exception {
         ConfigurationAdmin configAdmin = null;
         Boolean forwardMetrics = false;
         String metricTopic = null;
         boolean disableMetricsSplitting = false;
         String metricFilter = null;
+        boolean metricRoutingEnabled = false;
```
cgorantla, line 78: One thing to note with activator, this might not reload bundle when config gets updated. So if config gets updated after feature resolution, bundle/feature might need to be restarted

cgorantla (commented):
This looks good. Some comments and also this needs to be re-targeted on `foundation-2026`

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/shell/MetricRoutingTest.java
christianpape: This works when metric routing is disabled. In this case, a message appears stating that it is disabled and that the default theme is being used.

features/kafka/producer/src/main/java/org/opennms/features/kafka/producer/collection/KafkaPersisterActivator.java
christianpape: Added a note to the docs.

cgorantla (dismissed):
LGTM!

cgorantla (approved):
LGTM!

PR #8824, opened 2026-08-27
NMS-20154: Migrate the Generate System Report page to Vue UI

joseanesONMS:
Replaces the legacy Generate System Report JSP form with a Vue page backed by new REST endpoints, so report generation no longer depends on the JSP/servlet flow.

The report content is unchanged — the generate endpoint ports the existing `FormatterView` (resolve formatter, run the selected plugins, stream the result) rather than adding anything to what a report contains.

- New `GET /api/v2/system-report/plugins` and `/formatters` expose the visible plugins and formatters as JSON; all three endpoints are admin-gated in `applicationContext-spring-security.xml`.
- New `POST /api/v2/system-report/generate` streams the report as a file attachment and returns 400 for an unknown/non-streaming formatter or an empty plugin set.
- The Vue page at `/system-report` (Onms-XXX components: `OnmsCard`/`OnmsCheckbox`/`OnmsSelect`/`OnmsInputText`/`OnmsButton`) offers plugin checkboxes with an All toggle, a report-type select, an optional file name, and Generate; the route is admin-gated.
- Generate calls the endpoint via axios and saves the response with `composables/useDownload` (matching `SnmpConfigUploadDownloadTab`); a failed generation is surfaced to the user instead of silently doing nothing.
- The default report type stays `text`, the file name is sanitized the way the server sanitizes it, and the Support menu entry is repointed to the new page.
- `MenuHeaderIT` targets the new page's breadcrumb link and `OnmsButton`; covered by a `SystemReportRestService` unit test and a `SystemReport.vue` component test.
- `opennms-openapi-docs` now declares the system-report feature at provided scope so its swagger scan can reflect over `SystemReportRestService` (the CI failure).

---
Supersedes #8735 (same commits, `452ed818e80f`); the branch was renamed with a `-smoke` suffix so CircleCI runs the smoke tests before merge.


marshallmassengill (approved):
I think this is good now.

PR #8826, opened 2026-08-27
NMS-20271: Intermittent deadlock in DataBlocksOffHeapQueue during memory to off-heap transition

mershad-manesh:
Discovered while investigating a flaky test case.

### All Contributors

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20271

### DIFF
--- a/core/ipc/sink/off-heap/src/main/java/org/opennms/core/ipc/sink/offheap/DataBlocksOffHeapQueue.java
+++ b/core/ipc/sink/off-heap/src/main/java/org/opennms/core/ipc/sink/offheap/DataBlocksOffHeapQueue.java
@@ -293,15 +293,14 @@ public class DataBlocksOffHeapQueue<T> implements DispatchQueue<T> {
         var head = mainQueue.peek();
         if (head.size() > 0) {
             data = head.dequeue();
-            if (headTailEqual) {
-                tailLock.unlock();
-            } else {
-                removeHeadBlockIfNeeded();
-            }
+        }
+        if (headTailEqual) {
+            tailLock.unlock();
         } else {
-            if (headTailEqual) {
-                tailLock.unlock();
-            }
+            // Also runs when the head was already empty: a block drained while it was still the
+            // tail is never evicted at drain time, so without this it stays head forever and
+            // every dequeue() spins on it.
+            removeHeadBlockIfNeeded();
         }
         return data;
     }
--- a/core/ipc/sink/off-heap/src/main/java/org/opennms/core/ipc/sink/offheap/DataBlocksOffHeapQueue.java
+++ b/core/ipc/sink/off-heap/src/main/java/org/opennms/core/ipc/sink/offheap/DataBlocksOffHeapQueue.java
@@ -69,6 +69,10 @@ public class DataBlocksOffHeapQueue<T> implements DispatchQueue<T> {
 
     private final Lock headLock = new ReentrantLock(true);
     private final Lock tailLock = new ReentrantLock(true);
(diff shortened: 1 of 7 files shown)


marshallmassengill:
I can't recreate this as is.  I sent this over to @mershad-manesh to see what I managed to get after some analysis: https://github.com/OpenNMS/opennms/compare/mem/NMS-20271...marshallmassengill:opennms:mm/NMS-20271-smoke?expand=1

marshallmassengill (dismissed):
I'm good with this... but I think some others should review as well.

core/ipc/sink/off-heap/src/main/java/org/opennms/core/ipc/sink/offheap/OffHeapDataBlock.java
```diff
@@ -186,25 +186,20 @@ private void flushToDisk() {
...
             }
         });
-        diskLock.unlock();
     }
 
     // make sure data is in memory queue
     public synchronized void enableQueue() throws ReadFailedException, InterruptedException {
+        // Wait for an in-flight flush before taking diskLock. The flush task acquires diskLock
+        // itself, so holding it while waiting on the future deadlocks the two against each other.
+        while (future != null && !future.isDone()) {
```
christianpape, line 202: The new while (future != null && !future.isDone()) this.wait(10) is unbounded, and the flush future runs on the static fixed pool of 10 that also runs notifyNextDataBlock() prefetch tasks — which call enableQueue() and block in that same loop. Ten prefetch tasks waiting on flushes queued behind them in the same pool means no flush ever runs, and any consumer touching such a block blocks forever inside dequeue() holding headLock. Same silent wedge the PR is fixing, relocated. Needs a separate pool for prefetch, or no blocking on pool threads.

core/ipc/sink/off-heap/src/main/java/org/opennms/core/ipc/sink/offheap/DataBlocksOffHeapQueue.java
```diff
@@ -289,15 +293,14 @@ private Map.Entry<String, T> readData() throws ReadFailedException, InterruptedE
...
+        if (headTailEqual) {
+            tailLock.unlock();
         } else {
-            if (headTailEqual) {
-                tailLock.unlock();
-            }
+            // Also runs when the head was already empty: a block drained while it was still the
+            // tail is never evicted at drain time, so without this it stays head forever and
+            // every dequeue() spins on it.
+            removeHeadBlockIfNeeded();
```
christianpape, line 303: removeHeadBlockIfNeeded() is now reachable for a block whose flush is racing. OffHeapDataBlock.size() reads queue/offHeapQueueSize, written by the flush task under diskLock only while readers hold the object monitor — no happens-before, neither field volatile. A reader seeing queue == null with the initial offHeapQueueSize == -1 gets size() == -1, and the new call then evicts a head block still on disk, dropping a whole batch. Previously that path only stalled. Making both fields volatile closes it.

christianpape (changes requested):

core/ipc/sink/off-heap/src/main/java/org/opennms/core/ipc/sink/offheap/OffHeapDataBlock.java
mershad-manesh: Thanks for the feedback, I have attempted to resolve it in the latest commit. In addition, I'm adding two tests

christianpape (approved):
LGTM.

PR #8828, opened 2026-08-28
NMS-20269: Expose event-conf source order and per-event order on Rest and UI

cgorantla:
Add `eventconf_events.event_order`  so the evaluation order of events within a source is explicit;
Expose eventOrder in the event DTO  and show order(evaluationOrder) on sources and 
order for each event on a given source

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20269


cgorantla:
Added order in both event sources and for the events in a given source

Screenshot 2026-08-28 at 2 03 24 PM
Screenshot 2026-08-28 at 2 02 59 PM

marshallmassengill (changes requested):
These two seem to be worth addressing:

- Give the pessimistic locks a timeout. new LockOptions(LockMode.PESSIMISTIC_WRITE) defaults to WAIT_FOREVER in nextEventOrder and lockForUpdate; combined with REQUIRES_NEW on persistEventConfFile and reorderSourcesFromEventConf, any future caller already holding the lock hangs an HTTP worker permanently with no deadlock detection.
- Enforce the catch-all pin where fileOrder is allocated. createOrUpdateSource and createEventConfSource hand MAX+1 to a re-created opennms.catch-all.events, which then matches every event first. Pre-existing, but this PR is where the invariant gets named.

The rest of these may be as well though I don't think any of these are blockers - more just cleanup and getting tests dealt with:

- Add UNIQUE (source_id, event_order) DEFERRABLE INITIALLY DEFERRED, or say why events differ from sources. Nothing catches it if one appears.
- Flip the dedupe tie-break to id ASC. As written, the newer of two colliding sources ends up evaluated later, contradicting MAX+1. 
- Make a failed reorderSourcesFromEventConf visible in the response status, not just errorList under a 200. Also pick an error key that cannot collide with a file basename of eventconf.
- Change defaultNullValue="0" to 1, or remove it. It writes a value outside the documented 1..N contract.
- Remove the dead private EventConfPersistenceService.saveEventsToDatabase(); it assigns fileOrder = 1, 2, 3… and would now violate the new unique constraint.
- Fold the duplicated sortOrder assignment in EventConfEventDaoHibernate.findBySourceId into the initial one, and rename the direction parameter now that eventOrder is also a sort field.
- Remove the unused Optional, LinkedHashSet, HashMap imports from EventConfRestService.
- In EventConfEventOrderConcurrencyIT, stamp firstCommittedAt after the transaction returns, and capture/rethrow the worker threads' throwables so failures surface at their origin.
- Assert on the exception type or SQLState in testFileOrderMustBeUnique rather than on message text containing the constraint name.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/EventConfPersistenceService.java
christianpape, line 256: saveEventsToDatabase() is the only caller of the new if (getFileOrder() > 0) branch and assigns 1,2,3… in load order. That collides with the values the migration pins, so with the new uk_eventconf_sources_file_order it now hard-fails at commit instead of silently overwriting — and it inverts precedence. Unreachable today (private, no callers), but it's the system-migration path: delete it or switch to nextFileOrder().

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/EventConfPersistenceService.java
```diff
@@ -163,7 +223,13 @@ private EventConfSource createOrUpdateSource(final EventConfSourceMetadataDto ev
             source = new EventConfSource();
             source.setCreatedTime(eventConfSourceMetadataDto.getNow());
         }
-        source.setFileOrder(eventConfSourceMetadataDto.getFileOrder());
+        // fileOrder <= 0 means "decide here": an existing source keeps the position it has in this
+        // transaction's snapshot, a new one is allocated MAX+1 under the allocation lock
+        if (eventConfSourceMetadataDto.getFileOrder() > 0) {
```
christianpape, line 228: createOrUpdateSource grabs nextFileOrder() as the first statement of the REQUIRES_NEW transaction. That's a transaction-scoped row lock on the single EVENTCONF_SOURCES_ACCESS row, held across the parse and the full batched saveAll. Uploading one large vendor trap file blocks every other source creation (plugin sync, programmatic saves, parallel uploads) for the whole insert — no lock_timeout. Allocate fileOrder last, or from a sequence / INSERT … RETURNING.

opennms-dao/src/main/java/org/opennms/netmgt/dao/hibernate/EventConfSourceDaoHibernate.java
christianpape, line 173: filterEventConfSource catches Exception at DEBUG and still returns a non-zero totalRecords with an empty list. Since evaluationOrder is now the UI's default sort key (ui/src/stores/eventConfigStore.ts:22) and it's a formula rendered as a correlated subquery in ORDER BY, any failure there shows an empty Sources table with a paginator claiming N rows. Log WARN, return totalRecords: 0.

cgorantla:
@christianpape @marshallmassengill  new commits should resolve this.  Also added older eventconf.xml  with all the event sources to etc/examples/events.  It will facilitate single update within the folder with order.

christianpape (dismissed):
LGTM.

marshallmassengill (changes requested):
Some things to look at:

- Looks like CI is failing. EventConfRestIT.testUploadAllEventConfFilesAtOnce: expected:<233> but was:<232> — the test globs *.xml in etc/examples/events and asserts every file lands in successList, but the new eventconf.xml is deliberately consumed by parseEventConfOrder and never appears there.
- 36.0.4-eventconf-sources-name-dedupe and -name-unique are dead changesets. uk_eventconf_sources_name already exists from 35.0.0-eventconf-sources-name-unique (verified on the live 36.0.3-SNAPSHOT DB).
- enableDisableConfSourcesEvents → EventConfEventDaoHibernate.updateEventEnabledFlag calls no lock helper, so no SET LOCAL lock_timeout is issued for its transaction and its UPDATE … WHERE source_id = ? waits indefinitely behind an upload — with no global lock_timeout/statement_timeout underneath it, all 0.
- A malformed eventconf.xml still returns 200. parseEventConfOrder swallows the parse failure and returns null, leaving orderApplied true; only a failure inside reorderSourcesFromEventConf produces the new 500 (applies to API and UI).

Did my best to validate all of those on my running dev instance.  At least two of them I think need to be addressed so you may want to look at the others.

cgorantla:
@marshallmassengill Latest updates should handle these.

marshallmassengill (dismissed):
LGTM!

marshallmassengill (approved):
LGTM!

PR #8830, opened 2026-08-31
NMS-4471: Daemon Management page with per-daemon config reload

joseanesONMS:
Adds a "Daemon Management" page under Tools (NMS-4471) that reloads one daemon's configuration at a time.

Per the scope discussion: reload-only, via the existing events API — no new backend; a daemon status API and surfacing status on this page are follow-up tickets, and start/stop needs backend groundwork first.

- The page lists all 22 daemons that implement a `reloadDaemonConfig` handler, with what each reload applies; handlers match `daemonName` case-insensitively.
- Reload publishes `uei.opennms.org/internal/reloadDaemonConfig` with the `daemonName` parm through `POST /rest/events` — the same event `opennms:reload-daemon` sends — behind a confirmation dialog.
- The page states that the outcome arrives asynchronously as a `reloadDaemonConfigSuccessful`/`Failed` event; the About dialog explains the mechanism.
- Menu entry under Tools (`ROLE_ADMIN`) in all four templates; the route is admin-gated; `MenuHeaderIT` covers the new entry.
- Uses Onms-XXX components on the standard container/table-card layout; eslint/`vue-tsc` clean; tests cover the event payload, the daemon set, and the confirmation flow.
- Verified end to end on a running instance: Pollerd, Provisiond, Vacuumd, Discovery, Ackd, and Translator answer `reloadDaemonConfigSuccessful`, including lower/upper-cased names.

### DIFF
--- a/ui/src/components/DaemonManagement/DaemonManagementAbout.vue
+++ b/ui/src/components/DaemonManagement/DaemonManagementAbout.vue
@@ -1,0 +1,22 @@
+<template>
+  <div data-test="daemon-management-about">
+    <p>
+      Most OpenNMS daemons can re-read their configuration while running. Reloading applies changes made to a
+      daemon's configuration files (directly or through the configuration pages) without restarting OpenNMS or
+      interrupting the other daemons.
+    </p>
+    <p>
+      A reload request is delivered as the <em>uei.opennms.org/internal/reloadDaemonConfig</em> event and handled
+      asynchronously: the daemon answers with a <em>reloadDaemonConfigSuccessful</em> or
+      <em>reloadDaemonConfigFailed</em> event, which you can find in the event list. A failed reload leaves the
+      previous configuration in effect.
+    </p>
+    <p>
+      Only the daemons listed here support config reload. Starting and stopping daemons, and live daemon status,
+      are not yet available from this page.
+    </p>
+  </div>
+</template>
+
+<script setup lang="ts">
+</script>
--- a/ui/src/components/DaemonManagement/DaemonManagementAbout.vue
+++ b/ui/src/components/DaemonManagement/DaemonManagementAbout.vue
@@ -12,8 +12,10 @@
       previous configuration in effect.
     </p>
     <p>
-      Only the daemons listed here support config reload. Starting and stopping daemons, and live daemon status,
-      are not yet available from this page.
+      This page lists the daemons that handle the reload event. Some specialized components — provisioning
+      adapters (addressed as <em>Provisiond.&lt;AdapterName&gt;</em>) and correlation engines — also accept
+      targeted reloads but are configured through their own subsystems and are not listed here. Starting and
+      stopping daemons, and live daemon status, are not yet available from this page.
     </p>
   </div>
 </template>
--- a/ui/src/components/DaemonManagement/DaemonManagementAbout.vue
+++ b/ui/src/components/DaemonManagement/DaemonManagementAbout.vue
@@ -1,21 +1,20 @@
 <template>
   <div data-test="daemon-management-about">
     <p>
-      Most OpenNMS daemons can re-read their configuration while running. Reloading applies changes made to a
-      daemon's configuration files (directly or through the configuration pages) without restarting OpenNMS or
-      interrupting the other daemons.
+      Reloading makes a daemon re-read its configuration files without restarting OpenNMS or interrupting the
+      other daemons. This page lists the daemons that handle the reload event.
     </p>
     <p>
-      A reload request is delivered as the <em>uei.opennms.org/internal/reloadDaemonConfig</em> event and handled
-      asynchronously: the daemon answers with a <em>reloadDaemonConfigSuccessful</em> or
-      <em>reloadDaemonConfigFailed</em> event, which you can find in the event list. A failed reload leaves the
+      The request is sent as the <em>uei.opennms.org/internal/reloadDaemonConfig</em> event and handled
+      asynchronously. The daemon answers with a <em>reloadDaemonConfigSuccessful</em> or
+      <em>reloadDaemonConfigFailed</em> event, both visible in the event list. A failed reload leaves the
       previous configuration in effect.
     </p>
     <p>
-      This page lists the daemons that handle the reload event. Some specialized components — provisioning
-      adapters (addressed as <em>Provisiond.&lt;AdapterName&gt;</em>) and correlation engines — also accept
-      targeted reloads but are configured through their own subsystems and are not listed here. Starting and
-      stopping daemons, and live daemon status, are not yet available from this page.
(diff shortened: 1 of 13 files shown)


marshallmassengill (changes requested):
One major blocker and some minor stuff:

The About text "Only the daemons listed here support config reload" is wrong: Provisiond, Reportd, Ticketd, Translator, Vacuumd, Enlinkd, Bsmd... all handle reloadDaemonConfig.... and I think there may be others.  This needs real attention.

1. menu-template-legacy.json gets no entry, so the page is unreachable from the legacy menu — SCV, File Editor, Logs, Device Configs and External Requisitions all appear there. The PR body's "all three templates" is really four.
2. RELOADABLE_DAEMONS omits Discovery, a DaemonReloadEnum member that ReloadDaemonIT proves reloads successfully (features/discovery/.../Discovery.java:191) — yet the comment says "Mirrors DaemonReloadEnum"  and the test pins the 8-name list as if complete. 
3. "casing matters" / "reload matching is case-sensitive per daemon" is false - it's something that I think is in the docs somewhere but I don't think it actually applies anymore and this should be validated.  Better yet, it's not needed on the page.
4. { label: 'Home', to: '/' } diverges from all nine peer containers, which use menuStore.mainMenu.homeUrl with isAbsoluteLink: true and a final to: '#' crumb; as written Home lands on the SPA root, not the OpenNMS home page.
5. TableCard is padding: 25px 0px, and every other consumer adds padding: 25px on the card itself; this page wraps it in an unpadded div, so the title and table sit flush against the card's left and right edges.
6. New code calls the legacy useSnackbar adapter, whose own comment says new code should call useOnmsToast() directly (NMS-20029).
7. catch (_err) drops the server's message; POST /rest/events returns 400 with the validation reason, which is the only actionable detail a failed request carries.
8. Smaller deviations from the ManageGroups pattern: .card-title at 1.25rem vs 1.1rem, no page-title heading, and table markup in the container rather than a components/DaemonManagement/DaemonManagementTable.vue. A reload rebuilds Pollerd's whole schedule, so a confirmation step is worth considering.

Throw a screenshot on the PR if you can as well.

joseanesONMS:
@marshallmassengill all items addressed in `bd0e7c68d0d`:

1. **Daemon list (the blocker)** — the page now lists all 22 daemons that implement a `reloadDaemonConfig` handler; `DaemonReloadEnum` turned out to be only the Karaf command's completion set. Verified live: Provisiond, Vacuumd, Discovery, Ackd, and Translator all answer `reloadDaemonConfigSuccessful`. The About text no longer claims the list is exhaustive and notes that provisioning adapters (`Provisiond.<Adapter>`) and correlation engines take targeted reloads as components.
2. **Legacy menu** — `menu-template-legacy.json` gets the entry; PR body corrected to four templates.
3. **Casing** — you were right: every handler compares with `equalsIgnoreCase` (checked each handler site, confirmed live with `pollerd` and `PROVISIOND`), so the caveat is gone from the page and the code.
4. **Breadcrumbs** — Home now uses `menuStore.mainMenu.homeUrl` with `isAbsoluteLink` and a final `#` crumb like the peer containers.
5. **Padding** — the card carries `padding: 25px` itself now.
6. **Toast** — `useOnmsToast()` directly.
7. **Errors** — a failed publish surfaces the server's plain-text validation reason in the toast.
8. **Pattern** — table extracted to `components/DaemonManagement/DaemonManagementTable.vue`, 1.1rem card title, `h1.page-title` on the container (`MenuHeaderIT` updated), and reloads sit behind a confirmation dialog that mentions the schedule rebuild.

Screenshot to follow shortly.

joseanesONMS:
Screenshot 2026-09-01 at 12 15 39
Screenshot 2026-09-01 at 12 15 21
Screenshot 2026-09-01 at 12 15 10
Screenshot 2026-09-01 at 12 14 58
Screenshot 2026-09-01 at 12 14 46

@marshallmassengill @synqotik

ui/src/components/DaemonManagement/DaemonManagementAbout.vue
```diff
@@ -0,0 +1,24 @@
...
+      asynchronously: the daemon answers with a <em>reloadDaemonConfigSuccessful</em> or
+      <em>reloadDaemonConfigFailed</em> event, which you can find in the event list. A failed reload leaves the
+      previous configuration in effect.
+    </p>
+    <p>
+      This page lists the daemons that handle the reload event. Some specialized components — provisioning
+      adapters (addressed as <em>Provisiond.&lt;AdapterName&gt;</em>) and correlation engines — also accept
+      targeted reloads but are configured through their own subsystems and are not listed here. Starting and
+      stopping daemons, and live daemon status, are not yet available from this page.
+    </p>
```
marshallmassengill, line 18:
```suggestion
    <p>
      Reloading makes a daemon re-read its configuration files without restarting OpenNMS or interrupting the
      other daemons. This page lists the daemons that handle the reload event.
    </p>
    <p>
      The request is sent as the <em>uei.opennms.org/internal/reloadDaemonConfig</em> event and handled
      asynchronously. The daemon answers with a <em>reloadDaemonConfigSuccessful</em> or
      <em>reloadDaemonConfigFailed</em> event, both visible in the event list. A failed reload leaves the
      previous configuration in effect.
    </p>
```

marshallmassengill (changes requested):
This one needs to be blocked for now because of downstream stuff where we need to fix some issues with how the reloading works.

- Several of these don't actually do what they should be doing:
Enlinkd (features/enlinkd/daemon/.../EventProcessor.java:119 just calls m_linkd.reloadConfig()), Scriptd (opennms-services/.../scriptd/Executor.java:354), Threshd (.../threshd/ThresholdingServiceImpl.java:219) and Ticketd (.../ticketd/TroubleTicketer.java:248) emit no reloadDaemonConfigSuccessful/Failed. 
- Alarmd's "northbound rules" is wrong. Alarmd.handleReloadEvent (opennms-alarms/daemon/.../Alarmd.java:83) forwards to NorthbounderManager.handleReloadEvent, which matches daemonName.contains(nbi.getName()) — case-sensitive substring, so Alarmd matches no NBI. The reload only does m_droolsAlarmContext.reload().
- Provisiond's "detection settings" is wrong. The handler (Provisioner.java:753) only calls m_importSchedule.rebuildImportSchedule().

Let me dig into this a bit more and get bugs opened first for some of this.  This is looking promising though.

joseanesONMS:
@marshallmassengill the three things in this diff are corrected in `aad017f197c`: the About text takes your suggested wording, Alarmd now reads "Alarm correlation rules (Drools)", and Provisiond "Requisition import schedule". I've left the four daemons that emit no reload response event (Enlinkd, Scriptd, Threshd, Ticketd) untouched pending the bugs you're opening — happy to flag them in the table once those land. Push carries `!smoke`.

PR #8831, opened 2026-09-01
NMS-20273: Open side menu submenus on hover, in both rail states

synqotik:
# NMS-20273: Open side menu submenus on hover, in both rail states

Hovering a top-level entry in the side menu now flies its submenu out, instead of requiring a
click. This applies whether the rail is collapsed or pinned open.

---

## Background

The rail used to expand on hover *and* open its flyouts (NMS-18287). NMS-20167 removed both,
because the combination was jarring, leaving click-to-open. That turned out to be dialed back too
far: the jarring part was the rail resizing from under the pointer — reflowing the page and moving
every icon the user was aiming at — not the flyouts themselves.

So the flyouts come back, and the rail's width does not move. `isPinned` is now the only thing that
sets it, and only the toggle button and `Ctrl+\` change `isPinned`.

## Behavior

- Hovering a top-level entry opens its submenu after a **150ms dwell**. The rail keeps its width.
- Once the first flyout is open the rail is in hover mode: moving between entries switches flyouts
  **instantly**, with no further delay — the macOS menu-bar feel.
- Leaving the rail closes any open flyout after a **200ms grace**, and leaves hover mode, so
  re-entering needs a fresh dwell.
- Identical in both rail states. Click-to-open still works and is unchanged; so is keyboard
  navigation.

The dwell is the point of the design. Opening on bare `mouseenter` fires a flyout at every entry
the pointer merely crosses on its way somewhere else, which is the same class of noise the old
hover mode had. 150ms is long enough to ignore a pass-through and short enough to read as instant
on a deliberate hover.

The close grace matters because a flyout is not always beside its own entry: `positionFlyouts`
clamps tall flyouts to the rail's vertical band, so reaching one can take the pointer briefly
outside the rail. Closing on the first `mouseleave` would cut that travel off.

## Tooltips

While collapsed, top-level entries had a label tooltip, since their labels are hidden. Now:

- Entries **with** a submenu no longer show one.
- Entries **without** a submenu (Topology Map, Geographical Map — direct links, nothing to fly out)
  keep theirs, and its `showDelay` drops from 300ms to the same 150ms as the flyout, so every
  top-level entry responds on the same beat regardless of which kind it is.
- Expanded, labels are visible and tooltips stay suppressed, as before.

## Notes for reviewers

- **This reaches into TieredMenu internals** — the `dirty` flag, and a synthetic event aimed at a
  PrimeVue-owned element. A PrimeVue upgrade could change either. It is contained to this one
  component, which already carries a comment explaining why TieredMenu is used unwrapped rather
  than behind an `onms-ui` seam component, now extended to cover this. The alternative — calling
  `onItemChange()` with a processed item pulled off the instance — couples to more internals, not
  fewer, and skips PrimeVue's own handler.
- **An expanded-rail flyout now closes when the pointer leaves the rail**, where before it stayed
  until an outside click. Deliberate: the interaction should not differ between the two states.
- **Touch is unaffected.** There is no hover to dwell on, and click-to-open is untouched.
- The two timing constants are the only tuning knobs; both are named and commented at their
  definitions.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20273

### DIFF
--- a/ui/src/components/Menu/SideMenu.vue
+++ b/ui/src/components/Menu/SideMenu.vue
@@ -116,6 +116,45 @@ const topPanels = computed<OnmsMenuItem[]>(() => {
 
 const tieredMenuRef = ref<any>(null)
 
+// Close any open flyout and drop out of hover mode, so the next hover has to
+// dwell again.
+//
+// With nothing open, clearing `dirty` is the whole job (hide() would have done
+// it as a side effect): TieredMenu leaves it set after hovering an entry that
+// has no submenu, so skipping this would keep the rail in hover mode with no
+// flyout to show for it, and the next hover would open instantly with no dwell.
+//
+// hide() also resets focusedItemInfo to index -1. That is right for a flyout
+// the pointer merely hovered open — hover never focuses the menubar, so there
+// is no position to lose — but not for one the user clicked or tabbed into,
+// where it would discard their place in the menu. So when the menu actually
+// holds focus, put the focused index back on the root entry that was open,
+// which is where TieredMenu itself lands after closing a submenu (see its
+// onItemClick). A deeper position is not restored as-is: it points into a
+// submenu that is now closed, and TieredMenu would resolve the stale index
+// against the root list.
+const closeFlyouts = () => {
+  const menu = tieredMenuRef.value
+
+  if (!menu) {
+    return
+  }
+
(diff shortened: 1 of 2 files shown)


mershad-manesh (approved):
LGTM

PR #8834, opened 2026-09-01
NMS-20283: Count notifications, not deliveries, in the notification s…

marshallmassengill:
/notifications/summary derived userUnacknowledgedCount, teamUnacknowledgedCount and the newest-notifications list from a criteria join to usersNotified, and countMatching counts join rows. A notification carries one usersNotified row per notification method, so a user notified by both email and pager was counted twice and appeared twice in the list. On a real install the personal count came back larger than the total unacknowledged count, and the header badge disagreed with the notification list it links to.

This fixes that.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20283


christianpape:
Maybe I’m wrong, but wouldn’t this be solved by just adding .distinct() to the three criterias in NotificationRestService.getInfo() line 124, 129 and 137?

marshallmassengill:
> Maybe I’m wrong, but wouldn’t this be solved by just adding .distinct() to the three criterias in NotificationRestService.getInfo() line 124, 129 and 137?

Yes... yes it does.  Should be a lot simpler now.

christianpape (approved):
LGTM.

PR #8835, opened 2026-09-01
NMS-20280: Time support for OnmsDatePicker, and TimeControls fixes

synqotik:
# NMS-20280: Time support for `OnmsDatePicker`, and `TimeControls` fixes

Adds time-of-day support to the `OnmsDatePicker` seam wrapper, then uses it to
replace the fake time picker in `TimeControls` — which turned out to have a
silent 13-hour bug — and cleans up several long-standing annoyances in that
component.

## `OnmsDatePicker` (`onms-ui` component seam layer)

You can use `timeOnly` to show only time controls; `showTime` to include time controls under the date picker; or just display the date picker.

Now declares the following props.

| prop | type | default |
| --- | --- | --- |
| `showTime` | `boolean` | `false` |
| `timeOnly` | `boolean` | `false` |
| `showSeconds` | `boolean` | `false` |
| `hourFormat` | `OnmsHourFormat` (`'12' \| '24'`) | `'24'` |
| `stepHour` / `stepMinute` / `stepSecond` | `number` | `1` |
| `placeholder` | `string` | — |
| `minDate` / `maxDate` | `Date \| null` | — |
| `inputId` | `string` | — |

Plus new `show` / `hide` emits reporting overlay visibility.

Notes for reviewers:

- **No `OnmsTimePicker` is needed.** `timeOnly` works on its own — PrimeVue
  gates the time panel on `showTime || timeOnly` and the date panel on
  `!timeOnly`.
- **Every default matches PrimeVue's own**, so this changes no existing call
  site's rendering. These are not "notable deviations" in the README sense.
- `hourFormat` is typed as a new owned `OnmsHourFormat` in
  `packages/onms-ui/src/types.ts`, not PrimeVue's `HintedString<'12' | '24'>` —
  seam rule 4 (no PrimeVue types in a public signature).
- `minDate` / `maxDate` accept `Date | null` and collapse `null` to `undefined`
  before reaching PrimeVue, which types them `Date | undefined` and treats any
  non-`undefined` value as a live bound. This keeps a `?? undefined` out of
  every call site that holds its bounds in a nullable ref.
- `inputId` lands on the rendered `<input>`, not the wrapper, so a `FormField`
  label's `for` can reach the real control — matching what `OnmsSelect`,
  `OnmsInputNumber` and the other input wrappers already declare.
- These props were previously reachable only by Vue attribute fallthrough,
  which seam rule 2 classes as unsupported. Declaring them makes the API real.

## `TimeControls`

**Moved to `components/Common`**, along with `utils/timeRangeOptions.ts` —
`AdhocGraphBuilder.vue` and `adhocUrlState.ts` were already importing that out
of `components/Resources/`. Both moves are recorded as git renames.

### Bug fix: custom ranges were silently shifted 13 hours

The old UI was a date picker plus a separate whole-hour `OnmsSelect`, and
`applyCustomTime` **added** the select's value to the picked date. The select
defaulted to `1 PM`, so unless the user touched both selects, every custom range
came out 13 hours off. The test that caught it:

```
expected startTime 1788330600, got 1788283800   // Δ = 46800s = 13h
```

One `showTime` picker per range end replaces the pair. Ranges are now
minute-precise instead of whole-hour. `HOUR_OPTIONS` becomes dead and is
removed with its test.

### Clicking the preset list while date picker open no longer closes the popover

Previously, clicking the preset list ("Last hour", …) while a date picker
overlay was open selected a preset *and* closed the popover. Now that click only
dismisses the picker, keeping its selection, and leaves the popover open.

### Range validity

An explicit ordering check, **not** the `:minDate` / `:maxDate` cross-wiring. PrimeVue does not enforce those strictly enough to be a correctness guarantee. The Apply button is disabled unless
`end > start`, the End field explains why, and `applyCustomTime` re-checks rather than trusting the button's disabled state.

### Labels and accessibility

- **New optional `label` prop** renders a bold label beside the trigger button.
  The **Resource Graphs page gains the "Time Range:" label it was missing**, and
  `AdhocChartToolbar` drops its local `<span>` + CSS in favor of the prop.
- The trigger button is named by `"<label id> <own id>"`, so its accessible name
  is "Time Range: **Last day**" — label *and* current selection.
- The Start and End `FormField`s pass a `for` matching each picker's `inputId`,
  so those labels are associated with their inputs for the first time.

### Layout

- Dropped `.wrapper`'s fixed `height: 70px` (a 70px box around a 35px button) so
  the label and button text align on a baseline.
- "Apply custom time" is now `variant="ghost"`, with `1em` between the Start,
  End and Apply blocks.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20280

### DIFF
--- a/ui/packages/onms-ui/README.md
+++ b/ui/packages/onms-ui/README.md
@@ -269,6 +269,42 @@ tech, and a `tooltip`-only button is named from the tooltip text. Positioning
 modifiers (`v-onms-tooltip.top`) have no prop equivalent — a call site needing
 one keeps using the directive.
 
+## OnmsDatePicker: dates, times, and time-only (NMS-20280)
+
+`OnmsDatePicker` covers all three shapes, so there is no `OnmsTimePicker`:
+
+| prop | type | default | effect |
+| --- | --- | --- | --- |
+| `showTime` | `boolean` | `false` | adds hour/minute spinners below the calendar |
+| `showSeconds` | `boolean` | `false` | adds a second spinner (requires `showTime` or `timeOnly`) |
+| `timeOnly` | `boolean` | `false` | hides the calendar entirely — a pure time picker |
+| `hourFormat` | `OnmsHourFormat` (`'12' \| '24'`) | `'24'` | clock convention |
+| `stepHour` / `stepMinute` / `stepSecond` | `number` | `1` | spinner increments |
+
+`timeOnly` needs no companion `showTime`: PrimeVue gates the time panel on
+`showTime || timeOnly` and the date panel on `!timeOnly`.
+
+`hourFormat` is typed as the owned `OnmsHourFormat` rather than PrimeVue's
+`HintedString<'12' | '24'>`, per rule 4. Every default above matches installed
+primevue@4.5.5's own, so these are not "notable deviations" — declaring them
+changed no existing call site's rendering. `placeholder` / `minDate` / `maxDate`
+are declared for the same reason: passing them by attribute fallthrough happens
+to reach PrimeVue, but rule 2 classes that as unsupported.
+
+Beyond `update:modelValue`, the wrapper emits `show` / `hide` when the overlay
+panel opens and closes. `TimeControls` needs them: PrimeVue dismisses the panel
+from a document-level **mousedown** listener, so a consumer that wants to tell
+"this click dismissed a picker" from "this click chose something else" has to
+sample overlay visibility on mousedown. Without these emits its only route is
+reaching into PrimeVue internals.
+
+`minDate` / `maxDate` accept `Date | null` and collapse `null` to `undefined`
+before reaching PrimeVue, which types them `Date | undefined` and treats any
+non-undefined value as a live bound. That keeps a `?? undefined` out of every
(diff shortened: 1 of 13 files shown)


joseanesONMS:
Screenshot 2026-09-02 at 09 13 12
Screenshot 2026-09-02 at 09 13 05
Screenshot 2026-09-02 at 09 12 53
Screenshot 2026-09-02 at 09 12 31
Screenshot 2026-09-02 at 09 12 22
@synqotik Users modified using this.

joseanesONMS:
Looks good — One minor thing to consider, non-blocking:

`openPickerCount--` on `hide` can in principle go negative if a hide ever fires without a matched show (unmount ordering, or a future third picker). Harmless today since only `> 0` is read, but clamping (`Math.max(0, n - 1)`) would make the latch immune to it.

joseanesONMS (dismissed):

synqotik:
> Looks good — One minor thing to consider, non-blocking:
> 
> `openPickerCount--`

Fixed.

joseanesONMS (approved):

PR #8836, opened 2026-09-01
NMS-19840: Kafka producer json metric payloads

dino2gnt:
When enabled, the metrics persister encodes CollectionSets with protobuf JsonFormat instead of protobuf binary. Message splitting measures the encoded payload size, serializes once per message, and won't recurse forever on a single oversized attribute. Adds the gson bundle the wrapped protobuf-java-util requires, puts the opennms-kafka-producer feature under karaf verify, and documents the the option.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-19840

### DIFF
--- a/container/features/pom.xml
+++ b/container/features/pom.xml
@@ -530,6 +530,7 @@
                           <feature>opennms-timeseries-api</feature>
                           <feature>opennms-timeseries-api</feature>
                           <feature>opennms-grpc-exporter</feature>
+                          <feature>opennms-kafka-producer</feature>
                           <feature>vaadin</feature>
 
                           <!-- minion features -->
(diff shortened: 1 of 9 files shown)


christianpape:
I assume that JsonFormat follows the canonical Proto3 mapping, so that a non-finite Double value is output as "NaN" or "Infinity". CollectionSetMapper.visitAttribute explicitly sets Double.NaN as soon as getNumericValue() returns null. When "metrics.useJson=true", a data record may therefore contain "value": "NaN", whilst neighbouring data records contain "value": 123.0 - any type-based consumer will trigger an error in this case. I’m not sure whether this poses a major problem, but it would be worth investigating.

dino2gnt:
> any type-based consumer will trigger an error in this case

Is there a more robust option I could use in this case that doesn't mix numbers and string types?  We can't know what the consumer side will look like, but we shouldn't make things needlessly more difficult for them.

christianpape:
> > any type-based consumer will trigger an error in this case
> 
> Is there a more robust option I could use in this case that doesn't mix numbers and string types? We can't know what the consumer side will look like, but we shouldn't make things needlessly more difficult for them.

I’m afraid the only option is to write numeric values as strings as well, which I think is rather unsightly. I think we’ll leave everything as it is and you’ll simply add a note to the documentation pointing out this issue.

marshallmassengill (approved):
LGTM!

PR #8837, opened 2026-09-02
NMS-20275: CVE-2026-39006 test hardening

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-20275

* snmp4j-agent is test infrastructure and ships in nothing. It is declared in exactly one place, `dependencies/snmp4j-agent/pom.xml`, and consumed only by `tests/mock-snmp-agent`. It is absent from every assembly, RPM, Debian package, container image and Karaf feature. `container/karaf/src/main/filtered-resources/etc/custom.properties` exports `org.snmp4j.*` but never `org.snmp4j.agent*`. Nothing reaches `$OPENNMS_HOME/lib`.
* No daemon uses the agent API. No production source imports `org.snmp4j.agent.*`. Note two name collisions that are not usages: `Snmp4JAgentConfig` is OpenNMS's own manager-side target configuration, and `bin/snmp-request` invokes `org.snmp4j.tools.console.SnmpRequest` from snmp4j core.
* The vulnerable MIB is not registered even in the test fixtures. `MockSnmpAgent` overrides`registerSnmpMIBs()` without calling `super`, so `Snmp4jConfigMib` never enters its MOServer. The agent in `SnmpSetIT` does call `super.registerSnmpMIBs()`, but that call registers nothing: `SnmpTargetMIB.registerMOs()` raises `DuplicateRegistrationException`, which `BaseAgent` catches at the head of the registration chain. Verified by dumping `DefaultMOServer.getRegistry()` at runtime — it holds only the two scalars the test itself registers.

`SnmpSetIT`'s agent now unregisters `snmp4jConfigMIB` explicitly and a guard test asserts that nothing under `1.3.6.1.4.1.4976.10.1.1.2` is registered. The unused test-scope dependency was also dropped from `core/snmp/impl-snmp4j`, which had `Embed-Dependency: snmp4j*` and would have embedded the agent jar into a shipped bundle had that scope ever been widened.


marshallmassengill (approved):
LGTM!

PR #8839, opened 2026-09-02
NMS-20281: Vue UI User editor page with tabs, split time zone, date-picker duty schedules

joseanesONMS:
Converts Add/Edit User from a dialog to a routed page at `ui/index.html#/admin/users/:userId` (NMS-20281), with `create` as the new-user marker per the SNMP source-detail convention.

The page keeps every safety property the dialog had while reorganizing the form for space the dialog never offered.

- Three tabs per the ticket: General Information, Duty Schedule, Roles.
- General Information is sectioned: account fields first; Email and Mobile Phone always visible; the pager/PIN fields fold into a TogglePanel that auto-expands for users who have any of them set.
- The time zone is a Region + City/Area pair of filterable selects: underscores shown as spaces, case-insensitive search, region-only zones (UTC) complete on their own, and stored zones the browser list doesn't know still render.
- Duty schedules are a Day of Week / Start / End add row using `OnmsDatePicker` `timeOnly` (NMS-20280) plus a table with per-row delete; entries are only added or removed, so multi-day and hand-edited legacy strings round-trip byte-identically and overnight windows stay legal.
- Dialog semantics carried over: unexposed contact types (XMPP, microblog) round-trip untouched, changed-input-only validation, in-page server errors, and a deep link to a deleted user gets a not-found guard instead of a blank form.
- Merged with develop after #8835 (NMS-20280) landed, so the diff now shows only the NMS-20281 changes.
- 19 new UI tests; verified live against a running instance.


joseanesONMS:
Screenshot 2026-09-02 at 09 30 30
Screenshot 2026-09-02 at 09 13 12
Screenshot 2026-09-02 at 09 13 05
Screenshot 2026-09-02 at 09 12 53
2026-09-02 User Information
Screenshot 2026-09-02 at 09 12 22

synqotik (changes requested):
Need to wait until the `OnmsDatePicker` changes I made make their way into `develop`.

PR #8842, opened 2026-09-03
NMS-20286: Manage WS-Man configuration page (Vue UI)

joseanesONMS:
Adds a Manage WS-Man page under Manage Inventory so WS-Man server definitions, default connection settings and data collection can be maintained without editing XML by hand. The page writes the existing `wsman-config.xml` and `wsman-datacollection*.xml` files through a new admin-only `/api/v2/wsman-config` service, and both DAOs pick the changes up without a restart.

- Server definitions gain an optional `requisition` attribute (XSD, model) so Sync can provision a definition's addresses as nodes with the WS-Man service and add its ranges to discovery; sync only ever adds.
- The status column comes from the poller (servers, responding, down, provisioned but not polled), and a readiness banner can add the WS-Man poller service and rescan requisitions, since the shipped poller configuration has no WS-Man service.
- `FakeWsManAgent`, a new JDK-only test module under `tests/`, is a SOAP fake with a runtime-changeable Windows dataset; tests drive the real CXF client, detector, monitor and collector against it, and smoke `WsmanConfigIT` copies it into the container and runs the whole flow there.
- Two obsolete Woodstox jars pulled in by abdera are excluded from `cxf-dependencies`; they shadow woodstox-core 7 on unsorted classpaths and break CXF's secure parser fallback, which the end-to-end IT exposed.
- Creating new data collection objects is deliberately held back (buttons commented out) until those files move into the database.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20286


marshallmassengill (changes requested):
Blockers... I'm not 100% on the placement for the test agent but this should at least be a better start for where it should go:

- features/wsman/src/main/java/org/opennms/core/wsman/fake/FakeWsManAgent.java ships in the production bundle and gives it a mandatory Import-Package: com.sun.net.httpserver (confirmed in the built manifest). No Karaf etc/*.properties exports that package, and wsman-integration is a Minion boot feature (container/features/src/main/resources/features-minion.xml:116), so Minion stops resolving. Core smoke tests never see this because core loads the jar from lib/. Move the fake to a test-jar or its own module like tests/mock-snmp-agent.
  - features/wsman/pom.xml:20-29 reads ../../opennms-base-assembly/src/main/filtered/etc by filesystem path while base-assembly depends on wsman (inverted module graph), copies it unfiltered so the bundled root file carries a literal ${install.share.dir} placeholder, and resetDataCollection writes that placeholder into the user's etc/ whenever the current root is unreadable (WsmanConfigRestService.java:488-498). It also bundles resource-types.d/wsman-*.xml, which the service never reads.
  - Definition.java:461-489 equals/hashCode omit the new requisition field, and WsmanConfig.equals delegates to the definition list, so a definition differing only by requisition compares equal.
  - WsmanConfigRestServiceIT.java:203-218 resolves opennms.home to the tracked opennms-base-assembly/src/main/filtered, so sync and enable-polling rewrite the git-tracked discovery and poller XML; restore is content-compare in @After, and the cached discoveryFactory bean keeps the added range in memory for later IT classes in the same context.

joseanesONMS:
Thanks, all four addressed in 4e16bbb.

- Fake agent: moved to a new JDK-only `tests/mock-wsman-agent` module (test scope everywhere). The wsman bundle manifest no longer imports `com.sun.net.httpserver`; the smoke test copies the jar into the container and runs it from `/tmp` instead of `lib/`.
- Feature pom: the base-assembly resource copy is gone. Reset reads `share/etc-pristine` only and returns 409 on installs without it, before touching any file; the IT covers both paths. Tests read the shipped files from the base-assembly source tree like the other config tests do.
- `Definition.equals/hashCode` include `requisition` (and `productVersion` in the hash).
- IT: `opennms.home` and the poller factory's file point at a scratch copy for each test, and the discovery and poller singletons reload from the tracked files in `@After`; `git status` stays clean after a run.

opennms-base-assembly/src/main/filtered/etc/snmp-graph.properties.d/microsoft-winrm-graph.properties
```diff
@@ -77,6 +78,28 @@ report.winrm.shmipc.command=--title="Windows IPC Objects (WinRM)" \
  GPRINT:semaphores:MIN:"Min \\: %10.2lf %s" \
  GPRINT:semaphores:MAX:"Max \\: %10.2lf %s\\n"
 
+report.winrm.memory.physical.name=Windows Physical Memory (WinRM)
+report.winrm.memory.physical.columns=totalVisibleMem,freePhysMem
```
dino2gnt: I'd drop this change, make it a separate PR targeting foundation-2024 or -2025, and break it out into its own `.properties` file distinct from anything existing, so that it doesn't create an `.rpmnew` on upgrade.

opennms-base-assembly/src/main/filtered/etc/wsman-datacollection.d/microsoft-windows.xml
```diff
@@ -22,6 +22,32 @@
         <attrib name="PercentInterruptTime" alias="wrmOSCpuPctIntrTime" type="Gauge"/>
         <attrib name="PercentUserTime" alias="wrmOSCpuPctUserTime" type="Gauge"/>
     </group>
+    <!-- feeds the shipped Windows Memory graphs (wrmOSMem*) -->
+    <group name="winrm-memory"
+           resource-uri="http://schemas.microsoft.com/wbem/wsman/1/wmi/root/cimv2/Win32_PerfFormattedData_PerfOS_Memory"
```
dino2gnt: Same feedback as the graph.  I'd fix this in an older foundation and break it out into its own file so it upgrades cleanly.

dino2gnt (dismissed):
You might also want to check this against the `dcy/wsman-134-updates-f2026` branch.

opennms-base-assembly/src/main/filtered/etc/snmp-graph.properties.d/microsoft-winrm-graph.properties
joseanesONMS: Dropped in a771a0031b8; the shipped files are back to develop. The memory/objects groups and the physical memory graph will come as a separate PR in new files against an older foundation.

opennms-base-assembly/src/main/filtered/etc/wsman-datacollection.d/microsoft-windows.xml
joseanesONMS: Dropped in a771a0031b8; the shipped files are back to develop. The memory/objects groups and the physical memory graph will come as a separate PR in new files against an older foundation.

joseanesONMS:
Checked against `dcy/wsman-134-updates-f2026`. Overlap is `Definition.java` and `wsman-config.xsd`: that branch adds `kerberos-encryption` on the root config and on definitions, this one adds `requisition` on definitions, so the merge is mechanical. The one thing that matters: the page rewrites `wsman-config.xml` from its DTO, so once `kerberos-encryption` reaches develop the DTO and the form need the field or a save would drop it. I'll add it as soon as that branch lands (tri-state next to GSS authentication, same as `gss-auth`), or in this PR if it merges first.

joseanesONMS:
Integration-test failures (StatusRestServiceIT, BusinessServiceRestService*IT) were the webapp-rest context failing to introspect the WS-Man service without the integration API on those modules' classpath. Fixed in b62d72f50bf by resolving the TimeSeriesStorage interface by name; the webapp no longer depends on the integration API at compile time.

joseanesONMS:
Core smoke failure was the new WsmanConfigIT: etc/java.conf in the container holds the java binary path, not a JAVA_HOME, so the fake agent never started. Fixed in 05d1d138cde (Minion and Sentinel smoke and the integration stage passed on the previous head).

joseanesONMS:
Second core-smoke run got past the fake agent and failed on the definition PUT: the test round-trips the GET body, and the update models rejected the read-only hasPassword field with a 500. Fixed in 79532a078a0 by ignoring unknown fields on the update models, with an IT for the round trip. Integration, Minion and Sentinel stages passed on the previous head.

joseanesONMS:
Checked against #8845 (wsman 1.3.7, `kerberos-encryption`). Merging that branch into this one conflicts only in `Definition.equals/hashCode` and the XSD, both trivially both-sides. The page-side support (DTOs, update model, tri-state under Advanced next to GSS, definitions badge, help text, IT round trip through the file and the DAO endpoint) is prepared as one commit on a local preview branch built on #8845, with the REST IT and the fake-agent tests green against 1.3.7; it will be added here as soon as #8845 reaches develop, so a save from the page never drops the attribute.

joseanesONMS:
Rebased onto develop now that #8845 (wsman 1.3.7, `kerberos-encryption`) has landed there; the only conflicts were the XSD and `Definition.equals/hashCode`, which now carry both attributes. The page-side Kerberos support from #8848 is folded in as 1b0a7be359a: tri-state under Advanced next to GSS, definitions badge, help text, and an IT that saves the flag on the defaults and a definition, checks the file, and checks the DAO's endpoint. Verified locally on this head: fake-agent and end-to-end tests against 1.3.7, 17 REST ITs, OpenAPI generation, smoke compile, 66 UI tests, lint, build.

joseanesONMS:
build-ui on 1b0a7be failed on an unhandled error after the tests passed: PrimeVue's TabList schedules a 150 ms ink-bar timer on mount and never clears it, so with the Manage WS-Man page left mounted it fired after the container test file's DOM was torn down (ReferenceError: HTMLElement is not defined from @primeuix/utils), only in CI's file ordering. Fixed in ccf66fe by letting the timer fire and unmounting after each test in that file; full UI suite runs locally with no unhandled errors.

marshallmassengill (changes requested):
A few things to adjust still:

1. Control-endpoint writes race in-flight SOAP reads. The control handler calls the setter on a pool thread, mutating a plain list and map that the Enumerate and Get handlers iterate on other pool threads. The result surfaces as a 500. The smoke test changes values while collection runs, so this is a live flake path.
2. The fake agent leaks its thread pool. Close stops the server but never shuts down the executor, which the server's stop method does not touch.
3. A new JAX-RS client per request, never closed. Four polling loops call the request helper repeatedly, so a run builds dozens of clients. The same test already has a REST client to hand.
4. The end-to-end test rebinds the port it just released, so another process can take it in the gap.
5. The storage field is never populated, so its one consumer is dead. Both service mappers rebuild the response without it, while the panel reads it to pick a column header. The RRD-only hint never appears on a non-RRD install (remove the RRD hint I think?).
6. Timeout and max elements cannot be set to zero, though both validators accept it. The retry field next to them already gets this right.
7. A success message survives later actions. Only the sync handler clears it; the other four set just the error.
8. The create branch of the data-collection editors is unreachable. All three Add buttons are commented out, but the emit, the null-item default, and each dialog's editable file selector and new-file validation branch remain live.
9. The tab-timer sleep is unnecessary and diverges from the sibling test, which simply unmounts.
10. That end-to-end test's sleep comment is wrong. It cites mtime granularity, but the length change already defeats that check. The second actually covers the reload interval, and setting that to zero removes the wait.
11. Tab panels are declared out of order. Cosmetic but worth fixing.

joseanesONMS:
@marshallmassengill all eleven addressed in `9cc3741d57d`:

1. The fake agent swaps immutable snapshots of its class data under a lock; handlers iterate whichever snapshot is current, and a control PUT applies all its lines as one update.
2. close() shuts the executor down.
3. The smoke test uses one JAX-RS client for the run, closed after the class.
4. The agent gained a password setter, so the end-to-end test rotates the credential in place and never releases or rebinds its port.
5. Both UI mappers keep the storage block the server sends, so the RRA header can tell RRD from other stores (kept the hint now that it has data).
6. Timeout and max elements accept zero, like retries.
7. Every action outcome replaces the previous message.
8. The unreachable create branch in the three data-collection editors is gone, with the commented-out Add buttons and the add event.
9. The container test runs the tab timer out under fake timers instead of sleeping.
10. The end-to-end sleep is gone; the config DAO checks the file on every access.
11. Tab panels are declared in tab order.

Fake agent and end-to-end tests, smoke-test compile, webapp-rest build, eslint, `vue-tsc` build and the WS-Man UI tests are green; push carries `!smoke`.

marshallmassengill (approved):
I think this is ok to merge but I'd really like @synqotik to take a look first for this one.  It seems ok though.

synqotik (approved):
I'm approving, so people can start using and testing this.

I opened NMS-20339 for me to do a review and add tasks for things to fix (too many to list here). For the most part, it's functional but there's a lot of UX/UI improvements and streamlining / conforming to current patterns to work on.

PR #8843, opened 2026-09-03
Docs: Update trapd REST docs

marshallmassengill:
Updating the docs to fix the upload paths since NMS-19724 repurposed /api/v2/trapd/upload for JSON and added /upload/xml for XML, but the docs page still pointed XML uploads at /upload.


dino2gnt (approved):
LGTM

PR #8844, opened 2026-09-03
NMS-20289: Eventd drops every event until the webapp loads the database eventconf

marshallmassengill:
Events received during startup are discarded with "event does not have a logmsg", because nothing populates the database-backed eventconf until the web application starts.

Related to this, checkEventSanityAndDoWeProcess throws from inside a stream filter over the whole event log, and DefaultEventHandlerImpl breaks out of the processor loop, so one unmatched event discards every event batched with it. A MATCH-ANY-UEI catch-all masks this once the config is loaded, so today it only compounds losses inside the startup window.

This should resolve these.  Thanks to @dino2gnt for spotting this.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20289

### DIFF
--- a/features/events/daemon/src/main/java/org/opennms/netmgt/eventd/EventConfDbBootstrap.java
+++ b/features/events/daemon/src/main/java/org/opennms/netmgt/eventd/EventConfDbBootstrap.java
@@ -1,0 +1,86 @@
+/*
+ * Licensed to The OpenNMS Group, Inc (TOG) under one or more
+ * contributor license agreements.  See the LICENSE.md file
+ * distributed with this work for additional information
+ * regarding copyright ownership.
+ *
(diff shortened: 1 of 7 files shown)


cgorantla (approved):
This looks good.  

Just a note that this will block OpenNMS startup for few seconds as we load all eventconf from DB (webapp was loading this asynchronously).

marshallmassengill:
> This looks good.  
> 
> Just a note that this will block OpenNMS startup for few seconds as we load all eventconf from DB (webapp was loading this asynchronously).
> 
> 
> 

Yeah... I think this is an ok compromise.  If we can come up with something more clever then we can rework it.

cgorantla:
There is chance for us to optimize this. Created https://opennms.atlassian.net/browse/NMS-20299

PR #8845, opened 2026-09-03
NMS-20287: Update to wsman 1.3.7 and add support for new features

dino2gnt:
Update wsman to 1.3.7: Kerberos message encryption, WinRS monitor and detector
  
  Library
  - wsmanVersion 1.3.4 → 1.3.5 (WinRS command execution, Kerberos message
    encryption per MS-WSMV 2.2.9.1, AutoCloseable clients; 1.3.5 fixes WinRS
    inside OpenNMS, where Xalan rejects the JAXP 1.5 transformer attributes).
 - wsmanVersion 1.3.5 → 1.3.6 (whoops, you didn't see nothing)
 - wsmanVersion 1.3.6 → 1.3.7 (Don't allow wsman to populate and then
   destroy the default CXF bus)

  Configuration
  - New wsman-config.xml attribute kerberos-encryption (XSD, JAXB model,
    WsmanAgentConfig, endpoint map). Implies gss-auth and uses the canonical
    host name for the SPN like gss-auth does.
  - WsmanEndpointUtils now carries username/password for GSS and Kerberos
    endpoints; previously monitors and detectors silently dropped them and
    fell back to the JAAS entry (also fixes gss-auth with a password).

  Client lifecycle
  - CachingWSManClientFactory (AutoCloseable): one shared client per
    Kerberos-encrypted endpoint so the login, GSS context, and connection are
    reused across polls; non-Kerberos endpoints pass straight through.
  - try-with-resources at every call site (collector, monitors, detectors,
    asset adapter); destroy() on the collector, monitors, and detector
    factories wired as blueprint destroy-method.

  New components
  - WsManShellMonitor and WsManShellDetector (+ factory): run a command via
    WinRS, match stdout against a banner (substring, or regex when prefixed
    with ~, matched with DOTALL), optionally require an exit code. args is
    passed to WinRS verbatim since the server joins command and arguments
    with spaces and adds no quoting. command/args are excluded from
    placeholder substitution (they reach cmd.exe as written). timeout bounds
    every exchange with the host; timeout and retry default from
    wsman-config.xml, service definition wins. Registered for core
    (META-INF/services) and Minion (blueprint), added to the Minion smoke
    test lists, documented.

  Docs
  - New reference page configuration/wsman-kerberos.adoc: modes,
    prerequisites (AES256 only, reverse DNS for the SPN, clock, ports),
    krb5.conf, keytab (JAAS WSManClient) vs password (incl. SCV), system
    properties for core/Minion/containers, connection handling,
    verification with the wsman CLI, troubleshooting. Cross-linked from the
    collector, both monitors, and all three detector pages.

  Testing
  - Unit: features/wsman 66 tests (endpoint map round trips, config
    parsing, caching factory incl. 32-thread concurrency, monitor and
    detector behaviour, XML &quot; round trip through the poller JAXB model);
    wsman library adds CxfShellOperationsTest for the Xalan case.
  - **Live, against a Windows Server 2019 DC** (dc18) with WinRM on plain HTTP
    5985 and Kerberos message encryption, using a Horizon container built
    from this tree and a Minion container built from the same tree sharing
    the core's network namespace (broker failover:tcp://127.0.0.1:61616):
    * Core (Default location): WS-Man identify and WQL detectors, both shell
      detectors positive and one negative, WsManMonitor GET, two shell
      services up, one intentionally down with the exit-code reason in the
      lost-service event, WS-Man collector collecting; keytab path and
      vault-backed password path (${scv:...}) both verified.
    * Minion (MINION location): same detectors and monitors executed on the
      Minion with credentials resolved on the core and Kerberos on the
      Minion; session reuse confirmed from the client-creation log lines.
  - CI: Minion and Sentinel smoke tests pass after the blueprint fix
    (destroy-method is only valid on top-level beans).

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20287
* https://opennms.atlassian.net/browse/NMS-9350
* https://github.com/OpenNMS/wsman/pull/32
* https://github.com/OpenNMS/wsman/pull/31

### DIFF
--- a/features/wsman/src/main/java/org/opennms/core/wsman/utils/CachingWSManClientFactory.java
+++ b/features/wsman/src/main/java/org/opennms/core/wsman/utils/CachingWSManClientFactory.java
@@ -113,8 +113,12 @@ public class CachingWSManClientFactory implements WSManClientFactory, AutoClosea
                 return m_delegate.getClient(endpoint);
             });
             return new SharedClient(shared);
-        } catch (ExecutionException e) {
-            throw new WSManException("Failed to create WS-Man client for " + endpoint, e.getCause());
+        } catch (ExecutionException | UncheckedExecutionException e) {
+            final Throwable cause = e.getCause();
+            if (cause instanceof WSManException) {
+                throw (WSManException) cause;
+            }
+            throw new WSManException("Failed to create WS-Man client for " + endpoint, cause);
         }
     }
 
--- a/features/wsman/src/main/java/org/opennms/core/wsman/utils/CachingWSManClientFactory.java
+++ b/features/wsman/src/main/java/org/opennms/core/wsman/utils/CachingWSManClientFactory.java
@@ -75,7 +75,7 @@ import com.google.common.cache.RemovalNotification;
  * threads polling the same endpoint through the same factory instance take turns.
  * That is a protocol constraint, not a choice made here.
  */
-public class CachingWSManClientFactory implements WSManClientFactory {
+public class CachingWSManClientFactory implements WSManClientFactory, AutoCloseable {
     private static final Logger LOG = LoggerFactory.getLogger(CachingWSManClientFactory.class);
 
     static final Duration EXPIRE_AFTER = Duration.ofHours(1);
@@ -121,6 +121,7 @@ public class CachingWSManClientFactory implements WSManClientFactory {
     /**
      * Closes and forgets every cached client.
      */
+    @Override
     public void close() {
         m_clients.invalidateAll();
         m_clients.cleanUp();
--- a/features/wsman/src/main/java/org/opennms/core/wsman/utils/CachingWSManClientFactory.java
+++ b/features/wsman/src/main/java/org/opennms/core/wsman/utils/CachingWSManClientFactory.java
@@ -1,0 +1,272 @@
+/*
+ * Licensed to The OpenNMS Group, Inc (TOG) under one or more
+ * contributor license agreements.  See the LICENSE.md file
+ * distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * TOG licenses this file to You under the GNU Affero General
+ * Public License Version 3 (the "License") or (at your option)
+ * any later version.  You may not use this file except in
+ * compliance with the License.  You may obtain a copy of the
+ * License at:
+ *
+ *      https://www.gnu.org/licenses/agpl-3.0.txt
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied.  See the License for the specific
+ * language governing permissions and limitations under the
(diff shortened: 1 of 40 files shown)


features/wsman/src/main/java/org/opennms/core/wsman/utils/CachingWSManClientFactory.java
```diff
@@ -0,0 +1,273 @@
...
+import org.opennms.core.wsman.exceptions.WSManException;
+import org.opennms.core.wsman.shell.CommandResult;
+import org.opennms.core.wsman.shell.ShellOptions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.w3c.dom.Node;
+
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.RemovalNotification;
```
marshallmassengill, line 47:
```suggestion
import com.google.common.cache.RemovalNotification;
import com.google.common.util.concurrent.UncheckedExecutionException;
```

features/wsman/src/main/java/org/opennms/core/wsman/utils/CachingWSManClientFactory.java
```diff
@@ -0,0 +1,273 @@
...
+        }
+        final EndpointKey key = new EndpointKey(endpoint);
+        try {
+            final WSManClient shared = m_clients.get(key, () -> {
+                LOG.debug("Creating shared Kerberos-encrypted WS-Man client for {}", endpoint);
+                return m_delegate.getClient(endpoint);
+            });
+            return new SharedClient(shared);
+        } catch (ExecutionException e) {
+            throw new WSManException("Failed to create WS-Man client for " + endpoint, e.getCause());
```
marshallmassengill:
```suggestion
        } catch (ExecutionException | UncheckedExecutionException e) {
            final Throwable cause = e.getCause();
            if (cause instanceof WSManException) {
                throw (WSManException) cause;
            }
            throw new WSManException("Failed to create WS-Man client for " + endpoint, cause);
```

marshallmassengill (changes requested):
Left a couple of suggestions and sent you a mm message for another item that's up stream.

marshallmassengill (dismissed):
LGTM!

pom.xml
```diff
@@ -2002,7 +2002,7 @@
     <xmlSchemaVersion>2.3.1</xmlSchemaVersion>
     <xmlgraphicsCommonsVersion>2.9</xmlgraphicsCommonsVersion>
     <xstreamVersion>1.4.21</xstreamVersion>
-    <wsmanVersion>1.3.3</wsmanVersion>
+    <wsmanVersion>1.3.6</wsmanVersion>
```
mershad-manesh: we should use 1.3.7

dino2gnt: :joy:

christianpape (approved):
LGTM.

marshallmassengill (approved):
Excelsior!

dino2gnt:
FULL SEND!

PR #8733, opened 2026-08-03
NMS-20153: CSV and PDF export for Resource Graphs

joseanesONMS:
Adds CSV and PDF export to the new /ui Resource Graphs page.

The export reads from the same measurement data the graphs already hold, so it needs no new backend and reflects exactly what is plotted.

- Each graph gains a **CSV** button that downloads its data as one column per plotted metric with one ISO-timestamped row per sample, gaps left blank.
- The page gains a **PDF** button that renders the loaded graphs — chart image plus legend stats — into a single landscape PDF via jsPDF (new UI dependency, ~100 KB gzip).
- CSV omits metrics that have no column in the response (rather than emitting blank columns) and neutralizes spreadsheet formula-injection prefixes in header cells.
- PDF clamps each chart to fit a page, paginates, wraps long legends, and reports how many graphs were exported — only graphs currently mounted are captured, since the page loads them lazily on scroll.
- Covered by 11 unit tests (`buildGraphCsv`, `safeFileStem`, `exportGraphsToPdf`); full `ui` suite green.

* External References
            Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20159

PR #8847, opened 2026-09-07
NMS-20296: Export the dashboard to PDF (Vue UI)

joseanesONMS:
Adds an Export to PDF button to the Vue dashboard toolbar so operators can hand the dashboard around as a document. The export is a capture of the panels as shown, laid out the way they are on screen, rather than a re-rendering of their data.

- Each panel frame is captured with html2canvas (the one new dependency) and placed on landscape A4 pages at its on-screen position, scaled so the grid fills the page width; a panel that would cross a page boundary starts the next page, and the page background is the dashboard's own.
- jsPDF and the download helper are reused from the Resource Graphs export (#8733).
- The file is named `<yyyy-mm-dd>-<dashboard name>.pdf`, and the layout document gains an optional `name` shown in the toolbar in place of the fixed "Home", so a renamed dashboard carries its name into the export; the server stores the layout as a free-form document, so no backend change is needed.
- Capturing is CPU-bound and takes a few seconds, so a full-page overlay reports "Capturing panel n of m" while it runs, with frame yields so the browser can paint it.
- Tests cover filename, pagination, background contrast, a panel that fails to render, progress reporting, and the toolbar states; the whole UI lints clean.


joseanesONMS:
Screenshot 2026-09-07 at 15 40 57
[2026-09-07-Home.pdf](https://github.com/user-attachments/files/31926009/2026-09-07-Home.pdf)

joseanesONMS:
Screenshot 2026-09-07 at 15 47 37

PR #8848, opened 2026-09-08
NMS-20286: Manage Kerberos message encryption on the WS-Man page (needs #8845)

joseanesONMS:
Follow-up to #8842 for the `kerberos-encryption` attribute that #8845 (wsman 1.3.7) adds to the root config and to definitions, so a save from the Manage WS-Man page keeps it instead of dropping it.

- The page reads and writes the flag like the other connection settings: DTOs, update model, a tri-state under Advanced next to GSS authentication, a badge in the definitions table, and help text noting that it implies GSS authentication and makes plain HTTP safe.
- The REST IT saves the flag on the defaults and a definition, checks the file, and checks that the DAO's endpoint for that server has Kerberos encryption enabled.
- CI cannot pass until #8845 reaches develop: this commit calls `setKerberosEncryption`/`isKerberosEncryption`, which develop's `Definition` and `WsmanAgentConfig` do not have yet; the same commit on a local branch with #8845 merged in passes the REST IT (17), the WS-Man UI tests (66), lint, the build, and the fake-agent tests against 1.3.7.
- Stacked on the #8842 branch so the diff is only this change; retarget to develop once #8842 merges.


joseanesONMS:
Folded into #8842 as 1b0a7be359a now that #8845 is on develop; closing.

PR #8851, opened 2026-09-09
NMS-20300: Say the topology map needs WebGL instead of failing blank

marshallmassengill:
Browsers without WebGL will show a blank page and throw "TypeError: can't access property getParameter, n is null" in the console.  This will improve that so there's a real error message.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20300


dino2gnt (approved):
If it works, ship it.

synqotik (approved):
Seems reasonable!

PR #8852, opened 2026-09-09
NMS-20031: Vue Node Details - Complete Basic Layout

synqotik:
# NMS-20031: Vue Node Details screen

Continuing work that brings the Vue Node Details page up to parity with the legacy JSP node page
(`element/node.jsp`), so we can replace the current JSP-based UI.

## What the page shows

**Title row** — "Node Details for _node-label_", with a node actions menu on the right
offering the same links the legacy page did: events, alarms, outages,
assets, metadata, hardware inventory, availability, site status, resource graphs,
rescan, admin, SNMP update, schedule an outage, topology map, and node link
details. Site status and SNMP update appear only when the node has the data they need — a
building, and an SNMP-primary interface.

**Status badges** — node status, label, ID and monitoring location. Status is
color-coded: active reads as success, deleted as danger, unknown as neutral.

**SNMP Attributes** — name, sys object ID, location, contact and description.

**Surveillance Category Memberships** — the node's categories, with an edit
action for admins that opens the category admin screen for this node.

**Notifications** — links to the current user's outstanding and acknowledged
notifications for this node, plus a link to the Notifications page.

**Availability** — availability over the last 24 hours for the node, each of its
interfaces, and each monitored service, with the timeline graphics from the
legacy page. The `XXX` in the interface and service rows is a placeholder for
status icons still to come; this panel is being reworked in a subsequent PR.

**Node Interfaces** — IP and SNMP interfaces in tabs. More work to be done on these in a subsequent PR.

**Recent Events** — paginated, with a link to the full event list for this node
and CSV/JSON download.

**Recent Outages** — paginated, with a link to the full outage list for this node
(showing both current and resolved) and CSV/JSON download. Each row links out to
the outage, its interface and its service; lost and regained times are shown as
dates, and an outage that has not been resolved yet has its lost time flagged.

## Downloads

The events and outages panels can both export what they are showing as CSV or
JSON, covering the full record rather than only the visible columns. The rows are
already on hand, so a download uses that data instead of making another API request,
and so the file always agrees with the table. The user can change the rows-per-page
to get more or less data; exporting a node's entire history is left to
the Events and Outages pages (not yet implemented - will be done as part of converting
those pages to Vue).

The CSV neutralises anything a spreadsheet would run as a formula, since event
messages arrive from traps and syslog, and gives nested values their own columns
rather than a blob of JSON in one cell.

## When the node changes, or cannot be loaded

The page keeps one component instance as the user moves between nodes, so every
panel follows the node id rather than reading it once — otherwise a table would
keep showing one node while the links beside it pointed at another.

A node that cannot be fetched — e.g. a 404 for an id that does not exist — clears
all data from any previous node. The panels that describe a node are held
back until there is one, the heading reads "Node Details for N/A", and the
interfaces, outages, events and availability are cleared. A fetch still in
flight is treated as different from one that failed.

We will revisit this in a subsequent PR, e.g. using skeletons and spinners which loading
and a different display for empty or errored state.

## Node list

The per-row actions menu picks up the new node link details action, so the node
list and the Node Details page now offer the same set, including the
SNMP-primary-aware SNMP update link.

## Also in here

- The node asset record is now fully typed, which tidied up some latitude and
  longitude handling in the map components.
- A few `Outage` fields were corrected to match what the API actually sends.
- The panels share one card component, and the two download implementations share
  one helper, rather than each panel carrying its own copy.
- The node action links now live in one place instead of two lists that had
  drifted apart.

## Testing

Unit tests cover each panel, the download and link building, and the store
additions — including the empty and missing-data cases (no categories, no
outages, an outage with no interface or an unrecognised service, a node with no
building, a node that fails to load, and moving between nodes). The full `ui`
suite, lint and type check pass.

## Follow-ups, not in this PR

- The Availability panel is being reworked, including the status icons its `XXX`
  placeholders stand in for. The bar graphs will be rendered client-side.
- SNMP interface ordering, shading
- `useRole` caches the auth store at module level, which makes role-dependent
  component tests order-dependent. Worked around in the tests here; the
  composable itself should be fixed.
- Icon-button actions could eventually move into `OnmsCard` instead of the Node Details
  card component.
- Downloading a node's complete event and outage history belongs on the Events
  and Outages pages. We will implement that as part of the Vue rewrite.
- The node list's own CSV export still quotes only on commas, escapes no
embedded quotes, and does not guard against formulas — unlike the events and
outages export added here.
- We need better handling when REST data is not available, or is slow to populate. Perhaps spinners on each
panel or something similar.
- Better date display, ideally on 2 lines (date/time). Tricky because user can customize it.
- Various other panels that only display when certain data is available or options are selected.
- We expect to rearrange the panels, perhaps add tabs, etc.

### Vue Node Details - top section

view-node-details-1

### Vue Node Details - bottom section

view-node-details-2

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20031

### DIFF
--- a/ui/src/components/Map/LeafletMap.vue
+++ b/ui/src/components/Map/LeafletMap.vue
@@ -284,7 +284,7 @@ const getNodeCoordinateMap = computed(() => {
   const map = new Map<string, Coordinates>()
 
   allNodes.value.forEach((node: Node) => {
-    const coord = { latitude: node.assetRecord.latitude, longitude: node.assetRecord.longitude }
+    const coord = { latitude: node.assetRecord.latitude ?? 0, longitude: node.assetRecord.longitude ?? 0 }
     map.set(node.id, coord)
     map.set(node.label, coord)
   })
--- a/ui/src/components/Map/MapNodesGrid.vue
+++ b/ui/src/components/Map/MapNodesGrid.vue
@@ -124,7 +124,7 @@ const nodes = computed<Node[]>(() => mapStore.getNodes())
 const nodeLabelAlarmSeverityMap = computed(() => mapStore.getNodeAlarmSeverityMap())
 
 const doubleClickHandler = (node: Node) => {
-  const coordinate: Coordinates = { latitude: node.assetRecord.latitude, longitude: node.assetRecord.longitude }
+  const coordinate: Coordinates = { latitude: node.assetRecord.latitude ?? 0, longitude: node.assetRecord.longitude ?? 0 }
   mapStore.setMapCenter(coordinate)
 }
 
--- a/ui/src/components/Map/MarkerPopup.vue
+++ b/ui/src/components/Map/MarkerPopup.vue
@@ -45,20 +45,27 @@ import { OnmsIcon } from '@opennms/onms-ui'
 import Location from '@opennms/onms-ui/icons/action/Location.vue'
 import { LPopup } from '@vue-leaflet/vue-leaflet'
 import { Node } from '@/types'
-import { mapPopupOptions, stringToFixedFloat } from './utils'
+import { mapPopupOptions } from './utils'
 
 const props = defineProps({
-  baseHref: { type: Object as PropType<string> },
-  baseNodeUrl: { type: Object as PropType<string> },
+  baseHref: { type: String, required: true },
+  baseNodeUrl: { type: String, required: true },
   node: { type: Object as PropType<Node>, default: () => {
     return
   } },
-  ipAddress: { type: Object as PropType<string> },
+  ipAddress: { type: String },
   nodeLabelToAlarmSeverity: { type: Function as PropType<(label: string) => string>, required: true }
 })
 
-const latitude = computed(() => stringToFixedFloat(props.node.assetRecord.latitude, 6))
-const longitude = computed(() => stringToFixedFloat(props.node.assetRecord.longitude, 6))
+const numToFixedFloat = (num: number | null, decimalPlaces: number): string => {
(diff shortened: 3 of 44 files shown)


marshallmassengill:
I don't know what I would change about them exactly but the availability graphic looks incomplete to me.  Like it's missing a border maybe?  It just looks out of place.

synqotik:
> I don't know what I would change about them exactly but the availability graphic looks incomplete to me. Like it's missing a border maybe? It just looks out of place.

@joseanesONMS 
Yes, it's a known issue, I'm re-doing it in a subsequent PR, please just ignore it for now.

marshallmassengill (changes requested):
One thing surfaced as a blocker:

Node switching is last-write-wins, and the Update SNMP link can carry the previous node's address. Neither getNodeById (nodeStore.ts:78-88) nor getNodeSnmpPrimaryInterface (:120) records which id it was asked for, so a slower response overwrites a newer one. Proven with two tests against the real store: with the page on node 2, store.node.id is '1' and snmpPrimaryIpAddress is node 1's address, making the action build admin/updateSnmp.jsp?node=2&ipaddr=<node 1's IP> — a form that rewrites SNMP configuration for an address that is not the node's, which is the exact hazard nodeActionLinks.ts:69-71 reasons about. Capture the requested id in both actions and discard a stale response.

synqotik:
@marshallmassengill 

> One thing surfaced as a blocker:

Should now be fixed.

marshallmassengill (approved):
I think this is good to go now.

PR #8853, opened 2026-09-10
NMS-16196: Hardening match() method

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-16196

The database part was already fixed by me in NMS-16199. This PR hardens the match() method and prevents the method to fail and prevent the topology to be generated.


cgorantla (approved):
LGTM!

PR #8855, opened 2026-09-10
NMS-20308: Fix Sentinel dropwizard-metrics version range in featuresProcessing

mershad-manesh:
This issue causes Sentinel smoke tests to fail (on `release-2025.x` branch).

Widen the range to also catch CXF's own bundled dropwizard-metrics version.

### All Contributors

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20308


cgorantla (approved):
LGTM!

This is what probably happening.

This PR widens the Sentinel Karaf bundle-replacement rule in features-processing.xml from [3, 4.2.10) to [3, 10) for metrics-core and metrics-jvm. Karaf applies that rule to every feature repository as it loads, both in the karaf-maven-plugin assembly and in the runtime FeaturesService. So the Newts descriptor's reference to metrics-core 4.2.25 is rewritten to 4.2.10 on the fly. Both cassandra-driver definitions then point at the same jar, so it no longer matters which one wins.

Before this PR, `cassandra-driver` brought 4.2.25 and core has 4.2.10 and that produced conflict.

christianpape (approved):
LGTM.

PR #8856, opened 2026-09-11
NMS-20309: Vue UI dashboard panels for the legacy Charts

joseanesONMS:
Makes the three charts of the legacy Charts page available as panels in the Vue dashboard's Add Panel picker (NMS-20309).

The charts stay data-driven from `etc/chart-configuration.xml`, so admin-defined charts work too and nothing about the legacy page changes in this PR.

- New read-only `GET /api/v2/charts` and `GET /api/v2/charts/{name}` run the configured series queries and return categories plus one value series per series definition, with severity keys labelled as before.
- One `ConfiguredChartPanel` draws the chart with Chart.js in the Resource Graphs style: translucent bars edged in the configured color, value labels, soft grid, shared HTML legend.
- Three registry entries, "Charts - Alarms", "Charts - Last 7 Day Outages" and "Charts - Node Inventory", pin the shipped charts through a new `defaultOptions` seed; the panel options dialog can switch a panel to any configured chart.
- The Charts page and its menu entry are untouched.
- `ChartRestServiceIT` covers the list, the severity labels, the inventory counts and the 404; UI tests cover the service, the panel and the option seeding.

### DIFF
--- a/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/ChartRestService.java
+++ b/opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/ChartRestService.java
@@ -1,0 +1,173 @@
+/*
+ * Licensed to The OpenNMS Group, Inc (TOG) under one or more
+ * contributor license agreements.  See the LICENSE.md file
+ * distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * TOG licenses this file to You under the GNU Affero General
+ * Public License Version 3 (the "License") or (at your option)
+ * any later version.  You may not use this file except in
+ * compliance with the License.  You may obtain a copy of the
+ * License at:
+ *
+ *      https://www.gnu.org/licenses/agpl-3.0.txt
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied.  See the License for the specific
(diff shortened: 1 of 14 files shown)


joseanesONMS:
Screenshot 2026-09-11 at 09 41 28
Screenshot 2026-09-11 at 09 04 01

joseanesONMS:
Screenshot 2026-09-11 at 09 45 51

dino2gnt:
You shouldn't waste the time on modernizing this feature.  Let it die the slow, silent and ignoble death it deserves.

joseanesONMS:
That is insightful :)  I would prefer to give it a quick, merciful death and just eliminate the Surveillance Dashboard, Heatmap, Trends, and Charts with a single blow.   Say the word and I will be happy to do so.  @marshallmassengill 

If we want a more cautious approach, we can also provide an alternative.

Either way, I want to get to the point where there isn't a trace of our dated UI.  It has served us well, but it is time to clean it up.

marshallmassengill:
So the surveillance dashboard is worth keeping I think?  It needs to be updated though.  I do think the heatmap, trends, and charts can all go away though... along with the statsd/Metrics Statistics pages.  I would do one PR for removing each of those separately.

joseanesONMS:
@marshallmassengill @dino2gnt @synqotik  Will create a ticket and PR to remove heatmap, trends, and charts

marshallmassengill:
Close this one out when you can @joseanesONMS.  I don't think it's needed given the removals.

joseanesONMS:
No longer needed; closing without merge.

PR #8857, opened 2026-09-11
NMS-20312: Topology Preview viewport doesn't honor background size correctly

marshallmassengill:
The new topology preview viewport ends up having issues when adding a custom background or adding a new node.  Similar to what Morteza actually had me looking into with adding boxes a while back. 

On a new custom view the background image, labels, shapes and other overlays positioned through graphToViewport() were drawn squashed and off-centre, then jumped into place on the first node drop. viewportToGraph() was unaffected because sigma always recomputes the inverse matrix, which is why drops still landed under the cursor.  This should fix it.

Assisted by Anthropic Claude Opus 5/Fable 5.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20312


cgorantla (approved):
LGTM!

PR #8864, opened 2026-09-11
NMS-20318: Drop the Guava 17/21 features and update wsman to 1.3.3

cgorantla:
### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20318

### DIFF
--- a/container/features/pom.xml
+++ b/container/features/pom.xml
@@ -394,10 +394,7 @@
                           <feature>fst</feature>
                           <feature>guava</feature>
                           <feature>guava-dependencies</feature>
-                          <feature>guava17</feature>
                           <feature>guava18</feature>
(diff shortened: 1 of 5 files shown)


cgorantla:
This needs some more testing to validate we are not breaking anything.

dino2gnt:
The only customer on PB2023 that I am aware of has no use for wsman. I'm not sure it matters if it's broken or not. :)

cgorantla:
> The only customer on PB2023 that I am aware of has no use for wsman. I'm not sure it matters if it's broken or not. :)

Needed indirectly to update Guava.  Requirement coming from security scans flagging Old Guava versions.

christianpape (approved):
LGTM.

PR #8866, opened 2026-09-11
NMS-20313: Remove the Trends visualization

joseanesONMS:
Removes the Trends visualization from the Dashboards menu and from the code base; second of four PRs splitting NMS-20313 by menu item, following the review on #8858.

The page saw almost no use and carried its own configuration model and jQuery plugin, so this takes the whole feature out rather than hiding the entry.

- Trend controllers and pages, the `org.opennms.netmgt.config.trend` model and schema, `trend-configuration.xml` and its integration-test module.
- The jQuery sparkline plugin and trend-line style that only this page used, the menu entry in every template and the smoke-test step.
- `WillItUnmarshalIT` and `ConfigTesterTest` no longer expect the file; the release note says it can be deleted.

PR #8867, opened 2026-09-11
NMS-20313: Remove the Charts visualization

joseanesONMS:
Removes the Charts visualization from the Dashboards menu and from the code base; third of four PRs splitting NMS-20313 by menu item, following the review on #8858.

The page saw almost no use and pulled JFreeChart into the webapp, so this takes the page, servlet and module out rather than hiding the entry.

- The charts page, `PurdyChartServlet` with its web.xml wiring, and the `features/charts` module with its pom, Karaf export and page-sequence references.
- The menu entry in every template and the smoke-test step.
- `chart-configuration.xml`, `ChartConfigFactory` and the `org.opennms.netmgt.config.charts` model and schema go with the page; nothing else read them. The config meta-tests no longer expect the file.

External References
Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20313


joseanesONMS:
@dino2gnt could you review and approve when you get a chance? Merged with develop after #8866 (Trends) landed, so this is mergeable again; it carries the fixes from the #8858 review rounds.

marshallmassengill (changes requested):
One item worth addressing I think:

The shipped config names a class this PR deletes. opennms-base-assembly/src/main/filtered/etc/chart-configuration.xml:2 keeps sub-label-class="org.opennms.web.charts.SeveritySubLabels", which goes with the charts module. Nothing loads it at head, so it is inert rather than broken, but the file ships pointing at a class that is no longer in the product. The attribute is optional in the schema, so dropping it is a one-attribute edit here, with no schema change and no dependency on #8856.

joseanesONMS:
@marshallmassengill done in `10fb39bcea4`: the `sub-label-class` attribute is dropped from the shipped chart-configuration.xml, and the release note and PR body no longer refer to #8856, which has been closed.

dino2gnt (dismissed):
LGETM.

marshallmassengill (dismissed):
I think this is ok now.  You may want to wait for @cgorantla to look at it before merging.

synqotik (dismissed):
LGTM

cgorantla (commented):
Why are we keeping `etc/chart-configuration.xml` ?
Can this be removed if we are no longer have the charts page

joseanesONMS:
@cgorantla agreed, nothing reads it once the page is gone. `e29bf885de9` removes the shipped chart-configuration.xml, the `org.opennms.netmgt.config.charts` model and schema, ChartConfigFactory and the config-tester wiring; WillItUnmarshalIT and ConfigTesterTest no longer expect the file. WillItUnmarshalIT and MenuProviderTest are green locally; CI is running.

joseanesONMS:
The push that removed chart-configuration.xml dismissed the earlier approvals. @dino2gnt @synqotik @marshallmassengill could you re-approve once CI is green? The only change since your reviews is that removal.

joseanesONMS:
The sentinel smoke failure on `e29bf885de9` was mine: the chart-factory removal also took the neighbouring collectdConfigFactory and tokenAuthConfigFactory-init beans out of the config-tester context, and the container entrypoint runs the config tester before startup, so every Horizon container exited. Restored in `88fee319040`; CI re-running.

dino2gnt (approved):
Good enough for the girls we date.

PR #8868, opened 2026-09-11
NMS-20313: Remove the Statsd daemon and the Metrics Statistics pages

joseanesONMS:
Removes the Statsd daemon and the Metrics Statistics (statsd) pages from the Dashboards menu and from the code base; last of four PRs splitting NMS-20313 by menu item, following the review on #8858.

Statsd was the only producer of the statistics-report rows, so the daemon, its persistence and everything that displayed those rows go together.

- The daemon, its Spring contexts, configuration model, schema and DAO, the `StatisticsReport` and `ResourceReference` entities and DAOs, the web service layer and statistics-report pages, and its docs.
- The service-catalog entries in every shipped catalog, JMX collection and graph definitions, and the two database reports only Statsd could feed.
- No upgrade migration: a Statsd entry left in an existing `service-configuration.xml` is dropped by the catalog merge with a warning, and the release note says the stale files can be deleted.
- The `statisticsreport` tables and the Vacuumd purge are kept so existing rows still age out; the migrator tests count one active service fewer.

Verified locally with `ServiceCatalogParityTest`, the core/upgrade migrator tests (service counts), `WillItUnmarshalIT`, `MenuProviderTest` and a full build of the combined change.

External References
Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20313


joseanesONMS:
@dino2gnt could you review and approve when you get a chance? Merged with develop after #8866 (Trends) landed, so this is mergeable again; it carries the fixes from the #8858 review rounds.

core/upgrade/src/main/java/org/opennms/upgrade/implementations/StatsdRemovalMigratorOffline.java
```diff
@@ -0,0 +1,144 @@
...
+                final StringWriter sw = new StringWriter();
+                sw.write("<?xml version=\"1.0\"?>\n");
+                JaxbUtils.marshal(currentCfg, sw);
+                try (FileWriter fw = new FileWriter(configFile)) {
+                    fw.write(sw.toString());
+                }
+            }
+        } catch (final Exception e) {
+            throw new OnmsUpgradeException("Can't remove Statsd from the services configuration because " + e.getMessage(), e);
+        }
```
cgorantla: We don't need to programmatically update service configuration. Just update it directly in base assembly service-configuration.xml.

Also no need to move these files to archive statsd-configuration.xml. Just remove them.

cgorantla (changes requested):

core/upgrade/src/main/java/org/opennms/upgrade/implementations/StatsdRemovalMigratorOffline.java
joseanesONMS: Done in `b890fd1d2bd`: the migrator and its test are gone, so the removal lives only in the shipped catalogs (base assembly, defaults, example, container template, smoke overlay). A stale entry in an existing service-configuration.xml is dropped by the catalog merge with a warning, and the release note now says that entry and statsd-configuration.xml can be deleted. For the record, dino2gnt had asked for the migrator on #8858; this follows your call.

cgorantla: We should remove statsd from service-configuration.xml and also remove etc/statsd-configuration.xml directly. 

Don't need to do it from upgrade job.

joseanesONMS: That is exactly what the current head (`3159414d9ba`) does: the Statsd entry is removed directly from the shipped service-configuration.xml (base assembly, config-model defaults, the example, the container template and the smoke overlay), etc/statsd-configuration.xml is deleted, and there is no upgrade job at all since `b890fd1d2bd` removed the migrator and its test. This thread is anchored on that deleted file, which is why it still shows. CI is green; could you re-review?

joseanesONMS: To make sure nothing is left: `e91a170656b` also drops the Statsd block from core/upgrade's 14.0.0 baseline, the file the pre-14 migrator uses to re-add missing services, so no upgrade path can bring the entry back. With that, the only Statsd mentions on the branch are the historical test fixtures under core/upgrade/src/test/resources. The shipped service-configuration.xml has no Statsd entry, etc/statsd-configuration.xml is deleted, and there is no upgrade job.

dino2gnt: > drops the Statsd block from core/upgrade's 14.0.0 baseline

I don't think there is a need to introduce changes to historical upgrade tasks.

joseanesONMS: Fair enough; `87f2a12b10c` reverts that, so the 14.0.0 baseline is back to what develop has and the branch touches no upgrade task at all. What remains is exactly the shipped catalogs and the deleted etc file.

cgorantla: May be I looked at wrong commit https://github.com/OpenNMS/opennms/pull/8868/changes/b890fd1d2bd37c6af4b339ce116a7a0d88e5750a

Looks like Statsd was removed from etc/service-configuration.xml

I'm good with it. Just revert the commit https://github.com/OpenNMS/opennms/pull/8868/changes/e91a170656b6eb0acd72d901f5bfd055c3930e44
and we are good to go.

cgorantla (approved):
LGTM!

PR #8869, opened 2026-09-11
NMS-20321: Resolve OpenAPI schemas from the JSON wire names, not bean property names

marshallmassengill:
This registers a resolver in opennms-openapi-docs that introspects with the Jackson 1 annotations first and JAXB second, matching the provider's runtime order, and applies Jackson 1's wrapper-name rule. One endpoint referenced the wrong type outright and is corrected alongside: /config/trapd now points at TrapdConfiguration, the object on the wire, rather than the TrapdConfig interface.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20321

### DIFF
--- a/opennms-openapi-docs/src/main/java/org/opennms/openapi/WireFormatModelResolver.java
+++ b/opennms-openapi-docs/src/main/java/org/opennms/openapi/WireFormatModelResolver.java
@@ -1,0 +1,161 @@
+/*
+ * Licensed to The OpenNMS Group, Inc (TOG) under one or more
+ * contributor license agreements.  See the LICENSE.md file
+ * distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * TOG licenses this file to You under the GNU Affero General
+ * Public License Version 3 (the "License") or (at your option)
+ * any later version.  You may not use this file except in
+ * compliance with the License.  You may obtain a copy of the
+ * License at:
+ *
+ *      https://www.gnu.org/licenses/agpl-3.0.txt
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied.  See the License for the specific
+ * language governing permissions and limitations under the
+ * License.
+ */
+package org.opennms.openapi;
+
+import java.lang.reflect.Type;
+
+import org.codehaus.jackson.annotate.JsonBackReference;
+import org.codehaus.jackson.annotate.JsonIgnore;
+import org.codehaus.jackson.annotate.JsonProperty;
+import org.codehaus.jackson.annotate.JsonValue;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.databind.AnnotationIntrospector;
+import com.fasterxml.jackson.databind.BeanDescription;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.PropertyName;
+import com.fasterxml.jackson.databind.cfg.MapperConfig;
+import com.fasterxml.jackson.databind.introspect.Annotated;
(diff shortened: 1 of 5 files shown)


opennms-openapi-docs/src/main/java/org/opennms/openapi/WireFormatModelResolver.java
```diff
@@ -0,0 +1,161 @@
...
+ */
+final class WireFormatModelResolver extends ModelResolver {
+
+    WireFormatModelResolver() {
+        super(jaxbAwareMapper());
+    }
+
+    private static ObjectMapper jaxbAwareMapper() {
+        final ObjectMapper mapper = Json.mapper().copy();
+        // Jackson annotations before JAXB, as in the provider's default (JACKSON, JAXB) order.
```
cgorantla, line 78: A JAXB adapter converts a Java object into a simpler value when producing JSON. The resolver uses the original Java type without accounting for that conversion.

For example, OnmsSnmpInterface.nodeId holds an OnmsNode in Java, but the API returns "nodeId": 42. The schema incorrectly tells clients to expect an object. Likewise, Event.interface is returned as a string
such as "192.0.2.1", but documented as an object.

Fix: Use the adapter’s output type when generating the property’s schema.

opennms-openapi-docs/src/main/java/org/opennms/openapi/WireFormatModelResolver.java
```diff
@@ -0,0 +1,161 @@
...
+        mapper.setAnnotationIntrospector(AnnotationIntrospector.pair(
+                AnnotationIntrospector.pair(
+                        mapper.getSerializationConfig().getAnnotationIntrospector(),
```
cgorantla, line 81: An enum can have different names in Java and XML. The resolver uses the XML names, while the runtime JSON provider expects the Java names.

For SnmpTrapSink.version, the schema now advertises values such as "v1" and "v2-inform". The JSON API actually accepts "V1" and "V2_INFORM" and rejects those lowercase alternatives.

   Fix: Discover enum values using the same naming behavior as the runtime Jackson 1 provider.

opennms-openapi-docs/src/main/java/org/opennms/openapi/WireFormatModelResolver.java
```diff
@@ -0,0 +1,161 @@
...
+            return value == null ? null : value.value();
+        }
+    }
+
+    /**
+     * Jackson 1 took the XmlElementWrapper name as the property name outright. Jackson 2's
+     * USE_WRAPPER_NAME_AS_PROPERTY_NAME renames after collection, so two wrapped lists sharing an
+     * XmlElement name collide before the rename (AlarmStatistics has four "alarm" wrappers).
+     */
+    private static final class WrapperNamingJaxbIntrospector extends JaxbAnnotationIntrospector {
```
cgorantla, line 227: The resolver copies an XML default into the JSON schema without checking whether the JSON API accepts it.

For SyslogDestination.ip-protocol, the schema correctly lists "UDP" and "TCP" as allowed values, but also declares "udp" as the default. A client that fills in this default will send a value the API rejects.

Fix: Omit incompatible XML defaults, or translate them to the corresponding valid JSON value—here, "UDP".

marshallmassengill: Pushed an update for this.

opennms-openapi-docs/src/main/java/org/opennms/openapi/WireFormatModelResolver.java
marshallmassengill: Pushed an update for this.

opennms-openapi-docs/src/main/java/org/opennms/openapi/WireFormatModelResolver.java
marshallmassengill: Pushed an update for this.

cgorantla (approved):
LGTM!

PR #8870, opened 2026-09-14
Docs: Document the deprecation of Newts and Cassandra time series storage

marshallmassengill:
Adds a Breaking Changes entry to the release notes and a note on the newts page.  We missed this.  I borrowed the wording from the blog but let me know if this needs to be adjusted.  I assume we'll tackle H37 when the time comes and removing things if needed.


dino2gnt (approved):
Sure

PR #8871, opened 2026-09-15
NMS-19934: Updated JLine version

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-19934


cgorantla (commented):
Could we upgrade Karaf to 4.4.9? It may reduce the number of dependency patches we need to maintain.

  Looking at https://github.com/OpenNMS/opennms/pull/8262, the Karaf upgrade appears manageable if we separate it from the Java 21 changes and keep our current Java version.

christianpape:
> #8262

Ok. I'll check.

christianpape:
> Could we upgrade Karaf to 4.4.9? It may reduce the number of dependency patches we need to maintain.
> 
> Looking at #8262, the Karaf upgrade appears manageable if we separate it from the Java 21 changes and keep our current Java version.

See PR https://github.com/OpenNMS/opennms/pull/8877

Will close this one, in the case #8877 is merged.

christianpape:
Closed in favor of NMS-20332.

PR #8873, opened 2026-09-16
NMS-20328: Scheduled Outages follow-up (Vue UI)

joseanesONMS:
Follow-up to the Scheduled Outages page (#8820) from synqotik's review, plus layout fixes found while testing (NMS-20328).

The time-span rows move to the shared date picker and the page's spacing and buttons follow the other Vue admin pages.

- Time spans use `OnmsDatePicker`: time-only with seconds for daily, weekly and monthly outages, date and time for specific ones; the wire strings are formatted from the picked Date so the length-keyed parser still gets 20 or 8 characters.
- The Create New Outage dialog gets room between its rows and a right-aligned action row; both Cancel buttons use the ghost variant.
- The editor keeps its Create/Cancel buttons under the schedule instead of below the whole grid, and a collapsed "How to fill in this form" panel walks through the five steps.
- The list page pads its card like its siblings, and Time Spans sit 1rem below the outage type.

External References
Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20328


marshallmassengill (changes requested):
- Clearing a time field and pressing Add throws. Manual input is on by default in the installed PrimeVue, blank text parses to null, the validity check accepts null, and the model is updated to null. Both span fields are typed as a plain date and passed straight to the formatter, which calls an hour accessor on null. The clear button reaches the same place. The discrete selects this PR replaces could never be empty, so the picker swap introduces it.
- The new help panel promises a rule nothing enforces. It states the end must come after the start. The component emits without comparing, and the container's handler rejects duplicates only, by which point both values are already formatted strings.
- The year is the one field the specific formatter does not pad, so a year below 1000 yields nineteen characters and is read as a daily span. Reachability is unconfirmed, but the fix is one call and the comment above says the length is load-bearing.
- No test covers an empty or out-of-order span.

joseanesONMS:
@marshallmassengill all four addressed in `bb56fe89fc6`:

- A cleared start or end (manual input or the clear button) now disables Add and shows "Enter both a start and an end time." instead of reaching the formatter; the field type is nullable so the compiler enforces the check.
- The end must come after the start is enforced before formatting: full date and time for a specific outage, time of day for daily/weekly/monthly.
- The year is padded to four digits in the specific format, so the 20-character length holds for any year.
- Tests cover a cleared field, an out-of-order span (both types), the recurring time-of-day comparison and the year padding.

marshallmassengill (approved):
I think this is good to go for now pending a review from @synqotik.  That being said, the workflow for this is what needs more thought and effort still.

PR #8875, opened 2026-09-16
NMS-20301: Vue Node Details - interfaces panel and page layout

synqotik:
Updates to the **Node Interfaces** panel on the Vue Node Details page, along with layout and other changes.

**Please ignore the Availability panel (other than its placement in the page) for purposes of this PR. It will be revamped in an upcoming PR.**

## Node Interfaces panel

**SNMP interfaces**

- Seven columns: Index, Name, Status, Alias, Speed, Descr, Flows.
- Rows order by `ifIndex` ascending by default.
- New **Status** column showing `UP` / `DOWN` / `DISABLED` | `TESTING` | `UNKNOWN` status plus a "severity" type color.
A tooltip shows the two raw IF-MIB statuses it derives from including Admin and Operational status. Example:
  (`Admin Status: 1 (Up) | Operational Status: 1 (Up)`).
`SnmpIfStatus` covers the full `ifOperStatus` range (1–7), not just up/down/testing, because devices do report `dormant` and `notPresent`.
- `ifSpeed` is formatted — `100000000` now reads `100 Mbps`.
- `ifIndex` links to the SNMP Interface details page, `element/snmpinterface.jsp`.

**IP interfaces**

- Four columns: IP Address, IP Host Name, SNMP ifIndex, Status.
- The address now links to the Interface details screen, `element/interface.jsp`.
- **Status** spells out the `isManaged` code: `Managed`, `Unmanaged`, `Forced Unmanaged`,
  `Not Monitored`, `Deleted`. The column used to show the bare letter.

**Both tables**

- Every displayed column sorts, case-insensitively.
- A debounced (175 ms) search box filters on what the columns actually show.
- Both scroll horizontally, with per-column min-widths, rather than compressing their columns
  into the half-page panel.
- A table-scoped spinner appears if a fetch outlives 175 ms.
- The panel moves into `NodeDetailsPanel`, so it matches the other panels on the page.

## Page layout

- **`NodeSnmpAttributes.vue` deleted.** The node actions menu's `Info...` — previously only on the
  node list — is now on this page too, displaying an Info dialog, which already contained every attribute the panel
  showed (`sysName`, `sysObjectId`, `sysLocation`, `sysContact`, `sysDescription`). The
  Availability graph moves up into the freed space, so it's now near the top, where it should probably be (please
note this panel will be revamped later on, it is expected to have display issues).
- **`NodeCategoriesPanel.vue` deleted.** Category names move onto the header as a badge (first
  two, then an ellipsis, with all of them in a hover tooltip once some are hidden). The panel's
  Edit button becomes a **Surveillance Categories** action in the node menu. This saves some screen
real-estate.
- Node info dialog rows were cramped and now have vertical spacing.

## Shared code

- **`ui/src/lib/linkUtils.ts` (new)** — one place for creating links to the legacy
  JSP/htm pages. This used to be spread across six different call sites.
- **`ui/src/services/flowService.ts` (new)** — resolves one interface's flow graph URL.
- **`useDebouncedSearch` / `useDelayedLoading` (new hooks)** — shared by both tables.
- **`OnmsTable` gains a `rowClass` prop** (`NMS-20081`), with a seam contract test. Initially
I was going to use this for giving SNMP interfaces items a background color, but we're using badges
instead. However, seems like `rowClass` could be a useful feature, so left it in.

---

## Things worth a careful look

**1. Both interface tables stop being `lazy`.**

Both tables fetch every interface for the node in one request
(`limit=0`, which the v2 API documents as no limit), which is also what the legacy AngularJS
interfaces page did. Paging is client-side over the filtered, sorted set.

In most use cases, there are probably well under 100 interfaces for a node, in which case fetching
all of them at once and use client side sorting / filtering / paging is extremely fast.

There is some valid concern for things like a chassis switch which might have 10,000+ interfaces
(many of them virtual).

We measured rather than assume the performance implications.
With 20,000 synthetic rows in the store: sorting a column
takes ~210 ms, jumping to the last page ~13 ms, and only the visible page is ever in the DOM.
Sorting is the one interaction with a perceptible cost, and it arrives well beyond the scale of a
real node. What does not scale is the payload — 5,000 rows is ~2.35 MB, about 1.9 s on a 10 Mbit
link — which is what the deferred fetch below is for.

We should consider having an IP/SNMP details page showing all interfaces for a node, in which case we could
consider server-side fetch to handle very large numbers of interfaces. However, we should make
sure that there really are in fact performance issues with fetching / sorting 20,000+ interfaces.

Again, the legacy AngularJS did non-lazy fetch, and client side processing.

**2. Sorting is `Intl.Collator(undefined, { numeric: true })`, and that is not ours to change.**

Both tables sort using PrimeVue's own comparator. `DataTable.sortSingle`
(`primevue/datatable/index.mjs:4668`) builds it with `localeComparator()` from
`@primeuix/utils/object`, which is exactly:

```js
new Intl.Collator(undefined, { numeric: true }).compare
```

That one line explains two behaviors these tables are tested for:

- **Case-insensitive ordering** on every text column (`alpha, Alpha, beta`) — the collator's
  default strength, not anything this PR added.
- **`10.0.0.2` sorting before `10.0.0.10`** — `numeric: true` reads runs of digits inside a
  string as numbers, the way a file manager puts `file2` before `file10`.

**No IP address is being parsed.** These stay plain string comparisons; the collator just happens
to read each dotted group as a number, which for IPv4 coincides with address order. The
coincidence breaks on IPv6, where the groups are hex: `fe80::10` is *sixteen* as an address but
*ten* to the collator, so it sorts before `fe80::a` instead of after it. `IpInterfacesTable.test.ts`
pins that deliberately, with a comment saying that if address-aware sorting is ever added the test
should fail and be updated rather than worked around.

PrimeVue 4 has no `sortFunction`/`customSort` prop, so the comparator cannot be replaced. Changing
the semantics of a column means sorting on a decorated field instead of the raw one — which is
what Speed (sorts on the raw `ifSpeed`, displays a formatted label), Status and Flows already do.

**3. The IP interfaces fetch drops an `_s` term that was wrong (bug fix).**

The panel asked for `isManaged==U,isManaged==P,isManaged==N,isManaged==M`. Two things wrong with
that list: `P` is an `isSnmpPrimary` code (Polled), not an interface status, so it matched nothing;
and `F` was missing, which is the state the admin *Unmanage* action writes
(`ManageNodesServlet.unmanageInterfaces`). **Every interface an operator had explicitly unmanaged
was absent from this table**, with no indication any row was being withheld. The term dates to
2021.

The fetch now asks for every interface and the Status column says which is which, matching what
the legacy page did.

**4. Flow graph URLs are resolved on click, not up front (deliberate divergence from legacy).**

`rest/flows/flowGraphUrl` takes one `exporterNode` + `ifIndex` at a time, and
`FlowRestServiceImpl.getFlowGraphUrlInfo` runs `getFlowCount()` against the flow repository for
each call to populate a field the table never reads. The legacy page loops over every
flow-carrying interface as the table loads, so a switch with a thousand of them pays a thousand
flow queries for links nobody clicked.

Here the request happens on the click. Nothing is retained afterwards, which is why this is a
service call straight from the component rather than a store action and a map.

Whether an interface carries flows needs none of that — `hasIngressFlows` / `hasEgressFlows` are
already on the row — so the tags, their tooltips and the column's sort key all derive from the
row, and sorting groups the flow-carrying interfaces together whether or not any URL has ever
been resolved.

Two details in the click handler that look odd and are not:

- The tab is opened **before** the `await`, and pointed at the URL when it resolves. A tab opened
  after the await is blocked, the user gesture having expired.
- It is opened **without** `noopener`, which would return `null` instead of the handle needed to
  navigate it. `tab.opener` is nulled separately, while the tab is still `about:blank` and
  same-origin.

Three failure modes each say so rather than leaving a blank tab: no URL configured (`flowGraphUrl`
is empty by default and the endpoint then answers `204` with no body — confirmed against a running
instance), the lookup failed, or the browser blocked even the blank tab.

**5. Admin-only node actions are now gated (behavior change).**

These 5 actions are admin-only (need `ROLE_ADMIN`) on the server side, so we are also gating them here.

| Action | Destination | Rule |
| --- | --- | --- |
| Surveillance Categories | `admin/categories.htm` | `/admin/**` |
| Node Rescan | `element/rescan.jsp` | its own `intercept-url` |
| Admin / Node Management | `admin/nodemanagement/index.jsp` | `/admin/**` |
| Update SNMP Information | `admin/updateSnmp.jsp` | `/admin/**` |
| Schedule an Outage | `admin/sched-outages/editoutage.jsp` | `/admin/**` |

All five are gated together in legacy `element/node.jsp` (`<c:if test="${model.admin}">`).

Two findings behind this worth confirming:

- **`rescan` is admin-only** even though it isn't under `admin/` —
  `applicationContext-spring-security.xml` pins `/element/rescan.jsp` to `ROLE_ADMIN` on its own
  line. Easy to miss.
- **`updateSnmp` is plain `ROLE_ADMIN`**, *not* `useRole`'s `snmpRole` (which also admits
  `ROLE_PROVISION`) — `admin/updateSnmp.jsp` falls under `/admin/**` like the rest, so a
  provisioner does not get it.

`mapLink` refuses an admin-only link whenever `isAdmin` is absent, so a caller that forgets the
context **fails closed**. `NodeActionLinkContext` carries a boolean rather than a role list, so
the role vocabulary stays in `useRole`. Gating applies to the node list's menu too, since the link
list is shared.

If gating is needed based on things other than just `ROLE_ADMIN`, then we can revisit having just
the `isAdmin` boolean flag.

**6. Two deliberate divergences from the legacy labels.**

*SNMP status* — the JSP had three states and folded every non-up `ifAdminStatus` into unknown.
`DISABLED` and `TESTING` are pulled out of that: an interface that is administratively down was
deliberately turned off, and one under test is mid-change; neither is unknown and neither is a
fault, hence `warn` and `info` rather than `danger`. `UNKNOWN` now means what it says — no usable
`ifAdminStatus`. `ifAdminStatus` is tested first, since it says what the interface is being *asked*
to do.

We are using `DISABLED` rather than `ADMIN-DOWN`, partly to be more concise, partly as it
keep column width a bit narrower, partly because easier to filter on (different from any other status).

*IP interface status* — labels come from `ElementUtil.m_interfaceStatusMap`, the only place in
the product that turns these codes into words. `A` is in our map too: it is the `ifservices` code
for Managed and should not be used on `ipinterface`, but three admin servlets query for it, so it
reads `Unknown (A)` rather than looking like a bug in this table. We could remove it if we are sure this
is an invalid status (may have been a copy-and-paste error). Anything else reads
`Unknown (X)`, and a missing code reads `Unknown` — the column is an unconstrained `char(1)` with
no enum and no check constraint behind it, so naming an undefined code beats the JSP's blank cell.

**7. A global tooltip change with a side effect outside this feature.**

PrimeVue renders tooltip text in a plain div, so `\n` collapses to a space. `primevue-overrides.scss`
now sets `.p-tooltip-text { white-space: pre-line }` (it can't be scoped — tooltips teleport to
`<body>`). This also affects the validation-error tooltip in `EventConfigUploadFilesTab`, which
already joined its errors with `\n` and now breaks one per line. That looks like the original
intent, but it is a change nobody asked for.

**8. One IPv6 bug fixed by the link consolidation.**

`NodesTable` built its interface link without encoding the address, so a zone index like
`fe80::1%eth0` reached the legacy page corrupted. Fixed and tests added.

**9. Interface fetches are deferred until their tab is shown.**

PrimeVue 4 has no lazy `TabPanel`, so both tables mount at once and both used to fetch on mount.
Every node details load made an SNMP interface on the node — an expensive request on a
large switch — even for a user who never opened that tab. Each table now fetches when its tab is
first shown, and again if the node changes while it is showing.

Together with the IP interfaces table filling `nodeToIpInterfaceMap` from its own fetch (the page
used to issue a second, near-identical `limit=0` request purely to feed the info dialog), this
takes a node details load from **four interface requests to two**.

**10. The tables clear their rows when a fetch starts.**

The details page keeps one table instance across node changes, so rows left standing during a
fetch belong to the previous node — and an `ifIndex` collides across nodes, so a link built from
the live route id over a stale row goes somewhere real and wrong rather than 404ing. Clearing
first is also why the spinner needs two pieces of state: one for the delayed spinner, one to
suppress the empty state for the whole fetch, so the table never claims *No results found.* while
it is still waiting.

**11. Column widths are driven by headers, not values.**

Omitting the details, but IP/SNMP Interface headers were adjusted to fit better in smaller screen widths.

**Note** that we may end up putting `Node Interfaces` on a single row instead of in just the
left column. Leaving this for another layout PR after other work is done and we do a UX/UI review.

---

## Known issue, deliberately left

The **Availability** panel's graph overflows its card — the timeline bars and the percentage label
run past the right edge. This is pre-existing (`NodeAvailabilityGraph` has a hard-coded
`TIMELINE_WIDTH = 400` that doesn't respond to its container) and its column width is unchanged by
this PR, but moving the panel up makes it more prominent. We are revamping this panel in an
upcoming PR, so not worth fixing now. Please ignore this panel for purposes of reviewing this PR.

---

### Node Details - Top

node-details-top

### Node Details - IP Interfaces

node-details-ip-interfaces

### Node Details - SNMP Interfaces

node-details-snmp

### Node Details - Links Dropdown

node-details-links

### Node Details - Info Dialog

node-details-info-dialog

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20301


ui/src/components/Nodes/SnmpInterfacesTable.vue
```diff
@@ -1,79 +1,187 @@
...
+  UP: 'success',
+  DOWN: 'danger',
+  UNKNOWN: 'info'
+}
+
+// Every interface for the node in one request (limit 0 is "no limit"), the way the legacy
+// interfaces page fetched them. Sorting and filtering are client-side, and neither can be
+// correct over a single server page: filtering would only ever search the rows already on
+// screen, so a term matching something on page 3 would read as "no results".
+const queryParameters = { limit: 0 }
```
joseanesONMS, line 157: A chassis or core switch can carry thousands of SNMP interfaces, and every visit now pulls all of them and decorates each row. Fine for the stated assumption; a comment here or a soft cap with a "showing the first N" note would make the trade-off visible when a big node shows up.

ui/src/components/Nodes/SnmpInterfacesTable.vue
```diff
@@ -1,79 +1,187 @@
...
-const queryParameters = ref({
-  limit: DEFAULT_PAGE_SIZE,
-  offset: 0
-})
+const { searchTerm, appliedTerm, onSearch, clearSearch } = useDebouncedSearch()
+
+const statusSeverity: Record<SnmpInterfaceStatus, OnmsTagSeverity> = {
+  UP: 'success',
+  DOWN: 'danger',
+  UNKNOWN: 'info'
```
joseanesONMS, line 150: An administratively-down interface renders as a blue informational tag, the same colour as ordinary info badges. `secondary` or `warn` would set it apart. Cosmetic.

ui/packages/onms-ui/src/components/OnmsTable.vue
```diff
@@ -87,6 +88,9 @@ const props = withDefaults(defineProps<{
   // PrimeVue accepts either an array of row instances or an object keyed by
   // dataKey (its DataTableExpandedRows shape) when a table uses dataKey.
   expandedRows?: any[] | Record<string, boolean>
+  // Per-row CSS classes, as PrimeVue's own (data) => class signature. Returns a
+  // class string or a Vue class object; undefined leaves the row unclassed.
+  rowClass?: (data: any) => string | Record<string, boolean> | undefined
```
joseanesONMS, line 94: The description says the badge replaced the only use of `rowClass`, so this lands as seam surface with no consumer and a contract test to maintain. Consider dropping it here and adding it with its first user.

ui/src/components/Nodes/IpInterfacesTable.vue
```diff
@@ -1,77 +1,132 @@
...
+          <span v-else>N/A</span>
+        </template>
+      </OnmsColumn>
+      <OnmsColumn style="width: 24%" field="hostName" header="IP Host Name" sortable>
+        <template #body="{ data }">{{ data.hostName || 'N/A' }}</template>
+      </OnmsColumn>
+      <OnmsColumn style="width: 22%" field="ifIndex" header="SNMP ifIndex" sortable>
+        <template #body="{ data }">{{ data.ifIndex || 'N/A' }}</template>
+      </OnmsColumn>
+      <OnmsColumn style="width: 26%" field="isManaged" header="Managed" sortable>
```
joseanesONMS: The filter matches displayed text and this column shows single letters, so typing `m` matches every managed row. Rendering M/U/N/P as words would fix both readability and filtering. Pre-existing rendering, only worth it because filtering is new.

ui/tests/components/Nodes/SnmpInterfacesTable.test.ts
```diff
@@ -194,9 +497,27 @@ describe('SnmpInterfacesTable.vue', () => {
...
       })
       ;(useRoute() as any).params.id = '42'
     })
+
+    // The old node's search term says nothing about the new node's interfaces.
+    it('clears the filter', async () => {
+      nodeStore.snmpInterfaces = [{ id: 1, ifIndex: 1, ifName: 'eth0' }] as any
+      await nextTick()
+      await wrapper.find('[data-test="snmp-interfaces-search"]').setValue('nothing-matches')
+      await new Promise(r => setTimeout(r, SEARCH_DEBOUNCE_MS + 25))
```
joseanesONMS: This case sleeps on a real timer while the neighbouring filter tests use fake timers; `vi.useFakeTimers()` + `advanceTimersByTime` would keep it deterministic and fast.

joseanesONMS (commented):
Read through the whole diff and ran the Nodes, lib and onms-ui suites on a merge with develop (1429 tests, lint and typed build green). Checked on a live node that `limit=0` returns every interface, and that the admin gating matches the security context. Nothing blocking; the inline notes are suggestions only.

ui/tests/containers/NodeDetails.test.ts
```diff
@@ -137,21 +134,68 @@ describe('NodeDetails.vue', () => {
...
+
+      expect(wrapper.findComponent({ name: 'NodeDetailsDialog' }).props('visible')).toBe(false)
+    })
+
+    // The dialog picks the node's best IP from nodeToIpInterfaceMap, which only the node list
+    // used to fill -- so without this fetch its IP Address row is blank on this page.
+    it('fetches the IP interfaces the dialog needs to show an address', async () => {
+      const { wrapper, nodeStore } = mountComponent('42', true)
+      await flushPromises()
+      void wrapper
```
joseanesONMS: `void wrapper` only exists to dodge the unused-variable lint; destructuring just `nodeStore` avoids it.

ui/tests/components/Nodes/IpInterfacesTable.test.ts
```diff
@@ -149,38 +208,144 @@ describe('IpInterfacesTable.vue', () => {
...
+      nodeStore.ipInterfaces = [
+        { id: '1', ipAddress: '10.0.0.7', hostName: 'beta', ifIndex: 3, isManaged: 'M' },
+        { id: '2', ipAddress: '10.0.0.2', hostName: 'Alpha', ifIndex: 1, isManaged: 'U' },
+        { id: '3', ipAddress: '10.0.0.5', hostName: 'charlie', ifIndex: 2, isManaged: 'M' }
+      ] as any
+      nodeStore.ipInterfacesTotalCount = 3
+      await nextTick()
+    })
+
+    it('defaults to IP address ascending', () => {
```
joseanesONMS, line 232: Worth one case with `10.0.0.10` and `10.0.0.2`: the ascending order here depends on PrimeVue's numeric collator, and a lexical sort would pass the current fixtures too.

ui/src/styles/primevue-overrides.scss
```diff
@@ -42,6 +42,15 @@
   text-transform: uppercase;
 }
 
+// Let a tooltip whose text carries newlines render as multiple lines. PrimeVue
+// puts the text in a plain div, so '\n' otherwise collapses to a space and a
+// label/value tooltip runs together on one line. `pre-line` rather than `pre`
+// so ordinary single-line tooltips still wrap normally at the max width.
+// Tooltips teleport to <body>, so this cannot live in a scoped block.
+.p-tooltip-text {
```
joseanesONMS, line 50: No objection, and the event-config upload tooltip getting real line breaks is an improvement. Worth a line in the release notes in case any other tooltip embeds literal newlines on purpose.

ui/src/components/Nodes/SnmpInterfacesTable.vue
synqotik: Updated this logic a bit. Added `TESTING` and `DISABLED` status types.

`DISABLED` is when `ifAdminStatus` is `DOWN`, and it's displayed as `warn`. 

`TESTING` is when either `ifAdminStatus` is `TESTING`; or `ifAdminStatus` is `UP` and `ifOperStatus` is `TESTING`. This is an `info`.

This goes a bit more fine-grained than the JSP we are replacing, but more accurate and helpful.

ui/src/components/Nodes/SnmpInterfacesTable.vue
synqotik: I delved into this a bit, had some perf testing done, etc. Decided to keep client-side sorting/filtering and keep fetching "all at once" for now. For the vast majority of cases, this will be a better user experience since for anything under say 2000 items, client-side sorting/filtering is very fast, way faster than having to make network calls for each page of data, and the total JSON payload isn't too much.

For cases over 10,000 items (chassis/core switches), this will result in a noticeable delay of several seconds to get a larger 10 MB+ payload, parse it, then filter etc. We made a couple of changes to lessen the impact, plus added a spinner for the SNMP interfaces table. We can see in practice if this is an issue.

I did some testing simulating 20,000 items. Client side sorting / filtering / pagination was very fast. Main issue is just the extra payload and parsing it which could take up a few seconds and of course memory footprint in browser.

Probably what we really need is a dedicated SNMP interfaces page with a server-side paginated/etc. table, and we can filter by node and link to it from Node Details.

I didn't want a mix of client and server side processing on the Node Details page.

ui/src/components/Nodes/IpInterfacesTable.vue
synqotik: Revamping this to display full status, so this will be fixed. Also fixed a bug where we weren't querying for correct statuses, and now we get all IP interfaces and map them to labels and severity colors, plus we handle any unknown status codes. User can now type e.g. `managed` to get just managed interfaces.

ui/tests/components/Nodes/SnmpInterfacesTable.test.ts
synqotik: Fixed (note it's now on different lines).

ui/tests/containers/NodeDetails.test.ts
synqotik: Fixed (note, may be on a different line now).

ui/src/styles/primevue-overrides.scss
synqotik: Should be ok as-is. There will be a line in the PR description about this.

ui/tests/components/Nodes/IpInterfacesTable.test.ts
synqotik: Done. Also will add a note that PrimeVue sorts lexically but treats runs of digits inside a string as a number and sorts them numerically. For IPv4, this ends up working the way we'd prefer. For IPv6, it won't handle hex, etc., so those are more likely to be purely lexical. Added tests and notes for both IPv4 and IPv6 situations.

ui/packages/onms-ui/src/components/OnmsTable.vue
synqotik: Leaving as-is for now.

marshallmassengill (approved):
I think this is good to go.  Some minor stuff that you may want to know about and deal with in the future:

  - A failed interface fetch is indistinguishable from an empty node and never retried: the service returns false on error, the store leaves the rows cleared, and fetchedNodeId is already set, so re-showing the tab does nothing until the node id changes. Resetting fetchedNodeId when the store call yields nothing would fix it. SnmpInterfacesTable.vue:239, IpInterfacesTable.vue:142.
  - getNodeIpInterfaces replaces the shared nodeToIpInterfaceMap wholesale. Safe today because the node list rebuilds it on every load and its readers tolerate a missing entry. Worth a sentence in the store comment, no code change. nodeStore.ts:204.
  - IP status codes live in two maps that must stay in step: labels in utils.ts:207, severities in IpInterfacesTable.vue:70. Nit.
  - Agree with the existing reviewer that rowClass lands with no consumer. Author chose to keep it. Not blocking.

synqotik:
> I think this is good to go. Some minor stuff that you may want to know about and deal with in the future:
> 

Thanks, I'll look at those as part of the next PR.

PR #8876, opened 2026-09-17
NMS-19837: Create UI to configure threshd and thresholds

christianpape:
* JIRA: https://opennms.atlassian.net/browse/NMS-19837
* JIRA: https://opennms.atlassian.net/browse/NMS-19838


opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/ThreshdConfigRestService.java
```diff
@@ -0,0 +1,485 @@
...
+                    failure.set(Response.status(Status.CONFLICT)
+                            .entity("Threshd package '" + packageName + "' already exists.").build());
+                    return;
+                }
+
+                final List<Package> packages = new ArrayList<>(config.getPackages());
+                packages.add(ThreshdConfigMapper.toEntity(pkg));
+                config.setPackages(packages);
+
+                threshdDao.saveConfig();
```
cgorantla, line 209: this updates DAO's live dbConfig before saveConfig() runs.
If cms.updateConfiguration throws, nothing is rolled back.

This may need update at Dao layer.

opennms-config-model/src/main/resources/xsds/thresholding.xsd
```diff
@@ -26,50 +31,7 @@
...
         </element>
-
-        <element maxOccurs="unbounded" minOccurs="0" ref="this:thresholder">
-          <annotation>
-            <documentation>Service thresholders</documentation>
-          </annotation>
-        </element>
       </sequence>
-
-      <attribute name="threads" type="int" use="required">
```
cgorantla, line 37: Keep the attribute in the XSD as optional and ignored (<attribute name="threads" type="int" use="optional"/>, documented as unused). Old files then pass validation, and the Java model can stay without the field. The same goes for thresholder. 

This will allow  users to upload their updated threshold config without taking the change we are forcing.

cgorantla (commented):
I'm still reviewing this, updating couple of important ones.

opennms-webapp-rest/src/main/java/org/opennms/web/rest/v2/ThreshdConfigRestService.java
christianpape: done.

opennms-config-model/src/main/resources/xsds/thresholding.xsd
christianpape: done.

PR #8878, opened 2026-09-17
NMS-20335: Fix missing chunk for Sink messages with large buffer sizes

cgorantla:
subtle bug when chunkNum is zero.
Test was not taking effect as config setting was not right.
Fixed test as well.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20335


marshallmassengill (dismissed):
I think this is good... if I understand it correctly.

Some things we may want to address but not blocking:

- max.buffer.size of 0 or negative is not rejected. Zero makes IntMath.divide throw on every dispatch; negative yields a non-positive chunk count, the send loop never runs, and every message is dropped with no error. One line: fall back to the default when the parsed value is not positive. KafkaRemoteMessageDispatcherFactory.java:273.
- A dropped out-of-order chunk logs at debug. With the new strict ordering, a genuinely lost chunk now stalls that message until cache expiry with nothing visible. A rate-limited warn or a counter would help the next person.
- The test asserts only UEI and node id. Asserting the 5 KB description and log message round-trip would also catch a mis-ordered concatenation.
- The "ConfigAdmin may hand us an Integer" comment is true for programmatic config like the IT, but .cfg files deliver Strings, so production was already honoring the setting. Nit.

marshallmassengill (approved):
LGTM!

christianpape:
Problems found

1. Silent restart of accumulation mid-message — KafkaMessageConsumerManager.java:201 
byteString == null ? content : concat treats a missing buffer as "start fresh". largeMessageCache and currentChunkCache are two independent Guava caches (maximumSize=1000, expireAfterWrite=10m), and the content entry is always written first, so it always expires first. If it lapses while the counter survives, accumulation restarts at chunk k, receivedChunks == totalChunks still fires, and a truncated payload reaches module.unmarshal — either an exception swallowed by the catch (RuntimeException) at line 232, or a partially populated message dispatched downstream. Guard: if expectedChunk > 0 && byteString == null, invalidate both and skip.

3. Lost-update race on the shared caches — same line
Both caches are fields of the outer manager (lines 109–110), shared across every KafkaConsumerRunner thread, and getIfPresent → put is a non-atomic read-modify-write spanning two caches. The producer's partition-change retry deliberately puts chunks of one messageId on two partitions; with getNumConsumerThreads() > 1 those are polled concurrently. Two threads read the same ByteString, both put, one chunk's bytes vanish while the counter advances → corrupt dispatch. Holding {expectedChunk, bytes} in a single cache entry updated via asMap().compute(...) fixes this and number 1 together.

4. Newly reachable duplicate dispatch — KafkaRemoteMessageDispatcherFactory.java:109
sendMessageChunkToKafka delivers the chunk before the partition comparison at lines 141–146, so a partition change on the final chunk means that chunk is already on the new partition. With the consumer now fixed, the message completes and the cache entries are invalidated; dispatch's do { } while (partitionChanged) then re-sends every chunk under the same messageId, and the consumer reassembles and dispatches the same event/flow twice. This is new behavior — before the PR the first attempt could never complete.

cgorantla:
@christianpape Handled these edge cases at the producer end. Removed partition-change retry loop. Now it pins all chunks to same partition.

marshallmassengill (changes requested):
- The cross-thread race is narrowed, not closed. A group rebalance while a multi-chunk message is in flight can hand the partition to another runner in the same JVM while the first is still processing its batch, and the two caches are still updated by separate non-atomic get-then-put calls. Rare, and the other reviewer's single-entry compute() would remove it; worth a follow-up ticket rather than a hold.
  - After the eviction guard drops a message, every remaining chunk of it logs "Missing chunk ... will be dropped" at warn, one line per chunk. I think this is the real thing worth fixing here.

cgorantla:
> * The cross-thread race is narrowed, not closed. A group rebalance while a multi-chunk message is in flight can hand the partition to another runner in the same JVM while the first is still processing its batch, and the two caches are still updated by separate non-atomic get-then-put calls. Rare, and the other reviewer's single-entry compute() would remove it; worth a follow-up ticket rather than a hold.
> * After the eviction guard drops a message, every remaining chunk of it logs "Missing chunk ... will be dropped" at warn, one line per chunk. I think this is the real thing worth fixing here.

I think we are over optimizing here.  Group rebalance is a rare event and even in that case, the partition may get switched but unlikely that two threads polling from the same partition unless there is a stuck consumer which is rare.

 "Missing chunk ... will be dropped"  - Unlikely this is a problem, having couple of extra log lines seems to be fine. We are not dealing with 100MB sink messages.  We are probably dealing with 1-5 MB  buffers.  This is also very rare or mostly theoretical  for Sink messages that's why no one noticed this bug.

marshallmassengill (approved):
Ship it!

PR #8880, opened 2026-09-17
NMS-20334: Minion/Sentinel ignore JAVA_MIN_MEM and JAVA_MAX_MEM from /etc/sysconfig/minion

mershad-manesh:
### All Contributors

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20334


container/shared/src/main/resources/etc/container.init
```diff
@@ -49,6 +49,11 @@ fi
 EXTRA_JAVA_OPTS="$JAVA_OPTS $EXTRA_JAVA_OPTS"
 unset JAVA_OPTS
 
+# turn JAVA_MIN_MEM/JAVA_MAX_MEM into -Xms/-Xmx; prepended, so an explicit
```
cgorantla: These two lines run before COMMAND is set, so the -Xms/-Xmx flags end up in EXTRA_JAVA_OPTS for every verb. Karaf's setupDefaults appends EXTRA_JAVA_OPTS to JAVA_OPTS unconditionally and only the stop branch clears JAVA_OPTS, so `minion status` and `bin/client` now launch their small helper JVMs with the full server heap. On a memory-tight host with the server already holding its heap, status can fail to reserve memory and report the service as stopped.

Suggest moving the two lines into the start) branch, just before "$CONTAINER_HOME"/bin/start. EXTRA_JAVA_OPTS is already exported, so the reassignment still reaches the child, and restart / try-restart / force-reload all go through `$0 start`

cgorantla (commented):
Also if possible, merge all commits into one commit with a proper commit message.

container/shared/src/main/resources/etc/container.init
mershad-manesh: Great suggestion, I'll fix it

cgorantla (commented):
I think this is broken for a while. 
May need re-target this to foundation-2024

cgorantla (approved):
LGTM!

PR #8882, opened 2026-09-18
NMS-20341: Improve files we store under evidence folder for flaky tests

mershad-manesh:
### All Contributors

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20341


cgorantla (commented):
1. Change */target/logs/*${SIMPLE_CLASS}* to */target/logs/${TEST_CLASS}-* and match only the test-level directory. That stops ThresholdingIT from grabbing SentinelThresholdingIT logs, and stops each container folder from being copied twice. 
2. Keep container logs out of ~/test-results. That path is what store_test_results parses as JUnit output. Copy the non-XML evidence to a sibling folder and store only that as an artifact.

cgorantla (commented):
Could we also copy `target/logs/startup-failures` into `${FLAKY_EVIDENCE_DIR}` after the per-class loop?

This will give us logs when containers doesn't start properly

mershad-manesh:
> Could we also copy `target/logs/startup-failures` into `${FLAKY_EVIDENCE_DIR}` after the per-class loop?
> 
> This will give us logs when containers doesn't start properly

good idea, but we should copy it per attempt. 
If we want to make that file to be per-class, then we need to change how we create that file.

PR #8883, opened 2026-09-18
NMS-20336: Optimize OpenConfigClient to remove bottlenecks

cgorantla:
Restructured the client around asynchronous readiness (notifyWhenStateChanged) with a connection timeout to resolve twin config bottlenecks when dealing with 1000s of devices. 
This also fixes the leaked ManagedChannel on retry that produced ManagedChannelOrphanWrapper errors under heap pressure. retries defaults to 0 (unlimited), interval has a 1 s floor; a missing port fails once instead of retrying.

### External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20336
* Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20338


indigo423:
@cgorantla thanks for taking a stab on these two issues. I'm working right now to see if this is fixed in my lab environment.

indigo423 (commented):
I've tested your patch in my lab deployment and was successful. Establishing the gRPC sessions took now 68 seconds for ~10k devices rather than 33.3 hours for 7.5k :)

```
┌────────────────────────────────────────┬──────────────────┬────────────┐
│                                        │ 36.0.4 unpatched │  PR 8883   │
├────────────────────────────────────────┼──────────────────┼────────────┤
│ Sessions restored after Minion restart │ 7,500            │ 10,748     │
├────────────────────────────────────────┼──────────────────┼────────────┤
│ Wall clock                             │ 33.3 hours       │ 68 seconds │
├────────────────────────────────────────┼──────────────────┼────────────┤
│ Per device                             │ 16.0 s           │ 0.006 s    │
└────────────────────────────────────────┴──────────────────┴────────────┘
```

## Here is how I've tested it

The changes where in `features/openconfig/telemetry-client`, so I've cherry-picked a1719beb onto the `opennms-36.0.4-1` tag and hot-swapped that single OSGi bundle into a running Minion. The cherry-pick applied cleanly with no conflicts, and all 34 tests in the PR pass when built against 36.0.4.

Environment: Core 16 vCPU / 48 GiB, one Minion 4 vCPU / 16 GiB with a 10 GB heap, Kafka IPC, 10 748 simulated devices, gNMI on port 9339, 300 s sample interval, subscribed to `/interfaces/interface/state/counters`.

## Result: subscription restore after a Minion restart

| | 36.0.4 | this PR |
|---|---|---|
| Sessions restored | 7,500 | 10,748 |
| Wall clock | 33.3 hours | 68 seconds |
| Per device | 16.0 s | 0.006 s |

## This also fixes NMS-20338

The commit references only NMS-20336, but `retry()` now calls `close(current)`, which shuts down the channel a failed attempt leaves behind. That is precisely the leak described in NMS-20338.

## Caution 1: the parallel dial burst is hard on the targets

With this PR the simultaneous connect burst opens a lot of gRPC connections not on parallel. This could be an issue if gNMI devices are not in the same network and need to go through routing or firewall hops.

There is no concurrency limit and no ramp. `CONNECTION_TIMEOUT_MILLIS` is hardcoded at 5000 and the pool is `Math.max(2, Runtime.getRuntime().availableProcessors())`. A configurable cap on in-flight connection attempts would make this safer to roll out.'

In my opinion this is now a nice to have and not super urgent. We don't have a requirement for this topic yet and this is not a blocker.

indigo423 (approved):
Approved from the real world testing perspective :)

christianpape:
LGTM. Two things you can look at:

Unbounded zero-delay reconnect loop — OpenConfigClientImpl.java:349. retry() schedules with delay 0 whenever current.received is true, and accept() (:403-405) also resets remainingRetries. Since received is per-Attempt, every attempt that delivers even one message earns another zero-delay reconnect and a fresh budget. A flapping device (accepts subscription, emits one sample, drops) puts the client into a permanent full-speed reconnect storm — TLS handshake + LOG.warn + handler.onError per cycle, forever, regardless of configured retries. The old code always waited interval (default 300s). The comment at :340-342 promises "one immediate reconnect" but nothing enforces the "one".
Retry budget resets only on received data, not on successful subscribe — :405. The old scheduleSubscription() recomputed a fresh local retries per invocation and reset on a successful subscribe. Now a device that connects and subscribes fine but whose stream dies before any sample arrives (sparse telemetry, server idle timeout, frequency > server stream timeout) burns one retry per episode. With retries=3, after the third drop the client logs "Giving up", cancels everything and goes permanently dark — and the connector is never told. Reset remainingRetries when subscribed is set, not just on first data.

christianpape (approved):
LGTM.

marshallmassengill (approved):
I think this is good.  The issues @christianpape found above I also found but I'm not sure how real they are to be honest.  I don't see any issues and this moves us forward for more scale testing.

cgorantla:
> LGTM. Two things you can look at:
> 
> Unbounded zero-delay reconnect loop — OpenConfigClientImpl.java:349. retry() schedules with delay 0 whenever current.received is true, and accept() (:403-405) also resets remainingRetries. Since received is per-Attempt, every attempt that delivers even one message earns another zero-delay reconnect and a fresh budget. A flapping device (accepts subscription, emits one sample, drops) puts the client into a permanent full-speed reconnect storm — TLS handshake + LOG.warn + handler.onError per cycle, forever, regardless of configured retries. The old code always waited interval (default 300s). The comment at :340-342 promises "one immediate reconnect" but nothing enforces the "one". Retry budget resets only on received data, not on successful subscribe — :405. The old scheduleSubscription() recomputed a fresh local retries per invocation and reset on a successful subscribe. Now a device that connects and subscribes fine but whose stream dies before any sample arrives (sparse telemetry, server idle timeout, frequency > server stream timeout) burns one retry per episode. With retries=3, after the third drop the client logs "Giving up", cancels everything and goes permanently dark — and the connector is never told. Reset remainingRetries when subscribed is set, not just on first data.

Keeping this as is for now.  As current code uses STREAM_MODE, we may not hit these.

PR #8884, opened 2026-09-18
NMS-20343: Windows event log support over WS-Man

joseanesONMS:
Adds WsManEventLogd, a daemon that reads the Windows event logs of monitored hosts over WS-Man and creates one OpenNMS event per record, so Windows logs get the same treatment syslog already has.

The read runs on the Minion at the node's location through a new RPC module, keeps a per-node, per-log cursor in the key-value store, and maps records to events through a configurable Event ID to UEI mapping.

- New `wsman-eventlog-configuration.xml` with packages (filter, logs, mappings), shipped with System and Application on and Security off with curated audit Event IDs.
- Event Logs tab on Manage WS-Man: packages with filter preview, logs, mappings with an inline event definition editor, and per-node read status.
- REST under `wsman-config/event-log` for config, filter preview, status and definitions; definitions are written only to the daemon's own eventconf source.
- WS-Man and event log metrics collected through JMX with eight resource graphs.
- The Get-WinEvent read mode for Applications and Services logs is implemented but held back, since it has not been tested against a Windows host.

## External References

- Jira (Issue Tracker): https://opennms.atlassian.net/browse/NMS-20343


joseanesONMS:
2026-09-18 WS Man Log Configure What is Read
2026-09-18 WS Man Log Event Mappings
2026-09-18 WS Man Log Choose Logs
2026-09-18 WS Man Log Sample Events


Further discussion is necessary.  We wil need to figure out if we want to poll (as in with this daemon), or subscribe (as in telemetryd), or both. @marshallmassengill @dino2gnt @cgorantla

marshallmassengill:
I like this idea but there is a lot more to consider here because of how Windows does logging I think.  I think it comes down to these potential options:

1. WMI poll over WinRM. Win32_NTLogEvent via WQL. No host changes. Classic logs only.
2. PowerShell Get-WinEvent over WinRS. Native Event Log API, every channel, server-side XPath.
3. Windows Event Forwarding. Hosts push by Group Policy to a collector. Two ways to receive: a Windows Event Collector that OpenNMS then polls, or the Minion itself as the collector by implementing the WEF receiving side (WinRM HTTPS listener, Kerberos or cert auth, acknowledged delivery).
4. WS-Eventing subscribe from the Minion. Per-host subscription state on the Minion.  Maybe not well documented?

joseanesONMS:
Closing without merging. The work stays on jira/NMS-20343-wsman-eventlog-smoke (head 8882f836751, CI green); the design record with the options considered, the decisions and how the Windows hosts were simulated for testing is attached to NMS-20343.

